PR #73, opened 2023-01-04
ADR-100: Data Companion Push API

thanethomson:
:book: [Rendered](https://github.com/cometbft/cometbft/blob/thane/adr-082-data-companion-api/docs/architecture/adr-100-data-companion-push-api.md)

Original PR: https://github.com/tendermint/tendermint/pull/9437

Since then, I've considered another alternative design to this: a data companion _pull_ API, inspired by https://github.com/tendermint/tendermint/pull/9437#discussion_r1034037844 - see #82.

---

#### PR checklist

      documentation updates needed


docs/architecture/adr-082-data-companion-push-api.md
```diff
@@ -0,0 +1,440 @@
...
+## Alternative Approaches
+
+1. One clear alternative to this would be the approach outlined in [ADR
+   075][adr-075]. This approach:
+
+   1. Still leaves Tendermint responsible for maintaining a query interface and
+      event indexing functionality, increasing the long-term maintenance burden of,
+      and the possibility of feature sprawl in, that subsystem. To overcome this,
+      we could remove the query interface altogether and just always publish all
+      data.
```
ValarDragon, line 74: As a user of the SDK, I'm personally pretty unconcerned about a Tendermint API for this, and am more than happy to do this in the SDK.

Not sure how other ABCI users feel.

docs/architecture/adr-100-data-companion-push-api.md
```diff
@@ -0,0 +1,441 @@
...
+      data _and_ we would need to panic if the sliding window of events fills up.
+   3. Allows for multiple consumers. While desirable from typical web services,
+      Tendermint is a consensus engine and, as such, we should prioritize consensus
+      over providing similar guarantees to a scalable RPC interface (especially
+      since we know this does not scale, and if such a job could just be
+      outsourced). If we have a single consumer, we can easily automatically prune
+      data that was already sent. We could overcome this by limiting the API to a
+      single consumer and single event window.
+
+   Making the requisite modifications to ADR-075 basically converges on the
```
sergio-mena, line 90: I think it would be worth saying (probably later on, when we describe this ADR's solution) that ADR075 is somewhat orthogonal to what is proposed in this ADR, in the sense that a "candidate" companion could be (a slightly modified version of) what we have today as RPC, with ADR075 on top.

docs/architecture/adr-100-data-companion-push-api.md
```diff
@@ -0,0 +1,441 @@
...
+2. All or part of the following data _must_ be obtainable by the companion, and
+   as close to real-time as possible:
+   1. Committed block data
+   2. `FinalizeBlockResponse` data, but only for committed blocks
+
+3. The companion _must_ be able to establish the earliest height for which the
+   node has all of the requisite data.
+
+4. The API _must_ be (or be able to be) appropriately shielded from untrusted
+   consumers and abuse. Critical control facets of the API (e.g. those that
```
sergio-mena, line 145: Line 20 above says "single trusted external companion service". So I would clarify that the shielding we talk about here does not include a (trusted) companion, but unauthorized actors.

docs/architecture/adr-100-data-companion-push-api.md
```diff
@@ -0,0 +1,441 @@
...
+
+### Entity Relationships
+
+The following simple diagram shows the proposed relationships between
+Tendermint, a socket-based ABCI application, and the proposed data companion
+service.
+
+```
+     +----------+      +------------+      +----------------+
+     | ABCI App | <--- | Tendermint | ---> | Data Companion |
```
sergio-mena, line 176:
```suggestion
     | ABCI App | <--- |  CometBFT  | ---> | Data Companion |
```

docs/architecture/adr-100-data-companion-push-api.md
```diff
@@ -0,0 +1,441 @@
...
+    // field is optional, depending on the Tendermint node's configuration.
+    optional tendermint.abci.FinalizeBlockResponse finalize_block_response = 2;
+}
+
+// BlockCommittedResponse is either empty upon succes, or returns one or more
+// errors. Note that returning any errors here will cause Tendermint to crash.
+message BlockCommittedResponse {
+    // Zero or more errors occurred during the companion's processing of the
+    // request.
+    repeated Error errors = 1;
```
sergio-mena, line 234: Not sure about this, but I'd give it a thought about dropping "repeated" here, since we are crashing.

docs/architecture/adr-100-data-companion-push-api.md
```diff
@@ -0,0 +1,441 @@
...
+At present, this data is already stored on disk, and so practically we would
+need to implement some form of background pruning mechanism to remove the data
+we know has been shared with the companion service.
+
+In order to not allow disk storage to grow unbounded, _and_ ensure that we do
+not lose any critical data, we need to impose some kind of limit on the number
+of requests we will buffer before crashing the Tendermint node. In case there is
+manual intervention needed to restore the companion service, Tendermint should
+attempt to flush this buffer upon startup before continuing normal operations.
+
```
sergio-mena, line 291: Also, it would be wise IMO to define the use case where the companion is started for the first time, much _after_ CometBFT itself was started. Similarly to how state-sync works today, the companion should be aware that, in such a use case, it would have a "gap" in its history.
Later version may improve on this, but this simplistic use case should is as good as what we have today when state-sync is used.

docs/architecture/adr-100-data-companion-push-api.md
```diff
@@ -0,0 +1,441 @@
...
+# Address at which the gRPC companion service server is hosted. It is
+# recommended that this companion service be co-located at least within the same
+# data center as the Tendermint node to reduce the risk of network latencies
+# interfering in node operation.
+addr = "http://localhost:26659"
+
+# Use the experimental gzip compression call option when submitting data to the
+# server. See https://pkg.go.dev/google.golang.org/grpc#UseCompressor
+experimental_use_gzip = false
```
sergio-mena, line 312: I would keep this our of the first version.
I we collect numbers that justify its introduction, we can do it incrementally

sergio-mena: Ah OK, I see line 323. But, following that line, consider removing this option to keep the initial design lean

docs/architecture/adr-100-data-companion-push-api.md
```diff
@@ -0,0 +1,441 @@
...
+
+## Implications
+
+1. We will be able to mark the following RPC APIs as deprecated and schedule
+   them for removal in a future release:
+   1. The [WebSocket subscription API][websocket-api]
+   2. [`/tx_search`]
+   3. [`/block_search`]
+   4. [`/block_results`]
+
```
sergio-mena, line 355: How about 
`/block`
`block_byhash`
`blockchain`
`tx`

docs/architecture/adr-100-data-companion-push-api.md
```diff
@@ -0,0 +1,441 @@
...
+### Release Strategy
+
+As captured in the [requirements](#requirements), we should be able to release
+this as an additive, opt-in change, thereby not impacting existing APIs. This
+way we can evolve this API until such time that consumers are satisfied with
+removal of the old APIs mentioned in [implications](#implications).
+
+This also means we could potentially release this as a non-breaking change (i.e.
+in a patch release).
```
sergio-mena, line 376: If nobody (i.e., no user) is asking for this, I wouldn't promise anything along these lines.
It will already be delicate enough to release it in a smooth way in a new major release.

docs/architecture/adr-100-data-companion-push-api.md
sergio-mena: I see this is mentioned towards the end, but maybe a (forward) reference here would help the reader understand the potential "orthogonality" of the two solutions (ADR075 and ADR100)

sergio-mena (commented):
I didn't pinpoint all places to rename to CometBFT (please search-and-replace)

docs/architecture/adr-100-data-companion-push-api.md
```diff
@@ -0,0 +1,441 @@
...
+5. The node _must_ know, by way of signals from the companion, which heights'
+   associated data are safe to automatically prune.
+
+6. The API _must_ be opt-in. When off or not in use, it _should_ have no impact
+   on system performance.
+
+7. It _must_ not cause back-pressure into consensus.
+
+8. It _must_ not cause unbounded memory growth.
+
```
peterbourgon, line 159: As with #82, these requirements 7 and 8 (that the companion entity neither blocks consensus nor produces unbounded memory growth) are mutually exclusive with requirement 2 (that the companion entity guarantees that committed block data is available). Messages from the node to the companion API can always fail, and that failure must necessarily either block consensus (crash the node) or impact memory usage (add to a queue).

If the system model is two discrete components, one which performs consensus and another which captures and serves state data, then

thanethomson: I think there's a misunderstanding here regarding what I initially meant by "back-pressure". I was thinking of a specific class of back-pressure that occurs in the current (flawed) event bus system where slow WebSocket subscribers actually slow down consensus operations - sometimes to the point that the node falls out of sync with the rest of the network. I've clarified this wording now.

thanethomson:
I've addressed most of the comments on this ADR so far, and given the recent comments/feedback I realized there's a need to more clearly articulate why this work seems necessary in the first place _and_ the longer-term roadmap within which this fits toward more clearly separating application- and consensus-level concerns. See the updated context: https://github.com/cometbft/cometbft/blob/thane/adr-082-data-companion-api/docs/architecture/adr-100-data-companion-push-api.md#context

ADR-101 is a different approach with different tradeoffs, but its context is precisely the same as ADR-100. I'll work on updating ADR-101, incorporating everyone's feedback, next.

docs/architecture/adr-100-data-companion-push-api.md
```diff
@@ -0,0 +1,513 @@
...
+At present, this data is already stored on disk, and so practically we would
+need to implement some form of background pruning mechanism to remove the data
+we know has been shared with the companion service.
+
+In order to not allow disk storage to grow unbounded, _and_ ensure that we do
+not lose any critical data, we need to impose some kind of limit on the number
+of requests we will buffer before crashing the CometBFT node. In case there is
+manual intervention needed to restore the companion service, CometBFT should
+attempt to flush this buffer upon startup before continuing normal operations.
```
peterbourgon, line 332: Delivering a message to a buffer does not guarantee that it will be received by any specific consumer process. Storing data on disk does not provide any delivery guarantees. If reliable delivery is a requirement, then data cannot be buffered, neither on disk nor in memory; it must be synchronously delivered to its recipients.

thanethomson: I'd like to extend my offer again here: https://github.com/cometbft/cometbft/pull/82#discussion_r1157749867

docs/architecture/adr-100-data-companion-push-api.md
thanethomson: I'm not sure whether we can remove these yet. I was trying to keep the RPC endpoint removals to the bare minimum here. We can definitely consider removing more endpoints over time as we provide more ways for integrators to access this data through data companions.

thanethomson:
I've updated this ADR, addressing most of the comments afaik, and have marked it as rejected to keep it for posterity as an indication that we've considered this kind of option. It is, however, rejected at present in favour of #82.

sergio-mena (approved):

Issue #15, opened 2022-12-23
Formal specification of the Gossip layer


lasarojc:
This issue tracks the work of documenting and specifying the interaction between the consensus reactor and the P2P Layer, as well as the interactions inside the consensus reactor itself, between the consensus implementation and the consensus gossip layer.

First, this work will produce an initial set of draft specifications for the abstract behavior expected of the gossip layer and of its interactions Consensus and P2P, both in English and in Quint.
Quint specs will contain tests and will be model checked, once the tooling allows it.

Second, the specs will be refined to produce lower level abstractions, closer to the implementation level.
A primary concern here is to ensure that the abstractions can be efficiently implemented.

Third, the current implementation will be described in documented, in English, and abstracted in Quint.
The implementation will be matched to the abstract specs from the previous steps in a high level refinement mapping.
This work will possibly be completed with formal refinement mapping proofs, manual and/or automated.

The following list breaks these tasks down for easier tracking:
- Draft specification of the P2P/Gossip/Consensus interaction
   - [x] Exploratory study of the interaction between Consensus and Gossip (done in #16)
       - [x] Identify main abstractions to specify.
           - It has been identified that Gossip may be implemented as a 2P-Set CRDT.
           - It has been identified that a "garbage collection" must run in the 2P-Set; the definition will happen later.
   - [x] Draft of high level GOSSIP-I, in English (done in #16)
   - [x] Draft of high level GOSSIP-I, in Quint (done in #16)
   - [x] Draft of high level P2P-I, in English (done in #16) 
   - [x] Draft of high level P2P-I, in Quint (done in #16) 
       - A very simple interface was defined here and further work will be done in collaboration with other team members.
   - [ ] Quint tests for provided specs
        - [ ] P2P-I tests
        - [ ] GOSSIP-I tests
        - [ ] Quint mocks of GOSSIP
        - [ ] Quint mocks of P2P
   - [ ] Specification of the "garbage collection", GC (#608)
        - [ ] English
        - [ ] Quint
        - [ ] Tests

- Further Investigate CRDT, including implementations.
    - [x] #610
    - [ ] #751

- Refine specification of the P2P/Gossip/Consensus interaction (#17)
   - [ ] Refine GOSSIP-I, in english
   - [ ] Refine GOSSIP-I, in Quint
   - [ ] Refine P2P-I, in English (collaboration with other spec work)
   - [ ] Refine P2P-I, in Quint (collaboration with other spec work)
   - [ ] Update Quint tests.

[...]

PR #74, opened 2023-01-04
Consensus Reactor communication specification

lasarojc:
Original PR: https://github.com/tendermint/tendermint/pull/9870

Contributes to #15


cason (commented):
Great document. I've left a number of questions, I hope they help with the definitions and abstractions.

spec/consensus/reactor/reactor.md
```diff
@@ -0,0 +1,413 @@
...
+| If $m1$ and $m2$ are broadcast by any processes and $m2.\text{SSS}(m1)$, then the delivery of $m1$ is not required.
+
+[^todo1]: :clipboard: **TODO**: Is this really required? Should this be a requirement of GOSSIP?
+
+To be useful, however, some legitimate effort has to be made to deliver messages.
+
+|Best-Effort Communication with Supersession|
+|-----|
+| (i) If a correct process $p$ broadcasts some message $m$ at time $t$, $m$ is not superseded, and no failures or network partitions happen after $t$, then eventually every correct process delivers $m$.    
+| (ii) If a correct process $p$ receives some message $m$ at time $t$, $m$ is not superseded, and no failures or network partitions happen after $t$, then eventually every correct process delivers $m$.
```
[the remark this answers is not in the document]
lasarojc: It was confusing. Fixing.

spec/consensus/reactor/README.md
```diff
@@ -0,0 +1,113 @@
+# Consensus Reactor
+
+CometBFT is a state machine replication framework and part of the stack used in the Cosmos ecosystem to build distributed applications, such as the Cosmos SDK.
```
sergio-mena, line 3: I wouldn't define the Cosmos SDK as an app. Consider mentioning the Cosmos Hub instead

spec/consensus/reactor/README.md
```diff
@@ -0,0 +1,113 @@
...
+This set of documents focuses on the interactions between the P2P layer and the Consensus Reactor.
+
+```
+                                                      SDK Apps
+                                                   ==================
+ Applications                                        Cosmos SDK    
+======================================ABCI============================ ┐
+ [Mempool Reactor] [Evidence Reactor] [Consensus Reactor] [PEX] [...]  |
+- - - - - - - - - - - - - - - - P2P-I - - - - - - - - - - - -- - - - - | Tendermint
+                                  P2P                                  | Core
```
sergio-mena, line 18:
```suggestion
- - - - - - - - - - - - - - - - P2P-I - - - - - - - - - - - -- - - - - | CometBFT
                                  P2P                                  |
```

spec/consensus/reactor/README.md
```diff
@@ -0,0 +1,113 @@
...
+2. Provide equivalent Quint specifications, used to mechanically check the properties
+3. Provide an english description of how the current implementation matches (or not) the specified behavior, that is, a very loose refinement mapping.
+
+
+# Outline
+
+The specification is divided in multiple documents
+* [reactor.md](./reactor.md): specification in English
+* [reactor.qnt](./reactor.qnt): corresponding specifications in [Quint](https://github.com/informalsystems/quint)
+* [implementation.md](./implementation.md): a description of what is currently implemented in Tendermint Core, in English.
```
sergio-mena, line 58:
```suggestion
* [implementation.md](./implementation.md): a description of what is currently implemented in CometBFT, in English.
```

spec/consensus/reactor/README.md
```diff
@@ -0,0 +1,113 @@
...
+* [implementation.qnt](./implementation.qnt): Quint model of current behavior, for model checking of provided properties.
+
+
+# Conventions
+
+* MUST, SHOULD, MAY...
```
sergio-mena, line 64:
```suggestion
* MUST, SHOULD, MAY... according to RFC2119.
```

spec/consensus/reactor/reactor.md
```diff
@@ -1,196 +1,141 @@
...
 
-> **Warning**    
-> This is a Work In Progress
+The Tendermint algorithm also assumes that this property is used to provide **Gossip Communication**:
 
-> **Warning**    
-> Permalinks to excerpts of the Quint specification are provided throughout this document for convenience, but may outdated.
+|Gossip communication|
+|-----|
+| (i) If a correct process $p$ sends some message $m$ at time $t$, all correct processes will receive $m$ before $\text{max} \{t,\text{GST}\} + \Delta$.
```
sergio-mena, line 16: Suggestion: would these definitions still be correct if we worded like this ...?

(i) If a correct process $p$ sends some message $m$ at time $t$, at least one correct processes will receive $m$ before $\text{max} \{t,\text{GST}\} + \Delta$.

spec/consensus/reactor/reactor.md
```diff
@@ -1,196 +1,141 @@
...
-| [REQ-CONS-GOSSIP-DELIVERY.2]      | X | [PROV-GOSSIP-CONS-DELIVERY.2]         | X | X |  |
-| [REQ-GOSSIP-CONS-SUPERSESSION.1]  | X | [PROV-CONS-GOSSIP-SUPERSESSION.1]     | X | X |  |
-| [REQ-GOSSIP-CONS-SUPERSESSION.2]  | X | [PROV-CONS-GOSSIP-SUPERSESSION.2]     |   |   |  |
-|                                   |   | [PROV-CONS-GOSSIP-SUPERSESSION.3]     | X |   |  |
-| [REQ-GOSSIP-P2P-CONNECTION.1]     | X |                                       |   |   |  |
-| [REQ-GOSSIP-P2P-UNICAST.1]        | X |                                       |   |   |  |
-| [REQ-GOSSIP-P2P-UNICAST.2]        | X |                                       |   |   |  |
-| [REQ-GOSSIP-P2P-CONCURRENT_CONN]  | X |                                       |   |   |  | 
+Fortunately, while Gossip Communication is a sufficient condition for the Tendermint algorithm to terminate, it is not strictly necessary.
+First, Tendermint will keep progressing if periods in which communication is timely for long enough for rounds to complete continue to happen. Even though "long enough" periods, or **Local Stabilization Periods**, is still hard to define, it is certainly less strict than GST.
```
sergio-mena, line 24:
```suggestion
First, the Tendermint algorithm will keep progressing if periods in which communication is timely for long enough for rounds to complete continue to happen. Even though "long enough" periods, or **Local Stabilization Periods**, is still hard to define, it is certainly less strict than GST.
```

spec/consensus/reactor/reactor.md
lasarojc: Seems correct. Any proofs of termination would have to consider $2\Delta$ as the actual deadline for decision, that's the only consequence, I think.
I wouldn't change the wording here, though, as this has been copied from the paper.

spec/consensus/reactor/reactor.md
```diff
@@ -0,0 +1,318 @@
...
+
+
+# Part 2: CONS/GOSSIP interaction
+CONS, the Consensus Reactor State Layer, is where the actions of the Tendermint BFT are implemented.
+Actions are executed once certain pre-conditions apply, such as timeout expirations or reception of information from particular subsets of the nodes in the system, neighbors or not.
+
+An action may require communicating with applications and other reactors, for example to gather data to compose a proposal or to deliver decisions, and with the P2P layer, to communicate with other nodes.
+
+## Northbound Interaction - ABCI
+Here we assume that all communication with the Application and other reactors are performed through the Application Blockchain Interface, or [ABCI](../../abci/).
```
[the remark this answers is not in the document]
lasarojc: fair enough. simply pointed to ABCI and to some future work.

spec/consensus/reactor/reactor.md
```diff
@@ -0,0 +1,318 @@
...
+Both approaches should be equivalent and not impact the specification much, even if the corresponding implementations would be much different.
+For now, we follow the first approach by having CONS read the sets directly.[^setsorpred]
+
+[^setsorpred]: **TODO**: should we not specify these shared variables and instead pass predicates to GOSSIP from consensus? Variables make it harder to separate the CONS from GOSSIP, as the the variables are shared, but is highly efficient. Predicates are cleaner, but harder to implement efficiently. For example, when a vote arrives, multiple predicates may have to be tested independently, while with variables the tests may collaborate with each other.
+
+CONS and GOSSIP also share a vocabulary of CONS messages and an operator to test for message supersession.
+
+[VOC-CONS-GOSSIP]    
+* Message Types[^messagetypes]
+    * ProposalMessage
```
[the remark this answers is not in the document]
lasarojc: BlockPart is not part of the CONS vocabulary. it is an GOSSIP only message.

spec/consensus/reactor/reactor.md
```diff
@@ -0,0 +1,327 @@
...
+In the specific case of the Proposal step, only the proposer of the round can have a matching entry.
+
+- $\lang H, R, Proposal, *, * \rang$ returns at most one value.
+
+A violation of these rules is a proof of misbehavior.
+
+
+### Convergence Requirements
+
+We now formalize the requirements of the Tendermint algorithm in terms of an eventually consistent Tuple Space, which may be implemented using reliable communication (Gossip Communication), some best *best-effort* communication primitive that only delivers messages that are still useful, or some Gossip/Epidemic/Anti-Entropy approach for state convergence.
```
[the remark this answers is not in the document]
adizere: It can be confusing here to mention 3 communication mechanisms for ensuring convergence (reliable/gossip, best-effort, epidemic). While reading this, I feel like there is an insight here behind the reason for mentioning all of them, but I don't get the insight. Can you clarify, or simplify?

spec/consensus/reactor/reactor.md
```diff
@@ -0,0 +1,327 @@
...
+
+$\Delta$ encapsulates the assumption that, during stable periods, timeouts eventually do not expire precociously, given that they all can be adjusted to reasonable values, and the steps needed to converge on entries can be accomplished within $\Delta$.
+Without precocious timeouts, all new entries should help algorithms progress.
+In the Tendermint algorithm, for example, no votes for Nil are added and the round should end if a proposal was made.
+
+
+
+## The tuple space as a CRDT
+
+The two approaches described in the [previous section](#nodes-state-as-a-tuple-space), without the deletion of entries, correspond to operation-based and state-based [Grow-only SET](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type#G-Set_(Grow-only_Set)) CRDT (G-Set).
```
adizere, line 141: > The two approaches described in the [previous section](#nodes-state-as-a-tuple-space)

Not sure I understand which were the _two_ approaches. It seems like there is only one.

spec/consensus/reactor/reactor.md
```diff
@@ -0,0 +1,327 @@
...
+ABCI specifies both what CONS [requires from the applications](../../abci/abci%2B%2B_app_requirements.md) and on what CONS [provides to Applications](../../abci/abci%2B%2B_tmint_expected_behavior.md).
+
+Interactions with other reactors, such as with the Mempool reactor to build tentative proposals, will be covered elsewhere.
+
+
+
+## Southbound Interaction - GOSSIP-I
+CONS interacts southbound only with GOSSIP, to update the gossip state and to evaluate the current state to check for conditions that enable actions.
+
+To update the state, CONS passes a tuple GOSSIP a tuple with the exact content to be added to the tuple space through a functional call.
```
adizere, line 182: It's not obvious where the tuple space lives. I would have expected it to live inside the CONS layer intuitively. But here it's implied that GOSSIP holds the tuple space. Might be good to clarify which layer holds $t_p$.

adizere (commented):
Did a first pass, good start here so far! I only had minor comments, looking forward to seeing this mature.

spec/consensus/reactor/reactor.md
```diff
@@ -0,0 +1,333 @@
...
+
+Three kinds of messages are exchanged in the Tendermint algorithm: `PROPOSAL`, `PRE-VOTE`, and `PRE-COMMIT`.
+The algorithm progresses when certain conditions are satisfied over the set of messages received.
+For example, in order do decide on a value `v`, the set must include a `PROPOSAL` for `v` and more than two thirds of the number processes in `PRE-COMMIT` for the same `v`, during the same round.
+
+To ensure progress and termination, the algorithm assumes a **Global Stabilization Time (GST)**, after which communication is reliable and timely (Eventual $\Delta$-Timely Communication), which is used to provide **Gossip Communication**.
+
+| Eventual $\Delta$-Timely Communication|
+|-----|
+|There is a bound $\Delta$ and an instant GST (Global Stabilization Time) such that if a correct process $p$ sends a message $m$ at a time $t \geq \text{GST}$ to a correct process $q$, then $q$ will receive $m$ before $t + \Delta$.
```
peterbourgon, line 15: This means that a process can be correct from the perspective of one node in a network, and incorrect from the perspective of a different node in the same network. In other words, "correct" is not a property of the process itself, but is rather a property of a process relative to a given observer. Is that well understood?

peterbourgon (commented):
Curious definition of "gossip" here, I posed some questions :)

spec/consensus/reactor/implementation.md
```diff
@@ -0,0 +1,212 @@
...
+
+1. Superseding is not explored to drop superseded messages from the outgoing queues.
+
+## [LOCAL_CAST.1]
+For each of the neighbors of the node, looping go-routines continuously evaluate the conditions to send messages to other nodes.
+If a message must be sent, it is enqueued for transmission using TCP and will either be delivered to the destination or the connection will be dropped.
+New connections reset the state of Communication layer wrt the newly connected node (in case it is a reconnection) and any messages previously sent (but possibly not delivered) will be resent if the conditions needed apply. If the conditions no longer apply, it means that the message has been superseded and need no be retransmitted.
+
+1. Every message added to BCast is enqueued or will be enqueued later.
+2. Every message enqueued will be delivered using TCP or the connection will break.
```
peterbourgon, line 131: It's possible that an "enqueued" message is not delivered, and the corresponding connection is not broken, for an indefinite period of time. Defining a timeout at the application layer can break the connection at a certain point, but even in that case the message may still be delivered to the remote host. Hopefully I'm just preaching to the choir and this is all well understood 🙃 but if not please speak up as I guess this is important?

spec/consensus/reactor/implementation.md
```diff
@@ -0,0 +1,212 @@
...
+1. Broadcast messages are put in outgoing queues for current connections.
+1. If the condition persists, then they are put in the queue again on later iterations, for certain messages.
+1. If the condition persists, then they are put in the queues for new connections in later iterations.
+1. If the condition does not persist, then some superseding messages must have been sent and the superseded one may be dropped (same as remaining in BMsgs but never being actually sent again on the network).
+
+1. Superseding is not explored to drop superseded messages from the outgoing queues.
+
+## [LOCAL_CAST.1]
+For each of the neighbors of the node, looping go-routines continuously evaluate the conditions to send messages to other nodes.
+If a message must be sent, it is enqueued for transmission using TCP and will either be delivered to the destination or the connection will be dropped.
```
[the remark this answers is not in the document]
peterbourgon: TCP does not guarantee that a given message is delivered or else the underlying connection is terminated.

spec/consensus/reactor/reactor.md
```diff
@@ -0,0 +1,333 @@
...
+
+
+
+
+> Warning/TODO: A word about tombstones, that is $D$
+> - Only state that is not required should be deleted/tombstone'd.
+> - Instead of tombstones, add new entries that trigger removal of other entries (for example, state about a new height); each node must be given information to realize by itself that an entry is no longer needed.
+> - Tombstones are an optimization, kept to prevent data recreation and redeletion.
+> - Tombstones should be garbage collected at some point; imprecision shouldn't affect correctness/termination, since this is an optimization (as long as deleted state is never required again).
+> - Tombstones are not to be gossiped; if they were, they would need to carry proof for the reason they were created, defeating their point.
```
peterbourgon, line 155: In a 2P-Set, both the add set `A` and the remove set `D` are grow-only sets, i.e. G-Sets: they only accrue elements over time, they never lose elements. Elements in the remove set `D` can be considered tombstones, but those tombstones are not optimizations: they're equally important to elements in the add set `A`, when determining the ultimate set membership. 

It's possible to GC elements in a 2P-Set, but that GC can't be done on the "tombstones" in the remove set in isolation. It has to be performed holistically, to "reset" the state of both the add and remove sets together, and in a way that is reliably the same across all nodes in the network.

In summary, this description of a 2P-Set — and specifically the description of the remove set subcomponent of a 2P-Set — has some problems, at least as far as I understand the concepts at play. I'm happy to be corrected! But based on my current understanding, I don't think the state space can be modeled as a CRDT, as described here..

spec/consensus/reactor/reactor.md
```diff
@@ -0,0 +1,333 @@
...
+In other words, the algorithm does not require all messages to be delivered, only messages that advance the state of the processes.
+
+## Node's state as a Tuple Space
+
+One way of looking at the information used by CometBFT nodes is as a distributed tuple space to which all nodes contribute with signed entries;
+to share a proposal, the proposer adds it to the tuple space and to vote for a proposal, the node adds the vote.
+Each update includes the author name and therefore they are all non-conflicting with each other and update forgery is prevented with signing each update.
+
+Since we are talking about an asynchronous distributed system, individual nodes can only maintain approximations of the tuple space.
+Nodes may broadcast the update operations to all nodes, including themselves, and, if the communication is reliable, as guaranteed by Gossip Communication, the tuple space will eventually converge.
```
peterbourgon, line 40: I'm not sure I follow how gossip guarantees that communication between nodes is reliable. 

You define a concept of GST that asserts an upper bound for "correct" processes to receive messages, but in the case of a network fault, all processes would be classified as "incorrect" — it's not clear to me how this outcome produces reliable communication?

spec/consensus/reactor/reactor.md
lasarojc: Thanks for the comments, @peterbourgon. The quoted part was a set of notes for myself, things to think about, so please disregard them as they may indeed be imprecise. But just to share what I was thinking, the idea here is that the tombstone can be used to GC elements from A and prevent them from being re-added and if tombstones are gossiped, then the removals occur everywhere, but gossiping the tombstones is not ideal either. So we are working on a way to have implicit tombstones, but the idea is still being developed.

spec/consensus/reactor/implementation.md
lasarojc: The point here is that if the message is not delivered once enqueued, it will be enqueued again, eventually.
And eventually it will be delivered or it will not be needed anymore.

spec/consensus/reactor/implementation.md
```diff
@@ -0,0 +1,212 @@
...
+
+
+
+
+
+
+
+### The inner workings
+
+CONS uses `broadcast` to send messages to all processes, but since the network may not be fully connected, communication happens at a **local** level, in which information is delivered to the neighbors of a node, and a **global level**, in which information is forwarded on to neighbors of neighbors and so on.
```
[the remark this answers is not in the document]
lasarojc: I would say the assumption is that the network is connected over time, but not necessarily connected on any instant. I was thinking of using time dependent graph, but that idea is parked for now.

spec/consensus/reactor/reactor.md
lasarojc: Being correct is a an absolute property of a node/process, not relative.
Nodes can wrongly suspect each other, but the definition of Gossip Communication does not take this into account.

spec/consensus/reactor/reactor.md
```diff
@@ -0,0 +1,333 @@
...
+To ensure progress and termination, the algorithm assumes a **Global Stabilization Time (GST)**, after which communication is reliable and timely (Eventual $\Delta$-Timely Communication), which is used to provide **Gossip Communication**.
+
+| Eventual $\Delta$-Timely Communication|
+|-----|
+|There is a bound $\Delta$ and an instant GST (Global Stabilization Time) such that if a correct process $p$ sends a message $m$ at a time $t \geq \text{GST}$ to a correct process $q$, then $q$ will receive $m$ before $t + \Delta$.
+
+|Gossip communication|
+|-----|
+| (i) If a correct process $p$ sends some message $m$ at time $t$, all correct processes will receive $m$ before $\text{max} \{t,\text{GST}\} + \Delta$.
+| (ii) If a correct process $p$ receives some message $m$ at time $t$, all correct processes will receive $m$ before $\text{max}\{t,\text{GST}\} + \Delta$.
```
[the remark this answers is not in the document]
lasarojc: Gossip Communication is defined in the tendermint paper, from 2018.
Eventual $\Delta$-Timely Communication is being defined here, as a first step to Gossip Communication.

spec/consensus/reactor/reactor.md
lasarojc: When I say Gossip Communication I am not talking about gossip as a general technique. I am talking about the very precise definition given in the previously in the document, which came from the Tendermint paper.
Gossip Communication, as defined, is reliable for all messages seen by correct processes.

spec/consensus/reactor/reactor.md
peterbourgon: Got it.

Why is gossiping tombstones not ideal?

spec/consensus/reactor/reactor.md
peterbourgon: > Being correct is an absolute property of a node/process, not relative

From https://arxiv.org/pdf/1807.04938.pdf

> Formally, we model the network communication using a variant of the partially synchronous system model [6]: in all executions of the system there is a bound ∆ and an instant GST (Global Stabilization Time) such that all communication among correct processes after GST is reliable and ∆-timely, i.e., if a correct process p sends message m at time t ≥ GST to a correct process q, then q will receive m before t + ∆5. In addition to the standard partially synchronous system model [6], we assume an auxiliary property that captures gossip-based nature of communication6:

If "all communication among correct processes ... is reliable" within some upper bound (GST), and if processes are connected over a network, then "correctness" is necessarily subject to the network link(s) between processes. Those links are not equivalent for all processes, they're different for each connection: a given node may be reachable (correct) from one peer, but unreachable (incorrect) from another. As far as I can see, that's a  relative definition, until/unless those differences are resolved via e.g. consensus. 

With that said...

> Nodes can wrongly suspect each other, but the definition of Gossip Communication does not take this into account.

I guess this isn't the place to debate whether or not Gossip Communication is correct as defined :)

spec/consensus/reactor/implementation.md
peterbourgon: I understand these are the expectations at the application layer, but how are those expectations guaranteed? What re-enqueues messages that aren't delivered? What determines if an un-delivered message is no longer necessary and removes it from the queue?

spec/consensus/reactor/reactor.md
lasarojc: Say a message A is in the set and a message B arrives, causing A to become useless. We GC A and create a tombstone nA (not A). If nA is gossiped to another node, then it must carry B as well, to prove that nA was created in good faith, but in this case the receiver of nA can simply apply the same rule as the sender and use B to remove nA. Gossiping nA does not contribute in any way here.

spec/consensus/reactor/implementation.md
lasarojc: > What determines if an un-delivered message is no longer necessary
The analysis of the current state of the node. If the node has moved to heigh X, there is no point in sending messages for height X-2, for example.
> and removes it from the queue?
There is no need to remove it from the queue. Just don't enqueue it again.

spec/consensus/reactor/reactor.md
lasarojc: A node may be correct and still be completely isolated from all the other nodes because its links to such nodes are faulty. I'd agree that, In practice, this node use useless, but it is still a correct node according to the model.

> I guess this isn't the place to debate whether or not Gossip Communication is correct as defined :)
💯

spec/consensus/reactor/reactor.md
peterbourgon: Ah, OK, and that all rests on the assumption that Gossip Communication provides reliable delivery of messages to nodes, correct?

spec/consensus/reactor/implementation.md
peterbourgon: OK, I think I get it. This all assumes that delivery success/failure is knowable from the perspective of the sender. That's a property asserted by Gossip Communication (which is probably where you're speaking from) but not actually provided by e.g. TCP connections over a physical network (which is where I've been speaking from).

spec/consensus/reactor/reactor.md
peterbourgon: > Gossip Communication, as defined, is reliable for all messages seen by correct processes.

That's fine, as long as it's understood that this definition of Gossip Communication is not actually satisfied by the  current CometBFT implementation.

spec/consensus/reactor/implementation.md
lasarojc: Actually, the sender does not need to know, as long as it keep insisting and sending any messages that might still be useful. TCP cannot guarantee the delivery, but if you insist (reconnect if needed), in practice, stable periods will happen and deliver will succeed. Again, the sender may never know when this happens, so it needs to keep retransmitting. 

This is one place where theory does not meet practice. We write algorithms and prove that they will progress after GST, but we cannot guarantee that there will be a GST. However, as long GST is required only for liveness, not safety, and as long as the algorithm may still make progress during small stability periods, we get progress even if GST never arrives. This is the case with TM.

peterbourgon: The problem is that it's not possible for a sender to know if a given message has actually been received by its destination. An acknowledgement (ACK) is most likely a true-positive signal that the message was received, but a NAK is not a true-negative signal that the message was not received. That means the application can't treat a NAK as a signal to delete a message from its transmit queue, and can only stop trying to transmit a message once an alternative, out-of-band signal is received, that indicates that message is no longer useful. I guess that's the committed block height received over consensus. That's fine. But what happens if that latest block height data is delayed indefinitely? It means that GST is unbounded. You can classify such a node as incorrect, and thereby remove it from consideration of your system model, but that classification relies on information that is not available to the node itself, or (potentially) other nodes that also did not receive the latest block height data. So we're back to square one in terms of correctness being a relative measure. Again, this is all kind of fine, in terms of correctness! The only consequence I'm trying to point out is that GST is not guaranteed by the protocol, and can't be assumed by any implementation.

lasarojc: Please read my previous message again, where I clearly stated that "...the sender does not need to know, as long as it keep insisting and sending any messages that might still be useful"  and "the sender may never know when this happens, so it needs to keep retransmitting."

> can only stop trying to transmit a message once an alternative, out-of-band signal is received, that indicates that message is no longer useful.
It is not out of band, it is in the consensus protocol, which is the reason for sending the messages.

> It means that GST is unbounded.
This phrase does not make sense.

> You can classify such a node as incorrect, and thereby remove it from consideration of your system model, but that classification relies on information that is not available to the node itself, or (potentially) other nodes that also did not receive the latest block height data.
I am not the one that wants to classify an incommunicado node as faulty.

> So we're back to square one in terms of correctness being a relative measure. 
No we are not. A correct process is a correct process. A faulty process is a faulty process. The same applies to links, independently.

> GST is not guaranteed by the protocol...
Indeed it is not. It is assumed of the environment. FLP implies that if you want to prove that your algorithm eventually makes progress and terminates then you need to extend your asynchronous system somehow, be it with randomization oracles, perfect failure detectors or, in the case of tendermint, GST. 

> and can't be assumed by any implementation.
That's exactly what you have to do. You write you algorithm so that it makes progress when GST arrives and expect that it will progress even if it is not really GST, but just some stable period. You cannot guarantee it will be enough, but it is best you can do, given FLP.

spec/consensus/reactor/reactor.md
lasarojc: Indeed it is and this is one of the reasons for this spec.

> Fortunately, while Gossip Communication is a sufficient condition for the Tendermint algorithm to terminate, it is not strictly necessary, because the conditions to progress and terminate are evaluated over the messages of subsets of rounds executed, not all of them, and as new rounds are executed, messages in old rounds may be become obsolete and be ignored and forgotten, saving memory.
> In other words, the algorithm does not require all messages to be delivered, only messages that advance the state of the processes.

spec/consensus/reactor/implementation.md
peterbourgon: > A correct process is a correct process. A faulty process is a faulty process. The same applies to links, independently.

I don't understand how this is true.

Some process A can be correct from the perspective of an observer process X, and incorrect from the perspective of a different observer process Y. (If this is not evident, let me know, I can provide relevant examples.) Is that process A correct or incorrect? AFAICT, it is neither; there is no well-defined answer, only relative perspectives.

Similarly, the link between A and X can be correct from the perspective of A, and incorrect from the perspective of X. Is that A-X link correct or incorrect? AFAICT, same outcome as above.

It's entirely possible I'm missing something here. What am I missing?

--

> FLP implies that if you want to prove that your algorithm eventually makes progress and terminates then you need to extend your asynchronous system somehow, be it with randomization oracles, perfect failure detectors or, in the case of tendermint, GST.

I think the problem I'm observing here is that this invariant -- that the algorithm will eventually make progress -- is fine as an assumption for a system model, but not fine as an assumption for a real-world implementation. Basically, FLP asserts that it is not actually possible to for an algorithm to make progress and terminate, in any real-world context. Following: that GST is not guaranteed, or guarantee-able, in any real-world context. Does that mean that an algorithm does not necessarily make progress and/or terminate? Yes! It does mean that! Randomization oracles don't solve this problem; perfect failure detectors are not feasible in any real-world network.

spec/consensus/gossip/gossip.md
```diff
@@ -0,0 +1,333 @@
...
+Hence, processes execute in rounds in which they wait for conditions to be met for sometime and, if they timeout, send negative messages that will lead to new rounds.
+Again, Gossip Communication guarantees that eventually the conditions for deciding are met, even if only after GST is reached.
+
+Implementing Gossip Communication, however, is hard because even messages sent before the GST need to be buffered to be reliably delivered between correct processes, but since GST may take indefinitely long to arrive, in practice implementing this property would require unbounded memory.
+
+Fortunately, while Gossip Communication is a sufficient condition for the Tendermint algorithm to terminate, it is not strictly necessary, because the conditions to progress and terminate are evaluated over the messages of subsets of rounds executed, not all of them, and as new rounds are executed, messages in old rounds may be become obsolete and be ignored and forgotten, saving memory.
+In other words, the algorithm does not require all messages to be delivered, only messages that advance the state of the processes.
```
08d2, line 31: No real-world implementation of a distributed system over a geographically diverse network can satisfy the definition of GST which asserts that "if a correct process P sends a message at time T to a correct process Q, then Q will receive that message before T+∆".

Nothing can guarantee the delivery of a message from P to Q. You can say that delivery failure to Q marks Q as "incorrect", but that means "correctness" is necessarily a property that's relative to a specific observer. Q can be incorrect from P1, but correct from P2. There is no single global truth about Q's correctness.

Weakening Gossip Communication to be "not strictly necessary" because "the algorithm [only requires] messages that advance the state of the processes" doesn't really solve anything. Those requisite messages are still (presumably) delivered over gossip channels, right? If so, then they still require reliable delivery; in fact delivery guarantees at an e.g. gossip layer are not really different based on properties of any individual message. 

Gossip communication is not guaranteed, and is not reliable, just by definition. So you can't assume reliable delivery of gossiped messages at a higher layer e.g. CometBFT. And that means you can't rely on any specific guarantees of higher-layer concepts like GST in your application code.

spec/consensus/gossip/gossip.md
```diff
@@ -0,0 +1,333 @@
...
+to share a proposal, the proposer adds it to the tuple space and to vote for a proposal, the node adds the vote.
+Each update includes the author name and therefore they are all non-conflicting with each other and update forgery is prevented with signing each update.
+
+Since we are talking about an asynchronous distributed system, individual nodes can only maintain approximations of the tuple space.
+Nodes may broadcast the update operations to all nodes, including themselves, and, if the communication is reliable, as guaranteed by Gossip Communication, the tuple space will eventually converge.
+
+Otherwise, nodes may periodically compare their approximations with each other to identify and correct differences by adding missing entries, using some gossip/anti-entropy protocol.
+
+Updates are commutative from the point of view of the tuple space, even if they are not commutative from the Tendermint algorithm's point of view; however, nodes observing different approximations of the tuple space may decide at different point in time but cannot violate any correctness guarantees.
+And the eventual convergence of tuple space implies the eventual termination of the algorithm.
```
08d2, line 45: Commutativity is not by itself sufficient to establish a CRDT, updates must also be associative and idempotent. And if you say that entries are removed from the set based on application-level properties of added (received) updates, then the set is not a CRDT by definition.

That's because elements which modify the value space based on prior state (which such values necessarily do) are not idempotent. Receiving elements {A B C} can produce a different outcome compared to receiving elements {A C B}.

spec/consensus/gossip/gossip.md
```diff
@@ -0,0 +1,333 @@
...
+
+- $\lang H, R, Proposal, *, * \rang$ returns at most one value.
+
+A violation of these rules is a proof of misbehavior.
+
+
+### Convergence Requirements
+
+We now formalize the requirements of the Tendermint algorithm in terms of an eventually consistent Tuple Space, which may be implemented using reliable communication (Gossip Communication), some best *best-effort* communication primitive that only delivers messages that are still useful, or some Gossip/Epidemic/Anti-Entropy approach for state convergence.
+All of these can be made to progress after GST but should also progress during smaller stability periods.
```
08d2, line 114: It's insufficient for this kind of consensus-impacting data to be subject to delivery constraints that provide only best-effort semantics. (Gossip is not reliable.)

spec/consensus/gossip/implementation.md
```diff
@@ -0,0 +1,212 @@
...
+1. Broadcast messages are put in outgoing queues for current connections.
+1. If the condition persists, then they are put in the queue again on later iterations, for certain messages.
+1. If the condition persists, then they are put in the queues for new connections in later iterations.
+1. If the condition does not persist, then some superseding messages must have been sent and the superseded one may be dropped (same as remaining in BMsgs but never being actually sent again on the network).
+
+1. Superseding is not explored to drop superseded messages from the outgoing queues.
+
+## [LOCAL_CAST.1]
+For each of the neighbors of the node, looping go-routines continuously evaluate the conditions to send messages to other nodes.
+If a message must be sent, it is enqueued for transmission using TCP and will either be delivered to the destination or the connection will be dropped.
```
08d2, line 127: The condition that a message is either delivered, or the connection is dropped, is not guaranteed, and cannot be assumed.

spec/consensus/gossip/gossip.md
```diff
@@ -0,0 +1,333 @@
...
+
+
+
+
+> Warning/TODO: A word about tombstones, that is $D$
+> - Only state that is not required should be deleted/tombstone'd.
+> - Instead of tombstones, add new entries that trigger removal of other entries (for example, state about a new height); each node must be given information to realize by itself that an entry is no longer needed.
+> - Tombstones are an optimization, kept to prevent data recreation and redeletion.
+> - Tombstones should be garbage collected at some point; imprecision shouldn't affect correctness/termination, since this is an optimization (as long as deleted state is never required again).
+> - Tombstones are not to be gossiped; if they were, they would need to carry proof for the reason they were created, defeating their point.
```
08d2, line 155: 2P-Sets require that elements in the "remove" set are communicated between nodes in the same way as elements in the "add" set. If you're not gossiping tombstones (remove set elements) that's fine, but it means you're not implementing a 2P-Set.

spec/consensus/gossip/gossip.md
```diff
@@ -0,0 +1,333 @@
...
+```
+
+P2P must expose functionality to allow 1-1 communication with connected nodes.
+
+```qnt reactor.gen.qnt
+<<DEF-UNICAST>>
+```
+
+### Requires from P2P - P2P-I
+Message to nodes that remain connected are reliably delivered.
```
08d2, line 297: Not guaranteed.

spec/consensus/gossip/implementation.md
```diff
@@ -0,0 +1,212 @@
...
+> **TODO**: Add permalink
+
+
+#### Current implementations
+
+[REQ-GOSSIP-P2P-UNICAST]   
+- `Send(Envelope)`/`TrySend(Envelope)`
+    - Enqueue and forget. 
+    - Disconnection and reconnection causes drop from queues.
+    - Enqueuing may block for a while in `Send`, but not on `TrySend`
```
cason, line 195: Moreover, if enqueuing is not possible (because the remote peer is slow or crashed), both methods return `false`: `TrySend` immediately, `Send` after 10s.

spec/consensus/gossip/implementation.md
```diff
@@ -0,0 +1,212 @@
...
+
+> **TODO**: Add permalink
+
+
+#### Current implementations
+
+[REQ-GOSSIP-P2P-UNICAST]   
+- `Send(Envelope)`/`TrySend(Envelope)`
+    - Enqueue and forget. 
+    - Disconnection and reconnection causes drop from queues.
```
cason, line 194: More precisely, in case of a graceful disconnection, all enqueued messages are guaranteed to be sent to and delivered by the peer. Otherwise, no guarantees can be provided.

spec/consensus/gossip/gossip.md
```diff
@@ -0,0 +1,333 @@
...
+
+```qnt reactor.gen.qnt
+<<VOC-CONS-GOSSIP-TYPES>>
+```
+
+### Requires from GOSSIP
+
+CONS is provided with functions to add and remove tuples from the space.[^removal]
+
+[^removal]: removal of tuples has no equivalent in the Tendermint algorithm. **TODO** This is something to be added here.
```
cason, line 212:
```suggestion
[^removal]: removal of tuples has no equivalent in the Tendermint consensus algorithm. **TODO** This is something to be added here.
```

spec/consensus/gossip/gossip.md
```diff
@@ -0,0 +1,333 @@
...
+
+CONS is provided access to the local view.
+
+
+```qnt reactor.gen.qnt
+<<DEF-READ-TUPLE>>
+```
+
+> **Note**
+> If you read previous versions of this draft, you will recall GOSSIP was aware of supersession. In this version, I am hiding supersession in REQ-CONS-GOSSIP-REMOVE and initially attributing the task of identifying superseded entries to CONS, which then removes what has been superseded. A a later refined version of this spec will clearly specify how supersession is handled and translated into removals.
```
cason, line 226: This is the only occurrence of REQ-CONS-GOSSIP-REMOVE.

spec/consensus/gossip/gossip.md
```diff
@@ -0,0 +1,333 @@
...
+
+
+
+
+# Part III: GOSSIP requirements and provisions
+GOSSIP, the Consensus Reactor Communication Layer, provides on its northbound interface the facilities for CONS to communicate with other nodes by adding and removing tuples and exposing the eventually converging tuple space.
+On its southbound interface, GOSSIP relies on the P2P layer to implement the gossiping.
+
+## Northbound Interaction - GOSSIP-I
+Northbound interaction is performed through GOSSIP-I, whose vocabulary has been already [defined](#gossip-i-vocabulary).
```
cason, line 255: The link is broken. Does it refer to the Shared Vocabulary section?

spec/consensus/gossip/gossip.md
```diff
@@ -0,0 +1,333 @@
...
+
+### Requires from P2P - P2P-I
+Message to nodes that remain connected are reliably delivered.
+
+```qnt reactor.gen.qnt
+<<REQ-GOSSIP-P2P-UNICAST>>
+```
+
+The neighbor set of $p$ is never larger than `maxCon(p)`.
+> TODO: can maxConn change in runtime?
```
cason, line 304: No.

spec/consensus/gossip/files.md
```diff
@@ -0,0 +1,125 @@
...
+
+## reactor.go
+- Entry point for the consensus reactor (State+Gossip)
+    - `NewReactor`
+    - `OnStart`
+    - `OnStop`
+    - `SwitchToConsensus`
+
+- Implements the Gossip Layer
+    - Peer management
```
cason, line 13: Implements the `p2p.Reactor` interface.

spec/consensus/gossip/files.md
```diff
@@ -0,0 +1,125 @@
...
+        - `InitPeer`
+        - `AddPeer`
+        - `RemovePeer`
+
+    - Message delivery
+        - `Receive`
```
cason, line 19: Implements the `p2p.Reactor` interface.

spec/consensus/gossip/files.md
```diff
@@ -0,0 +1,125 @@
...
+            - Processes messages delivered by the Gossip layer and updates consensus (`State`) and peer information (`PeerState`)
+            - Messages are received through multiple channels
+                - `StateChannel`: `NewRoundStepMessage`, `NewValidBlockMessage`, `HasVoteMessage`, `VoteSetMaj23Message`
```
cason, line 22: All these messages are from the GOSSIP layer.

spec/consensus/gossip/files.md
```diff
@@ -0,0 +1,125 @@
...
+                - `DataChannel`: `ProposalMessage`, `ProposalPOLMessage`, `BlockPartMessage`
+                - `VoteChannel`: `VoteMessage`
+                - `VoteBitsChannel`: `VoteSetBitsMessage`
+
+        - `PeerState` update
```
cason, line 27: The distinction here is:
- `Set*` methods refer to messages produced and processed by the consensus protocol implementation.
- `Apply*` methods refer to "control" messages, exchanged by the consensus reactors of connected peers; they don't have "payload" are not delivered to the consensus protocol implementation, that is why I used "control" for them.

spec/consensus/gossip/files.md
```diff
@@ -0,0 +1,125 @@
...
+            - `ApplyNewValidBlockMessage`
+            - `ApplyHasVoteMessage`
+            - `SetHasProposal`
+            - `ApplyProposalPOLMessage`
+            - `SetHasProposalBlockParts`
+            - `SetHasVote`
+            - `ApplyVoteSetBitsMessage`
+
+        - Peer information requests
+            - `VoteSetMaj23Message` (CODE TODO: Move handling from receive to helper function.)
```
cason, line 38: The response of which is a `VoteSetBitsMessage`.

spec/consensus/gossip/files.md
```diff
@@ -0,0 +1,125 @@
...
+
+    - [Consensus state](#state.go) update
```
cason, line 40: More specifically, messages are delivered to the consensus implementation, via the shared `peerMsgQueue`.

spec/consensus/gossip/files.md
```diff
@@ -0,0 +1,125 @@
...
+
+    - Consensus state monitoring
+        - `subscribeToBroadcastEvents` - event bus (Why not simply using channels and go-routines?)
+            - `broadcastNewRoundStepMessage`
+            - `broadcastNewValidBlockMessage`
+            - `broadcastHasVoteMessage`
```
cason, line 46: `EventVote` produced by the consensus implementation upon receiving a valid and "timely" (with respect with its current state) `Vote` message.

spec/consensus/gossip/files.md
```diff
@@ -0,0 +1,125 @@
...
+            - `ApplyVoteSetBitsMessage`
+
+        - Peer information requests
+            - `VoteSetMaj23Message` (CODE TODO: Move handling from receive to helper function.)
+
+    - [Consensus state](#state.go) update
+
+    - Consensus state monitoring
+        - `subscribeToBroadcastEvents` - event bus (Why not simply using channels and go-routines?)
+            - `broadcastNewRoundStepMessage`
```
cason, line 44: `EventNewRoundStep` produced by the consensus implementation.

spec/consensus/gossip/files.md
```diff
@@ -0,0 +1,125 @@
...
+            - `broadcastNewValidBlockMessage`
```
cason, line 45: `EventValidBlock` produced by the consensus implementation.

spec/consensus/gossip/files.md
```diff
@@ -0,0 +1,125 @@
...
+            - `broadcastHasVoteMessage`
+        - `go updateRoundStateRoutine`
```
cason, line 47: This is dirty workaround to prevent accessing the consensus state (and acquiring its lock) every time the gossip implementation needs to read it, to do it instead periodically (every 100 micros!?). In any case, directly or though this routine, the consensus state is (shallowly) copied.

cason: Having said that, I would not put that as "monitoring" routine: this is a workaround for actual shared memory communication between the protocol and the reactor.

spec/consensus/gossip/files.md
```diff
@@ -0,0 +1,125 @@
...
+
+
+    - [Consensus state](#state.go) gossip 
+        - `conR.gossipDataRoutine`
+            - `BlockPart`
+                 - if peers agree on block parts hash
+                 - if peer is lagging, through `gossipDataForCatchup`
+            - `Proposal`
+            - `ProposalPOL`
```
cason, line 56: This is a "control" message, with no payload. The goal is to inform which of the POLRound votes the node has to offer.

spec/consensus/gossip/files.md
```diff
@@ -0,0 +1,125 @@
...
+	    - `conR.gossipVotesRoutine`
+            - `PickSendVote` - sends a `Vote` of the specified set, which depends on the step of self and peer
+                - `gossipvotesForHeight` - if H matches
+                    - `LastCommit`
+                    - POL `Prevotes`
+                    - `Prevotes`
+                    - `Precommits`
+
+	    - `conR.queryMaj23Routine`
```
cason, line 65: Intended to handle Byzantine behavior, more specifically, the situation on which validators produce votes for different values in the same height, round, and step.

spec/consensus/gossip/files.md
```diff
@@ -0,0 +1,125 @@
...
+
+    - Message type definitions
+        - ProposalMessage
+            - Source: Round proposer
+            - Layer: State
+            - Step: Propose
+            - Once added to the roundState (`PeerState.PeerRoundState.Proposal*`) and accepted (validations passed and all [block parts](`BlockPartMessage`) have been received), will be gossiped.
```
cason, line 72: The link is broken.

cason: Not clear here whether the distinction between the algorithm `PROPOSAL` message and the code `Proposal` message has been considered.

spec/consensus/gossip/files.md
```diff
@@ -0,0 +1,125 @@
...
+
+        - ProposalPOLMessage
+        - BlockPartMessage
+        - VoteMessage
+        - HasVoteMessage
+        - VoteSetMaj23Message
+        - VoteSetBitsMessage
+            - Source: full node
+            - Step:
+            - Sent in response to `VoceSetMaj23Message`; 1-to-1 message.
```
cason, line 85: Every message here, except `Proposal`, `BlockPart`, and `Vote` messages are 1-to-1.

cason (commented):
Nice work.

I only covered in this review the English specification, as I am not yet enough familiar with Quint.

I've noticed that the interaction between CONS and GOSSIP was been updated and this is really the tricky part of this specification. In particular, the [Southbound Interaction - GOSSIP-I](https://github.com/cometbft/cometbft/blob/702fc04776623927dbdb185cc24498e83a9e66de/spec/consensus/gossip/gossip.md#southbound-interaction---gossip-i) section is a part that we should discuss in more detail, better if synchronously.

spec/consensus/gossip/gossip.md
```diff
@@ -0,0 +1,333 @@
...
+| If there exists a correct process $p$ such that $e \in t_p$, then, eventually, for every correct process $q$, $e \in t_q$ or there exists a correct process $r$ such that $\bar{e} \in t_r$.|
+
+> **Note**
+> Nodes may learn of an entry deletion before learning of its addition.
+
+In order to ensure convergence even in the presence of failures, the network must be connected in such a way to allow communication around any malicious nodes and to provide redundant paths between correct ones.
+This may not be feasible at all times, but should happen at least during periods in which the system is stable.
+
+In other words, during periods in which messages are timely delivered between correct processes long enough for multiple communication rounds to succeed, processes will identify and solve differences in their states.
+We call "long enough" $\Delta$.
```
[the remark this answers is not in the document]
lasarojc: Consensus algorithms such as Tendermint are always safe. 
It is a well known result [FLP] that Consensus cannot be guaranteed to always make progress in asynchronous distributed systems in which failures may happen. 
Assuming stable periods, bounded or unbounded, is one of the assumption an implementation can make to guarantee progress. It is the one used by algorithm like Paxos and Raft, even if encapsulated inside leader election oracles. It is a very reasonable assumption on most systems. And it is the case of Tendermint.

spec/consensus/gossip/gossip.md
lasarojc: It is impossible to guarantee progress of consensus in asynchronous distributed systems in which failures can happen. Period.

GST is used as tool to prove that if there are stable periods, the algorithm will progress, the separate algorithms from others that wouldn't progress even if there were stable periods.

No one expects GST to happen in practice. Still, implementations of Tendermint, Paxos, Raft and many others have been used for years, and they keep making progress during stable periods. If those stable periods cease, all these systems will stop working.

In conclusion, your WEB3 application working is a gift from the stable periods. You can rely on them.

spec/consensus/gossip/gossip.md
lasarojc: Thanks for you comment, but it will only make sense to try disprove what I intend here as a CRDT once I have formally defined the merge operator. Please keep in mind that this is a draft PR and is incomplete.

spec/consensus/gossip/gossip.md
```diff
@@ -0,0 +1,333 @@
...
+Even so, for efficiency, entries that are removed should not be added again, which may be prevented by keeping "tombstones" for the entries removed, which must be smaller than the entries themselves.
+We note the tombstone for an entry $e$ as $\bar{e}$.
+
+With time, even small tombstones may accrue and need to be garbage collected, which may cause the entry being added again, which continues to be safe.
+
+
+### Querying the Tuple Space
+
+The tuple space contains information regarding steps taken by validators during possibly many rounds of possibly many heights.
+Each entry has form $\lang Height, Round, Step, Validator, Value \rang$ and corresponds to the message Validator sent in Step of Round of Height; Value is a tuple of the message contents.
```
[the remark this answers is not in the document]
lasarojc: I never checked the rendering on GH itself. I will what would be the html equivalent and replace I the future.

spec/consensus/gossip/gossip.md
```diff
@@ -0,0 +1,333 @@
...
+
+We now formalize the requirements of the Tendermint algorithm in terms of an eventually consistent Tuple Space, which may be implemented using reliable communication (Gossip Communication), some best *best-effort* communication primitive that only delivers messages that are still useful, or some Gossip/Epidemic/Anti-Entropy approach for state convergence.
+All of these can be made to progress after GST but should also progress during smaller stability periods.
+
+|Eventual Convergence|
+|-----|
+| If there exists a correct process $p$ such that $e \in t_p$, then, eventually, for every correct process $q$, $e \in t_q$ or there exists a correct process $r$ such that $\bar{e} \in t_r$.|
+
+> **Note**
+> Nodes may learn of an entry deletion before learning of its addition.
```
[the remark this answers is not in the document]
lasarojc: that is the idea of bar{e}, to be a tombstone. I will clean this up at some point.

spec/consensus/gossip/gossip.md
08d2: > It is impossible to guarantee progress of consensus in asynchronous distributed systems in which failures can happen. Period.

Definitely! This is also established by e.g. CAP. You pick either availability (which weakens your consistency model if there is a partition) or consistency (which halts your system if there is a partition). You can have progress, or consistency, but not both. No problem! Systems like Paxos and Raft and I guess Tendermint pick consistency. But that means, in the presence of a partition, the system will necessarily halt. It will pick up again when the partition is healed, but how long that takes is unknowable. Concepts like stability periods or a GST are I guess different ways to characterize partition-free windows of time in a system, but those concepts are emergent and descriptive properties of that system, not guarantees, and actually unknowable at any individual node. Protocols execute at individual nodes and without whole-system knowledge, so those system-level concepts can't be part of a protocol definition.

This is understood by e.g. Raft and Paxos, neither of which define or rely on any notion of a stability period or a GST, or assume anything about stability in the guarantees they provide to their consumers.

In contrast, this spec asserts stuff like

> To ensure progress and termination, the algorithm assumes a GST after which communication is reliable and timely

which isn't really, er, possible. Certainly neither Raft nor Paxos provide this assurance. In fact no system can ensure both consistency and progress/termination. And a distributed system protocol cannot rely on an assumption that delivery will be reliable or timely within some time boundary. If your system model assumes a GST as specified here, it only means that the model becomes invalid when that assumption fails. Which it will.

spec/consensus/gossip/gossip.md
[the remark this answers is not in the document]
08d2: Neither Paxos nor Raft guarantee progress, nor do they assume any concept of a stable period at the protocol level.

spec/consensus/gossip/gossip.md
lasarojc: Let me repeat myself.

> GST is used as a tool to prove that if there are stable periods, the algorithm will progress, to separate algorithms from others that wouldn't progress even if there were stable periods.
> No one expects GST to happen in practice. Still, implementations of Tendermint, Paxos, Raft and many others have been used for years, and they keep making progress during stable periods. If those stable periods cease, all these systems will stop working.

Again, it is a tool to prove that under the right conditions, the algorithm progresses and If these conditions do not happen. Why do we do it then? To differentiate these algorithms (which Guerraoui et al. called indulgent) from algorithms that are safe but don't progress even under the right conditions (which are technically correct but useless). So we need to make these assumptions here and show that when they hold, the algorithm progresses.

WRT your statement

> This is understood by e.g. Raft and Paxos, neither of which define or rely on any notion of a stability period or a GST, or assume anything about stability in the guarantees they provide to their consumers.

Both Paxos and Raft make timing assumptions. For [example](https://lamport.azurewebsites.net/pubs/lamport-paxos.pdf)

> Progress could not be guaranteed unless enough legislators stayed in the Cham- ber for a long enough time. Because Paxon legislators were unwilling to curtail their outside activities, it was impossible to ensure that any decree would ever be passed. However, legislators were willing to guarantee that, while in the Chamber, they and their aides would act promptly on all parliamentary matters. This guar- antee allowed the Paxons to devise a parliamentary protocol satisfying the following progress condition.
> > If a majority of the legislators2 were in the Chamber and no one entered or left the Chamber for a sufficiently long period of time then any decree proposed by a legislator in the Chamber would be passed, and every decree that had been passed would appear in the ledger of every legislator in the Chamber.

and 

> Achieving the progress condition required that legislators be able to measure the passage of time, so they were given simple hourglass timers.

The fact that timing assumptions are often hidden behind Omega oracles, which ensure eventual stable leader election, makes it harder to perceive them, but they are there.

spec/consensus/gossip/gossip.md
lasarojc: That is fine. It will eventually be a 2P-Set or I will give it a different name. Remember that these are todo notes on a draft PR.

spec/consensus/gossip/gossip.md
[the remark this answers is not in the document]
lasarojc: see the comment on the following conversation.

spec/consensus/gossip/gossip.md
08d2: Does a Paxos node know itself whether or not it is a valid legislator, or that it is in the Chamber, for a given round?

spec/consensus/reactor/reactor.md
```diff
@@ -0,0 +1,413 @@
...
+However, while Gossip Communication is a sufficient condition for Tendermint BFT to terminate, it is not strictly necessary that all messages from correct processes reach all other correct processes.
+What is required is for either the message to be delivered or that, eventually, some newer message, with information that **supersedes** that in the first message, to be timely delivered.
+In Tendermint BFT, this property is seen, for example, when nodes ignore proposal messages from prior rounds.
+
+|Supersession|
+|----|
+| Given messages $m1$ and $m2$, we say that **$m2$ supersedes $m1$** if after receiving $m2$ a process would make at least as much progress as it would by receiving $m1$ and we note it as $m2.\text{SSS}(m1)$.   
+| Supersession is transitive, i.e., if $m3.\text{SSS}(m2)$ and $m2.\text{SSS}(m1)$, then $m3.\text{SSS}(m1)$
+
+> :clipboard: **TODO**: Better way of saying "would make at least as much progress"?
```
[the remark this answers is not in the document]
lasarojc: This is handled by using a CRDT, AFAIU, so closing for now.

spec/consensus/reactor/reactor.md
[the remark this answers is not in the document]
lasarojc: Not relevant anymore.

lasarojc:
instead of this large PR, breaking the (updated) changes into a series of smaller PRs

PR #76, opened 2023-01-05
kvindexer/0.34/bugfixes in range queries and result deduplication

jmalicevic:
Original PR tendermint/tendermint#9884

This PR fixes two bugs when matching attributes within events:
- When multiple events match in the same transaction or block, the query would return multiple times the same result. 
- When the query contains a height range and another range, the height range would get ignored under certain conditions. 
- A query containing both a height range and height equality would process them both. Now the equality is ignored when a range exists. 
- Queries containing `CONTAINS` and `EXISTS` would ignore the height requirement when `match.events = 1`.

There have been minor improvements to the readability of the code:
- As we need to know the height range or the requested height to match events at and both whether the height related conditions are the only ones (beside match.events), there is now a `heightInfo ` struct that encapsulates all of that. Additionally, spuradious checks on whether the length of the conditions is a certain number has been replaced with simple `bool` flags which are derived within the `dedupHeight` method.


sergio-mena:
Pushing a commit with the block indexing UTs I came up with...

state/txindex/kv/kv_test.go
```diff
@@ -314,19 +349,80 @@ func TestTxSearchOneTxWithMultipleSameTagsButDifferentValues(t *testing.T) {
...
+			q:     "match.events = 1 AND account.number CONTAINS 'bla'",
+			found: false,
+		},
+	}
+
 	ctx := context.Background()
 
-	results, err := indexer.Search(ctx, query.MustParse("account.number >= 1"))
-	assert.NoError(t, err)
+	for _, tc := range testCases {
```
sergio-mena, line 416: I refactored this to be able to add new TCs easily

state/txindex/kv/utils.go
```diff
@@ -66,16 +66,18 @@ func ParseEventSeqFromEventKey(key []byte) (int64, error) {
...
+	found := false
 	heightRangeExists := false
 	var heightCondition []query.Condition
 	heightInfo.onlyHeightEq = true
 	heightInfo.onlyHeightRange = true
 	for _, c := range conditions {
 		if c.CompositeKey == types.TxHeightKey {
 			if c.Op == query.OpEqual {
-				if heightRangeExists {
+				if heightRangeExists || found {
```
sergio-mena, line 77: Do we need to back/forward port this?

sergio-mena (approved):

state/txindex/kv/utils.go
jmalicevic: IT has to be included in the open PR for 0.37. I have it in my todo for that PR. As well as porting the UTs you added to this PR.

PR #77, opened 2023-01-05
v0.37.x:state/kvindexer: port 0.34 query fix

jmalicevic:
Original PR tendermint/tendermint#9876

This is a forward port of tendermint/tendermint#9759 to fix the querying behaviour where attributes should be returned only if they match within the same event. It will include fixes for the above mentioned PR: #76  and tendermint/tendermint#9877


.changelog/unreleased/bug-fixes/77-kvindexer-fix-evattribute-indexing.md
```diff
@@ -0,0 +1 @@
+- [state/kvindexer] \#77 Fixed the default behaviour of the kvindexer to index and query attributes by events in which they occur. In 0.34.25 this was mitigated by a separated RPC flag.  (@jmalicevic)
```
thanethomson, line 1:
```suggestion
- `[state/kvindexer]` \#77 Fixed the default behaviour of the kvindexer to index and query attributes by events in which they occur. In 0.34.25 this was mitigated by a separated RPC flag.  (@jmalicevic)
```

Otherwise it's interpreted as a Markdown link.

thanethomson (approved):

PR #78, opened 2023-01-05
rfc: Banning peers based on ResponseCheckTx

### LINKED ISSUE #66: mempool/p2p: Research implications of peer disconnect based on ResponseCheckTx
Original issue: tendermint/tendermint#9546

As mentioned in  tendermint/tendermint#7918 and tendermint/tendermint#2185, it might be of interest to disconnect from a peer when receiving transactions that could never have been valid. 

Before implementing this behaviour we want to look into more details of such a behaviour. Namely:

- The use case where a node sends transactions that could never have been valid;
- Implications on the security and correctness model of Tendermint (how does the application know that it is this particular peer that should be removed, is this scenario not covered already by other layers in Tendermint, etc.);
- What other changes in other parts of Tendermint should the implementation of this require;
- If solution/problem still seem valuable, propose an implementation path taking into account potential need to refactor and change the p2p layer itself to support this.

jmalicevic:
Original PR: tendermint/tendermint#9675. Please look at the comments from the original PR as there was a lot of discussion on this already. 

--- 

This RFC captures the changes required to implement the proposals in tendermint/tendermint#7918 , tendermint/tendermint#2185, tendermint/tendermint#6523. The gist of the document is that the mempool should be able to trigger baning of peers if they either send us transactions that fail too often, or transactions that fail due to a particular code that indicates that the application has determined a transaction to never be valid, in any state.

It closes #66 and if accepted will be followed by implementing the proposed changes.


thanethomson:
@jmalicevic and I were talking about this synchronously today and I was wondering about the case where an attacker discovers a particular transaction that they know would be _accepted, and therefore propagated_, by >1/3 of the voting power on the network, but rejected by the rest. This, as you mentioned before @sergio-mena, would result in halting the network for the period for which we blacklist "misconfigured" peers, because >1/3 of the voting power would be blacklisted by the remaining peers, right?

cason:
I find a little debatable the sentence, or concept, of a transaction being "accepted, and therefore propagated, by >1/3 of the voting power on the network". The reason is that a validator node is not really expected to propagate transactions. In fact, the goal of the mempool is to allow transactions to be submitted to nodes that are not validators, with some guarantees that those transactions will eventually reach a correct validator.

Notice that understand that the sentence, expression, concept above could be translated in practice to nodes that are controlled, because used as sentry nodes, by validators with 1/3+ of the voting power of the network. But also in this case, this doesn't mean that transactions will not eventually reach a correct validator, as the propagation of transactions is not, or at least has never been formalized as, a process that requires 2/3+ of the (mempool) participants to be correct...

jmalicevic:
But validators still propagate the transactions to their peers? Even though the tx propagation was not designed as them being the main gossipers? What Thane is saying I think is that, if a validator propagates the tx to a validator who will not accept it, the one that will not accept it will disconnect from the sender and thus potentially partition the network if this happens often enough?

Although at a high level, is this a setting that we should even account for, this is a very messy configuration and it is debatable whether it should be allowed to exist.

cason:
> if a validator propagates the tx to a validator

Validators are rarely connected to each other. Usually they are connected only to their sentry nodes, which should run the same code (version of the application) as the validator itself. So a validator should never receive a transaction that would be rejected by its application, as the transaction will almost always come from a node under its control, running the same code.

The discussion here, that is my opinion, should be restricted to public nodes, not including the validators.

hvanz:
> their sentry nodes, which should run the same code (version of the application) as the validator itself. So a validator should never receive a transaction that would be rejected by its application, as the transaction will almost always come from a node under its control, running the same code.

Apart from the same app code, sentry nodes and validators should also have the same app configuration. The [problem on the issue](https://github.com/cosmos/gaia/issues/2073#issuecomment-1404846489) mentioned above on the Gaia testnet was about having different values for `minimum-gas-price`. So because of this misconfiguration, in this case it could have happened that sentry nodes were accepting txs that their validator was rejecting.

thanethomson (approved):
Is there still anything outstanding to do here? If not, please renumber with an RFC number >= 100 and merge.

Remember, RFCs are useful for explorations of the problem space, which I think this RFC does really well. We don't need to make any decisions yet as to implementation - that can come later as our priorities allow, and this will serve as a valuable reference then if similar discussions come up again.

docs/rfc/rfc-026-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+### Current support for peer banning
+
+The p2p layer implements banning peers by marking them
+as bad and removing them from the list of peers to connect to for *at least* a predefined amount of time. This is done by calling the `MarkBad` routine implemented by the `Switch`. If the node does not set the amount of time to be banned, a default value is used. 
+Note that the timing parameter sets the lower bound for when a peer will be unbanned. 
+But the p2p layer will only try to connect to banned peers if the node is not sufficiently connected. Thus the node has no
+explicit control on when a reconnect attempt will be triggered.
+
+The application can blacklist peers via ABCI if the 
+[`filterPeers`](../../spec/abci/abci%2B%2B_app_requirements.md#peer-filtering)
```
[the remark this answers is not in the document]
cason: Are you sure that a flag is needed?

docs/rfc/rfc-026-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+specific design considerations. 
+
+## Discussion
+
+If this feature is to be implemented we need to clearly define the following:
+1. What does banning a peer mean:
+   1. A peer can be simply disconnected from. 
+   2. Peer is disconnected from and banned.
+   3. Conditions for the peer to be banned.
+2. If `CheckTx` signals a peer should be banned, retrieve the ID of peers to ban.
```
cason, line 96: This is available. If the message received by the mempool reactor is `e p2p.Envelope`, `e.Src.ID()` returns its ID.

docs/rfc/rfc-026-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+   3. Conditions for the peer to be banned.
+2. If `CheckTx` signals a peer should be banned, retrieve the ID of peers to ban.  
+3. Are there possible attack scenarios or unexpected behaviours by allowing this. 
+
+Any further mentions of `banning` will be agnostic to the actual way banning is implemented by the p2p layer. 
+
+
+### 1. What does banning a peer mean 
+
+Tendermint recognizes that peers can accept transactions into their mempool as valid but then when the state changes, they can become invalid.
```
cason, line 104:
```suggestion
CometBFT recognizes that peers can accept transactions into their mempool as valid but then when the state changes, they can become invalid. 
```

docs/rfc/rfc-026-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+
+A possible solution is to ban peers temporarily. In addition to the question on how long is temporarily, setting specific time limits for banning on a peer basis
+is currently not supported by the p2p layer.
+
+*Banning a peer in case of duplicate transactions*
+
+Currently, a peer can send the same valid (or invalid) transaction [multiple times](https://github.com/tendermint/tendermint/blob/ff0f98892f24aac11e46aeff2b6d2c0ad816701a/mempool/v0/clist_mempool.go#L247). Peers do not 
+gossip transactions to peers that have sent them that same transaction. But there is no check on whether 
+a node has already sent the same transaction to this peer before. There is also no check whether the transaction
+that is being gossiped is currently valid or not (assumint that invalid transactions could become valid).
```
cason, line 158:
```suggestion
that is being gossiped is currently valid or not (assuming that invalid transactions could become valid).
```

docs/rfc/rfc-026-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+ **Discussion**
+
+ The problem with supporting this scenario is the definition of the above mentioned parameters. It is very hard to estimate, at the Tendermint level, what these parameters should be. A possible solution is 
+ to allow the application to set these parameters. What is unclear, how will the application know that these parameters are not well set if, due to a bug or network problems, transactions start to fail? 
+The network could end up with all nodes banning everyone. How would an application developer know to debug this, what to look for?
+
+A possible solution is to ban peers temporarily. In addition to the question on how long is temporarily, setting specific time limits for banning on a peer basis
+is currently not supported by the p2p layer.
+
+*Banning a peer in case of duplicate transactions*
```
cason, line 153: Add a header level?

docs/rfc/rfc-026-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+
+#### **Banning for sending *never-valid* transactions**
+
+If a transaction fails since it could never have been valid, `CheckTx` returns a `ResponseCheckTx.code` 
+value greater than 1. In this case, the peer should be disconnected from and banned immediately without keeping count on how often 
+this has happened. 
+
+The question is whether this transaction should be kept track of in the cache? We can still store it in 
+the cache so that we don't run `CheckTx` on it again, but if this peer is immediately banned, maybe there is no need
+to store its information.
```
cason, line 174: The same txs can be received again from other peer. That is the point of keeping them on the cache, actually.

docs/rfc/rfc-026-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+we might need more information on the actual reason for rejection. This could be done by an additional set of response codes provided by the application. 
+
+### 2. Choosing the peer to ban
+
+Each transaction gossiped contains the ID of the peer that sent that transaction. Upon receiving a transaction,
+a node saves the peer ID of the peer(s) that have sent it. As each peer had to have
+run `CheckTx` on this transaction before adding it to its own mempool, we can assume this peer
+can be held accountable for the validity of transactions it gossips. Invalid transactions are kept only in the mempool
+cache and thus not gossiped. 
+As nodes have to complete a cryptographic handshake at the p2p layer, Tendermint guarantees that a malicious peer
```
cason, line 186:
```suggestion
As nodes have to complete a cryptographic handshake at the p2p layer, CometBFT guarantees that a malicious peer
```

docs/rfc/rfc-026-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+cannot lie about who the sender of the transaction is.
```
cason, line 187: This would actually be impersonating another node.

docs/rfc/rfc-026-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+a node saves the peer ID of the peer(s) that have sent it. As each peer had to have
+run `CheckTx` on this transaction before adding it to its own mempool, we can assume this peer
+can be held accountable for the validity of transactions it gossips. Invalid transactions are kept only in the mempool
+cache and thus not gossiped. 
+As nodes have to complete a cryptographic handshake at the p2p layer, Tendermint guarantees that a malicious peer
+cannot lie about who the sender of the transaction is. 
+
+*Transactions received from users*
+
+For transactions submitted via `broadcastTxCommit`, the `SenderID` field is empty.
```
cason, line 191: There are 3 `broadcastTx` RPC endpoints.

docs/rfc/rfc-026-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+**Note**  Do we have mechanisms in place to handle cases when `broadcastTxCommit` submits
+failing transactions (can this be a form of attack)?
+
+### 3. Attack scenarios
+
+While an attack by simply banning peers on failing `CheckTx` is hard to imagine, as the incentive for doing so is not clear, there are considerations with regards to the current mempool gossip implementation.
+
+Should we keep transactions that could never have been valid in the cache? Assuming that receiving such transactions is rare, and the peer that sent them is banned, do we need to occupy space in the mempool cache with these transactions?
+
+- What if nodes run different versions of Tendermint and banning is not supported in one of the versions?
```
cason, line 202:
```suggestion
- What if nodes run different versions of CometBFT and banning is not supported in one of the versions? 
```

docs/rfc/rfc-026-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+
+In summary, if we want to support banning peers based on the frequency with which they submit invalid transactions, we need to have **additional datastructures**:
+1. One to keep track of past invalid transactions
+2. A datastructure to differentiate between valid and invalid *cached* transactions. If the `KeepInvalidTxsInCache` configuration parameter is not set, this datastructure
+   is not needed. 
+
+We propose two ways to implement peer banning based on the result of `CheckTx`:
+
+1. Introduce banning when transactions are received
+2. Adapt the recheck logic to support this
```
cason, line 240: The problem of 2. is that we don't know who to punish/ban in the case the transaction has been received from multiple peers.

docs/rfc/rfc-026-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+if !mem.config.KeepInvalidTxsInCache || r.CheckTx.Code == abci.NeverValid {
+  // remove from cache (it might be good later or is never valid, we'll most likely don't see it again)
+    mem.cache.Remove(tx)
+}
+```
+
+- We do keep it in the cache as long as possible to avoid running `CheckTX` on it because we know, for sure, that it will never be valid. As it is rare enough, it 
+might not take that much space. In this case though, as we ban the sending peer immediately, we can save space by not storing peer information for this transaction. 
+
+The question is which one is more costly, doing `CheckTx` more then once, or keeping an extra entry in the cache?
```
cason, line 291: The space used in the cache is not really a problem. The cache has a fixed size, and should store hashes. The point here is that we are using one of the limited entries of the cache for an invalid transaction. Notice that as the time passes, all entries in the cache are replaced by more recently seen ones.

docs/rfc/rfc-026-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+We propose two ways to implement peer banning based on the result of `CheckTx`:
+
+1. Introduce banning when transactions are received
+2. Adapt the recheck logic to support this
+
+
+
+**Peer banning when transactions are received**
+
+If a transaction fails `CheckTx` the [first time it is seen](https://github.com/tendermint/tendermint/blob/ff0f98892f24aac11e46aeff2b6d2c0ad816701a/mempool/v0/clist_mempool.go#L409),  the peer can be banned right there:
```
cason, line 246: Update link to the code.

docs/rfc/rfc-026-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+```
+
+- We do keep it in the cache as long as possible to avoid running `CheckTX` on it because we know, for sure, that it will never be valid. As it is rare enough, it 
+might not take that much space. In this case though, as we ban the sending peer immediately, we can save space by not storing peer information for this transaction. 
+
+The question is which one is more costly, doing `CheckTx` more then once, or keeping an extra entry in the cache?
+
+
+
+As said, this code will [never be executed](https://github.com/tendermint/tendermint/blob/ff0f98892f24aac11e46aeff2b6d2c0ad816701a/mempool/v0/clist_mempool.go#L239) for transactions whose signature is found
```
cason, line 295: Update link to the code.

cason, line 295: By signature you mean hash?

cason (commented):
Left some suggestions, in particular renaming to CometBFT.

docs/rfc/rfc-026-p2p-bad-peers-checktx.md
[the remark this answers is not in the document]
jmalicevic: Are you sure we should refer to main rather than to the relative path? In  case the spec changes on main at some point..although I guess the RFCs will always be on main?

jmalicevic:
@thanethomson , @cason , I will document our discussion and current take on things and merge this RFC. If we decide to implement it, I will edit it with the corresponding info. But I have one more point I would like to raise before closing it. 

From the current discussion ,it seems that implementing it has too many concerns and we did not get sufficient concrete examples where it would benefit. 

However, based on the comment made by @sergio-mena  above, I was thinking that adding at least a comment on a transaction failing `CheckTx` might help. But it is absolutely possible for a node to have gossiped a tx that became invalid due to a state change, no? This would create a lot of noise again.  So we  distinguish two cases:
a) a tx fails `CheckTx` because of a state change and 
b) a tx  fails due to a misconfiguration.

To facilitate the debugging in a case of b), we could introduce a special code for a transaction that could never have been valid (say code 999) but not ban peers as a result, rather output this log message.

docs/rfc/rfc-026-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+### Current state of mempool/p2p interaction
+
+Transactions received from a peer are handled within the [`Receive`](https://github.com/tendermint/tendermint/blob/ff0f98892f24aac11e46aeff2b6d2c0ad816701a/mempool/v0/reactor.go#L158) routine. 
+
+Currently, the mempool triggers a disconnect from a peer in the case of the following errors:
+
+  - [Unknown message type](https://github.com/tendermint/tendermint/blob/ff0f98892f24aac11e46aeff2b6d2c0ad816701a/mempool/v0/reactor.go#L184)
+
+However, disconnecting from a peer is not the same as banning the peer. The p2p layer will close the connecton but 
+the peer can reconnect without any penalty, and if it as a persistent peer, a reconnect will be initiated
```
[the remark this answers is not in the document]
jmalicevic: Clarified, will push new version shortly.

docs/rfc/rfc-026-p2p-bad-peers-checktx.md
jmalicevic: Yes, this is further clarified below when this is discussed

docs/rfc/rfc-026-p2p-bad-peers-checktx.md
jmalicevic: Yes -  changed the wording

docs/rfc/rfc-101-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+ Whenever a transaction fails, if the `now - lastFailure <= failureResetInterval`, we increment the `numFailures` for this particular peer and set the `lastFailure` to `now`. 
+ Otherwise, we set `lastFailure` to `now` and set `numFailures`  to 1. 
+ Once the value for `numFailures` for a peer reaches `maxAllowedFailures`, the peer is disconnected from and banned. 
+
+The reason for this logic is as follows: We deem it acceptable if every now and then a peer sends us an invalid transaction.
+ But if this happens very frequently, then this behaviour can be considered as spamming and we want to disconnect from the peer. 
+  
+ **Discussion**
+
+ The problem with supporting this scenario is the definition of the above mentioned parameters. It is very hard to estimate, at the CometBFT level, what these parameters should be. A possible solution is
```
ValarDragon, line 146: I agree, I'm generally concerned about how we want to reason about this. I think its probably useful to iterate towards this, but also expose configs so node operators can adjust.

Also this type of ban, should likely always be timebound

docs/rfc/rfc-101-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+
+#### **Banning for sending *never-valid* transactions**
+
+If a transaction fails since it could never have been valid, `CheckTx` returns a `ResponseCheckTx.code` 
+value greater than 1. In this case, the peer should be disconnected from and banned immediately without keeping count on how often 
+this has happened. 
+
+The question is whether this transaction should be kept track of in the cache? We can still store it in 
+the cache so that we don't run `CheckTx` on it again, but if this peer is immediately banned, maybe there is no need
+to store its information.
```
ValarDragon, line 174: My impression is it doesn't need to be in the cache, since anyone who sends it has a notable punishment, and must keep establishing new peer connections to DOS.

If they were doing a significant effort to establish new peer connections, I believe that with ease, they could create new malicious payloads with costly compute that get banned (rather than reuse the same payload)

ValarDragon (approved):
Awesome! Glad peer banning for never valid txs seems like its happening! Can start mitigating a lot of bad signature DOS concerns!

jmalicevic:
Thanks for the reviews @ValarDragon ! Can you tell us more abut the DOS you mentioned? Have you seen this often, nodes sending TXs with bad signatures?

docs/rfc/rfc-101-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+**Note**  Do we have mechanisms in place to handle cases when `broadcastTxCommit` submits
+failing transactions (can this be a form of attack)?
+
+### 3. Attack scenarios
+
+While an attack by simply banning peers on failing `CheckTx` is hard to imagine, as the incentive for doing so is not clear, there are considerations with regards to the current mempool gossip implementation.
+
+Should we keep transactions that could never have been valid in the cache? Assuming that receiving such transactions is rare, and the peer that sent them is banned, do we need to occupy space in the mempool cache with these transactions?
+
+- What if nodes run different versions of CometBFT and banning is not supported in one of the versions?
```
lasarojc, line 213: Being able to ban bad actors is a desirable feature and serves as incentive to upgrade.

docs/rfc/rfc-101-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+ 
+
+## Implementation considerations
+
+**Indicating a new type of `CheckTx` failure**
+
+The initial proposal is to reserve a special response code to indicate that the transaction could never have been valid. 
+Due to concerns of this being a breaking change for applications that have already reserved this code for internal
+purposes, there is an alternative implementation: expanding `ResponseCheckTx` with an additional field. 
+This field `neverValidTx` would be `false` by default. If a transaction could never have been valid,
```
lasarojc, line 225: instead of a boolean, maybe use an extended error code, which may be used for indicating other scenarios that could show up later.

docs/rfc/rfc-101-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+```
+
+- We do keep it in the cache as long as possible to avoid running `CheckTX` on it because we know, for sure, that it will never be valid. As it is rare enough, it 
+might not take that much space. In this case though, as we ban the sending peer immediately, we can save space by not storing peer information for this transaction. 
+
+The question is which one is more costly, doing `CheckTx` more then once, or keeping an extra entry in the cache?
+
+
+
+As said, this code will [never be executed](https://github.com/tendermint/tendermint/blob/ff0f98892f24aac11e46aeff2b6d2c0ad816701a/mempool/v0/clist_mempool.go#L239) for transactions whose signature is found
```
lasarojc:
```suggestion
As said, this code will [never be executed](https://github.com/cometbft/cometbft/blob/ff0f98892f24aac11e46aeff2b6d2c0ad816701a/mempool/v0/clist_mempool.go#L239) for transactions whose signature is found
```

docs/rfc/rfc-101-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+in the cache. 
+Instead of remembering the cached transactions, we could have had a valid/invalid bit per transaction within the cache. As transactions themselves do not
```
lasarojc, line 317: can we keep the error code from the first check? this way, if the transactions shows up again we can avoid rechecking it and still returning the same error code as before.

docs/rfc/rfc-101-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+**Implementing peer banning on recheck**
+
+Currently the recheck logic confirmes whether once **valid** transactions, 
+, where `ResponseCheckTx.code == 0`, are still valid. 
+
+As this logic loops through the transactions in any case, we can leverage it to check whether we can ban peers. 
+However, this approach has several downsides:
+- It is not timely enough. Recheck is executed after a block is committed, leaving room for a bad peer to send 
+us transactions the entire time between two blocks. 
+- If we want to keep track of when peers sent us a traansaction and punish them only if the misbehaviour happens
```
lasarojc, line 405:
```suggestion
- If we want to keep track of when peers sent us a transaction and punish them only if the misbehaviour happens 
```

docs/rfc/rfc-101-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+Currently the recheck logic confirmes whether once **valid** transactions, 
+, where `ResponseCheckTx.code == 0`, are still valid. 
+
+As this logic loops through the transactions in any case, we can leverage it to check whether we can ban peers. 
+However, this approach has several downsides:
+- It is not timely enough. Recheck is executed after a block is committed, leaving room for a bad peer to send 
+us transactions the entire time between two blocks. 
+- If we want to keep track of when peers sent us a traansaction and punish them only if the misbehaviour happens 
+frequently enough, this approach makes it hard to keep track of when exactly was a transaction submitted. 
+- Rechecking if optional and node operators can disable it.
```
lasarojc, line 407:
```suggestion
- Rechecking is optional and node operators can disable it. 
```

docs/rfc/rfc-101-p2p-bad-peers-checktx.md
```diff
@@ -0,0 +1,436 @@
...
+- If we want to keep track of when peers sent us a traansaction and punish them only if the misbehaviour happens 
+frequently enough, this approach makes it hard to keep track of when exactly was a transaction submitted. 
+- Rechecking if optional and node operators can disable it. 
+- Furthermore, rechecking is a node local configuration parameter. This means that, some nodes might be performing this check,
+  while others will be unaware of this. 
+
+On the plus side this would avoid adding new logic to the mempool caching mechanism and keeping additional information about
+transaction validity. But we would still have to keep the information on peers and the frequency at which they send us bad transactions.
+
+Transactions that became invalid on recheck should not be cause for peer banning as they have not been gossiped as invalid transactions.
```
lasarojc, line 414: This is not necessarily true. With prepare and process proposal, bad actors could send transactions that are valid but not ripe for proposal, which would always become invalid later. I think only the app can be the judge of it.

Issue #46, opened 2022-12-23
Establish and implement the relevant metrics to understand storage workloads


lasarojc:
Was https://github.com/tendermint/tendermint/issues/9773

We need to identify the set of metrics to understand the storage workloads of CometBFT. The metrics should help us identify:

* The access patterns (sequential / random access)
* How often the data is read/written
* Who reads/writes the data
* Is the data accessed by multiple components or just one.
* How much of the total height time is spent in storage - on a small network, on a big network? (Is storage a bottleneck? )

Open questions: Do we want information on which CometBFT Blockstore / StateStore method call triggered the access or is read/write/delete count and timing enough?

Draft implementation: https://github.com/tendermint/tendermint/pull/9774

### DoD
* [ ]  Set of candidate metrics is established, discussed and well understood.
* [ ]  The identified metrics are added into the Tendermint codebase.

jmalicevic:
For Q4 2023, we will focus on:
- Storage access time (last point in the issue description) 
- Which data is most frequently read 
- The amount of storage taken by different workloads with and without pruning
- The throughout with and without pruning - This was derived from a call with Injective whose users' latency was heavily impacted by pruning being enabled. 

This list might be altered after an in person meeting of the team deciding on the exact measurments that will help us achieve our Q4 goals. 

The main result of this testing should be :
- Showing that the impact of pruning has improved - compaction works
- Database access times have decreased
- Impact of the new pruning mechanism on throughput
- Impact of different key representation + new pruning mechanism on throughput.

melekes:
> The throughout with and without pruning - This was derived from a call with Injective whose users' latency was heavily impacted by pruning being enabled.

The throughput can be hard to measure as it depends on many factors. Is it better to measure the duration of pruning (if possible)? This is similar to how the Golang core team optimised their garbage collector by looking at the duration of GC pauses.

PR #79, opened 2023-01-05
metrics: add a candidate set of metrics for measuring db access times.

jmalicevic:
Original PR: tendermint/tendermint#9774.

This PR addresses #46.  It provides a proposed implementation of metrics to measure the DB access times of different Tendermint components.
This is still work in progress. Do not merge!


jmalicevic:
Closing as it is superceeded by #1974

PR #80, opened 2023-01-05
WIP Spec: Draft for top-level p2p spec

jmalicevic:
Original PR: tendermint/tendermint#9732 

-- 
This is the beginning of a high-level p2p spec addressing #20  for discussions. I started top down trying to understand what p2p is supposed to do rather than how p2p should do things. This collects findings from several discussion.


spec/p2p/reactor-survey.md
```diff
@@ -0,0 +1,247 @@
...
+  > As a result of these two properties, once a transaction is known to all validators (present and future), it 
+cannot be overtaken by infinitely many transaction that arrived in the system later (there
+is still some possible overtaking due to faulty proposers); cf. no starvation.
+
+
+We arrive at the requirement on the mempool: 
+
+#### [CM-REQ-MEMP-GOSSIP.0]
+
+1. every transaction submitted to a correct full node must eventually be included in the mempool of a correct validator
```
josef-widder, line 154: We should figure out whether we want the same guarantees for transactions submitted to any fullnode, of e.g. for MEV, you submit your transaction to a special full node (or a set of full nodes) and you get better guarantees. Best effort transactions vs. transactions with guarantees.
Might be a question for the mempool spec.

josef-widder: Daniel: Define stable nodes vs. transient nodes is important. We can give guarantees wrt stable nodes, while we allow transient nodes.

spec/p2p/reactor-survey.md
```diff
@@ -0,0 +1,247 @@
...
+
+> Note that in this document, we want to capture (an overapproximation of) the requirements the mempool puts on p2p. The notions of "pending" and "valid" seem mostly to be conditions that the mempool can use to drop transactions (and exempt them from beeing spread over the network). Thus for the purpose of this document we will largely ignore them (they should eventually be addressed [here](https://github.com/cometbft/cometbft/issues/223)). 
+
+As 
+- full nodes add transactions to the mempool following client requests, and
+- validators read transactions from the mempool to compose proposed blocks,
+
+the mentioned end-to-end requirements on transactions translate into requirements 
+on the mempool, e.g., every transaction that is submitted to a full node should 
+eventually be present at the mempool of a validator/proposer.
```
08d2, line 112: No way to guarantee this property.

spec/p2p/reactor-survey.md
```diff
@@ -0,0 +1,247 @@
...
+on the mempool, e.g., every transaction that is submitted to a full node should 
+eventually be present at the mempool of a validator/proposer. 
+
+For the purpose of this document (specifying what the mempool needs from p2p), 
+we do not strive to precisely state the end-to-end requirements. Rather, we 
+will show that the arguably strongest requirement on the mempool do translate roughly
+into requirements on the p2p layer that are similar to those of the
+consensus reactor.
+
+- end-to-end requirement: "every transaction submitted should be eventually put into a block"
```
08d2, line 120: Not possible to guarantee.

spec/p2p/reactor-survey.md
```diff
@@ -0,0 +1,247 @@
...
+the transaction is proposed by every correct proposer in every height until it is committed
+into a block (assuming infinitely often correct proposers get their proposal through)  
+> leads to
+- current solution to address this requirement: 
+    - the mempool of correct full nodes is organized as a list. 
+    This list can be understood as a shared data structure with relaxed semantics, that is,
+    not all nodes need to agree on the order of the elements, but we have a property that 
+    approximately says "if at time *t* a transaction tx1 is in the list of **all** nodes, and a 
+    client submits transaction tx2 after time t, then it is never the case the tx2 is before
+    tx1 in a list of a correct validator.
```
08d2, line 132: This property is not provided by the current spec or implementation. It is not possible to guarantee in a weakly-connected, eventually-consistent, gossip-based network as is defined by the p2p layer.

spec/p2p/reactor-survey.md
08d2, line 154: It is definitely not true that every transaction submitted to a correct full node will, or can, be included in the mempool of any other arbitrary "correct" validator.

spec/p2p/reactor-survey.md
```diff
@@ -0,0 +1,247 @@
...
+cannot be overtaken by infinitely many transaction that arrived in the system later (there
+is still some possible overtaking due to faulty proposers); cf. no starvation.
+
+
+We arrive at the requirement on the mempool: 
+
+#### [CM-REQ-MEMP-GOSSIP.0]
+
+1. every transaction submitted to a correct full node must eventually be included in the mempool of a correct validator
+2. if a transaction is included in the mempool of a correct validator eventually it is included in the mempool of all correct validators
```
08d2, line 155: This is also definitely not true.

spec/p2p/reactor-survey.md
```diff
@@ -0,0 +1,247 @@
...
+1. when evidence is put into a block (similar to the mempool we may assume that correct proposers always put all the evidence they are aware of in a block, but CometBFT cannot enforce that)
+1. when evidence is submitted
+
+In face of this uncertainty, there are only few reasonable requirement we can postulate
+
+#### [CM-PROTOCOL-ER-COMM.0]
+1. An evidence transaction reaches all validator nodes as "fast as possible".
+1. Evidence is not written into a block after it has expired
+
+> With more control over the four points, in principle we might define a real-time property which formalizes "if evidence is submitted at least a day before it expires, then evidence is written into a block before it expires".
```
08d2, line 202: Nothing guarantees that evidence submitted at some wall-clock time X will be written into a block before some later wall-clock time Y.

spec/p2p/reactor-survey.md
08d2: It is important to point out that gossipped information has no delivery guarantees whatsoever. No message that goes through the p2p layer is, by itself, reliably delivered to any specific destination. Gossip does not guarantee that a message is reliably delivered to any/every node in any tine frame. Only the information which goes through consensus can be assumed as true.

08d2 (commented):
:|

spec/p2p/reactor-survey.md
josef-widder: Thanks for looking at the document and giving feedback. Please note that this is a working document that should describe requirements one may expect from the reactors in order to formalize, in turn, what the reactors expect from the p2p system below. 

By describing stronger properties (than those we actually need) here, and deriving properties for p2p from them, we can eventually specify a p2p system that may do a bit more than is needed and thus suffices for what the reactors need in reality. We had noted that on line 104, which says 

> Note that in this document, we want to capture (an overapproximation of) the requirements the mempool puts on p2p. 

We will try to make this clearer in a revision to avoid future confusion.

spec/p2p/reactor-survey.md
josef-widder: See [this comment](https://github.com/cometbft/cometbft/pull/80/files/f2f73d279306cc15b9460b1e217a82ace92fd09d#r1158159787).

spec/p2p/reactor-survey.md
josef-widder: See [this comment](https://github.com/cometbft/cometbft/pull/80/files/f2f73d279306cc15b9460b1e217a82ace92fd09d#r1158159787)

spec/p2p/reactor-survey.md
josef-widder: I agree. The statement also doesn't contradict this.

spec/p2p/reactor-survey.md
08d2: I think this may boil down to the definition of a "correct" node.  If "correct" means that a node will reliably receive any message gossiped from some other "correct" node within some time bound, then "correct" is relative to an observer, it has no well-defined value in general, and it can't be used to assert stuff like what is asserted here.

If you want to define "correct" in this way, and use that definition to say things like "it is never the case that tx2 is before tx1 among correct nodes", then you're defining a node that suffers any kind of network error to be incorrect, which is unknowable by the protocol as evaluated at any given node.

It is entirely possible that a transaction arrives in the mempool of only (say) 70% of the validators in a network, and if the proposer is one of those 70% of validators that transaction gets put into a proposed block, and that block is validated and committed, all without that transaction ever entering the mempool of (say) 30% of the validators of that network. Those 30% validators can fail to receive the gossip messages that would put the transaction in their mempool, while at the same time successfully receiving (and approving) the consensus messages required to commit a block which included that transaction to the chain.

So "if at time t1 a transaction tx1 is in the list of all nodes" is never possible to assume, because there is no situation where any given transaction is guaranteed to be known by all validators, unless it is put through consensus (and even then it is only guaranteed to be known to the approving quorum); "and a client submits transaction tx2 after time t1" is not something that can even be evaluated, as times t1 and t2 are not reliably comparable, and even if they were that comparison can only be made by nodes which know about both tx1 and tx2, which are not guaranteed to be a quorum (or even a majority) of nodes at any time; and finally if "it is never the case that tx2 is before tx1 in a list of a correct validator" defines order as time order, which is not necessarily true (app side mempools, and alternative prepare proposal hooks, can re-order mempool tx order by any criteria) even if time is reliable (which it isn't, afaiu).

spec/p2p/reactor-survey.md
08d2: Gossip provides no guarantees about delivery that can be assumed by the protocol at any specific node or time. It guarantees _eventual_ delivery, over an unbounded time window, yes, but that doesn't allow the protocol as executed at any specific node to assert specific assumptions like a GST, or that the mempool will definitely contain any specific set of elements. Gossipping message M1 from node N1 at time T1 does not guarantee that M1 will be received by node N2 at time T2 within any time limit. T2 could be T1 plus 1s, or T1 plus 1h, or T1 plus 1d, or T1 plus 1w.

spec/p2p/reactor-survey.md
08d2: That's all fine, I think. But it is not the case that "every transaction that is submitted to a full node should eventually be present at the mempool of a validator/proposer".

spec/p2p/reactor-survey.md
08d2: Does the statement not assert that evidence submitted at wall-clock time X (where X is at least a day before the expiration time of that evidence) will be written into a block before some later wall-clock time Y (where Y is the expiration time of the evidence)"?

josef-widder:
I close as the target location of the text is unclear. I have moved relevant parts into our new knowledge base.

Issue #81, opened 2023-01-05
rpc: Add gRPC support


thanethomson:
At present, we support a flavor of JSON-RPC as one of the primary ways of integrating with a running node.

From discussions with users (e.g. @hdevalence), it's clear there's a need for a full gRPC interface to a node. This would dramatically simplify client generation for interaction with a node, and would help get around our very odd (inherited) JSON serialization quirks.

There's some context provided by [ADR-057](https://github.com/cometbft/cometbft/blob/main/docs/architecture/tendermint-core/adr-057-RPC.md), but no clear decision was documented in that ADR. Previously, the team had decided to deprecate and remove gRPC support (see tendermint/tendermint#7121 and tendermint/tendermint#9683), but that decision has now been reversed.

## Definition of Done

There are several deliverables associated with this issue, after which it could be considered done:

- [x] #650 
- [x] #850
- [ ] #2693
   - After it's on `main`, we can consider whether we should backport this to older releases.

Original issue: tendermint/tendermint#9751

thanethomson:
Some of the endpoints we should aim to provide in the first iteration of our API should at least provide the same functionality as those exposed by Penumbra's `TendermintProxyService` here: https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.client.v1alpha1#penumbra.client.v1alpha1.TendermintProxyService

andynog:
If the work in #2693 is completed then this tracking issue can be closed.

PR #82, opened 2023-01-05
ADR-101: Data Companion Pull API

thanethomson:
:book: [Rendered](https://github.com/cometbft/cometbft/blob/thane/adr-084-data-companion-pull-api/docs/architecture/adr-101-data-companion-pull-api.md)

A competing alternative to #73, inspired by https://github.com/tendermint/tendermint/pull/9437#discussion_r1034037844

To be clear, if we decide to move forward with the "data companion" architecture, only one of either ADR-100 or ADR-101 must be chosen and implemented.

See also the standalone RPC solution proposed in ADR 102 (#658) as a reference implementation of a specific companion service.

---

#### PR checklist

      documentation updates needed


thanethomson:
> One area of concern I would have here is that this change in 82/84 would affect exchange integration with Cosmos chains.

How do you see this as affecting exchange integration?

lasarojc (approved):

tac0turtle:
> There isn't really a team owning Rosetta integration but some risk assessment of the impact on Rosetta/exchange integration is needed.

rosetta has a maintenance team, sdk team along side zondax are the maintainers of it. We will support changes in this sense. If the rosetta needs to become stateful we can support that.

thanethomson:
Given the greater simplicity of this approach compared to #73, I'm currently leaning toward this pull-based approach.

Are there any strong arguments in favor of the push-based approach in #73?

adizere (commented):
Great work Thane!

Most of my comments are nits and questions to help my understanding.

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,358 @@
...
+
+1. **Data companion disabled**: Here, the node prunes as per normal based on the
+   `retain_height` parameter supplied by the application via the
+   [`Commit`][abci-commit] ABCI response.
+2. **Data companion enabled**: Here, the node prunes _blocks_ based on the lower
+   of the retain heights specified by the application and the data companion,
+   and the node prunes _block results_ based on the retain height set by the
+   data companion.
+
+Enabling the `discard_abci_responses` flag under the `[storage]` section in the
```
sergio-mena, line 106: I think we should go on and deprecate `discard_abci_responses` as part of this

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,358 @@
...
+    //
+    // The lower of this retain height and that set by the application in its
+    // Commit response will be used by the node to determine which heights' data
+    // can be pruned.
+    rpc SetRetainHeight(SetRetainHeightRequest) returns (SetRetainHeightResponse) {}
+
+    // GetRetainHeight returns the retain height set by the companion and that
+    // set by the application. This can give the companion an indication as to
+    // which heights' data are currently available.
+    rpc GetRetainHeight(GetRetainHeightRequest) returns (GetRetainHeightResponse) {}
```
sergio-mena, line 234: How are these two methods behave in the presence of a crash-recovery? Can the latest values returned by CometBFT be trusted also in that case?

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,358 @@
...
+    // The retain height as set by the data companion.
+    int64 data_companion_retain_height = 1;
+    // The retain height as set by the ABCI application.
+    int64 app_retain_height = 2;
+}
+```
+
+With this API design, it is technically possible for an integrator to attach
+multiple data companions to the node, but only one of their retain heights will
+be applied.
```
sergio-mena, line 255: So, semantically (as far as the interface is concerned) it is still "at most one"

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,358 @@
...
+syntax = "proto3";
+
+package tendermint.block_service.v1;
+
+import "tendermint/abci/types.proto";
+import "tendermint/types/types.proto";
+import "tendermint/types/block.proto";
+
+// BlockService provides information about blocks.
+service BlockService {
```
sergio-mena, line 151: How does the companion know that there are new blocks it can process?

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,358 @@
...
+method = "basic"
+
+# Path to the file containing basic authentication credentials to access the
+# data companion service. If the data companion is enabled and this password
+# file is not supplied, or the file does not exist, or the file is of an invalid
+# format, the node will fail to start.
+#
+# See https://httpd.apache.org/docs/current/programs/htpasswd.html for details
+# on how to create/configure this file.
+password_file = "/path/to/.htpasswd"
```
sergio-mena, line 312: In terms of UX, I'm not sure what is more cumbersome, whether this, or simply having the companion gRPC separate and bound to _localhost_.

sergio-mena (commented):
Thanks for putting this together. Most comments are minor/editorial (didn't signal all needed renaming to CometBFT).
There are, though, a few other comments on things I'd like to understand more clearly how they will be handled.

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,358 @@
...
+
+## Context
+
+Following from the discussion around the development of [ADR 100][adr-100], an
+alternative model is proposed here for offloading certain data from nodes to a
+"data companion". This alternative model inverts the control of the data
+offloading process, when compared to ADR 082, from the node to the data
+companion.
+
+Overall, this approach provides slightly weaker guarantees than that of ADR 082,
```
jmalicevic, line 20:
```suggestion
Overall, this approach provides slightly weaker guarantees than that of ADR 100,
```

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,358 @@
...
+### Pruning Behaviour
+
+There are two "modes" of pruning that need to be supported by the node:
+
+1. **Data companion disabled**: Here, the node prunes as per normal based on the
+   `retain_height` parameter supplied by the application via the
+   [`Commit`][abci-commit] ABCI response.
+2. **Data companion enabled**: Here, the node prunes _blocks_ based on the lower
+   of the retain heights specified by the application and the data companion,
+   and the node prunes _block results_ based on the retain height set by the
```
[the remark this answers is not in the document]
jmalicevic: We should be aware that pruning is applied only to the block store (not the state store, nor the indexer for example). So the data companion could, in the way things are right now retrieve data from other stores... We should also understand whether we have use cases where people are fine with block pruning but still need events around (as they were so far stored in the indexer). If we have such users and we do not store things in the indexer, we might break them. What we could do, is separate the block store from the event store, and delete only the block store but not the events.

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,358 @@
...
+   data companion.
+
+Enabling the `discard_abci_responses` flag under the `[storage]` section in the
+configuration is incompatible with enabling a data companion. If
+`storage.discard_abci_responses` and `data_companion.enabled` are both `true`,
+then the node _must_ fail to start.
+
+### gRPC API
+
+At the time of this writing, it is proposed that Tendermint implement a full
```
jmalicevic, line 113:
```suggestion
At the time of this writing, it is proposed that CometBFT implement a full
```

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,358 @@
...
+2. Implement a separate RPC API on a different port to the standard gRPC
+   interface. This allows for a clearer distinction between the standard and
+   data companion-specific gRPC interfaces, but complicates the server and
+   client interaction models.
+
+Due to the poorer operator experience in option 2, it would be preferable to
+implement option 1, but have certain endpoints be
+[access-controlled](#access-control) by default.
+
+With this in mind, the following gRPC API is proposed, where the Tendermint node
```
jmalicevic, line 132:
```suggestion
With this in mind, the following gRPC API is proposed, where the CometBFT node
```

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,358 @@
...
+    // GetLatestBlockID returns a stream of the latest block IDs as they are
+    // committed by the network.
+    rpc GetLatestBlockID(GetLatestBlockIDRequest) returns (stream GetLatestBlockIDResponse) {}
+
+    // GetBlock attempts to retrieve the block at a particular height.
+    rpc GetBlock(GetBlockRequest) returns (GetBlockResponse) {}
+
+    // GetBlockResults attempts to retrieve the results of block execution for a
+    // particular height.
+    rpc GetBlockResults(GetBlockResultsRequest) returns (GetBlockResultsResponse) {}
```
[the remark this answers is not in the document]
jmalicevic: Could the data companion not be periodically checking the status, or have a user be interesting in progress, and in this scenario pulling the entire block every time might be costly?

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,358 @@
...
+    tendermint.types.Block block = 1;
+}
+
+message GetBlockResultsRequest {
+    // The height of the block results to get.
+    int64 height = 1;
+}
+
+message GetBlockResultsResponse {
+    // All events produced by the ABCI BeginBlock call for the block.
```
jmalicevic, line 191: Given that 0.38+ will not have these functions, will this be too big of an overhead in supporting older versions of CometBFT? What I am saying should just aim at providing the service from 0.38 onwards?

adizere: > What I am saying should just aim at providing the service from 0.38 onwards?

Makes sense to me. It's an additional incentive for users to upgrade faster.

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,358 @@
...
+7. The companion _must_ be able to handle the possibility that a node may
+   disable and then later re-enable the companion interface, potentially causing
+   the companion to have missing data in between those two heights.
+
+8. The API _must_ be opt-in. When off or not in use, it _should_ have no impact
+   on system performance.
+
+9. It _must_ not cause back-pressure into consensus.
+
+10. It _must_ not cause unbounded memory growth.
```
peterbourgon, line 70: It is not possible for something to meet the requirement that it  "_must_ ...  [provide] Committed block data" while also meeting the requirement that it "_must_ not cause back-pressure into consensus". These are mutually exclusive.

thanethomson: I'm interested to know: how do you see these two aspects as mutually exclusive?

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,358 @@
...
+### Requirements
+
+Similar requirements are proposed here as for [ADR-100][adr-100].
+
+1. A node _must_ support at most one data companion.
+
+2. All or part of the following data _must_ be obtainable by the companion, and
+   as close to real-time as possible:
+   1. Committed block data
+   2. `FinalizeBlockResponse` data, but only for committed blocks
```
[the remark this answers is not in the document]
thanethomson: Technically, right now `FinalizeBlock` is called prior to `Commit`. The consensus engine only "knows" that a block is committed and that it's in sync with the application state after `Commit` successfully completes.

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,358 @@
...
+socket-based ABCI application, and the proposed data companion service.
+
+```
+     +----------+      +------------+      +----------------+
+     | ABCI App | <--- | Tendermint | <--- | Data Companion |
+     +----------+      +------------+      +----------------+
+```
+
+In this diagram, it is evident that Tendermint connects out to the ABCI
+application, and the companion connects to the Tendermint node.
```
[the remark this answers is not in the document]
thanethomson: In ADR-101 yes, Comet is the server and the companion is a client.

In ADR-100 it's the other way around.

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,358 @@
...
+   interface. This allows for a clearer distinction between the standard and
+   data companion-specific gRPC interfaces, but complicates the server and
+   client interaction models.
+
+Due to the poorer operator experience in option 2, it would be preferable to
+implement option 1, but have certain endpoints be
+[access-controlled](#access-control) by default.
+
+With this in mind, the following gRPC API is proposed, where the Tendermint node
+will implement these services.
```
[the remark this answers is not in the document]
thanethomson: I'll clarify. I was hoping by this point that we'd be able to easily roll out gRPC support, as per #81, but it seems we're still a way off there.

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,358 @@
...
+[data_companion]
+
+# Is the data companion gRPC API enabled at all? Default: false
+enabled = false
+
+# The maximum number of blocks to allow the data companion's retain height to
+# lag behind the current height. The node will block all other operations until
+# the companion's retain height is less than `max_retain_lag` blocks behind the
+# current height. Default: 100
+max_retain_lag = 100
```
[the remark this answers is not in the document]
thanethomson: Yes, I was actually planning to remove this parameter entirely.

docs/architecture/adr-101-data-companion-pull-api.md
peterbourgon: If CometBFT wants to assume that sending messages to a configured data companion process is reliable, then it has to treat any message delivery error as fatal to the node. But even in that (definitely not viable in any real-world context) case, there's no way to guarantee that it will "not cause back-pressure to consensus", because message delivery success or failure is ultimately only known to the sender after a timeout, and that timeout necessarily impacts consensus.

thanethomson: Are you perhaps thinking of the model proposed by ADR-100 in #73? Because the model in this ADR doesn't propose any requirements regarding guaranteed delivery of specific data to the companion; only that the node must facilitate some mechanism by which the companion can access that specific data should it need to. Requirement 2 is largely about which specific _kinds_ of data the node should make available to the companion.

In this model, it's up to the companion to establish whether or not it's reliably stored the data and manage the "retain height" that it indicates to the node accordingly (which influences the node's pruning mechanisms). Also in this model, data is "buffered" on disk, and not in memory. So inappropriate construction of a companion can result in unbounded _storage_ growth, yes, but that's not something within our control in this model.

Think of this model as an RPC interface for the companion to query whatever data it wants, whenever it wants, including the latest data. The companion also has a say in how the node prunes its data, and the node can only provide the data that it has (over which the companion has a fair degree of control).

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,358 @@
...
+be applied.
+
+### Access Control
+
+As covered in the [gRPC API section](#grpc-api), it would be preferable to
+implement some form of access control for sensitive, data companion-specific
+APIs. At least **basic HTTP authentication** should be implemented for these
+endpoints, where credentials should be obtained from an `.htpasswd` file, using
+the same format as [Apache `.htpasswd` files][htpasswd], whose location is set
+in the Tendermint configuration file.
```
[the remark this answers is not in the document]
thanethomson: While I initially would have wanted to make this the default for the whole gRPC interface in its entirety, the trouble with forcing binding to `localhost` only is that, if folks are running their processes in Docker containers, an external container running the companion service won't be able to access the gRPC service bound to `localhost` within the node's container.

docs/architecture/adr-101-data-companion-pull-api.md
thanethomson: Via the `GetLatestBlockID` RPC call. You'll notice it has a `stream` response - see https://grpc.io/docs/what-is-grpc/core-concepts/#server-streaming-rpc

I'm considering renaming this to just `GetLatestHeight`, as while building the PoC recently I realized that the block ID is not available at the point where I need it in order to return it here; plus it's not strictly necessary anyways. The companion just needs to know what the latest height is such that it can send a `GetBlock` or `GetBlockResults` request.

docs/architecture/adr-101-data-companion-pull-api.md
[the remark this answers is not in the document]
thanethomson: A couple of reasons:
1. The quantity of data stored for each is application-dependent, and only offering a single call could impose dramatic overhead costs in some cases.
2. Right now our RPC distinguishes between these two types of data (see https://docs.cometbft.com/v0.37/rpc/#/Info/block and https://docs.cometbft.com/v0.37/rpc/#/Info/block_results), and I've envisaged this `BlockService` as part of #81 (which should be exposing a sub-set of our current JSON-RPC API).
3. Eventually, once we no longer need to store block results at all, we can just deprecate the `GetBlockResults` API call without affecting clients who rely on `GetBlock`.

docs/architecture/adr-101-data-companion-pull-api.md
thanethomson: My current aim is to produce a PoC on a feature branch based off of `main`, and we still need to decide to which releases it needs to be backported.

It would incentivize users to upgrade faster, but at the same time, folks using the v0.34 release series (which will still be in circulation for quite some time) will not benefit. Plus functionality like the IBC node will then only be possible from v0.38 onwards.

docs/architecture/adr-101-data-companion-pull-api.md
peterbourgon: If a companion process is allowed to "lag" behind the block data of the node it is configured to pull from -- I agree that this is unavoidable, necessary, and ultimately not a problem -- then it doesn't satisfy requirement 1. Or, minimally, it means "as close to real time as possible" has no useful definition: it can be infinite.

In this case, the companion is simply a component to which the node sends state data, and proxies subsequent requests for that state data. But what then is the purpose? Whatever work the companion does, writing data to disk or whatever, could equally be done by the node itself, right? This physical separation of concerns has to be motivated by some requirement that is best satisfied by a separate process.

thanethomson: > If a companion process is allowed to "lag" behind the block data of the node it is configured to pull from

This is precisely the difference between the two competing proposals of ADR-100 and ADR-101. I'll see if I can capture this more clearly in the ADR context.

> Or, minimally, it means "as close to real time as possible" has no useful definition: it can be infinite.

How would you articulate it better? The goal is to offer the data to the companion as quickly as reasonably possible after it's committed, with minimal interference with the node's core operations. I think it's important to keep some pragmatic perspective here on where we _are_ right now (see below), and it's pretty clear that infinite delivery time here would be useless. It's also clear that there's no way to _guarantee_ real-time delivery without the possibility of halting the node (which is the whole premise of ADR-100).

> But what then is the purpose?

Was the [context](https://github.com/cometbft/cometbft/blob/thane/adr-082-data-companion-api/docs/architecture/adr-100-data-companion-push-api.md#context) in ADR-100 not clear? We're ultimately trying to carve out a path here that:

1. Allows us to more clearly separate application- and consensus-related concerns
2. Doesn't immediately break the entire ecosystem

Currently, a node:

1. Handles a mixture of consensus- and application-related concerns, which results in massive storage bloat with application-specific data, and results in us having to maintain multiple indexers, complex query mechanisms, etc. (again, application-level concerns). This ultimately results in code bloat as we're forced to cater for more and more diverse application-specific use cases.
2. Has an incredibly flaky RPC that has a tendency to blow up the node the moment it incurs marginal load (and therefore is not scalable). This includes the _existing_ problem of slow WebSocket subscribers causing back-pressure into consensus. It's also not scalable for operators to have to run multiple full nodes in order to offer a robust RPC service to users, which is what some operators are forced to do, and it would, e.g., be preferable to scale the RPC independently of the node. But the best possible RPC here (especially the querying mechanisms) is application-specific.
3. Contains many layers of heavily intertwined technical debt, which significantly slows our progress in making meaningful changes to the system. We can't get rid of much of this right now due to (1) and (2).

We believe that we can solve many of these and related problems by more clearly separating application- and consensus-related concerns - even separating them into separate processes. We suspect that the first step is explicitly offering some way of offloading non-consensus-related functionality to a separate sidecar process (as captured in the two competing proposals of ADR-100 and ADR-101).

Once the necessary infrastructure is built out to support this functionality independently of the node, we will be in a better position to start to migrate more of the application-level concerns into the SDK/application, and we'll be able to further simplify the consensus engine.

docs/architecture/adr-101-data-companion-pull-api.md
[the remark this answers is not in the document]
sergio-mena: In such a case, I would assume the communicating containers would be part of (or can easily be made part of) a non-publicly-routable subnet (I mean, non-publicly-routable as defined in RFC1918). This is a quite common setup, and operators can then safely replace `localhost` by the IP address of the node's non-publicly-routable interface

docs/architecture/adr-101-data-companion-pull-api.md
sergio-mena: I like the idea of renaming it to `GetLatestHeight`

docs/architecture/adr-101-data-companion-pull-api.md
peterbourgon: I definitely understand why you want to decouple consensus-related concerns from application-related concerns. But I don't understand how that is actually possible, because (AFAIU) the consensus layer is not invariant to the application layer. Or, equivalently: nothing prevents application-layer state from influencing consensus-layer decisionmaking. 

I could be mistaken! Am I mistaken? Are consensus-related concerns _guaranteed_ to be invariant to application-related concerns? If so, then I withdraw my objections 😉 But if not, then consensus is _necessarily_ coupled to the application.

thanethomson: It's important to be concrete here about which specific application concerns bleed over into consensus where it's not necessary. The two areas articulated in these ADRs, specifically, are:

1. Events: https://github.com/cometbft/cometbft/blob/6088604212de8bed47c0c6cc7ebd7275069dcefc/proto/tendermint/abci/types.proto#L346
2. Transaction execution results: https://github.com/cometbft/cometbft/blob/6088604212de8bed47c0c6cc7ebd7275069dcefc/proto/tendermint/abci/types.proto#L351

This data is not necessary to facilitate consensus, and can only be properly interpreted by the application. We dump it to a database (in some cases taking up huge quantities of disk space) and offer it (rather poorly) via the RPC. This seems to have been incorporated into the application <-> consensus engine interaction model as a convenience for application developers, where all Tendermint Core-based nodes would offer a common RPC interface through which to query this data (e.g. `/block_search`, `/tx_search`, `/block_results`, the WebSocket subscription interface, etc.). This is one of those cases where, in trying to cater to everyone, we end up catering to no-one.

There are also more application concerns that have snuck into the consensus engine that really shouldn't be there now that we've got `PrepareProposal` and `ProcessProposal`, such as the notion of gas. But that's a whole other discussion (see [RFC 011](https://github.com/cometbft/cometbft/blob/main/docs/rfc/tendermint-core/rfc-011-delete-gas.md)).

One day in the distant future, the only thing that the consensus engine might be facilitating consensus on is an opaque set of bytes given to it by the application (which could represent transactions, an app hash, etc. - but should the consensus engine care about what those bytes mean?). Even the notion of "transactions" is effectively an application-level concern. But there's still a _lot_ of work to do to get there.

peterbourgon: Whether or not an application concern _should_ bleed into the consensus layer isn't really relevant. The question is if those application concerns _can possibly_ influence the consensus layer. If they can, then hiding application state behind an RPC means that state is not deterministic across nodes, which (AFAICT) breaks consensus.

thanethomson: It's not clear to me that we're actually understanding each other here effectively. Feel free to [sign up](https://groups.google.com/g/cometbft-community/) to join one of our community calls to discuss this synchronously, or reach out to @adizere or @alijnmerchant21 if you'd like to have a more direct conversation!

docs/architecture/adr-101-data-companion-pull-api.md
thanethomson: You don't think that some folks may want to run a full node that doesn't have a companion attached to it _and_ also doesn't store ABCI responses?

docs/architecture/adr-101-data-companion-pull-api.md
thanethomson: I'm envisaging that the data companion and application's retain heights be stored on disk in the state store.

The application's retain height could be set here: https://github.com/cometbft/cometbft/blob/6088604212de8bed47c0c6cc7ebd7275069dcefc/state/execution.go#L295

The companion's retain height would be set during the gRPC call.

docs/architecture/adr-101-data-companion-pull-api.md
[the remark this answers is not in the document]
thanethomson: So could operators use that approach to simply circumvent our forcing them to bind to `localhost`? If so, what security advantage would forcing them to bind to `localhost` provide?

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,358 @@
...
+### Positive
+
+- Facilitates offloading of data to an external service, which can be scaled
+  independently of the node
+  - Potentially reduces load on the node itself
+  - Paves the way for eventually reducing the surface area of a node's exposed
+    APIs
+- Allows the data companion more leeway in reading the data it needs than the
+  approach in [ADR 100][adr-100]
+- Simpler implementation than [ADR 100][adr-100]
```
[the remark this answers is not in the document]
thanethomson: It's hard to give a very accurate estimate here without actually implementing both, but ADR-101 is probably going to be far simpler in its implementation because the majority of the functionality is effectively provided by #81.

With ADR-100, it still needs to explicitly cater for crash recovery (sending the last block upon startup), and the gRPC implementation is probably going to be more complex than ADR-101.

docs/architecture/adr-101-data-companion-pull-api.md
sergio-mena: > For example, ABCI 1.0 defines a ProcessProposal method (which is part of consensus) that can reject proposed blocks. The implementation of that method is arbitrary

> But, otherwise, consensus-related concerns and application-related concerns are necessarily coupled, and (AFAICT) you can't evaluate one without knowledge of the other.

Did you bother to read the [ABCI 1.0 spec](https://github.com/cometbft/cometbft/blob/main/spec/abci/abci%2B%2B_app_requirements.md#formal-requirements)?

If the application is to repsect the ABCI 1.0 specification, it's implementation must definitely _not_  be arbitrary.

docs/architecture/adr-101-data-companion-pull-api.md
yihuang: I hope there's a decoupling between consensus layer and rpc layer, the consensus layer only need to store the raw blocks and some light weight consensus states, the ABCI responses and tx indexers belongs to the rpc layer.

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,370 @@
...
+    // be retained by the node. This data includes block data and block
+    // execution results.
+    //
+    // Setting a retain height lower than a previous setting will result in an
+    // error.
+    //
+    // The lower of this retain height and that set by the application in its
+    // Commit response will be used by the node to determine which heights' data
+    // can be pruned.
+    rpc SetRetainHeight(SetRetainHeightRequest) returns (SetRetainHeightResponse) {}
```
jmalicevic, line 215: Since we are storing both block data and block results, I am wondering whether we should have a separate retain_height for block data and for abci result data? Also there was a suggestion to store the results separately from the block, it might not be a bad idea.

adizere:
Might be good to check on this similar project: https://github.com/terra-money/mantlemint

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,401 @@
...
+### Pruning Behaviour
+
+Two parameters are proposed as a necessary part of the pruning API:
+
+- **Pruning service block retain height**, which influences the height to which
+  the node will retain blocks. This is different to the **application block
+  retain height**, which is set by the application in its response to each ABCI
+  `commit` message.
+
+  The node will prune blocks to whichever is lower between the pruning service
```
jmalicevic, line 81: Responding to a question from @glnro , I realized that we might have the following situation:
- The application has not set a retain height. 
- The data companion sets it to 6 and we start pruning blocks from 1-6
- The application tries to set a retain height of 4. We do not accept this height because blocks have already been pruned ( a higher retain height exists). 


However, at this point, we have no way of letting the application know that this has happened (we log this info, but nothing else). Also, should the data companion (ran potentially by a node operator which is not the same as the app? ), have the power to delete blocks not yet deleted by the application. In other words, should we not block the data companion from pruning until we get an initial `retainHeight` from the application? 

We could assume a scenario where operators have special nodes where they store all the data the application needs on a subset of nodes but prunes data on other nodes. In that case, the application would have all the data it needs. But we are not guaranteeing this.

thanethomson: We should assume that the application's block retain height is set to 0, which means even if the companion set a higher retain height, we wouldn't prune any blocks until the application updates its retain height.

jmalicevic: Ok, I think we need to update the code to do that. Thanks for clarifying!

sergio-mena (approved):
I took a look at the latest changes (from 20 July) and they LGTM. I came here to check things up while reviewing #1150

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,401 @@
...
+# Configuration specifically for the gRPC pruning service, which is considered a
+# privileged service.
+#
+[grpc.privileged.pruning_service]
+
+# Only controls whether the pruning service is accessible via the gRPC API - not
+# whether a previously set pruning service retain height is honoured by the
+# node.
+#
+# To disable the influence of previously set pruning service retain height(s) on
```
jmalicevic, line 343: I [commented](https://github.com/cometbft/cometbft/pull/1154/files/024e917eb0259f56ab1b59a99ad7ec4f09ef5622#r1285770952) on this in @mzabaluev 's PR and it is still not clear to me what behaviour is this sentence describing.

thanethomson: It's no longer relevant. I've removed it now.

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,397 @@
...
+### Metrics
+
+The following metrics are proposed to be added to monitor the health of the
+interaction between a node and its data companion:
+
+- `pruning_service_block_retain_height` - The current block retain height as
+  requested by the pruning service.
+- `pruning_service_block_results_retain_height` - The current block results
+  retain height as requested by the pruning service.
+
```
jmalicevic, line 356: I propose we extend this list with the following:
 - `application_block_retain_height`
 - `block_store_base_height`
 - `abci_results_base_height`
 
 The first one I am suggesting because this allows an operator to see whether there is currently a lower retain height set and why the blocks the data companion wants pruned are not pruned. 
 The other two are there to indicate how far has pruning gotten. I also had a conversation with Milan today where he mentioned that sometimes they are not aware a block is not there and they try to get a block etc.

thanethomson: I'll update the ADR accordingly!

thanethomson: Updated

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,403 @@
...
+## Status
+
+Accepted
+
+## Context
+
+See the [context for ADR-100][adr-100-context].
+
+The primary novelty introduced in this ADR is effectively just a new gRPC API
```
cason, line 18:
```suggestion
The primary novelty introduced in this ADR is effectively a new gRPC API
```

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,403 @@
...
+- 2023-02-28: Renumber from 084 to 101 (@thanethomson)
+- 2022-12-18: First draft (@thanethomson)
+
+## Status
+
+Accepted
+
+## Context
+
+See the [context for ADR-100][adr-100-context].
```
cason, line 16:
```suggestion
See the [context for the Data Companion Push API (ADR-100)][adr-100-context].
```

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,403 @@
...
+CometBFT node (a server).
+
+### Pruning Behaviour
+
+Two parameters are proposed as a necessary part of the pruning API:
+
+- **Pruning service block retain height**, which influences the height to which
+  the node will retain blocks. This is different to the **application block
+  retain height**, which is set by the application in its response to each ABCI
+  `commit` message.
```
cason, line 79:
```suggestion
  `Commit` message.
```

cason: The spec link: https://github.com/cometbft/cometbft/blob/main/spec/abci/abci%2B%2B_methods.md#commit

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,403 @@
...
+
+  The node will prune blocks to whichever is lower between the pruning service
+  and application block retain heights.
+
+- **Pruning service block results retain height**, which influences the height
+  to which the node will retain block results.
+
+These parameters need to be durable (i.e. stored on disk).
```
cason, line 87: durable: persistent?

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,403 @@
...
+syntax = "proto3";
+
+package tendermint.services.block_results.v1;
+
+// BlockResultsService provides information about the execution results for
+// specific blocks.
+service BlockResultsService {
+    // GetByHeight attempts to retrieve the execution results associated with a
+    // block of a certain height.
+    rpc GetByHeight(GetByHeightRequest) returns (GetByHeightResponse)
```
cason, line 189:
```suggestion
    rpc GetByHeight(GetByHeightRequest) returns (GetByHeightResponse) {}
```

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,403 @@
...
+```protobuf
+syntax = "proto3";
+
+package tendermint.services.pruning.v1;
+
+// PruningService provides privileged access to specialized pruning
+// functionality on the CometBFT node to help control node storage.
+service PruningService {
+    // SetBlockRetainHeightRequest indicates to the node that it can safely
+    // prune all block data up to the specified retain height.
```
cason, line 223: I assume this is inclusive, right? So if the parameter is `H`, we can prune all heights <= `H`.

docs/architecture/adr-101-data-companion-pull-api.md
```diff
@@ -0,0 +1,403 @@
...
+}
+```
+
+With this API design, it is technically possible for an integrator to attach
+multiple data companions to the node, but only one of their retain heights will
+be considered by the node.
+
+### Configuration
+
+The following configuration file update is proposed to support the data
```
cason, line 284: Those additions are to CometBFT's configuration file, right?

docs/architecture/adr-101-data-companion-pull-api.md
cason: I don't think we need to repeat this description of ADR-100 thorough the document, but at least here, the first time it is used, it would be good for the reader.

docs/architecture/adr-101-data-companion-pull-api.md
cason: Actually, `[abci-commit]` is a defined link for the document.

cason (approved):
Great work.

It is true that most of the content is shared with ADR-100, which is already merged. I've left just a couple of minor comments for improved readability.

docs/architecture/adr-101-data-companion-pull-api.md
jmalicevic: persisted, as said stored on disk.

docs/architecture/adr-101-data-companion-pull-api.md
jmalicevic: No, the retain heights, as defined in current versions of Comet and TM, are non inclusive. So you will prune **up** to the retain height but that height stays.

docs/architecture/adr-101-data-companion-pull-api.md
jmalicevic: Yes, these are sections added to `config.toml` . I can make it more specific in the text if you prefer.

PR #253, opened 2023-02-01
Removing all the vuepress related build files and references 

### LINKED ISSUE #249: Remove the .vuepress folder and logic to build the docs
Since we are switching the documentation to build using Jekyll and the logic is in the `cometbft-docs` repository, it's better to remove all the logic to build the documentation using vuepress. For that we need to:

- Remove the .vuepress folder
- Update the content of `/docs/DOCS_README.md` to point to `cometbft-docs` repo for instructions on how to build the documentation locally
- Remove the `build-docs` from the `Makefile`
- Remove:
    docs/package.json
    docs/package-lock.json
    docs/post.sh
    docs/pre.sh
    docs/presubmit.sh
    docs/stop-words.txt
    docs/versions

andynog:
Closes #249 

This PR removes all the files related to the old vuepress build and all the references (Makefile, READMEs) since the new `docs` build will be handled by the logic in the `cometbft-docs` repository

#### PR checklist

- [X] Tests written/updated, or no tests needed
- [X] `CHANGELOG_PENDING.md` updated, or no changelog entry needed
- [X] Updated relevant documentation (`docs/`) and code comments, or no
      documentation updates needed


andynog:
If this needs to be back-ported please assign the proper labels

.github/workflows/docs-toc.yml
```diff
@@ -1,20 +0,0 @@
...
-  check:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v3
-      - uses: technote-space/get-diff-action@v6
-        with:
-          PATTERNS: |
-            docs/architecture/**
-            docs/rfc/**
-      - run: make check-docs-toc
```
thanethomson, line 19: This workflow would still be useful actually as it purely ensures that, if someone's submitting an RFC or ADR, that it has a corresponding link in either the RFC or ADR table of contents.

If we remove this here I'd recommend that, in a follow-up PR where we restructure the ADRs/RFCs, we reintroduce it. Otherwise rather just keep this workflow, the `check-docs-toc` Makefile target, and `docs/presubmit.sh` script.

andynog: OK, I've reverted these changes

docs/architecture/README.md
```diff
@@ -89,7 +88,7 @@ therefore they have links to it and refer to CometBFT as "Tendermint" or "Tender
 
 ### Deprecated
 
-None
+- [ADR-035: Documentation](tendermint-core/adr-035-documentation.md)
```
thanethomson, line 91: Moving this ADR into the `tendermint-core` directory is probably causing the TOC check workflow to break. I'd recommend doing the move in a separate follow-up PR, and rather keep this ADR along with all the others.

andynog: it broke I believe because of a file permission which I fixed and the workflow executes, but I can revert this change. Makes sense to leave to an ADR refactor issue

sergio-mena:
Wow! This PR is eye candy! Almost 26K lines to remove!

jmalicevic (approved):

thanethomson:
> Wow! This PR is eye candy! Almost 26K lines to remove!

Haha yes, the vast majority of that's just the `package.lock` file for the documentation build. Still blows my mind how many JavaScript dependencies one needs to accomplish relatively simple tasks nowadays.

Issue #33, opened 2022-12-23
v0.38 release plan


sergio-mena:
Was tendermint/tendermint#9428

The primary focus of the v0.38 release will be rolling out ABCI++ vote extensions and `FinalizeBlock` functionality.
  * [ ]  #12
      * [x]  We need to improve our QA process during Q4 in order to reduce the amount of manual work involved
  * [x] #66
  * [x] #39
  * [x] #595
  * [x] Cut first alpha release
  * [x] Cut release candidate
  * [x] docs: Enable build for `v0.38.x` branch
  * [x] #31 
    * [x] Capture v0.37 baseline metrics (Optional, to do only if existing v0.37 metrics can't be reused)
    * [x] qa: Capture v0.38 pre-release metrics
    * [x] Validate that there's no meaningful, substantial regression from v0.37 to v0.38
  * [x] Execute on the QA plan for integration with the Cosmos SDK
  * [x] Cut final v0.38.0 release
  * [x] Either remove priority mempool or forward port the PR associated to #39 to main and v0.38.x (if exists)
  * [x] https://github.com/cometbft/cometbft/pull/423
  * [x] #571

### Deprioritised from this release:

  * [ ]  Operators should have more control over bandwidth consumption
    * [ ]  Investigate where Tendermint bandwidth is being consumed (perhaps capture this in an RFC as an articulation of the problem space)
    * [x]  Find short-term wins/low-hanging fruit where bandwidth consumption can be reduced (ideally in Q4 of 2022)
* [ ] #54

elmatrwy:
> Was tendermint/tendermint#9428
> 
> The primary focus of the v0.38 release will be rolling out ABCI++ vote extensions and `FinalizeBlock` functionality.
>   * [ ]  #12
>       * [x]  We need to improve our QA process during Q4 in order to reduce the amount of manual work involved
>   * [x] #66
>   * [x] #39
>   * [x] #595
>   * [x] Cut first alpha release
>   * [x] Cut release candidate
>   * [ ] docs: Enable build for `v0.38.x` branch
>   * [x] #31 
>     * [x] Capture v0.37 baseline metrics (Optional, to do only if existing v0.37 metrics can't be reused)
>     * [x] qa: Capture v0.38 pre-release metrics
>     * [x] Validate that there's no meaningful, substantial regression from v0.37 to v0.38
>   * [ ] Execute on the QA plan for integration with the Cosmos SDK
>   * [ ] Cut final v0.38.0 release
[...]

PR #281, opened 2023-02-06
Remove priority mempool

### LINKED ISSUE #260: Remove priority mempool from v0.38
We need to remove the priority mempool, as decided in #39.

**Definition of Done**
- Implementation code complete and all PRs merged
- Any changes needed to the documentation/spec are merged

hvanz:
Closes #260 

Note to reviewers: moving the files from `mempool/v0` to `mempool` creates a dependency cycle `mempool/clist_mempool_test.go -> internal/test/block.go ->  state -> mempool`. That's why I moved the function `makeBlocks` from `internal/test/block.go` to `consensus/replay_test.go` (the only place where it is used). The function `MakeVoteSet` in `internal/test/commit.go` was also creating a cycle, and it was unused, so I removed it.

## Tasks

- [x] Remove the files in `mempool/v1`.
- [x] Remove the `Version` field from `MempoolConfig`.
- [x] Remove v1 cases from node.
- [x] Remove v1 cases from consensus and node tests.
- [x] Remove mempool version from manifest in node generator for e2e tests.
- [x] Remove last three fields (`sender`, `priority`, `mempool_error`) of `ResponseCheckTx`, which are related to the priority mempool.
- [x] Move `mempool/v0/*` to `mempool/`.
- [x] Update docs
- [x] Update changelog


config/config.go
```diff
@@ -27,11 +27,6 @@ const (
 	// DefaultLogLevel defines a default log level as INFO.
 	DefaultLogLevel = "info"
 
-	// Mempool versions. V1 is prioritized mempool, v0 is regular mempool.
-	// Default is v0.
-	MempoolV0 = "v0"
```
lasarojc, line 32: I feel like we should leave some of this information in, to keep the memory of previous versions easily seen in the code so that if someone comes up with a new mempool, it does not end up as V1 again. I am not sure what would be the best approach, though. Maybe just comment out v1 and add a deprecation note in the comment as well, and leave the Version setting in, with the default value only, so people looking into the mempool will necessarily bump into the history.

hvanz: Thanks for the comment! I thought the same and I discussed it with @sergio-mena. The conclusion was that we should keep just one default implementation (the current one) and roll back the code as it was before the priority mempool was added. The reasons:
- in the future different mempools will be implemented at the application level through ABCI 1.0,
- we don't know exactly how we will structure the mempool module once we modularize Comet,
- we don't want to maintain an unused configuration entry.

Also, I think we may want to have only one default mempool implementation in Comet, and leave to the users the option to implement their own versions at the application level.

Now, I think you're right that we should document somehow that versions v0 and v1 existed in past releases (though I don't think it will be a real problem if version numbers are reused).

spec/mempool/general.md
```diff
@@ -0,0 +1,32 @@
+
```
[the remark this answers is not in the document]
hvanz: Oops, this was added by mistake.

config/config.go
cason: I do agree, and would add that `v[0-9]` is not really a good name for different implementations.

lasarojc: Cool. A good comment near MempoolConfig should suffice.

lasarojc (approved):
The removal of v1 and renaming of v0 looks good to me

proto/tendermint/abci/types.proto
```diff
@@ -230,12 +230,6 @@ message ResponseCheckTx {
   repeated Event events     = 7
       [(gogoproto.nullable) = false, (gogoproto.jsontag) = "events,omitempty"];
   string codespace = 8;
-  string sender    = 9;
-  int64  priority  = 10;
-
-  // mempool_error is set by CometBFT.
-  // ABCI applictions creating a ResponseCheckTX should not set mempool_error.
-  string mempool_error = 11;
```
sergio-mena, line 238: I know `v0.38.x` is ABCI-breaking w.r.t. `v0.37.x`, so this change should not be problematic.
However, we should add somewhere in the UPGRADING notes that we are removing fields from `ResponseCheckTx`.
Could you please add a line in #33 to not forget to do it? Alternatively, open an issue and hook it up to #33. Thanks

config/config.go
```diff
@@ -754,39 +745,20 @@ type MempoolConfig struct {
...
-	// insertion time into the mempool is beyond TTLDuration.
-	TTLDuration time.Duration `mapstructure:"ttl-duration"`
-
-	// TTLNumBlocks, if non-zero, defines the maximum number of blocks a transaction
-	// can exist for in the mempool.
-	//
-	// Note, if TTLDuration is also defined, a transaction will be removed if it
-	// has existed in the mempool at least TTLNumBlocks number of blocks or if
-	// it's insertion time into the mempool is beyond TTLDuration.
-	TTLNumBlocks int64 `mapstructure:"ttl-num-blocks"`
```
sergio-mena, line 772: Please add a mention to these suppressed config parameters in the issue on UPGRADING.

Do we need to deprecate these in `v0.37.x`?

internal/test/block.go
```diff
@@ -91,33 +90,3 @@ func MakeHeader(t *testing.T, h *types.Header) *types.Header {
 
 	return h
 }
-
-func MakeBlock(state sm.State) *types.Block {
```
sergio-mena, line 95: MakeBlocks, MakeVote, MakeVoteSet tend to have multiple occurrences across our code, I've done some quick tidy-up in the last months, but any improvement (such as the one done here) is super-welcome :-)

sergio-mena (approved):
Thanks @hvanz for this!

config/config.go
hvanz: PR #365 will mark these fields as deprecated.

proto/tendermint/abci/types.proto
lasarojc: The protobuf documentation indicates reserving field numbers as a possible solution here

https://protobuf.dev/programming-guides/proto/#reserved

sergio-mena: Yeah, we've used that in ABCI. I think it is a good idea.
So, can we... ?
* reserve those numbers here, maybe with a comment explaining why
* add a note in #33 to also mention these fields are disappearing in `v0.38.x`

sergio-mena: Please ignore bullet 2 above. I see you already did it in your latest commit

hvanz: Thanks for pointing this out @sergio-mena! I already updated the UPGRADING notes.

hvanz: That's a good point, @lasarojc! Afaiu, this could have been a security issue in proto2, but not in proto3, where every field is implicitly optional. So the new generated code will ignore data with the old fields that may come from an application. We may still want to mark these fields as reserved so that nobody tries to use them in an incompatible way. What do you think?

config/config.go
hvanz: Done. I updated UPGRADING.

proto/tendermint/abci/types.proto
lasarojc: Thanks for looking into it. I think it is a good idea to ensure that the field ids don't get reused and avoid any clashes.

config/config.go
hvanz: I added a comment about the mempool versions to `MempoolConfig` in `config/config.go`.

proto/tendermint/abci/types.proto
hvanz: So at the end I reverted back the changes and marked these fields as reserved in the proto file. Thanks @lasarojc and @sergio-mena!

PR #286, opened 2023-02-08
Remove unused code

hvanz:
This PR cleans:
- unused fields in the node struct
- unused wrapper functions on the node's fields 
- unused return values on a couple of functions


sergio-mena (approved):
Thanks @hvanz, I agree with @lasarojc that `Node` is a major "entanglement hub", so touching that code is usually a delicate matter.
As I understand that this PR is exclusively removing code, and that you have been guided by the compiler to do so, I'd say it is safe to go ahead.

node/node.go
```diff
@@ -699,41 +694,11 @@ func (n *Node) Switch() *p2p.Switch {
...
-}
-
 // Mempool returns the Node's mempool.
 func (n *Node) Mempool() mempl.Mempool {
 	return n.mempool
 }
 
-// PEXReactor returns the Node's PEXReactor. It returns nil if PEX is disabled.
-func (n *Node) PEXReactor() *pex.Reactor {
-	return n.pexReactor
```
tac0turtle, line 729: this should be a required change log, this breaks some tests in sdk and the changelog doesnt have anything

Issue #283, opened 2023-02-07
QA of CMT 0.34.x


lasarojc:
This task has 3 goals: 
- [x] Confirm performance compatibility between CMT 0.34.27 and Informal's TM 0.34.26
    - [x] Run 200-node QA test, collect and compare metrics with previous run.
            - #291
            - #312
- [x] Confirm interoperability between CMT 0.34.27 and Informal's TM 0.34.26
    - [x] Run a 200-node QA test with 1/3 of nodes running CometBFT and the remaining running TM @jmalicevic 
            #292
    - [x] Run a 200-node QA test with 2/3 of nodes running CometBFT and the remaining running TM @lasarojc    
            #297 
    - [ ] Run e2e with an extended manifest, allowing to test blocksync in a mixed network (saving us from running the rotating node test) #375
    - [ ] Add the extended manifest in a GitHub workflow to be run weekly #425 
- [x] Finalize report

PR #292, opened 2023-02-09
QA: v0.34 - 2/3 TM 1/3 CM

jmalicevic:
Contributes to #283.


docs/qa/v034/README.md
```diff
@@ -560,7 +575,7 @@ Version: xxxxxxxxxxxxxxxxx
 
 #### 2/3 Tendermint Core - 1/3 CometBFT
 
-**Result: ????**
+**Result: PASS**
 
 Date: 2023-02-08
```
sergio-mena, line 581: Please change the "version" to "versions", and add
CometBFT: 3b783434f26b0e87994e6a77c5411927aad9ce3f (if that's the one you used)
Tendermint Core: [add here the hash of the Tendermint Core version]

sergio-mena (approved):
Thanks! Looking excellent!

PR #297, opened 2023-02-09
QA: v0.34 - 1/3 TM 2/3 CMT

lasarojc:
Contributes to #283


sergio-mena (approved):
Thanks Lásaro! In particular for the new, better way of producing the Prometheus plots.
Some of the plots seem to start some time before the experiment. Although I agree it can be interesting to see how the system goes from idle to busy, it adds some noise when comparing them with existing reports (baseline, `v0.37.x`). The main criterion for PASS or FAIL is how similar new plots are to existing ones.

PR #313, opened 2023-02-10
e2e: add manifest fields to enable Prometheus on nodes

### LINKED ISSUE #294: e2e: add manifest field to enable Prometheus on nodes
In order to collect metrics from e2e test runs, we need to enable Prometheus in node configs.

hvanz:
Closes #294 

---

#### PR checklist

      documentation updates needed


test/e2e/pkg/manifest.go
```diff
@@ -82,6 +82,11 @@ type Manifest struct {
 	LoadTxSizeBytes   int `toml:"load_tx_size_bytes"`
 	LoadTxBatchSize   int `toml:"load_tx_batch_size"`
 	LoadTxConnections int `toml:"load_tx_connections"`
+
+	// When true, enables Prometheus metrics for all nodes.
+	// When false, just apply the Prometheus flag of each individual node, if present.
+	// Defaults to false.
+	Prometheus bool `toml:"prometheus"`
```
sergio-mena, line 89: Have you tested the case when this is false, but the one added below is true for several nodes?
Won't there be a socket binding error?

test/e2e/runner/setup.go
```diff
@@ -73,6 +73,10 @@ func Setup(testnet *e2e.Testnet, infp infra.Provider) error {
 			}
 		}
 
+		if testnet.Prometheus {
+			node.Prometheus = true
```
sergio-mena, line 77: Will this override the setting you introduced in `test/e2e/pkg/manifest.go` line 173 ?

test/e2e/pkg/manifest.go
hvanz: As discussed offline, I removed the config field for a node and kept just one top-level flag that applies to all the nodes. This will make things simpler for the user, as this seems to be the most typical scenario where the user just wants to enable the metrics for all nodes or none.

test/e2e/runner/setup.go
hvanz: This code is no longer relevant, as I removed `node.Prometheus` flag (as explained above).

sergio-mena (approved):
Thanks!

hvanz:
@Mergifyio backport v0.34.x

PR #316, opened 2023-02-10
Release v0.34.27-alpha.1

thanethomson:
We're cutting an alpha of v0.34.27 in order to help facilitate integration with this release series.

Once we have confirmed that our QA is done, we will cut v0.34.27.

Also includes some of the changes from #317 in order to slightly improve the pre-release/release processes.

---

#### PR checklist

- [x] Tests written/updated, or no tests needed
- [x] `CHANGELOG_PENDING.md` updated, or no changelog entry needed
- [x] Updated relevant documentation (`docs/`) and code comments, or no
      documentation updates needed

PR #317, opened 2023-02-10
ci: Update release and pre-release workflows and process

thanethomson:
This updates the release and pre-release workflows to simplify, fix and clarify them. It also updates the `RELEASES.md` file to refer to unclog when building the changelog.

Needs to be backported to v0.37.x, since I'm updating this for the v0.34.x branch in #316.

---

#### PR checklist

      documentation updates needed

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -20,7 +20,8 @@ jobs:
         with:
           go-version: '1.18'
 
-      - run: |
+      - name: Generate release notes
+        run: |
           VERSION="${GITHUB_REF#refs/tags/}"
           CHANGELOG_URL="https://github.com/cometbft/cometbft/blob/${VERSION}/CHANGELOG.md"
           echo "See the [CHANGELOG](${CHANGELOG_URL}) for changes available in this pre-release, but not yet officially released." > ../release_notes.md
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -27,8 +27,10 @@ jobs:
           version: latest
           args: build --skip-validate   # skip validate skips initial sanity checks in order to be able to fully run
 
-      # Link to CHANGELOG_PENDING.md as release notes.
-      - run: echo https://github.com/cometbft/cometbft/blob/${GITHUB_REF#refs/tags/}/CHANGELOG_PENDING.md > ../release_notes.md
+      # Link to CHANGELOG.md as release notes (ensure that changelog is built
+      # with unreleased entries prior to cutting a pre-release)
+      - run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/pre-release.yml
```diff
@@ -20,22 +20,16 @@ jobs:
         with:
           go-version: '1.18'
 
-      - name: Build
-        uses: goreleaser/goreleaser-action@v4
-        if: ${{ github.event_name == 'pull_request' }}
```
thanethomson, line 25: We never run this on pull requests, only when we push tags, so this job is useless.

.github/workflows/pre-release.yml
```diff
@@ -20,22 +20,16 @@ jobs:
...
-        with:
-          version: latest
-          args: build --skip-validate   # skip validate skips initial sanity checks in order to be able to fully run
-
-      # Link to CHANGELOG_PENDING.md as release notes.
-      - run: echo https://github.com/cometbft/cometbft/blob/${GITHUB_REF#refs/tags/}/CHANGELOG_PENDING.md > ../release_notes.md
+      - run: |
+          VERSION="${GITHUB_REF#refs/tags/}"
+          CHANGELOG_URL="https://github.com/cometbft/cometbft/blob/${VERSION}/CHANGELOG.md"
+          echo "See the [CHANGELOG](${CHANGELOG_URL}) for changes available in this pre-release, but not yet officially released." > ../release_notes.md
```
thanethomson, line 26: This attempts to clarify what we're doing here by building the `release_notes.md` file.

.github/workflows/pre-release.yml
```diff
@@ -20,22 +20,16 @@ jobs:
...
 
       - name: Release
         uses: goreleaser/goreleaser-action@v4
-        if: startsWith(github.ref, 'refs/tags/')
```
thanethomson, line 35: We only ever run this job when tags are pushed, so there's no need to check that we're pushing tags.

.github/workflows/pre-release.yml
```diff
@@ -20,22 +20,16 @@ jobs:
...
         with:
           version: latest
-          args: release --rm-dist --release-notes=../release_notes.md
+          args: release --clean --release-notes ../release_notes.md
```
thanethomson, line 32: The `--rm-dist` flag has been deprecated in favor of the `--clean` flag according to https://goreleaser.com/deprecations/#-rm-dist

.goreleaser.yml
```diff
@@ -26,7 +26,7 @@ checksum:
 
 release:
   prerelease: auto
-  name_template: "{{.Version}}"
+  name_template: "v{{.Version}}"
```
thanethomson, line 29: Without the `v` prefix, the releases look like this: https://github.com/tendermint/tendermint/releases/tag/v0.34.24

This doesn't match the tags that we push, which include a `v` prefix.

PR #345, opened 2023-02-14
Allow inclusion of repository in the tag name, so ghcr and dockerub c…

lasarojc:
…an be used, including in the same experiment

Should be ported to 0.37 and main.

---

#### PR checklist

- [ ] Tests written/updated
- [ ] Changelog entry added in `.changelog` (we use
  [unclog](https://github.com/informalsystems/unclog) to manage our changelog)


sergio-mena (approved):
Thanks for this!

PR #356, opened 2023-02-17
Updates golang to 1.20

lasarojc:
---

#### PR checklist

- [x] Tests written/updated


adizere (approved):

andynog:
Doesn't this needs to be also updated in other places like Dockerfiles so tests/images are using the right version ?
https://github.com/cometbft/cometbft/blob/main/DOCKER/Dockerfile

lasarojc:
Closing this for now. We'll be pushing for 1.19, which also includes the fix.

Issue #259, opened 2023-02-02
Mark priority mempool as deprecated in v0.37.x


hvanz:
We need to mark the priority mempool as deprecated, as decided in #39.

Since these changes only affect the 0.37 release, we should make the PR on v0.37.x, instead of main.

## Tasks

- [x] Add comments with special Deprecated label to relevant public functions.
- [x]  Log a warning message on node startup when "v1" is the mempool version set in the config. For that, update CheckDeprecate() in config/config.go.
- [x] Update docs

**Definition of Done**
- Implementation code complete and all PRs merged
- Any changes needed to the documentation/spec are merged

PR #365, opened 2023-02-21
Mark priority mempool config fields as deprecated

hvanz:
The priority mempool was marked as deprecated for v0.37.x in #259. The mempool config fields `TTLDuration` and `TTLNumBlocks` are only used by the priority mempool and will be deleted in v0.38.x, but we forgot to marked them as deprecated. This PR fixes it.


sergio-mena (changes requested):
The changes LGTM, however, could we also add a warning message, similar to what was done [here](https://github.com/cometbft/cometbft/blob/v0.37.x/config/config.go#L159-L161)?

hvanz:
> The changes LGTM, however, could we also add a warning message, similar to what was done [here](https://github.com/cometbft/cometbft/blob/v0.37.x/config/config.go#L159-L161)?

Thanks @sergio-mena, I added the warning messages.

sergio-mena (approved):
Excellent!
Ship it!

Issue #319, opened 2023-02-10
docs: fix broken links


andynog:
This issue will be the master issue to track the fixes for the several broken links in the documentation site across releases. 

A new tool has been implemented in the `cometbft-docs` Makefile that allows [checking for these broken links](https://github.com/cometbft/cometbft-docs/pull/31) locally and reports are generated with these broken links that makes it easier to fix them in the `/docs` folder.

Fix broken links for:
- [ ] v0.34
- [ ] v0.37-rc2
- [ ] main

PR #368, opened 2023-02-21
Docs fixes

andynog:
closes: #319 

This fixes many broken links, typos, references, landing pages, navigation links. 

This PR will be used to make the docs.cometbft.com to be ready for production deployment

#### PR checklist

- [ ] Tests written/updated


docs/tutorials/go-built-in.md
```diff
@@ -700,7 +700,7 @@ The additional logic at the end of the file allows the program to catch SIGTERM.
 
 Our application is almost ready to run, but first we'll need to populate the CometBFT configuration files.
 The following command will create a `cometbft-home` directory in your project and add a basic set of configuration files in `cometbft-home/config/`.
-For more information on what these files contain see [the configuration documentation](https://github.com/cometbft/cometbft/blob/v0.37.x/docs/core/configuration.md).
+For more information on what these files contain see [the configuration documentation](../core/configuration.md).
```
cason, line 703: The version used in the tutorial is the 0.37.0. That is why a full link was used here.

docs/tutorials/go.md
```diff
@@ -598,7 +598,7 @@ which connects to our server and send us transactions and other messages.
 
 Our application is almost ready to run, but first we'll need to populate the CometBFT configuration files.
 The following command will create a `cometbft-home` directory in your project and add a basic set of configuration files in `cometbft-home/config/`.
-For more information on what these files contain see [the configuration documentation](https://github.com/cometbft/cometbft/blob/v0.37.x/docs/core/configuration.md).
```
cason, line 601: The version used in the tutorial is the 0.37.0. That is why a full link was used here.

cason (commented):
Thanks for this effort.

I've left a couple of comments on the tutorials. We have to come up with a way they are generic, and do not apply to a single version. This is a little more complex since ABCI is under changes from v0.37.

docs/app-dev/app-architecture.md
```diff
@@ -52,4 +52,4 @@ See the following for more extensive documentation:
 - [Interchain Standard for the Light-Client REST API](https://github.com/cosmos/cosmos-sdk/pull/1617) (legacy/deprecated)
 - [CometBFT RPC Docs](https://docs.cometbft.com/main/rpc/)
 - [CometBFT in Production](../core/running-in-production.md)
-- [ABCI spec](https://github.com/cometbft/cometbft/tree/main/spec/abci)
+- [ABCI spec](../spec/abci/README.md)
```
cason, line 55: Will this link work from Github? I don't think so.

docs/tutorials/go-built-in.md
andynog: thanks for the context @cason, I'll revert this change. We can figure out in the future a better way to link that when we do the docs revamp for example.

docs/tutorials/go.md
andynog: I've reverted this change too

docs/app-dev/app-architecture.md
andynog: it will not work, good catch. We can also link to the documentation site `https://docs.cometbft.com/main/spec/abci`, what do you think ? I can't think now a way to make it work both ways.

thanethomson (approved):

PR #369, opened 2023-02-22
Don't specify the patch version in go-vulncheck

lasarojc:
go-setup now supports 1.19.6 when specifying just 1.19, so we can drop the patch version.

---

#### PR checklist

- [x] Tests written/updated


sergio-mena (approved):
Ship it!

thanethomson (approved):

PR #375, opened 2023-02-23
Improve wait for node in e2e runner for long catch-up times

sergio-mena:
Contributes to #283

The e2e runner has a fix timeout when waiting for a process to reach a particular height.
In longer e2e tests (with complex manifests) this hard-code timeout value may not be well adapted (too short).
This PR renders the logic more flexible in that it only applies the timeout from the time it sees the node is not advancing heights.

This PR also adds the long-running manifest that motivated the improvement (thanks @lasarojc)

---

#### PR checklist

- [x] Tests written/updated


test/e2e/networks/long.toml
```diff
@@ -0,0 +1,254 @@
...
+    validator08 = 50
+  [validator_update.2005]
+    validator09 = 50
+  [validator_update.3005]
+    validator10 = 50
+
+[node]
+  [node.full01]
+    mode = "full"
+    version = "ghcr.io/informalsystems/tendermint-e2e-node:v0.34.x"
```
lasarojc, line 38: How can we make this general for all branches?
Maybe the update should be from `cometbft/e2e-node:latest` to `cometbft/e2e-node:local`?

thanethomson: Would this not actually break if merging into `main`? We'd be running a v0.34-based node alongside v0.38-based ones, right?

sergio-mena: Hi, sorry for delay coming back to this. I looked at @lasarojc's commits below (thanks for them!) and I think they address this concern.
I have also removed `backport-to-v0.34x` as that will need to be a special case.
Are we still missing something here?

test/e2e/networks/long.toml
```diff
@@ -0,0 +1,254 @@
+ipv6 = true
+initial_height = 2
+key_type = ""
+abci_protocol = "tcp"
+upgrade_version = "cometbft/e2e-node:local"
```
jmalicevic, line 5: Trying to run this locally (not the same manifest but with these parameters), it is complaining about the lack of `e2e-node:local`. I do have a `e2e-node:local-version` tag in my docker image list and that is also what you have in the Makefile so I don't know whether this should be changed here.

lasarojc (approved):

test/e2e/networks/long.toml
lasarojc: Yes, it must be local-version, but it should be fixed now. Pls pull.

test/e2e/runner/wait.go
```diff
@@ -30,5 +30,5 @@ func WaitUntil(testnet *e2e.Testnet, height int64) error {
 // waitingTime estimates how long it should take for a node to reach the height.
 // More nodes in a network implies we may expect a slower network and may have to wait longer.
 func waitingTime(nodes int) time.Duration {
-	return time.Duration(20+(nodes*2)) * time.Second
+	return time.Duration(20+(nodes*nodes)) * time.Second
```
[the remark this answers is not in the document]
sergio-mena: Can you suggest something along those lines? (ideally liner w.r.t. number of nodes and liner w.r.t height)

PR #378, opened 2023-02-23
ci: Remove Buf tokens from workflow

thanethomson:
Attempts to fix the current failing proto lint test in CI.

---

#### PR checklist

- [ ] Tests written/updated


andynog (approved):
looks good to me, if we get rate limiting issues we can try to put the github token back, but I agree with removing it.

PR #382, opened 2023-02-24
txindex v0.37.x: Enable the indexer to parse slashes in query

jmalicevic:
The transaction indexer indexes transaction events by creating a custom key containing the following: 
- event attribute key
- event attribute value
- txresult
- height
- event sequence

These elements are separated with a '/'. This however causes problems when the value (over which we have no control), contains a '/' . This PR fixes this by making sure to isolate the element containing the value.

The code was tested for backwards compatibility, making sure we can parse events indexed with older versions of Tendermint (without this fix). 

---

#### PR checklist

- [x] Tests written/updated


state/txindex/kv/kv.go
```diff
@@ -588,45 +589,59 @@ func isTagKey(key []byte) bool {
 	// The check for 3 was added to allow data indexed before (w/o the event number)
 	// to be retrieved.
 	numTags := strings.Count(string(key), tagKeySeparator)
-	return numTags == 4 || numTags == 3
+	return numTags >= 3
```
lasarojc, line 592: Pls update the comment `This should be always 4` to `This should be at least 4`.

jmalicevic: Done! :)

state/txindex/kv/kv.go
```diff
@@ -584,49 +585,63 @@ REMOVE_LOOP:
 // Keys
 
 func isTagKey(key []byte) bool {
-	// This should be always 4 if data is indexed together with event sequences
-	// The check for 3 was added to allow data indexed before (w/o the event number)
-	// to be retrieved.
+	// This should be always at least 4 if data is indexed together with event sequences
+	// But we need to support versions with 3 tags as well in case the data was indexed with an old
+	// version of CometBFT.
```
lasarojc, line 590: Maybe make the comment neutral, since we don't have old versions of CometBFT.
```suggestion
	// But we need to support versions with 3 tags as well in case the data was indexed with an old
        // format.
```

lasarojc (approved):

.changelog/unreleased/bug-fixes/382-txindexer-fix-slash-parsing.md
```diff
@@ -0,0 +1 @@
+- `[state/kvindexer]` \#382 Resolved crashes when event values containted slashes, introduced after adding event sequences in \#77.  (@jmalicevic)
```
thanethomson, line 1:
```suggestion
- `[state/kvindexer]` \#382 Resolved crashes when event values contained slashes, introduced after adding event sequences in \#77.  (@jmalicevic)
```

thanethomson (approved):

PR #383, opened 2023-02-24
txindexer v0.34.x: Enable the indexer to parse slashes in query 

jmalicevic:
The transaction indexer indexes transaction events by creating a custom key containing the following: 
- event attribute key
- event attribute value
- txresult
- height
- event sequence

These elements are separated with a '/'. This however causes problems when the value (over which we have no control), contains a '/' . This PR fixes this by making sure to isolate the element containing the value.

The code was tested for backwards compatibility, making sure we can parse events indexed with older versions of Tendermint (without this fix). 

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/state/txindex/kv/kv_test.go
+++ b/state/txindex/kv/kv_test.go
@@ -71,6 +71,7 @@ func TestTxSearch(t *testing.T) {
 	txResult := txResultWithEvents([]abci.Event{
 		{Type: "account", Attributes: []abci.EventAttribute{{Key: []byte("number"), Value: []byte("1"), Index: true}}},
 		{Type: "account", Attributes: []abci.EventAttribute{{Key: []byte("owner"), Value: []byte("Ivan"), Index: true}}},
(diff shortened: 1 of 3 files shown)


lasarojc (approved):

state/txindex/kv/kv_test.go
```diff
@@ -101,6 +102,16 @@ func TestTxSearch(t *testing.T) {
 		{"account.number < 10000 AND account.owner = 'Ivan'", 1},
 		// search using a prefix of the stored value
 		{"account.owner = 'Iv'", 0},
+		// search for owner with dash in name
```
thanethomson, line 105: Should this not be:

```suggestion
		// search for owner with slash in name
```

And similar for all the code comments that follow...

thanethomson (approved):

state/txindex/kv/kv_test.go
jmalicevic: You are right :D good catch!

Issue #409, opened 2023-02-28
Deprecate blocksync-related configuration in `v0.37.x` that disables blocksync. Remove it from `v0.38.x`


sergio-mena:
_Tracked by: #41 and #10_

After some discussion on two alternatives on how to tackle an error path with vote extensions in `v0.38.x`, the CometBFT decided to remove the possibility for blocksync to be configurable (it _must_ always be active).

Therefore:
* [x] In `v0.37.x` we need to deprecate the CometBFT configuration options that allows the operator to disable blocksync #414
* [x] In `v0.38.x` these options will be removed, as part of vote extension implementation #415

## DoD

* There is an (early) release of `v0.37.x` that warns the operator with a deprecation message if they configure their node to disable blocksync
* It is not possible for the operator to disable blocksync in `v0.38.x`

PR #415, opened 2023-02-28
config: remove `block_sync` key

hvanz:
Part of #409 

---

#### PR checklist

- [x] Tests written/updated


docs/core/block-sync.md
```diff
@@ -18,17 +18,13 @@ consensus gossip protocol.
...
-`--block_sync=false`.
-
-In this mode, the CometBFT daemon will sync hundreds of times faster
-than if it used the real-time consensus process. Once caught up, the
-daemon will switch out of Block Sync and into the normal consensus mode.
-After running for some time, the node is considered `caught up` if it
-has at least one peer and its height is at least as high as the max
-reported peer height.
-See [the IsCaughtUp method](https://github.com/cometbft/cometbft/blob/main/blocksync/pool.go#L168).
+Nodes start in Block Sync mode by default. In this mode, the CometBFT daemon
```
sergio-mena, line 21:
```suggestion
When starting from scratch, nodes will use the Block Sync mode.
In this mode, the CometBFT daemon
```

test/e2e/networks/ci.toml
```diff
@@ -61,7 +61,6 @@ database = "rocksdb"
 perturb = ["pause"]
 
 [node.validator05]
-block_sync = "v0"
```
sergio-mena, line 64: should that be renamed to `block_sync_version`, rather than deleted? (I saw you've kept the version as part of the generation)

test/e2e/runner/setup.go
```diff
@@ -209,10 +209,8 @@ func MakeConfig(node *e2e.Node) (*config.Config, error) {
 		return nil, fmt.Errorf("unexpected mode %q", node.Mode)
 	}
 
-	if node.BlockSync == "" {
-		cfg.BlockSyncMode = false
-	} else {
-		cfg.BlockSync.Version = node.BlockSync
+	if node.BlockSyncVersion != "" {
```
sergio-mena, line 212: But I can't be empty now, can it?

test/e2e/networks/ci.toml
hvanz: That's right. Before, this line was needed to enable BlockSync, not just to asign the version.

test/e2e/runner/setup.go
hvanz: You're right. I simplified the code.

sergio-mena (approved):
Thanks!

PR #426, opened 2023-03-01
ci: Add release version check

thanethomson:
Since I forgot to update the version in `version/version.go` yesterday, I thought I'd cut another RC, this time introducing a workflow to check that the version has indeed been updated.

This workflow strips the `release/v` prefix from a release branch name to obtain the version number, and compares it to the version output by `go run ./cmd/cometbft/ version` (which should just output the version from `version/version.go`).

---

#### PR checklist

- [ ] Tests written/updated


thanethomson:
Seems I'm going to have to introduce this on `main` first, backport it and then attempt to test it out.

PR #469, opened 2023-03-06
spec:abci2.0 - clarify crash recovery mechanism

### LINKED ISSUE #203: spec: Clearly document when an app is expected to persist state
Per a recent conversation with @hdevalence regarding the ABCI `InitChain` request, it isn't 100% clear to app developers as to what to do with multiple such `InitChain` requests in the case that the interaction between the node and the app fails before the first `Commit` call for the first block.

We should clarify in the spec that our crash-recovery model requires both the app and CometBFT to crash together, and that the app is expected to persist state only upon calling `Commit`.

This should be clear in the specs on all release branches.

jmalicevic:
Closes #203 .


spec/abci/abci++_app_requirements.md
```diff
@@ -810,7 +810,16 @@ implementation of
 
 On startup, CometBFT calls the `Info` method on the Info Connection to get the latest
 committed state of the app. The app MUST return information consistent with the
-last block it successfully completed Commit for.
+last block it successfully completed `Commit` for. 
+
+The application is expected to persist it's state during the `Commit` method.
```
lasarojc, line 815:
```suggestion
The application is expected to persist its state during the `Commit` method.
```

spec/abci/abci++_app_requirements.md
```diff
@@ -810,7 +810,16 @@ implementation of
...
+CometBFT and the application are expected to crash together and there should not 
+exist a scenario where the application has persisted state that CometBFT has not.
+
+In other words, if `Commit` for a block H has not been executed and CometBFT crashed,
+the application
+should not have persisted and relied on any state between the `Commit` for H - 1 and 
+H.
```
lasarojc, line 822: Not really part of this PR, but in line 824, below, it states that either Commit completed and last_block_height is now H or Commit failed and last_block_height is still H-1, but even if Commit crashed midway, it may have crashed after the advancing to H. So the restriction should be last_block_height and last_block_app_hash agree, not that they are agree on H-1.

Suggestion
```
If the app successfully committed block `H`, then `last_block_height = H` and `last_block_app_hash = <hash returned by Commit for block H>`. 
If the app failed during the Commit of block `H`, then either 
- `last_block_height = H-1` and `last_block_app_hash = <hash returned by Commit for block H-1, which is the hash in the header of block H>`; or, 
- `last_block_height = H` and `last_block_app_hash = <hash returned by Commit for block H>`.

**Review by lasarojc — APPROVED** [review]
Nitpicking.

File: spec/abci/abci++_app_requirements.md [hunk 3]
```diff
@@ -810,7 +810,16 @@ implementation of
...
-last block it successfully completed Commit for.
+last block it successfully completed `Commit` for. 
+
+The application is expected to persist it's state during the `Commit` method.
+CometBFT and the application are expected to crash together and there should not 
+exist a scenario where the application has persisted state that CometBFT has not.
+
+In other words, if `Commit` for a block H has not been executed and CometBFT crashed,
+the application
+should not have persisted and relied on any state between the `Commit` for H - 1 and
```
cason, line 821: Didn't really understand the "and relied on" part.

spec/abci/abci++_app_requirements.md
cason: I would, first, refer that `last_block_height` and `last_block_app_hash` are the fields returned by the `Info` method (https://github.com/cometbft/cometbft/blob/7c7b48039453aa66fd6b128af6701651e1398319/spec/abci/abci%2B%2B_methods.md). This because it was not clear to me, and these variables are only used here on the text.

cason: Second, is it not possible to have one of the fields updated (to H) but the other not?

cason: Third, I agree with Lasaro here that the app could crash during the `Commit` of height H and it will have the two fields updated to H state, although Comet will not be aware of this because the crash occurs before returning the control to Comet.

spec/abci/abci++_app_requirements.md
```diff
@@ -837,6 +846,12 @@ First, some simple start conditions:
```
cason, line 846: Also commenting something that is not part of this PR, ehe.

> Note also CometBFT never calls Commit on an ABCI app twice for the same height.

Comet can call `Commit` for height H, then we have a crash, a recover, and since the application has returned to have state at height H, Comet will call `Commit` for height H "again". So, in fact, if `Commit` has returned for height H, then Comet will never call it again for height H.

spec/abci/abci++_app_requirements.md
```diff
@@ -837,6 +846,12 @@ First, some simple start conditions:
 
 If `appBlockHeight == 0`, then call InitChain.
```
cason, line 847:
```suggestion
If `appBlockHeight == 0`, then call `InitChain`.
```

spec/abci/abci++_app_requirements.md
```diff
@@ -837,6 +846,12 @@ First, some simple start conditions:
...
 
+Note that, in case there is a failure to commit the first block and
+a crash occurs,
+upon restart, `InitChain` will be called again. As explained,
+the application should not see this as an issue,
```
cason, line 852: The render of this part is odd. The lines are "broken" like in the source code.

spec/abci/abci++_app_requirements.md
```diff
@@ -837,6 +846,12 @@ First, some simple start conditions:
 
 If `appBlockHeight == 0`, then call InitChain.
 
+Note that, in case there is a failure to commit the first block and
+a crash occurs,
+upon restart, `InitChain` will be called again. As explained,
+the application should not see this as an issue,  
+as it is expected to crash together with Comet.
```
cason, line 853: To call it Comet reveals a certain intimacy with him. :P

But in the remaining of the document, CometBFT has been adopted.

cason (approved):
Nitpicking++.

spec/abci/abci++_app_requirements.md
jmalicevic: I guess a more accurate descrption here would be that `Commet` never completes the execution twice? As in, if it crashes before the return of `Commit` (and state persistance), it is technically possible to call it again?

cason: Yes. But I didn't want to make this description (on the sentence) too long.

cason: Actually, I am not sure anymore, after reading the comment block here: https://github.com/cometbft/cometbft/blob/main/consensus/state.go#L1672

spec/abci/abci++_app_requirements.md
jmalicevic: Is the value assigned mid-commit before the crash persisted? As in, is the state not updated atomically (or at least saved)?

cason: I hope #493 can help us.

lasarojc: > Is the value assigned mid-commit before the crash persisted? As in, is the state not updated atomically (or at least saved)?

It must be the case that both values are updated during commit, very likely as the the last steps, and the problem is ensuring that both updates are part of one atomic operation, probably along with putting the final nail on the old's state coffin. Since we don't control that, as it happens on the app side, we must require the app to implement this behavior.

spec/abci/abci++_app_requirements.md
lasarojc: Maybe the problem is referring to too things that are too general as synchronization points. I mean, for example, we shouldn't refer to "execute" commit or "return from" commit, but to the atomic operation of updating the height, hash, and the app state inside commit.

jmalicevic: That's a good point. because in my mind, a successful commit means the state is update as well as all other relevant fields. Maybe we should first specify what exactly is a successful execution of commit (on our side at least).

spec/abci/abci++_app_requirements.md
cason: The application typically uses something like leveldb. Correct me if I am wrong, but this kind of databases do not support transactions. So, while both writes are probably atomic, if a crash occurs between them the state can become inconsistent.

spec/abci/abci++_app_requirements.md
```diff
@@ -810,9 +810,21 @@ implementation of
...
+last block it successfully completed `Commit` for. 
 
-If the app successfully committed block H, then `last_block_height = H` and `last_block_app_hash = <hash returned by Commit for block H>`. If the app
+The application is expected to persist its state during the `Commit` method.
+CometBFT and the application are expected to crash together and there should not 
+exist a scenario where the application has persisted state that CometBFT has not.
+
+In other words, if `Commit` for a block H has not been executed and CometBFT crashed,
+the application
+should not have persisted any state between the `Commit` for H - 1 and
```
sergio-mena, line 821: It's actually H-N (I know it's surprising, but it's like that). The App does not need to persist every height

spec/abci/abci++_app_requirements.md
```diff
@@ -810,9 +810,21 @@ implementation of
 
 On startup, CometBFT calls the `Info` method on the Info Connection to get the latest
 committed state of the app. The app MUST return information consistent with the
-last block it successfully completed Commit for.
+last block it successfully completed `Commit` for. 
 
-If the app successfully committed block H, then `last_block_height = H` and `last_block_app_hash = <hash returned by Commit for block H>`. If the app
+The application is expected to persist its state during the `Commit` method.
+CometBFT and the application are expected to crash together and there should not 
+exist a scenario where the application has persisted state that CometBFT has not.
```
sergio-mena, line 817:
```suggestion
exist a scenario where the application has persisted state of a height greater than the latest one persisted by CometBFT.
```

spec/abci/abci++_app_requirements.md
sergio-mena: The main invariant is that the App state is always committed after consensus persists the results (returned by the app) of executing the block.
So the app and consensus "persist" actions don't need to be atomic. Upon recovery, there is some logic, poorly documented ATM in a section of the spec, to replay anything the app has "forgotten" (this is figured out with the handshake after recovery, using `Info` method).

sergio-mena (approved):

spec/abci/abci++_app_requirements.md
lasarojc: Yes, I think that would work.

spec/abci/abci++_app_requirements.md
```diff
@@ -808,67 +808,61 @@ implementation of
...
-stateBlockHeight = height of the last block for which CometBFT completed all
-    block processing and saved all ABCI results to disk
-appBlockHeight = height of the last block for which ABCI app successfully
-    completed Commit
+last block it successfully completed `Commit` for. 
+
+The three steps performed before the state of a height is considered persisted are: 
+- The block is stored by CometBFT in the blockstore
+- CometBFT has stored the state (results returned from the application)
+- The application has committed it's state within `Commit`.
```
lasarojc, line 825:
```suggestion
- CometBFT has stored the state returned by the application throuch `FinalizeBlockResult`
- The application has committed its state within `Commit`. 
```

spec/abci/abci++_app_requirements.md
```diff
@@ -808,67 +808,61 @@ implementation of
...
+
+``` 
+APP:                                              Execute block                         Persist state
+                                                 /     return ResultFinalizeBlock            /
+                                                /                                           /  
+Event: ------------- block_stored -------------/ -------------state_stored ----------------/---app_persisted_state
+                          |                   /                   |                       /     |
+CometBFT: Decides - Persistes block ---Call `FinalizeBlock`  Persist results------------Commit---- 
+            on        in the                                (txResults, validator
+          Block      block store                                updated,..)
```
lasarojc, line 839:
```suggestion
``` 
APP:                                              Execute block                           Persist application
                                                / return ResultFinalizeBlock            / state
                                               /                                       /    |
Event: -------------- block_stored -----------/-------------state_stored -------------/---app_persisted_state
                          |                  /                   |                   /     
CometBFT: Decide ---- Persist ----------- Call ------------- Persist ------------ Call Commit---- 
          on Block    block in the        FinalizeBlock      results 
                      blockstore                             (txResults, validator updated,..)
```

File: spec/abci/abci++_app_requirements.md [hunk 12]
```diff
@@ -808,67 +808,61 @@ implementation of
...
-If `appBlockHeight < stateBlockHeight`
-    replay all blocks in full from `appBlockHeight` to `storeBlockHeight-1`,
-    and replay the block at `storeBlockHeight` using the WAL.
-This happens if the app forgot the last block it committed.
-
-If `appBlockHeight == stateBlockHeight`,
-    replay the last block (storeBlockHeight) in full.
-This happens if we crashed before the app finished Commit
+As these three steps are not atomic, we observe different cases based on which steps have been executed
+before the crash occured
```
lasarojc, line 844:
```suggestion
before the crash occurred
```

spec/abci/abci++_app_requirements.md
```diff
@@ -808,67 +808,61 @@ implementation of
...
+- `block_stored` and `state_stored`: As the app did not persist its state within `Commit`, we need to re-execute
+  `FinalizeBlock` to retrieve the results and compare them to the state stored by CometBFT within `state_stored`. 
+  The expected case is that the states will match, otherwise CometBFT panics.
+- `block_stored`, `state_stored`, `app_persisted_state`: we move on to the next height. 
+
+Based on the sequence of these events, CometBFT will panic if any of the steps in the sequence happen out of order,
+that is if: 
+- The application has persisted a block at a height higher than the blocked saved during `state_stored`.
+- The `block_stored` step persisted a block at a height smaller than the `state_stored`
+- And the difference between the heights of the blocks persited by `state_stored` and `block_stored` is more
```
lasarojc, line 858:
```suggestion
- And the difference between the heights of the blocks persisted by `state_stored` and `block_stored` is more 
```

spec/abci/abci++_app_requirements.md
```diff
@@ -808,84 +808,61 @@ implementation of
...
+ABCI functions that are called and executed by CometBFT and the application:
+
+
+``` 
+APP:                                              Execute block                         Persist state
+                                                 /     return ResultFinalizeBlock            /
+                                                /                                           /  
+Event: ------------- block_stored -------------/ -------------state_stored ----------------/---app_persisted_state
+                          |                   /                   |                       /     |
+CometBFT: Decides - Persistes block ---Call `FinalizeBlock`  Persist results------------Commit----
```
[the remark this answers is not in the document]
lasarojc: moved the suggestion and made a few others to a comment with larger scope.

spec/abci/abci++_app_requirements.md
jmalicevic: This is rendered weirdly :) I got the suggestion I think, but please confirm :)

spec/abci/abci++_app_requirements.md
```diff
@@ -804,80 +804,65 @@ The expectation is for there to be some number of high level paths
...
-the app.
-
-```md
-storeBlockHeight = height of the last block CometBFT saw a commit for
-stateBlockHeight = height of the last block for which CometBFT completed all
-    block processing and saved all ABCI results to disk
-appBlockHeight = height of the last block for which ABCI app successfully
-    completed Commit
+In practice, persisting the state of a height consists of three steps, the last of which 
+is the call to the application's `Commit` method, where the application is expected to
```
sergio-mena, line 816:
```suggestion
is the call to the application's `Commit` method, the only place where the application is expected to
```

spec/abci/abci++_app_requirements.md
```diff
@@ -804,80 +804,65 @@ The expectation is for there to be some number of high level paths
...
+persist/commit its state.
+On startup, CometBFT calls the `Info` method on the Info Connection to get the latest
```
sergio-mena, line 818:
```suggestion
On startup (upon recovery), CometBFT calls the `Info` method on the Info Connection to get the latest
```

spec/abci/abci++_app_requirements.md
```diff
@@ -804,80 +804,65 @@ The expectation is for there to be some number of high level paths
...
+committed state of the app. The app MUST return information consistent with the
+last block for which it successfully completed `Commit`. 
+
+The three steps performed before the state of a height is considered persisted are: 
+- The block is stored by CometBFT in the blockstore
+- CometBFT has stored the state returned by the application through `FinalizeBlockResult`
```
sergio-mena, line 824:
```suggestion
- CometBFT has stored the state returned by the application through `FinalizeBlockResponse`
```

sergio-mena (approved):
Awesome improvement of the crash-recovery section, finally! Thanks for taking care of this!

Issue #10, opened 2022-12-23
ABCI 2.0: Final adjustments


sergio-mena:
Tasks:

  - [x] #8
  - [x] #90
    - see what is missing from tendermint/tendermint#8158
  - [x] #97
  - [x] Double-check if there is something we need from tendermint/tendermint#8414 (probably not)
  - [x] #471 (backport leftovers to `v0.37.x` #479)
    - see what metrics we are still missing from tendermint/tendermint#8480 (mainly vote extensions)
  - [x] Redundant params in `FinalizeBlock`, and obsolete methods (`BeginBlock`, etc). Double check that changes in tendermint/tendermint#8598 and/or tendermint/tendermint#8633 are unnecessary
  - [x] #478
    - move `app_hash` from `Commit` to `FinalizeBlock`: follow tendermint/tendermint#8664
  - [x] #488
  - [x] #490
  - [x] #485
  - [x] Leftovers
    - #514 
    - #515
  - [x] Bugfixes
    - #539
    - #542
    - #546
    - #562
  - [x] #547
  - [x] Once we are code complete, and all CI tests pass, merge branch back into `main`
    - #568
    - #572
  - [x] #510
  - [x] Second part of #409 (remove config to disable blocksync)

DoD:

  - No errors in CI
  - No major concerns if testnets were run
  - Branch `feature/abci++vef` has been merged back into `main`

Original issue: tendermint/tendermint#9916

sergio-mena:
Closing as all tasks have been completed

PR #471, opened 2023-03-06
Backport vote extension metrics

sergio-mena:
Contributes to #10

Partial cherry-pick of #8480

The rest of #8480 was backported for `v0.37.x`.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/consensus/metrics.go
+++ b/consensus/metrics.go
@@ -91,6 +91,11 @@ type Metrics struct {
 	//metrics:Interval in seconds between the proposal timestamp and the timestamp of the latest prevote in a round where all validators voted.
 	FullPrevoteDelay metrics.Gauge `metrics_labels:"proposer_address"`
 
+	// VoteExtensionReceiveCount is the number of vote extensions received by this
+	// node. The metric is annotated by the status of the vote extension from the
+	// application, either 'accepted' or 'rejected'.
+	VoteExtensionReceiveCount metrics.Counter
+
 	// ProposalReceiveCount is the total number of proposals received by this node
 	// since process start.
 	// The metric is annotated by the status of the proposal from the application,
@@ -99,6 +104,8 @@ type Metrics struct {
 
 	// ProposalCreationCount is the total number of proposals created by this node
 	// since process start.
+	// The metric is annotated by the status of the proposal from the application,
+	// either 'accepted' or 'rejected'.
 	ProposalCreateCount metrics.Counter
 
 	// RoundVotingPowerPercent is the percentage of the total voting power received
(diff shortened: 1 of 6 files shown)


docs/core/metrics.md
```diff
@@ -44,6 +44,7 @@ The following metrics are available:
 | consensus\_block\_gossip\_parts\_received  | Counter   | matches\_current | Number of block parts received by the node                                                                                                 |
 | consensus\_quorum\_prevote\_delay          | Gauge     |                  | Interval in seconds between the proposal timestamp and the timestamp of the earliest prevote that achieved a quorum                        |
 | consensus\_full\_prevote\_delay            | Gauge     |                  | Interval in seconds between the proposal timestamp and the timestamp of the latest prevote in a round where all validators voted           |
+| consensus\_vote\_extension\_receive\_count | Counter   | status           | Number of vote extensions received                                                                                                         |
```
jmalicevic, line 47: This is just a backport so it's fine but to mee somehow this metric is showing two things (number and status) and I am not sure whether one could extract stats on the number of rejects or accepts. Maybe it is not even relevant, but was a bit counterintuitive at first.

consensus/metrics.go
```diff
@@ -138,6 +153,14 @@ func (m *Metrics) MarkRound(r int32, st time.Time) {
...
 	roundTime := time.Since(st).Seconds()
 	m.RoundDurationSeconds.Observe(roundTime)
+
+	pvt := cmtproto.PrevoteType
+	pvn := strings.ToLower(strings.TrimPrefix(pvt.String(), "SIGNED_MSG_TYPE_"))
+	m.RoundVotingPowerPercent.With("vote_type", pvn).Set(0)
+
+	pct := cmtproto.PrecommitType
+	pcn := strings.ToLower(strings.TrimPrefix(pct.String(), "SIGNED_MSG_TYPE_"))
+	m.RoundVotingPowerPercent.With("vote_type", pcn).Set(0)
```
jmalicevic, line 163: How are these changes related to vote extensions and why is the count here set to 0 - because we start a new round?

sergio-mena: The original patch (the one I'm cherry-picking from) was not limited to vote extensions. This hunk is a left-over that hadn't been backported from the original patch.
This is the typical example why we're following this process (i.e., cherry-picking patches that are supposed to have already been backported manually)

docs/core/metrics.md
sergio-mena: My interpretation is that you want to count both, accepted and rejected, as two "sub-metrics" (that's my understanding of `MarkVoteExtensionReceived`

jmalicevic: You should definitely count both towards the number of VE seen. I was just wondering whether someone could further count the number within each sub-metric. But not really a major issue, just discussing.

jmalicevic (approved):

Issue #385, opened 2023-02-24
Upgrade to use go 1.20


lasarojc:
As asked in #174, in general, what are our criteria for updating to support a new version of Go?

sergio-mena:
In abstract, I would tend to be more conservative overall. However, with the (über-strict IMO) policy of only supporting the last to versions of golang, I don't think we can afford to be that conservative.
So I support upgrading `main` to the latest version as soon as it is out.

cason:
Yes, I don't think Golang gives us much choice here. We had 3 versions (1.18 - 1.20) in less than one year. We should hope this doesn't become the norm.

thanethomson:
Yes, the longer we wait to update to the latest version of Go on `main`, the more likely our releases' minimum Go versions are going to end up being unsupported, potentially _increasing_ the likelihood of issues for networks upgrading to new versions (i.e. being more conservative in this case potentially _increases_ risk overall).

lasarojc:
I also like the idea of keeping up with Go and giving the comments above we should move HEAD to Go 1.20. I guess the sooner the better, so people are well warned about it.

What about 0.37? Should we start coordinating with our customers to update to 1.20? Some of them are already there.

thanethomson:
In general we should upgrade to the latest release of Go as soon as pragmatically feasible after it comes out to maximize the lifespan of a release using a single version of Go.

What would dramatically reduce any risks there is if we had some way of testing mixed networks of app chains (e.g. with Gaia) where some binaries are built using one version of Go and other are built using a different version. That's where the real risk comes in. Not during coordinated upgrades.

cason:
An important point to consider is that go 1.20 is not available as a package in some important Linux distros:

- Debian offers 1.19, even for the for the unstable/sid suite: https://packages.debian.org/search?keywords=golang&searchon=names&exact=1&suite=all&section=all
- Ubuntu 22.04 and 22.10 offers 1.19, only the development `lunar` suite, which will become 23.04, offers 1.20: https://packages.ubuntu.com/search?keywords=golang&searchon=names&exact=1&suite=all&section=all

thanethomson:
Does anyone exclusively use the standard versions of Go provided by specific distros? Personally, I either manually download and symlink the latest version of Go, or use Homebrew/Linuxbrew.

Some users only use the Docker images we build, and some will build their own Docker images using the Go 1.20 base images.

sergio-mena:
[...]

PR #360, opened 2023-02-20
Upgrade to Go 1.19 on `0.34.x`

lasarojc:
Partially addresses #174 

The SDK uses Go 1.18 from their release that integrates with v0.34, but they should upgrade

---

#### PR checklist

- [ ] Tests written/updated
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments

PR #474, opened 2023-03-06
Post-release root docs update

thanethomson:
Updates our root [changelog](https://github.com/cometbft/cometbft/blob/thane/post-release-docs-update/CHANGELOG.md) and [upgrading guidelines](https://github.com/cometbft/cometbft/blob/thane/post-release-docs-update/UPGRADING.md) on `main` to sync them with what we have on `v0.34.x` and `v0.37.x`.

Also makes a few minor changes to the [README](https://github.com/cometbft/cometbft/blob/thane/post-release-docs-update/README.md).

---

#### PR checklist

- [ ] Tests written/updated


.changelog/v0.34.27/breaking-changes/360-update-to-go-119.md
```diff
@@ -0,0 +1,2 @@
+- Use Go 1.19 to build CometBFT, since Go 1.18 has reached end-of-life.
```
sergio-mena, line 1: Since there is \#385 listed below, I'm not sure including this done achieves anything (think of a user reading these). It could even be misleading.

I'd remove this one

sergio-mena (approved):

sergio-mena:
Question: let's say we backport something (next week) from `main` to `v0.37.x`. How do we make sure the changelog for that change also lands on the `v0.37x` section on branch `main`?

.changelog/v0.34.27/breaking-changes/360-update-to-go-119.md
thanethomson: I'll do this in a follow-up PR.

thanethomson:
> Question: let's say we backport something (next week) from `main` to `v0.37.x`. How do we make sure the changelog for that change also lands on the `v0.37x` section on branch `main`?

This is a good question. I've been thinking about this recently, and I would say that we should probably only add changelog entries to PRs in two cases:
1. When we're submitting a PR to `main` that won't be backported (i.e. it'll land on the new backport branch we create when we start work towards a new version, so we'd want a changelog entry for it by then).
2. When we're submitting a PR to a backport branch (e.g. `v0.34.x` and `v0.37.x`). In this case, if the initial PR to `main`, then we shouldn't include the changelog entry there - only on the backport PRs to the backport branches.

This way, when we cut a new release, it's easy to just copy across all the `.changelog` entries to `main` for those releases when we cut them. Otherwise we'll end up with duplicate or incorrect changelog entries that we then need to de-duplicate (this is also a valid approach, but arguably more difficult and time-consuming than doing the work up-front).

I'll see if there's an easy way to codify this in our PR template and contributing guidelines.

.changelog/v0.34.27/breaking-changes/360-update-to-go-119.md
thanethomson: Oh, but isn't #385 relevant to v0.37.0, while #360 is relevant to v0.34.27?

Issue #94, opened 2023-01-11
proto: Rename packages from `tendermint` -> `cometbft`


thanethomson:
The packages of all Protobuf definitions in the `proto` folder of on branch `main` must be renamed from `tendermint` to `cometbft`. We intend on making substantial changes to the architecture and interfaces in the coming year and, as such, the protos could diverge substantially from Tendermint Core over time (only across major releases, of course). We also have no intention of remaining compatible with Tendermint Core in future major releases (this is already actually the case with our v0.37 release). As such, it makes very little sense, from a product perspective, to continue producing interface/data structure changes for Tendermint Core.

Having our proto package as `cometbft` will make publishing these protos to registries, e.g. https://buf.build/cometbft/cometbft, easier for us and for integrators who consume them (especially once we version our protos, in #95).

e.g. https://github.com/cometbft/cometbft/blob/754f1ae9e29a84754aa42ac969f0cd369266f019/proto/tendermint/types/block.proto#L2

must be renamed to:

```protobuf
package cometbft.types;
```

This includes regenerating all of the Go code and resolving any import errors.

Branches `v0.37.x` and `v0.34.x` will stay as they are:

* in order to minimize disruption in the ecosystem
* all protos in those branches are reused verbatim from the Tendermint Core repo

Further, we cannot do this for the `v0.34.x` branch because we would be breaking the message URLs associated with these messages, meaning CometBFT would potentially be incompatible with Tendermint Core, meaning that networks would need to perform a coordinated upgrade to CometBFT. Ideally we want our first v0.34-series release to be a drop-in replacement, so that a coordinated upgrade isn't necessary.

mzabaluev:
There are gRPC services defined in these files:

| File | Service name |
| ---- | ------------------- |
| `abci/types.proto` | `ABCIApplication` |
| `rpc/grpc/types.proto` | `BroadcastAPI` |

Changing the package name will result in changing the gRPC endpoints for these services as used by the generated code.
Do we want to go forward with it as part of this renaming, or should I keep these packages in the `tendermint` namespace?

thanethomson:
> Changing the package name will result in changing the gRPC endpoints for these services as used by the generated code. Do we want to go forward with it as part of this renaming, or should I keep these packages in the tendermint namespace?

[...]

Issue #95, opened 2023-01-11
proto: Implement versioning


thanethomson:
Relates to tendermint/tendermint#9197

Depends on #94

In order to better support integrators, we need to version our Protobuf definitions as per [Buf's guidelines](https://docs.buf.build/best-practices/module-development#module-layout). The goal here is to have:
- [x] Our Protobuf definitions in the `/proto` folder in the repo. https://github.com/cometbft/cometbft/pull/495
- [x] Our generated Protobuf Go code in a new `/api` folder in the repo. https://github.com/cometbft/cometbft/pull/495
- [x] Document how to translate protobuf messages across versions by annotating the `.proto` files
  - [x] #780

mzabaluev:
After discussion with @thanethomson, I understand that we need to take care of snapshots of proto definitions going back to 0.34.

The proposed packaging versioning schema should look as follows:

* Any message definitions that were new in `v0.34.x` get packaged with the suffix `.v1` appended to their domain package name.
* Message definitions that have changed in subsequent release streams (`v0.37.x` and `main`/0.38 at the moment) in ways that introduce breaking changes to generated code in strictly typed languages (our primary concern being Go and Rust), are put into packages with incremental suffixes `.v2`, `.v3`, and so on. There is no direct correspondence between a package versioning suffix and any CometBFT protocol version; the suffixes are only to track evolution of message types.
* Messages newly introduced since 0.34 may land into `.v1` or later packages. This depends on which package's versioned namespace looks more natural considering type crosslinks and changes to adjacent message types in the same domain package.

thanethomson:
> Regarding the output folder for the generated files, I'd prefer a single location including the ABCI stubs.
This would make buf.gen.yaml the single source of truth for the protobuf generation process, removing the tricky post-build Makefile actions.

This is precisely the idea.

mzabaluev:
I've hit a problem in #495 with the `Wrap` API method that is manually implemented for generated variant types of `oneof` content fields for types like `Message`, e.g. in `consensus`.

This method ([permalink](https://github.com/cometbft/cometbft/blob/29765b988e45ea1488e7f1ce1bafd5aceb1cf132/proto/cometbft/consensus/v1/message.go#L21)), for example, has to be defined in `consensus/v1`, because `VoteSetBits` is defined in the corresponding proto package and the generated struct ends up in `consensus/v1/types.pb.go`:

```golang
func (m *VoteSetBits) Wrap() proto.Message {
[...]

PR #482, opened 2023-03-08
proto: rename packages from `tendermint` to `cometbft`

mzabaluev:
Work on #94

---

#### PR checklist

- [ ] Tests written/updated


mzabaluev:
As initially pushed, this is just a mechanical rename of the packages and regeneration of the `*.pb.go` files on top of the feature branch's baseline.

This PR is currently posted only for reference.
I'm working on an attempt to introduce version suffixes for #95 on top of these changes, before integrating the generated code into the rest of the code base. Doing the massive renaming twice would be a waste of effort.

abci/types/types.pb.go
```diff
@@ -3844,7 +3843,7 @@ func NewABCIClient(cc grpc1.ClientConn) ABCIClient {
 
 func (c *aBCIClient) Echo(ctx context.Context, in *RequestEcho, opts ...grpc.CallOption) (*ResponseEcho, error) {
 	out := new(ResponseEcho)
-	err := c.cc.Invoke(ctx, "/tendermint.abci.ABCI/Echo", in, out, opts...)
```
mzabaluev, line 3847: This is the gRPC protocol breakage I'm talking about in https://github.com/cometbft/cometbft/issues/94#issuecomment-1456520852

thanethomson:
If you move all of the `*.proto` files instead of copying them and then renaming them, how do we continue to offer wire-compatible `tendermint.*` package protos to users as previously discussed?

mzabaluev:
Superseded by #495.

PR #484, opened 2023-03-08
spec: abci2.0 -Incorporate content of RFC100 into the spec

### LINKED ISSUE #483: Turn RFC100 into part of the spec
The content in RFC100  aims at defining the changes within CometBFT needed to support vote extensions, as well as information on how to upgrade from earlier versions of ABCI to ABCI 2.0. Part of that text should be contained in the abci specification.

jmalicevic:
Closes #483.


spec/abci/abci++_app_requirements.md
```diff
@@ -1015,6 +1015,8 @@ from the genesis file and light client RPC servers. It also calls `Info` to veri
   current height's block header
 
 Once the state machine has been restored and CometBFT has gathered this additional
-information, it transitions to block sync (if enabled) to fetch any remaining blocks up the chain
-head, and then transitions to regular consensus operation. At this point the node operates like
+information, it transitions to block sync. As of ABCI 2.0, if vote extensions are enabled, 
+block sync is required to sync
+at least one block before switching to consensus. This way, the node is guaranteed to have obtained
+the vote extensions (stored within the extended commit in the block store), which are needed to participate in consensus. The node can then transition to regular consensus operation. At this point the node operates like
```
sergio-mena, line 1021: This is good, but could you add a reference here to RFC100, to the place we explain the condition for switching in detail?

spec/abci/abci++_comet_expected_behavior.md
```diff
@@ -126,6 +126,24 @@ Let us now examine the grammar line by line, providing further details.
...
+>sync-blocks         = request-block apply-block update-state
+>```
+
+Thus the entire sequence of startup via syncing protocols before switching to consensus would look as follows:
+
+>```abnf
+>startup          = state-sync block-sync recovery
+>```
+
+
```
sergio-mena, line 146: I would not add this, as the present grammar does not distinguish between block sync and normal consensus behavior. This was done of purpose as the app shouldn't be able to tell.

spec/abci/abci++_app_requirements.md
```diff
@@ -1015,6 +1015,39 @@ from the genesis file and light client RPC servers. It also calls `Info` to veri
...
+
+Introducing vote extensions requires changes to the configuration of the application.
+
+First of all, switching to a version of CometBFT with vote extensions, requires a coordinated upgrade. 
+For a detailed description on the upgrade path, please refer to the corresponding 
+[section](./../../docs/rfc/rfc-100-abci-vote-extension-propag.md#upgrade-path) in RFC-100.
+
+The configuration file now contains a [**new consensus parameter**](./abci%2B%2B_app_requirements.md#abciparamsvoteextensionsenableheight): `VoteExtensionsEnableHeight`. 
+This parameter represents the height after which vote extensions are 
+required for consensus to proceed, with 0 being the default value (no vote extensions).
```
jmalicevic, line 1036: @sergio-mena , maybe I have not a clear understanding, but let's say a chain is just starting and wants to have vote extensions, but the default value here is 0. Should we not say something like - we have to set this to at least 1? Because someone starting from scratch with a version of the software having extensions enabled, might assume they are enabled by default?

spec/abci/abci++_app_requirements.md
```diff
@@ -1015,6 +1015,39 @@ from the genesis file and light client RPC servers. It also calls `Info` to veri
...
+Once the (coordinated) upgrade to ABCI 2.0 has taken place, at height  *h<sub>u</sub>*,
+the value of `VoteExtensionsEnableHeight` MAY be set to some height, *h<sub>e</sub>*,
+which MUST be higher than the current height of the chain. Thus the earliest value for 
+ *h<sub>e</sub>* is  *h<sub>u</sub>* + 1. 
+Once a node reaches the configured height,
+for all heights *h ≥ h<sub>e</sub>*, the consensus algorithm will
+reject any votes that do not have vote extension data as invalid.
+Likewise, for all heights *h < h<sub>e</sub>*, any votes that *do* have vote extensions
+will be considered an error condition.
+Height *h<sub>e</sub>* is somewhat special, as calls to `PrepareProposal` MUST NOT
```
lasarojc, line 1046: Is this comment necessary here, as this behavior will be enforced by CometBFT itself?

spec/abci/abci++_comet_expected_behavior.md
```diff
@@ -224,3 +227,35 @@ As for the new methods:
...
+## Accomodating for vote extensions
+
+In a manner transparent to the application, CometBFT ensures the node is provided with all
+the data it needs to participate in consensus. 
+
+In the case of recovering from a crash, or joining the network via state sync, CometBFT will make
+sure the node acquires the neccessary vote extensions before switching to conensusus. 
+
+If a node is already in consensus but falls behind, during catch-up, CometBFT if providing the node with
```
lasarojc, line 239: confusing.
Should `CometBFT if providing` become `CometBFT will provide`?

spec/abci/abci++_comet_expected_behavior.md
```diff
@@ -224,3 +227,35 @@ As for the new methods:
...
+
+## Handling upgrades to ABCI 2.0 
+
+If applications upgrade to ABCI 2.0, CometBFT internally ensures that the [application setup](./abci%2B%2B_app_requirements.md#application-configuration-required-to-switch-to-abci-20) is reflected in its opertaion. 
+
+Namely, upon saving the block for a given height *h* in the block store at decision time
+- if *h ≥ h<sub>e</sub>*, the corresponding extended commit is saved as  well
+- if *h < h<sub>e</sub>*, there are no changes to the data saved
+
+In the catch-up mechanism, when a node f realizes that another peer is at height hp, which is more than 2 heights behind,
```
lasarojc, line 254:
```suggestion
In the catch-up mechanism, when a node *f* realizes that another peer is at height *h<sub>p</sub>*, which is more than 2 heights behind,
```

spec/abci/abci++_comet_expected_behavior.md
jmalicevic: Yes, spelling mistake.

spec/abci/abci++_app_requirements.md
jmalicevic: I was also thinking about this.. However, this felt to me like information the app should be aware in terms of ,it is the one that has to provide this. CometBFT will enforce it and make sure they are there, otherwise crash.. But you as the application have to know that for this height you need to implement this behaviour.

spec/abci/abci++_comet_expected_behavior.md
```diff
@@ -242,3 +227,39 @@ As for the new methods:
...
+
+Introducing vote extensions requires changes to the configuration of the application. Additionally, 
+operators and application developers need to be aware of the impact vote extensions have on
+expected runtime behaviour.
+
+First of all, switching to a version of CometBFT with vote extensions, requires a coordinated upgrade. 
+For a detailed description on the upgrade path, please refer to the corresponding 
+[section](./../../docs/rfc/rfc-100-abci-vote-extension-propag.md#upgrade-path) in RFC-100.
+
+Once upgraded, the configuration file is updated with a [**new consensus parameter**](./abci%2B%2B_app_requirements.md#abciparamsvoteextensionsenableheight): `VoteExtensionsEnableHeight`.
```
sergio-mena:
```suggestion
Once upgraded, the new version contains a [**new consensus parameter**](./abci%2B%2B_app_requirements.md#abciparamsvoteextensionsenableheight): `VoteExtensionsEnableHeight`. 
```

spec/abci/abci++_comet_expected_behavior.md
```diff
@@ -242,3 +227,39 @@ As for the new methods:
...
+Once upgraded, the configuration file is updated with a [**new consensus parameter**](./abci%2B%2B_app_requirements.md#abciparamsvoteextensionsenableheight): `VoteExtensionsEnableHeight`. 
+This parameter represents the height after which vote extensions are 
+required for consensus to proceed, with 0 being the default value (no vote extensions).
+Once the (coordinated) upgrade to ABCI 2.0 has taken place, at height  *h<sub>u</sub>*,
+the value of `VoteExtensionsEnableHeight` MAY be set to some height, *h<sub>e</sub>*,
+which MUST be higher than the current height of the chain. Thus the earliest value for 
+ *h<sub>e</sub>* is  *h<sub>u</sub>* + 1. 
+Once a node reaches the configured height,
+for all heights *h ≥ h<sub>e</sub>*, the consensus algorithm will
+reject any votes that do not have vote extension data as invalid.
```
sergio-mena:
```suggestion
reject as invalid any precommit messages that do not have signed vote extension data.
A 0-length vote extensions is allowed, but it MUST be signed.
```

spec/abci/abci++_comet_expected_behavior.md
```diff
@@ -242,3 +227,39 @@ As for the new methods:
...
+Height *h<sub>e</sub>* is somewhat special, as calls to `PrepareProposal` MUST NOT
+have vote extension data, but all precommit votes in that height MUST carry a vote extension.
+Height *h<sub>e</sub> + 1* is the first height for which `PrepareProposal` MUST have vote
+extension data and all precommit votes in that height MUST have a vote extension.
+
+
+### Changes to data stored by CometBFT
+
+Upon saving the block for a given height *h* in the block store at decision time
+    - if *h ≥ h<sub>e</sub>*, the corresponding extended commit is saved as  well
```
sergio-mena:
```suggestion
    - if *h ≥ h<sub>e</sub>*, the corresponding extended commit that was used to decide locally is saved as well
```

sergio-mena (dismissed):
Thanks!
EDIT: Sorry, I must have done something wrong, because I just noticed I was reviewing an old version of this PR. Bear with me

spec/abci/abci++_app_requirements.md
```diff
@@ -1015,6 +1015,39 @@ from the genesis file and light client RPC servers. It also calls `Info` to veri
...
+
+## Application configuration required to switch to ABCI 2.0
+
+Introducing vote extensions requires changes to the configuration of the application.
+
+First of all, switching to a version of CometBFT with vote extensions, requires a coordinated upgrade. 
+For a detailed description on the upgrade path, please refer to the corresponding 
+[section](./../../docs/rfc/rfc-100-abci-vote-extension-propag.md#upgrade-path) in RFC-100.
+
+The configuration file now contains a [**new consensus parameter**](./abci%2B%2B_app_requirements.md#abciparamsvoteextensionsenableheight): `VoteExtensionsEnableHeight`.
```
sergio-mena, line 1034:
```suggestion
There is a newly introduced [**new consensus parameter**](./abci%2B%2B_app_requirements.md#abciparamsvoteextensionsenableheight): `VoteExtensionsEnableHeight`. 
```

spec/abci/abci++_app_requirements.md
```diff
@@ -1015,6 +1015,39 @@ from the genesis file and light client RPC servers. It also calls `Info` to veri
...
+required for consensus to proceed, with 0 being the default value (no vote extensions).
+Once the (coordinated) upgrade to ABCI 2.0 has taken place, at height  *h<sub>u</sub>*,
+the value of `VoteExtensionsEnableHeight` MAY be set to some height, *h<sub>e</sub>*,
+which MUST be higher than the current height of the chain. Thus the earliest value for 
+ *h<sub>e</sub>* is  *h<sub>u</sub>* + 1. 
+Once a node reaches the configured height,
+for all heights *h ≥ h<sub>e</sub>*, the consensus algorithm will
+reject any votes that do not have vote extension data as invalid.
+Likewise, for all heights *h < h<sub>e</sub>*, any votes that *do* have vote extensions
+will be considered an error condition.
```
sergio-mena, line 1045:
```suggestion
Likewise, for all heights *h < h<sub>e</sub>*, any votes that *do* have vote extensions
will be considered an error condition.
```
```suggestion
Likewise, for all heights *h < h<sub>e</sub>*, any precommit messages that *do* have vote extensions
will also be rejected as malformed.
```

spec/abci/abci++_app_requirements.md
sergio-mena: Yes, I think we need to mention that "special" height, as it will be noticeable to the app: they will see no extensions in `PrepareProposal`, but then they will be asked to extend and verify vote extensions at the same height. So they (app developers) need to be aware of this quirk

spec/abci/abci++_comet_expected_behavior.md
```diff
@@ -118,7 +118,10 @@ Let us now examine the grammar line by line, providing further details.
   to provide the Application with all the snapshots needed, in order to reconstruct the state locally.
   A successful attempt must provide at least one chunk via `ApplySnapshotChunk`.
   At the end of a successful attempt, CometBFT calls `Info` to make sure the reconstructed state's
-  _AppHash_ matches the one in the block header at the corresponding height.
+  _AppHash_ matches the one in the block header at the corresponding height. Note that the state 
+  of  the application does not contain vote extensions itself. The application can rely on 
+  [CometBFT to ensure](./../../docs/rfc/rfc-100-abci-vote-extension-propag.md#base-implementation-persist-and-propagate-extended-commit-history)
+  the node has all the relevant data by the time it switches to consensus.
```
sergio-mena, line 124:
```suggestion
  the node has all the relevant data to proceed with the execution beyond this point. 
```
I'm skeptical about using "switching to consensus" in this section, as, from the point of view of the grammar (i.e., the App) consensus is not distinguishable from blocksync.

spec/abci/abci++_comet_expected_behavior.md
```diff
@@ -224,3 +227,35 @@ As for the new methods:
...
+## Accomodating for vote extensions
+
+In a manner transparent to the application, CometBFT ensures the node is provided with all
+the data it needs to participate in consensus. 
+
+In the case of recovering from a crash, or joining the network via state sync, CometBFT will make
+sure the node acquires the neccessary vote extensions before switching to conensusus. 
+
+If a node is already in consensus but falls behind, during catch-up, CometBFT will provide the node with 
+vote extensions from past heights by storing the extensions within `ExtendedCommit` for old heights.
```
sergio-mena, line 240:
```suggestion
vote extensions from past heights by retrieving the extensions within `ExtendedCommit` for old heights that it had previously stored.
```

spec/abci/abci++_comet_expected_behavior.md
```diff
@@ -224,3 +227,35 @@ As for the new methods:
...
+
+We realize this is sub-optimal due to the increase in storage needed to store the extensions, we are 
+working on an optimization of this implementation which should eliviate this concern. The decision to store
```
sergio-mena, line 243:
```suggestion
working on an optimization of this implementation which should alleviate this concern.
However, the application can use the existing `retain_height` parameter to decide how much
history it wants to keep, just as is done with the block history. The network-wide implications
of the usage of `retain_height` stays the same.
The decision to store 
```

spec/abci/abci++_comet_expected_behavior.md
```diff
@@ -224,3 +227,35 @@ As for the new methods:
...
+If a node is already in consensus but falls behind, during catch-up, CometBFT will provide the node with 
+vote extensions from past heights by storing the extensions within `ExtendedCommit` for old heights.
+
+We realize this is sub-optimal due to the increase in storage needed to store the extensions, we are 
+working on an optimization of this implementation which should eliviate this concern. The decision to store 
+historical commits and potential optimizations, are discussed in detail in [RFC-100](./../../docs/rfc/rfc-100-abci-vote-extension-propag.md#current-limitations-and-possible-implementations)
+
+## Handling upgrades to ABCI 2.0 
+
+If applications upgrade to ABCI 2.0, CometBFT internally ensures that the [application setup](./abci%2B%2B_app_requirements.md#application-configuration-required-to-switch-to-abci-20) is reflected in its opertaion.
```
sergio-mena, line 248:
```suggestion
If applications upgrade to ABCI 2.0, CometBFT internally ensures that the [application setup](./abci%2B%2B_app_requirements.md#application-configuration-required-to-switch-to-abci-20) is reflected in its operation. 
```

spec/abci/abci++_comet_expected_behavior.md
```diff
@@ -224,3 +227,35 @@ As for the new methods:
...
+We realize this is sub-optimal due to the increase in storage needed to store the extensions, we are 
+working on an optimization of this implementation which should eliviate this concern. The decision to store 
+historical commits and potential optimizations, are discussed in detail in [RFC-100](./../../docs/rfc/rfc-100-abci-vote-extension-propag.md#current-limitations-and-possible-implementations)
+
+## Handling upgrades to ABCI 2.0 
+
+If applications upgrade to ABCI 2.0, CometBFT internally ensures that the [application setup](./abci%2B%2B_app_requirements.md#application-configuration-required-to-switch-to-abci-20) is reflected in its opertaion. 
+
+Namely, upon saving the block for a given height *h* in the block store at decision time
+- if *h ≥ h<sub>e</sub>*, the corresponding extended commit is saved as  well
```
sergio-mena, line 251: You probably need to redefine *h<sub>e</sub>* here (since you can't assume the reader is reading the whole spec)

spec/abci/abci++_app_requirements.md
sergio-mena: Good point.
First, I realize we're not mentioning that the ConsensusParam can be set at genesis. So, typically, for chains starting from scratch and having vote extensions from their app's first version, they just need to set `VoteExtensionsEnableHeight` to 1 at genesis and forget about the upgrade path. We need to say this somewhere in the spec.
Second, to your question. We (our team) discussed about the default value at genesis last spring in some PR (I'll try to pull it up). I remember thinking default value at genesis should be 1, but then people convincing me that 0 is better. Now, when I think of it, it makes total sense to me that the default value is zero, for compatibility with existing genesis docs.
Let me know if you're still unclear in some aspect.

sergio-mena: > they just need to set VoteExtensionsEnableHeight to 1 
actually, to be precise, they need to set it to their `InitialHeight` parameter

jmalicevic: Yes, I know and agree it should be 0 by default. But as you said, I think we should state it that, if they do want vote extensions they have to change it, they will not magically be turned on. Of course, it can be deducted from the spec, but I think it's better to be clear on this as people might just skim through some parts of the spec.

spec/abci/abci++_comet_expected_behavior.md
jmalicevic: Ah I reread the sentence and you comment. You are not talking about removing the switching to consensus section, just about rewording this. I agree ,we can change this.

spec/abci/abci++_app_requirements.md
sergio-mena: > I'll try to pull it up

[Here](https://github.com/tendermint/tendermint/issues/8453#issuecomment-1116135296) it is (took some time :-) )

sergio-mena: Agreed.

Just make sure it's clear that there are two way of enabling vote extensions:
* Via app's logic at a given height, by changing the ConsensusParam
* Via setting it to `InitialHeight` in the genesis. This way, the app does not need to do anything to have extensions enabled from the very beginning.
BTW, this distinction also applies to other parts of CometBFT's state, such as validator updates.

spec/abci/abci++_app_requirements.md
```diff
@@ -1009,6 +1009,47 @@ from the genesis file and light client RPC servers. It also calls `Info` to veri
...
+
+First of all, switching to a version of CometBFT with vote extensions, requires a coordinated upgrade. 
+For a detailed description on the upgrade path, please refer to the corresponding 
+[section](./../../docs/rfc/rfc-100-abci-vote-extension-propag.md#upgrade-path) in RFC-100.
+
+There is a newly introduced [**consensus parameter**](./abci%2B%2B_app_requirements.md#abciparamsvoteextensionsenableheight): `VoteExtensionsEnableHeight`. 
+This parameter represents the height at which vote extensions are 
+required for consensus to proceed, with 0 being the default value (no vote extensions).
+A chain can enable vote extensions either:
+* at genesis by setting `VoteExtensionsEnableHeight` to be equal to the `InitialHeight`
```
adizere, line 1032: Is it valid to have `InitialHeight + 1` instead of `InitialHeight`?

spec/abci/abci++_app_requirements.md
adizere, line 1: This is out of the scope of this PR, but a general comment would be to consider moving the section "Formal Requirements" later in this doc.

Also: adding a table of content to this doc would help get an overview of it, since right now it's very large and has a lot of valuable information, but the structure is not obvious to me.

adizere: I'd be glad to open an issue to track my comments on this file, and fix that in a subsequent PR if it sounds like a nice-to-have later on.

adizere (commented):
I think I'm starting to understand ABCI++ better, the docs are very thorough!

spec/abci/abci++_app_requirements.md
jmalicevic: Yes this is also valid, but then vote extensions are not enabled from the start, rather frm the height right above the starting height. So if they want vote extensions enavbled from the very beginning it has to be set to `InitialHeight`

spec/abci/abci++_comet_expected_behavior.md
```diff
@@ -224,3 +227,41 @@ As for the new methods:
...
+However, the application can use the existing `retain_height` parameter to decide how much
+history it wants to keep, just as is done with the block history. The network-wide implications
+of the usage of `retain_height` stay the same.
+The decision to store 
+historical commits and potential optimizations, are discussed in detail in [RFC-100](./../../docs/rfc/rfc-100-abci-vote-extension-propag.md#current-limitations-and-possible-implementations)
+
+## Handling upgrades to ABCI 2.0 
+
+If applications upgrade to ABCI 2.0, CometBFT internally ensures that the [application setup](./abci%2B%2B_app_requirements.md#application-configuration-required-to-switch-to-abci-20) is reflected in its operation. 
+CometBFT retrieves from the application configuration the value of *h<sub>e</sub>*, the height at which vote extensions are required for consensus to proceed, and uses it to determine the data it stores and data it sends to a peer
```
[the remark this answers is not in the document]
jmalicevic: Good idea, done!

spec/abci/abci++_app_requirements.md
jmalicevic: I added to TOC to this file but refactoring the sections is better done in a follow up PR I believe..as this is supposed to be just a backport and the later sections might refer to the previous one atm. So refactoring the text might require a bit more time. I can check with Sergio on this as well.

jmalicevic:
@sergio-mena , do you mind taking a last look at the TOC and  the latest changes so we can merge this?

spec/abci/abci++_app_requirements.md
```diff
@@ -5,6 +5,32 @@ title: Requirements for the Application
...
+    - [Updating Consensus Parameters](#updating-consensus-parameters)
+  - [Query](#query)
+    - [Query Proofs](#query-proofs)
+    - [Peer Filtering](#peer-filtering)
+    - [Paths](#paths)
+  - [Crash Recovery](#crash-recovery)
+  - [State Sync](#state-sync)
+- [Application configuration required to switch to ABCI2.0](#application-configuration-required-to-switch-to-abci-20)
+
+
```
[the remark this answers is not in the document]
adizere: I [suggested](https://github.com/cometbft/cometbft/pull/484/files#r1141823124) to add a TOC, because I didn't realize GH has one. I managed to find it in the left-hand side of the view:

Screenshot 2023-03-21 at 09 33 49 

Thanks for highlighting that Lasaro! I still think having the TOC directly in the file does not hurt.

> something to think about going forward when writing specs, RFC and ADR.

Agree. We might want to have a consistent approach. Good discussion to have internally another time.

sergio-mena:
On it! (sorry I couldn't get back to this earlier)

spec/abci/abci++_app_requirements.md
```diff
@@ -1025,16 +1051,19 @@ First of all, switching to a version of CometBFT with vote extensions, requires
 For a detailed description on the upgrade path, please refer to the corresponding 
 [section](./../../docs/rfc/rfc-100-abci-vote-extension-propag.md#upgrade-path) in RFC-100.
 
-There is a newly introduced [**new consensus parameter**](./abci%2B%2B_app_requirements.md#abciparamsvoteextensionsenableheight): `VoteExtensionsEnableHeight`. 
+There is a newly introduced [**consensus parameter**](./abci%2B%2B_app_requirements.md#abciparamsvoteextensionsenableheight): `VoteExtensionsEnableHeight`. 
 This parameter represents the height at which vote extensions are 
 required for consensus to proceed, with 0 being the default value (no vote extensions).
-
+A chain can enable vote extensions either:
+* at genesis by setting `VoteExtensionsEnableHeight` to be equal to the `InitialHeight`
```
sergio-mena, line 1058:
```suggestion
* at genesis by setting `VoteExtensionsEnableHeight` to be equal, e.g., to the `InitialHeight`
```
I'm suggesting this to tackle @adizere's concern, I think he's got a point (shouldn't imply that, if you set it at genesis then you MUST set it to _initialHeight_)

PR #488, opened 2023-03-08
Compare `types.proto` and `feature/abci++vef` and tip of `v0.36.x`

sergio-mena:
Contributes to #10

Only minor renaming needed

---

#### PR checklist

- [ ] Tests written/updated


proto/tendermint/abci/types.proto
```diff
@@ -352,8 +352,8 @@ message ResponseFinalizeBlock {
   repeated ValidatorUpdate validator_updates = 3 [(gogoproto.nullable) = false];
   // updates to the consensus params, if any.
   tendermint.types.ConsensusParams consensus_param_updates = 4;
-  // agreed_app_data is the bytes that all nodes reach consensus upon in the following height. It is often the hash of the applications' state which can be used to confirm that execution of the transactions was deterministic.
-  bytes agreed_app_data = 5;
+  // app_hash is the hash of the applications' state which is used to confirm that execution of the transactions was deterministic. It is up to the application to decide which algorithm to use.
+  bytes app_hash = 5;
```
jmalicevic, line 356: As we renamed this parameter, should it keep the same sequence number of should we give it a 6?

sergio-mena: The parameter is new in this branch (doesn't exist out of it)

lasarojc (approved):

PR #489, opened 2023-03-08
Pass signatures of vote extensions in `RequestPrepareProposal`

### LINKED ISSUE #485: A proposer needs to forward vote extension information to be verified in `ProcessProposal`
We need to modify the ABCI interface (`PrepareProposal`) to be able to deliver extra information the the App can use to prove to other nodes that it behaved correctly when preparing the proposal.

- [x] #489 
- [x] #509 
- [x] #540

sergio-mena:
Contributes to #485

Contents of this PR:

* This PR introduces the changes in the ABCI protobufs to pass the vote extension signature together with the extensions in `PrepareProposal`
* It demonstrates how to use this data at the application level
  * In the unit tests, with a small example
  * In the e2e app:
    * The e2e app can now find all the data needed to validate the signature
      * some extra functionality was introduced to keep track of the chain ID and the validators' public keys.
    * The e2e app adds all that data in a special transaction at `PrepareProposal` so that it is also available at `ProcessProposal`
    * The e2e app uses the data to validate the signature:
      * In `PrepareProposal` which panics if something is unexpected, as the data comes from the local CometBFT, which should have validated everything itself
      * In `ProcessProposal` which, rather than panicking, returns _reject_ as the data cannot have come from a correct node (which, in this example, would have panicked while doing the same verification)

Note: The checks on `PrepareProposal` are done here because e2e app is for testing purposes, and also to demonstrate how to do it. However, real applications will probably be OK with skipping verification at `PrepareProposal` and thus trusting the local CometBFT. In that case `PrepareProposal` would just be forwarding the information in TXs to `ProcessProposal` (similarly to how it is done in this PR), and having `ProcessProposal` verify the extensions and signatures.

---

#### PR checklist

- [x] Tests written/updated


test/e2e/app/app.go
```diff
@@ -32,12 +35,14 @@ const (
...
-	cfg             *Config
-	restoreSnapshot *abci.Snapshot
-	restoreChunks   [][]byte
+	logger            log.Logger
+	state             *State
+	snapshots         *SnapshotStore
+	cfg               *Config
+	restoreSnapshot   *abci.Snapshot
+	restoreChunks     [][]byte
+	chanID            string
```
thanethomson, line 44: `chainID`?

test/e2e/app/app.go
```diff
@@ -308,12 +315,39 @@ func (app *Application) PrepareProposal(
...
+		//... and verify
+		valAddr := crypto.Address(vote.Validator.Address)
+		pub, ok := app.currentValidators[valAddr.String()]
+		if !ok {
+			panic("PrepareProposal: received vote from unknown validator")
+		}
+
+		if !pub.VerifySignature(extSignBytes, vote.ExtensionSignature) {
+			panic("PrepareProposal: received vote with invalid signature")
+		}
```
thanethomson, line 344: What would the recommended behaviour be for correct applications here? I assume it's not to panic, right? (Just asking preemptively for app developers' sake - I can see how this is useful for picking up problems in our E2E tests)

thanethomson (commented):
One minor nit, but otherwise seems good to me!

I'll approve once the SDK team's happy that the solution works for them.

test/e2e/app/app.go
```diff
@@ -17,7 +17,10 @@ import (
 	"github.com/cometbft/cometbft/abci/example/kvstore"
 	abci "github.com/cometbft/cometbft/abci/types"
 	"github.com/cometbft/cometbft/crypto"
+	cryptoenc "github.com/cometbft/cometbft/crypto/encoding"
 	"github.com/cometbft/cometbft/libs/log"
+	"github.com/cometbft/cometbft/libs/protoio"
+	cmtproto "github.com/cometbft/cometbft/proto/tendermint/types"
```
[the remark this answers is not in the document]
thanethomson: Doesn't really matter very much right now, I think.

With the reorg that @mzabaluev will be doing, this will end up changing anyways and will be more consistent across all code that imports our generated Protobuf code.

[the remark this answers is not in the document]
sergio-mena: Actually, I just did a quick search on the SDK repo and it's used in production code, even some protobufs import them (e.g., _query_, _block_, and _evidence_ proto definitions).
So we should be good on e2e app.

test/e2e/app/app.go
sergio-mena: Good catch!

test/e2e/app/app.go
sergio-mena: So, my understanding is that the signature we're verifying is local (it comes from local CometBFT, which is supposed to have verified it itself when it received the info). A real app could even skip this verification in `PrepareProposal` under the argument it trusts CometBFT has verified it.

The missing piece of this PR (hence it's still a draft) is shipping all necessary info in special transactions so that `ProcessProposal` can also verify (which is where a real app would verify). In that case it wouldn't be a panic, but a `ResponseProcessProposal` with `reject`

state/execution.go
```diff
@@ -487,12 +487,14 @@ func buildExtendedCommitInfo(ec *types.ExtendedCommit, store Store, initialHeigh
 				panic(fmt.Errorf("commit at height %d received with missing vote extensions data", ec.Height))
 			}
 			ext = ecs.Extension
+			extSig = ecs.ExtensionSignature
```
lasarojc, line 490: in line 434, above, Tendermint -> CometBFT

alexanderbez (approved):
LGTM :)

Thanks for also providing an example in the tests for how to perform verification app-side 👍

test/e2e/app/app.go
```diff
@@ -296,66 +304,74 @@ func (app *Application) ApplySnapshotChunk(_ context.Context, req *abci.RequestA
...
 	if extCount > 0 {
 		var totalBytes int64
-		extTxPrefix := fmt.Sprintf("%s=", voteExtensionKey)
-		extTx := []byte(fmt.Sprintf("%s%d", extTxPrefix, sum))
-		app.logger.Info("preparing proposal with custom transaction from vote extensions", "tx", extTx)
-		// Our generated transaction takes precedence over any supplied
-		// transaction that attempts to modify the "extensionSum" value.
-		txs := make([][]byte, len(req.Txs)+1)
-		for i, tx := range req.Txs {
+		extCommitBytes, err := req.LocalLastCommit.Marshal()
```
lasarojc, line 331: As this serves as an example app, this whole loop could be applied irrespective of `extCount>0` to guard against transactions starting with `extTxPrefix` and `reservedKey`.
Then build and append the special transaction only if `extCount>0`.

test/e2e/app/app.go
```diff
@@ -376,12 +392,15 @@ func (app *Application) ProcessProposal(_ context.Context, req *abci.RequestProc
...
 		}
 		// Additional check for vote extension-related txs
-		if k == voteExtensionKey {
-			_, err := strconv.Atoi(v)
-			if err != nil {
-				app.logger.Error("malformed vote extension transaction", k, v, "err", err)
+		switch {
+		case k == voteExtensionKey:
+			if err := app.verifyExtensionTx(req.Height, v); err != nil {
+				app.logger.Error("vote extension transaction failed verification, rejecting proposal", k, v, "err", err)
```
lasarojc, line 398: Add a comment regarding using this proposal as evidence of misbehavior by the proposer?

test/e2e/app/app.go
```diff
@@ -17,14 +18,20 @@ import (
...
+	cryptoproto "github.com/cometbft/cometbft/proto/tendermint/crypto"
+	cmtproto "github.com/cometbft/cometbft/proto/tendermint/types"
 	"github.com/cometbft/cometbft/version"
 )
 
 const (
 	appVersion                 = 1
 	voteExtensionKey    string = "extensionSum"
 	voteExtensionMaxVal int64  = 128
+	reservedKey         string = "reservedTxKey"
```
lasarojc, line 33: Eventhough addresses do not contain `y`, it might be a good idea to add a separator to the end of the key.
```suggestion
	reservedKey         string = "reservedTxKey"
```

proto/tendermint/abci/types.proto
```diff
@@ -444,6 +444,8 @@ message ExtendedVoteInfo {
   bool signed_last_block = 2;
   // Non-deterministic extension provided by the sending validator's application.
   bytes vote_extension = 3;
```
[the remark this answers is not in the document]
sergio-mena: I prefer to err on the safe side (long but clear name), unless someone has a strong opinion on this

proto/tendermint/abci/types.proto
```diff
@@ -444,6 +444,8 @@ message ExtendedVoteInfo {
...
+  // Vote extension signature created by CometBFT
+  bytes extension_signature = 4;
```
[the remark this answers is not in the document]
sergio-mena: Not possible, we shouldn't confuse it with the signature of the vote itself

test/e2e/app/app.go
sergio-mena: I like your suggestion, let me see what I can do.
Also, for the future, we can strengthen the logic of this function, because we do know when vote extensions are active and when they are not. The way it is now is a bit loose, and, since this is a testing app we should probable make if panic if we receive extensions when we shouldn't (or if they are missing when they shouldn't)

test/e2e/app/app.go
sergio-mena: Yeah, good idea

test/e2e/app/app.go
alexanderbez: In the case of the app, which verifies vote exts "injected" by the proposer, if any such verification fails, rather if more than 2/3 voting power fails, then it would REJECT the proposal.

lasarojc: I didn't get this last comment. The proposal contains extensions from 2/3 of the voting power from the previous round, out of which 1/3+1 is guaranteed to be from correct validators. What do you mean by "if more than 2/3 voting power fails"?
Maybe I am not understanding how you plan on using it. Is there anything you could share?

state/execution.go
sergio-mena: Thanks for spotting. I'm changing this in an upcoming related PR, which I am doing separately to avoid bloating this one.

test/e2e/app/app.go
sergio-mena: Also, comming in the next PR

test/e2e/app/app.go
sergio-mena: Actually, I decided to go for the "tightening up" the checks we do on vote extensions. Everything is pretty optional in this PR. I have an upcoming PR lined up which does that, and simplifies this loop massively. Please take a look when it's up

test/e2e/app/app.go
sergio-mena: Added a NOTE

test/e2e/app/app.go
alexanderbez: > The proposal contains extensions from 2/3 of the voting power from the previous round, out of which 1/3+1 is guaranteed to be from correct validators.

Yes, given to the proposer during `PrepareProposal`. The remaining validators that receive the manually injected vote extensions during `ProcessProposal` have to ensure they get at least 2/3 from the proposer.

state/execution.go
sergio-mena: #509

test/e2e/app/app.go
sergio-mena: #509

cason:
This PR is missing a changelog entry. And it is a breaking change.

cason:
Ref #4458.

Issue #370, opened 2023-02-22
Render the consensus `needProofBlock` mehtod more robust (WIP)


cason:
The `needProofBlock` method from [`consensus/state.go`](https://github.com/cometbft/cometbft/blob/45a08f3f90a170bd862fa5615aeb9bba12e7f55b/consensus/state.go#L1050) informs whether the execution of the last committed block has caused the application hash to change. If the application state (hash) has changed, a new block is needed as it serves as a proof of the new state of the application, which is the reason for the name of the method.

The information provided by the `needProofBlock` is relevant when nodes are starting the first round (`round == 0`) of a height. If the application state has changed from the execution of the last block, which can be considered the common case, the node should immediately proceed to the `propose` round step to propose a block or wait for a proposed block. The alternative flow at this point is to wait for transactions being available in the mempool before entering the propose step, as a way to prevent the proposal of empty blocks (with no transactions). This alternative flow is not taken by default, it has to be configured (see this [line](https://github.com/cometbft/cometbft/blob/45a08f3f90a170bd862fa5615aeb9bba12e7f55b/consensus/state.go#L1037)),  and it can only be taken when the `needProofBlock` method returns `false` . The method returns false when the application state (hash) has not changed with the previous block, for instance because the previous block carried no transactions.

Given this analysis and providing an initial conclusion, it is **safe** for this method to return `true` when the information it should provide cannot be retrieved.

TODO: describe the example scenario in which the information needed by the method is not available.

adizere:
For reference, the immediate solution was this:

```diff
diff --git a/consensus/state.go b/consensus/state.go
index 6d2fab2bd..b8264534a 100644
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -1050,7 +1050,7 @@ func (cs *State) needProofBlock(height int64) bool {

        lastBlockMeta := cs.blockStore.LoadBlockMeta(height - 1)
        if lastBlockMeta == nil {
-               panic(fmt.Sprintf("needProofBlock: last block meta for height %d not found", height-1))
+               return true
        }

        return !bytes.Equal(cs.state.AppHash, lastBlockMeta.Header.AppHash)
```

https://github.com/informalsystems/tendermint/commit/ffbf7ececac58c59ef49ab4f59e1beed7e2b37a0

TobiaszCudnik:
Adding a bit more context regarding "proof blocks".

[...]

PR #491, opened 2023-03-09
Added `abci.RequestBeginBlock.IsProofBlock` for `create_empty_blocks=false`

TobiaszCudnik:
Ported from https://github.com/tendermint/tendermint/pull/10004
Closes: [cosmos-sdk#10240](https://github.com/cosmos/cosmos-sdk/issues/10240)
Related to #370

This PR extends `abci.RequestBeginBlock` with `IsProofBlock`, used when `create_empty_blocks=false`.

Main changes in:
- `consensus/state.go`
- `proto/tendermint/abci/types.proto`

The `IsProofBlock` flag is used by modules in `BeginBlock` methods to avoid write ops for those blocks. "Proof blocks" happen when:
- no txs in the mempool
- current apphash differs form the prev block's apphash
- empty blocks should NOT be created

"Proof block" is currently defined in `tendermint/consensus/state.go:1041`:
https://github.com/cometbft/cometbft/blob/24040f5a010728cab301011b4c8eb489e151b897/consensus/state.go#L1050-L1060
and used for `create_empty_blocks` in `enterNewRound`:
https://github.com/cometbft/cometbft/blob/24040f5a010728cab301011b4c8eb489e151b897/consensus/state.go#L1037-L1047

### Related links
- [PR for Cosmos SDK](https://github.com/cosmos/cosmos-sdk/pull/15148)
- [Ticket cosmos-sdk#10240](https://github.com/cosmos/cosmos-sdk/issues/10240)
- [Ticket cosmos-sdk#12356](https://github.com/cosmos/cosmos-sdk/issues/12356)

---

#### PR checklist

- [ ] Tests written/updated


adizere:
Thanks for the PR @TobiaszCudnik !

There's a couple of concerns here:

### Nature of change

The PR changes `proto/tendermint/abci/types.proto`, so it will be a Go API/ABCI change. To my understanding, we can treat this change as forward compatible due to protobuf support, i.e., nodes running current version (which does not have the new field [`is_proof_block`](https://github.com/cometbft/cometbft/pull/491/files#diff-8384a26745c978d00d66edad2513d2c0518fab0ddb4123c613f475b88ed95667R79)) should be able to parse responses from nodes running the version with your proposed change (and simply ignore the new field, without breaking). Let me know if you think this is in/correct as it will help us assess the impact of the change.

### Relation to SDK 

This change is directly dependent on impact and priority in the SDK. So we'll take their feedback into account, [in particular this coment](https://github.com/cosmos/cosmos-sdk/issues/10240#issuecomment-1446460037). Thanks @tac0turtle or team for weighing in here (not urgent!)

### Priority

We're working at full capacity to release v0.38 (alpha) to unblock multiple teams in the ecosystem. Give us a few weeks to sort that out. Then we'll get back to you to schedule including this potentially in v0.38 or v0.39. Thanks for the patience! If this blocks anything on your side, don't hesitate to let us know.

TobiaszCudnik:
> nodes running current version (which does not have the new field [is_proof_block](https://github.com/cometbft/cometbft/pull/491/files#diff-8384a26745c978d00d66edad2513d2c0518fab0ddb4123c613f475b88ed95667R79)) should be able to parse responses from nodes running the version with your proposed change

That is correct, only the known IDs will be parsed. Itd break if someone adds another field with the same proto ID (`5`), eg from another consensus implementation. Although if some validators wont update, they'll keep producing empty blocks (AFAIK).

> Give us a few weeks to sort that out

Take your time, Im still figuring out the best way to test this. It would probably need a regression test for Cosmos and a regular one for Comet.

PR #496, opened 2023-03-10
ABCI calls should crash/exit in all cases when the call itself reports and error

### LINKED ISSUE #490: ABCI: Crash in all cases of undefined app behavior
Was tendermint/tendermint#8206

During the spec/implementation of ABCI 2.0 in `v0.36.x`, [it was decided](tendermint/tendermint#8124) that CometBFT should _crash_ if it received an error from the application from one of the ABCI methods.

This behavior was decided as correct since an application responding with invalid data or that has become unresponsive is no longer in a consistent state. Additionally, the Tendermint algorithm can no longer make progress safely.

We should update the rest of the consensus code to crash when ABCI responds with an error or produces invalid data.

sergio-mena:
Closes #490

Followed the call hierarchy of ABCI methods to make sure that reporting an error ultimately results in a crash/exit (also followed new methods in branch `feature/abci++vef`).

Found one blind spot that should be considered a bug, and should probably be backported. Hence, I'm opening this against `main` and not `feature/abci++vef`

---

#### PR checklist

- [ ] Tests written/updated


jmalicevic (approved):
I assume we will fix the vulnerability failures.

PR #497, opened 2023-03-10
Start of a Draft of p2p quint spec following Daniel's English spec

josef-widder:
This is for discussion on the p2p spec work.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/spec/p2p/p2p.qnt
+++ b/spec/p2p/p2p.qnt
@@ -1,0 +1,140 @@
+module p2p {
+
+    pure val peerStates = Set("dialing", "inbound", "reconnecting", "outbound", "unconditional")
+    // a work around as constants are not working in the REPL
+
+    pure val connectionStates = Set("init", "open", "closed")
+
+    type Address = {ID:str, ipAddress:str}
+ //   type State = {peerState:str, connectionState: str}
+
+    //type Peer = {address:Address, state:State}
+    //var Peers : Set[Peer]
+
+    // p2p variables 
+    var peerState: Address -> str
+    var connectionState: Address -> str
+    var IsUnconditional: Set[Address]
+
+    // PEX variable
+    var AddressBook: Set[Address]
+
+    // Environment. These are the nodes that may dial us
+    var Environment: Set[Address]
+ 
+
+    pure val MaxNumOutboundPeers = 5
+    pure val MaxNumInboundPeers = 10
+
+    def Peers = peerState.keys()
+
+    def peersOutbound = size(Peers.filter(a => peerState.get(a) == "outbound"))
+    def peersInbound = size(Peers.filter(a => peerState.get(a) == "inbound"))
+    def peersDialing = size(Peers.filter(a => peerState.get(a) == "dialing"))
+
+    def require(pred:bool):bool = pred
+    //pure def putset (m,s)
+
+    // triggered by "the network"
+    action loseConnection(a:Address):bool = all{
+        require(connectionState.get(a) != "closed"),
+        connectionState' = connectionState.set(a, "closed"),
+        all { //unchanged
+            AddressBook' = AddressBook,
+            peerState' = peerState,
+            Environment' = Environment,
+            IsUnconditional' = IsUnconditional,
+        }
+    }
+
+    action transportAccept(a: Address):bool = all {
+        or {
+            peersInbound < MaxNumInboundPeers,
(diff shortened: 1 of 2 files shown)


spec/p2p/p2p.qnt
```diff
@@ -0,0 +1,140 @@
+module p2p {
+
+    pure val peerStates = Set("dialing", "inbound", "reconnecting", "outbound", "unconditional")
+    // a work around as constants are not working in the REPL
+
+    pure val connectionStates = Set("init", "open", "closed")
+
+    type Address = {ID:str, ipAddress:str}
```
cason, line 8:
```suggestion
    type Address = {ID:str, address:str}
```

Address is a network address: IP:port

spec/p2p/p2p.qnt
```diff
@@ -0,0 +1,140 @@
...
+            IsUnconditional' = IsUnconditional,
+        }
+    }
+
+    action transportAccept(a: Address):bool = all {
+        or {
+            peersInbound < MaxNumInboundPeers,
+            IsUnconditional.contains(a)
+        },
+        peerState' = peerState.put(a, "inbound"),  //shall we check if it is already outbound and open?
```
cason, line 55: Yes, we should. But it can be put on the conditions, right? Peers is not connected.

spec/p2p/p2p.qnt
```diff
@@ -0,0 +1,140 @@
...
+    }
+        // inform all reactors then
+    
+
+    action ensurePeers(a:Address):bool =
+        all {
+                require(peersOutbound + peersDialing < MaxNumOutboundPeers),
+                require(not(Peers.contains(a))),
+                all {
+                    // I guess we don't remove AddressBook' = AddressBook.exclude(Set(a)),
```
cason, line 82: Yes, we don't change the Address book.

spec/p2p/p2p.qnt
```diff
@@ -0,0 +1,140 @@
...
+                        Environment' = Environment,
+                        IsUnconditional' = IsUnconditional,
+                    }
+                }
+            }
+
+    action dialPeers(S:Set[Address]):bool = // check what the code actually is doing. Now we don't override existing peers in this spec
+        all {
+            peerState' = Peers.union(S).mapBy(a => if (Peers.contains(a)) peerState.get(a) else "dialing"), 
+            connectionState' = Peers.union(S).mapBy(a => if (Peers.contains(a)) connectionState.get(a) else "init"),
```
cason, line 97: We should have `dialPeer` also here. I am assuming that this method either set a peer to the "outbound" state or produce an error, right?

cason: Since this method is invoked from "outside", the addresses are also added to the Address Book.

spec/p2p/p2p.qnt
```diff
@@ -0,0 +1,140 @@
...
+                        IsUnconditional' = IsUnconditional,
+                    }
+        }
+
+    action init =
+        all {
+            peerState' = Map(),
+            connectionState' = Map(),
+            AddressBook' = Set({ID: "igor", ipAddress: "127.0.0.1"}, {ID: "josef", ipAddress: "127.0.0.2"}),
+            Environment' = Set({ID: "lasaro", ipAddress: "127.0.0.7"}, {ID: "daniel", ipAddress: "127.0.0.6"}),
```
cason, line 110: For testing purposes, it will be good to have the same address as incoming and as candidate to be dialed.

spec/p2p/p2p.qnt
```diff
@@ -0,0 +1,140 @@
...
+            IsUnconditional' = Set({ID: "zarko", ipAddress: "127.0.0.42"}),
+        }
+
+    action step = any {
+        all {
```
cason, line 115: The `ensurePeers` semantics provides it should choose an address at random to dial. Here this step is done outside this method, right?

josef-widder: Yes. This is for debugging purposes. Like this I can tell the REPL on the command line which peer to chose. 
I guess it makes sense to have another definition that pulls the choice in for defining the steps of the system...

spec/p2p/p2p.qnt
josef-widder: I have added an alternative init action `action initWithSameSets`

PR #509, opened 2023-03-13
Tighten the checks we do on vote extensions in e2e app

sergio-mena:
Contributes to #485

The current checks on vote extensions in the e2e app take vote extensions as something always optional. This is OK, but we can leverage the app to more thoroughly test several cases:
* upgrade path: make sure extensions are present _iff_ they are active
* test the subtleties of the upgrade height: `PrepareProposal`/`ProcessProposal` don't have extensions, `ExtendVote`/`VerifyVoteExtension` do.
* Decide what to do with `Nil` votes in a commit

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/test/e2e/app/app.go
+++ b/test/e2e/app/app.go
@@ -30,10 +30,10 @@ const (
 	appVersion                 = 1
 	voteExtensionKey    string = "extensionSum"
 	voteExtensionMaxVal int64  = 128
-	reservedKey         string = "reservedTxKey"
-	suffixChainID       string = "_ChainID"
-	suffixVoteExtHeight string = "_VoteExtensionsHeight"
-	suffixInitialHeight string = "_InitialHeight"
+	prefixReservedKey   string = "reservedTxKey_"
+	suffixChainID       string = "ChainID"
+	suffixVoteExtHeight string = "VoteExtensionsHeight"
+	suffixInitialHeight string = "InitialHeight"
 )
 
 // Application is an ABCI application for use by end-to-end tests. It is a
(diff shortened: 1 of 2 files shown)


test/e2e/app/app.go
```diff
@@ -452,22 +444,21 @@ func (app *Application) ExtendVote(_ context.Context, req *abci.RequestExtendVot
...
+	appHeight, areExtensionsEnabled := app.checkHeightAndExtensions(false, req.Height, "VerifyVoteExtension")
+	if !areExtensionsEnabled {
+		panic(fmt.Errorf("received call to VerifyVoteExtension at height %d, when vote extensions are disabled", appHeight))
+	}
+	// We don't allow vote extensions to be optional
 	if len(req.VoteExtension) == 0 {
+		app.logger.Error("received empty vote extension")
 		return &abci.ResponseVerifyVoteExtension{
-			Status: abci.ResponseVerifyVoteExtension_ACCEPT,
+			Status: abci.ResponseVerifyVoteExtension_REJECT,
```
jmalicevic, line 455: Is it not allowed to have an empty vote extension?

sergio-mena: Not in the current application's logic. I chose to do it this way (more stringent) to be able to better detect bugs... and it worked ;-)

test/e2e/app/app.go
```diff
@@ -533,22 +566,36 @@ func parseTx(tx []byte) (string, string, error) {
...
+		if !vote.SignedLastBlock {
+			continue
+		}
+		if !areExtensionsEnabled {
+			if len(vote.VoteExtension) != 0 {
+				return 0, fmt.Errorf("non-empty vote extension at height %d, which has extensions disabled", currentHeight)
+			}
+			continue
+		}
+		if len(vote.VoteExtension) == 0 {
```
jmalicevic, line 582: One last question related to the comment above. Here the length is allowed to be 0 (as it is supposed to be), but then it will fail later? I am just asking to confirm that you don't want to impose your more stringent condition here as well.

jmalicevic (approved):
Other than the comment, LGTM.

test/e2e/app/app.go
sergio-mena: That's an excellent question. :-)
The answer is coming in one of my next PRs. The problem is that the current ABCI data structures can only say if the sender of the vote "signed or not" (i.e., a boolean). But if the sending validator has signed for `nil`, `SignedLastBlock` will also be true; however as it is a `nil` precommit it won't have an extension.
It took me a few hours of troubleshooting with e2e tests to find this out.
We discussed this in @ebuchman's office hours on Monday, and we decided we're gonna fix the ABCI interface (a boolean is not enough), since anyway we're introducing breaking changes.
This is what's coming in one of my next PRs.

sergio-mena: #540

PR #534, opened 2023-03-14
docs: improve Execute code block

rootulp:
1. `InitialHeight` is specified twice
2. The contents of `Execute` refer to the parameter `state` not `s`


adizere (approved):
Thanks for the contribution @rootulp, nice catch!

PR #536, opened 2023-03-15
Fixes the order of guides and removes stale guides

lasarojc:
the stale guides were added back on a previous PR.
the order change makes guides appear in the following order
- install
- Quickstart
- go built in app
- go app
- upgrade from TM


thanethomson (approved):

docs/guides/java.md
```diff
@@ -1,6 +1,6 @@
----
+<!---
 order: 6
----
+--->
```
thanethomson, line 3: Oh, forgot to mention, we _require_ frontmatter in our docs. So please don't comment this out - perhaps just remove the order field? Otherwise it'll break the Jekyll build.

Otherwise please just delete the files :slightly_smiling_face:

lasarojc: I think Jekyll only includes files that have the front-matter. If we include it without the order field, it would just order the files last.

lasarojc: I will double check locally.

andynog: I'd say, if we don't need the files anymore it's better to delete than to keep them, even though you could hide from the documentation site, they would still show up in Github and if we're not supporting tutorials in these languages in the short term, it's better to remove them. Probably when we revamp the docs we can add future tutorials on more popular languages (e.g. typescript).

PR #538, opened 2023-03-15
Add metrics to BlockSync

### LINKED ISSUE #543: Generate Prometheus metrics during block sync
### Summary

Currently when the BlockSync reactor is being executed, it only produces one metric called `Syncing` which flags whether BlockSync is catching up or not.

### Problem Definition

Having metrics for the latest block that BlockSync has processed would be useful for 

### Proposal

Add new metrics for:
- the height of the latest block processed by BlockSync, 
- number and size of transactions of the latest block, and
- total number of transactions processed.

The new metrics should be in Prometheus namespace for BlockSync. We should not reuse Consensus' metrics.

Also, when BlockSync finishes and switches to Consensus, the metrics should not be registered anymore, as they would be not relevant.

Note that there used to be an [issue](https://github.com/tendermint/tendermint/issues/3507) for this feature that was closed for `v0.36.x`, but it's missing from `main`. Only the unused method `RecordConsMetrics` was left, which should be removed.

hvanz:
Closes #543 

---

#### PR checklist

- [x] Tests written/updated


blocksync/reactor.go
```diff
@@ -240,7 +240,15 @@ func (bcR *Reactor) Receive(e p2p.Envelope) {
 // NOTE: Don't sleep in the FOR_LOOP or otherwise slow it down!
 func (bcR *Reactor) poolRoutine(stateSynced bool) {
 	bcR.metrics.Syncing.Set(1)
-	defer bcR.metrics.Syncing.Set(0)
+	defer func() {
```
lasarojc, line 243: If we set the metric back to 0 before deregistering will we capture the end of block sync in prometheus? If we want to query for running periods, this is easier than figuring out when metrics stopped being updated.

hvanz: I tried setting the `Syncing` metric to 0 before deregistering but it doesn't work. Prometheus doesn't have time to see the change to 0 before unregistering the metric. 

The advantage of not changing to 0 is that the graph for this metric is cleaner because it doesn't have to display the 0 values. And for querying running periods, there is the option of comparing whether it's equal to 1 or different from `nil`, instead of checking that it's not equal to 0. But if you think it's better to set it to 0, we could do just that; it's not a big difference.

hvanz: For the other metrics, I think it's necessary to unregister them. For example, this is what we would see if we want to display `cometbft_consensus_height` and `cometbft_blocksync_latest_block_height` at the same time. The red line for the height during syncing would get a permanent fixed value even after syncing has finished.
Screenshot 2023-03-17 at 15 27 45

lasarojc: > I tried setting the `Syncing` metric to 0 before deregistering but it doesn't work. Prometheus doesn't have time to see the change to 0 before unregistering the metric.
> 
> The advantage of not changing to 0 is that the graph for this metric is cleaner because it doesn't have to display the 0 values. And for querying running periods, there is the option of comparing whether it's equal to 1 or different from `nil`, instead of checking that it's not equal to 0. But if you think it's better to set it to 0, we could do just that; it's not a big difference.

I see. As long as we can query, it LGTM.

lasarojc (approved):

blocksync/metrics.go
```diff
@@ -16,4 +22,44 @@ const (
...
+func (m *Metrics) recordBlockMetrics(block *types.Block) {
+	m.NumTxs.Set(float64(len(block.Data.Txs)))
+	m.TotalTxs.Add(float64(len(block.Data.Txs)))
+	m.BlockSizeBytes.Set(float64(block.Size()))
+	m.LatestBlockHeight.Set(float64(block.Height))
+}
+
+// unregisterGauge removes a gauge metric from the Prometheus registry. No new
+// data will be generated in the http /metrics endpoint for this metric.
+func unregisterGauge(g metrics.Gauge) bool {
```
thanethomson, line 44: I discussed this with @hvanz on Slack - I really don't think we should be unregistering Prometheus metrics. It seems risky, especially if external metrics systems come to rely on specific metrics being present. Plus it's really non-standard.

If we do really want to unregister metrics, there needs to be a _very_ good reason.

sergio-mena: We discussed it synchronously. The best solution will be to unify certain metrics, such as the height, into a node-wide set of metrics. But that will need a bit more work on infra.
I agree we probably want to not unregister upon catching up, even if the resulting graph looks somewhat awkward.

hvanz: Thanks for the discussions! As a result, we won't unregister the metrics, and we'll have to think about a better solution for metrics at the node level.

PR #539, opened 2023-03-15
Fix bugs: (1) infinite loop in e2e tests (2) missing Vote Extensions at `PrepareProposal` time

sergio-mena:
Contributes to #10

While running e2e tests locally, I detected a hard-to-diagnose bug with vote extensions. This PR contains the fix.

Since I did this in isolation (due to remote work), I'll make an effort at explaining the troubleshooting process in this PR. Hence, this PR will be a succession of commits which show the sketch on how I diagnosed and fixed the bug. So, if the reviewer wants to understand what the problem was and how it was solved, the presence of the little ❌ or ✅ next to each commit below is important (you can click and check how the e2e tests failed/passed).

Additionally, I also found that the e2e tests where blocking forever when a node was crashing. This is fixed in the first commit.

* Commit#1: fix the infinite loop in e2e runner that may block the execution indefinitely.
* Commit#2: adapt `ci.toml` to only have one validator, since the bug appears when a recovered node has to propose.
* Commit#3: under the hypothesis that something is wrong with `VoteExtensionsEnableHeight`, we plant a `panic` to have the backtrace if it happens.
  * The enable height in `ci.toml` is set to 1007 from genesis, so we should never return zero in this test case.
* Commit#4: contains the fix. Every time we call `reconstructLastCommit` method of `State`, immediately after we call `updateToState`. Therefore the state object _whose method_ we are calling is incomplete. The fix just queries the _complete_ object, which is passed as a normal parameter (and which will be given to `updateToState` immediately after).
  * Note that, although some checks are failing on that commit (hence the ❌), e2e are now passing
* Commit#5: remove diagnosis-related code, i.e., commit#2 and commit#3

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/test/e2e/runner/rpc.go
+++ b/test/e2e/runner/rpc.go
@@ -94,14 +94,14 @@ func waitForNode(ctx context.Context, node *e2e.Node, height int64, timeout time
 		case <-timer.C:
 			status, err := client.Status(ctx)
 			switch {
+			case time.Since(lastChanged) > timeout:
+				return nil, fmt.Errorf("timed out waiting for %v to reach height %v", node.Name, height)
 			case err != nil:
 			case status.SyncInfo.LatestBlockHeight >= height && (height == 0 || !status.SyncInfo.CatchingUp):
 				return status, nil
 			case curHeight < status.SyncInfo.LatestBlockHeight:
 				curHeight = status.SyncInfo.LatestBlockHeight
 				lastChanged = time.Now()
-			case time.Since(lastChanged) > timeout:
-				return nil, fmt.Errorf("timed out waiting for %v to reach height %v", node.Name, height)
 			}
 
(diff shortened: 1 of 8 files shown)


thanethomson (approved):
Great catch!

test/e2e/runner/rpc.go
cason, line 1: No changes in this file, after all? I don't think changing the order of events in a select block has any practical effect.

sergio-mena: It removes an infinite loop ;-)

cason: I read the code wrongly, sorry.

sergio-mena: NP

cason (approved):

PR #540, opened 2023-03-16
Fix deficiency in ABCI interface: `SignedLastBlock`

sergio-mena:
Contributes to #485

Current ABCI 2.0 implementation has `VoteInfo` and `ExtendedVoteInfo` to send vote signature information to the application at several points in the interaction:
* `PrepareProposal`
* `ProcessProposal`
* `FinalizeBlock`

The problem with `VoteInfo` and `ExtendedVoteInfo` is that they contain a `bool`, called `SignedLastBlock`, which collapses several cases: (a) did the sending validator precommit for `nil`? (b) did it precommit for the decided block? (c) did they not precommit (or no precommit received, absent)?

`SignedLastBlock` is collapsing cases (a) and (b) into value `true` of the boolean, hiding potentially useful information to the application.

With vote extensions, this has become a bigger problem, as we are unable to "ensure" from a testing application that the vote extensions are present or absent at the right time. If we (CometBFT team) are unable to test this properly, then real apps may have a very hard time troubleshooting potential problems on the field, due to some info being _trapped_ south of ABCI.

After a discussion with @ebuchman, we have decided to fix this problem, which was already fixed long ago in the block-related protos (see [discussion](https://github.com/tendermint/tendermint/issues/3485)). The conclusion was that this is the right moment, as we are heavily changing ABCI in this release, and it's not certain when we will need to make substantial changes to it in the mid-term.

This PR contains the changes needed in the protobufs and in the code.

---

#### PR checklist

- [ ] Tests written/updated


test/e2e/app/app.go
```diff
@@ -568,30 +568,48 @@ func parseTx(tx []byte) (string, string, error) {
...
-		if !vote.SignedLastBlock {
-			continue
+		if vote.BlockIdFlag == cmtproto.BlockIDFlagUnknown || vote.BlockIdFlag > cmtproto.BlockIDFlagNil {
+			return 0, fmt.Errorf("vote with bad blockID flag value at height %d; blockID flag %d", currentHeight, vote.BlockIdFlag)
 		}
-		if !areExtensionsEnabled {
+		if vote.BlockIdFlag == cmtproto.BlockIDFlagAbsent || vote.BlockIdFlag == cmtproto.BlockIDFlagNil {
 			if len(vote.VoteExtension) != 0 {
-				return 0, fmt.Errorf("non-empty vote extension at height %d, which has extensions disabled", currentHeight)
+				return 0, fmt.Errorf("non-empty vote extension at height %d, for a vote with blockID  flag %d",
```
jmalicevic, line 585: Essentially, if this validator had voted correctly, and there is an extension, the flag had to have been Commit?

jmalicevic (approved):
Minor comments, otherwise good.

proto/tendermint/abci/types.proto
```diff
@@ -434,18 +435,22 @@ message ValidatorUpdate {
...
 message ExtendedVoteInfo {
   // The validator that sent the vote.
   Validator validator = 1 [(gogoproto.nullable) = false];
-  // Indicates whether the validator signed the last block, allowing for rewards based on validator availability.
-  bool signed_last_block = 2;
   // Non-deterministic extension provided by the sending validator's application.
   bytes vote_extension = 3;
   // Vote extension signature created by CometBFT
   bytes extension_signature = 4;
+  // block_id_flag indicates which BlockID the signature is for
```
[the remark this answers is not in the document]
sergio-mena: You're right. That was me abusing copy-paste

test/e2e/app/app.go
sergio-mena: Exactly:
* BlockIDFlagAbsent means it didn't precommit so, how come we got a vote extension?
* BlockIDFlagNil mean it precommitted for `nil`, but `nil` precommits don't have vote extensions

In both cases the info we get is contradictory, so we should reject it.

PR #562, opened 2023-03-20
Fixed two bugs related to metrics

sergio-mena:
Contributes to #10

When testing in prometheus-enabled testnets, we caught two places where the metrics gathering were causing a panic.
This PR also includes adjustments to e2e `ci.toml` config that is able to trigger these bugs.
Finally (not important, but for the record), I have included the time window I used to extract metrics from the 200-node test.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/test/e2e/networks/ci.toml
+++ b/test/e2e/networks/ci.toml
@@ -11,6 +11,7 @@ process_proposal_delay = "100ms"
 check_tx_delay = "0ms"
 # The most common case (e.g. Cosmos SDK-based chains).
 abci_protocol = "builtin"
+prometheus = true
 
 [validators]
 validator01 = 100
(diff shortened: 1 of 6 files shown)


adizere (approved):

test/e2e/networks/ci.toml
```diff
@@ -11,6 +11,7 @@ process_proposal_delay = "100ms"
 check_tx_delay = "0ms"
 # The most common case (e.g. Cosmos SDK-based chains).
 abci_protocol = "builtin"
+prometheus = true
```
jmalicevic, line 14: So we are running prometheus enabled nodes in our CI as well? Is that going to cause too much noise?

sergio-mena: Prometheus is enable from now on, but no server set up. Although it might seem a bit silly, it would have caught the two panics I saw when I started the DO testnet.

jmalicevic (approved):

PR #568, opened 2023-03-23
Merge `main` into `feature/abci++vef` (preparation merge)

sergio-mena:
This merge is in preparation for the final merge, which will be in the other direction: `feature/abci++vef` --> `main`

Performing two merges, rather than one merge in one shot, allows us resolve all existing conflicts now, do some testing, etc, without blocking commits to `main`: to avoid a _dirty_ `merge` from `feature/abci++vef` to `main`, no commit should land on `main` during that merge operation. In contrast, this one only blocks commits to `feature/abci++vef` which is not a problem since all those commits have already landed.

Kudos to @jmalicevic for solving some of the unit test failures.

---

#### PR checklist

- [x] Tests written/updated


abci/example/kvstore/kvstore.go
```diff
@@ -45,6 +45,10 @@ type Application struct {
 	// validator set
 	valUpdates         []types.ValidatorUpdate
 	valAddrToPubKeyMap map[string]cryptoproto.PublicKey
+
```
sergio-mena, line 48: @jmalicevic please review the detail of this file. All changes here were conflicts I have to solve manually. Please make sure I didn't make any mistake.

rpc/client/main_test.go
```diff
@@ -19,6 +19,8 @@ func TestMain(m *testing.M) {
 	}
 
 	app := kvstore.NewPersistentApplication(dir)
+	// If testing block event generation
+	// app.SetGenBlockEvents() needs to be called here
```
sergio-mena, line 23: @jmalicevic please review this one as well (manual conflict resolution), is this well-inserted here?

abci/example/kvstore/kvstore.go
jmalicevic: Reviewed and updated.. All good!

rpc/client/main_test.go
jmalicevic: Yes this is the place to have this!

jmalicevic (approved):
My review focused mostly on indexer related things and I am not finding anything that is wrong. I locally did a diff between this branch and main, and don't find anything out of the ordinary.

state/indexer/block/kv/kv.go
```diff
@@ -49,7 +54,7 @@ func (idx *BlockerIndexer) Has(height int64) (bool, error) {
 // The following is indexed:
 //
 // primary key: encode(block.height | height) => encode(height)
-// FinalizeBlock events: encode(eventType.eventAttr|eventValue|height|finalize_block) => encode(height)
+// FinalizeBlock events: encode(eventType.eventAttr|eventValue|height|finalize_block|eventSeq) => encode(height)
```
jmalicevic, line 57: In line 74 below, we now do not distinguish the different types of events and store in our database always the same string `finalize_block`. This is taking up space and is completely useless. I wonder whether we should, at some point open a PR to alter the indexer to remove this? (It is trivial to remove but we would need to account that previous versions of the indexer stored this `typ` field so we'd need to parse it)

PR #572, opened 2023-03-23
Final merge of `feature/abci++vef` into `main`

sergio-mena:
Contributes to #10 

This is the final merge of feature branch `feature/abci++vef`. After this merge, the feature branch will cease to exist technically.

We will need to make some adjustment to `UPGRADING.md` after the merge with extra info on how `FinalizeBlock` events are indexed, but that shouldn't block this merge.

IMPORTANT: Please do not review the detail of this merge, as this code has already been reviewed on `feature/abci++vef`.

---

#### PR checklist

- [x] Tests written/updated


docs/app-dev/indexing-transactions.md
```diff
@@ -105,20 +105,20 @@ Type: "transfer",
    },
 ```
 
-will be represented as follows in the store, assuming these events result from the EndBlock call for height 1: 
+will be represented as follows in the store, assuming these events result from the `finalize_block` call for height 1:
```
jmalicevic, line 108:
```suggestion
will be represented as follows in the store, assuming these events result from the `FinalizeBlock` call for height 1: 
```

jmalicevic (approved):
Looks good, amazing work!

sergio-mena:
4790ea3e4 represents the original merge commit + all subsequent commits of this PR squashed onto the merge commit

Issue #576, opened 2023-03-23
p2p/e2e: Validators not connecting to each other


thanethomson:
A follow-up to #575.

It appears as though, in our E2E tests on `v0.34.x`, when using the `test/e2e/networks/ci.toml` manifest, at least one of the validators (e.g. `validator02`) fails to connect to the other validators on the same network.

The full output log of such a run is available here: https://gist.github.com/thanethomson/3b85c8fa8ce3a731d748e52d5bdd0105

This was not reproducible on `v0.37.x` or `main`. It could be architecture- or platform-specific, as we have not yet been able to replicate this on macOS, and has only manifested on Ubuntu so far using Docker.

thanethomson:
After applying the patch in #577 to `v0.34.x`, when encountering this issue, the E2E tests will hang indefinitely.

cason:
So, from the logs, `validator02` connects to `seed02` at startup. It request peer addresses to the seed node, which does not report any new address (likely reports its own address), then disconnects from `validator02`, which is standard for a seed node. Since `seed02` is the only node `validator02` knows, it attempts connecting to it again. Since `validator02` is the only node `seed02` knows, it attempts connecting to it again, and succeeds.

`seed02` fails connecting to `seed01`, which the configured seed node, and vice-versa: `seed01` fails connecting to `seed02`. I couldn't understand why this happens. There are connections rejected because they are duplicated and EOF errors. It could be the case that both, by having two connections with the remote peer, one on each direction, choose the "wrong one" to close. The result is that no connection is established.

So, in summary, `validator02` and `seed02` are only able connect to each other, forming a network partition.

cason:
Is this still an issue?

sergio-mena:
I haven't seen it for ages, so if nobody saw it recently, I'd say we can close it for the time being (we can always reopen if needed)

PR #577, opened 2023-03-23
light client: Address panic in response to empty signed header

### LINKED ISSUE #575: light client: Empty RPC commit response triggers panic in E2E tests
### Setup

**CometBFT version** (use `cometbft version` or `git rev-parse --verify HEAD` if installed from source): `v0.34.x`

**ABCI app** (name for built-in, URL for self-written if it's publicly available): E2E

**Environment**:
- **OS** (e.g. from /etc/os-release): Ubuntu 22.04.2 LTS

### What happened?

Running our E2E tests locally on my ThinkPad running Ubuntu 22.04 LTS from the `v0.34.x` branch causes the light client to panic and the E2E tests to fail. This was not reproducible on other team members' Macbooks, but @andynog could reproduce this on his Ubuntu machine too.

```bash
cd test/e2e
make -j2 docker runner generator
./run-multiple.sh ./networks/ci.toml
```

The light client's logs: https://gist.github.com/thanethomson/7a3eeddff839adb33bf87af275b2c338#file-light01-log

`validator02`'s logs: https://gist.github.com/thanethomson/7a3eeddff839adb33bf87af275b2c338#file-validator02-log

It appears as though this is due to the fact that, for some unknown reason, `validator02` is not connecting to the rest of the network, and therefore has no blocks. This results in the `/commit` RPC endpoint returning a nil signed header because there's no block metadata for the latest block: https://github.com/cometbft/cometbft/blob/4a9076f8a95cde44b8a5071f13ed02b413aa7cc0/rpc/core/blocks.go#L125

The light client simply accepts this, because there's no error response from the RPC server: https://github.com/cometbft/cometbft/blob/4a9076f8a95cde44b8a5071f13ed02b413aa7cc0/light/provider/http/http.go#L183

It then panics here because the signed header (`sh`) is `nil`: https://github.com/cometbft/cometbft/blob/4a9076f8a95cde44b8a5071f13ed02b413aa7cc0/light/provider/http/http.go#L86

For the light client to not panic when it encounters this situation, but instead handle it more gracefully.

Also, does it make sense for the RPC to return a normal, non-error response here? Is it not technically better to return an error indicating that the relevant commit (which happens to be the latest commit) cannot be found?

### How to reproduce it

This is hard to reproduce, and it seems as though it's non-deterministic, since it doesn't always happen.

thanethomson:
Closes #575.

This at least addresses the panic in the light client, instead turning it into a temporary error (i.e. that the node does not yet have the signed header for the requisite height, which is true here).

I've tested this patch on the `v0.34.x` branch and it prevents the panic, but instead causes the E2E tests to hang indefinitely. As such, I've logged #576.

The code change is identical across `main`, `v0.37.x` and `v0.34.x`, so backports should be clean. I'll add changelog entries on the backport branches prior to merging them.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/light/provider/http/http.go
+++ b/light/provider/http/http.go
@@ -180,6 +180,15 @@ func (p *http) signedHeader(ctx context.Context, height *int64) (*types.SignedHe
 		commit, err := p.client.Commit(ctx, height)
 		switch {
 		case err == nil:
+			// See https://github.com/cometbft/cometbft/issues/575
+			// If the node is starting at a non-zero height, but does not yet
+			// have any blocks, it can return an empty signed header without
+			// returning an error.
+			if commit.SignedHeader.IsEmpty() {
+				// Technically this means that the provider still needs to
+				// catch up.
+				return nil, provider.ErrHeightTooHigh
+			}
 			return &commit.SignedHeader, nil
 
 		case regexpTooHigh.MatchString(err.Error()):
--- a/types/light.go
+++ b/types/light.go
@@ -120,6 +120,11 @@ type SignedHeader struct {
 	Commit *Commit `json:"commit"`
 }
 
+// IsEmpty returns true if both the header and commit are nil.
+func (sh SignedHeader) IsEmpty() bool {
+	return sh.Header == nil && sh.Commit == nil
+}
+
 // ValidateBasic does basic consistency checks and makes sure the header
 // and commit are consistent.
 //


types/light.go
```diff
@@ -120,6 +120,11 @@ type SignedHeader struct {
 	Commit *Commit `json:"commit"`
 }
 
+// IsEmpty returns true if both the header and commit are nil.
```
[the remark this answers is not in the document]
thanethomson: I needed a way here to test for an exactly empty signed header. Would a partially empty signed header not be picked up by one of the subsequent checks after the `signedHeader` call here? https://github.com/cometbft/cometbft/blob/4790ea3e46475064d5475c787427ae926c5a9e94/light/provider/http/http.go#L75

light/provider/http/http.go
```diff
@@ -180,6 +180,15 @@ func (p *http) signedHeader(ctx context.Context, height *int64) (*types.SignedHe
 		commit, err := p.client.Commit(ctx, height)
 		switch {
 		case err == nil:
+			// See https://github.com/cometbft/cometbft/issues/575
+			// If the node is starting at a non-zero height, but does not yet
+			// have any blocks, it can return an empty signed header without
+			// returning an error.
+			if commit.SignedHeader.IsEmpty() {
```
sergio-mena, line 187: I'm not convinced by this. Here, we are changing behavior in production.
The way a light client should work is propagating results from full nodes (after verifying them).
So, to me, the expected behavior from a light client is that it should reply the same as the full nodes that are backing it.
Is it possible to "coalesce" `ErrHeightTooHigh` and _empty signed header_ (my understanding of this change) in testing code rather than production code?
Please let me know if you would like to discuss this synchronously.

thanethomson: The question I would ask here is, is the current behaviour in production (panicking when receiving, what appears to be, a valid response from our RPC) correct? I don't think the behaviour's correct. I think we're hitting a corner case here that wasn't thought of previously, i.e. the node starting from a non-zero height, but not yet having started catching up (i.e. a bug).

Several other endpoints also return empty OK responses if the node doesn't yet have data. See the following cases:

https://github.com/cometbft/cometbft/blob/5d8789647b6759a3213f078c28279b0d16d1c754/rpc/core/blocks.go#L99

https://github.com/cometbft/cometbft/blob/5d8789647b6759a3213f078c28279b0d16d1c754/rpc/core/blocks.go#L114

https://github.com/cometbft/cometbft/blob/5d8789647b6759a3213f078c28279b0d16d1c754/rpc/core/blocks.go#L132

https://github.com/cometbft/cometbft/blob/5d8789647b6759a3213f078c28279b0d16d1c754/rpc/core/blocks.go#L142

sergio-mena: I took a closer look at this. I was under the impression that the modified method (`signedHeader`) was the implementation of a public RPC endpoint, but it is not the case, it's only used by method `LightBlock`.
So, even if coalescing "empty header" with `ErrHeightTooHigh` feels like we're making the caller of `signedHeader` lose info, in all fairness I don't see a better way of handling the panic we're hitting (I mean, better than what is proposed here).
So this convinces me that this change makes sense. Nevertheless, you might consider moving the `IsEmpty()` check to line [79](https://github.com/cometbft/cometbft/blob/5d8789647b6759a3213f078c28279b0d16d1c754/light/provider/http/http.go#L79), and have `LightBlock` return a specific error, similarly to what is done in line [82](https://github.com/cometbft/cometbft/blob/5d8789647b6759a3213f078c28279b0d16d1c754/light/provider/http/http.go#L82).
Thanks for your patience.

sergio-mena (approved):

light/provider/http/http.go
thanethomson: Not a problem!

The meaning associated with an empty response is: the chain has no data yet. This is the same response that would be returned if the light client requested a header for height 1 but the chain is at height 0. This particular arm of the switch statement is triggered in such a case (i.e. querying for height 1's data when the chain has no data yet): https://github.com/cometbft/cometbft/blob/5d8789647b6759a3213f078c28279b0d16d1c754/light/provider/http/http.go#L185

The error tells the light client that this is effectively a transient problem and the peer should not be removed as a provider, as per: https://github.com/cometbft/cometbft/blob/5d8789647b6759a3213f078c28279b0d16d1c754/light/provider/errors.go#L9-L11

In the case that I'm catering for here, it's only possible because in our E2E tests we start the chain at height 1000, but it doesn't yet have any blocks. The RPC server handles these two cases differently - if the chain's at height 0 and the request is for height 1, it returns an error; if the chain's at height 1000 and it has no data, it returns an empty OK response.

So our options here are to:
1. Handle the possibility of getting an empty OK response in the light client.
2. Change the behaviour of the RPC to return an error when the chain has no data, regardless of the base height of the chain.
3. Both 1 and 2.

I'm not sure what side effects option (2) would have though.

thanethomson: What's the benefit of distinguishing between the two errors? Would the light client knowing which of the two errors occurred make dealing with the error more actionable in some way? (It's only visible to the light client, and light clients generally don't have any control/influence over the nodes to which they're connecting)

To me the impact/effect of both errors is precisely the same so far, but I'm always open to being convinced otherwise.

thanethomson: Let's merge this now to at least remove the panic, and then I can follow up with a separate PR if there's value in distinguishing the error into two different errors?

sergio-mena: Sorry, long time to get back to this. I'm OK with leaving it as is. A common pattern I developed is to try to give as much info in error as possible, to allow (manual) troubleshooting. So, usually, I try to keep different error separate if possible (so a human troubleshooting can distinguish them if needed).
But again, just a suggestion. The PR is good as merged.

PR #579, opened 2023-03-24
rpc: tweaked the block search test

jmalicevic:
Improved the BlockSearch test within the RPC to make sure it matches against a value if the block events are generated. 

The previous version had a query that would never match and this way it is harder to test whether indexing was performed correctly. 

---

#### PR checklist

- [x ] Tests written/updated

### DIFF
--- a/rpc/client/rpc_test.go
+++ b/rpc/client/rpc_test.go
@@ -529,12 +529,16 @@ func TestBlockSearch(t *testing.T) {
 		require.NoError(t, err)
 	}
 	require.NoError(t, client.WaitForHeight(c, 5, nil))
-	// This cannot test match_events as it calls the client BlockSearch function directly
-	// It is the RPC request handler that processes the match_event
-	result, err := c.BlockSearch(context.Background(), "begin_event.foo = 100 AND begin_event.bar = 300", nil, nil, "asc")
+	result, err := c.BlockSearch(context.Background(), "begin_event.foo = 100", nil, nil, "asc")
 	require.NoError(t, err)
 	blockCount := len(result.Blocks)
-	require.Equal(t, blockCount, 0)
+	// if we generate block events within the test (by uncommenting)
+	// the code in line main_test.go:L23 then we expect len(result.Blocks)
+	// to be at least 5
+	// require.GreaterOrEqual(t, blockCount, 5)
+
+	// otherwise it is 0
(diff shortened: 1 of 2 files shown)


rpc/client/rpc_test.go
```diff
@@ -529,12 +529,16 @@ func TestBlockSearch(t *testing.T) {
...
 	require.NoError(t, client.WaitForHeight(c, 5, nil))
-	// This cannot test match_events as it calls the client BlockSearch function directly
-	// It is the RPC request handler that processes the match_event
-	result, err := c.BlockSearch(context.Background(), "begin_event.foo = 100 AND begin_event.bar = 300", nil, nil, "asc")
+	result, err := c.BlockSearch(context.Background(), "begin_event.foo = 100", nil, nil, "asc")
 	require.NoError(t, err)
 	blockCount := len(result.Blocks)
-	require.Equal(t, blockCount, 0)
+	// if we generate block events within the test (by uncommenting)
+	// the code in line main_test.go:L23 then we expect len(result.Blocks)
```
sergio-mena, line 536:
```suggestion
	// if we generate block events within the test (by uncommenting
	// the code in line main_test.go:L23) then we expect len(result.Blocks)
```

rpc/client/rpc_test.go
```diff
@@ -529,12 +529,16 @@ func TestBlockSearch(t *testing.T) {
...
+	// to be at least 5
+	// require.GreaterOrEqual(t, blockCount, 5)
+
+	// otherwise it is 0
+	require.GreaterOrEqual(t, blockCount, 0)
```
sergio-mena, line 541: If it _is_ 0 as the comment says, shouldn't the test be `require.Equal`, rather than `require.GreaterOrEqual`?

sergio-mena:
Should this be backported?

rpc/client/rpc_test.go
jmalicevic: I am sorry, indeed, I copied the statement before and forgot to revert this condition back to `Equal`.

sergio-mena (approved):
Thanks!

jmalicevic:
I added the backport labels, confirmed it works on all branches, but will need some tweaking for 0.34 as it has this match_events keyword. I'll take care of it.

PR #581, opened 2023-03-24
kvindexer: Removing event function type as it is always the same

### LINKED ISSUE #571: kvindexer: Remove event type from block event
## Feature Request

By introducing `FinalizeBlock` in v0.38.x, all block events will be returned in `ResponseFinalizeBlock`. Previously, 
the events were returned from both `BeginBlock` and `EndBlock`. The indexer would then store the type of the function the event belongs to. With `FinalizeBlock` replacing them, the current events are indexed with *one* type of function, thus storing this information is redundant and useless, while taking storage space. 

We propose removing the `finalize_block` value from the event key. 

### Summary

The information about the event function type should be removed from the event keys in the block indexer. 

### Problem Definition

Implementing this, removes extra bytes from the event key. THis might prove to have a significant impact as the indexer is anyways taking up a lot of storage.

The only possible disadvantage is supporting events indexed with older version (pre v0.38.x). Those events will contain event function type in their keys. But this can dealt with by accounting for that extra information on traversal. Potentially, we could offer the users the option to re-index the data and set a flag that the data is re-indexed. This would avoid checking for the parameter on every record. 

### Proposal

Indexing: Remove the `typ` parameter from the composite key in the `Index` method in `state/indexer/block/kv/kv.go`. 

Querying: In all the functions that read the events from the database, implement a check whether the `typ` value exists or not. Ignore it if it does. 

Potential improvement: Enable users t oconfigure a flag saying that the data indexed does not have `typ` stored so we can avoid the check described above. 

## DoD 

- [ ] Even function type removed
- [ ] Queries work with data indexed with older version and data indexed with new versions

jmalicevic:
Closes #571 .

Removes the event function type identifier form the event key. Earlier having this in the key made sense as it differentiated between `BeginBlock` and `EndBlock` events. With `FinalizeBlock` it will always be the same and has no point. 

The only issue is that we need to support data indexed with old versions where the key has the function identifier key (see `util.go:L123`. 

**Note**: I am piggybacking updates to the tests because there were duplicate UTs and UTs refering to `match_events` which do not exist in this version.

---

#### PR checklist

- [x] Tests written/updated
- [x] Backwards compatibility test
- [x] Run e2e upgrading test


state/indexer/block/kv/kv_test.go
```diff
@@ -229,26 +229,14 @@ func TestBlockIndexerMulti(t *testing.T) {
 			q:       query.MustCompile("block.height = 1"),
 			results: []int64{1},
 		},
-		"query return all events from a height - exact - no match.events": {
-			q:       query.MustCompile("block.height = 1"),
-			results: []int64{1},
-		},
```
sergio-mena, line 235: I don't understand why these test cases are no longer needed

state/indexer/block/kv/util.go
```diff
@@ -108,24 +107,36 @@ func parseHeightFromEventKey(key []byte) (int64, error) {
...
+	remaining, err := orderedcode.Parse(string(key), &compositeKey, &eventValue, &height)
 	if err != nil {
 		return 0, fmt.Errorf("failed to parse event key: %w", err)
 	}
 
-	// This is done to support previous versions that did not have event sequence in their key
+	// We either have an event sequence or a function type (potentially) followed by an event sequence.
+	// Potential scenarios:
+	// 1. Events indexed with v0.38 and later, will only have an event sequence
+	// 2. Events indexed up to v0.34.27 and v0.37.x will have a function type and an event sequence
```
sergio-mena, line 123:
```suggestion
	// 2. Events indexed between v0.34.27 and v0.37.x will have a function type and an event sequence
```

state/indexer/block/kv/util.go
```diff
@@ -108,24 +107,36 @@ func parseHeightFromEventKey(key []byte) (int64, error) {
...
-			return 0, fmt.Errorf("failed to parse event key: %w", err)
-		}
-		if len(remaining) != 0 {
-			return 0, fmt.Errorf("unexpected remainder in key: %s", remaining)
+		var typ string
+		remaining2, err := orderedcode.Parse(remaining, &typ) // Check if 2 or 3
+		if err != nil {                                       // If it cannot parse the event function type, it could be 1
+			remaining, err2 := orderedcode.Parse(string(key), &compositeKey, &eventValue, &height, &eventSeq)
+			if err2 != nil || len(remaining) != 0 { // We should not have anything else after the eventSeq
+				return 0, fmt.Errorf("failed to parse event key: %w", err)
```
sergio-mena, line 132: Do we also want to report `err2`?

state/indexer/block/kv/util.go
```diff
@@ -108,24 +107,36 @@ func parseHeightFromEventKey(key []byte) (int64, error) {
...
+			}
+
+		} else {
+			remaining, err2 := orderedcode.Parse(remaining2, &eventSeq) // If 2, , retrieve the eventSeq, otherwise ignore
+			if err2 != nil || len(remaining) != 0 {                     // We should not have anything else after the eventSeq
+				return 0, fmt.Errorf("failed to parse event key: %w", err)
```
sergio-mena, line 138: Report `err2`?

state/txindex/kv/kv_test.go
```diff
@@ -212,23 +212,19 @@ func TestTxSearchEventMatch(t *testing.T) {
...
+		" Match range with special character 2": {
 			q:             "account.number <= 2 AND account.owner = '/Ivan/.test' AND tx.height > 0",
 			resultsLength: 1,
 		},
-		" Match range with match events contains with multiple items": {
+		" Match range with contains with multiple items": {
 			q:             "account.number <= 2 AND account.owner CONTAINS '/Iv' AND account.owner CONTAINS 'an' AND tx.height = 1",
 			resultsLength: 1,
 		},
-		" Match range with match events contains": {
```
sergio-mena, line 231: Are we testing all "if" branches in the changes to file `state/indexer/block/kv/util.go` above?

state/indexer/block/kv/kv_test.go
jmalicevic: BEcause they are the exact duplicate of the existing tests - they were there because we used to test with and without the match-events keywords and then when I fixed the tests for not use the keyword, I did not delete the duplicate ones.

state/indexer/block/kv/util.go
```diff
@@ -108,24 +107,36 @@ func parseHeightFromEventKey(key []byte) (int64, error) {
...
+		var typ string
+		remaining2, err := orderedcode.Parse(remaining, &typ) // Check if 2 or 3
+		if err != nil {                                       // If it cannot parse the event function type, it could be 1
+			remaining, err2 := orderedcode.Parse(string(key), &compositeKey, &eventValue, &height, &eventSeq)
+			if err2 != nil || len(remaining) != 0 { // We should not have anything else after the eventSeq
+				return 0, fmt.Errorf("failed to parse event key: %w", err)
+			}
+
+		} else {
+			remaining, err2 := orderedcode.Parse(remaining2, &eventSeq) // If 2, , retrieve the eventSeq, otherwise ignore
```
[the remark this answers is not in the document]
jmalicevic: If 2 - the case under number 2. above , that have both an event sequence and event type. 
I will rephrase this, but essentially ignore means, that we do not have the event sequence, but rather just the event function type (case 3.), and we will ignore. Err2 should still be returned.

state/txindex/kv/kv_test.go
jmalicevic: I could not think of how to test the cases with UTs because those if branches are handling events indexed using old versions of the indexer. Unless we create different indexing options in our UTs. We could provide a custom indexing function to generate the event key differently and then test. That is basically what i did manually. And I did this with the backwards compatibility tests when running on top of data indexed with old versions.

state/indexer/block/kv/util.go
```diff
@@ -49,13 +49,12 @@ func heightKey(height int64) ([]byte, error) {
...
 }
 
-func eventKey(compositeKey, typ, eventValue string, height int64, eventSeq int64) ([]byte, error) {
+func eventKey(compositeKey, eventValue string, height int64, eventSeq int64) ([]byte, error) {
 	return orderedcode.Append(
 		nil,
 		compositeKey,
 		eventValue,
 		height,
-		typ,
```
thanethomson, line 58: Since this changes the format in which data is stored in the index, this is a breaking change, is it not? If so, what does the upgrade path look like for users with existing indexes from v0.34/v0.37? Do they need to reindex? If so I'd recommend adding some instructions in the `UPGRADING.md` in the backport branch that Mergify creates.

(We shouldn't have a changelog entry on `main` or extra `UPGRADING.md` instructions if this is going to be released from a backport branch)

jmalicevic: Thane, sorry, I am just seeing this. People will not need to re-index if they do not want because this is handled by all the if-else statements in parsing the event. But info on this has already been added to the [upgrading.md ](https://github.com/cometbft/cometbft/pull/591). Let me know if you feel we should revisit the text there.

state/indexer/block/kv/util.go
```diff
@@ -108,24 +107,38 @@ func parseHeightFromEventKey(key []byte) (int64, error) {
...
-		if len(remaining) != 0 {
-			return 0, fmt.Errorf("unexpected remainder in key: %s", remaining)
+		var typ string
+		remaining2, err := orderedcode.Parse(remaining, &typ) // Check if we have scenarios 2. or 3. (described above).
+		if err != nil {                                       // If it cannot parse the event function type, it could be 1.
+			remaining, err2 := orderedcode.Parse(string(key), &compositeKey, &eventValue, &height, &eventSeq)
+			if err2 != nil || len(remaining) != 0 { // We should not have anything else after the eventSeq.
+				return 0, fmt.Errorf("failed to parse event key: %w; and %w", err, err2)
+			}
+
```
sergio-mena, line 134:
```suggestion
```

state/indexer/block/kv/util.go
```diff
@@ -108,24 +107,38 @@ func parseHeightFromEventKey(key []byte) (int64, error) {
...
+		} else {
+			remaining, err2 := orderedcode.Parse(remaining2, &eventSeq) // If the event follows the scenario in 2.,
+			// retrieve the eventSeq, otherwise, there should be no remainder
+			// and no error and this parsing has no effect.
+			if err2 != nil || len(remaining) != 0 { // We should not have anything else after the eventSeq if in 2.
+				return 0, fmt.Errorf("failed to parse event key: %w", err2)
+			}
```
sergio-mena, line 141:
```suggestion
			if len(remaining2) != 0 { // Are we in case 2 or 3
				remaining, err2 := orderedcode.Parse(remaining2, &eventSeq) // the event follows the scenario in 2.,
				// retrieve the eventSeq
				// there should be no error
				if err2 != nil { // We should not have anything else after the eventSeq if in 2.
					return 0, fmt.Errorf("failed to parse event key: %w", err2)
				}
			}
```

I'm suggesting this, because I don't clearly see the path taken by scenario 3 in the present code.
I also see two main advantages with my suggestion:
* Easier to distinguish paths followed by scenarios 2 and 3
* You don't call `Parse` unless you're sure you need to call it (cleaner)

state/indexer/block/kv/util.go
```diff
@@ -108,24 +107,38 @@ func parseHeightFromEventKey(key []byte) (int64, error) {
...
 		}
 	}
```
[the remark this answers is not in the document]
jmalicevic: Agree.

state/indexer/block/kv/util.go
jmalicevic: I am fine with the chaneg but we still need to check whether len(remaining) != 0 because that is invalid. I will push a change including both. Let me know if you agree.

.changelog/unreleased/improvements/581-remove-event-function-type-from-indexer.md
```diff
@@ -0,0 +1,2 @@
+- `[kvindex]` Remove event function type from event key in kvindexer. The function type is not always `FinalizeBlock`.
+  ([\#581](https://github.com/cometbft/cometbft/pull/581))
```
thanethomson, line 2: What's the impact here on the user? If there's no direct impact on the user, it's not clear to me that we'd need a changelog entry. If the overall goal is to reduce storage consumption, then I'd recommend rewording to something that explicitly highlights that. Users may not understand what this changelog entry means for them otherwise.

Also, if it's being backported to v0.38, then the PR to `main` shouldn't have a changelog entry - only the backport PR should have a changelog entry (from a user's perspective, they'd be receiving this change in the v0.38 release series, so there'd be no need to also be notified about it in the v0.39 series).

jmalicevic: Makes sense. I will remove the changelog entry here. As for the impact on the user, there is no direct impact as far as I can see. We have info on this in the upgrading docs for 0.38 because the format of data on disk changes. However, people who have forks that play with the indexer, they might be impacted.

state/indexer/block/kv/util.go
```diff
@@ -108,25 +107,42 @@ func parseHeightFromEventKey(key []byte) (int64, error) {
...
+		return 0, fmt.Errorf("failed to parse event sequence: %w", err)
 	}
 
-	// This is done to support previous versions that did not have event sequence in their key
+	// We either have an event sequence or a function type (potentially) followed by an event sequence.
+	// Potential scenarios:
+	// 1. Events indexed with v0.38.x and later, will only have an event sequence
+	// 2. Events indexed between v0.34.27 and v0.37.x will have a function type and an event sequence
+	// 3. Events indexed before v0.34.27 will only have a function type
+	// function_type = 'being_block_event' | 'end_block_event'
```
lasarojc, line 125: I'd suggest putting the remaining == 0 case first and reducing the indentation level of the what's in the current if block.

```
if len(remaining) != 0 {
		return 0, fmt.Errorf("failed to parse event sequence, invalid event format")
	}
```

state/indexer/block/kv/util.go
```diff
@@ -108,25 +107,42 @@ func parseHeightFromEventKey(key []byte) (int64, error) {
...
+			}
+		} else {
+			if len(remaining2) != 0 { // Are we in case 2 or 3
+				remaining, err2 := orderedcode.Parse(remaining2, &eventSeq) // the event follows the scenario in 2.,
+				// retrieve the eventSeq
+				// there should be no error
+				if err2 != nil || len(remaining) != 0 { // We should not have anything else after the eventSeq if in 2.
+					return 0, fmt.Errorf("failed to parse event sequence: %w", err2)
+				}
+			}
```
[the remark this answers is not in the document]
jmalicevic: Looks indeed cleaner. Pushing now. Sorry, this logic was super tricky so when I wrote it down, I was just following a thread in my head. Thanks for the suggestion!

sergio-mena (approved):
Thanks for the refactoring of the intricate code!

jmalicevic:
Thank you guys for improving it! @lasarojc , let me know if you are happy with the changes and I can merge this :)

lasarojc (approved):

Issue #601, opened 2023-03-28
spe/p2p: Document execution flows for establishing connections


cason:

cason:
> Link here the knowledge-base spec in Quint.

Added: https://github.com/cometbft/knowledge-base/pull/12/commits/443dafce0a120e9c81aabe25766d7b106affddea

Issue #521, opened 2023-03-13
Bug: losing all peers with `peer switch stopped` error


czarcas7ic:
# CometBFT peer issue

## Bug Report

For the last two months, nodes on Osmosis are frequently halting with:
```
ERR Won’t start a peer – switch is not running module=p2p peer={“Data”:{}, “Logger”:{}}
```
(presumably coming from: https://github.com/cometbft/cometbft/blob/877b0629578c7695dec77b16cf0021563e060d82/p2p/switch.go#L817-L821C3 )

Validators are saying that they are having issues such as:
```
We have a working node, that then stops progressing and faces the same error every 2hrs. Node is getting stuck connecting to peers.
Updating live peers and restarting is working, but have to keep doing it often to keep the node alive.
```

### Setup

**CometBFT version** Tendermint v0.44.22

**Have you tried the latest version**: no (trying to upgrade to it, working through some SDK dependency issues)

**ABCI app**: Osmosis

**Environment**:
- **OS** (e.g. from /etc/os-release): Ubuntu 20.04
- **Install tools**: 
- **Others**:

**node command runtime flags**:

### Config

https://pastebin.com/Xs6nKUku

### What happened?

Nodes halt frequently facing peer issues. They appear to be working fine with peers prior to this halt point. They halt with the error previously explained.

Some interestings notes:

* Validators / node operators seem to experience it together (i.e. for an unknown reason, the last couple days validators / node operators have not experienced the issue)
* I have not experienced this issue on our testnet despite utilizing the same binary

### What did you expect to happen?

Node to continue running.

### How to reproduce it

Run a node on mainnet Osmosis (I have been told it happens on other networks as well, but can not personally confirm).

The current "fix" is to continuously restart the node at regular intervals to prevent this error, which is not ideal.

### Logs

```
ERR Won’t start a peer – switch is not running module=p2p peer={“Data”:{}, “Logger”:{}}
```

### `dump_consensus_state` output

Not a consensus bug

cason:
Hi, can you provide a full log of the problematic node for our analysis? The error is very likely produced where you pointed out, but I couldn't understand why the `Switch` would not be running (i.e., why it was stopped) on the reported situation.

Moreover, can you confirm the adopted CometBFT/Tendermin version?

andynog:
another useful information that might be helpful for troubleshooting is the P2P configuration under:
```
#######################################################
###           P2P Configuration Options             ###
#######################################################
[p2p]
```
[...]

PR #616, opened 2023-03-30
WIP: Multi process p2p peer management Quint spec

josef-widder:
This is a spec for p2p peer management in Quint. In contrast to the previous draft PR, 
- this is set up for multiple nodes (the nodes we want to model explicitly can be determined in the initial state)
- the specification proposes several sections that mirror what we had in English specs
   - data structures
   - local functions (shall replace pseudo code)
   - definition of a transition system
       - variables that describe a global state
       - actions that describe how the local functions (defined above) transform the global state
       - tests (sequences of actions with assertions)
       - classic transition system definitions
           - initial state(s)
           - a step function (currently it is quite coarse as it doesn't link the modeled nodes yet)

Next steps:
- I would like to get feedback on the structure
- Together with @cason, 
    - we will make sure the function definitions capture what we need
    - formulate test scenarios that are interesting to understand what works in p2p right now, and what are problematic cases
    - perhaps encode a proper step action for the multi-process case to be able to help us analyze the system.  

---

#### PR checklist

- [ ] Tests written/updated


spec/p2p/p2p.qnt
```diff
@@ -0,0 +1,557 @@
...
+
+    pure def Peers(s: LocalState): Set[Address] = s.peerState.keys()
+
+    pure def activePeers(s: LocalState): Set[Address] = 
+        s.Peers().filter(p =>
+            and {
+                s.peerState.get(p) != "reconnecting",
+                or {
+                    s.connections.contains({address: p, state: "outgoing"}),
+                s.connections.contains({address: p, state: "incoming"})
```
lasarojc, line 77:
```suggestion
                    s.connections.contains({address: p, state: "incoming"})
```

spec/p2p/p2p.qnt
```diff
@@ -0,0 +1,557 @@
...
+    action step = any {
+        nondet a = oneOf(modeledNodes) 
+        nondet outgoing = oneOf(config.get(a).AddressBook)
+        nondet incoming = oneOf(Environment)
+        nondet conn = oneOf(config.get(a).connections.union(Set({address: {ID: "default", networkAddress: "0.0.0.0"}, 
+                                                                 state: "outgoing"})))
+        nondet dial = oneOf(OperatorAddressBook.powerset())
+        nondet seedResponse = oneOf(Environment.powerset())
+            any {            
+                actionDialPeer(a, outgoing),
```
lasarojc, line 526: `actionDialPeer` always succeeds, even if `a` is not dialed so this `any` may always succeed here and the action is in practice just a stutter.

josef-widder: Yes. I agree. I didn't put much effort in making `step` precise yet. What I also didn't do, for instance, is to set up `step` so that actually two modelled nodes are "connecting". 

First I really would like to get more clarity on what we would like to investigate. Then I would envisage a workflow where we first write tests, and only then think in more detail about making `step` more precise.

.github/workflows/specs.yml
```diff
@@ -0,0 +1,17 @@
+name: Specs
```
josef-widder, line 1:
```suggestion
on:
  pull_request:

name: Specs
```

spec/p2p/p2p.qnt
```diff
@@ -0,0 +1,577 @@
...
+                         .union(Set({address: a, state: "lost"}))
+        s.with("connections", y)
+    }
+
+ // return DialPeerSuccess -> if connection.address.id not in Peers -> peerState(address) = "outbound" 
+ //                           otherwise -> delete connection from map 
+ // in the spec right now: the peer might not exists, be already outbound (stutter), be reconnecting
+ // TODO: check the cases
+    action lowerLevelDialPeerSuccess(node: Address, a: Address): bool = all {
+        require(config.get(node).connections.contains({address: a, state: "outgoing"})),
```
cason, line 339: This line summarize the `isDialing` predicate, right? Actually it should be called `isDialingOrOutboundPeer`.

spec/p2p/p2p.qnt
```diff
@@ -0,0 +1,577 @@
...
+        s.with("connections", y)
+    }
+
+ // return DialPeerSuccess -> if connection.address.id not in Peers -> peerState(address) = "outbound" 
+ //                           otherwise -> delete connection from map 
+ // in the spec right now: the peer might not exists, be already outbound (stutter), be reconnecting
+ // TODO: check the cases
+    action lowerLevelDialPeerSuccess(node: Address, a: Address): bool = all {
+        require(config.get(node).connections.contains({address: a, state: "outgoing"})), 
+        if (config.get(node).Peers().filter(x => x.ID == a.ID) == Set()) all {
```
cason, line 340: For me this is a high level logic (at the `Switch`), not low level (at the `Transport`).

spec/p2p/p2p.qnt
```diff
@@ -0,0 +1,577 @@
...
+            s.connections.exclude(Set({address: a, state: "outgoing"}, {address: a, state: "incoming"}))
+                         .union(Set({address: a, state: "lost"}))
+        s.with("connections", y)
+    }
+
+ // return DialPeerSuccess -> if connection.address.id not in Peers -> peerState(address) = "outbound" 
+ //                           otherwise -> delete connection from map 
+ // in the spec right now: the peer might not exists, be already outbound (stutter), be reconnecting
+ // TODO: check the cases
+    action lowerLevelDialPeerSuccess(node: Address, a: Address): bool = all {
```
cason, line 338: This method, in my opinion, should create an inbound connection at the destination peer. So, it would be something like the `transportAccept` at the destination peer.

spec/p2p/p2p.qnt
```diff
@@ -0,0 +1,577 @@
...
+        require(config.get(node).connections.contains({address: a, state: "outgoing"})), 
+        if (config.get(node).Peers().filter(x => x.ID == a.ID) == Set()) all {
+            // If I don't have the ID in my peers already, add the address
+            config' = config.set(node, config.get(node).putPeer(a, "outbound"))
+        }
+        else {
+            // I have the ID already. That is, it might be incoming, or I might have dialed over a different
+            // (network) address
+            config' = config.set(node, config.get(node).removeConnectionOutgoing(a))
```
cason, line 347: This is not a successful connection.

spec/p2p/p2p.qnt
```diff
@@ -0,0 +1,577 @@
...
+        },
+        environmentUnchanged,
+    }
+
+    action lowerLevelDialPeerFailed(node: Address, a: Address): bool = all {
+        require(config.get(node).connections.contains({address: a, state: "outgoing"})),
```
cason, line 353: This is the same as `isDialing`.

spec/p2p/p2p.qnt
```diff
@@ -0,0 +1,577 @@
...
+            // (network) address
+            config' = config.set(node, config.get(node).removeConnectionOutgoing(a))
+        },
+        environmentUnchanged,
+    }
+
+    action lowerLevelDialPeerFailed(node: Address, a: Address): bool = all {
+        require(config.get(node).connections.contains({address: a, state: "outgoing"})), 
+        // TODO: perhaps add some checks
+        config' = config.set(node, config.get(node).removeConnectionOutgoing(a)),
```
cason, line 355: This would be the same as remove the peer from the dialing state, right?

spec/p2p/p2p.qnt
cason: Although this appears to be needed in order to remove the peer from the dialing state.

spec/p2p/p2p.qnt
```diff
@@ -0,0 +1,577 @@
...
+        }
+    }
+
+    action pexEnsurePeers(node: Address): bool = all {
+        //val addr = s.AddressBook.chooseSome() // addressbook.pickAddress(Bias)
+        nondet addr = oneOf(config.get(node).AddressBook)
+        config' = config.set(node, config.get(node).ensurePeers(addr)),
+        environmentUnchanged,
+    }
+
```
cason, line 442: Regarding `pexEnsurePeers` and `ensurePeers`.

In the current logic `ensurePeers` checks the condition (number of outbound peers), then chooses a random address to dial. So `pexEnsurePeers` should call `s.dialPeer(addr)`and be invoked by `ensurePeers`.

spec/p2p/p2p.qnt
```diff
@@ -0,0 +1,577 @@
...
+    pure val connectionStates = Set("incoming", "outgoing", "lost")
+
+    // (public_key, ipAddress:port)
+    type Address = {ID:str, networkAddress:str}
+
+    type Connection = {
+        address : Address,
+        state: str, // "incoming" "outgoing" "lost"
+    }
+
```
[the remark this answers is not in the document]
cason: `Config` has several elements that are inside `LocalState` now and `ListenAddress`.

When we connect to the node the incoming connection must have address (node.ID, node.Config.ListenAddress). Nodes do not need to have a `ListenAddrs`. Those nodes of course don't accept connections.

cason (commented):
I am still a little confused or not fully satisfied with the modelling of connections and peers, we can discuss it synchronously before I do a full pass on the spec.

cason:
I wonder whether we should include the reference English [file](https://github.com/cometbft/cometbft/blob/91f8dc3ea0490fd824d1d3d346fb0da61de2f848/spec/p2p/architecture/peers-connect.md) in this PR.

It does not define types and abstractions properly, though. Maybe we need another English-spec file defining it, with a semi-direct mapping to the Quint spec.

spec/p2p/p2p.qnt
```diff
@@ -0,0 +1,577 @@
...
+ * ************************************************************************* */
+
+    // TODO: "pure val" is a work around as we don't have enums yet
+    pure val peerStates = Set("reconnecting", "inbound", "outbound")
+    pure val connectionStates = Set("incoming", "outgoing", "lost")
+
+    // (public_key, ipAddress:port)
+    type Address = {ID:str, networkAddress:str}
+
+    type Connection = {
```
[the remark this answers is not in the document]
josef-widder: Into the transition system:

- type rawConnection = {from : Address, to:Address}
- network = Set[rawConnection]
- establishedConnection =  Set[rawConnection]
- -> actionDialPeer we are adding a rawconnection to network
- -> actionTransportAccept is guarded by node appears in to field
- -> and the follow up in 'step' regarding which actions should be executed to success/failed

spec/p2p/p2p.qnt
```diff
@@ -0,0 +1,577 @@
...
+   // node dials a. This starts lower lever computations. 
+    action actionDialPeer (node: Address, a:Address):bool = all {
+        config' = config.set(node, config.get(node).dialPeer(a)),
+        environmentUnchanged,
+    }
+
+    action actionTransportAccept (node: Address, a:Address):bool = all {
+        config' = config.set(node, config.get(node).transportAccept(a)),
+        environmentUnchanged,
+    }
```
josef-widder, line 304: if (config.get(node) == config.get(node).transportAccept(a)), that is, if the node doesn't accept connection, then record that in the rawConnection. 
Then having a failed rawConnection should enable dialpeerFailed on the other side (if the other side is  a modeledNode

spec/p2p/p2p.qnt
josef-widder: merge the pure def and the action into an action so that the 'if' is on the right place

adizere:
For my reference: why did we not finish/merge this spec?

I am asking because I was looking to understand the P2P part better. There's quite a few issues related to P2P (eg https://github.com/cosmos/gaia/issues/2347, https://github.com/cometbft/cometbft/issues/521) and the logic is very intertwined across modules, making it difficult to onboard, triage, debug. Specs would be helpful!

cason:
Hey Adi, there is an issue tracking this effort: https://github.com/cometbft/cometbft/issues/601

cason:
It is on my plan for this quarter, but for now it should be the last one of the 4 I have planned.

josef-widder:
I thought at this point it doesn't make sense to keep it open and pollute the repo. I have moved the spec over to the [knowledge base](https://github.com/cometbft/knowledge-base/tree/main/quint/p2p-connection-management), where it waits until Daniel is ready to resume.

Issue #604, opened 2023-03-28
Formatting with gofumpt


thanethomson:
There have been discussions previously about potentially formatting (and enforcing formatting) with [gofumpt](https://github.com/mvdan/gofumpt) (see tendermint/tendermint#9738, tendermint/tendermint#9208, tendermint/tendermint#8478).

The main argument against this was that doing a big-bang reformat of the whole codebase significantly hampers our ability to do code archeology, which, with a codebase like this, is incredibly important for our team to be able to make meaningful progress.

A compromise here would be to start to enforce this locally - i.e. on each of our own machines when saving our code, that only the files each of us touch in a specific PR get formatted with gofumpt. As we progressively refactor the system, I'm fairly confident we'll end up touching most of the codebase, simultaneously progressively formatting the code.

At some point, once the diff between formatting and not formatting the codebase with gofumpt drops below a certain level (say a few hundred LoC), we can simply format the remainder of the code in a single PR and implement a formatting enforcement check as part of CI (at which point we can close this issue).

PR #625, opened 2023-03-31
Fix lints

thanethomson:
This touches a large number of files, but I believe it's necessary as part of our tech debt cleanup. There are so many functions littering the codebase where:

1. We have unused parameters and they should have been refactored away, but they weren't
2. We have unused parameters and they should be named according to proper Go conventions (e.g. in interface implementations where a particular function signature is required, but the variable names should either be left out or `_` when unused)
3. We have redundant code (a whole bunch of redundant `if` statements and error checks, for example)
4. We use bad naming conventions for variables, like `copy` or `len`

I'm also tired of having the linter fail locally. And if we don't do this, it will all just rot even more.

This PR targets `main`, but should be applied across all the backport branches (which will, of course, require some conflict resolution, but I'm fine with doing that). I've left TODOs in some places where changes need to be made in follow-up PRs.

Commits are organized by package so that it's hopefully easier to review.

If you pick up on formatting changes, please see #604.

---

#### PR checklist

- [ ] Tests written/updated


cmd/cometbft/commands/root.go
```diff
@@ -98,7 +98,8 @@ var RootCmd = &cobra.Command{
 }
 
 // deprecateSnakeCase is a util function for 0.34.1. Should be removed in 0.35
-func deprecateSnakeCase(cmd *cobra.Command, args []string) {
+// TODO(thane): Remove this across all releases
```
thanethomson, line 101: See #626.

lasarojc (approved):
while I dislike some of the the practices pushed by the Lint, I all all for uniformity.
LGTM

p2p/test_util.go
```diff
@@ -178,11 +178,9 @@ func StartSwitches(switches []*Switch) error {
 func MakeSwitch(
 	cfg *config.P2PConfig,
 	i int,
-	network, version string,
```
sergio-mena, line 181: Good catch!

rpc/grpc/api.go
```diff
@@ -12,12 +12,12 @@ type broadcastAPI struct {
...
 }
 
-func (bapi *broadcastAPI) Ping(ctx context.Context, req *RequestPing) (*ResponsePing, error) {
+func (bapi *broadcastAPI) Ping(context.Context, *RequestPing) (*ResponsePing, error) {
 	// kvstore so we can check if the server is up
 	return &ResponsePing{}, nil
 }
 
-func (bapi *broadcastAPI) BroadcastTx(ctx context.Context, req *RequestBroadcastTx) (*ResponseBroadcastTx, error) {
+func (bapi *broadcastAPI) BroadcastTx(_ context.Context, req *RequestBroadcastTx) (*ResponseBroadcastTx, error) {
```
sergio-mena, line 20: We are setting the variable name to "_" in line 20, and removing the name altogether in line 15?
Can we treat the two cases the same way'

sergio-mena (approved):
Thanks for this @thanethomson
I like it because there is zero risk of breaking something and the changes are sparse enough not to hinder git-blame significantly.
One thought: since this is strictly refactoring, is it worth your time backporting it to other branches?

rpc/grpc/api.go
thanethomson: The convention in Go is that, if _all_ function parameters are unused, then you can have unnamed parameters. But the moment that one of your parameters is used, they all have to be named, in which case the unused ones must be called `_`.

thanethomson:
>  One thought: since this is strictly refactoring, is it worth your time backporting it to other branches?

If we don't, our local linter's going to fail when linting those branches, so I think I'll have to backport this.

PR #626, opened 2023-03-31
cmd: Remove snake case deprecation warning

thanethomson:
While working on #625 I discovered this. We have aliases for both types of commands, so I don't think there's any need to issue a warning here.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/cmd/cometbft/commands/reset.go
+++ b/cmd/cometbft/commands/reset.go
@@ -24,8 +24,9 @@ var keepAddrBook bool
 
 // ResetStateCmd removes the database of the specified CometBFT core instance.
 var ResetStateCmd = &cobra.Command{
-	Use:   "reset-state",
-	Short: "Remove all the data and WAL",
+	Use:     "reset-state",
+	Aliases: []string{"reset_state"},
+	Short:   "Remove all the data and WAL",
 	RunE: func(cmd *cobra.Command, args []string) (err error) {
 		config, err = ParseConfig(cmd)
 		if err != nil {
--- a/cmd/cometbft/commands/reset.go
+++ b/cmd/cometbft/commands/reset.go
@@ -18,16 +18,14 @@ var ResetAllCmd = &cobra.Command{
 	Aliases: []string{"unsafe_reset_all"},
 	Short:   "(unsafe) Remove all the data and WAL, reset this node's validator to genesis state",
 	RunE:    resetAllCmd,
-	PreRun:  deprecateSnakeCase,
 }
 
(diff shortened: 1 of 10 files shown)


cmd/cometbft/commands/reset.go
```diff
@@ -18,16 +18,14 @@ var ResetAllCmd = &cobra.Command{
...
 }
 
 var keepAddrBook bool
 
 // ResetStateCmd removes the database of the specified CometBFT core instance.
 var ResetStateCmd = &cobra.Command{
-	Use:    "reset-state",
-	Short:  "Remove all the data and WAL",
-	PreRun: deprecateSnakeCase,
+	Use:   "reset-state",
```
adizere, line 27: FYI most other commands have a snake_case alias (eg `gen_validator`, `probe_upnp`..) but `reset-state` does not have.

adizere (approved):

sergio-mena:
The changes make sense technically. However, I have the impression that they added the "deprecate snake case warning" because they were planning to remove the (snake_case) aliases in a subsequent breaking release.
(take a look at [this](https://github.com/tendermint/tendermint/pull/5786) and [this](https://github.com/tendermint/tendermint/pull/5777))

thanethomson:
Do we still want to remove the snake_case aliases then? Or shouldn't we just offer both? (kebab-case is easier to type when using the CLI)

And wasn't the snake_case deprecation part of the broader effort to kebab-case the config file?

sergio-mena:
> Do we still want to remove the snake_case aliases then? Or shouldn't we just offer both? (kebab-case is easier to type when using the CLI)

I think that deserves a (maybe short) synchronous discussion? I've added a point for tomorrow's internal meeting. I apologize for adding a point to a meeting I'll probably not be able to attend :-|
But I am OK with whatever you folks decide.

> And wasn't the snake_case deprecation part of the broader effort to kebab-case the config file?

I think so, although, by the age of tendermint/tendermint#5786 and tendermint/tendermint#5777, I'd say they first tackled this, and later (as of `v0.35.x`) they tackled the `config.toml`.

thanethomson:
> I think that deserves a (maybe short) synchronous discussion?

We spoke about it in the team call today and we decided it'd be less disruptive to users to:
1. Continue to offer kebab-case CLI commands as the default, as it's generally more ergonomic to type than snake_case when using the CLI.
2. Continue to provide snake_case CLI aliases to not break folks who depend on them.
3. Remove the deprecation warning.

PR #637, opened 2023-04-03
mempool: Add metric to measure how many times a tx was received

### LINKED ISSUE #634: mempool: Add metric to measure how many times a tx was received
## Feature Request

Related to #11. Prerequisite for #613.

### Problem Definition

Currently transactions are propagated to all peers except to the peer that sent the transaction. Then a transaction may reach a node multiple times. A metric for counting the number a times a transaction was received by a node would help to improve the gossiping mechanism.

### Proposal

Exposing one metric for each tx would add a huge number of Prometheus entries at each measurement. A better way would be to add a new metric of type histogram with sizes of buckets increasing exponentially. For example, buckets of size 1 (that is, number of transactions received at most once), 2, 4, 8, and more than 8.

hvanz:
Closes #634 

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/mempool/metrics.gen.go
+++ b/mempool/metrics.gen.go
@@ -52,16 +52,25 @@ func PrometheusMetrics(namespace string, labelsAndValues ...string) *Metrics {
 			Name:      "recheck_times",
 			Help:      "Number of times transactions are rechecked in the mempool.",
 		}, labels).With(labelsAndValues...),
+		TimesTxsWereReceived: prometheus.NewHistogramFrom(stdprometheus.HistogramOpts{
+			Namespace: namespace,
+			Subsystem: MetricsSubsystem,
+			Name:      "times_txs_were_received",
+			Help:      "Histogram of times a transaction was received.",
+
+			Buckets: stdprometheus.ExponentialBuckets(1, 2, 5),
+		}, labels).With(labelsAndValues...),
 	}
 }
 
 func NopMetrics() *Metrics {
 	return &Metrics{
-		Size:         discard.NewGauge(),
-		TxSizeBytes:  discard.NewHistogram(),
-		FailedTxs:    discard.NewCounter(),
-		RejectedTxs:  discard.NewCounter(),
-		EvictedTxs:   discard.NewCounter(),
-		RecheckTimes: discard.NewCounter(),
+		Size:                 discard.NewGauge(),
+		TxSizeBytes:          discard.NewHistogram(),
+		FailedTxs:            discard.NewCounter(),
+		RejectedTxs:          discard.NewCounter(),
+		EvictedTxs:           discard.NewCounter(),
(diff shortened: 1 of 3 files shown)


mempool/metrics.gen.go
```diff
@@ -40,28 +40,37 @@ func PrometheusMetrics(namespace string, labelsAndValues ...string) *Metrics {
...
 			Namespace: namespace,
 			Subsystem: MetricsSubsystem,
 			Name:      "recheck_times",
 			Help:      "Number of times transactions are rechecked in the mempool.",
 		}, labels).With(labelsAndValues...),
+		AlreadyReceivedTxs: prometheus.NewCounterFrom(stdprometheus.CounterOpts{
+			Namespace: namespace,
+			Subsystem: MetricsSubsystem,
+			Name:      "already_received_txs",
+			Help:      "Number of times transactions were received more than once.",
```
lasarojc, line 53:
```suggestion
			Help:      "Number of duplicate transaction reception (conservative estimate).",
```

mempool/metrics.go
```diff
@@ -31,13 +34,41 @@ type Metrics struct {
...
-	// transactions that passed CheckTx and existed in the mempool but were later
-	// evicted to make room for higher priority valid transactions that passed
-	// CheckTx.
-	//metrics:Number of evicted transactions.
-	EvictedTxs metrics.Counter
-
 	// Number of times transactions are rechecked in the mempool.
 	RecheckTimes metrics.Counter
+
+	// Number of times transactions were received more than once.
```
lasarojc, line 40:
```suggestion
	// Number of times transactions were received after the first time. 
	//This is a conservative as new receptions may happen after cache purge.
```

lasarojc (approved):

mempool/metrics.gen.go
hvanz: This is an auto generated file. Cannot make changes here.

.changelog/v0.37.0/improvements/634-mempool-metric-times-tx-was-received.md
```diff
@@ -0,0 +1,3 @@
+- `[mempool]` Add two metrics to measure how many times a tx was received (a
```
thanethomson, line 1: This shouldn't be added directly to the v0.37.0 changelog, since we've already released that version.

Which version(s) of the software are you targeting? If it's just `main`, then a changelog entry in `.changelog/unreleased/improvements` is fine. Otherwise please don't add a changelog to this PR, but add it instead on the backport branch you're targeting.

hvanz:
The final version of this PR just adds the simple counter to measure duplicate transactions. I think that will be enough for our needs for the moment. If we see that in the future the histogram is useful, we'll create a separate PR for it.

PR #639, opened 2023-04-03
p2p: kcp instead of tcp

faddat:
I am nearly certain that this isn't a priority, and that's no big deal.  

From time to time I would remember this PR, which I wrote while playing about with Sei and getting 40% uptime from Vietnam. 

Theoretically speaking (and indeed I do think in practice) this replaces the tcp transport that comet uses with kcp, which....

https://github.com/skywind3000/kcp

https://ims.improbable.io/insights/kcp-a-new-low-latency-secure-network-stack

https://mirror-networking.gitbook.io/docs/manual/transports/kcp-transport

------

Theoretically.....

this would make comet faster, but we'd need a (crazy) testnet to know. 

PS: this thing likely f(gfw) -- I have seen it do exactly that. 

https://stratcomcoe.org/cuploads/pfiles/belarus_protests_web_nato_stratcom_coe.pdf

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/p2p/netaddress.go
+++ b/p2p/netaddress.go
@@ -14,13 +14,8 @@ import (
 	"strings"
 	"time"
 
-<<<<<<< HEAD
 	tmp2p "github.com/cometbft/cometbft/proto/tendermint/p2p"
-=======
 	kcp "github.com/xtaci/kcp-go/v5"
-
-	tmp2p "github.com/tendermint/tendermint/proto/tendermint/p2p"
->>>>>>> notional/main-kcp
 )
 
 // EmptyNetAddress defines the string representation of an empty NetAddress
--- a/p2p/transport.go
+++ b/p2p/transport.go
@@ -9,19 +9,12 @@ import (
 	"golang.org/x/net/netutil"
 
 	"github.com/cosmos/gogoproto/proto"
-<<<<<<< HEAD
 
 	"github.com/cometbft/cometbft/crypto"
 	"github.com/cometbft/cometbft/libs/protoio"
 	"github.com/cometbft/cometbft/p2p/conn"
 	tmp2p "github.com/cometbft/cometbft/proto/tendermint/p2p"
-=======
-	"github.com/tendermint/tendermint/crypto"
-	"github.com/tendermint/tendermint/libs/protoio"
-	"github.com/tendermint/tendermint/p2p/conn"
-	tmp2p "github.com/tendermint/tendermint/proto/tendermint/p2p"
 	kcp "github.com/xtaci/kcp-go/v5"
->>>>>>> notional/main-kcp
 )
 
 const (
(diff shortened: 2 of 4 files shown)


p2p/netaddress.go
```diff
@@ -41,10 +42,10 @@ func IDAddressString(id ID, protocolHostPort string) string {
 // panic. Panics if ID is invalid.
 // TODO: socks proxies?
 func NewNetAddress(id ID, addr net.Addr) *NetAddress {
-	tcpAddr, ok := addr.(*net.TCPAddr)
+	tcpAddr, ok := addr.(*net.UDPAddr)
 	if !ok {
 		if flag.Lookup("test.v") == nil { // normal run
-			panic(fmt.Sprintf("Only TCPAddrs are supported. Got: %v", addr))
+			panic(fmt.Sprintf("Only UDPAddrs are supported. Got: %v", addr))
```
peterbourgon, line 48: If you want to add support for a novel transport-level protocol like KCP (which also switches the network-layer protocol from TCP to UDP) you can't just redefine a core function like NewNetAddress to use your alternative by default, you have to make it opt-in.

p2p/transport.go
```diff
@@ -248,7 +249,8 @@ func (mt *MultiplexTransport) Close() error {
 
 // Listen implements transportLifecycle.
 func (mt *MultiplexTransport) Listen(addr NetAddress) error {
-	ln, err := net.Listen("tcp", addr.DialString())
+
+	ln, err := kcp.Listen("0.0.0.0:26656")
 	if err != nil {
 		return err
 	}
```
peterbourgon, line 256: You're totally ignoring addr.

peterbourgon (changes requested):
CometBFT's P2P layer is not defined by this particular implementation, it is defined by [a specification](https://github.com/cometbft/cometbft/blob/main/spec/p2p/node.md) which explicitly specifies [P2P connections as TCP](https://github.com/cometbft/cometbft/blob/main/spec/p2p/connection.md). 

More generally, changes in fundamental subsystems like the p2p layer need to be motivated by well-defined requirements that are agreed upon by all stakeholders, and they need to substantiated by concrete evidence that the proposed changes satisfy those requirements. This PR provides neither agreed-upon requirements, nor evidence that proposed change satisfies those requirements.

faddat:
Oh yes totally that is all accurate sir. I put this out mainly as a proof
of concept and well dealing with ADDR sounds like a really good idea, and
when I first made this PR the idea was actually to support TCP,  quic, and
KCP.

But that was back in the days before comet you see.

Also it wasn't quite done I was just really surprised that anything worked
with it to be quite honest with you sir.

Is this an interesting direction? Or is it a potshot?

Thank you for your review

Issue #592, opened 2023-03-27
ADR 102: RPC Companion architecture


thanethomson:
As part of #574, we want to offer users a standalone proof-of-concept RPC service to serve as an example of a data companion. To this end, we want an ADR that captures:

1. The overall architecture of our standalone RPC service (implemented as a data companion)
2. Which of our [RPC endpoints](https://docs.cometbft.com/v0.34/rpc/) we'll be exposing as `/v1`
3. The decision as to which database to use to facilitate (1). The best database here will depend on the data access patterns defined by (1).

Future efforts could possibly involve productionizing this to the point that it may be able to scale independently of a full node, possibly reducing the costs associated with providing an RPC as a service, but this work needs to be broken down in separate follow-up issues.

jmalicevic:
I am wondering whether the access patterns will depend on the endpoints exposed? If we have a preliminary list, I could make sure to gather the info on access patterns during the storage investigation.

andynog:
I've started adding information about the RPC endpoints that will be implemented. I've used a table format because I believe it's easier to understand. There are a lot more info to add so it's WIP. For each endpoint I have to dig in into the code to understand it a bit better to assess the complexities on implement them. I also plan to add more sections outlining on how things will work, e.g. translating queries or return a response mapping the database records to a struct (or protobuf).

PR #658, opened 2023-04-05
docs: ADR-102 - RPC Companion

andynog:
[Rendered](https://github.com/cometbft/cometbft/blob/andy/adr-102-standalone-rpc/docs/architecture/adr-102-rpc-companion.md)

closes: #592

This is a WIP document for the RPC Companion ADR


docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,344 @@
...
+## Context
+
+This ADR proposes an architecture of a ***RPC Companion*** solution implemented based on the proposed [ADR-101 Data Companion Pull API](https://github.com/cometbft/cometbft/blob/thane/adr-084-data-companion-pull-api/docs/architecture/adr-101-data-companion-pull-api.md). This solution can run as a sidecar concurrently with the full node, and it is optional.
+
+This ADR provides a reference implementation of a system that can be used to offload queryable data from a CometBFT
+full node to a database and offering a service that exposes the same JSONRPC methods on a endpoint as the regular JSONRPC methods of an CometBFT node endpoint,
+which makes it easier for integrators of RPC clients such as client libraries and applications to switch to this
+***RPC Companion*** with as minimum effort as possible.
+
+This architecture also make it possible to scale horizontally the querying capacity of a full node by running multiple
```
jmalicevic, line 23:
```suggestion
This architecture also makes it possible to scale horizontally the querying capacity of a full node by running multiple
```

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,344 @@
...
+### High-level architecture
+
+![High-level architecture](images/adr-102-architecture.jpeg)
+
+This diagram shows all the required components for a full RPC Companion solution. The solution implementation contains
+many parts and each one is described below:
+
+### [Ingest Service](#ingest-service)
+
+The **Ingest Service** pulls the data from the full node RPC endpoint and store the information retrieved in
```
jmalicevic, line 146:
```suggestion
The **Ingest Service** pulls the data from the full node RPC endpoint and stores the information retrieved in
```

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,344 @@
...
+
+Another potential solution for this scenario is to find a way in the database that can support "versioning" of data structures. For example, let's
+assume there's a `Block` structure, let's call it `v1`. If in the future there's a need to modify this structure that is not compatible with the
+previous data structure, then the database would support a `v2` schema for `ResultBlock` and an `index` table could determine the criteria on which
+data structure should be used for inserting or querying data.
+
+There's also a possibility that the data structure could be stored in parallel with the normalized data but in one blob, e.g. in a `jsonb` field
+that contains all the information needed to return the client, this could be an optimized way to serve the data but could add to the storage
+requirements. This could improve the performance for reading data from the database. An alternative approach to this problem that can offer
+similar results would be to use a "caching" layer.
```
jmalicevic, line 291: Thanks Andy for explaining all these options. My only worry with separate versioned data is support for queries (somehow you'd need to query more than one table (or not? ) etc.) Maybe we could do the first thing proposed (copy the old one into the new structure), but add a version column and perhaps do something with the fields that differ on query response? (like if its `v1` and the field is `null` put a default value, or remove it ? 

Also, note that returning `nil` in the fields that did not exist before might be needed, because if people want to query data both in the old format and new format, they can't receive entries in different formats, its either one or the other. We could support both but as a user, I'd find that harder to manage. I'd rather have everything in one format with a version field indicating that I can then do some extra parsing after I get the results.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,344 @@
...
+
+- Alternative and optional **RPC Companion** that is more scalable and reliable with a higher query throughput.
+- Less backpressure on the full node that is running consensus.
+- Possibility for future additional endpoints (v2).
+- Allow users to create better and faster indexers and analytics solutions.
+
+### Negative
+
+- Additional infrastructure complexity to set up and maintain.
+- Additional infrastructure cost.
```
jmalicevic, line 329: This might actually not be only a cost because it should save on some storage (especially if we port the indexes), and CPU cost when querying the full nodes whose resources we were taking up until now. Of course it will cost more but maybe not *that* much more.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -40,109 +40,128 @@ TBD
...
-If the **ingest service** becomes unavailable (e.g. stops), then it should resume synchronization with the full node when it is back online.
-The **ingest service** should query the full node for the last `retain height` and the **ingest service** should request
+If the **Ingest Service** becomes unavailable (e.g. stops), then it should resume synchronization with the full node when it is back online.
+The **Ingest Service** should query the full node for the last `retain height` and the **Ingest Service** should request
 and process all the heights missing on the database until it catches up with the full node latest height.
 
-In case the **ingest service** becomes unavailable for a long time and there are a lot of height be caught up, it is
-important for the **ingest service** to do it in a throttled way in order not to stress the full node and cause issues in its consensus processing.
+In case the **Ingest Service** becomes unavailable for a long time and there are several heights to be synchronized, it is
+important for the **Ingest Service** to do it in a throttled way in order not to stress the full node and cause issues in its consensus processing.
```
lasarojc, line 162: "consensus processing" sounds weird to me. What about 
```suggestion
important for the **Ingest Service** to do it in a throttled way in order not to stress the full node and hinder block generation.
```

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -186,16 +205,74 @@ func (db *GoLevelDB) Set(key []byte, value []byte) error {
...
+) TABLESPACE pg_default;
+```
+
+Additional tables will be required to persist normalized data related to the block structure. All database schema will be available in the `rpc-companion` Github repository under the `/database/schema` folder.
+
 ##### Data types
 
-This solution will implement a data schema in the database using its built-in data types. By using a relation database, there's a possibility
-to better normalize the data structures, this might provide savings in storage but might add to the complexity of returning a particular
+This solution will implement a data schema in the database using Postgresql built-in data types. By using a relational database, there's a possibility
```
lasarojc, line 250:
```suggestion
This solution will implement a data schema in the database using PostgreSQL built-in data types. By using a relational database, there's a possibility
```

lasarojc: search and replace

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,344 @@
...
+| `abci_query`         | * path <br/> * data <br/> * height <br/> * prove                     | Query information from the application | This companion endpoint will return the same response structure as the CometBFT endpoint. The RPC companion service will have to implement a proper abci parameter to sql query translation  |
+| `block_search`       | * query <br/> * page <br/> * per_page <br/> * order_by               | Query information about a block        | This companion endpoint will return the same response structure as the CometBFT endpoint. The RPC companion service will have to implement a proper query parameter to sql query translation |
+| `tx_search`          | * query <br/> * page <br/> * per_page <br/> * prove <br/> * order_by | Query information about transactions   | This companion endpoint will return the same response structure as the CometBFT endpoint. The RPC companion service will have to implement a proper query parameter to sql query translation |
+
+These endpoints will proxy the request to the full node. Since they are not dependent on data retrieval from the RPC Companion database they should just act as proxies to the full node.
```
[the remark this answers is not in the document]
andynog: This is something we can evaluate. I think once I have more data that should be indexed (e.g. events) in the database we will see what's the best way. I think the idea is that if we can create indexes in the database and all the data is there then we could offload this to the Companion.

docs/architecture/adr-102-rpc-companion.md
andynog: We might need to thinks this through and see what's the best way to approach this. I'm wondering how this might have been handled in CometBFT in the past, like the data structures changing and how that was handled from the RPC perspective.

docs/architecture/adr-102-rpc-companion.md
andynog: Maybe I should be more specific on infrastructure cost. I was thinking mostly the whole setup if using multiple servers for the RPC Companion service, also the database depending on the setup (master/replica), load balancer, etc. But agree, some of the cost might be offset if the data is properly stored and normalized.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -40,109 +40,128 @@ TBD
...
 
-The **Standalone RPC** solution shall:
+The **RPC Companion** solution shall:
 
-1. Provide an **ingestion service** implemented as a data companion that can pull data from the node and store it on
+1. Provide an **[Ingest Service](#ingest-service)** implemented as a data companion that can pull data from the node and store it on
 its own storage (database)
-2. Provide its own storage (database) that can handle a high-load of reads and also have a good performance on inserting data.
-3. Implement a schema that needs to be backwards compatible with the current CometBFT RPC.
+2. Provide its own storage ([Database](#database)) that can handle a high-load of reads and also have a good performance on inserting data.
```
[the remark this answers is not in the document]
andynog: The idea here from a database perspective is that the ratio of read/writes is much higher, since the ingest will insert only a every 5 or so seconds but if load balanced so many reads.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -40,109 +40,128 @@ TBD
...
+3. Implement a [database schema](#database-schema) that needs to be backwards compatible with the current CometBFT RPC.
```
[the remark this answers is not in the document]
andynog: actually I meant to say it would be compatible with Comet data structures

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -40,109 +40,128 @@ TBD
...
-
-> NOTE: The RPC Companion server instances should not implement endpoints that can modify state in the blockchain such as
-the `/broadcast_tx_*` endpoints. Since there might be many load balanced RPC server instances, this might cause issues with
-transactions, for example sequential transactions might be relayed in the wrong order causing the full node to reject some
-transactions with sequence mismatch errors. These requests will proxy to the full node as outlined above.
+| **JSONRPC method**     | *JSONRPC Parameters** | **Description**                           | **Notes**                                                                                                                                                                                                                                       | Proxy |
```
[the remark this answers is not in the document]
thanethomson: This is outlined on line 110. The JSON-RPC interface is not something that's easy to proxy - all method calls happen via POST requests to a single endpoint.

If we don't proxy these methods, then clients need to know which endpoints to target for which requests, which complicates client architecture substantially. There is no way afaik with software like Nginx to proxy such requests, since routing would depend on data in the body of the message.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -40,109 +40,128 @@ TBD
...
+This diagram shows all the required components for a full RPC Companion solution. The solution implementation contains
 many parts and each one is described below:
 
-### Ingest Service
+### [Ingest Service](#ingest-service)
 
-The **ingest service** pulls the data from the full node via its RPC endpoints and store the information retrieved in
-the database. In the future, if a gRPC interface is implemented in the full node this might be used to pull the data
+The **Ingest Service** pulls the data from the full node RPC endpoint and store the information retrieved in
+the RPC Companion database.
```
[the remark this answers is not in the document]
thanethomson: The ingest service is an example/instance of a data companion. Perhaps it'd clarify things to make that a bit more explicit? But just calling it a "data companion" may be confusing.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -40,109 +40,128 @@ TBD
...
+
+| **JSONRPC method**   | **JSONRPC Parameters**                 | **Description**                                 | **Notes**                                                                                                                                                                                                                                                                               |
+|----------------------|----------------------------------------|-------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `abci_info`          |                                        | Get information about the application           | This companion endpoint will return the same response structure as the CometBFT endpoint. The companion endpoint will return the latest information stored in its database that was retrieved from the full node.                                                                       |
```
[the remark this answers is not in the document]
thanethomson: Yes, technically the JSON-RPC API is only accessed via a single endpoint. Every other facet here is a method, not an endpoint.

docs/architecture/adr-102-rpc-companion.md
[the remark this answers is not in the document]
mzabaluev: Is it possible to switch to REST style in this API? That is, the endpoints will be selected by the query string; for "write" methods, the request method will still be POST, while read methods would be as GET.

This will enable HTTP request routing with commonly used reverse proxy/load balancer solutions.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+## Status
+
+Accepted | Rejected | Deprecated | Superseded by
+
+## Context
+
+This ADR proposes an architecture of an ***RPC Companion*** solution implemented based on the proposed [ADR-101 Data Companion Pull API](https://github.com/cometbft/cometbft/blob/thane/adr-084-data-companion-pull-api/docs/architecture/adr-101-data-companion-pull-api.md). This solution can run as a sidecar concurrently with the full node, and it is optional.
+
+This ADR provides a reference implementation of a system that can be used to offload queryable data from a CometBFT
+full node to a database and offering a service that exposes the same JSONRPC methods on a endpoint as the regular JSONRPC methods of a CometBFT node endpoint,
```
cason, line 16:
```suggestion
full node to a database and offer a service exposing the same JSONRPC methods on a endpoint as the regular JSONRPC methods of a CometBFT node endpoint,
```

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+
+- 2022-03-27: First draft (@andynog)
+
+## Status
+
+Accepted | Rejected | Deprecated | Superseded by
+
+## Context
+
+This ADR proposes an architecture of an ***RPC Companion*** solution implemented based on the proposed [ADR-101 Data Companion Pull API](https://github.com/cometbft/cometbft/blob/thane/adr-084-data-companion-pull-api/docs/architecture/adr-101-data-companion-pull-api.md). This solution can run as a sidecar concurrently with the full node, and it is optional.
```
cason, line 13: A sidecar here is a separate process?

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+
+This ADR provides a reference implementation of a system that can be used to offload queryable data from a CometBFT
+full node to a database and offering a service that exposes the same JSONRPC methods on a endpoint as the regular JSONRPC methods of a CometBFT node endpoint,
+which makes it easier for integrators of RPC clients such as client libraries and applications to switch to this
+***RPC Companion*** with as minimum effort as possible.
+
+This architecture also makes it possible to scale horizontally the querying capacity of a full node by running multiple
+copies of the ***RPC Companion*** server instances that can be behind a scalable load-balancer (e.g. Cloudflare)  and can serve
+the data in a more reliable way.
```
cason, line 22: Why reliable?

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+
+## Alternative Approaches
+
+Currently, there aren't any alternative solutions that are based on a data companion architecture. It is
```
cason, line 26: I would rephrase that in a way that makes it clear that ADR 101 is a new feature that will potentially be adopted by users once it is well-established. And that this ADR proposes a first use-case for this feature.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+### Requirements
+
+The target audience for this solution are operators and integrators that want to alleviate the load on their nodes by offloading
+the queryable data requests to the **RPC Companion**.
+
+This solution shall meet the following requirements in order to provide real benefits to these users.
+
+The **RPC Companion** solution shall:
+
+1. Provide an **[Ingest Service](#ingest-service)** implemented as a data companion that can pull data from the node and store it on
```
cason, line 46: Add a reference to the data companion ADR.

cason, line 46:
```suggestion
1. Provide an **[Ingest Service](#ingest-service)** implemented as a data companion that can pull data from a CometBFT node and store it on
```

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+its own storage (database)
+2. Provide its own storage ([Database](#database)) that can handle a high-load of reads and also have a good performance on inserting data.
```
cason, line 48:
```suggestion
2. Provide a storage ([Database](#database)) that can handle a high-load of reads and also have a good performance on inserting data.
```

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+3. Implement a [database schema](#database-schema) that needs to be backwards compatible with the current CometBFT RPC.
```
cason, line 49:
```suggestion
3. Implement a [database schema](#database-schema) to store the pulled data that is backwards compatible with the current CometBFT RPC.
```

docs/architecture/adr-102-rpc-companion.md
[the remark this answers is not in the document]
cason: Agree with Lasaro here. We need to ensure that the same data available through RPC can also be found on the database, using the same schema (data structures).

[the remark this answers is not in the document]
cason: In fact, it doesn't matter how we store data, provide that we can provide the same interface as the CometBFT RPC interface.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+4. Do not enforce any breaking changes to the existing RPC.
+5. Ensure the responses returned by the [RPC Companion v1 endpoint](#rpc-endpoint) is wire compatible with the existing CometBFT endpoint.
```
cason, line 51:
```suggestion
5. Ensure the responses returned by the [RPC Companion v1 endpoint](#rpc-endpoint) is wire compatible with the existing CometBFT RPC endpoint.
```

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+6. Implement tests to verify backwards compatibility.
+7. Be able to handle multiple concurrent requests and return idempotent responses.
```
cason, line 53: Idempotent = Independent?

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+8. Do not crash or panic if the querying demand is very high or large responses are returned.
```
cason, line 54: This requirement is... odd.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+9. Storage optimized for an access pattern of faster read throughput than write.
```
cason, line 55:
```suggestion
9. Be optimized for an access pattern of faster read throughput than write throughput.
```

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+10. Provide metrics to allow operators to properly monitor the services and infrastructure.
+
+It is ***NOT*** in the scope of the **RPC Companion**:
+
+1. Provide an authentication mechanism for the RPC Companion endpoint.
+
+### [RPC Endpoint](#rpc-endpoint)
+
+The RPC Companion endpoint will be the same as the CometBFT endpoint but with a `/v1` appended to it. The RPC Companion endpoint
```
cason, line 64:
```suggestion
The RPC Companion endpoint will be the same as the CometBFT RPC endpoint but with a `/v1` appended to it. The RPC Companion endpoint
```

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+might also use a different port than the default CometBFT RPC port (e.g. `26657`).
```
cason, line 65: This second part is about network addresses, right? The address can be different, so it can be the ports.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+
+For example, suppose these are the URLs for each RPC endpoint:
+
+CometRPC -> `http://cosmos.host:26657`
+
+RPC Companion -> `http://rpc-companion.host:8080/v1`
+
+To make a request for a `block` at height `5` using the CometBFT endpoint:
+
+`curl --header "Content-Type: application/json" --request POST --data '{"method": "block", "params": ["5"], "id": 1}' http://cosmos.host:26657`
```
cason, line 75: Suggestion: replace the `` around the command by a tab in the beginning of the line. This produces a code block formatting.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+
+The make the same request to the RPC Companion endpoint:
+
+`curl --header "Content-Type: application/json" --request POST --data '{"method": "block", "params": ["5"], "id": 1}' http://rpc-companion.host:8080/v1`
+
+> Note that only the URL changes between these two `curl` commands
+
+The RPC Companion will accept JSONRPC requests, the same way as the CometBFT RPC endpoint does.
+
+The methods for the RPC Companion endpoint on the following table are the ones that can be implemented first since it seems it will be straightforward and less complex
```
cason, line 85: Did you build this table from scratch, or is it present elsewhere in the documentation? If the second case, it would be good to reference the original table (if one). In any case, it would be good to reference when the reader can find a full description of the methods.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+|----------------------|----------------------------------------|-------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `abci_info`          |                                        | Get information about the application           | This companion endpoint will return the same response structure as the CometBFT endpoint. The companion endpoint will return the latest information stored in its database that was retrieved from the full node.                                                                       |
+| `block`              | * height                               | Get block at a specified height                 | This companion endpoint will return the same response structure as the CometBFT endpoint. The data retrieved from the companion database for a particular block will have to be properly serialized into the `block` struct in order to be returned as a response.                      |
```
cason, line 90: It appears that all methods in this table will return the same response as the CometBFT RPC corresponding endpoint. Maybe it is good do add this information to the table caption, instead of repeating it on each line.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -40,109 +40,128 @@ TBD
...
+| `abci_query`         | * path <br/> * data <br/> * height <br/> * prove                     | Query information from the application | This companion endpoint will return the same response structure as the CometBFT endpoint. The RPC companion service will have to implement a proper abci parameter to sql query translation  |
+| `block_search`       | * query <br/> * page <br/> * per_page <br/> * order_by               | Query information about a block        | This companion endpoint will return the same response structure as the CometBFT endpoint. The RPC companion service will have to implement a proper query parameter to sql query translation |
+| `tx_search`          | * query <br/> * page <br/> * per_page <br/> * prove <br/> * order_by | Query information about transactions   | This companion endpoint will return the same response structure as the CometBFT endpoint. The RPC companion service will have to implement a proper query parameter to sql query translation |
+
+These endpoints will proxy the request to the full node. Since they are not dependent on data retrieval from the RPC Companion database they should just act as proxies to the full node.
```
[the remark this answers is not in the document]
cason: "These" here refers to the methods in the following table, right?

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+The **Ingest Service** should query the full node for the last `retain height` and the **Ingest Service** should request
+and process all the heights missing on the database until it catches up with the full node latest height.
+
+In case the **Ingest Service** becomes unavailable for a long time and there are several heights to be synchronized, it is
+important for the **Ingest Service** to do it in a throttled way in order not to stress the full node and cause issues in its consensus processing.
+
+### [Database](#database)
+
+The database stores the data retrieved from the full node and provide this data for the RPC server instance. Since the frequency
+that blocks are generated on the chain are in the range from 5 seconds to 7 seconds on average, the _write_ back pressure is not
```
cason, line 164: This data (5-7 seconds) is from which network? I would make this affirmation less network-dependent, while providing this data (5-7s) as an example of a specific network. Moreover, this is the _period_ not the _frequency_.

For instance, one can say that the frequency is not more than 2 blocks per second, which would lead to 0.5s period, and more likely in the order of 8-12 blocks per minute?

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+very high from a modern database perspective. While the frequency and number of requests for reading the data from the database will
```
cason, line 165:
```suggestion
very high from a modern database perspective. While the frequency and number of requests for reading the data from the database can
```

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+be much larger due to the fact that the RPC service instance can be scaled. Therefore, a database that provides a high read
+throughput should be favored.
+
+For this is initial solution implementation it is proposed that the relational database [Postgresql](https://www.postgresql.org/) should be used in order to support
+the RPC scalability and this will also provide more flexibility when implementing the RPC v2 endpoints that can return data
```
cason, line 170: I would break this sentence before "and".

cason: Also, this is the first mention to v2 endpoints.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+in different forms and database indexes might also be leveraged in order to boost the query responses performance.
+
+The data needs to be available both for the Ingest Service (writes) and the RPC server instance (reads) and these service might be running from different machines so an embedded database
```
cason, line 173: Data or database?

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+is not recommended in this case since accessing the data remotely might not be optimal for an embedded key-value database. Also
+since the RPC might have many server instances (or processes) running that will need to retrieve data concurrently it is recommended to use
+a well-known robust database engine that can support such a load.
+
+Also, a database that can support ACID transactions is important to provide more guarantees that
```
cason, line 178: Actually, I think the most important point of having ACID transactions is to ensure that clients (RPC) don't read incomplete data, namely data that was only partially written to the database. Any database provides write guarantees, but if the write is made of multiple steps, readers can access incomplete data.

cason (commented):
Great write up.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+## Context
+
+This ADR proposes an architecture of an ***RPC Companion*** solution implemented based on the proposed [ADR-101 Data Companion Pull API](https://github.com/cometbft/cometbft/blob/thane/adr-084-data-companion-pull-api/docs/architecture/adr-101-data-companion-pull-api.md). This solution can run as a sidecar concurrently with the full node, and it is optional.
+
+This ADR provides a reference implementation of a system that can be used to offload queryable data from a CometBFT
+full node to a database and offering a service that exposes the same JSONRPC methods on a endpoint as the regular JSONRPC methods of a CometBFT node endpoint,
+which makes it easier for integrators of RPC clients such as client libraries and applications to switch to this
+***RPC Companion*** with as minimum effort as possible.
+
+This architecture also makes it possible to scale horizontally the querying capacity of a full node by running multiple
```
sergio-mena, line 20: If this is an important use case, maybe we should revisit the decision we made in ADR101 about the data companion interface only supporting one client. If horizontal scaling becomes the norm, the fact that all RPC-companion clients have to coordinate to "act as one single data-companion client" might entail a lot of overhead.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+
+## Changelog
+
+- 2022-03-27: First draft (@andynog)
+
+## Status
+
+Accepted | Rejected | Deprecated | Superseded by
+
+## Context
```
sergio-mena, line 11: Is it worth mentioning that one advantage of this setup is that the full node can turn off indexing (which is a known source of contention and performance bottleneck)?

docs/architecture/adr-102-rpc-companion.md
sergio-mena, line 48:
```suggestion
2. Provide its own storage ([Database](#database)) that can handle a high-load of reads and also have a good performance on inserting data concurrently to the reads.
```

docs/architecture/adr-102-rpc-companion.md
sergio-mena: I would just say "Be able to handle multiple concurrent requests correctly"

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+
+CometRPC -> `http://cosmos.host:26657`
+
+RPC Companion -> `http://rpc-companion.host:8080/v1`
+
+To make a request for a `block` at height `5` using the CometBFT endpoint:
+
+`curl --header "Content-Type: application/json" --request POST --data '{"method": "block", "params": ["5"], "id": 1}' http://cosmos.host:26657`
+
+The make the same request to the RPC Companion endpoint:
```
sergio-mena, line 77:
```suggestion
To make the same request to the RPC Companion endpoint:
```

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+| `abci_query`         | * path <br/> * data <br/> * height <br/> * prove                     | Query information from the application | This companion endpoint will return the same response structure as the CometBFT endpoint. The RPC companion service will have to implement a proper abci parameter to sql query translation  |
+| `block_search`       | * query <br/> * page <br/> * per_page <br/> * order_by               | Query information about a block        | This companion endpoint will return the same response structure as the CometBFT endpoint. The RPC companion service will have to implement a proper query parameter to sql query translation |
+| `tx_search`          | * query <br/> * page <br/> * per_page <br/> * prove <br/> * order_by | Query information about transactions   | This companion endpoint will return the same response structure as the CometBFT endpoint. The RPC companion service will have to implement a proper query parameter to sql query translation |
```
sergio-mena, line 108: According to my understanding:

* offloading `block_search` and `tx_search` methods to the companion has the biggest potential of improving full node's performance
* other methods whose implementation can simply access the block store or state store directly are interesting from a storage perspective, but less impactful performance-wise

So would be very beneficial to have `block_search` and `tx_search` in the RPC companion. The problem I see for including them in an MVP (maybe that's why they are in this list?) is that, ideally, we would use the underlying database's SQL querying, rather than migrating our ad-hoc indexing code. If so, we couldn't offer _the same_ querying language to the RPC-companion's users as full nodes implement.

Is this discussion worth including in this ADR?

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+The **Ingest Service** pulls the data from the full node RPC endpoint and stores the information retrieved in
+the RPC Companion database.
+
+> In the future, if a gRPC interface is implemented in the full node this might be used to pull the data
+from the server.
+
+The **Ingest Service** can control the pruning on the full node via a mechanism to [pruning service](https://github.com/cometbft/cometbft/blob/thane/adr-084-data-companion-pull-api/docs/architecture/adr-101-data-companion-pull-api.md#pruning-service).
+Once the Ingest Service pulls the data from the full node and is able to process it, and it gets an acknowledgement from the database that the data was inserted,
+the **Ingest Service** can communicate with the full node notifying it that a specific height has been processed and set the processed
+height as the `retain height` on the full node signaling this way to the node that this height can be pruned.
```
sergio-mena, line 152: See my previous comment about horizontal scaling

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+If the **Ingest Service** becomes unavailable (e.g. stops), then it should resume synchronization with the full node when it is back online.
+The **Ingest Service** should query the full node for the last `retain height` and the **Ingest Service** should request
+and process all the heights missing on the database until it catches up with the full node latest height.
+
+In case the **Ingest Service** becomes unavailable for a long time and there are several heights to be synchronized, it is
+important for the **Ingest Service** to do it in a throttled way in order not to stress the full node and cause issues in its consensus processing.
+
+### [Database](#database)
+
+The database stores the data retrieved from the full node and provide this data for the RPC server instance. Since the frequency
```
sergio-mena, line 163:
```suggestion
The database stores the data retrieved from the full node and provides this data for the RPC server instance. Since the frequency
```

docs/architecture/adr-102-rpc-companion.md
sergio-mena: I'm just thinking out loud... the Crypto.com folks implemented the SQL sink (AFAIK), which is backed by PostgreSQL, is there something there we can reuse?
Check [this dir](https://github.com/cometbft/cometbft/blob/b7c3e9964/state/indexer/sink/psql)

docs/architecture/adr-102-rpc-companion.md
sergio-mena: Adding to this, by reading further in the text, I understand that the horizontal scaling won't (in principle) affect the Ingest service, which will still be only one instance write to the DB, from which a horizontally-scaled RPC servers will read.

If that is the case, consider explaining this somewhere in the ADR

docs/architecture/adr-102-rpc-companion.md
thanethomson: Yes. That should probably be emphasized here.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+```
+
+Since the RPC Companion stores the information in a relational database, there are opportunities to better structure and normalize the data. Here is the schema definition for a table to persist a normalized `ResultBlock` data structure in the database (Postgresql)
+
+```sql
+-- Table: comet.result_block
+
+CREATE TABLE IF NOT EXISTS comet.result_block
+(
+    block_id_hash bytea NOT NULL,
```
sergio-mena, line 212: Shouldn't we be using foreign keys for this kind of fields?

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+    block_header_proposer_address bytea NOT NULL,
+    block_header_last_block_id_hash bytea NOT NULL,
+    block_header_last_block_id_parts_hash bytea NOT NULL,
+    block_header_last_block_id_part_total comet.uint32 NOT NULL,
+    block_last_commit_height comet.uint64 NOT NULL,
+    block_last_commit_round comet.uint32 NOT NULL,
+    block_last_commit_block_id_hash bytea NOT NULL,
+    block_last_commit_block_id_parts_total comet.uint32 NOT NULL,
+    block_last_commit_block_id_parts_hash bytea NOT NULL,
+    CONSTRAINT block_pkey PRIMARY KEY (block_header_height),
```
sergio-mena, line 237: Should we make chainID part of the primary key, in case an operator decides to reuse the same DB for more than one chains?

docs/architecture/adr-102-rpc-companion.md
thanethomson: Yes, I suspect we should remove this requirement. Not sure it's implementable.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+    block_header_next_validators_hash bytea NOT NULL,
+    block_header_consensus_hash bytea NOT NULL,
+    block_header_app_hash bytea NOT NULL,
+    block_header_last_results_hash bytea NOT NULL,
+    block_header_evidence_hash bytea NOT NULL,
+    block_header_proposer_address bytea NOT NULL,
+    block_header_last_block_id_hash bytea NOT NULL,
+    block_header_last_block_id_parts_hash bytea NOT NULL,
+    block_header_last_block_id_part_total comet.uint32 NOT NULL,
+    block_last_commit_height comet.uint64 NOT NULL,
```
sergio-mena, line 232: This should probably also be a (null-able) foreign key to the same table.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+##### Data types
+
+This solution will implement a data schema in the database using Postgresql built-in data types. By using a relational database, there's a possibility
+to better normalize the data structures, this might provide savings in storage but might also add to the complexity of returning a particular
+dataset because the data joins that will be required. Also, it would be important ensure the referential integrity is not violated since
+this can cause issues to the clients consuming the data.
+
+In order to accurately ensure the database is storing the full node data types properly, the database might implement custom data types (`domains` in Postgresql).
+
+For example, Postgresql doesn't have an unsigned `uint64` datatype, therefore in order to support this in the database, you can use a `domain`, which is a base types with additional constrains. For example, this is the definition of a `uint64` domain:
```
sergio-mena, line 254:
```suggestion
For example, Postgresql doesn't have an unsigned `uint64` datatype, therefore in order to support this in the database, you can use a `domain`, which is a base type with additional constrains. For example, this is the definition of a `uint64` domain:
```

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+
+##### Schema migration
+
+Another point to consider is when the data structures change across CometBFT releases. There are a couple of ways to provide a solution to this
+problem.
+
+One way is to support a mechanism to migrate the old data to the new data structures. This would require additional logic for the migration process
+that would need to be run before ingesting data from an upgraded full node that contains the new data structures. But this approach might also
+cause some issues, for example, if the new data structure has a new field that there's no corresponding value in the "old" data structure, probably
+the value would need to be set to `[null]` but this can have unintended consequences.
```
sergio-mena, line 278: IMO, this is the responsibilty of the designers of the new feature, or the fix that caused the changes.

All the instances of this I can remember (in the last 18 months) have been well thought out, as the same problem exists in the current block store, state store, and indexing DBs. So, probably, the design of a fix or a new feature, once this is out, should also include how the PostgreSQL DB should handle the change.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+
+Another potential solution for this scenario is to find a way in the database that can support "versioning" of data structures. For example, let's
+assume there's a `Block` structure, let's call it `v1`. If in the future there's a need to modify this structure that is not compatible with the
+previous data structure, then the database would support a `v2` schema for `ResultBlock` and an `index` table could determine the criteria on which
+data structure should be used for inserting or querying data.
```
sergio-mena, line 283: I like this approach, as it is close to what we have in mind for RPC, which is what users will see.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+### RPC server instance
+
+The **RPC server instance** is a node that runs the RPC API process for the data companion. This server instance provides an RPC API (v1) with
+the same JSONRPC methods in the full node endpoint. The RPC Companion service will expose the same JSONRPC methods and will accept the same request types and
+return wire compatible responses (should match the same response as the equivalent full node RPC endpoint).
+
+The **RPC server instance**, when serving a particular request, retrieves the required data from the database in order to
+fulfill the request. The data should be serialized in a way that makes it wire compatible with the CometBFT RPC endpoint.
+
+Identical requests should return idempotent responses, no side effects should cause the RPC service to return different responses.
```
sergio-mena, line 300: This is not true for all methods (e.g. methods that have height as param, but the param is left blank, should retrieve the data for the latest height)

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,341 @@
...
+
+
+## Consequences
+
+### Positive
+
+- Alternative and optional **RPC Companion** that is more scalable and reliable with a higher query throughput.
+- Less backpressure on the full node that is running consensus.
+- Possibility for future additional endpoints (v2).
+- Allow users to create better and faster indexers and analytics solutions.
```
sergio-mena, line 321:
```suggestion
- Can act as a basis for users to create better and faster indexers and analytics solutions.
```

docs/architecture/adr-102-rpc-companion.md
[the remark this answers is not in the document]
lasarojc: Ok. My point is that "good performance" is too abstract and claiming you met the requirement in the end will be subjective.

docs/architecture/adr-102-rpc-companion.md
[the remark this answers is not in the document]
lasarojc: Yes, I think it would benefit from stating that the ingest service is (or will be) an instance of the data companion.

docs/architecture/adr-102-rpc-companion.md
lasarojc: Good point.
Also, while requests can be idempotent I don't think it makes sense to talk about "idempotent responses". 
Maybe "requests are idempotent, even if some parameters are determined during the request processing itself" or something like it.

docs/architecture/adr-102-rpc-companion.md
sergio-mena: Agreed. Probably we could include this somewhere in this ADR's text

docs/architecture/adr-102-rpc-companion.md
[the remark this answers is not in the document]
thanethomson: > Is it possible to switch to REST style in this API? That is, the endpoints will be selected by the query string; for "write" methods, the request method will still be POST, while read methods would be as GET.

We're aiming to not change the API in the first iteration of the RPC companion (accessible via the `/v1` endpoint) - we're aiming to provide as much compatibility with the existing RPC interface as possible to prevent client breakage.

We're of course free to do whatever we want with a `/v2`, but that's outside the scope of this work.

docs/architecture/adr-102-rpc-companion.md
thanethomson: @andynog, I suspect what @cason means here is that the word "reliable" has some strong implications in certain contexts.

But my understanding of it here is that our existing RPC is flaky, and many different applications (e.g. relayers, wallets, light clients) rely on the RPC for their functionality. So the more reliable the RPC, the more reliable those add-on services can be.

docs/architecture/adr-102-rpc-companion.md
thanethomson: > I'm just thinking out loud... the Crypto.com folks implemented the SQL sink (AFAIK), which is backed by PostgreSQL, is there something there we can reuse?

@ancazamfir and @romac used the SQL DB schema defined by the PostgreSQL sink and found it to be very suboptimal for the purposes of the relayer.

docs/architecture/adr-102-rpc-companion.md
thanethomson: The DB may need to scale horizontally too, but that's something that's taken care of via PostgreSQL configuration.

And the ingest service should not have to scale - there should only ever be a single ingest service for the RPC companion service/cluster.

docs/architecture/adr-102-rpc-companion.md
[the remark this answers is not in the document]
andynog: makes sense, let me rephrase that

docs/architecture/adr-102-rpc-companion.md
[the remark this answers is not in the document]
andynog: yes, the Companion RPC interface (or a future gRPC) should provide wire compatible responses to the CometBFT own RPC interface. I think this can be re-worded better. Mixing the schema with RPC is indeed confusing. I'll try to make it more clear.

docs/architecture/adr-102-rpc-companion.md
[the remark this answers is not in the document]
andynog: Thanks, I'll go over all the comments again, they should be clear as Thane said that there's one endpoint and different methods. I'll reword if something is not giving that idea. Thanks.

docs/architecture/adr-102-rpc-companion.md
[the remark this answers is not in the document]
andynog: Yes, agree with Thane, for v1 we're not changing the interfaces/methods, the client should be able to call the Companion just changing the URL.

docs/architecture/adr-102-rpc-companion.md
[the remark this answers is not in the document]
andynog: added more clarity on that in https://github.com/cometbft/cometbft/pull/658/commits/8a036c28c1e13d20335513d69c62e72fad4606fd

docs/architecture/adr-102-rpc-companion.md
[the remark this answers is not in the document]
andynog: Added some information related to indexing data

docs/architecture/adr-102-rpc-companion.md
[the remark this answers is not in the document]
andynog: I removed the blurb about performance, it was very subjective and hard to justify through more concrete numbers

docs/architecture/adr-102-rpc-companion.md
andynog: Fixed

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -80,53 +83,56 @@ The make the same request to the RPC Companion endpoint:
...
+| `tx`                 | * hash <br/> * prove                   | Get a transaction by its hash                   | This method will return the same response structure as the equivalent CometBFT method.                                                                                                                                                                                                |
+| `validators`         | * height <br/> * page <br/> * per_page | Get validator set at a specified height         | This method will return the same response structure as the equivalent CometBFT method.                                                                                                                                                                                                |
+
+The following methods can also be implemented, but might require some additional effort and complexity to be implemented.
+These are mostly the ones that provides `search` and `query` functionalities. These methods will proxy the request to the
```
sergio-mena, line 106:
```suggestion
These are mostly the ones that provide `search` and `query` functionalities. These methods will proxy the request to the
```

docs/architecture/adr-102-rpc-companion.md
andynog: fixed

docs/architecture/adr-102-rpc-companion.md
andynog: re-worded

docs/architecture/adr-102-rpc-companion.md
andynog: fixed

docs/architecture/adr-102-rpc-companion.md
andynog: makes sense, removed

docs/architecture/adr-102-rpc-companion.md
andynog: re-worded

docs/architecture/adr-102-rpc-companion.md
andynog: re-worded

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -40,109 +40,128 @@ TBD
...
+| `tx`                 | * hash <br/> * prove                   | Get a transaction by its hash                   | This companion endpoint will return the same response structure as the CometBFT endpoint.                                                                                                                                                                                               |
+| `validators`         | * height <br/> * page <br/> * per_page | Get validator set at a specified height         | This companion endpoint will return the same response structure as the CometBFT endpoint.                                                                                                                                                                                               |
+
+The following endpoints can also be implemented, but might require some additional effort and complexity to be implemented. These are mostly the `search` and `query` endpoints.
```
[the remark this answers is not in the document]
andynog: resolved

docs/architecture/adr-102-rpc-companion.md
andynog: resolved

docs/architecture/adr-102-rpc-companion.md
andynog: reworded, thanks

docs/architecture/adr-102-rpc-companion.md
andynog: there's a reference on the document to the PR for ADR-101

docs/architecture/adr-102-rpc-companion.md
andynog: makes sense, thanks, re-wrote it.

docs/architecture/adr-102-rpc-companion.md
andynog: added, thanks

docs/architecture/adr-102-rpc-companion.md
andynog: This comment is along the lines with Jasimna comments and I think this is probably a totally different discussion (or even an ADR). There are still many unknowns to validate before we can attest that this idea will work. I agree and think that the "search" and "query" would be the most beneficial but requires some effort to get it to work (if feasible). So I'll just leave the additional conversation to the near future but not part of this ADR

docs/architecture/adr-102-rpc-companion.md
andynog: Yes, Thane is right. One Ingest / Multiple RPC if needed. I'll try to emphasize that the ingest service should be a "singleton"

docs/architecture/adr-102-rpc-companion.md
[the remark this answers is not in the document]
andynog: I've rewrote this a bit. This excerpt was to emphasize that reads might occur more than writes (assuming there are a lot of queries to the server)

docs/architecture/adr-102-rpc-companion.md
andynog: that's a good point but in our design we assume one database per chain, so if they want to have a RPC Companion for another chain they would need to store the data in a separate database

docs/architecture/adr-102-rpc-companion.md
andynog: the schema is still evolving, this is the latest schema and it might change. The comments on schema are helpful but I'd think they would be worth it to have in the PR for the rpc-companion when it's submitted.

docs/architecture/adr-102-rpc-companion.md
andynog: same comment as above as schema specific feedback

docs/architecture/adr-102-rpc-companion.md
andynog: makes sense, added that information.

docs/architecture/adr-102-rpc-companion.md
andynog: agree

docs/architecture/adr-102-rpc-companion.md
andynog: I've just removed this section for now, it was causing more confusion than benefits. I think it's OK to leave it out.

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,358 @@
...
+
+The **RPC server instance**, when serving a particular request, retrieves the required data from the database in order to
+fulfill the request. The data should be serialized in a way that makes it wire compatible with the CometBFT JSONRPC endpoint.
+
+The RPC service endpoint should be exposed through an external load-balancer service such as Cloudflare or AWS ELB, or
+a server running its own load balancer mechanism (e.g. nginx).
+
+The RPC clients should make requests to the **RPC Companion** server instances through this load balancer.
+
+The **RPC server instance** will also implement logic for the proxied requests to the full node. It should properly handle the proxy requests and responses from the full node.
```
melekes, line 322: Let's also mention that the `RPC server instance` should cache responses from the full node to avoid requesting the same data over and over again. Wdyt?

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,358 @@
...
+
+## Consequences
+
+### Positive
+
+- Alternative and optional **RPC Companion** that is more scalable and reliable with a higher query throughput.
+- Less backpressure on the full node that is running consensus.
+- Possibility for future additional (e.g a `/v2`) with additional methods not available in the `/v1` endpoint.
+- Can act as a basis for users to create better and faster indexers and analytics solutions.
+- Possibility to turn off indexing on the full node if that can be offload to the RPC Companion.
```
melekes, line 337: RPC Companion or Data Companion?

docs/architecture/adr-102-rpc-companion.md
```diff
@@ -0,0 +1,358 @@
...
+This solution can run as a sidecar which is a separate process that runs concurrently with the full node, but the RPC Companion is optional meaning that the full node will still provide RPC services that can be queried directly if operators don't want to run
+an RPC Companion service.
+
+This ADR provides a reference implementation of a system that can be used to offload queryable data from a CometBFT
+full node to a database and offer a service exposing the same JSONRPC methods on a endpoint as the regular JSONRPC methods of a CometBFT node endpoint,
+which makes it easier for integrators of RPC clients such as client libraries and applications to switch to this
+***RPC Companion*** with as minimum effort as possible.
+
+This architecture also makes it possible to scale horizontally the querying capacity of a full node by running multiple
+copies of the ***RPC Companion*** server instances that can be behind a scalable load-balancer (e.g. Cloudflare) which
```
melekes, line 24: Would you happen to know how it's done right now? How do operators scale their API? Have we conducted any user interviews? Are there any issues from users related to the theme of this ADR?

melekes:
@andynog what's the timeline for this work? Should we merge this PR and resume once we have time?

andynog:
> @andynog what's the timeline for this work? Should we merge this PR and resume once we have time?

I started to look into this again and will take it to the finish line updating all the information needed since the ADR-101 Data Companion was implemented

docs/architecture/adr-102-rpc-companion.md
andynog: most operators might use a proxy server (Nginx) or a cloud load balancer (e.g. AWS ELB). Scaling the API is not just a matter of load balancing servers but eventually having a caching layer, geographical distribution, etc. The scalability topic is very complex and it is beyond the scope of this ADR in my opinion

docs/architecture/adr-102-rpc-companion.md
andynog: added additional text for clarification

docs/architecture/adr-102-rpc-companion.md
andynog: added clarification, thanks

docs/architecture/adr-102-rpc-companion.md
andynog: good tip, thanks

docs/architecture/adr-102-rpc-companion.md
andynog: it was built from scratch. We do have an RPC documentation in the docs.

docs/architecture/adr-102-rpc-companion.md
andynog: since we now have proto versioning, I like the idea of using the same versioning as the proto files since the db schema should match the proto structure. This might be the easiest way to handle data structure changes.

docs/architecture/adr-102-rpc-companion.md
andynog: added text for caching, thanks

docs/architecture/adr-102-rpc-companion.md
andynog: make it more clear that is an external data storage with indexing

adizere (approved):

PR #673, opened 2023-04-07
Fix race condition in Consensus code

### LINKED ISSUE #487: Intermittent data race in consensus.State during startup
When starting a set of validators using gaia v9, built with the race detector, and inside linux/amd64 Docker containers, I was seeing an intermittent but somewhat frequent data race. The top of the stack looked like this:

```
WARNING: DATA RACE
Write at 0x00c00029d9c8 by goroutine 67:
  github.com/tendermint/tendermint/consensus.(*State).reconstructLastCommit()
      github.com/tendermint/tendermint@v0.34.26/consensus/state.go:577 +0xb7
  github.com/tendermint/tendermint/consensus.(*Reactor).SwitchToConsensus()
      github.com/tendermint/tendermint@v0.34.26/consensus/reactor.go:113 +0xd7
  github.com/tendermint/tendermint/blockchain/v0.(*BlockchainReactor).poolRoutine()
      github.com/tendermint/tendermint@v0.34.26/blockchain/v0/reactor.go:324 +0x1a73
  github.com/tendermint/tendermint/blockchain/v0.(*BlockchainReactor).OnStart.func1()
      github.com/tendermint/tendermint@v0.34.26/blockchain/v0/reactor.go:112 +0x3b

Previous read at 0x00c00029d9c8 by goroutine 70:
  runtime.racereadrange()
      <autogenerated>:1 +0x1b
  github.com/tendermint/tendermint/consensus.(*Reactor).updateRoundStateRoutine()
      github.com/tendermint/tendermint@v0.34.26/consensus/reactor.go:546 +0x149
  github.com/tendermint/tendermint/consensus.(*Reactor).OnStart.func2()
      github.com/tendermint/tendermint@v0.34.26/consensus/reactor.go:82 +0x39
```

Version details are:

- gaia v9.0.0
- cosmos-sdk@v0.45.13-ics
- tendermint v0.34.26

The setup for starting the testnet in this configuration is a bit complicated, but luckily we saw a seemingly identical data race during CI on the Cosmos SDK: https://github.com/cosmos/cosmos-sdk/actions/runs/4295303107/jobs/7485538761#step:6:41

So then I attempted to reproduce that failure. It took more runs than I expected, but I have been able to reproduce it on a Linux docker image. It might reproduce on an M1 Mac, but I didn't see a repro within a few dozen runs. The Linux session looked like this:

```console
$ docker run --rm -it golang:1.19

root@b33bddfadfe0:/go# git clone --depth=1 -b v0.45.15-rc.0 https://github.com/cosmos/cosmos-sdk

root@b33bddfadfe0:/go# cd cosmos-sdk/testutil/network/

root@b33bddfadfe0:/go/cosmos-sdk/testutil/network# echo 'package network_test

import (
	"testing"

	"github.com/cosmos/cosmos-sdk/testutil/network"
	"github.com/stretchr/testify/require"
)

func TestNetwork(t *testing.T) {
	n := network.New(t, network.DefaultConfig())

	_, err := n.WaitForHeight(1)
	require.NoError(t, err)
}' > ./network_test.go
root@b33bddfadfe0:/go/cosmos-sdk/testutil/network# go test -race -c .

root@b33bddfadfe0:/go/cosmos-sdk/testutil/network#  i=0; while time ./network.test -test.count=1; do date; echo "$i"; i=$((i+1)); done; echo "failed after $i runs"
PASS

real	1m2.042s
user	0m49.385s
sys	0m22.292s
Thu Mar  2 21:41:22 UTC 2023
0
...
```

Full stack from testnet run: [gaia9-race.1.txt](https://github.com/cometbft/cometbft/files/10921689/gaia9-race.1.txt)

Full stack from a unit test repro run: 
[unit-test-repro-race.txt](https://github.com/cometbft/cometbft/files/10921804/unit-test-repro-race.txt)

Overall this data race looks about as innocuous as a data race can be. But, because it risks happening at startup, this risks interfering with searches for less innocuous data races in application chains. For example, using `GORACE=halt_on_error=1` is meant to stop execution as soon as the first data race is encountered; if this data race happens first, it is an inconvenience that the application will need to be restarted again in hopes of not encountering this data race.

sergio-mena:
Closes #487

This PR contains changes in the test infra to repro #487, and the the fix needed to avoid the race condition.
Similarly to #539, I have pushed the commits that make up this PR one by one, so that those interested can track how I reproed the race condition using e2e tests, and how the fix prevents the race from happening again.

* Commit#1: Add race option to e2e tests for builtin app. We also activate "halt on error", so e2e test are able to catch this (otherwise it just shows up in the logs, but the tests pass)
* Commit#2: The fix.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/test/e2e/Makefile
+++ b/test/e2e/Makefile
@@ -14,7 +14,7 @@ docker:
 # order to build a binary with a CometBFT node in it (for built-in
 # ABCI testing).
 node:
-	go build $(BUILD_FLAGS) -tags '$(BUILD_TAGS)' -o build/node ./node
+	go build -race $(BUILD_FLAGS) -tags '$(BUILD_TAGS)' -o build/node ./node
 
 generator:
 	go build -o build/generator ./generator
--- a/test/e2e/docker/Dockerfile
+++ b/test/e2e/docker/Dockerfile
@@ -25,6 +25,7 @@ RUN cd test/e2e && make node && cp build/node /usr/bin/app
 WORKDIR /cometbft
 VOLUME /cometbft
 ENV CMTHOME=/cometbft
+ENV GORACE "halt_on_error=1"
 
 EXPOSE 26656 26657 26660 6060
 ENTRYPOINT ["/usr/bin/entrypoint"]


consensus/reactor.go
```diff
@@ -107,14 +107,17 @@ func (conR *Reactor) OnStop() {
...
+	conR.conS.mtx.Lock()
+	// We have no votes, so reconstruct LastCommit from SeenCommit
 	if state.LastBlockHeight > 0 {
 		conR.conS.reconstructLastCommit(state)
 	}
 
 	// NOTE: The line below causes broadcastNewRoundStepRoutine() to broadcast a
 	// NewRoundStepMessage.
 	conR.conS.updateToState(state)
+	conR.conS.mtx.Unlock()
```
08d2, line 120: reconstructLastCommit and updateToState can panic, which will leave conS.mtx in a locked state, producing deadlock for any concurrent lockers.

thanethomson: @08d2, are you deeply familiar with the `SwitchToConsensus` function and where it's called? If so, please point out where, exactly (which files/lines of code), you see `conS.mtx` being locked concurrently during this specific `SwitchToConsensus` call.

sergio-mena: By design, CometBFT never tries to recover from a panic (crash failures always being preferable to general byzantine behaviour in both crash-tolerant and BFT algorithms), so, strictly speaking, this is a non-issue.
Nevertheless, I've changed the locking I had added to a RAII-style

thanethomson (approved):

consensus/reactor.go
08d2: > @08d2, are you deeply familiar with the `SwitchToConsensus` function and where it's called? If so, please point out where, exactly (which files/lines of code), you see `conS.mtx` being locked concurrently during this specific `SwitchToConsensus` call.

My mistake for qualifying the error as only for "concurrent callers". This was not accurate. If any caller, concurrent or otherwise, were to panic in the ways I described previously, the conS mutex would deadlock for all subsequent callers.

PR #692, opened 2023-04-11
Fix race condition in `gossipVotesRoutine`

### LINKED ISSUE #674: Race condition in `gossipVotesRoutine`
### Setup

**CometBFT version** : `main` (but present in all active releses)

**ABCI app** : e2e app, builtin

**Environment**:
- **OS** : linux/arm
- **Install tools**: e2e test

**node command runtime flags**: those for running e2e

### Config

N/A (extracted from manifest `test/e2e/networks/ci.toml`)

### What happened?

Race condition when building the builtin node with "-race"

No race

### How to reproduce it

* Add a for loop in function `gossipVotesRoutine` so that it accesses `ConsensusParams.ABCI.VoteExtensionsEnabled` many times
* Run e2e tests after building the docker image with "-race" option

### Logs

```
full02       | WARNING: DATA RACE
full02       | Write at 0x00c0002fe5d8 by goroutine 344:
full02       |   github.com/cometbft/cometbft/consensus.(*State).updateToState()
full02       |       github.com/cometbft/cometbft/consensus/state.go:727 +0x99c
full02       |   github.com/cometbft/cometbft/consensus.(*State).finalizeCommit()
full02       |       github.com/cometbft/cometbft/consensus/state.go:1758 +0xf94
full02       |   github.com/cometbft/cometbft/consensus.(*State).tryFinalizeCommit()
full02       |       github.com/cometbft/cometbft/consensus/state.go:1651 +0x3d0
[...]
full02       | Previous read at 0x00c0002fe5d8 by goroutine 217:
full02       |   github.com/cometbft/cometbft/consensus.(*Reactor).gossipVotesRoutine()
full02       |       github.com/cometbft/cometbft/consensus/reactor.go:749 +0xa6c
full02       |   github.com/cometbft/cometbft/consensus.(*Reactor).AddPeer.func2()
full02       |       github.com/cometbft/cometbft/consensus/reactor.go:201 +0x58
```

N/A

A patch is ready to test

sergio-mena:
Closes #674

This PR contains changes in the test infra to repro #674, and the the fix needed to avoid the race condition.
Similarly to #673 I have pushed the commits that make up this PR one by one, so that those interested can track how I repro'ed the race condition using e2e tests, and how the fix prevents the race from happening again. To do so, click on the little ❌ or ✅ next to the relevant commits.

* Commit#1: Add race option to e2e tests for builtin app. We also activate "halt on error", so e2e tests are able to catch this (otherwise it just shows up in the logs, but the tests pass).
* Commit#2 (pushed together with Commit#1): Added a for loop to increase the probability of the race condition happening.
* Commit#3: Commit#2 was enough to repro the race locally, but e2e tests passed in this PR for Commit#2. This commit is a second attempt at hitting the race condition. This time, however, almost all nodes are hitting it (check the corresponding ❌).
* Commit#4: Fix the race condition (1st version, see Commit#6). The first e2e run from this commit stalled (from the logs, it looks like the for introduced to exacerbate the race condition seems to also have side effects on peers' connectivity), however the race condition didn't show up. The second e2e run, although extremely slow, passed.
* Commit#5: Reverts Commit#2, to get the PR ready for merging (after approval)
* Commit#6 (pushed together with Commit#5): turned the locking added into RAII-style

Note that, unlike #673, we are leaving the race detection in e2e tests in place. Several runs in my local environment show no other _obvious_ race conditions are lurking, so leaving the race detection is useful to detect any further race condition early on in the process.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/consensus/reactor.go
+++ b/consensus/reactor.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
-	"runtime"
 	"sync"
 	"time"
 
@@ -747,7 +746,7 @@ OUTER_LOOP:
 			var ec *types.ExtendedCommit
 			var veEnabled bool
 			for i := 0; i < 10000; i++ {
-				runtime.Gosched()
+				time.Sleep(30 * time.Microsecond)
 				veEnabled = conR.conS.state.ConsensusParams.ABCI.VoteExtensionsEnabled(prs.Height)
 			}
 			if veEnabled {
--- a/consensus/reactor.go
+++ b/consensus/reactor.go
@@ -744,9 +744,12 @@ OUTER_LOOP:
 			// Load the block's extended commit for prs.Height,
 			// which contains precommit signatures for prs.Height.
 			var ec *types.ExtendedCommit
-			conR.conS.mtx.RLock()
-			veEnabled := conR.conS.state.ConsensusParams.ABCI.VoteExtensionsEnabled(prs.Height)
-			conR.conS.mtx.RUnlock()
+			var veEnabled bool
+			func() {
+				conR.conS.mtx.RLock()
+				defer conR.conS.mtx.RUnlock()
+				veEnabled = conR.conS.state.ConsensusParams.ABCI.VoteExtensionsEnabled(prs.Height)
+			}()
 			if veEnabled {
 				ec = conR.conS.blockStore.LoadBlockExtendedCommit(prs.Height)
 			} else {
--- a/consensus/reactor.go
+++ b/consensus/reactor.go
@@ -744,12 +744,8 @@ OUTER_LOOP:
 			// Load the block's extended commit for prs.Height,
 			// which contains precommit signatures for prs.Height.
 			var ec *types.ExtendedCommit
-			var veEnabled bool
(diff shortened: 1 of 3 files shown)


consensus/reactor.go
```diff
@@ -745,10 +745,12 @@ OUTER_LOOP:
 			// which contains precommit signatures for prs.Height.
 			var ec *types.ExtendedCommit
 			var veEnabled bool
+			conR.conS.mtx.RLock()
```
jmalicevic, line 748: Should we surround the sleep with a lock?

sergio-mena: Sorry, I'm not done with this yet (my bad I marked it ready for review), there are still two commits missing (and some description missing as well)

jmalicevic: My bad, sorry! :D

jmalicevic (approved):

consensus/reactor.go
sergio-mena: > My bad, sorry! :D

I disagree, it was my fault since I didn't think of working on a draft PR first. Sorry again

PR #714, opened 2023-04-14
spec/p2p: Reactor specification (WIP)

cason:
This is a **draft** and a work in progress.

Please only review this PR if you have context on the topic. If you would like to get some context on the content of the PR, feel free to contact me.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/spec/p2p/reactor/README.md
+++ b/spec/p2p/reactor/README.md
@@ -1,0 +1,27 @@
+# Reactors
+
+Reactor is the generic name for a component that interacts with the P2P communication layer.
+
+This document aims to specify the operation of CometBFT reactors.
+
+This is a work in progress, tracked by the issue [#599](https://github.com/cometbft/cometbft/issues/599).
+
+## Interface
+
+To become a reactor, a component has to implement the
+[`p2p.Reactor`](../../../p2p/base_reactor.go) interface.
+
+Much of the expected operation of a reactor can be derived from the
+documentation of this interface.
+
+## Grammar
+
+The expected sequence of calls to a reactor from the P2P layer will be
+described using a [Grammar](./grammar.md).
+
+
+## Quint
+
+The expected operation of a reactor will be modelled using
+[Quint](https://github.com/informalsystems/quint),
+an executable specification language.


spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,233 @@
...
+
+	/*
+	 * SetSwitch allows setting a switch.
+	 */
+	def SetSwitch(s: ReactorState, switch: bool): ReactorState = {
+		s.with("switch", switch)
+	}
+
+	// Code: Switch.AddReactor(name string, reactor Reactor)
+	action register() = all {
```
josef-widder, line 85: I would separate pure defs and actions in the code.

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,233 @@
...
+		val chIDs = channels.map(c => c.ID)
+		channels.size() == chIDs.size(), // channel IDs are unique
+		state' = state.SetSwitch(true),
+	}
+
+	/*
+	 * Start the service.
+	 * If it's already started or stopped, will return an error.
+	 */
+	def OnStart(s: ReactorState): ReactorState = {
```
josef-widder, line 97:
```suggestion
	pure def OnStart(s: ReactorState): ReactorState = {
```

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,233 @@
...
+		state.switch,
+		not(state.started) and not(state.stopped),
+		state' = state.OnStart()
+	}
+
+	/*
+	 * Stop the service.
+	 * If it's already stopped, will return an error.
+	 */
+	def OnStop(s: ReactorState): ReactorState = {
```
josef-widder, line 112:
```suggestion
	pure def OnStop(s: ReactorState): ReactorState = {
```

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,233 @@
...
+		state.started and not(state.stopped),
+		// Either no peer was added or all peers were removed
+		state.peers.size() == 0,
+		state' = state.OnStop()
+	}
+
+	/*
+	 * Return true if the service is running.
+	 */
+	def IsRunning(s: ReactorState): bool = {
```
josef-widder, line 127:
```suggestion
	pure def IsRunning(s: ReactorState): bool = {
```

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,233 @@
...
+		s.started and not(s.stopped)
+	}
+
+	/*
+	 * InitPeer is called by the switch before the peer is started. Use it to
+	 * initialize data for the peer (e.g. peer state).
+	 */
+	def InitPeer(s: ReactorState, peer: Peer): (ReactorState, Peer) = {
```
josef-widder, line 135:
```suggestion
	pure def InitPeer(s: ReactorState, peer: Peer): (ReactorState, Peer) = {
```

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,233 @@
...
+		not(peer.ID.in(state.peers)),
+		state' = state.InitPeer(peer)._1 // Updated peer not used here
+	}
+
+	/* 
+	 * AddPeer is called by the switch after the peer is added and successfully
+	 * started. Use it to start goroutines communicating with the peer.
+	 */
+	def AddPeer(s: ReactorState, peer: Peer): ReactorState = {
+		val startedRoutines = Set( {name: "ioRoutine", peer: peer} )
```
josef-widder, line 152: Is this just an example, or does each reactor has this 'ioRoutine'

josef-widder (approved):
Looks great!
As discussed synchronously, in Quint, I would move the guards from the actions into the `step`. I guess this would allow to see the correspondence to the grammar more easily. I suggest we first merge and then do these changes.

Also, I think it would be good to add a couple of tests.

spec/p2p/reactor/reactor.qnt
cason: This is an example. Should I point this out or maybe change the name?

cason: Moreover, should `AddPeer` be `pure def` as well?

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,233 @@
...
+	/*
+	 * GetChannels returns the list of MConnection.ChannelDescriptor. Make sure
+	 * that each ID is unique across all the reactors added to the switch.
+	 */
+	// TODO: we need a way to configure the returned channels!
+	def GetChannels(s: ReactorState): Set[ChannelDescriptor] = {
+		Set({ID: 3, Priority: 1})
+		.union(Set({ID: 7, Priority: 2}))
+		//.union(Set({ID: 3, Priority: 2})) // registration should fail!
+	}
```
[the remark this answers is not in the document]
cason: The reactor must implement the `GetChannels()` method. I am not sure (the TODO represents this) how to make the return value of this method something more configurable. In any case, this is not something that should/could change during the operation of a reactor.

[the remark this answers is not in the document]
josef-widder: Ah, OK.

spec/p2p/reactor/reactor.qnt
josef-widder: yeah. perhaps this is a matter of taste. Perhaps we can leave it like this for now. 
However, I think best readability would be
- pure defs
- vals (global state)
- actions (without guards)
- step (with guards to see correspondence to grammar)
but we can keep this issue for later.

spec/p2p/reactor/reactor.qnt
josef-widder: yes, add a comment that says this is an example. yes. should be a pure def :-)

spec/p2p/reactor/reactor.qnt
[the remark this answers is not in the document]
josef-widder: We might even leave it as is, and just add a comment that says that this is an example, and that it is up to the reactors to define their own channels.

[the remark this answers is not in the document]
cason: Added a comment for clarification.

spec/p2p/reactor/reactor.qnt
cason: Added a comment (new commits) to clarify.

spec/p2p/reactor/reactor.qnt
cason: Yes, this is a good organization. Lets see what other reviewers think.

spec/p2p/reactor/README.md
```diff
@@ -0,0 +1,224 @@
...
+- `Src`: the source or sender `Peer`, from which the message was received
+- `Message`: the actual message's payload, unmarshalled using protocol buffers
+
+Two important observations regarding the implementation of the `Receive` method:
+
+1. Concurrency: the implementation should consider concurrent invocations of
+   the `Receive` method carrying messages from different peers, as the
+   interaction with different peers is independent and messages can be received in parallel.
+1. Non-blocking: the implementation of the `Receive` method is expected not to block,
+   as it is invoked directly by the receive routines.
```
lasarojc, line 222: Where were the "...receive routines..." defined?

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,238 @@
...
+
+	var state: ReactorState
+
+	/*
+	 * GetChannels returns the list of MConnection.ChannelDescriptor. Make sure
+	 * that each ID is unique across all the reactors added to the switch.
+	 */
+	pure def GetChannels(s: ReactorState): Set[ChannelDescriptor] = {
+		// This method should return a list of channels, e.g.:
+		Set({ID: 3, Priority: 1})
```
lasarojc, line 72: Thinking in terms of testing the Quint spec, instead of having this hardcoded example here, could have a set of ChanelDescriptor as part of the ReactorState and returned by this function.
The set would be initialized by the test functions.

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,238 @@
...
+	 * If it's already stopped, will return an error.
+	 */
+	pure def OnStop(s: ReactorState): ReactorState = {
+		// Shutdown procedures should come here.
+		s.with("stopped", true)
+	}
+
+	// Code: Switch.OnStop()
+	action stop(): bool = all {
+		state.started and not(state.stopped),
```
lasarojc, line 119:
```suggestion
		IsRunning(state),
```

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,238 @@
...
+	 */
+	pure def OnStart(s: ReactorState): ReactorState = {
+		// Startup procedures should come here.
+		s.with("started", true)
+	}
+
+	// Code: Switch.OnStart()
+	action start(): bool = all {
+		state.switch,
+		not(state.started) and not(state.stopped),
```
lasarojc, line 104:
```suggestion
		NeverStarted(state)
```

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,238 @@
...
+		s.started and not(s.stopped)
+	}
+
+	/*
+	 * InitPeer is called by the switch before the peer is started. Use it to
+	 * initialize data for the peer (e.g. peer state).
+	 */
+	pure def InitPeer(s: ReactorState, peer: Peer): (ReactorState, Peer) = {
+		// This method can update the received peer, which is returned.
+		val updatedPeer = peer
```
lasarojc, line 138: does this have any practical effect, in Quint?

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,238 @@
...
+		state' = state.InitPeer(peer)._1 // Updated peer not used here
+	}
+
+	/* 
+	 * AddPeer is called by the switch after the peer is added and successfully
+	 * started. Use it to start goroutines communicating with the peer.
+	 */
+	pure def AddPeer(s: ReactorState, peer: Peer): ReactorState = {
+		// This method can be used to start routines to handle the peer.
+		// Below an example of an arbitrary 'ioRoutine' routine.
```
lasarojc, line 155: Where do the available routines come from? Maybe add them in a set in the ReactorState and them them here and then start.
Are they all started together? If so, start all here. Otherwise, chose one in addPeer and pass as a parameter to AddPeer

spec/p2p/reactor/reactor.qnt
cason: Yes, this is probably the best solution. Although one thing I don't like on it is that this value should be static, to never change.

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,238 @@
...
+	pure def SetSwitch(s: ReactorState, switch: bool): ReactorState = {
+		s.with("switch", switch)
+	}
+
+	// Code: Switch.AddReactor(name string, reactor Reactor)
+	action register() = all {
+		not(state.switch),
+		val channels = state.GetChannels()
+		val chIDs = channels.map(c => c.ID)
+		channels.size() == chIDs.size(), // channel IDs are unique
```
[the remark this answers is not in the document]
cason: Yes! 

To test this we should also have a field (can be part of state) with the list of other channelIDs already registered by other reactors.

In terms of the implementation, the `Switch` panics when a reactor tries to register a channel already allocated to another reactor.

spec/p2p/reactor/reactor.qnt
cason: This is an abstraction, we don't intend here to represent the operation of a specific reactor. In fact, each reactor is free to do whatever it wants at this method and each reactor has its own set of routines. The idea here is to represent that this is _right_  point where to start routines. Routines should not be started at `InitPeer`, as the `Peer` passed to that method is not started yet.

cason: > Otherwise, chose one in addPeer and pass as a parameter to AddPeer

We want to preserve the same parameters passed to the interface `Reactor` methods. The method only receives a `Peer` on the API, so we want to reproduce the same here.

cason: More broadly, in a next refinement/extension of the specification we might represent routines interacting with peers. These routines, for instance, send messages to a peer using the `Peer` handler. For now routines are very simpler, but we might make them more "useful".

spec/p2p/reactor/README.md
cason: Ok, this is maybe too much specific. I will replace by IO/handling routines or something like that.

cason: Actually, they are mentioned before, see lines 130, 142, 172, and 195.

spec/p2p/reactor/reactor.qnt
lasarojc: If it is part of the state, you can initialize it normally.
Another option is to have the channels as `const` which are initialized on `import`, like in the following example: 

```
module reactor {
   type test = int

   type State = {bla: test}

   const state: State
}

module tests {
    import reactor as R (
          state = {bla: 1}
    )
}
```

Mention it in case you want to use the feature in the future. Right now it is unstable and I don't recommend it.

spec/p2p/reactor/reactor.qnt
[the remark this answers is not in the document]
lasarojc: Yes, my recommendation would be in these lines, that is, to have a witness variable for all the registered channels.

spec/p2p/reactor/reactor.qnt
lasarojc: Got it. I just found it weird to have these examples as part of the spec. Then again, I don't know how to provide examples in a better way.
I wish we could define just the prototype here, for type checking, and force a refined instantiation when testing.

lasarojc (approved):

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,238 @@
...
+
+		// Service implementation.
+		started: bool,
+		stopped: bool,
+
+		// Indicates whether a switch was set: reactor registration.
+		switch: bool,
+	}
+
+	var state: ReactorState
```
hvanz, line 64: There is no need to have only one variable holding the whole state of a reactor. One could have one variable for each field in `ReactorState`. This will result in one indirection less.

hvanz (approved):
The text looks very good!

The Quint spec is a good first version though it may need a couple of iterations in the future. In particular I cannot run it in the repl, it complains about some error on the `state` variable. Also, the spec is for one reactor in one node. Things start to get interesting when there is more than one node.

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,238 @@
...
+		s.with("peers", s.peers.exclude(Set(peer.ID)))
+		 .with("routines", s.routines.exclude(stoppedRoutines))
+	}
+
+	// Code: Switch.stopAndRemovePeer(peer Peer, reason interface{})
+	action removePeer(peer: Peer, reason: str): bool = all {
+		state.switch and state.IsRunning(),
+		peer.ID.in(state.peers), // InitPeer(peer) and not RemovePeer(peer)
+		// Routines might not be started, namely: not AddPeer(peer)
+		// Routines could also be already stopped if Peer has erroed.
```
sergio-mena, line 184:
```suggestion
		// Routines could also be already stopped if Peer has errored.
```

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,238 @@
...
+		peer: Peer,
+	}
+
+	type ReactorState = {
+		// Peers that have been initialized but not yet removed.
+		// The reactor should expect receiving messages from them.
+		peers: Set[NodeID],
+
+		// The reactor runs multiple routines.
+		routines: Set[Routine],
```
sergio-mena, line 54: Do we really need to model routines? That's up to the Reactor, isn't it?
Besides, the fact that we need to keep track of the routines in some actions complicates those actions

Another way to express this is: I don't see any way in which `routines` help drive the model (they only appear once in an `all` clause of an actions, but it is not adding any restriction

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,238 @@
...
+	 */
+	pure def RemovePeer(s: ReactorState, peer: Peer, reason: str): ReactorState = {
+		// This method should stop routines created by `AddPeer(Peer)`.
+		val stoppedRoutines = s.routines.filter(r => r.peer.ID == peer.ID)
+		s.with("peers", s.peers.exclude(Set(peer.ID)))
+		 .with("routines", s.routines.exclude(stoppedRoutines))
+	}
+
+	// Code: Switch.stopAndRemovePeer(peer Peer, reason interface{})
+	action removePeer(peer: Peer, reason: str): bool = all {
```
sergio-mena, line 180: Do we need to model "reason", does is affect the behavior of the model?

sergio-mena (commented):
I really like this spec. Clear and concise.
It's been a long time I didn't learn so much (about the system I'm involved in) by reading so little.
Finally, bear in mind this will become a cornerstone of CometBFT's modularity management. My hope is that the Quint model (extended maybe?) can be used for testing the different reactors in isolation (similarly to @hvanz's work on ABCI++ apps last year)

spec/p2p/reactor/README.md
```diff
@@ -0,0 +1,224 @@
...
+specified in [IETF rfc7405](https://datatracker.ietf.org/doc/html/rfc7405)).
+It is inspired on the grammar produced to specify the interaction of CometBFT
+with an ABCI++ application, available [here](../../abci/abci%2B%2B_comet_expected_behavior.md).
+
+## Registration
+
+To become a reactor, a component has first to implement the `Reactor` interface,
+then to register the implementation with the p2p layer, using the
+`Switch.AddReactor(name string, reactor Reactor)` method, where `name` can be
+an arbitrary string.
```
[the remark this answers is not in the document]
cason: I don't think this check (single name per reactor) is made. I will double check.

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,238 @@
...
+
+		// The reactor runs multiple routines.
+		routines: Set[Routine],
+
+		// Service implementation.
+		started: bool,
+		stopped: bool,
+
+		// Indicates whether a switch was set: reactor registration.
+		switch: bool,
```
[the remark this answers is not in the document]
cason: In the next iteration, I think a `Switch` type will exist :D

spec/p2p/reactor/reactor.qnt
hvanz: Is it even a possibility to change channel ids after the node is initialized? Is it possible that different nodes are configured with different ids for the same channel name? And that we need to model and analyze this scenario? If not, we don't need to model this possibility and channels in the spec should be fixed (`val`s) for all nodes.

cason: > Is it even a possibility to change channel ids after the node is initialized?

No. The Reactor expect a call to `GetChannels()` at the registration phase, which only happens once.

> channels in the spec should be fixed (vals) for all nodes.

Yes, this is a good solution.

cason: > Is it possible that different nodes are configured with different ids for the same channel name? 

It is possible but nothing will work in this case. Channels don't have name, to be precise, but the are assumed to be _global_, meaning that all nodes use the same channel byte ID to implement the same channels (e.g., the mempool channel).

spec/p2p/reactor/reactor.qnt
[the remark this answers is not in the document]
cason: In the next round of specification, I should model the `Switch` from the point of view of a reactor. This will possible be included then.

spec/p2p/reactor/README.md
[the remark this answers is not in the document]
cason: So, if you register a reactor with a name that has already been used, you _overwrite_ the existing reactor. Notice, however, that registering two reactors claiming the same channel ID results in a panic. I will reword the description here.

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,238 @@
...
+	var state: ReactorState
+
+	/*
+	 * GetChannels returns the list of MConnection.ChannelDescriptor. Make sure
+	 * that each ID is unique across all the reactors added to the switch.
+	 */
+	pure def GetChannels(s: ReactorState): Set[ChannelDescriptor] = {
+		// This method should return a list of channels, e.g.:
+		Set({ID: 3, Priority: 1})
+		.union(Set({ID: 7, Priority: 2}))
```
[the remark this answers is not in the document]
cason: Yes, will rephrase.

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,238 @@
...
+		not(state.switch),
+		val channels = state.GetChannels()
+		val chIDs = channels.map(c => c.ID)
+		channels.size() == chIDs.size(), // channel IDs are unique
+		state' = state.SetSwitch(true),
+	}
+
+	/*
+	 * Start the service.
+	 * If it's already started or stopped, will return an error.
```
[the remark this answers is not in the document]
cason: The content within /*  */ are literally copied from the documentation present in the source code. In this case, this is the documentation of the `Service` interface.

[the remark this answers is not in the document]
cason: We don't model the error here. In fact, we implement the `OnStop()` method, while the method that cares about not stopping more than once is the `Stop()` method. The `Stop()` method is implemented in the `BaseService` and invokes the `OnStop()` method (re)implemented by every implementation of a `Service`. It is a mess, I know :P

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,238 @@
...
+	type ReactorState = {
+		// Peers that have been initialized but not yet removed.
+		// The reactor should expect receiving messages from them.
+		peers: Set[NodeID],
+
+		// The reactor runs multiple routines.
+		routines: Set[Routine],
+
+		// Service implementation.
+		started: bool,
```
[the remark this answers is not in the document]
cason: That is a good suggestion. Here I was blindly following the implementation (Go code). They use booleans here so that to use atomic read-compare-write which are very efficient, but renders the code hard to read.

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,238 @@
...
+		state' = state.RemovePeer(peer, reason)
+	}
+
+	/*
+	 * Receive is called by the switch when an envelope is received from any connected
+	 * peer on any of the channels registered by the reactor.
+	 */
+	pure def Receive(s: ReactorState, e: Envelope): ReactorState = {
+		// This method should process the message payload: e.Message.
+		s
```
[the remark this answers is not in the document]
cason: Yes, but what would I do then with the messages on this set, I mean, which tests or checks?

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,238 @@
...
+	 * Peer is an interface representing a peer connected on a reactor.
+	 */
+	type Peer = {
+		ID: NodeID,
+
+		// Other fields can be added to represent the p2p operation.
+	}
+
+	// Byte ID used by channels.
+	type Byte = int
```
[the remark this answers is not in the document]
cason: Thank you for this feedback and the explanation. Indeed, numbers here were used because it is like that on the code. The point is that IDs should be unique, and you are completely right regarding the unnecessary complexity added to the verification.

I will change it to strings in the next round of the spec. :)

spec/p2p/reactor/reactor.qnt
cason: The idea here is to render this spec re-usable. Also, with this approach one could model multiple reactor instances.

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,238 @@
+/*
+ * Reactor is responsible for handling incoming messages on one or more
+ * Channel. Switch calls GetChannels when reactor is added to it. When a new
```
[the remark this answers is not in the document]
cason: The content within /* */ are literally copied from the documentation present in the source code. In this case, this is the `Reactor` interface.

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,238 @@
...
+		// Peers that have been initialized but not yet removed.
+		// The reactor should expect receiving messages from them.
+		peers: Set[NodeID],
+
+		// The reactor runs multiple routines.
+		routines: Set[Routine],
+
+		// Service implementation.
+		started: bool,
+		stopped: bool,
```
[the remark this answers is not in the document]
cason: Quint doesn't have yet a good way to represent enums. But you are right, Hernán noticed the same.

I will replace this by a state (string) in the next interaction of the spec.

spec/p2p/reactor/reactor.qnt
```diff
@@ -0,0 +1,238 @@
...
+	pure def GetChannels(s: ReactorState): Set[ChannelDescriptor] = {
+		// This method should return a list of channels, e.g.:
+		Set({ID: 3, Priority: 1})
+		.union(Set({ID: 7, Priority: 2}))
+	}
+
+	/*
+	 * SetSwitch allows setting a switch.
+	 */
+	pure def SetSwitch(s: ReactorState, switch: bool): ReactorState = {
```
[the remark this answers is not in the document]
cason: In the next round of this spec, `Switch` should be a modeled type, so be more complex than just a boolean.

spec/p2p/reactor/reactor.qnt
cason: Yes, Lásaro has commented something on the same lines.

If I don't find an actually use for them in the next round of this specification, I will get rid of them, making the starting and stopping or routines a comment/suggestion.

spec/p2p/reactor/reactor.qnt
cason: Good point.

If I don't find an actually use for it in the next round of this specification, I will get rid of it.

spec/p2p/reactor/README.md
```diff
@@ -0,0 +1,224 @@
...
+Reactor is the generic name for a component that employs the p2p communication layer.
+
+The main source of documentation for a reactor is the [`Reactor`](../../../p2p/base_reactor.go) interface.
+In order to use communication services provided by the p2p layer,
+a component has to implement the p2p's package `Reactor` interface.
+
+The goal of this document is to complement the existing documentation by
+specifying the behaviour of the p2p when interacting with a reactor.
+So while the `Reactor` interface defines the methods invoked and determines
+what the p2p layer expects from a reactor, this documentation focuses on the
```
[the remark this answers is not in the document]
cason: No, you are not missing. The wording is confuse. I will fix it before merging.

spec/p2p/reactor/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+The main duty of a reactor is to handle incoming messages on the channels it
+has registered with the p2p layer.
+
+The _pre-condition_ for receiving a message from a `Peer` is that the p2p layer
+has previously invoked `InitPeer(Peer)`.
+This means that the reactor should be able to receive a message from a `Peer`
+_before_ `AddPeer(Peer)` is called.
+While this is not the usual scenario, it can happen because starting the peer's
+send and receive routines and adding the peer are done in parallel:
```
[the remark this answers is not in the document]
cason: This is a good point: do the reactors benefit from having two calls to add a peer?

I will add a few words of discussion on this in the next round of this spec.

spec/p2p/reactor/README.md
```diff
@@ -0,0 +1,224 @@
...
+The grammar is written in case-sensitive Augmented Backus–Naur form (ABNF,
+specified in [IETF rfc7405](https://datatracker.ietf.org/doc/html/rfc7405)).
+It is inspired on the grammar produced to specify the interaction of CometBFT
+with an ABCI++ application, available [here](../../abci/abci%2B%2B_comet_expected_behavior.md).
+
+## Registration
+
+To become a reactor, a component has first to implement the `Reactor` interface,
+then to register the implementation with the p2p layer, using the
+`Switch.AddReactor(name string, reactor Reactor)` method, where `name` can be
```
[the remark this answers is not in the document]
cason: The grammar refers to a single reactor. This should be made clear in the text before merging.

spec/p2p/reactor/README.md
```diff
@@ -0,0 +1,224 @@
...
+```abnf
+registration    = get-channels set-switch
+```
+
+The p2p layer retrieves from the reactor a list of channels the reactor is
+responsible for, using the `GetChannels()` method.
+The reactor implementation should thereafter expect the delivery of every
+message received by the p2p layer in the informed channels.
+
+The second method `SetSwitch(Switch)` concludes the handshake between the
```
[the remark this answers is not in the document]
cason: This is a handshake: the switch receives a reactor, and the reactor needs to receive the switch.

spec/p2p/reactor/README.md
```diff
@@ -0,0 +1,224 @@
...
+that `Peer` instance.
+
+## Start peer
+
+Once a `Peer` is initialized by every reactor, the p2p layer starts the peer's
+communication routines and adds the `Peer` to the set of connected peers.
+If both steps are concluded without errors, the reactor's `AddPeer(Peer)` is invoked:
+
+```abnf
+start-peer      = [*receive] (connected-peer / start-error)
```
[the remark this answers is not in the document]
cason: For the minor comment: yes, this is redundant. In a previous version we had `[receive]` with a comment that this could actually happen more than once. But then we realize that we cannot provide an "upper bound" on the number of times this happens.

I would maintain this wording which, although redundant, highlights that receiving messages here is not the "expected" behavior, but a possible behavior that has to be taken into account.

spec/p2p/reactor/README.md
```diff
@@ -0,0 +1,224 @@
...
+connected-peer  = add-peer *receive
+```
+
+In case of errors, a message is logged informing that the p2p layer failed to start the peer.
+This is not a common scenario and it is only expected to happen when
+interacting with a misbehaving peer. A practical example is reported on this
```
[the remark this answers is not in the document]
cason: Yes... and no. If the network is flaky, the connection will probably timeout. By then, we already called `AddPeer`.

spec/p2p/reactor/README.md
```diff
@@ -0,0 +1,224 @@
+# Reactors
```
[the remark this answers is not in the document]
cason: This is a good point that I will address.

cason:
"In particular I cannot run it in the repl, it complains about some error on the state variable"

@hvanz, due to changes made here on GitHub, the Quint model was not "compiling". Essentially, `IsRunning()` was used before being declared. Can you try again? It should be working now.

spec/p2p/reactor/reactor.qnt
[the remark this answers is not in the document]
cason: This was already changed, thank you for the suggestion.

spec/p2p/reactor/reactor.qnt
cason: Pure defs representing API methods and actions representing the p2p layer behavior are separated now.

spec/p2p/reactor/reactor.qnt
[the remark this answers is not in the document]
hvanz: Enums in Quint (and TLA+) could be just sets.

hvanz:
> Can you try again? It should be working now.
Yes, now it's working as expected :)

spec/p2p/reactor/README.md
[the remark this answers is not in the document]
cason: I think the added changes should have fixed this confusion.

PR #724, opened 2023-04-17
RFC 102: Improve forward compatibility of proto-generated Rust

mzabaluev:


thanethomson (approved):
I think this is great i.t.o. capturing the problem that our one Protobuf versioning approach introduces for certain programming languages.

sergio-mena:
Thanks for this RFC. The test is clear IMO.
My answer to the final question is still to go for the "Sensitive" approach. This RFC describes an elegant solution for the problem in Rust code generation, but the general aim of CometBFT should not be limited to Go and Rust, but a more comprehensive set of target languages.
Abandoning the "Sensitive" approach would risk bumping N times into the same problem, and having to apply similar ideas to those in this RFC every time for a different language.

docs/rfc/rfc-102-rust-gen-builders.md
```diff
@@ -0,0 +1,132 @@
...
+### References
+
+* Issue [#399](https://github.com/tokio-rs/prost/issues/399) in the prost
+  repository captures the general problem and discussion.
+* CometBFT PR [#495](https://github.com/cometbft/cometbft/pull/495) introduces
+  versioned protobuf definitions, currently with extra versioning applied
+  to accommodate API breakage caused in tendermint-rs by code generated with
+  prost-build.
+* [Notes](https://docs.google.com/document/d/1DoxKiYtUx44xZv5my-bkfWZKY6TklvxpSUrdX9yOpNw/edit?usp=sharing) on the discussion during a 13 Apr 2023 meeting, detailing the
+  considerations specific to CometBFT versioning.
```
sergio-mena, line 41: Please add a reference to ADR 103 when it is merged

mzabaluev: Done in https://github.com/cometbft/cometbft/pull/724/commits/3a212e43bcbed780671d79cbe7c43d9f8fcc149a

Issue #29, opened 2022-12-23
State sync from local snapshot: Implementation


sergio-mena:
Tasks:

- [ ] Implementation (existing implementation of ADR083: tendermint/tendermint#9541)
- [ ] Introduce necessary changes to documentation and/or spec to keep them up to date
- [ ] Check if tendermint/tendermint#4642 can be closed once implementation merged

DoD:
- Implementation code complete and all PRs merged
- Any changes needed to the documentation/spec are merged

Original issue: tendermint/tendermint#9948

chillyvee:
Happy to collaborate when you are ready.  We just updated the code for the latest Juno v13 and have some intermediate feedback based on previous discussions.

adizere:
Thank @chillyvee, appreciate it ! We'll get back to you in a few weeks time after we have explored all the existing material in tendermint/tendermint and have clarity over the solution. This work is part of our Q2 priorities.

adizere:
Hi @chillyvee would you kindly reconstruct your [original PR](https://github.com/tendermint/tendermint/pull/9541/commits/96202a30ad76ae31729abea2db84b3ddae24febf) and submit to this repo, with a base branch v0.34? We've now in the "understand problem & solution" space (#28) so it would be good to see this implementation.

Thanks in advance!

yihuang:
> Hi @chillyvee would you kindly reconstruct your [original PR](https://github.com/tendermint/tendermint/pull/9541/commits/96202a30ad76ae31729abea2db84b3ddae24febf) and submit to this repo as a draft PR preferably, with a base branch v0.34? We've now in the "understand problem & solution" space (#28) so it would be good to see this implementation.
> 
> Thanks in advance!

Love this feature, would like to help if needed, BTW, where can I find the ADR083?

From my understanding, state-sync can be decoupled into different components:
- snapshot discovery and chunks downloading
- snapshot chunks verification with light client protocol
- snapshot restoration
- node startup with the new state.

I think ideally we should have offline commands for each steps to provide maximum flexibility.

cason:
I think Adi referred to the open PR in the old repository: https://github.com/tendermint/tendermint/pull/9651

yihuang:
> I think Adi referred to the open PR in the old repository: [tendermint/tendermint#9651](https://github.com/tendermint/tendermint/pull/9651)
> 
[...]

PR #728, opened 2023-04-19
Enable local snapshot restore (backport: #733)

yihuang:
Closes: #29 
Original PR: https://github.com/tendermint/tendermint/pull/9541

Sync state in Handshake via RPC

CV Cleanup dead code

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/consensus/replay.go
+++ b/consensus/replay.go
@@ -2,6 +2,8 @@ package consensus
 
 import (
 	"bytes"
+	"context"
+	"errors"
 	"fmt"
 	"hash/crc32"
 	"io"
@@ -11,8 +13,11 @@ import (
 	abci "github.com/tendermint/tendermint/abci/types"
 	"github.com/tendermint/tendermint/crypto/merkle"
 	"github.com/tendermint/tendermint/libs/log"
+	"github.com/tendermint/tendermint/light"
 	"github.com/tendermint/tendermint/proxy"
 	sm "github.com/tendermint/tendermint/state"
+	"github.com/tendermint/tendermint/statesync"
+	"github.com/tendermint/tendermint/store"
 	"github.com/tendermint/tendermint/types"
 )
 
@@ -198,27 +203,31 @@ func makeHeightSearchFunc(height int64) auto.SearchFunc {
 //---------------------------------------------------
 
 type Handshaker struct {
-	stateStore   sm.Store
-	initialState sm.State
-	store        sm.BlockStore
-	eventBus     types.BlockEventPublisher
-	genDoc       *types.GenesisDoc
-	logger       log.Logger
+	stateStore    sm.Store
+	initialState  sm.State
+	store         sm.BlockStore
+	eventBus      types.BlockEventPublisher
+	genDoc        *types.GenesisDoc
+	logger        log.Logger
+	stateProvider statesync.StateProvider
 
 	nBlocks int // number of blocks applied to the state
 }
 
 func NewHandshaker(stateStore sm.Store, state sm.State,
-	store sm.BlockStore, genDoc *types.GenesisDoc) *Handshaker {
+	store sm.BlockStore, genDoc *types.GenesisDoc,
+	stateProvider statesync.StateProvider) *Handshaker {
 
 	return &Handshaker{
-		stateStore:   stateStore,
-		initialState: state,
-		store:        store,
-		eventBus:     types.NopEventBus{},
-		genDoc:       genDoc,
-		logger:       log.NewNopLogger(),
-		nBlocks:      0,
+		stateStore:    stateStore,
+		initialState:  state,
+		store:         store,
+		eventBus:      types.NopEventBus{},
+		genDoc:        genDoc,
+		logger:        log.NewNopLogger(),
+		stateProvider: stateProvider,
+
+		nBlocks: 0,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


consensus/replay.go
```diff
@@ -278,6 +287,53 @@ func (h *Handshaker) Handshake(proxyApp proxy.AppConns) error {
...
+		return sm.State{}, nil, errors.New("Missing State Provider")
+	}
+
+	pctx, pcancel := context.WithTimeout(context.TODO(), 30*time.Second)
+	defer pcancel()
+
+	// Optimistically build new state, so we don't discover any light client failures at the end.
+	state, err = h.stateProvider.State(pctx, appBlockHeight)
+	if err != nil {
+		h.logger.Error("failed to fetch and verify tendermint state", "err", err)
```
adizere, line 308: Nit: Should we use "comet" instead of "tendermint" ?

consensus/replay.go
```diff
@@ -278,6 +287,53 @@ func (h *Handshaker) Handshake(proxyApp proxy.AppConns) error {
...
+		if err == light.ErrNoWitnesses {
+			return sm.State{}, nil, err
+		}
+		return sm.State{}, nil, errors.New("snapshot was rejected")
```
adizere, line 312: Can we be more specific in this error message? It seems to me that what happened here was that Comet's light client verified the app hash present in the local snapshot of `app.db`, and this app hash was incorrect. Is this accurate?

consensus/replay.go
```diff
@@ -278,6 +287,53 @@ func (h *Handshaker) Handshake(proxyApp proxy.AppConns) error {
...
+	}
+
+	commit, err = h.stateProvider.Commit(pctx, appBlockHeight)
+	if err != nil {
+		h.logger.Error("failed to fetch and verify commit", "err", err)
+		if err == light.ErrNoWitnesses {
+			return sm.State{}, nil, err
+		}
+		return sm.State{}, nil, errors.New("snapshot was rejected")
```
adizere, line 321: Same comment as above: Would it be good to have more explicit error message here?

adizere (commented):
Thanks for the initial pass at this solution, very grateful for the work here!

I left some questions / suggestions. I would be glad to fix myself some of the concerns I raised directly in the dev branch, if that's ok with you @yihuang and @chillyvee; but will of course give you priority to make commits on the branch.

node/node.go
```diff
@@ -775,8 +791,10 @@ func NewNode(config *cfg.Config,
 	// Create the handshaker, which calls RequestInfo, sets the AppVersion on the state,
 	// and replays any blocks as necessary to sync CometBFT with the app.
 	consensusLogger := logger.With("module", "consensus")
-	if !stateSync {
-		if err := doHandshake(stateStore, state, blockStore, genDoc, eventBus, proxyApp, consensusLogger); err != nil {
+
+	// RPC recovery is required if blockStore Height is 0
+	if blockStore.Height() == 0 || !stateSync {
```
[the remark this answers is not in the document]
chillyvee: This is worth discussing since there are gaps in knowledge that need to be filled.  There are three cases where the comet height is zero.

1) Actually genesis
2) Statesync completely over the network as we use today
3) LocalStatesync where the application.db is already restored, but comet is at 0

One of the cosmos-sdk to comet integration challenges for local state sync is communicating the height of appstate/application.db

In the initial implementation, an extra applicationd.db height parameter was passed from cosmos-sdk to comet.  However a previous code review tried to eliminate that extra parameter and simply check if comet height was 0, and start RPC statesync.

However that actually causes some issues because a normal network statesync also has comet height == 0, but application.db is also 0. 

We need an agreeable way to determine the height of application.db from cosmos-sdk.  Since a previous code review determined a parameter was not desirable, is there a better way to query the application height?  Is there an object we should look at, or make an ABCI call?

consensus/replay.go
chillyvee: ok

consensus/replay.go
chillyvee: Referencing `func (s *lightClientStateProvider) State(ctx context.Context, height uint64) (sm.State, error) {`

The error conditions are failure to VerifyLightBlockAtHeight , RPC Client Creation Failure (Bad URL), and failure to fetch consensus parameters for the height.

It might be better to 1) return or 2) wrap the original error for the best accuracy.  Is there a preference?

consensus/replay.go
chillyvee: What do you think of wrapping the error to add detail?

yihuang:
Haven't looked into detail, but find an unexpected behavior: when the app height is 0, statesync enabled, it still try to init with genesis.json, which don't happen with normal version.

node/node.go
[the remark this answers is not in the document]
adizere: I see, thanks for the detailed notes and history here, I wasn't aware of the subtlety here!

> Is there an object we should look at, or make an ABCI call?

An ABCI call (`info`) would indeed provide the app height I believe. But not sure that's the best way.

consensus/replay.go
adizere: Yes, that's a good idea. Let's wrap the `err` to save the detail and return it to the caller.

node/node.go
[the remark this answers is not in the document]
chillyvee: Originally it was added as a variable to configure when cosmos-sdk created the Node{}, but it was asked to be removed during code-review, probably to avoid creating a new dependency between cosmos-sdk and comet

Now that we know the removal causes an issue, it's up to us to pick a new way to compare the heights of comet vs cosmos-sdk app. 

Is there general guidance on how to communicate variables between comet and cosmos-sdk?

consensus/replay.go
cason: There is not snapshot involved here.

The error should say that the application state was rejected, instead.

consensus/replay.go
```diff
@@ -278,6 +287,53 @@ func (h *Handshaker) Handshake(proxyApp proxy.AppConns) error {
...
+	}
+
+	if err := h.stateStore.Save(state); err != nil {
+		return sm.State{}, nil, err
+	}
+	if err := h.store.(*store.BlockStore).SaveSeenCommit(int64(appBlockHeight), commit); err != nil {
```
cason, line 327: Following the original implementation, which is `Node.startStateSync` (I don't know the branch to refer, so no link, sorry), you should use `state.LastBlockHeight` here for compatibility. It should be equal to `appBlockHeight`, except for uint/int stuff, but for coherence it is better to follow the same approach.

consensus/replay.go
```diff
@@ -299,6 +355,19 @@ func (h *Handshaker) ReplayBlocks(
...
+	// Check restorable conditions for localsync
+	if storeBlockHeight == 0 && appBlockHeight > 0 {
+		var syncErr error
+		if state, _, syncErr = h.localSync(uint64(appBlockHeight)); syncErr != nil {
+			panic("Unable to use RPC to update state")
+		}
+		h.logger.Info("localSync resulting state.Version.Consensus.Block", "state", state.Version.Consensus.Block)
+
+		// Assume Heights Restored, Update Heights for edge cases and constraints on the storeBlockHeight and storeBlockBase.
+		storeBlockHeight = appBlockHeight
```
cason, line 367: I would say that these values should be read from the stores, not blindly updated here.

node/node.go
```diff
@@ -323,7 +324,22 @@ func doHandshake(
 	proxyApp proxy.AppConns,
 	consensusLogger log.Logger,
 ) error {
-	handshaker := cs.NewHandshaker(stateStore, state, blockStore, genDoc)
+	var err error
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	stateProvider, err := statesync.NewLightClientStateProvider(
```
cason, line 330: If we are not using this "local sync", this would error, right?

We have to still support the case in which the node is simply recovering, and will restore all state from disk.

node/node.go
```diff
@@ -989,8 +1007,9 @@ func (n *Node) OnStart() error {
 		return fmt.Errorf("could not dial peers from persistent_peers field: %w", err)
 	}
 
-	// Run state sync
+	// Statesync reqeusting all data from P2P
```
cason, line 1010:
```suggestion
	// State sync fetching snapshots from peers
```

node/node.go
```diff
@@ -989,8 +1007,9 @@ func (n *Node) OnStart() error {
...
 	if n.stateSync {
+		// If state and blockStore.Height are both at the same height, skip the P2P Statesync and immediately enter consensus
```
cason, line 1012: I couldn't understand this comment.

cason:
Are we proceeding with this approach/solution attempt? If so, we should add the `wip` label to the PR. Otherwise, GitHub will close it.

yihuang:
> Are we proceeding with this approach/solution attempt? If so, we should add the `wip` label to the PR. Otherwise, GitHub will close it.

I think no need to pursue

adizere:
Closing, a different solution was chosen.

chillyvee:
Which other solution was chosen?  Just had a break in schedule to work on the Juno v14 local state sync demo today.

yihuang:
> Which other solution was chosen? Just had a break in schedule to work on the Juno v14 local state sync demo today.

https://github.com/cosmos/cosmos-sdk/pull/16061
https://github.com/cosmos/cosmos-sdk/pull/16067

chillyvee:
nice code @yihuang :)

PR #733, opened 2023-04-20
Enable local snapshot restore

yihuang:
Closes: #29 

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/node/setup.go
+++ b/node/setup.go
@@ -165,15 +165,31 @@ func createAndStartIndexerService(
 }
 
 func doHandshake(
+	config *cfg.StateSyncConfig,
 	stateStore sm.Store,
 	state sm.State,
 	blockStore sm.BlockStore,
 	genDoc *types.GenesisDoc,
 	eventBus types.BlockEventPublisher,
 	proxyApp proxy.AppConns,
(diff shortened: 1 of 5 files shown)


adizere:
Keeping this draft until we finish the review of #728.

node/setup.go
```diff
@@ -165,15 +165,31 @@ func createAndStartIndexerService(
...
+		ctx,
+		state.ChainID, state.Version, state.InitialHeight,
+		config.RPCServers, light.TrustOptions{
+			Period: config.TrustPeriod,
+			Height: config.TrustHeight,
+			Hash:   config.TrustHashBytes(),
+		}, consensusLogger)
+	if err != nil {
+		return fmt.Errorf("failed to set up light client state provider: %w", err)
+	}
```
cmwaters, line 190: We should only create the state provider if we need it for the local sync. If I understand this correctly, if a user hasn't set up the state sync config because they are simply restarting their node and just want to block sync, this will error (because it needs at least 2 rpc servers), causing start up to needlessly fail.

adizere:
Different solution pursued.

Issue #687, opened 2023-04-11
`ApplyBlock` does not fill the time field in FinalizeBlock


sergio-mena:
## Bug Report

`ApplyBlock` does not fill the time field in FinalizeBlock

### Setup

**CometBFT version** : `main` and `v0.38.x`

**ABCI app** (name for built-in, URL for self-written if it's publicly available):
N/A

**Environment**:
N/A

**node command runtime flags**:
N/A

### Config

N/A

### What happened?

`ApplyBlock` does not fill the time field in FinalizeBlock

### What did you expect to happen?

`ApplyBlock` must fill the time field in FinalizeBlock

### How to reproduce it

`TestFinalizeBlockDecidedLastCommit` can be easily modify to detect this condition

### Logs

N/A

### `dump_consensus_state` output

N/A

### Anything else we need to know

Patch available.

Thanks @ze97286 for catching this

PR #746, opened 2023-04-24
Fill `time` field in `ExecCommitBlock`

### LINKED ISSUE #739: `ExecCommitBlock` does not fill the time field in `FinalizeBlock`
## Bug Report

Related to #687

`ExecCommitBlock` does not fill the time field in `FinalizeBlock`

### Setup

**CometBFT version** : `main` and `v0.38.x`

N/A

**Environment**:
N/A

N/A

### Config

N/A

### What happened?

`ExecCommitBlock` does not fill the time field in `FinalizeBlock`. This shows up when replaying blocks

`ExecCommitBlock` must fill the time field in `FinalizeBlock`

### How to reproduce it

TODO

### Logs

N/A

N/A

Patch available.

Thanks again to @ze97286 for catching this

sergio-mena:
Closes #739

Similar to #687, `ExecCommitBlock` is not filling the `time` field. Further, there was no UT catching this.
This PR add logic to an existing test case to check for that field. The test case is initially failing. When added the production fix (and fixing the test case logic), the test case passes.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/state/execution_test.go
+++ b/state/execution_test.go
@@ -223,7 +223,8 @@ func TestFinalizeBlockValidators(t *testing.T) {
 		block := makeBlock(state, 2, lastCommit.ToCommit())
 
 		_, err = sm.ExecCommitBlock(proxyApp.Consensus(), block, log.TestingLogger(), stateStore, 1)
-		require.Nil(t, err, tc.desc)
+		require.NoError(t, err, tc.desc)
+		require.True(t, app.LastTime.After(now), "last_time %v now %v", app.LastTime, now)
 
 		// -> app receives a list of validators with a bool indicating if they signed
 		ctr := 0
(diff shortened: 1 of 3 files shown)


state/execution_test.go
```diff
@@ -206,10 +206,11 @@ func TestFinalizeBlockValidators(t *testing.T) {
 		desc                     string
 		lastCommitSigs           []types.ExtendedCommitSig
 		expectedAbsentValidators []int
+		shouldHaveTs             bool
```
thanethomson, line 209: The linter's complaining about this field name - is the "Ts" meant to represent "timestamp"?

sergio-mena: Yes. What do you suggest?

thanethomson: The linter was recommending `shouldHaveTS`, but any of those should do.

lasarojc (approved):

PR #755, opened 2023-04-25
Fix `TestStateOversizedBlock`

### LINKED ISSUE #754: Oversized blocks are not thoroughly tested
### Setup

`main` and `v0.38.x`

N/A

**Environment**:
- **OS** (e.g. from /etc/os-release): any
- **Install tools**: N/A

N/A

### Config

N/A

### What happened?

Test case `TestStateOversizedBlock` has several problems and is not thoroughly testing oversized blocks; e.g., if you feed a block with correct size, the test case is still passing.

Test case should fail if provided with a block with a correct size

### How to reproduce it

* Delete the lines modifying `propBlock.Data.Txs` and `propBlock.Header.DataHash` in test case `TestStateOversizedBlock`
* Run the test
* The test should fail, but it passes

### Logs

N/A

N/A

Patch exists, opening PR shortly

Once again, thanks to @ze97286 for drawing our attention to checking incoming oversized proposals

sergio-mena:
Closes #754

Test case `TestStateOversizedBlock` has several problems:

* The way if modifies the data hash doesn't work:
  * the call to `propBlock.Data.Hash()` will not recalculate the hash, if one is already calculated (which is the case here)
  * so, if you reduce the TX size so that the block size is correct, the consensus state will still not accept the proposal because the hash doesn't match
  * and the test will still "pass" thinking the consensus state rejected the value because it is too big (which is not the case)
* The test does not test the positive scenario (which probably caused the problem described above) 
* The test uses 2001 as an off-by-1 value for the block size but the current limit is actually way below (1720 at the time of writing this); so 2001 is oversized but it is not an off-by-1 value

This PR addresses all these problems

NOTE FOR REVIEWERS: For some reason, GitHub is showing a uselessly complex diff in the "Files Changed" tab. It is recommended to use the "hide whitespaces" option of GitHub. To do so, go to "Files changed" --> Click on the cog icon next to "Jump to" --> Click on "Hide whitespace" --> Click on "Apply and reload"

---

#### PR checklist

- [x] Tests written/updated


cason:
> NOTE FOR REVIEWERS: For some reason, GitHub is showing a uselessly complex diff in the "Files Changed" tab.

The diff is complex because the main method was moved into a new context (shifted by two tabs) to allow the execution of two versions of the test (success, failure).

The suggested command presented about the same diff as the File changed tab.

consensus/state_test.go
```diff
@@ -256,73 +257,88 @@ func TestStateBadProposal(t *testing.T) {
...
+			oversized: false,
+		},
+		{
+			name:      "off-by-1 max size, incorrect block",
+			oversized: true,
+		},
+	} {
+		t.Run(testCase.name, func(t *testing.T) {
+			cs1, vss := randState(2)
+			cs1.state.ConsensusParams.Block.MaxBytes = maxBytes
```
cason, line 277: Here `maxBytes` replaces `2000` (line 263, old).

consensus/state_test.go
```diff
@@ -256,73 +257,88 @@ func TestStateBadProposal(t *testing.T) {
...
-	propBlockParts, err := propBlock.MakePartSet(partSize)
-	require.NoError(t, err)
-	blockID := types.BlockID{Hash: propBlock.Hash(), PartSetHeader: propBlockParts.Header()}
-	proposal := types.NewProposal(height, round, -1, blockID)
-	p := proposal.ToProto()
-	if err := vs2.SignProposal(cs1.state.ChainID, p); err != nil {
-		t.Fatal("failed to sign bad proposal", err)
-	}
-	proposal.Signature = p.Signature
+			propBlock, propBlockParts := findBlockSizeLimit(t, height, maxBytes, cs1, partSize, testCase.oversized)
```
cason, line 283: This line replace lines 272-275 and 281-282 from the old version.

consensus/state_test.go
```diff
@@ -256,73 +257,88 @@ func TestStateBadProposal(t *testing.T) {
...
+			}
 
-	// and then should send nil prevote and precommit regardless of whether other validators prevote and
-	// precommit on it
-	ensurePrevote(voteCh, height, round)
-	validatePrevote(t, cs1, round, vss[0], nil)
+			// start the machine
+			startTestRound(cs1, height, round)
+
+			t.Log("Block Sizes;", "Limit", cs1.state.ConsensusParams.Block.MaxBytes, "Current", totalBytes)
```
cason, line 313: Replaces old line 304.

cason, line 313:
```suggestion
			t.Log("Block Sizes:", "Limit", cs1.state.ConsensusParams.Block.MaxBytes, "Current", totalBytes)
```

consensus/state_test.go
```diff
@@ -256,73 +257,88 @@ func TestStateBadProposal(t *testing.T) {
...
+				validateHash = nil
+				lockedRound = -1
+				// if the block is oversized cs1 should log an error with the block part message as it exceeds
+				// the consensus params. The block is not added to cs.ProposalBlock so the node timeouts.
+				ensureNewTimeout(timeoutProposeCh, height, round, cs1.config.Propose(round).Nanoseconds())
+				// and then should send nil prevote and precommit regardless of whether other validators prevote and
+				// precommit on it
+			}
+			ensurePrevote(voteCh, height, round)
+			validatePrevote(t, cs1, round, vss[0], validateHash)
```
cason, line 327: Replaces old line 312, `validateHash` in place of `nil`. The value will be `nil` as well if `testCase.oversized` (line 318, new).

consensus/state_test.go
```diff
@@ -256,73 +257,88 @@ func TestStateBadProposal(t *testing.T) {
...
+			require.NoError(t, err)
 
-	signAddVotes(cs1, cmtproto.PrevoteType, propBlock.Hash(), bps.Header(), false, vs2)
-	ensurePrevote(voteCh, height, round)
-	ensurePrecommit(voteCh, height, round)
-	validatePrecommit(t, cs1, round, -1, vss[0], nil, nil)
+			signAddVotes(cs1, cmtproto.PrevoteType, propBlock.Hash(), bps.Header(), false, vs2)
+			ensurePrevote(voteCh, height, round)
+			ensurePrecommit(voteCh, height, round)
+			validatePrecommit(t, cs1, round, lockedRound, vss[0], validateHash, validateHash)
```
cason, line 335: Replaces old line 321, `lockedRound` instead of `-1` and `validateHash` instead of `nil`. Again, if `testCase.oversized`, the old values are adopted as well in the new version.

cason (approved):
I left comments with the actual changes in the code, to help future reviewers.

As mentioned in the description, the set of changes (except of the new method) is small but the diff messes everything and it is hard to read.

sergio-mena:
> The suggested command presented about the same diff as the File changed tab.

That's surprising, in my case it produced [this](https://gist.github.com/sergio-mena/cf4b699d2c729fc98a930f8337a8eb4d). As you can see, it is only showing the parts of the function's body that I changed. The rest doesn't appear in the diff, even if the indentation doesn't match in "unchanged" lines. Much easier to review than what is in this PR.

sergio-mena:
> You probably have an enhanced git diff setup :D

Actually, I just learnt something on GitHub... you can get the same here...

Go to "Files changed" --> Click on the cog icon next to "Jump to" --> Click on "hide whitespace" --> Click on "Apply and reload"

updating the description...

cason:
Much better using this setup, it renders my comments useless, ehehe.

PR #763, opened 2023-04-25
Rename proto enum members to improve compliance with buf style guide

### LINKED ISSUE #736: proto: enum definitions should follow the `*_UNSPECIFIED` guideline
> One last thing from my side. I'm reading [this](https://protobuf.dev/programming-guides/dos-donts/), and I think it would be a good moment to rename all \*__UNKNOWN_ enums we have to \*__UNSPECIFIED_, to align with the recommendations.
Also, please check if we have an enum without an UNKNOWN/UNSPECIFIED zero-value

_Originally posted by @sergio-mena in https://github.com/cometbft/cometbft/issues/495#issuecomment-1515903441_

mzabaluev:
Fixes #736

---

#### PR checklist

- [ ] Tests written/updated


abci/types/types.go
cason, line 1: The standard here is to given capitalized names to constants (`const`)?

I noticed that this was adopted in lines 115-116 and 122-124, but not in the remaining `const` blocks. Personally, I do prefer the previous style, namely `<type name>_<VALUE>`, but the point here is that we should whenever possible, stick to a single standard.

types/evidence.go
```diff
@@ -80,7 +80,7 @@ func NewDuplicateVoteEvidence(vote1, vote2 *Vote, blockTime time.Time, valSet *V
 // ABCI returns the application relevant representation of the evidence
 func (dve *DuplicateVoteEvidence) ABCI() []abci.Misbehavior {
 	return []abci.Misbehavior{{
-		Type: abci.MisbehaviorType_DUPLICATE_VOTE,
+		Type: abci.MISBEHAVIOR_TYPE_DUPLICATE_VOTE,
```
cason, line 83: As already commented for file `abci/types/types.go`, I prefer the previous style for those fields. Which also has the advantage to require less renaming.

cason (commented):
I don't have much context on protobufs, and the multiple versions of our API, so I can't really provide much feedback here.

One thing that was not really clear to me is what the standard we are adopting for the names of the fields/constants renamed or reorganized by this PR.

abci/types/types.go
mzabaluev: Some of the `TypeName_VALUE` naming is generated for enums nested in message definitions. I did not change that yet, although the style guidelines suggest consistently prefixing also these enum values with the enum name.
The `TypeName_ENUM_NAME_VALUE` would make those names annoyingly long.

abci/types/types.go
```diff
@@ -112,22 +112,22 @@ func RegisterABCIServer(s grpc.Server, srv ABCIServer) {
 type CheckTxType = v1.CheckTxType
 
 const (
-	CheckTxType_New     CheckTxType = v1.CheckTxType_New
-	CheckTxType_Recheck CheckTxType = v1.CheckTxType_Recheck
+	CHECK_TX_TYPE_NEW     CheckTxType = v1.CHECK_TX_TYPE_NEW
```
mzabaluev, line 115: As discussed in #736, this cannot be renamed to `CHECK_TX_TYPE_UNSPECIFIED` due to prior usage. This means the enum cannot be extended in the future, as the newly added values will be mapped to 0 by protobuf bindings that have not been updated.

mzabaluev:
> One thing that was not really clear to me is what the standard we are adopting for the names of the fields/constants renamed or reorganized by this PR.

It's as recommended by the [protobuf style guide](https://protobuf.dev/programming-guides/style/#enums): the value names are in capitals with underscores, and the enum name forms a common prefix for its value names (the Rust code generator knows to erase the prefix, so the end result looks sensible).

sergio-mena (commented):
I went through the changes and I have no detailed feedback.
However there are several points I'd like to make:
* I see you are modifying old versions of the protobufs. I don't think we can do that. That would be very confusing because those protobufs, in that form, wouldn't represent any real usage, in any real CometBFT version.
* Pushing this argument, I don't think it is a good idea either to modify the latest versions of the protobuf. Bear in mind those are already in alpha stage in `v0.38.x` and I am starting pre-release QA shortly. Also, several teams (Vega, SDK, etc) have started integration weeks ago.
* This brings me to suggest that the modifications you are proposing should represent a _new_ version of the protobufs in this feature branch.

On another topic, I see you are also renaming consts and enums defined in non-autogenerated Go code. I think this is OK, but since they are public (API breaking), this PR should have a changelog entry (or a paragraph in `UPGRADING.md`) explaining these changes.

mzabaluev:
> I see you are modifying old versions of the protobufs.

The idea is that there are no "old" versions before the feature branch is merged and a CometBFT release is made (including on buf.build) featuring the new package names. So we are free to rename things there, like we already have done retroactively for some fields. The versions, at this point, only serve to separate the protocol releases that may have supported implementations in the field (e.g. `abci.v2` corresponds to 0.37, if you need to make sure you are not using any newer definitions).

sergio-mena:
I still feel the heavy renaming of enums proposed here will add confusion to users of existing releases. The changes in the `.proto` files of your previous PR (introducing protobuf versioning) were inevitable, because they are needed for the versioning itself. But now that we have versions of protobufs, I would avoid introducing further discrepancies between the protos as they are in this branch after your versioning PR and the `.proto` files still existing on each of the release branches.
While if might be convenient for developers of external systems or tools, it will add noise for those working with several release branches of CometBFT at the same time (including us, but not only).

sergio-mena:
Is the current version of this PR reflecting the outcome of the discussion we had a couple of weeks ago?
(enum renaming being part of the latest version only)

mzabaluev:
This will be superseded by a new PR, which is waiting on #971.

PR #771, opened 2023-04-27
v0.37 pubsub: Handle big ints

### LINKED ISSUE #726: Using pubsub query with big values results in an int64 overflow
### Setup

**Environment**:
- **Install tools**:
- **Others**:

### What happened?

```
```

### How to reproduce it

### Logs

See above

jmalicevic:
Closes #726 for 0.37.x.

This is a draft PR. The main goal is to  make sure that big integers are parsed properly without a loss of precision. Int64 values will be processed as before. 
It also fixes the old logic where ints and floats were treated the same if a float had a fraction due to the fact that the float was converted to an int before comparison. 

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/.changelog/unreleased/bug-fixes/771-kvindexer-parsing-big-ints.md
+++ b/.changelog/unreleased/bug-fixes/771-kvindexer-parsing-big-ints.md
@@ -1,3 +1,2 @@
-- `[state/kvindex]` Querying event attributes that are bigger than int64 is now enabled as well as properly converting float event values 
-from the database into integer values, even if the float is bigger than an int64. 
+- `[state/kvindex]` Querying event attributes that are bigger than int64 is now enabled. We are not supporting reading floats from the db into the indexer  nor parsing them into BigFloats to not introduce breaking changes in minor releases.
   ([\#771](https://github.com/cometbft/cometbft/pull/771))
--- a/.changelog/unreleased/bug-fixes/771-kvindexer-parsing-big-ints.md
+++ b/.changelog/unreleased/bug-fixes/771-kvindexer-parsing-big-ints.md
@@ -1,0 +1,3 @@
+- `[state/kvindex]` Querying event attributes that are bigger than int64 is now enabled as well as properly converting float event values 
+from the database into int64 values ,even if the float is bigger than an int64. 
+  ([\#771](https://github.com/cometbft/cometbft/pull/771))
--- a/.changelog/unreleased/bug-fixes/771-kvindexer-parsing-big-ints.md
+++ b/.changelog/unreleased/bug-fixes/771-kvindexer-parsing-big-ints.md
@@ -1,3 +1,3 @@
 - `[state/kvindex]` Querying event attributes that are bigger than int64 is now enabled as well as properly converting float event values 
-from the database into int64 values ,even if the float is bigger than an int64. 
+from the database into integer values, even if the float is bigger than an int64. 
   ([\#771](https://github.com/cometbft/cometbft/pull/771))
--- a/libs/pubsub/query/query.go
+++ b/libs/pubsub/query/query.go
@@ -480,6 +480,8 @@ func matchValue(value string, op Operator, operand reflect.Value) (bool, error) 
 				// if it is a number bigger than int64, it will not be parsed properly.
 				// If we use bigFloat and convert that to a string, the values will be rounded which
 				// is not what we want either.
+				// Here we are simulating the behaviour that int64(floatValue). This was the default behaviour
+				// before introducing BigInts and we do not want to break the logic in minor releases.
 				_, ok := v.SetString(strings.Split(filteredValue, ".")[0], 10)
 				if !ok {
 					return false, fmt.Errorf("failed to convert value %s from float to big int", filteredValue)
--- a/libs/pubsub/query/query.go
+++ b/libs/pubsub/query/query.go
@@ -468,45 +468,42 @@ func matchValue(value string, op Operator, operand reflect.Value) (bool, error) 
 		if reflect.TypeOf(operand.Interface()) != reflect.TypeOf(i) {
 			break
 		}
-
+		noFrac := true
 		filteredValue := numRegex.FindString(value)
+		operandVal := operand.Interface().(*big.Int)
 		var cmpRes int
+		v := new(big.Int)
 		if strings.ContainsAny(filteredValue, ".") {
-			floatVal := new(big.Float)
-			_, ok := floatVal.SetString(operand.Interface().(*big.Int).String())
-			if !ok {
-				return false, fmt.Errorf("failed to convert value %v from event attribute to float64", filteredValue)
+			v1, err := strconv.ParseFloat(filteredValue, 64)
+			if err != nil {
+				return false, fmt.Errorf("failed to convert value %v from event attribute to float64: %w", filteredValue, err)
 			}
-			v := new(big.Float)
-			v, ok = v.SetString(filteredValue)
-			if !ok {
-				return false, fmt.Errorf("failed to convert value %v from event attribute to float64", filteredValue)
+			if _, frac := math.Modf(v1); frac != 0 {
+				noFrac = false // the numbers cannot be equal if the floating point has anything else than 0 as fraction
 			}
-			cmpRes = floatVal.Cmp(v)
+			v = big.NewInt(int64(v1))
 		} else {
-			operandVal := operand.Interface().(*big.Int)
-			// try our best to convert value from tags to int64
-			v := new(big.Int)
 
-			v, ok := v.SetString(filteredValue, 10)
+			// try our best to convert value from tags to int64
+			_, ok := v.SetString(filteredValue, 10)
 
 			if !ok {
 				return false, fmt.Errorf("failed to convert value %v from event attribute to big int", filteredValue)
 			}
-			cmpRes = operandVal.Cmp(v)
-		}
 
+		}
+		cmpRes = operandVal.Cmp(v)
 		switch op {
 		case OpLessEqual:
-			return cmpRes == 0 || cmpRes == 1, nil
+			return (cmpRes == 0 && noFrac) || cmpRes == 1, nil
 		case OpGreaterEqual:
-			return cmpRes == 0 || cmpRes == -1, nil
+			return (cmpRes == 0 && noFrac) || cmpRes == -1, nil
 		case OpLess:
 			return cmpRes == 1, nil
 		case OpGreater:
-			return cmpRes == -1, nil
+			return cmpRes == -1 || (cmpRes == 0 && !noFrac), nil
 		case OpEqual:
-			return cmpRes == 0, nil
+			return cmpRes == 0 && noFrac, nil
 		}
 
 	case reflect.Int64:
--- a/libs/pubsub/query/query.go
+++ b/libs/pubsub/query/query.go
@@ -10,7 +10,6 @@ package query
 
 import (
 	"fmt"
-	"math"
 	"math/big"
 	"reflect"
 	"regexp"
@@ -288,8 +287,8 @@ func (q *Query) Matches(events map[string][]string) (bool, error) {
 				value, err := strconv.ParseFloat(number, 64)
 				if err != nil {
 					err = fmt.Errorf(
-						"got %v while trying to parse %s as float64 (should never happen if the grammar is correct)",
-						err, number,
+						"problem parsing %s as big float (should never happen if the grammar is correct)",
+						number,
 					)
 					return false, err
 				}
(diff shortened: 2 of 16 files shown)


state/indexer/block/kv/kv_test.go
```diff
@@ -249,26 +249,14 @@ func TestBlockIndexerMulti(t *testing.T) {
 			q:       query.MustParse("block.height = 1"),
 			results: []int64{1},
 		},
-		"query return all events from a height - exact - no match.events": {
```
jmalicevic, line 252: These are just duplicate tests, piggy backing their removal onto this PR. We already did this for 0.38 and main

libs/pubsub/query/query_test.go
```diff
@@ -11,6 +12,47 @@ import (
...
+		matches  bool
+		matchErr bool
+	}{
+
+		{"account.balance <= 10000000000000000000", map[string][]string{"account.balance": {"10000000000000000000"}}, false, true, false},
+		{"account.balance <= 10000000000000000000", map[string][]string{"account.balance": {"10000000000000000000.0"}}, false, true, false},
+		{"account.balance <= 20000000000000000000", map[string][]string{"account.balance": {"10000000000000000000"}}, false, true, false},
+		{"account.balance <= 10000000000000000000", map[string][]string{"account.balance": {"10000000000000000001"}}, false, false, false},
+		{"account.balance <= 10000000000000000002", map[string][]string{"account.balance": {"10000000000000000000.6"}}, false, true, false},
+		{"account.balance > 10000000000000000000", map[string][]string{"account.balance": {"10000000000000000000.6"}}, false, true, false},
```
lasarojc, line 29: Could you add the following checks for negative values?

```suggestion
		{"account.balance > 10000000000000000000", map[string][]string{"account.balance": {"10000000000000000000.6"}}, false, true, false},
				{"account.balance <= -10000000000000000000", map[string][]string{"account.balance": {"-10000000000000000000"}}, false, true, false},
		{"account.balance <= -10000000000000000000", map[string][]string{"account.balance": {"-10000000000000000000.0"}}, false, true, false},
		{"account.balance <= -20000000000000000000", map[string][]string{"account.balance": {"-10000000000000000000"}}, false, false, false},
		{"account.balance <= -10000000000000000000", map[string][]string{"account.balance": {"-10000000000000000001"}}, false, true, false},
		{"account.balance <= -10000000000000000002", map[string][]string{"account.balance": {"-10000000000000000000.6"}}, false, false, false},
		{"account.balance > -10000000000000000000", map[string][]string{"account.balance": {"-10000000000000000000.6"}}, false, false, false},

```

sergio-mena, line 29: Does this also pass?
```
{"account.balance > 10000000000000000000", map[string][]string{"account.balance": {"10000000000000000000.1"}}, false, true, false},
```

libs/pubsub/query/query_test.go
```diff
@@ -33,6 +75,7 @@ func TestMatches(t *testing.T) {
 		{"tx.gas > 7 AND tx.gas < 9", map[string][]string{"tx.gas": {"8"}}, false, true, false},
 		{"body.weight >= 3.5", map[string][]string{"body.weight": {"3.5"}}, false, true, false},
 		{"account.balance < 1000.0", map[string][]string{"account.balance": {"900"}}, false, true, false},
+
```
sergio-mena, line 78: ?

sergio-mena (approved):
LGTM. I payed a lot of attention, but added some questions on parts I didn't fully understand.

libs/pubsub/query/query_test.go
```diff
@@ -11,6 +12,47 @@ import (
...
+	for _, tc := range testCases {
+		q, err := query.New(tc.s)
+		if !tc.err {
+			require.Nil(t, err)
+		}
+		require.NotNil(t, q, "Query '%s' should not be nil", tc.s)
+
+		if tc.matches {
+			match, err := q.Matches(tc.events)
```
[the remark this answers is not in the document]
jmalicevic: Weirdly enough, `Conditions()` is called only from the indexer but I could add a few UTs to make sure the conditions are parsed properly.

libs/pubsub/query/query.go
```diff
@@ -452,42 +457,54 @@ func matchValue(value string, op Operator, operand reflect.Value) (bool, error)
...
-		var v int64
+	case reflect.Pointer:
+
+		switch operand.Interface().(type) {
+		case *big.Int:
+			filteredValue := numRegex.FindString(value)
+			operandVal := operand.Interface().(*big.Int)
+			var cmpRes int
+			v := new(big.Int)
+			if strings.ContainsAny(filteredValue, ".") {
```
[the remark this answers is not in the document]
jmalicevic: This is the value read from the DB so it could be anything.

libs/pubsub/query/query.go
```diff
@@ -452,42 +457,54 @@ func matchValue(value string, op Operator, operand reflect.Value) (bool, error)
...
 
-	case reflect.Int64:
-		var v int64
+	case reflect.Pointer:
+
+		switch operand.Interface().(type) {
+		case *big.Int:
+			filteredValue := numRegex.FindString(value)
+			operandVal := operand.Interface().(*big.Int)
+			var cmpRes int
```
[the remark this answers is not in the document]
jmalicevic: Ah nice catch. This was left over from the previous commit where I had two different cases and called Cmp() in both the if and else so I needed the declaration first. fixed.

libs/pubsub/query/query_test.go
jmalicevic: No it does not because the `10000000000000000000.1` will be converted to `10000000000000000000.1`.

jmalicevic: Looking at the error I get when I run this and at the rules of the [query grammar](https://github.com/cometbft/cometbft/blob/jasmina/726-pubsub-int-overflow/libs/pubsub/query/query.peg), it seems that negative numbers in the query are not allowed.

state/indexer/query_range.go
```diff
@@ -44,6 +45,9 @@ func (qr QueryRange) LowerBoundValue() interface{} {
 	switch t := qr.LowerBound.(type) {
 	case int64:
 		return t + 1
```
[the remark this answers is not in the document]
jmalicevic: I thought it was needed for the height but no, you are right, we always call this with a bigInt.

libs/pubsub/query/query_test.go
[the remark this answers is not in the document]
jmalicevic: Added new tests specifically for this.

libs/pubsub/query/query_test.go
jmalicevic: I will bring this up on Monday, it seems that the DB allows event values to be stored  as floats. But the indexer queries always get the event value and convert it an int. So you end up with this particular event not being detected. 
We can leave this behaviour for old versions but it feels incorrect. Converting to int might work for most cases but in this particular cases, we need to detect that this number is a float, then parse it and check that there is a decimal part if the non decimal part is the same. I think for main we should 100% do this. But I want us to reach consensus on whether we want to do this for 0.34/0.37

lasarojc: Thanks for checking. Indeed it does not seem like it. There is no limitation on the data itself, right? Is this a bug?

jmalicevic: You are right, you could technically submit anything as a tx event I believe. But querying is not supported. I am not sure tbh...in does seem wrong to disable querying something that can be provided as input :/

sergio-mena: From today's discussion, I understand that [my initial comment](https://github.com/cometbft/cometbft/pull/771#discussion_r1180588069) does not apply to `v0.37.x`/`v0.34.x` (which this PR is targeting), isn't it?

jmalicevic: Reverting this commit means that if there is a float in the DB, it will be converted to its value in int64.

state/indexer/block/kv/kv.go
```diff
@@ -293,10 +294,17 @@ LOOP:
...
+			v := new(big.Int)
+			v, ok := v.SetString(eventValue, 10)
+			if !ok {
+
+				v_f, err := strconv.ParseFloat(eventValue, 64)
+				if err != nil {
+					continue LOOP
+				}
+				v = new(big.Int)
+				v = v.SetInt64(int64(v_f))
```
sergio-mena, line 307: Two questions:
1) This part was not present in the original. Do we need it?
2) Should `ParseFloat` be replaced by a bog float?

jmalicevic: Hey, no it was not, this case basically covers the functionality we discussed today that we decided to ignore. This is when an event is stored in the DB as a float and it will not be parsed as BigInt (the `ok` in line 298 will be false). To mimic the original behaviour in 0.34 (before introducing BigInts), we need to still somehow parse the float into at least the int64. This same reasoning applies to the change in [this part ](https://github.com/cometbft/cometbft/blob/1dd27a7cb11df703065a29b173ed53b2e3cd374e/libs/pubsub/query/query.go#L476)of the code . Note that there I do use a big float. And I am actually considering now what the best approach is. Because using the BigFloat used in `query.go` rounds the float (which is not done with a simple cast of float64 to int64), but using a simple cast will not parse BigFloats and ints properly. Trying to find a solution and will mark the PR as draft and re-open it to avoid confusion.

sergio-mena: I think what you suggested in slack would be the simplest (to be consistent with the current behavior in `v0.34.x`/`v0.37.x`): if there is a dot, just split the string and use the whole number part

state/indexer/block/kv/kv.go
```diff
@@ -294,17 +295,21 @@ LOOP:
...
 		}
 
-		if _, ok := qr.AnyBound().(*big.Int); ok { // This means that in a range we can only have integers, floats will be ignored
+		if _, ok := qr.AnyBound().(*big.Int); ok {
 			v := new(big.Int)
 			v, ok := v.SetString(eventValue, 10)
-			if !ok {
-
-				v_f, err := strconv.ParseFloat(eventValue, 64)
+			if !ok { // If the number was not int it will might be a float so we get the int value
```
sergio-mena, line 301:
```suggestion
			if !ok { // If the number was not int it might be a float so we get the int value
```

libs/pubsub/query/query.go
```diff
@@ -452,42 +457,57 @@ func matchValue(value string, op Operator, operand reflect.Value) (bool, error)
...
-		operandInt := operand.Interface().(int64)
-		filteredValue := numRegex.FindString(value)
+		switch operand.Interface().(type) {
+		case *big.Int:
+			filteredValue := numRegex.FindString(value)
+			operandVal := operand.Interface().(*big.Int)
+			v := new(big.Int)
+			if strings.ContainsAny(filteredValue, ".") {
+				// We do this just to check whether the string can be parsed as a float
+				_, err := strconv.ParseFloat(filteredValue, 64)
```
sergio-mena, line 469: What if we get a very big float (e.g., 600000000000000000000000000000000000000000000000000000001), shouldn't we still parse it as a BigFloat (to make sure we don't lose the last 1) ?

libs/pubsub/query/query.go
```diff
@@ -452,42 +457,57 @@ func matchValue(value string, op Operator, operand reflect.Value) (bool, error)
...
-			v1, err := strconv.ParseFloat(filteredValue, 64)
-			if err != nil {
-				return false, fmt.Errorf("failed to convert value %v from event attribute to float64: %w", filteredValue, err)
-			}
+				// If yes, we get the int part of the  string.
+				// We could simply cast the float to an int and use that to create a big int but
+				// if it is a number bigger than int64, it will not be parsed properly.
+				// If we use bigFloat and convert that to a string, the values will be rounded which
+				// is not what we want either.
+				_, ok := v.SetString(strings.Split(filteredValue, ".")[0], 10)
```
sergio-mena, line 483: Suggestion: I'd add something like "we are simulating int64(v1), which was here before this patch"

.changelog/unreleased/bug-fixes/771-kvindexer-parsing-big-ints.md
```diff
@@ -0,0 +1,3 @@
+- `[state/kvindex]` Querying event attributes that are bigger than int64 is now enabled as well as properly converting float event values 
+from the database into int64 values ,even if the float is bigger than an int64.
```
lasarojc, line 2:
```suggestion
from the database into int64 values, even if the float is bigger than an int64. 
```

lasarojc (approved):

state/indexer/query_range.go
```diff
@@ -67,7 +69,9 @@ func (qr QueryRange) UpperBoundValue() interface{} {
 	switch t := qr.UpperBound.(type) {
 	case int64:
 		return t - 1
-
+	case *big.Int:
```
sergio-mena, line 72: (from sync discussion) Do we need to delete 70-71 to be consistent with `LowerBoundValue`

jmalicevic: Reverted the deletion in `LowerBoundValue`

sergio-mena (approved):
🥳

libs/pubsub/query/query.go
```diff
@@ -152,16 +153,18 @@ func (q *Query) Conditions() ([]Condition, error) {
...
-				if err != nil {
-					err = fmt.Errorf(
-						"got %v while trying to parse %s as int64 (should never happen if the grammar is correct)",
-						err, number,
+				valueBig := new(big.Int)
+
+				valueBig, ok := valueBig.SetString(number, 10)
+				if !ok {
+					err := fmt.Errorf(
+						"problem parsing %s as bigint (should never happen if the grammar is correct)",
```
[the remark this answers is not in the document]
jmalicevic: As this is the default message in a bunch of other places and it actually makes sense - comparing the query to the query grammar, I would leave this?

PR #780, opened 2023-05-01
First version (ABCI only ATM) of documenting protobuf version translation

sergio-mena:
Contributes to #95

This is my first attempt at documenting:

* whether it makes sense to interpret a protobuf message from the previous version
* if it does make sense, how to translate the data found in the previous version into the current one

... and the same for interpreting a protobuf message from the current version while in the previous version.

This should be helpful for integrators that, e.g., need to work with two versions at the same time. They could take advantage of these annotations in order to construct domain types that would support the required versions of the protos.

I kind of invented some _keywords_ along the way, with the intention of keeping the text concise. Those keywords should be self-explanatory. Let me know if that's not the case, and I'll add some explanations in the README.

Feedback is more than welcome. I have only done ABCI since I'd like to gather feedback to then proceed with the rest of the proto files.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/proto/cometbft/abci/v2/types.proto
+++ b/proto/cometbft/abci/v2/types.proto
@@ -17,6 +17,8 @@ import "google/protobuf/timestamp.proto";
 //----------------------------------------
 // Request types
 
+// Handling v1 as v2: REJECT
+// Handling v2 as v1: REJECT
 message Request {
   oneof value {
     v1.RequestEcho               echo                 = 1;
@@ -39,6 +41,11 @@ message Request {
   reserved 4;
 }
 
+
+// Handling v1 as v2: ACCEPT
+//     SET abci_version:v2 TO "<NOT PROVIDED>"
+// Handling v2 as v1: ACCEPT
+//     IGNORE abci_version:v2
 message RequestInfo {
   string version       = 1;
   uint64 block_version = 2;
@@ -46,6 +53,8 @@ message RequestInfo {
   string abci_version  = 4;
 }
 
+// Handling v1 as v2: REJECT
+// Handling v2 as v1: REJECT
 message RequestInitChain {
   google.protobuf.Timestamp time = 1
       [(gogoproto.nullable) = false, (gogoproto.stdtime) = true];
@@ -56,6 +65,8 @@ message RequestInitChain {
   int64                       initial_height         = 6;
 }
 
+// Handling v1 as v2: REJECT
+// Handling v2 as v1: REJECT
 message RequestBeginBlock {
   bytes                    hash                 = 1;
   cometbft.types.v1.Header header               = 2 [(gogoproto.nullable) = false];
@@ -63,6 +74,8 @@ message RequestBeginBlock {
   repeated Misbehavior     byzantine_validators = 4 [(gogoproto.nullable) = false];
 }
 
+// Handling v1 as v2: REJECT
+// Handling v2 as v1: REJECT
 message RequestPrepareProposal {
   // the modified transactions cannot exceed this size.
   int64 max_tx_bytes = 1;
@@ -78,6 +91,8 @@ message RequestPrepareProposal {
   bytes proposer_address = 8;
 }
 
+// Handling v1 as v2: REJECT
+// Handling v2 as v1: REJECT
 message RequestProcessProposal {
   repeated bytes       txs                  = 1;
   CommitInfo           proposed_last_commit = 2 [(gogoproto.nullable) = false];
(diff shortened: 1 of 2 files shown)


proto/cometbft/abci/v2/types.proto
```diff
@@ -17,6 +17,8 @@ import "google/protobuf/timestamp.proto";
 //----------------------------------------
 // Request types
 
+// Handling v1 as v2: REJECT
```
thanethomson, line 20: Just to clarify, does this mean that, if we receive `v1.Request`, but we're expecting `v2.Request`, we reject it?

sergio-mena: Exactly. And "reject" means:

* either "it doesn't make sense" (this is the case here)
* or it is an error condition

Let me know if you think it is worth distinguishing between REJECT and ERROR

thanethomson (approved):
I'd be interested to know what @mzabaluev thinks of this convention.

mzabaluev (changes requested):
In light of the above comment, I don't think most of this information is applicable in practice. The compatibility comments should be limited to handling situations when a field added in a newer version of the message is missing or set to the zero value (these conditions are not practically distinguishable by the proto consumer), possibly meaning that the sender used an older version to encode.

sergio-mena:
@mzabaluev I mostly agree with your reasoning, however your reasoning is focusing on marshalling/unmarshalling of protobufs. This PR is not really about that, but rather domain knowledge a developer needs if they are not expert in those protobufs. For instance, it would help someone writing code that handles more than one version at once.

I believe that's precisely what you're doing in informalsystems/tendermint-rs#1312. Further, [these lines](https://github.com/cometbft/cometbft/blob/b8b6b9dcf/proto/cometbft/abci/v3/types.proto#L265-L278) contain the information you had to ask over Slack a few days ago (and rightly so, because the info _was_ missing).

My impression is that the word "handle" used in the text added in this PR is somewhat misleading. Maybe:
* instead of "Handling v1 as v2"
* it should read "Translating v1 into v2" ?

I guess it's best if we discuss this synchronously.

mzabaluev:
Yes, the added comments should perhaps provide a migration guide between versions, in cases when just following the usual protobuf rules is not enough (if you can make it good by just filling the newly added fields with zero values, no additional information is needed). In the present form, it's not clear to me how the application developer should make use of the provided rules.

mzabaluev:
In the specific case of ABCI, the application should be instructed to handle `Request` as per the expected version.
With the changed API, applications built for 0.37 cannot be expected to work correctly with a 0.38 node, correct?

sergio-mena:
> With the changed API and semantics, applications built for 0.37 cannot be expected to work correctly with a 0.38 node and vice versa, correct?

Exactly. That's what all the REJECTs I wrote for those mean... the REJECT means "don't even try to mix up versions of these messages".

But there's more: the example I gave you is **missing info** with out this PR. Also, some messages were renamed, and without this PR, it is not obvious for someone examining the various versions for ABCI protobufs.

proto/cometbft/abci/v3/types.proto
```diff
@@ -223,13 +262,39 @@ message ResponseFinalizeBlock {
 //----------------------------------------
 // Blockchain Types
 
+// Handling v1 as v3: ACCEPT
```
sergio-mena, line 265: Differences from v1 to v3

proto/cometbft/abci/v2/types.proto
```diff
@@ -39,13 +41,20 @@ message Request {
   reserved 4;
 }
 
+
+// Handling v1 as v2: ACCEPT
+//     SET abci_version:v2 TO "<NOT PROVIDED>"
```
sergio-mena, line 46:
```suggestion
//     follow normal protobuf rules
```

sergio-mena: (from discussionwith @mzabaluev): Consider removing this, as this is considered as the default (so only document the "REJECT"s as counter-intuitive

proto/cometbft/abci/v2/types.proto
```diff
@@ -39,13 +41,20 @@ message Request {
...
+// Handling v2 as v1: ACCEPT
+//     IGNORE abci_version:v2
 message RequestInfo {
   string version       = 1;
   uint64 block_version = 2;
   uint64 p2p_version   = 3;
   string abci_version  = 4;
 }
 
+// Handling v1 as v2: REJECT
```
sergio-mena, line 56:
```suggestion
// Handling v1 as v2: REJECT  -- [explain why] 
```
everywhere

sergio-mena:
We decided not to pursue this approach

PR #783, opened 2023-05-02
Unsafe int cast in `kill` command

sergio-mena:
This addresses a [warning](https://github.com/cometbft/cometbft/security/code-scanning/2) from CodeQL

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/cmd/cometbft/commands/debug/kill.go
+++ b/cmd/cometbft/commands/debug/kill.go
@@ -3,6 +3,7 @@ package debug
 import (
 	"errors"
 	"fmt"
+	"math"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -123,11 +124,15 @@ func killProc(pid uint64, dir string) error {
 	go func() {
 		// Killing the CometBFT process with the '-ABRT|-6' signal will result in
 		// a goroutine stacktrace.
-		p, err := os.FindProcess(int(pid))
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "failed to find PID to kill CometBFT process: %s", err)
-		} else if err = p.Signal(syscall.SIGABRT); err != nil {
-			fmt.Fprintf(os.Stderr, "failed to kill CometBFT process: %s", err)
+		if pid <= math.MaxInt {
+			p, err := os.FindProcess(int(pid))
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "failed to find PID to kill CometBFT process: %s", err)
+			} else if err = p.Signal(syscall.SIGABRT); err != nil {
+				fmt.Fprintf(os.Stderr, "failed to kill CometBFT process: %s", err)
+			}
+		} else {
+			fmt.Fprintf(os.Stderr, "failed to convert PID to int (%v > %v)", pid, math.MaxInt)
 		}
 
 		// allow some time to allow the CometBFT process to be killed
--- a/cmd/cometbft/commands/debug/kill.go
+++ b/cmd/cometbft/commands/debug/kill.go
@@ -3,7 +3,6 @@ package debug
 import (
 	"errors"
 	"fmt"
-	"math"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


adizere (commented):
I tried reproducing the behavior. It killed my `iterm` window. Will try again...

cmd/cometbft/commands/debug/kill.go
```diff
@@ -123,11 +124,15 @@ func killProc(pid uint64, dir string) error {
 	go func() {
 		// Killing the CometBFT process with the '-ABRT|-6' signal will result in
 		// a goroutine stacktrace.
-		p, err := os.FindProcess(int(pid))
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "failed to find PID to kill CometBFT process: %s", err)
-		} else if err = p.Signal(syscall.SIGABRT); err != nil {
-			fmt.Fprintf(os.Stderr, "failed to kill CometBFT process: %s", err)
+		if pid <= math.MaxInt {
```
greg-szabo, line 127: Honestly, I'm scratching my head at this. I think the killProc function should already only accept int, if processes are stored as int. Accepting uint64 because that's a better validation seems counterintuitive, if it is then converted again.

OS processes should always be bigger than 0 for sure, but the high limit is dependent on the architecture. (32-bit vs 64-bit)

If FindProcess requires an int, I would make killProc require an int too. No need to make it more complex.

greg-szabo (changes requested):
I've left my note on how I would change this, if we touch it. (Maybe it's just not worth touching it right now. uint to int conversion should work fine.)

I'm more concerned about the actual input of the killProc function from line 36 (https://github.com/cometbft/cometbft/blob/feac56b6297eb0dd800d5e35b9c04a848758c612/cmd/cometbft/commands/debug/kill.go#L36). I thought Cobra's args[0] contains the application name. Converting that to uint will not make it a pid. (But most probably zero.)

That might explain also why Adi's test closed his iTerm. Or maybe I'm completely wrong.

cmd/cometbft/commands/debug/kill.go
sergio-mena: > I think the killProc function should already only accept int

TBH, I was hesitating whether to add the limit check, or go all the way until the CLI param... and went for the former. Main reason: limiting code changes on what will likely be backported to all branched.

Now that I see you're leaning towards the other solution, let's go for it. Bear with me

adizere (approved):
Seems to work, thanks Sergio!


```
% ./build/cometbft debug kill 999999999999999999999 /tmp/a --rpc-laddr http://localhost:27030 --home ~/.gm/node0
ERROR: strconv.Atoi: parsing "999999999999999999999": value out of range
```

adizere:
> I tried reproducing the behavior. It killed my `iterm` window. Will try again...

I managed to reproduce this.

The trick is to pass to `os.FindProcess(pid)` a value of `pid == math.MaxInt`. __Do not run this on local workstation, it reliably kills all open applications__:


```
% ./build/cometbft debug kill 9223372036854775807 /tmp/a --home ~/.gm/node0/ --rpc-laddr http://localhost:27030
```

Or the alterantive, simple recipe:

```go
package main

import (
	"fmt"
	"os"
	"syscall"
)

func main() {
	fmt.Printf("searching PID")
	p, err := os.FindProcess(9223372036854775806)
	fmt.Printf("found PID test %d | %v", p, err)
	if err != nil {
			fmt.Fprintf(os.Stderr, "failed to find PID to kill CometBFT process: %s", err)
		} else if err = p.Signal(syscall.SIGABRT); err != nil {
			fmt.Fprintf(os.Stderr, "failed to kill CometBFT process: %s", err)
		}
	fmt.Printf("after test %d | %v", p, err)
}
```

---

I don't have time to understand why this happens right now. Not sure we should do anything about it in Comet.

sergio-mena:
@adizere Thanks for the info, I also reproed it (you could see it with your eyes :-D).

greg-szabo (approved):

PR #784, opened 2023-05-02
Struct `Client` exposes sensitive data

sergio-mena:
This addresses a [warning](https://github.com/cometbft/cometbft/security/code-scanning/1) from CodeQL

Struct `Client` defined [here](https://github.com/cometbft/cometbft/blob/feac56b62/rpc/jsonrpc/client/http_json_client.go#L124-L133) is used in some print/log functions.
As it contains a "password" field, we need to make sure it is not printed. This PR addresses that problem by having `Client` implement `fmt.Stringer`.

---

#### PR checklist

- [ ] Tests written/updated


adizere (approved):
Pre-approve based on reading the code. Have not tested yet.

lasarojc (approved):

PR #791, opened 2023-05-04
[pubsub/indexer] Support for bigInt and bigFloat

### LINKED ISSUE #726: Using pubsub query with big values results in an int64 overflow

### LINKED ISSUE #769: pubsub: Properly parse big numbers
### Setup

0.38.x, main, 
yes

### What happened?
The performance optimizations introduced with tendermint/tendermint#7319 parse all numbers occurring in a query using `ParseFloat`, including integers. While this solves the problem of parsing big integers, it might lead to a loss of precision. 

One way to achieve this is introduce a `TFNumber` and `TIntNumber` in `query/syntax/scanner` which would replace the current `TNumber` type. The function `scanNumber` already detects potential `.` in the number. 

Integers are parsed using either `ParseInt` or parsed as big integers (for example by using `math/big`).

jmalicevic:
Closes #769 

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/libs/pubsub/query/query.go
+++ b/libs/pubsub/query/query.go
@@ -249,6 +249,8 @@ var opTypeMap = map[syntax.Token]map[syntax.Token]func(interface{}) func(string)
 		syntax.TNumber: func(v interface{}) func(string) bool {
 			return func(s string) bool {
 				w, err := parseNumber(s)
+				vval := v.(*big.Float)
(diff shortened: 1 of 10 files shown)


jmalicevic:
Superseded by #797 .

Issue #55, opened 2022-12-23
Update the e2e runner to be able to start and stop CometBFT testnet processes on non-local machines


sergio-mena:
Tasks:

- [x]  Familiarize with PR tendermint/tendermint#9801
- [x]  Clone it to Comet (merge to `main`)
- [x] Drive it to completion, including any non-CI testing needed
  - [x] Refactor to enable start and stop (#796)
  - [x] Add the DO infra and test
    - #846
    - cometbft/qa-infra#16

DoD:

- The Comet PR (cloned from tendermint/tendermint#9801), and subsequent PRs, are approved, tested, and merged

Original issue: tendermint/tendermint#9790

sergio-mena:
Closed by #796, #846, and cometbft/qa-infra#16

PR #796, opened 2023-05-05
Extend the infrastructure provider with `StartNodes` and `StopTestnet`

sergio-mena:
Contributes to #55

This PR is reusing tendermint/tendermint#9801 and driving it to completion:
* Two new methods added to infrastructure provider with `StartNodes` and `StopTestnet`
* Refactored the docker provider and the runner logic to use these two interface methods
* Prepared a new provider, `digitalocean` to implement these methods in a subsequent PR
* Distingush between internal and external IP
  * on DO, they will be the same
  * on Docker, the external IP will be `localhost` for all nodes (the proxy address)

---

#### PR checklist

- [ ] Tests written/updated


test/e2e/runner/start.go
```diff
@@ -102,15 +119,16 @@ func Start(ctx context.Context, testnet *e2e.Testnet) error {
 
 		logger.Info("Starting catch up node", "node", node.Name, "height", node.StartAt)
 
-		if err := execCompose(testnet.Dir, "up", "-d", node.Name); err != nil {
+		err := p.StartNodes(context.Background(), node)
```
lasarojc, line 122: in the previous loop, logically, we should first `waitForHeight(... node.StartAt ...)` and then update `networkHeight`.

test/e2e/runner/main.go
```diff
@@ -186,7 +196,7 @@ func NewCLI() *CLI {
 			if err != nil {
 				return err
 			}
-			return Start(cmd.Context(), cli.testnet)
+			return Start(cmd.Context(), cli.testnet, cli.infp)
```
lasarojc, line 199: in the short description, above, update to include DO?

test/e2e/runner/main.go
```diff
@@ -211,7 +221,7 @@ func NewCLI() *CLI {
 		Short: "Stops the Docker testnet",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			logger.Info("Stopping testnet")
-			return execCompose(cli.testnet.Dir, "down")
+			return cli.infp.StopTestnet(context.Background())
```
lasarojc, line 224: in the short description above, update to include DO?

test/e2e/runner/main.go
```diff
@@ -266,15 +276,15 @@ func NewCLI() *CLI {
 		Use:   "logs",
 		Short: "Shows the testnet logs",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return execComposeVerbose(cli.testnet.Dir, "logs")
+			return docker.ExecComposeVerbose(context.Background(), cli.testnet.Dir, "logs")
```
lasarojc, line 279: will this eventually also work for DO?
maybe add a TODO?

test/e2e/runner/main.go
```diff
@@ -266,15 +276,15 @@ func NewCLI() *CLI {
...
 		},
 	})
 
 	cli.root.AddCommand(&cobra.Command{
 		Use:   "tail",
 		Short: "Tails the testnet logs",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return execComposeVerbose(cli.testnet.Dir, "logs", "--follow")
+			return docker.ExecComposeVerbose(context.Background(), cli.testnet.Dir, "logs", "--follow")
```
lasarojc, line 287: ditto

test/e2e/runner/perturb.go
```diff
@@ -28,7 +29,7 @@ func Perturb(ctx context.Context, testnet *e2e.Testnet) error {
 // after recovering.
 func PerturbNode(ctx context.Context, node *e2e.Node, perturbation e2e.Perturbation) (*rpctypes.ResultStatus, error) {
 	testnet := node.Testnet
-	out, err := execComposeOutput(testnet.Dir, "ps", "-q", node.Name)
+	out, err := docker.ExecComposeOutput(context.Background(), testnet.Dir, "ps", "-q", node.Name)
```
lasarojc, line 32: will this eventually work also for DO?

test/e2e/pkg/exec/exec.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// CommandOutput executes a shell command and return the command's output.
+func CommandOutput(ctx context.Context, args ...string) ([]byte, error) {
+	//nolint: gosec
+	// G204: Subprocess launched with a potential tainted input or cmd arguments
+	cmd := osexec.CommandContext(ctx, args[0], args[1:]...)
+	out, err := cmd.CombinedOutput()
+	switch err := err.(type) {
+	case nil:
```
lasarojc, line 23: is it just me or do you also find it weird that you can test a type against nil?

test/e2e/pkg/exec/exec.go
```diff
@@ -0,0 +1,40 @@
...
+	osexec "os/exec"
+)
+
+// Command executes a shell command.
+func Command(ctx context.Context, args ...string) error {
+	_, err := CommandOutput(ctx, args...)
+	return err
+}
+
+// CommandOutput executes a shell command and return the command's output.
```
lasarojc, line 16:
```suggestion
// CommandOutput executes a shell command and returns the command's output.
```

lasarojc (approved):
Thanks for driving this. It will be a huge improvement for our QA.

Made a few suggestions, but nothing blocking.

test/e2e/runner/perturb.go
thanethomson: I think the point of the PR was just to start putting in some of the initial plumbing to get the DO nodes to start/stop. The original intention was to expand this functionality with perturbations in follow-up PRs.

thanethomson (approved):

test/e2e/pkg/exec/exec.go
sergio-mena: This was coming from William's patch, so I assumed it was good.
I've just taken a look at our code base, and we do this quite often, e.g. [here,](https://github.com/cometbft/cometbft/blob/bd9d7e/light/client.go#L731) [here](https://github.com/cometbft/cometbft/blob/bd9d7e/evidence/reactor.go#L88), or [here](https://github.com/cometbft/cometbft/blob/4790ea3e46475064d5475c787427ae926c5a9e94/light/detector.go#L59).

sergio-mena: Seems to be a standard way to handle errors...

lasarojc: It seems to be one of Go's idiosyncrasies

test/e2e/runner/main.go
```diff
@@ -30,6 +31,7 @@ type CLI struct {
 	testnet  *e2e.Testnet
 	preserve bool
 	infp     infra.Provider
+	ifd      e2e.InfrastructureData
```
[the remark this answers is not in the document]
sergio-mena: Good point... let me see what I can do (I need to check if that's OK with the UTs below)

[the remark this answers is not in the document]
sergio-mena: I was a bit more involved than expected, but the result is worth it, I think

test/e2e/runner/main.go
sergio-mena: Good catch. To avoid having to come back to it at every subsequent PR, I just removed references to Docker

test/e2e/runner/main.go
sergio-mena: Removed reference to Docker

test/e2e/runner/perturb.go
sergio-mena: To add on to @thanethomson's comment, `PerturbNode` will eventually become a method in `infra.Provider`, whose Docker version will be + or - it current implementation; and we will have to come up with mechanisms for implementing it for a DO testnet.

test/e2e/runner/start.go
sergio-mena: Do you mean the loop starting at line 54? Those nodes start all at height zero.
I had to change the logic a bit here, to make the DO way of doing things compatible with Docker:
* Before: nodes that start at height zero were started one by one. That is very inefficient with a large testnet (think of the 200-node testnet)
* After: nodes starting at height zero are all started at the same time. Then we wait for all of them to be started (`waitForNode` with height zero)

PR #797, opened 2023-05-05
pubsub/kvindexer:support for big numbers - v2

### LINKED ISSUE #767: kvindexer: enable parsing event values which exceed the max value of `int64`
### Setup

0.34.x, 0.37.x, 0.38.x 

yes

### What happened?

For parsing event values within the kvindexer we use `parseInt` which will not parse values bigger than the maximum value of int64. As pointed out in #726, there are many chains using big denoms and we might end up in a situation where their data is not indexed properly. 

`ParseInt` was used only for range queries in Comet and Tendermint core versions prior to 0.34.27. v0.34.27+ uses `ParseInt` in many other places. 

If an event of the form: `sender.amount = 10000000000000000000` exists, a query for it should return this event. Currently, an error is produced.

### LINKED ISSUE #769: pubsub: Properly parse big numbers

jmalicevic:
Closes #767 
Contributes to #769

This is a second version of the implementation adding support for big numbers - integers and floats.

In this PRs all numbers are converted to Big.Floats but when converting integers, the floating point precision is set to the number of bits needed to represent the integer. 

The following other updates and fixes are included in this PR:

kvindexer: 
- events stored as floats have been ignored, and not read from the database
- query conditions could not contain floating point values nor big int values
- added logger to log parsing errors

pubsub:
- query conditions could not containg big int nor big float values
- very big ints and floats would still loose precision even when using `big.ParseFloat` as the default precision is 64. 

---

#### PR checklist

- [x] Tests written/updated


libs/pubsub/query/query.go
```diff
@@ -248,7 +255,9 @@ var opTypeMap = map[syntax.Token]map[syntax.Token]func(interface{}) func(string)
 		syntax.TNumber: func(v interface{}) func(string) bool {
 			return func(s string) bool {
 				w, err := parseNumber(s)
-				return err == nil && w == v.(float64)
+				vval := v.(*big.Float)
+				fmt.Println(vval.Text('f', 64))
```
sergio-mena, line 259: Remove?

libs/pubsub/query/query.go
```diff
@@ -268,7 +277,9 @@ var opTypeMap = map[syntax.Token]map[syntax.Token]func(interface{}) func(string)
 		syntax.TNumber: func(v interface{}) func(string) bool {
 			return func(s string) bool {
 				w, err := parseNumber(s)
-				return err == nil && w < v.(float64)
+				vval := v.(*big.Float)
+				fmt.Println(vval.Text('f', 64))
```
sergio-mena, line 281: Remove?

libs/pubsub/query/query.go
```diff
@@ -288,7 +299,9 @@ var opTypeMap = map[syntax.Token]map[syntax.Token]func(interface{}) func(string)
 		syntax.TNumber: func(v interface{}) func(string) bool {
 			return func(s string) bool {
 				w, err := parseNumber(s)
-				return err == nil && w <= v.(float64)
+				vval := v.(*big.Float)
+				fmt.Println(vval.Text('f', 64))
```
sergio-mena, line 303: Remove?

libs/pubsub/query/query.go
```diff
@@ -328,7 +341,9 @@ var opTypeMap = map[syntax.Token]map[syntax.Token]func(interface{}) func(string)
 		syntax.TNumber: func(v interface{}) func(string) bool {
 			return func(s string) bool {
 				w, err := parseNumber(s)
-				return err == nil && w >= v.(float64)
+				vval := v.(*big.Float)
+				fmt.Println(vval.Text('f', 64))
```
sergio-mena, line 345: Remove?

libs/pubsub/query/query_test.go
```diff
@@ -173,6 +173,84 @@ var apiTypeEvents = []types.Event{
 	},
 }
 
+var apiBigNumTest = map[string][]string{
```
sergio-mena, line 176: Consider embedding this inside the test case that uses it

libs/pubsub/query/syntax/parser.go
```diff
@@ -68,17 +67,35 @@ func (a *Arg) String() string {
...
+			return nil
+		}
+		return f
+	}
+	// If it is indeed a big integer, we make sure to convert it to a float with enough precision
+	// to represent all the bits
+	bitLen := uint(intVal.BitLen())
+	var f *big.Float
+	var err error
+	if bitLen < 64 {
```
sergio-mena, line 89: <= ?

libs/pubsub/query/syntax/parser.go
```diff
@@ -68,17 +67,35 @@ func (a *Arg) String() string {
...
+		f, _, err = big.ParseFloat(a.text, 10, 0, big.ToNearestEven)
+	} else {
+		f, _, err = big.ParseFloat(a.text, 10, uint(intVal.BitLen()), big.ToNearestEven)
```
sergio-mena, line 92:
```suggestion
		f, _, err = big.ParseFloat(a.text, 10, bitLen, big.ToNearestEven)
```

state/indexer/query_range.go
```diff
@@ -44,7 +45,17 @@ func (qr QueryRange) LowerBoundValue() interface{} {
 	switch t := qr.LowerBound.(type) {
 	case int64:
 		return t + 1
-
+	case *big.Int:
+		tmp := new(big.Int)
+		return tmp.Add(t, big.NewInt(1))
```
sergio-mena, line 50: Why are we adding/subtracting 1 in the first place?

state/indexer/query_range.go
```diff
@@ -182,7 +197,7 @@ func conditionArg(c syntax.Condition) interface{} {
 	}
 	switch c.Arg.Type {
 	case syntax.TNumber:
-		return int64(c.Arg.Number())
+		return c.Arg.Number()
```
sergio-mena, line 200: Above, you did a similar conversion like this: `c.Arg.Number().Int64()`
Shouldn't these be consistent?

internal/indexer/indexer_utils.go
```diff
@@ -0,0 +1,94 @@
...
+		// For float we cannot decrease the condition by 1 as it would be wrong
+		// example int: x < 100; upperBound = 99; if x.Cmp(99) == 0 the condition holds
+		// example float: x < 100.0; upperBound = 100.0; if x.Cmp(100) ==0 then returning x
+		// would be wrong. Thus we check whether the float upper bound should be included
+		// if cmp == 0 && !include {
+		// 	return 1, nil
+		// }
+		return vF.Cmp(opVal), true, nil
+	default:
+		return -1, false, fmt.Errorf("unable to parse arguments")
```
sergio-mena, line 56: Please add the args' contents

internal/indexer/indexer_utils.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+}
+
```
sergio-mena, line 59: Please add a comment here explaining why `int`s and `float`s have different conditions

docs/app-dev/indexing-transactions.md
```diff
@@ -276,3 +276,10 @@ This behavior was fixed with CometBFT 0.34.26+. However, if the data was indexed
 Tendermint Core and not re-indexed, that data will be queried as if all the attributes within a height
 occurred within the same event.
 
+## Event attribute value types
```
lasarojc, line 279:
```suggestion
## Event attribute value types

```

docs/app-dev/indexing-transactions.md
```diff
@@ -276,3 +276,10 @@ This behavior was fixed with CometBFT 0.34.26+. However, if the data was indexed
...
+Users can use anything as an event value. However, if the even attrbute value is a number, the following needs to be taken into accound:
+- Negative numbers will not be properly retrieved when querying the indexer.
+- Event values are converted to big floats( from the `big/math` package). The precision of the floating point number is set to the bit length 
+of the integer it is supposed to represent. This was not present before CometBFT v0.38.x and all float values were ignored. 
+- As of CometBFT v0.38.x, queries can contain floating point numbers as well.
+- Note that comparing to floats can be imprecise with a high number of decimals.
```
lasarojc, line 285:
```suggestion
## Event attribute value types

Users can use anything as an event value. However, if the event attribute value is a number, the following needs to be taken into accound:

- Negative numbers will not be properly retrieved when querying the indexer.
- Event values are converted to big floats( from the `big/math` package). The precision of the floating point number is set to the bit length 
of the integer it is supposed to represent. This was not present before CometBFT v0.38.x and all float values were ignored. 
- As of CometBFT v0.38.x, queries can contain floating point numbers as well.
- Note that comparing to floats can be imprecise with a high number of decimals. 
```

docs/core/subscription.md
```diff
@@ -40,6 +40,25 @@ You can also use tags, given you had included them into DeliverTx
...
+While CometBFT imposes no restrictions on the application with regards to the type of 
+the event output, there are several considerations that need to be taken into account 
+when querying events. 
+
+- Queries convert all event values to `big.Float` , provided by `math/big`. Integers
+are converted into a float with a precision equal to the number of bits needed
+to represent this integer. This is done to avoid precision loss for big integers when they 
+are converted with the default precision (`64`). 
+- When comparing two values, if either one of them is a float, the other one will be represented 
+as a big float. Integeres are again parsed as big floats with a precision equal to the number
```
lasarojc, line 54:
```suggestion
as a big float. Integers are again parsed as big floats with a precision equal to the number 
```

internal/indexer/indexer_utils.go
```diff
@@ -0,0 +1,94 @@
...
+// we convert the int to float with a precision equal to the number of bits
+// needed to represent the integer to avoid rounding issues with floats
+// where 100 would equal to 100.2 because 100.2 is rounded to 100, while 100.7
+// would be rounded to 101.
+func compareInt(op1 *big.Int, op2 interface{}) (int, bool, error) {
+	switch opVal := op2.(type) {
+	case *big.Int:
+		return op1.Cmp(opVal), false, nil
+	case *big.Float:
+		vF, _, err := big.ParseFloat(op1.String(), 10, uint(op1.BitLen()), big.ToNearestEven)
```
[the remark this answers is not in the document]
jmalicevic: It should be `opVal.BitLen()` in both. If the precision is different from that of op1, it will be handled properly in the comparison. Will fix this.

internal/indexer/indexer_utils.go
```diff
@@ -0,0 +1,94 @@
...
+
+	"github.com/cometbft/cometbft/state/indexer"
+)
+
+// If the actual event value is a float, we get the condition and parse it as a float
+// to compare against
+func compareFloat(op1 *big.Float, op2 interface{}) (int, bool, error) {
+	switch opVal := op2.(type) {
+	case *big.Int:
+		vF, _, err := big.ParseFloat(opVal.String(), 10, op1.Prec(), big.ToNearestEven)
```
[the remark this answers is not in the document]
lasarojc: Good point. A +1 should fix it, right?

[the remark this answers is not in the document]
jmalicevic: See my response to Sergio's comment below. This is actually a mistake, it should be `opVal.BitLen` here and we should be fine.

internal/indexer/indexer_utils.go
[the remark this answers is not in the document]
lasarojc: I think the idea is to be as precise as needed to compare with `op1` so if `op1` is float then `Prec` is used and if `op1` is Int then `bitten` is used. But as in `compareFloat` you need to `+1`.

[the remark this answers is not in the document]
jmalicevic: So, if `op1` is a `float` then we will call `compareFloat`, not `compareInt`.  My point here is that we are converting `op1` to a `float` and we should use as many bits as needed to represent `op1` (whose value is `opVal`). And if the precision of `vF` and `op2` differ, the `Cmp` function will compare them properly (as it compares mantissa's bit by bit and the exponents separetely). So we should not worry about how closely precisions of op1 and op2 match.  
And a `+1` would not necessarily solve it because we might have an int whose precision is bigger than the float in big float so we would need to find the max of `op1.BitLen()` and `op2.Prec()`.

state/indexer/query_range.go
jmalicevic: This is only in places where we expect the argument to be the height. As the height is expected to be an `Int64`. This here will return `*big.Float`. then in functions where we expect it to return height, it will get the `int64` value using `.Int64()`.

jmalicevic:
> Thanks Jasmina. Some comments, but overall looking quite good!
> Do we need to add a changelog entry?

@sergio-mena , yes definitely, just uploaded them.

state/indexer/query_range.go
sergio-mena: Makes sense

docs/app-dev/indexing-transactions.md
```diff
@@ -276,3 +276,12 @@ This behavior was fixed with CometBFT 0.34.26+. However, if the data was indexed
 Tendermint Core and not re-indexed, that data will be queried as if all the attributes within a height
 occurred within the same event.
 
+## Event attribute value types
+
+Users can use anything as an event value. However, if the event attribute value is a number, the following needs to be taken into accound:
+
+- Negative numbers will not be properly retrieved when querying the indexer.
+- Event values are converted to big floats( from the `big/math` package). The precision of the floating point number is set to the bit length
```
sergio-mena, line 284:
```suggestion
- Event values are converted to big floats (from the `big/math` package). The precision of the floating point number is set to the bit length so that there is no loss of precision in the conversion.
```

docs/core/subscription.md
```diff
@@ -40,6 +40,25 @@ You can also use tags, given you had included them into DeliverTx
 response, to query transaction results. See [Indexing
 transactions](../app-dev/indexing-transactions.md) for details.
 
+## Query parameter and event type restrictions
+
+While CometBFT imposes no restrictions on the application with regards to the type of 
+the event output, there are several considerations that need to be taken into account 
+when querying events. 
+
+- Queries convert all event values to `big.Float` , provided by `math/big`. Integers
```
sergio-mena, line 49:
```suggestion
- Queries convert all numeric event values to `big.Float` , provided by `math/big`. Integers
```

docs/core/subscription.md
```diff
@@ -40,6 +40,25 @@ You can also use tags, given you had included them into DeliverTx
...
+to represent this integer. This is done to avoid precision loss for big integers when they 
+are converted with the default precision (`64`). 
+- When comparing two values, if either one of them is a float, the other one will be represented 
+as a big float. Integers are again parsed as big floats with a precision equal to the number 
+of bits required to represent them. 
+- As with all floating point comparisons, comparing floats with decimal values can lead to imprecise 
+results. 
+- Queries cannot include negative numbers 
+
+Prior to version v38.x, floats were not supported as query parameters.
```
sergio-mena, line 60:
```suggestion
Prior to version `v0.38.x`, floats were not supported as query parameters. 
```

internal/indexer/indexer_utils.go
```diff
@@ -59,36 +49,77 @@ func compareInt(op1 *big.Int, op2 interface{}) (int, bool, error) {
...
 	switch vVal := v.(type) {
 	case *big.Int:
 		if lowerBound != nil {
 			cmp, isFloat, err := compareInt(vVal, lowerBound)
-			if err == nil && (cmp == -1 || (isFloat && cmp == 0 && !ranges.IncludeLowerBound)) {
+			if err != nil {
+				return false
+			}
+			if cmp == -1 || (isFloat && cmp == 0 && !ranges.IncludeLowerBound) {
 				include = false
```
sergio-mena, line 88: If we hit this line, do we really need to go through lines 91-99? Is it gonna change anything?

libs/pubsub/query/query.go
```diff
@@ -227,6 +227,9 @@ func parseNumber(s string) (*big.Float, error) {
 	intVal := new(big.Int)
 	if _, ok := intVal.SetString(s, 10); !ok {
 		f, _, err := big.ParseFloat(extractNum.FindString(s), 10, 125, big.ToNearestEven)
+		if err != nil {
+			fmt.Println(err)
```
sergio-mena, line 231: rather log?

libs/pubsub/query/syntax/parser.go
```diff
@@ -77,6 +77,9 @@ func (a *Arg) Number() *big.Float {
 	if _, ok := intVal.SetString(a.text, 10); !ok {
 		f, _, err := big.ParseFloat(a.text, 10, 125, big.ToNearestEven)
 		if err != nil {
+
+			fmt.Println(err)
```
sergio-mena, line 81: log instead?

state/indexer/block/kv/kv_test.go
```diff
@@ -82,54 +82,54 @@ func TestBlockIndexer(t *testing.T) {
...
+		// 	results: []int64{},
+		// },
+		// "begin_event.proposer CONTAINS 'FCAA001'": {
+		// 	q:       query.MustCompile(`begin_event.proposer CONTAINS 'FCAA001'`),
+		// 	results: []int64{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11},
+		// },
+		// "end_event.foo CONTAINS '1'": {
+		// 	q:       query.MustCompile("end_event.foo CONTAINS '1'"),
+		// 	results: []int64{1, 10},
+		// },
```
sergio-mena, line 132: Temporary change?

libs/pubsub/query/query.go
```diff
@@ -223,8 +223,15 @@ func compileCondition(cond syntax.Condition) (condition, error) {
 // tests for, but we should probably get rid of that.
 var extractNum = regexp.MustCompile(`^\d+(\.\d+)?`)
```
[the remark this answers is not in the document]
jmalicevic: If you just type in 12.434CXXD it will fail because there is nothing to parse it. If you type in "12.34CXXD" it will parse it as a string but then fail when comparing it to a number because it will try to parse it as a float.

internal/indexer/indexer_utils.go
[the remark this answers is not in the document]
mzabaluev: Shouldn't there be a way in `math/big` to convert integers to floats without going through a string representation? If my reading of the documentation on [`SetInt`](https://pkg.go.dev/math/big#Float.SetInt) is correct, it's enough to create a `Float` variable and set its precision to 0 (or `op1.BitLen()`, but that will be used under the hood), so that `SetInt` can be used to losslessly convert the value of `op1`?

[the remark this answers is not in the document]
jmalicevic: You are absolutely right! Thank you for this, I submitted the updated version. I thought the`setInt` was setting it to the value of an `int64`.

internal/indexer/indexer_utils.go
```diff
@@ -12,13 +12,10 @@ import (
 func compareFloat(op1 *big.Float, op2 interface{}) (int, bool, error) {
 	switch opVal := op2.(type) {
 	case *big.Int:
-		vF, _, err := big.ParseFloat(opVal.String(), 10, uint(opVal.BitLen()), big.ToNearestEven)
-		if err != nil {
-			err = fmt.Errorf("failed to convert %s to float", opVal)
-			return -1, false, err
-		}
+		vF := new(big.Float)
+		vF.SetInt(opVal)
```
mzabaluev, line 16: Good, this is much more efficient now.

libs/pubsub/query/query.go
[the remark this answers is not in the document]
sergio-mena: Not sure I get why "12.434CXXD" and "12.34CXXD" will be treated differently.

Anyway, Mm main concern here is that the regex does not require the string to _end_ with digits... i.e., there is no `$` at the end of the regex

internal/indexer/indexer_utils.go
```diff
@@ -80,44 +76,44 @@ func CheckBounds(ranges indexer.QueryRange, v interface{}) bool {
...
 			}
 			if cmp == 1 || (cmp == 0 && isFloat && !ranges.IncludeUpperBound) {
-				return false
+				return false, err
 			}
 		}
 
 	default:
-		return false
+		return false, fmt.Errorf("invalid argument type in query")
```
sergio-mena, line 116: Please add the arg's contents in the error message to improve experience in potential troubleshooting of this code

internal/indexer/indexer_utils.go
sergio-mena: I just realized the switch is on the variable's type, so my previous comment is unclear... it should read "... the arg's type" (e.g., using %T)

libs/pubsub/query/query.go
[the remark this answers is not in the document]
jmalicevic: I had a typo in my previous comment but I did a bit more digging and found a discrepancy between query parsing in the kvindexer and the pubsub query parser. First of all, "12.434CXXD" and "12.34CXXD"  are exactly the same, I mistakenly did not write an extra 4. In pubsub it will convert this to 12.434.In the kvindexer it will fail on conversion to Float/Int. Same for 12.43CXXD. If you quote either one of those, it will parse them as a string and then your comparison is only on equality. 
I wanted to rewrite it initially as you suggested but, I noticed that the test 
`amount > 7` where `amount=8stake`  fails when we add the `$` but passes without it. I know some of our users indeed input amounts like that (`8stake, 8atom` etc). However, the same query on the kvindexer will fail. The reason is that the pubsub basically [parses out the leftmost ](https://github.com/cometbft/cometbft/blob/2e0c72b94f3d817569289d6d6baa284cbbb3619a/libs/pubsub/query/query.go#L227) match on the string. So  out of `12.434CXXD` convert `12.434` to float. I think we should introduce this behaviour in our kvindexer too, but I would to that in a separate PR because it is another module, and  not related to the change here.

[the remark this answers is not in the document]
sergio-mena: 100% agree. The "8atom" think makes sense and we shouldn't break them just because we don't agree with that. So I agree with making the behavior consistent in the kvindexer and, maybe, replacing the current TODO comment above the regex by a couple of sentences explaining the "8stake, 6atom" case.

sergio-mena (approved):
Thanks for this!
Consider tackling my last two minor comments... but in the end, up to you!

libs/pubsub/query/query.go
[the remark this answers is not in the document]
sergio-mena: I just noticed you're off this week. I'll try to merge this today on your behalf

PR #798, opened 2023-05-05
Backport of PR #771 to 0.34 : big int parsing

### LINKED ISSUE #726: Using pubsub query with big values results in an int64 overflow

jmalicevic:
Closes #726  for 0.34.x

This is a manual backport of #771  because Mergify did not trigger it.

### DIFF
--- a/.changelog/unreleased/bug-fixes/771-kvindexer-parsing-big-ints.md
+++ b/.changelog/unreleased/bug-fixes/771-kvindexer-parsing-big-ints.md
@@ -1,2 +1,2 @@
-- `[state/kvindex]` Querying event attributes that are bigger than int64 is now enabled. We are not supporting reading floats from the db into the indexer  nor parsing them into BigFloats to not introduce breaking changes in minor releases.
+- `[state/kvindex]` Querying event attributes that are bigger than int64 is now enabled. We are not supporting reading floats from the db into the indexer nor parsing them into BigFloats to not introduce breaking changes in minor releases.
   ([\#771](https://github.com/cometbft/cometbft/pull/771))
--- a/.changelog/unreleased/bug-fixes/771-kvindexer-parsing-big-ints.md
+++ b/.changelog/unreleased/bug-fixes/771-kvindexer-parsing-big-ints.md
@@ -1,0 +1,2 @@
+- `[state/kvindex]` Querying event attributes that are bigger than int64 is now enabled. We are not supporting reading floats from the db into the indexer  nor parsing them into BigFloats to not introduce breaking changes in minor releases.
+  ([\#771](https://github.com/cometbft/cometbft/pull/771))
--- a/.changelog/unreleased/bug-fixes/771-kvindexer-parsing-big-ints.md
+++ b/.changelog/unreleased/bug-fixes/771-kvindexer-parsing-big-ints.md
@@ -1,2 +1,2 @@
-- `[state/kvindex]` Querying event attributes that are bigger than int64 is now enabled. We are not supporting reading floats from the db into the indexer nor parsing them into BigFloats to not introduce breaking changes in minor releases.
+- `[state/kvindex]` Querying event attributes that are bigger than int64 is now enabled.
   ([\#771](https://github.com/cometbft/cometbft/pull/771))
--- a/docs/app-dev/indexing-transactions.md
+++ b/docs/app-dev/indexing-transactions.md
@@ -267,8 +267,11 @@ is ignored and the data is retrieved as if `match_events=false`.
 Additionally, if a node that was running Tendermint Core 
 when the data was first indexed, and switched to CometBFT, is queried, it will retrieve this previously indexed
 data as if `match_events=false` (attributes can match the query conditions across different events on the same height).
+
 # Event attribute value types
-Users can use anything as an event value. However, if the even attrbute value is a number, the following restrictions apply:
+
+Users can use anything as an event value. However, if the event attribute value is a number, the following restrictions apply:
+
 - Negative numbers will not be properly retrieved when querying the indexer
 - When querying the events using `tx_search` and `block_search`, the value given as part of the condition cannot be a float.
 - Any event value retrieved from the database will be represented as a `BigInt` (from `math/big`)
--- a/docs/app-dev/indexing-transactions.md
+++ b/docs/app-dev/indexing-transactions.md
@@ -267,3 +267,10 @@ is ignored and the data is retrieved as if `match_events=false`.
 Additionally, if a node that was running Tendermint Core 
 when the data was first indexed, and switched to CometBFT, is queried, it will retrieve this previously indexed
 data as if `match_events=false` (attributes can match the query conditions across different events on the same height).
+# Event attribute value types
+Users can use anything as an event value. However, if the even attrbute value is a number, the following restrictions apply:
+- Negative numbers will not be properly retrieved when querying the indexer
+- When querying the events using `tx_search` and `block_search`, the value given as part of the condition cannot be a float.
+- Any event value retrieved from the database will be represented as a `BigInt` (from `math/big`)
+- Floating point values are not read from the database even with the introduction of `BigInt`. This was intentionally done 
+to keep the same beheaviour as was historically present and not introduce breaking  changes. This will be fixed in the 0.38 series.
--- a/docs/app-dev/indexing-transactions.md
+++ b/docs/app-dev/indexing-transactions.md
@@ -268,6 +268,7 @@ Additionally, if a node that was running Tendermint Core
 when the data was first indexed, and switched to CometBFT, is queried, it will retrieve this previously indexed
 data as if `match_events=false` (attributes can match the query conditions across different events on the same height).
 
+
 # Event attribute value types
 
 Users can use anything as an event value. However, if the event attribute value is a number, the following restrictions apply:
--- a/docs/core/subscription.md
+++ b/docs/core/subscription.md
@@ -40,6 +40,20 @@ You can also use tags, given you had included them into DeliverTx
 response, to query transaction results. See [Indexing
 transactions](./indexing-transactions.md) for details.
 
+
+## Query parameter and event type restrictions
+
+While CometBFT imposes no restrictions on the application with regards to the type of 
+the event output, there are several restrictions when it comes to querying 
+events whose attribute values are numeric. 
+
+- Queries cannot include negative numbers
+- If floating points are compared to integers, they are converted to an integer
+- Floating point to floating point comparison leads to a loss of precision for very big floating point numbers
+(`10000000000000000000.0` is treated the same as `10000000000000000000.6`) 
+- When floating points do get converted to integers, they are not rounded, the decimal part is simply truncated.
+This has been done to preserve the behaviour present before introducing the support for BigInts in the query parameters. 
+
 ## ValidatorSetUpdates
 
 When validator set changes, ValidatorSetUpdates event is published. The
--- a/docs/core/subscription.md
+++ b/docs/core/subscription.md
@@ -50,8 +50,8 @@ events whose attribute values are numeric.
 - Queries cannot include negative numbers
 - If floating points are compared to integers, they are converted to an integer
 - Floating point to floating point comparison leads to a loss of precision for very big floating point numbers
-(`10000000000000000000.0` is treated the same as `10000000000000000000.6`) 
-- When floating points do get converted to integers, they are not rounded, the decimal part is simply truncated.
+(e.g., `10000000000000000000.0` is treated the same as `10000000000000000000.6`) 
+- When floating points do get converted to integers, they are always rounded down.
 This has been done to preserve the behaviour present before introducing the support for BigInts in the query parameters. 
 
 ## ValidatorSetUpdates
--- a/libs/pubsub/query/query.go
+++ b/libs/pubsub/query/query.go
@@ -153,7 +153,6 @@ func (q *Query) Conditions() ([]Condition, error) {
 				conditions = append(conditions, Condition{eventAttr, op, value})
 			} else {
 				valueBig := new(big.Int)
-
 				_, ok := valueBig.SetString(number, 10)
 				if !ok {
 					err := fmt.Errorf(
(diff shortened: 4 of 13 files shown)


state/indexer/block/kv/kv_test.go
```diff
@@ -358,3 +358,132 @@ func TestBlockIndexerMulti(t *testing.T) {
 		})
```
sergio-mena, line 358: There were some test cases removed from the this test in the original patch. Any reason for them to stay here?

state/indexer/block/kv/util.go
```diff
@@ -135,7 +136,7 @@ func parseEventSeqFromEventKey(key []byte) (int64, error) {
 func lookForHeight(conditions []query.Condition) (int64, bool, int) {
 	for i, c := range conditions {
 		if c.CompositeKey == types.BlockHeightKey && c.Op == query.OpEqual {
-			return c.Operand.(int64), true, i
+			return c.Operand.(*big.Int).Int64(), true, i
```
sergio-mena, line 139: This change in not present in `v0.37.x`, do we know why?

state/indexer/block/kv/util.go
```diff
@@ -196,7 +197,7 @@ func dedupMatchEvents(conditions []query.Condition) ([]query.Condition, bool) {
 	for i, c := range conditions {
 		if c.CompositeKey == types.MatchEventKey {
 			// Match events should be added only via RPC as the very first query condition
-			if i == 0 && c.Op == query.OpEqual && c.Operand.(int64) == 1 {
+			if i == 0 && c.Op == query.OpEqual && c.Operand.(*big.Int).Int64() == 1 {
```
sergio-mena, line 200: Same here, this does not appear in the `v0.37.x` version of the patch, do we know the reason?

state/txindex/kv/kv.go
```diff
@@ -349,7 +350,7 @@ func lookForHash(conditions []query.Condition) (hash []byte, ok bool, err error)
 func lookForHeight(conditions []query.Condition) (height int64, heightIdx int) {
 	for i, c := range conditions {
 		if c.CompositeKey == types.TxHeightKey && c.Op == query.OpEqual {
-			return c.Operand.(int64), i
+			return c.Operand.(*big.Int).Int64(), i
```
sergio-mena, line 353: Change no present in `v0.37.x`, reason?

state/txindex/kv/utils.go
```diff
@@ -33,7 +34,7 @@ func dedupMatchEvents(conditions []query.Condition) ([]query.Condition, bool) {
 	for i, c := range conditions {
 		if c.CompositeKey == types.MatchEventKey {
 			// Match events should be added only via RPC as the very first query condition
-			if i == 0 && c.Op == query.OpEqual && c.Operand.(int64) == 1 {
+			if i == 0 && c.Op == query.OpEqual && c.Operand.(*big.Int).Int64() == 1 {
```
sergio-mena, line 37: Not in `v0.37.x`, reason?

sergio-mena (commented):
Only minor comments on `v0.34.x` <--> `v0.37.x` discrepancies. I'm interested in the reasons for those

.changelog/unreleased/bug-fixes/771-kvindexer-parsing-big-ints.md
```diff
@@ -0,0 +1,2 @@
+- `[state/kvindex]` Querying event attributes that are bigger than int64 is now enabled. We are not supporting reading floats from the db into the indexer  nor parsing them into BigFloats to not introduce breaking changes in minor releases.
```
lasarojc, line 1:
```suggestion
- `[state/kvindex]` Querying event attributes that are bigger than int64 is now enabled. We are not supporting reading floats from the db into the indexer nor parsing them into BigFloats to not introduce breaking changes in minor releases.
```

docs/app-dev/indexing-transactions.md
```diff
@@ -267,3 +267,10 @@ is ignored and the data is retrieved as if `match_events=false`.
 Additionally, if a node that was running Tendermint Core 
 when the data was first indexed, and switched to CometBFT, is queried, it will retrieve this previously indexed
 data as if `match_events=false` (attributes can match the query conditions across different events on the same height).
+# Event attribute value types
```
lasarojc, line 270:
```suggestion

# Event attribute value types

```

docs/app-dev/indexing-transactions.md
```diff
@@ -267,3 +267,10 @@ is ignored and the data is retrieved as if `match_events=false`.
...
+Users can use anything as an event value. However, if the even attrbute value is a number, the following restrictions apply:
+- Negative numbers will not be properly retrieved when querying the indexer
```
lasarojc, line 272:
```suggestion
Users can use anything as an event value. However, if the event attribute value is a number, the following restrictions apply:

- Negative numbers will not be properly retrieved when querying the indexer
```

docs/app-dev/indexing-transactions.md
```diff
@@ -267,3 +267,10 @@ is ignored and the data is retrieved as if `match_events=false`.
...
+- When querying the events using `tx_search` and `block_search`, the value given as part of the condition cannot be a float.
+- Any event value retrieved from the database will be represented as a `BigInt` (from `math/big`)
+- Floating point values are not read from the database even with the introduction of `BigInt`. This was intentionally done
```
lasarojc, line 275: note to self: not read or read as int?

docs/core/subscription.md
```diff
@@ -40,6 +40,20 @@ You can also use tags, given you had included them into DeliverTx
...
+## Query parameter and event type restrictions
+
+While CometBFT imposes no restrictions on the application with regards to the type of 
+the event output, there are several restrictions when it comes to querying 
+events whose attribute values are numeric. 
+
+- Queries cannot include negative numbers
+- If floating points are compared to integers, they are converted to an integer
+- Floating point to floating point comparison leads to a loss of precision for very big floating point numbers
+(`10000000000000000000.0` is treated the same as `10000000000000000000.6`)
```
lasarojc, line 53:
```suggestion
(e.g., `10000000000000000000.0` is treated the same as `10000000000000000000.6`) 
```

docs/core/subscription.md
```diff
@@ -40,6 +40,20 @@ You can also use tags, given you had included them into DeliverTx
...
+
+While CometBFT imposes no restrictions on the application with regards to the type of 
+the event output, there are several restrictions when it comes to querying 
+events whose attribute values are numeric. 
+
+- Queries cannot include negative numbers
+- If floating points are compared to integers, they are converted to an integer
+- Floating point to floating point comparison leads to a loss of precision for very big floating point numbers
+(`10000000000000000000.0` is treated the same as `10000000000000000000.6`) 
+- When floating points do get converted to integers, they are not rounded, the decimal part is simply truncated.
```
lasarojc, line 54:
```suggestion
- When floating points do get converted to integers, they are always rounded down.
```

libs/pubsub/query/query.go
```diff
@@ -151,16 +152,18 @@ func (q *Query) Conditions() ([]Condition, error) {
 
 				conditions = append(conditions, Condition{eventAttr, op, value})
 			} else {
-				value, err := strconv.ParseInt(number, 10, 64)
-				if err != nil {
-					err = fmt.Errorf(
-						"got %v while trying to parse %s as int64 (should never happen if the grammar is correct)",
-						err, number,
+				valueBig := new(big.Int)
+
```
lasarojc, line 156:
```suggestion
```

libs/pubsub/query/query.go
```diff
@@ -298,11 +301,13 @@ func (q *Query) Matches(events map[string][]string) (bool, error) {
...
 				}
 			} else {
-				value, err := strconv.ParseInt(number, 10, 64)
-				if err != nil {
-					err = fmt.Errorf(
-						"got %v while trying to parse %s as int64 (should never happen if the grammar is correct)",
-						err, number,
+				value := new(big.Int)
+				_, ok := value.SetString(number, 10)
+
```
lasarojc, line 306:
```suggestion
```

libs/pubsub/query/query.go
```diff
@@ -451,42 +456,59 @@ func matchValue(value string, op Operator, operand reflect.Value) (bool, error)
...
+				// Here we are simulating the behavior that int64(floatValue). This was the default behavior
+				// before introducing BigInts and we do not want to break the logic in minor releases.
+				_, ok := v.SetString(strings.Split(filteredValue, ".")[0], 10)
+				if !ok {
+					return false, fmt.Errorf("failed to convert value %s from float to big int", filteredValue)
+				}
+			} else {
+				// try our best to convert value from tags to big int
+				_, ok := v.SetString(filteredValue, 10)
+
```
lasarojc, line 491: Removing these empty lines for consistency with other lines in the change. I am fine with another standard.

```suggestion
```

libs/pubsub/query/query.go
```diff
@@ -451,42 +456,59 @@ func matchValue(value string, op Operator, operand reflect.Value) (bool, error)
...
-			v = int64(v1)
-		} else {
-			var err error
-			// try our best to convert value from tags to int64
-			v, err = strconv.ParseInt(filteredValue, 10, 64)
-			if err != nil {
-				return false, fmt.Errorf("failed to convert value %v from event attribute to int64: %w", filteredValue, err)
 			}
-		}
+			cmpRes := operandVal.Cmp(v)
```
lasarojc, line 497: If you invert it here

```suggestion
			cmpRes := v.Cmp(operandVal)
```

then ...

libs/pubsub/query/query.go
```diff
@@ -451,42 +456,59 @@ func matchValue(value string, op Operator, operand reflect.Value) (bool, error)
...
+			case OpLessEqual:
+				return cmpRes == 0 || cmpRes == 1, nil
+			case OpGreaterEqual:
+				return cmpRes == 0 || cmpRes == -1, nil
+			case OpLess:
+				return cmpRes == 1, nil
+			case OpGreater:
+				return cmpRes == -1, nil
+			case OpEqual:
+				return cmpRes == 0, nil
```
lasarojc, line 508: then you can make the parameter match the comparison

```suggestion
			case OpLessEqual:
				return cmpRes <= 0, nil
			case OpGreaterEqual:
				return cmpRes >= 0, nil
			case OpLess:
				return cmpRes < 0, nil
			case OpGreater:
				return cmpRes > 0 , nil
			case OpEqual:
				return cmpRes == 0, nil
```

docs/app-dev/indexing-transactions.md
jmalicevic: Not read at all, they are simply ignored because the conversion to `BigInt` fails.

state/indexer/block/kv/kv_test.go
jmalicevic: The reason they are here is because in 0.34 we have two tests for the same query - one with `match.events`, and one without. In 0.37 they were left over and hence deleted. Here we need both.

state/indexer/block/kv/util.go
jmalicevic: Because  the function `lookForHeight` is not used at all, we use only `dedupHeight`. In 0.34.x we still keep the old behaviour hence using this function.

state/indexer/block/kv/util.go
jmalicevic: 0.37.x does not have `match.events`, hence no `dedupMatchEvents` function.

state/txindex/kv/kv.go
jmalicevic: Same as above.

state/txindex/kv/utils.go
jmalicevic: Same as for `dedupMatchEvents` in the block indexer.

libs/pubsub/query/query.go
jmalicevic: Yes, the old logic was a bit twisted. I can test this, I was just not sure whether to mess with this at all :)

lasarojc: no worries. Not a requirement for approval, since it is correct.

sergio-mena: As this is a backport, I'd lean toward not much refactoring

sergio-mena (approved):
LGTM.
Please make sure you address @lasarojc's comments

libs/pubsub/query/query.go
jmalicevic: @lasarojc let me know if its fine with you to leave it and whether the other changes make sense :)

.changelog/unreleased/bug-fixes/771-kvindexer-parsing-big-ints.md
```diff
@@ -0,0 +1,2 @@
+- `[state/kvindex]` Querying event attributes that are bigger than int64 is now enabled. We are not supporting reading floats from the db into the indexer nor parsing them into BigFloats to not introduce breaking changes in minor releases.
+  ([\#771](https://github.com/cometbft/cometbft/pull/771))
```
thanethomson, line 2: > Querying event attributes that are bigger than int64 is now enabled.

This makes sense to me as something to communicate to users.

> We are not supporting reading floats from the db into the indexer nor parsing them into BigFloats to not introduce breaking changes in minor releases.

What does this, however, mean for users?

jmalicevic: Maybe this does not have to be in the changelog, but only in the docs. This is basically the same behaviour as they had before, but they might have been unaware of this. I agree that maybe this should be removed frm the changelog but left in the docs.

libs/pubsub/query/query.go
lasarojc: not needed. resolving.

lasarojc (approved):

jmalicevic:
@thanethomson , I am closing this as the content is in the docs, but feel free to ping me to change the changelog message.

Issue #28, opened 2022-12-23
State sync from local snapshot: Understand problem, explore solutions


sergio-mena:
Tasks:

- [x] Analyze and understand the problem ADR083 is trying to address
- [x] Gauge users' interest in solving this problem (how painful it is _not_ to solve it?)
- [x] Explore all possible solutions, among them ADR083 (tendermint/tendermint#9651)
- [x] Adapt ADR083 according to the solution decided, finalize and merge it

DoD:
- We have made sure that users want the problem solved
- We have decided on a solution, and implemented it
- The ADR is merged

Original issue: tendermint/tendermint#9947

JayT106:
Many online projects need this feature leading ASAP (including us, the Cronos chain). Before the p2p protocol can run efficiently. I think this will be a short-term solution for solving the unstable statesync problem.

cason:
> Keep ADR 103 as it is, with the original solution. Cite ADR X as an alternative. Deprecate ADR 103 in favor of ADR X.

What I mean here is to keep ADR 103 proposing a solution that is not a local sync, as it involves no snapshot. Make it clear on its wording. Point out some disadvantages of this solution.

sergio-mena:
The way I see it (with the context I have):

* Set ADR 103 as rejected
* Write ADR X as proposed
* Add references to each other in a visible place

cason:
So, we have to consider here another option that was implemented (is already merged) by the SDK. It consists of two parts:

1. https://github.com/cosmos/cosmos-sdk/pull/16067

This PR creates a snapshot manager binary `simd` to manage snapshots of a SDK-based application. From my understanding, the SDK has already a module to manage application snapshots, so what this program does is to offer a CLI interface to interact with the snapshots manager.

In the context of this issue, 3 commands are relevant:

- `dump`: exports a snapshot produced by the application to a tarball (a tar.gz file). The snapshot is identified by the height it was taken.
- `load`: loads a snapshot from a tarball, adding it to the snapshots store. Together with `dump`, this enables out-of-band snapshot transfer.
- `restore`: apply a snapshot present in the store to the application. Since the application is stored in a leveldb-style DB, this is possible even with a running application, it is enough to lock the store.

These three commands, plus the snapshot manager CLI, enables transferring the application state between two nodes in a pretty safe way. The problem at this point is that the restored snapshot might produce an application state that is not compatible with the CometBFT state (putting it simple, the heights don't match).

[...]

PR #729, opened 2023-04-19
ADR 103 Local State Sync Support (ported from Tendermint Core repo)

adizere:
Closes #28 

[Rendered view of adr-103-local-state-sync.md](https://github.com/cometbft/cometbft/blob/adi/28_local_sync_adr/docs/architecture/adr-103-local-state-sync.md)

TODOs

- [ ] Cross-validate with the solution in #728 
- [ ] Update docs to reflect the new workflow, with instructions on how users can do local (out-of-band) state sync
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments

PR #801, opened 2023-05-05
ADR-104: Out-of-band state sync support

cason:
Closes https://github.com/cometbft/cometbft/issues/28.

This should be a replacement for https://github.com/cometbft/cometbft/pull/729.

[Rendered ADR](https://github.com/cometbft/cometbft/blob/cason/adr-104/docs/architecture/adr-104-out-of-band-state-sync.md)


yihuang (changes requested):
Can you describe more about the actual design, especially the UX part, what does user need to do to do certain task. Since we have multiple solutions here for the same problem, the UX would the decisive factor I think.

cason:
> Can you describe more about the actual design, especially the UX part

This is a work in progress. The above content should go into "Detailed Design" section.

In general terms, the operation should be the one described on the "Decision" section.

cason:
> Since we have multiple solutions here for the same problem, the UX would the decisive factor I think.

I agree we can have multiple designs for the user interfaces, but I think that this ADR (with its current content) reasonably covers the possible solutions.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -0,0 +1,215 @@
...
+obtained out-of-band by node operators.
+
+In other words, the following operation should be, in general terms, possible:
+
+1. When configuring a new node, operators should be able to define where (e.g.,
+   a local file system location) the client side of State sync should look for
+   application snapshots to install or restore;
+1. Operators should be able to instruct the server side of State sync on a
+   running node to export application snapshots, when they are available,
+   to a given location (e.g., at the local file system);
```
yihuang, line 141: Is it like a standalone command that'll connect to the p2p network and download the snapshots?

cason: No p2p, no network, no download. You instruct the node (e.g., via CLI command) to save a snapshot to disk.

cason: Of course, you can save it to a NFS shared disk, so you have network and a form of upload.

yihuang: do you mean export a snapshot from the state running in the local node, not discover and download a snapshot from the network?

cason: Discovering and fetching snapshots is what the existing State sync does.

The proposed extension assumes that the node operator will discover and fetch the snapshot in some way and just point to the bootstrapping node (see item 1.) where to find the snapshot data.

yihuang: yeah, but I'd like a standalone that simply download the snapshot from p2p network and save to disk, nothing more, but I guess this could be done on app side.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -0,0 +1,215 @@
...
+
+> 2. What is the problem?
+
+With the widespread adoption of State sync to bootstrap nodes, however,
+what should be one of its strengths - the ability to discover and fetch
+application snapshots from peers in the p2p network - has turned out to be one
+of its weakness.
+In fact, while downloading recent snapshots is very convenient for new nodes
+(clients of the protocol), providing snapshots to multiple peers (as servers of
+the protocol) is _bandwidth-consuming_.
```
tac0turtle, line 29: i would say this is less of a bandwidth issue, but more of a reason of space and system requirements.

docs/architecture/adr-104-out-of-band-state-sync.md
tac0turtle: i think all this can be done app side, its simpler if its done app side since there may be gotchas in the app that are hard to account for

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -0,0 +1,215 @@
...
+in order to adopt this solution.
+
+## Decision
+
+> This section records the decision that was made.
+> It is best to record as much info as possible from the discussion that happened.
+> This aids in not having to go back to the Pull Request to get the needed information.
+
+State sync should support the bootstrap of new nodes from application snapshots
+obtained out-of-band by node operators.
```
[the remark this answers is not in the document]
cason: > State sync should support the bootstrap of new nodes from application snapshots

I would add "local" to this sentence? Otherwise, it is what State sync already does...

docs/architecture/adr-104-out-of-band-state-sync.md
cason: > i think all this can be done app side

The point here is to create a way to a operator to instruct CometBFT to "export" a snapshot. The snapshot will still be produced by the application, using the same ABCI calls currently in use. The difference is that the snapshot retrieved from the application will not be sent through the network but instead saved to the file system.

tac0turtle: this is already how state sync works for the sdk. Def adding support for some of the items here makes sense, but we would only need a way to bootstrap the node, like huang yi did. 



> Makes sense to rely on the app-side for fulfilling this requirement! Is there already app-side support for it?

we are merging the prs to make it work. they are all open and approved

cason: Links, please?

tac0turtle: links to?

cason: > we are merging the prs to make it work. they are all open and approved

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -0,0 +1,215 @@
...
+The existing protocol is essentially altruistic: it assumes that established
+nodes will support the bootstrap of fresh nodes without receiving anything in
+return, other than having new peers joining their network.
+
+From an operational point of view, the design of the p2p layer should take into
+consideration the fact that not every node in the network is a "good" state
+sync server. Fresh nodes, which need support for bootstrapping, should then be
+able to discover peers in the network that are able or willing to provide
+application snapshots.
+The implementation of this feature would require a more complex content-based
```
jmalicevic, line 78: While I agree in principle with this statement, I am wondering, do we know whether in practice, at this moment, the set of nodes contacted via RPC for statesync is overlapping with the peers the node connects in any case? 

My reasoning is as follows:
1. If it overlaps, then we contact those peers anyways, and we could just have a peer signal whether it is also a statesync peer.. and then a node keeps track of those peers. 
2. If it does not, or it overlaps partially,we still need to pre-configure the set of peers to use for statesync, or at least have a special statesync seed node maybe that would give us useful peers for statesync. This might be a better/easier solution or complimentary to 1. because we still need to have enough statesyncing peers to be able to light client verify the headers.

docs/architecture/adr-104-out-of-band-state-sync.md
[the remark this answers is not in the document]
jmalicevic: Is out of band the same as local though? Local to me would mean the node had a snapshot it itself stored before. Out of band can mean it was just copied or obtanined in another manner.

docs/architecture/adr-104-out-of-band-state-sync.md
yihuang: > > we are merging the prs to make it work. they are all open and approved
> 
> 

https://github.com/cosmos/cosmos-sdk/pull/16067

https://github.com/cosmos/cosmos-sdk/pull/16061

docs/architecture/adr-104-out-of-band-state-sync.md
cason: This is a good point, but I have not idea. In principle, nodes that we connect via RPC to implement the light client should have no relation with the peers a node has in the p2p layer. I didn't fully follow you reasoning from this point, though.

docs/architecture/adr-104-out-of-band-state-sync.md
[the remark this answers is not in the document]
cason: I would say snapshots are local, for the sake of this definition. In general, I would say, from point 2. below, that they are local because obtained out-of-band?

docs/architecture/adr-104-out-of-band-state-sync.md
jmalicevic: Thanks! I know that in principle they do not have to be related, was just wondering whether in practice they are. 

My other point was more with regards to the discussion on strengthening the p2p based statesync. We would still need to have a user manually configure state syncing peers or have a new special node that serves as a seed for statesyncing peers. The node discovers peers who provide snapshots and gossips those around.

docs/architecture/adr-104-out-of-band-state-sync.md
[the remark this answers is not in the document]
jmalicevic: So local in this terminology is that the snapshots are applied from something eventually present on the node. Even though it originated from somewhere else. Sorry, this is just nit picking, nothing major :D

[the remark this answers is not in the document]
cason: No, that's important. Recall that we should still pick a name for the PR :)

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -0,0 +1,344 @@
...
+what should be one of its strengths - the ability to discover and fetch
+application snapshots from peers in the p2p network - has turned out to be one
+of its weakness.
+In fact, while downloading recent snapshots is very convenient for new nodes
+(clients of the protocol), providing snapshots to multiple peers (as servers of
+the protocol) is _bandwidth-consuming_, especially without a clear incentive for
+node operators to provide this service. 
+As a result, the number of nodes in production CometBFT networks offering the
+state sync service (i.e., servers offering snapshots) has been limited, which
+has rendered the service _fragile_ (from the client's point of view). It is very
```
cason, line 34:
```suggestion
has rendered the service _fragile_ (from the client's point of view). In other words, it is very
```

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -51,17 +54,29 @@ state sync, discovering and fetching application snapshots from peers in the
 network, in order to address the above mentioned limitations, while preserving
 most of state sync's operation.
 
+The ADR presents two solutions:
+
+> 1. The first solution is implemented by the application and was proposed and
```
cason, line 59: I would remove this `>` and render it a normal text block

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -51,17 +54,29 @@ state sync, discovering and fetching application snapshots from peers in the
 network, in order to address the above mentioned limitations, while preserving
 most of state sync's operation.
 
+The ADR presents two solutions:
+
+> 1. The first solution is implemented by the application and was proposed and 
+implemented by the SDK (PR #16067 and #16061). This ADR describes the solution
```
cason, line 60: I would add links to these PRs.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -51,17 +54,29 @@ state sync, discovering and fetching application snapshots from peers in the
 network, in order to address the above mentioned limitations, while preserving
 most of state sync's operation.
 
+The ADR presents two solutions:
+
+> 1. The first solution is implemented by the application and was proposed and 
+implemented by the SDK (PR #16067 and #16061). This ADR describes the solution 
+in a general way, proividing guidlines to non-SDK based applications if they wish
+to implement their own local state sync.
```
cason, line 62: Local or out-of-band? :-)

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -51,17 +54,29 @@ state sync, discovering and fetching application snapshots from peers in the
...
+implemented by the SDK (PR #16067 and #16061). This ADR describes the solution 
+in a general way, proividing guidlines to non-SDK based applications if they wish
+to implement their own local state sync. 
+> 2. The second part of the ADR proposes a more general solution, that uses ABCI
+to achieve the same behaviour achieved by the SDK.
+
+
 ## Alternative Approaches
 
-> This section contains information around alternative options that are considered
```
cason: This text is from the ADR template, we can remove in the final version.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -124,58 +138,173 @@ in order to adopt this solution.
 
 ## Decision
```
cason, line 139: I would maintain the previous content of this section, which provides a high-level overview of the proposed solution, to then include the observations you added.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -124,58 +138,173 @@ in order to adopt this solution.
...
+### Application-implemented local state sync
 
+This section describes the solution to local state sync implemented by the SDK. An application can 
+chose to implement local state sync differently, or implement only a subset of the functionalities
+implemented by the SDK.
 
-## Detailed Design
+This solution exposes a command line interface enabling a node to manipulate the snapshots
+including dumping existing snapshots to an exportable format, loading and deleting exported snapshots,
+as well as bootstraping a node's state based on a local snapshot.
```
cason, line 158: Node's state or application state?

To be precise here, `restore` installs a snapshot into the application. Then the bootstrap command configures the node's state and block store accordingly.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -124,58 +138,173 @@ in order to adopt this solution.
...
+including dumping existing snapshots to an exportable format, loading and deleting exported snapshots,
+as well as bootstraping a node's state based on a local snapshot. 
+
+The SDK exposes the following commands for snapshot manipulation:
+
+```script
+
+delete      Delete a local snapshot
+dump        Dump the snapshot as portable archive format
+export      Export app state to snapshot store
```
cason, line 166: I don't know whether you copied the description from the SDK PR. But here we are exporting a snapshot (not the application state, which can be in a newer state) from the application to the snapshot store.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -124,58 +138,173 @@ in order to adopt this solution.
...
-## Detailed Design
+This solution exposes a command line interface enabling a node to manipulate the snapshots
+including dumping existing snapshots to an exportable format, loading and deleting exported snapshots,
+as well as bootstraping a node's state based on a local snapshot. 
+
+The SDK exposes the following commands for snapshot manipulation:
+
+```script
+
+delete      Delete a local snapshot
```
cason, line 164: Delete a snapshot from the snapshot store.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -124,58 +138,173 @@ in order to adopt this solution.
...
+dump        Dump the snapshot as portable archive format
+export      Export app state to snapshot store
+list        List local snapshots
+load        Load a snapshot archive file into snapshot store
+restore     Restore app state from local snapshot
```
cason, line 169: So, "local" here means "in the snapshot store"?

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -124,58 +138,173 @@ in order to adopt this solution.
...
+
+```
+
+and the following command to bootstrap the state of CometBFT upon a completed state sync:
```
cason, line 173: Not a completed state sync, but after restoring/installing a snapshot.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -124,58 +138,173 @@ in order to adopt this solution.
...
+
+``` script
+
+comet bootstrap-state          Bootstrap the state of CometBFT's state and block store
```
cason, line 177: Bootstrap.... from the application state/height.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -124,58 +138,173 @@ in order to adopt this solution.
...
+
+```
+
+These commands enable the implementation of both the client and server side of statesync. 
+Namely, a statesync server can use `dump` to create a portable archieve format out existing snapshots,
```
cason, line 182: I believe that the operation is inverted here. The operator first `export` a snapshot to the store, to them `dump` the snapshot from the store to a file.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -124,58 +138,173 @@ in order to adopt this solution.
...
+or trigger snapshot creation using `export`. 
+
+The client side, restores the application state from a local snapshot that was previously exported, using
```
cason, line 185: The client should first `load` a snapshot from a file to then `restore` it. The latter restore/install a snapshot from the snapshot store, while the first adds a snapshot to the store.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -124,58 +138,173 @@ in order to adopt this solution.
...
+for snapshot manipulation. 
+The interface was derived from the SDK's implementation of local state sync.  
+
+```golang
+
+// Delete a snapshot for a certain height
+func (s *Store) Delete(height uint64, format uint32) error
+
+// Retrieves a snapshot for a certain height
+func (s *Store) Get(height uint64, format uint32) (*types.Snapshot, error)
```
cason, line 211: `types` here is ABCI types?

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -124,58 +138,173 @@ in order to adopt this solution.
...
+func (s *Store) PathChunk(height uint64, format, chunk uint32) string
+
+```
+
+In order to dump a snapshot, an application needs to retrieve all the chunks stored at a certain path. 
+
+**CometBFT state bootstrap**
+
+In addition to managing snapshots, it is neccessary to bootstrap (setup) the state and block store of CometBFT
+in order for CometBFT to continue with block sync and consensus after state sync.
```
cason, line 237: Actually, so that CometBFT can start properly. Notice that the `boostrap` command is an off-line command.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -124,58 +138,173 @@ in order to adopt this solution.
...
+
+**CometBFT state bootstrap**
+
+In addition to managing snapshots, it is neccessary to bootstrap (setup) the state and block store of CometBFT
+in order for CometBFT to continue with block sync and consensus after state sync. 
+At the moment of writing this ADR, there is no command line in CometBFT that supports this.
+Thus application developers have to, within their bootstraping command:
+
+- Create a state and block store
+- Launch a light client to verify the snapshot headers
```
cason, line 242: Not the headers. The light client is launched to learn the block header for the height of the snapshot.

cason (commented):
Good additions. 

I think the second part of the detailed design, from line 234, is a little speculative and dives into too much detail. My opinion is that this section should essentially tell the reader how the SDK commands can be implemented using ABCI calls (step by step), leaving out details on how and where we store stuff, for instance.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -0,0 +1,344 @@
...
+_orthogonal_ to the solution proposed in this ADR, which should have relevant use
+cases even for an enhanced version of state sync.
+
+### Manual application state transfer
+
+The proposed [ADR 083][adr083]/[ADR 103][adr103] inspired the discussions that
+led to this ADR. At first glance, their proposal might look identical to the
+solution discussed here. The main distinction is that their proposal does not
+involve installing application snapshots, but relies on node operators to
+manually synchronize the application state.
```
adizere, line 106: Nit: adding emphasis here.

```suggestion
solution discussed here. The main distinction is that these older proposals do not
involve installing application _snapshots_, but rely on node operators to
manually synchronize the application _state_.
```

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -124,58 +138,173 @@ in order to adopt this solution.
...
+CometBFT, we propose a solution using ABCI, that mimics the behaviour described in the previous section.
+
+Local state sync can be performed online, while CometBFT is running, by reading the snapshots from a 
+predefined location. For this to be possible, this ADR proposes the following additions to the configuration file:
+
+```bash
+[statesync]
+# State syncing from a local snapshot
+local_sync=false 
+# Path to snapshot, will be ignored if local_sync=true
```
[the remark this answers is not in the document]
adizere: Agree

docs/architecture/README.md
```diff
@@ -42,6 +42,8 @@ numbering our ADRs from 100 onwards.
 
 ### Proposed
 
+- [ADR-104: Out-of-band state sync support](./adr-104-out-of-band-state-sync.md)
```
adizere, line 45: Update title here also if we decide to rename.

adizere (approved):
The new ADR content helped answer many of my questions. Thanks Jasmina and Daniel!

I think we should fix any inaccuracies in the ADR and merge this.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -0,0 +1,344 @@
...
+- Launch a light client to verify the snapshot headers
+- Use the light client's `State` and `Commit` functions to 
+retrieve the proper state, verify the `AppHash` and retrieve the `LastCommit`.
+- Save the retrieved values into the state and block stores. 
+
+This code is essentially what the existing implementation of the statesync reactor does, once state syncing is complete 
+minus the functions that apply the snapshot chunks (as this has already been done offline). 
+
+
+## CometBFT based local state sync
```
adizere, line 251: This should be H3.

```suggestion
### CometBFT based local state sync
```

docs/architecture/adr-104-out-of-band-state-sync.md
jmalicevic: I did copy it and export does that basically they export the application state to the store. It is essentially instructing the application to take a snapshot and put it in the snapshot store.

docs/architecture/adr-104-out-of-band-state-sync.md
jmalicevic: Yes, I will add it to the description.

docs/architecture/adr-104-out-of-band-state-sync.md
jmalicevic: They do not have to be related. HEnce the 'or'. Snapshots are created anyways, and `dump` should get us those already created. `export` can just force the creation of a new one but `dump` is not dependent on executing it.

docs/architecture/adr-104-out-of-band-state-sync.md
jmalicevic: I think here it is SDK specific, I had it on my todo to verify the types and make sure they are Comet consistent.

docs/architecture/adr-104-out-of-band-state-sync.md
jmalicevic: Yes, that s what I had in mind but will clarify it better.

docs/architecture/adr-104-out-of-band-state-sync.md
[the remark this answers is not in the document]
jmalicevic: :) yes

jmalicevic:
> Good additions.
> 
> I think the second part of the detailed design, from line 234, is a little speculative and dives into too much detail. My opinion is that this section should essentially tell the reader how the SDK commands can be implemented using ABCI calls (step by step), leaving out details on how and where we store stuff, for instance.

Line 234 talks about the need to bootstrapCometBFT's state after an approach following the SDKs implementation. So I think we do need to have it and application developers should be aware of that. Hopefully we will replace this copy/past approach with the solution solving issue #803 . 

I will see and rephrase the rest but my point was that we do not want to provide the same commands as the SDK as we will be creating the snapshots online while we run normal operations. We could expand the CLI with the equivalent of `export` for example, which could be useful. What I wanted to propose is a solution for how will we know where the snapshot dump is stored and how it is retrieved.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -124,58 +138,173 @@ in order to adopt this solution.
...
+
+In addition to managing snapshots, it is neccessary to bootstrap (setup) the state and block store of CometBFT
+in order for CometBFT to continue with block sync and consensus after state sync. 
+At the moment of writing this ADR, there is no command line in CometBFT that supports this.
+Thus application developers have to, within their bootstraping command:
+
+- Create a state and block store
+- Launch a light client to verify the snapshot headers
+- Use the light client's `State` and `Commit` functions to 
+retrieve the proper state, verify the `AppHash` and retrieve the `LastCommit`.
```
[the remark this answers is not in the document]
jmalicevic: You are right the type is not LastCommit, but that is the last commit for the given height. Thanks for catching this.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -0,0 +1,463 @@
...
+is that it relies on the ability of node
+operators to properly synchronize the application state between two nodes.
+While experienced node operators are likely able to perform this operation in a
+proper way, we have to consider a broader set of users and emphasize that it is
+an operation susceptible to errors.
+Furthermore, it is an operation that is, by definition, application-specific:
+applications are free to manage their state as they see fit, and this includes
+how and where the state is persisted.
+Node operators would therefore need to know the specifics of each application
+in order to adopt this solution.
```
sergio-mena, line 137: Somewhere in this paragraph, I would also mention the risk of data corruption (e.g. in a manually transferred DB) going undetected with the manual process in ADR103

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -0,0 +1,463 @@
...
+
+``` script
+
+comet bootstrap-state          Bootstrap the state of CometBFT's state and block store from the
+                               application's state
+
+```
+
+These commands enable the implementation of both the client and server side of statesync. 
+Namely, a statesync server can use `dump` to create a portable archieve format out existing snapshots,
```
sergio-mena, line 201:
```suggestion
Namely, a statesync server can use `dump` to create a portable archive format out existing snapshots,
```

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -0,0 +1,463 @@
...
+application developer has to manually bootstrap CometBFTs state and block stores. 
+With support for local State sync, a node can simply load a snapshot from a predefined location and offer it to the application
+as is currently done via networked state sync. 
+
+On the server side, without any support for local State sync, an operator has to manually instruct the application
+to export the snapshots into a portable format (via `dump`). 
+
+Having support for this within CometBFT, the app can automatically perform this export when taking snapshots. 
+
+In order to supported local State sync, the following changes to CometBFT are necessary:
```
sergio-mena, line 291:
```suggestion
In order to support local State sync, the following changes to CometBFT are necessary:
```

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -0,0 +1,463 @@
...
+
+1. Adding new configuration options to the config file.
+2. Introduce a CLI command that can explicitly tell the application to create a snapshot export, in case 
+operators decide not to generate periodical exports.
+3. Extract a snapshot from the exported format.
+4. Alter existing ABCI calls to signal to the application that we want to create a snapshot export periodically.
```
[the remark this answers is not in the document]
sergio-mena: (I realize this is explained later on in the text)

sergio-mena (approved):
Thanks!

docs/architecture/adr-104-out-of-band-state-sync.md
jmalicevic: I think the actual risk is that the oeprator has to read the applications DB, then dump the contenst, without knowing the exact format of the application's snapshot DB. But def something to add.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -233,34 +259,67 @@ In order to dump a snapshot, an application needs to retrieve all the chunks sto
...
+- Launch a light client to obtain and verify the block header for a given height.
+- Use the light client's `State` function to 
+verify that the `AppHash` for that block matches the `AppHash` obtained via
+the call to `Info` (which is done on startup). 
+- Use `Commit` to retrieve the last commit for the snapshot height. 
 - Save the retrieved values into the state and block stores. 
 
-This code is essentially what the existing implementation of the statesync reactor does, once state syncing is complete 
-minus the functions that apply the snapshot chunks (as this has already been done offline). 
+This code is essentially what CometBFT does as part of node setup, once state syncing is complete.
```
cason, line 276:
```suggestion
This code essentially mimics what CometBFT does as part of node setup, once state sync is complete.
```

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -271,68 +330,125 @@ snapshot_dump_path=""
...
+An improvement to the second option mentioned above would be to add path parameters to the command,
+ and include the path into `RequestListSnapshots` instructing the application to generate a snapshot export
+at a given location.
+
+A third option is the introduction of a new ABCI call: `ExportSnapshots`, which will have the same effect as option 2 above.
+
+
+#### *Automatic snapshot exporting*
+
+Applications generate snapshots in fixed time intervals. The application itself measures the time passed since the last snapshot,
```
cason, line 362:
```suggestion
Applications generate snapshots with a regular cadence (fixed time intervals or heights). The application itself measures the time or number of heights passed since the last snapshot,
```

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -271,68 +330,125 @@ snapshot_dump_path=""
...
+snapshots into a compressed file, we need to make sure that we do not return files that with ongoing writes.
 An alternative is for CometBFT to provide a Snapshot export store with snapshot isolation. This store would essentially be pruned as soon as a new snapshot is written,
 thus not taking too much more space.
 
+If it is the application that exports the snapshots, it is something the application developer has to be aware of.
+
+#### Syncing a node using local snapshots
+
+On startup, if `local_sync` is set to `true`, CometBFT will try to look for an existing snapshot at the path
+given by the operator. The snapshots will be loaded and state restored as if they came from a peer in the current implementation.
```
cason, line 386:
```suggestion
given by the operator. If a snapshot is available, it will be loaded and state restored as if it came from a peer in the current implementation. 
```

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -271,68 +330,125 @@ snapshot_dump_path=""
...
+    | exportedSnapshot| [] byte | Snapshot export created by the application. | 1            |
+
+    Commit signals the application to persist application state. It takes no parameters.
+
+* **Response**:
+
+    | Name               | Type                                                    | Description                            | Field Number |
+    |--------------------|-------------------------------------------------------- |----------------------------------------|--------------|
+    | result             | [Result](../../spec/abci/abci%2B%2B_methods.md#result)  | The result of the snapshot offer.      | 1            |
+    | resultChunk        | ResultC                                                 | The result of applying the chunks.     | 2            |
```
cason, line 412: These should be multiple?

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -271,68 +330,125 @@ snapshot_dump_path=""
...
+If CometBFT is not the one who created the exported file, we introduce a new ABCI call `UnpackSnapshot` 
+to send the exported snapshot as a blob of bytes to the application, which uncompresses it, installs it 
+and responds whether the snapshot is accepted and the chunk application has passed. 
+
+
+* **Request**:
+
+    | Name            | Type    | Description                                 | Field Number |
+    |-----------------|---------|---------------------------------------------|--------------|
+    | exportedSnapshot| [] byte | Snapshot export created by the application. | 1            |
```
cason, line 403: So is Comet that saves the snapshot to a file?

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -271,68 +330,125 @@ snapshot_dump_path=""
...
+  enum ResultC {
+    UNKNOWN         = 0;  // Unknown result, abort all snapshot restoration
+    ACCEPT          = 1;  // The chunks were accepted.
+    ABORT           = 2;  // Abort snapshot restoration, and don't try any other snapshots.
+    RETRY_SNAPSHOT  = 3;  // Restart this snapshot from `OfferSnapshot`, reusing chunks unless instructed otherwise.
+    REJECT_SNAPSHOT = 4;  // Reject this snapshot, try a different one.
+  }
+```
+
+Unlike networked statesync, we do not refetch individual chunks, thus if the chunk application of the exported snapshot fails, snapshot application fails.
```
cason, line 424:
```suggestion
Unlike networked state sync, we do not re-fetch individual chunks, thus if the application of a chunk fails, then the application of the whole snapshot fails. 
```

cason (commented):
Nice writing up, I think this document is covering almost every possible solution for the discussed problem, while providing an overview of the specifics that would have to be taken into consideration by any implementation.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -154,40 +176,44 @@ chose to implement local state sync differently, or implement only a subset of t
...
-`restore`. Upon successful completion of the previous sequence of commands, the state of CometBFT is bootstrapped
+`restore`. Before `restore` can be called, the client has to `load` an archived snapshot into its
+local snapshot store.
+Upon successful completion of the previous sequence of commands, the state of CometBFT is bootstrapped
 using `bootstrap-state` and CometBFT can be launched. 
 
 
 There are three prerequisites for this solution to work when a node is syncing:
+
 1. The application has access to the snapshot database
```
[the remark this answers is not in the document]
jmalicevic: Yes, it is the same, I can clarify that.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -271,68 +330,125 @@ snapshot_dump_path=""
...
+
+
+#### *Automatic snapshot exporting*
+
+Applications generate snapshots in fixed time intervals. The application itself measures the time passed since the last snapshot,
+and CometBFT has no role in instructing the application when to take snapshots. 
+
+The State sync reactor currently retrieves a list of snapshots from a peer, who obtains these snapshots via ABCI (`RequestListSnapshots`). 
+
+Applications can thus themselves be in charge of dumping the snapshots into a given file format, the same way they generate snapshots.
```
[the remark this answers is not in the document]
jmalicevic: Yes exactly, the app does it in the background.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -271,68 +330,125 @@ snapshot_dump_path=""
...
+ If `auto_snapshot_dump` is true, 
+CometBFT instructs the application to export the snapshots periodically.
+
+An alternative solution is that CometBFT, itself, using the implementation of the `dump` command, whichever 
+is chosen at the time, creates or asks the application to create snapshot exports.
```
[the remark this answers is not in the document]
jmalicevic: To export existing ones. Although the application can do things differently if we ask the app to export it. Otherwise, if comet is doing it, we are just asking the app for current snapshots.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -271,68 +330,125 @@ snapshot_dump_path=""
...
+CometBFT instructs the application to export the snapshots periodically.
+
+An alternative solution is that CometBFT, itself, using the implementation of the `dump` command, whichever 
+is chosen at the time, creates or asks the application to create snapshot exports. 
+
+**Export file consistency**
 
 The same caveat of making sure the data exported is not corrupted by concurrent reads and writes applies here as well. If we decide to simply export
-snapshots into a compressed file, we need to make sure that we do not return files that are not already written. 
+snapshots into a compressed file, we need to make sure that we do not return files that with ongoing writes.
```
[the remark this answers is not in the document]
jmalicevic: Yes exactly, the point is whether the exported file is complete or not. The  snapshots themselves are not of concern here.

docs/architecture/adr-104-out-of-band-state-sync.md
jmalicevic: In this case no.. This call is reserved for reading an exported snapshot that comet is just reading in as bytes and sending to the app for parsing.

docs/architecture/adr-104-out-of-band-state-sync.md
jmalicevic: Actually no ,I think that if one chunk is invalid, we can't apply any of those so CometBFT should get a new snapshot export. Unless we want to expand the implementation to refetch individual chunks.

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -0,0 +1,465 @@
...
+- Additional complexity, with the possible addition of a CLI command to save
+  application snapshots to the file system
+
+## References
+
+- [State sync][state-sync] description, as part of ABCI++ spec
+- Original issue on Tendermint Core repository: [statesync: bootstrap node with state obtained out-of-band #4642][original-issue]
+- Original solution on Tendermint Core repository: [ADR 083: Supporting out of band state sync #9651][adr083]
+- Original proposal ported to CometBFT repository: [ADR 103: Local State Sync Support][adr103]
+
```
cason, line 458:
```suggestion
- SDK's implementation of local state sync: [local snapshot store management][sdk-pr2] and [CometBFT state bootstrap][sdk-pr1]
```

docs/architecture/adr-104-out-of-band-state-sync.md
```diff
@@ -0,0 +1,465 @@
+# ADR 104: State sync from local application snapshot
+
+## Changelog
+
+- 2023-05-05: Initial Draft (@cason)
+- 2023-24-05: Added description of SDK-based & a CometBFT based solution (@jmalicevic)
+
+## Status
+
+Accepted | Rejected | Deprecated | Superseded by
```
cason, line 10:
```suggestion
Accepted
```

PR #802, opened 2023-05-06
make proto-gen

faddat:
## Description
This PR just runs make proto-gen

I don't really know how to use unclog and never have but will give it a 
5th try

---

#### PR checklist

- [x] Tests written/updated
- [x] Changelog entry added in `.changelog` (we use
  [unclog](https://github.com/informalsystems/unclog) to manage our changelog)


faddat:
I just need to mention that if we're going to push unclog, then we need to include the commands for the ad file. Please please please cause that would be sweet.

faddat:
oh it just gobencher failing.  gobencher always fails. 🤷

thanethomson:
This fails our protobuf code generation check and doesn't build: https://github.com/cometbft/cometbft/actions/runs/4900600023/jobs/8752262026

How did you generate these changes? Are you perhaps using the cosmos/gogoproto gogofaster plugin locally? If so, please note that the migration to cosmos/gogoproto only happened for v0.38.x and v0.37.x (see https://github.com/tendermint/tendermint/pull/9356 and https://github.com/tendermint/tendermint/pull/9381).

> I just need to mention that if we're going to push unclog, then we need to include the commands for the ad file. Please please please cause that would be sweet.

Have you read the docs? https://github.com/informalsystems/unclog#adding-a-new-unreleased-entry

thanethomson:
@faddat what's the context for this PR?

PR #772, opened 2023-04-27
ADR-103: Protobuf definition versioning

thanethomson:
Relates to #95.

Since we have had some relatively nuanced discussions around the Protobuf versioning, I thought it would be valuable to capture a summary of them in an ADR.

I've tried to keep this ADR as simple and focused as possible on the versioning aspect of the Protobuf changes, without mention of the package name change from `tendermint` -> `cometbft`, the generation of Go code into its own new module, the idea of switching from gogoproto to Google's Go code generator (this perhaps deserves its own ADR), or switching from Google's gRPC implementation to [Connect](https://connect.build/).

[Rendered](https://github.com/cometbft/cometbft/blob/thane/adr-protobuf-versioning/docs/architecture/adr-103-proto-versioning.md)

@mzabaluev has already been working on implementing this on the [`feature/proto-upgrade` branch](https://github.com/cometbft/cometbft/tree/feature/proto-upgrade).

---

#### PR checklist

- [ ] Tests written/updated
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments

PR #817, opened 2023-05-08
ADR-103: Protobuf definition versioning (#772)

thanethomson:
Port of #772 to `main`.

If we don't reserve the numbers we may end up with clashes in future. If for some unlikely reason we decide not to version our protos, we can just mark this as rejected.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/docs/architecture/README.md
+++ b/docs/architecture/README.md
@@ -42,6 +42,8 @@ numbering our ADRs from 100 onwards.
 
 ### Proposed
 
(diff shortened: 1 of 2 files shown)


lasarojc (commented):
Nice write up. This seems to capture the issue well, but I will defer to @mzabaluev to approve.

thanethomson:
@sergio-mena already approved it in #772 - I'm just bringing the exact same doc to `main` to minimize the chance of conflicts with other ADRs currently underway.

lasarojc (approved):

PR #822, opened 2023-05-10
Specifies the CRDT used for Gossiping in Consensus

lasarojc:
Gossiping is specified in terms of a CRDT.
Pull requests over branch gossip-spec will serve to get early feedback until gossip-spec is ready to be merged into main.


spec/consensus/gossip/crdt.md
otrack, line 1: In the definition of Eventual $\Delta$-Timely Communication, $\Delta$ is unknown to the processes.

In the definition of Gossip Communication property, I do not understand that "all the correct processes will receive $m$ before $max(t, GST)+\Delta$". The upper bound $\Delta$ only holds after $GST$, hence delivery is only guaranteed after $GST+\Delta$, and $t$ has no role to play here.

spec/consensus/gossip/crdt.md
```diff
@@ -41,25 +41,19 @@ In other words, the algorithm does not require all messages to be delivered, onl
...
 
 These approaches work to reach convergence because the updates are commutative regarding the tuple space; each update simply adds an entry to a set.
-From the Tendermint algorithm's point of view, convergence guarantees progress but is not requirement for correctness.[^todo2]
-In other words, nodes observing different approximations of the tuple space may decide at different point in time but cannot violate any correctness guarantees and the eventual convergence of tuple space implies the eventual termination of the algorithm.
-
-[^todo1]: Formalize using [Making CRDTs Byzantine Fault Tolerant](https://martin.kleppmann.com/papers/bft-crdt-papoc22.pdf) as basis; "Many CRDTs, such as Logoot [44] and Treedoc [ 36], assign a unique identifier to each item (e.g. each element in a sequence); the data structure does not allow multiple items with the same ID, since then an ID would be ambiguous.”
+From the Tendermint algorithm's point of view, convergence guarantees progress but is not a requirement for correctness.
```
otrack, line 54: In approach two, I would replace "gossip/anti-entropy" with anti-entropy.

It worths pointing out that approach one requires unbounded space, so in some sense it is not practical.

spec/consensus/gossip/crdt.md
```diff
@@ -68,59 +62,64 @@ To save memory, entries should be removed from the tuple space as soon as they b
 For example, if a new height is started, all entries corresponding to previous heights become stale.
 
 In general, simply forgetting stale entries in the local view would save the most space.
-However, it could lead to entries being added back and never being completely purged from the system.
-Although stale entries do not affect the algorithm, or they would not be considered stale, not adding the entries back for performance and resource utilization sake.
+However, if the second approach described [above](#nodes-state-as-a-tuple-space), it could lead to entries being added back and never being completely purged from the system.
```
otrack, line 65: I am not sure to understand why if a new height is started, all the entries corresponding to previous heights become stale. We need to have a decision at that height.

spec/consensus/gossip/crdt.md
```diff
@@ -68,59 +62,64 @@ To save memory, entries should be removed from the tuple space as soon as they b
...
+Although stale entries do not affect the algorithm, or they would not be considered stale, not adding the entries back is important for performance and resource utilization sake.
 
-One approach to prevent re-adding entries may be prevented by keeping _tombstones_ for the removed entries.
-With time, however, even small tombstones may accrue and need to be garbage collected, in which cause the corresponding entry may be added again; again, this will not break correctness and as long as tombstones are kept for long enough, the risk of re-adding is minimal.
+One way to prevent re-adding entries is keeping _tombstones_ for the removed entries.
+A tombstone is nothing but an that supersedes a specific other entry.
```
otrack, line 69: Typo.

spec/consensus/gossip/crdt.md
```diff
@@ -68,59 +62,64 @@ To save memory, entries should be removed from the tuple space as soon as they b
...
 
-Let $\text{ValSet}_h \subseteq P$ be the set of validators of height $h$ and $\text{Prop}_{h,r}$ be the proposer of round $r$ of height $h$.
+Let $V_h \subseteq P$ be the set of validators of height $h$ and $\pi^h_r \in V_h$ be the proposer of round $r$ of height $h$.
+When the context eliminates any ambiguity on the height number, we might write these values simply as $V$ and $\pi_r$.
 
-Given that each validator can execute each step only once per round, a query that specifies height, round, step and validator must either return empty or a single tuple.
+Given that each validator can execute each step only once per round, a query that specifies height, round, step and validator SHOULD either return empty or a single tuple.
 
-- $\forall h \in \N, r \in \N, s \in \{\text{Proposal, PreVote, PreCommit}\}, v \in \text{ValSet}_h$,  $\cup_{p \in P} \lang h, r, s, v, * \rang_p$ contains at most one element.
+- $\forall h \in \N, \forall r \in \N, \forall s \in \{\text{Proposal, PreVote, PreCommit}\}, \forall v \in V_h$,  $\cup_{p \in P} \lang h, r, s, v, * \rang_p$ contains at most one element.
```
otrack, line 118: I do not understand $\cup{p \in P} (h,r,s,v,*)$ contains at most one element.

In general, is it not better to have these preconditions on the tuple spaces and consider that they are enforced at all time?

spec/consensus/gossip/crdt.md
```diff
@@ -130,48 +129,178 @@ Consider the following definition for **Eventual Convergence**.
 
 |Eventual Convergence|
 |-----|
-| If there exists a correct process $p$ such that $e \in t_p$, then, eventually, for every correct process $q$, $e \in t_q$ or there exists a correct process $r$ such that $\bar{e} \in t_r$.|
-
-> **Note**
-> Nodes may learn of an entry deletion before learning of its addition.
+| If there exists a correct process $p \in P$ such that $e \in t_p$, then, eventually, for every correct process $q \in P$, either $e \in t_q$ or $e$ is stale in $t_q$.
```
otrack, line 132: $t_p$ is not defined earlier. I would maybe opt for $tuples_p$, or possibly $p.tuples$ without the subscript.

the definition of _stale in_ is missing (additional comments about this below).

spec/consensus/gossip/crdt.md
```diff
@@ -130,48 +129,178 @@ Consider the following definition for **Eventual Convergence**.
...
+In order to ensure convergence even in the presence of failures, the network must be connected in such a way to allow communication around any malicious nodes, that is, to provide paths connecting correct nodes.
+Even if paths connecting correct nodes exist, effectively using them requires timeouts to not expire precociously and abort communication attempts.
+Timeout values can be guaranteed to eventually be enough for communication after a GST is reached, which implies that all communication between correct processes will eventually happen timely, which implies that the tuple space will converge and keep converging.
 Formally, if there is a GST then following holds true:
 
 | Eventual $\Delta$-Timely Convergence |
 |---|
-| If $e\in t_p$, for some correct process $p$, at instant $t$, then by $\text{max}(t,\text{GST}) + \Delta$, either $e \in t_q$, for every correct process $q$ or $e$ is stale in \bar{e} \in t_p$.
+| If $e\in t_p$, for some correct process $p \in P$, at instant $t$, then by $\text{max}(t,\text{GST}) + \Delta$, for every correct process $q \in P$, either $e \in t_q$ or $e$ is stale in $t_p$.
```
otrack, line 141: then _the_ following holds true

I am not sure to understand the benefit of $\Delta$-timely convergence over eventual converge, as the later suffices.

spec/consensus/gossip/crdt.md
```diff
@@ -130,48 +129,178 @@ Consider the following definition for **Eventual Convergence**.
...
+Operation-based CRDT use reliable communication to ensure that all updates (operations) are delivered to all replicas.
+If the reliable communication primitive precludes duplication, then applying all operations will lead to the same state, irrespective of the delivery order since operations are commutative.
+If duplications are allowed, then the operations must be made idempotent somehow.
+
+State-based CRDT do not rely on reliable communication.
+Instead they assume that replicas will compare their states and converge two-by-two using a merge function;
+as long as the function is commutative, associative and idempotent, the states will converge.
+For example, in the G-Set case, the merge operator is simply the union of the sets.
+
+The two approaches for converging the message sets in the Tendermint algorithm described [earlier](#nodes-state-as-a-tuple-space), without the deletion of entries, correspond to the operation- and state-based [Grow-only Set](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type#G-Set_(Grow-only_Set)) CRDT;
```
otrack, line 188: _earlier_ is not pointing toward the right reference I think.

spec/consensus/gossip/crdt.md
```diff
@@ -130,48 +129,178 @@ Consider the following definition for **Eventual Convergence**.
...
+if removals must be handled, then using a 2P-Set is an option.
+To the best of our knowledge, no existing CRDT supports superseding of elements, until now.
```
otrack, line 190: I would give some intuition about what does mean _supeseeding_ here. 

I believe that, on the contrary, this is the very essence of CRDTs, that is once a state $u$ is reached (whatever how), any of the prefixes of $u$ in the lattice are included/superseded.

spec/consensus/gossip/crdt.md
```diff
@@ -130,48 +129,178 @@ Consider the following definition for **Eventual Convergence**.
...
+
+### About supersession
+
+We call an entry of the tuple space an **Entry**[^tuple] and call **View** a set of Entry.
```
otrack, line 194: I would write "a set of entries".

spec/consensus/gossip/crdt.md
```diff
@@ -130,48 +129,178 @@ Consider the following definition for **Eventual Convergence**.
...
+
+[^tuple]: We use Entry instead of Tuple not to conflict with the use of tuple in Quint.
+
+Let $v_1, v_2, v_3$ be **View**.
```
otrack, line 198: be views

Beware, because $v$ is already used for values. One solution is to use blocks for values and keep $v$ for views

spec/consensus/gossip/crdt.md
```diff
@@ -130,48 +129,178 @@ Consider the following definition for **Eventual Convergence**.
...
+We say that $v_1$ is superseded by $v_2$ if all elements of $v_1$ are rendered stale by some subset of $v_2$ and note this relation as $v_1 < v_2$.
```
otrack, line 199: again, the definition of stale is missinig
imho, this should be something like:
consider two entries e=(h,r,s,v,p) and e'=(h',r',s',v',p')
then
e < e'= \lor h < h' // still unclear to me if we need h as a parameter
    	\lor h=h' \land r<r' \land PoER(r) // PoER is a proof that round r has ended
	\lor h=h' \land r=r' \land s<s' // with the usual order PROPOSAL < PRE-VOTE < PRE-COMMIT
by extension < can also be defined naturally across views
in which case, v < v' = \forall e \in v. \exists e' \in v'. e < e'

spec/consensus/gossip/crdt.md
```diff
@@ -130,48 +129,178 @@ Consider the following definition for **Eventual Convergence**.
...
+While an equivalent operations-based definition must exist, it is out of the scope of this document.
+
+
+#### The merge operator - $\sqcup$
+
+State based CRDT use a merge operation to combine two replica states.
+In our CRDT we embed the removal of superseded messages, as defined in the previous section, in the merge operation.
+
+Let $\mathcal{V}$ be the power set of entries, or the set of all views of a system.
+The merge operator $\sqcup: \mathcal{V} x \mathcal{V} \rightarrow \mathcal{V}$ is defined as the union of the input views, without the entries superseded by the other elements in the union, that is: $v_1 \sqcup v_2 = [e: e \in (v_1 \cup v_2) \land e < (v_1 \cup v_2 \setminus \{e\})]$
```
otrack, line 230: Unclear that this is actually returning the least upper bound of $\\{v_1,v_2\\}$. 

If the superseding relation is defined as proposed in the comment above, we may construct a merge operator as follows: starting from v_2, one iterates over all the elements of v_1. For each of them e, if e is superseding it is added and all the lower elements removed from the resulting set. (There might be a better implem. thought. For that, we can look at soa CRDT implementations.)

otrack (commented):
Overall, I believe that specifying the state of a (BFT) SMR protocol as a CRDT is a very interesting approach. From a curiosity point of view, it would be interesting to see if this can be done solely with set of messages. However, maybe it is easier to have also some state with predicate over it. For instance, as suggested in the comments below, a base predicate can be $PoE(r)$ which returns $true$ if there is a proof that some round $r$ has ended.

lasarojc:
> Overall, I believe that specifying the state of a (BFT) SMR protocol as a CRDT is a very interesting approach. From a curiosity point of view, it would be interesting to see if this can be done solely with set of messages. However, maybe it is easier to have also some state with predicates over it. For instance, as suggested in the comments below, a base predicate can be  which returns  if there is a proof that some round  has ended.

Agree that having these predicates will make it easier to specify and read. This is the approach I will be taking on the supersession operator and probably other places as well.

spec/consensus/gossip/crdt.md
lasarojc: t was used here to handle the case in which m was sent **after** GST. 
In this case, because the message may still require up to Delta to be delivered, it can only be guaranteed to be delivered after t+Delta, which may come much later than GST+delta..

spec/consensus/gossip/crdt.md
lasarojc: 👍 Fixed the second. The first is pointed out in the following section.

spec/consensus/gossip/crdt.md
lasarojc: You are right, the decision cannot be forgotten, but because we can get previous decisions out-of-band (from the tuple space point of view), we can forget them here.

spec/consensus/gossip/crdt.md
lasarojc: thanks. fixed on the follow up PR.

spec/consensus/gossip/crdt.md
lasarojc: I didn't understand your comment.
What I tried to express here is that there should be a single vote per validator per (height, round).
Before the vote is issued, the query result is empty. If the validator behaves, eventually there may be one entry in the set. But if the validator misbehaves, there could be more than one value in the result.

spec/consensus/gossip/crdt.md
lasarojc: thanks for the catch. it escaped from a previous write up. removed now on the follow up PR.

spec/consensus/gossip/crdt.md
lasarojc: hmmm... the important difference at the time of the writing was that e may get stale on q, in eventual convergence, but in eventual delta timely convergence it can get stale in $p$, and never forwarded to $q$, but I will have to think more about it being actually required or not.

spec/consensus/gossip/crdt.md
otrack: That's very true.

cason (commented):
I think the use of CRDTs is well supported by this document, that covers some empty gaps existing in the previous version.

However, I think this document would benefit from having the merge/tombstone operators being used, even as an example, for the actual consensus messages.

Further recommendations:
1.  Avoid changing code, even if correcting misspelled comments in code, in spec PRs
2. Try breaking long lines into smaller ones for easy reading and better version control
3. Check the math symbols adopted and make sure they are rendered by GitHub

spec/consensus/gossip/crdt.md
lasarojc: Good point. Although it is an implementation detail of `getValue` and `valid`, it might be worth capturing it here. I will consider it in specifying the superseding function, in the follow up PR.

spec/consensus/gossip/crdt.md
lasarojc: I see. $(h,r,s,v,*)$ is supposed to be a query, whose result is a set. I will improve on the notation.

You are absolutely right about more than one message existing here being related to that discussion.

spec/consensus/gossip/crdt.md
lasarojc: done.

consensus/state.go
```diff
@@ -387,7 +387,7 @@ func (cs *State) OnStart() error {
 }
 
 // timeoutRoutine: receive requests for timeouts on tickChan and fire timeouts on tockChan
-// receiveRoutine: serializes processing of proposoals, block parts, votes; coordinates state transitions
+// receiveRoutine: serializes processing of proposals, block parts, votes; coordinates state transitions
```
lasarojc, line 390:
```suggestion
// receiveRoutine: serializes processing of proposoals, block parts, votes; coordinates state transitions
```

consensus/state.go
```diff
@@ -2001,7 +2001,7 @@ func (cs *State) handleCompleteProposal(blockHeight int64) {
 	if cs.Step <= cstypes.RoundStepPropose && cs.isProposalComplete() {
 		// Move onto the next step
 		cs.enterPrevote(blockHeight, cs.Round)
-		if hasTwoThirds { // this is optimisation as this will be triggered when prevote is added
+		if hasTwoThirds { // this is an optimization as this will be triggered when prevote is added
```
lasarojc, line 2004:
```suggestion
		if hasTwoThirds { // this is an optimisation as this will be triggered when prevote is added
```

consensus/state.go
```diff
@@ -2001,7 +2001,7 @@ func (cs *State) handleCompleteProposal(blockHeight int64) {
 	if cs.Step <= cstypes.RoundStepPropose && cs.isProposalComplete() {
 		// Move onto the next step
 		cs.enterPrevote(blockHeight, cs.Round)
-		if hasTwoThirds { // this is optimisation as this will be triggered when prevote is added
+		if hasTwoThirds { // this is an optimisation as this will be triggered when prevote is added
```
lasarojc, line 2004:
```suggestion
		if hasTwoThirds { // this is optimisation as this will be triggered when prevote is added
```

spec/consensus/gossip/README.md
```diff
@@ -5,6 +5,8 @@
 This specification is divided into multiple documents and should be read in the following order:
 
 - [layers.md](layers.md): describes the architecture used in CometBFT and where this specification is focused;
+- [crdt.md](crdt.md): explains the rationale of using a CRDT in the gossiping and defines the CRDT used, SSE;
+- [sse.qnt](sse.qnt): Quint spec of our ;
```
lasarojc, line 9:
```suggestion
- [sse.qnt](sse.qnt): Quint spec with example instantiations of the proposed CRDT;
```

lasarojc:
> I think the use of CRDTs is well supported by this document, that covers some empty gaps existing in the previous version.
> 
> However, I think this document would benefit from having the merge/tombstone operators being used, even as an example, for the actual consensus messages.
> 
> Further recommendations:
> 
> 1. Avoid changing code, even if correcting misspelled comments in code, in spec PRs
> 2. Try breaking long lines into smaller ones for easy reading and better version control
> 3. Check the math symbols adopted and make sure they are rendered by GitHub

Wrt 1. reverted.
Wrt 2. I prefer breaking on period ends exactly for better version control, as I won't have to move words around to rebalance lines when some small edit happens.
Wrt 3. Will do in the follow up PR.

lasarojc:
@cason Are you ok with me addressing the the remaining comments in the follow up PR?

spec/consensus/gossip/crdt.md
```diff
@@ -130,48 +129,178 @@ Consider the following definition for **Eventual Convergence**.
...
+We say that $v_1$ is superseded by $v_2$ if all elements of $v_1$ are rendered stale by some subset of $v_2$ and note this relation as $v_1 < v_2$.
+For all $e \in v_1$, if $v_1 < v_2$ then we abuse the notation and terminology and simply say that $e$ is superseded by $v_2$ and note it as $e < v_2$
+
+The supersession $<$ relation must respect the following properties:[^not_contains]
+
+[^not_contains]: $v_1 \subset v_2 \implies v_1 < v_2$ but $v_1 < v_2 \not\implies v_1 \subset v_2$
+
+1. Transitivity: if $v_1 < v_2$ and $v_2 < v_3$, then $v_1 < v_3$;
+1. Reflexiveness: $v < v$;
+1. Anti-symmetry: if $v_1 < v_2$ and $v_3 < v_2$ then $v_1 = v_2$;
```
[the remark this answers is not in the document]
cason: Is this a typo, our didn't I get the point?

PR #826, opened 2023-05-12
Fixed distribution catalog API endpoint can lead to OOM via malicious…

iot-defcon:
… user input

The /v2/_catalog endpoint was designed specifically to do registry syncs with search or other API systems. Such an endpoint would create a lot of load on the backend system, due to overfetch required to serve a request in certain implementations.

Because of this, we strongly recommend to always this API endpoint behind heightened privilege and avoid leaving it exposed to the internet.

Background
==========

/v2/_catalog endpoint accepts a parameter to control the maximum amount of records returned (query string: n).

When not given the default n=100 is used. The server trusts that n has an acceptable value, however when using a maliciously large value, it allocates an array/slice of n of strings before filling the slice with data.

Have a running registry with at least one image on it. and pass a  sufficiently long
`n` to the `/v2/_catalog`.

---
$  = host machine shell A
%  = host machine shell B
#  = container
-- = comment

## Impact
Systems that run `distribution` built after a specific commit running on memory-restricted environments can suffer from denial of service by a crafted malicious `/v2/_catalog` API endpoint request.

CVE-2023-2253
`CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H`

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -102,7 +102,7 @@ require (
 	github.com/dgraph-io/ristretto v0.1.1 // indirect
 	github.com/dgryski/go-farm v0.0.0-20200201041132-a6ae2369ad13 // indirect
 	github.com/docker/cli v23.0.5+incompatible // indirect
-	github.com/docker/distribution v2.8.1+incompatible // indirect
+	github.com/docker/distribution v2.8.2+incompatible // indirect
 	github.com/docker/docker v23.0.5+incompatible // indirect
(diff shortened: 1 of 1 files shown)


thanethomson:
Thanks for this @iot-defcon, but the `go.sum` changes are missing from this PR. This will be superseded by #826.

This is an indirect dependency that isn't used by anything in production code, so it's not a security concern for us.

PR #829, opened 2023-05-12
rpc: Remove response data from response failure logs

### LINKED ISSUE #654: RPC server: Long Log Messages
### Summary

Syslog has trouble handling these long strings (see below). Not sure if they are even needed in logs, but it would surely be great to be able to suppress them, even if syslog gets adjusted to handle them (max message size), just to spare disk IO and reduce log size.

### Problem Definition

https://paste2.org/xbEIvUhb 

### Acceptance criteria

- [ ] truncate the log output in the example above
    - decide if the `txs` raw bytes are useful in the log in the first place (most likely not useful)
        - if useful: keep a truncated version;
        - if not useful: remove the `txs` bytes entirely from the log output

thanethomson:
Closes #654.

Backporting to v0.38, v0.37 and v0.34, so I'll add changelog entries on the backport branches.

---

#### PR checklist

- [ ] Tests written/updated


thanethomson:
@gaia do you think setting a max limit of 900 characters for the response data sufficient? (to try to keep overall log line length < 1024 chars)

gaia:
> @gaia do you think setting a max limit of 900 characters for the response data sufficient? (to try to keep overall log line length < 1024 chars)

IMHO, anything over 1024 chars total should be broken up into multiple lines and/or most of the time the information (in the Cosmos SDK universe) likely doesn't belong there.

That being said, as far as the RFCs and default settings involved in Syslog (protocol) and Rsyslog (most popular syslog daemon these days), they are 2K ([RFC 5426 section 3.2](https://www.rfc-editor.org/rfc/rfc5426#section-3.2)) and 8K ([manual](https://www.rsyslog.com/doc/master/configuration/global/index.html#true-global-directives)), respectively.

thanethomson:
> IMHO, anything over 1024 chars total should be broken up into multiple lines and/or most of the time the information (in the Cosmos SDK universe) likely doesn't belong there.

I'd be perfectly happy to remove the response logging entirely. It's not clear to me though whether there's anyone out there that actually relies on it.

gaia:
> > IMHO, anything over 1024 chars total should be broken up into multiple lines and/or most of the time the information (in the Cosmos SDK universe) likely doesn't belong there.
> 
> I'd be perfectly happy to remove the response logging entirely. It's not clear to me though whether there's anyone out there that actually relies on it.

that'd be great. by all means, if they do, should be redirected to a special log, non default.

lasarojc (approved):
LGTM

Issue #476, opened 2023-03-07
privval: Implement gRPC support


thanethomson:
Originally tracked in tendermint/tendermint#4698 and implemented in tendermint/tendermint#5725 in the v0.35/v0.36 release series.

We should bring these changes into whichever releases we can.

The corresponding work in tendermint-rs, for instance, is documented in [ADR-011](https://github.com/informalsystems/tendermint-rs/blob/0f6ce751095aebd37bf76ab0353bfc15d9a4d06a/docs/architecture/adr-011-validator-crate.md). See also informalsystems/tendermint-rs#1134.

adizere:
Needed for https://github.com/strangelove-ventures/horcrux/issues/39.

tony-iqlusion:
I had some suggestions for error handling I left in this comment I wanted to make sure don't get lost in the shuffle: https://github.com/cometbft/cometbft/pull/850#issuecomment-1664586786

It's been a major problem with multiple chain upgrades that prevents a validator with a remote signer from coming online, so it'd be really great to get it addressed.

PR #850, opened 2023-05-17
docs: ADR-106: gRPC API

thanethomson:
Partially addresses #81.

[:book: Rendered](https://github.com/cometbft/cometbft/blob/thane/81-grpc-adr/docs/references/architecture/adr-106-grpc-api.md)

There's already a partial implementation of this on the `feature/adr101-pull-companion` branch. See #818.

---

#### PR checklist

- [ ] Tests written/updated


docs/architecture/adr-105-grpc-api.md
```diff
@@ -0,0 +1,222 @@
...
+- `VersionService` - A simple service that aims to be quite stable over time in
+  order to be utilized by clients to establish the version of the software with
+  which they are interacting (e.g. to pre-emptively determine compatibility).
+  This could technically be part of the `NodeService`, but if the `NodeService`
+  interface were to be modified, a new version of the service would need to be
+  created, and all old versions would need to be maintained, since the
+  `GetVersion` method needs to be quite stable.
+  - `GetVersion` - Query the version of the software and protocols employed by
+    the node (e.g. CometBFT, ABCI, block, P2P and application version).
+- `NodeService` - Provides information about the node providing the gRPC
```
lasarojc, line 71: could the `NodeService` run on a machine different from the one running the full-node?
If so, is the information returned here from the node serving gRPC or the full node or both?

docs/architecture/adr-105-grpc-api.md
```diff
@@ -0,0 +1,222 @@
...
+
+- `VersionService` - A simple service that aims to be quite stable over time in
+  order to be utilized by clients to establish the version of the software with
+  which they are interacting (e.g. to pre-emptively determine compatibility).
+  This could technically be part of the `NodeService`, but if the `NodeService`
+  interface were to be modified, a new version of the service would need to be
+  created, and all old versions would need to be maintained, since the
+  `GetVersion` method needs to be quite stable.
+  - `GetVersion` - Query the version of the software and protocols employed by
+    the node (e.g. CometBFT, ABCI, block, P2P and application version).
```
lasarojc, line 70: could add "version of any companions" to the list, even though it does not need to be exhaustive.

docs/architecture/adr-105-grpc-api.md
thanethomson: No, Protobuf "services" are just logical ways of grouping sets of RPC methods. These services are always exposed from a full node or a validator.

docs/architecture/adr-105-grpc-api.md
thanethomson: I wouldn't. That would involve developing some more complex handshake mechanism with the data companion, for which I don't see a need at the moment.

docs/architecture/adr-105-grpc-api.md
thanethomson: Ah, I see. If the RPC companion had to also provide the same gRPC interface, it would proxy such methods. There are only a handful of methods where it would pull its data from a separate database, e.g. fetching blocks, block results, historical transactions, etc.

But methods like getting node status, performing an ABCI query, etc. have to be proxied through to the upstream full node to which the companion is attached.

@andynog is looking into this in ADR 102.

thanethomson: Nothing stops the companion from exposing _additional_ services that are specific to the companion itself. But I don't think it should modify or expand on any of the data or services it proxies.

romac:
Looks great!

romac:
> Glad to hear you think so! Are there any other endpoints that you'd like to see exposed here @romac?

Here's the full list of RPC endpoints we use from Hermes: https://hedgedoc.informal.systems/FmNN0xuESR2wS1ZoRhuU8g

Aside from the first two endpoints which are strictly required for relaying (`/health` & `/consensus_params`), the gRPC list is missing these three endpoints:

- `/header`
- `/tx_search`
- `/block_search`

While it would be nice to get rid of JSON-RPC altogether, since we'll have to maintain multi-version support for some time we could initially do without and keep hitting the JSON-RPC endpoints for those, and wait until they are added at a later time.

lasarojc (approved):

thanethomson:
@hdevalence I've added a `NetworkService` from which you can query the genesis data and currently connected peers - does that suffice? Do you need the listener info that the `/net_info` provides?

@romac I've updated the endpoints - will this suffice for the relayer?

romac:
> @romac I've updated the endpoints - will this suffice for the relayer?

I might be missing something but I believe the `/latest_commit`, `/commit`, and `/validators` endpoints are missing. Those are required by the light client to fetch and verify light blocks.

Aside from that, looks great! 🙏

tony-iqlusion:
@thanethomson what about privval support? (previously tracked as https://github.com/tendermint/tendermint/issues/9256)

If that's in the cards, I have some feedback on the design of that API based on some recent experiences (namely it would be nice if there were an error which KMS systems can use to communicate the last signed h/r/s back to CometBFT in the event of double signing, so CometBFT could potentially update its round counter in response)

thanethomson:
> what about privval support?

That could definitely be on the cards @tony-iqlusion. We do have a separate tracking issue for that in #476, but we'd be very interested to hear what your ideal API/interaction model would look like.

tony-iqlusion:
Really all I'd like is a documented standard way of communicating a double signing error from `sign_vote` and `sign_proposal` and including metadata that CometBFT can parse and use to inform future signing requests. Something like the following:

- On double signing error, return 409 Conflict to indicate a signature at the requested height already exists
- The error should include metadata for `last_height` in `h/r/s` format, which reflects the last height a signature was computed for

If `last_height` reflects the current block height, but CometBFT's round counter is lower than the KMS's, then CometBFT can advance the round counter to one more than the one included in `last_height`.

Right now CometBFT does not handle `privval` errors very well. This was pretty painful with the last Neutron halt, where the chain halted for several hours and validators needed to resync. When the halt height was reached syncing, it took an additional hour for validators using TMKMS to reach the correct round to participate in consensus. It seems settings like `timeout_prevote(_delta)` and `timeout_precommit(_delta)` are not properly honored in this case (which perhaps I should report as a separate bug).

melekes:
@thanethomson what's the state of this work? Should we take over here?

thanethomson:
> @thanethomson what's the state of this work? Should we take over here?

Yes please @melekes!

adizere (approved):
Pre-approving based on a sync review with Andy today.

What is left:
* double-check comments from Henry & Tony & Romain
* validate the Go API server instantiation
* double-check potential name conflict "ADR 105"

PR #852, opened 2023-05-17
`e2e` provider is extended with `Disconnect`, `Reconnect` and `CheckUpgraded`

### LINKED ISSUE #58: Update the e2e runner to disconnect a node from network in Digital Ocean testnet
Tasks:

- [x] Familiarize with tendermint/tendermint#9860
- [x]  Clone it to Comet (merge to `main`)
- [x] Drive tendermint/tendermint#9860 to completion, including any non-CI testing needed
  - [x] #852
- [x] Test on DO
  - [x] cometbft/qa-infra#17

DoD:

* Patch finalized, reviewed, approved, and merged

Original issue: tendermint/tendermint#9950

sergio-mena:
Closes #58

This PR is completing tendermint/tendermint#9860 with the extension of `Provider` interface with methods `Disconnect`, and `Reconnect`, and their implementation for the Digital Ocean infrastructure provider. Method `CheckUpgraded` is also introduced but only implemented for the Docker provider.

The DO implementation uses ansible (just as the other commands), and `iptables` to disconnect the node from the p2p network. The use of ansible follows its usage in `cometbft/qa-infra`.

See how this can be used on a Digital Ocean testnet in cometbft/qa-infra#17

---

#### PR checklist

- [ ] Tests written/updated


test/e2e/pkg/infra/digitalocean/digitalocean.go
```diff
@@ -24,33 +25,63 @@ func (p *Provider) Setup() error {
...
+func (p Provider) Reconnect(ctx context.Context, _ string, ip string) error {
+	playbook := ansiblePerturbConnectionBytes(false)
+	playbookFile := getNextPlaybookFilename()
+	if err := p.writePlaybook(playbookFile, playbook); err != nil {
+		return err
+	}
+	return execAnsible(ctx, p.Testnet.Dir, playbookFile, []string{ip})
+}
+
+func (p Provider) CheckUpgraded(ctx context.Context, node *e2e.Node) (string, bool, error) {
```
sergio-mena, line 79:
```suggestion
func (p Provider) CheckUpgraded(_ context.Context, node *e2e.Node) (string, bool, error) {
```

test/e2e/pkg/infra/digitalocean/digitalocean.go
```diff
@@ -24,33 +25,63 @@ func (p *Provider) Setup() error {
...
 }
 
-const ymlSystemd = "systemd-action.yml"
+var ymlPlaybookSeq int
+
+func getNextPlaybookFilename() string {
+	const ymlPlaybookAction = "playbook-action"
+	ymlPlaybookSeq++
+	return ymlPlaybookAction + strconv.Itoa(ymlPlaybookSeq) + ".yml"
+}
```
thanethomson, line 34: Is this so that playbooks don't clash with one another?

test/e2e/pkg/infra/digitalocean/digitalocean.go
```diff
@@ -60,25 +91,58 @@ func (p Provider) writePlaybook(yaml string, starting bool) error {
...
 
-// file as bytes to be written out to disk.
-// ansibleStartBytes generates an Ansible playbook to start the network
-func ansibleSystemdBytes(starting bool) string {
-	startStop := "stopped"
-	if starting {
-		startStop = "started"
-	}
-	playbook := fmt.Sprintf(`- name: start/stop testapp
+const basePlaybook = `- name: e2e custom playbook
```
thanethomson, line 94: While I'm pretty confident that this would work, it is rather difficult to know what the final playbook for each provider method would look like.

Would it not make it more readable to define a [template](https://pkg.go.dev/text/template) per provider method (e.g. one for `StartNodes`, one for `StopTestnet`, one for `Reconnect`, etc.) and simply parameterize the parts that need to be variable?

thanethomson (commented):
Glad to see this coming together, thanks @sergio-mena!

test/e2e/pkg/infra/digitalocean/digitalocean.go
sergio-mena: It's for troubleshooting, a late addition to this PR. I wanted to see what _actual_ playbook had been generated, but the runner was overwriting the file with new playbooks.
This way we keep the history, and can inspect.
All these files are part of the "testnet" dir, so they will be wiped out on the next run (or upon cleanup). But, if we see that the number of files is very large for a run (not the case now), we could extend this to only produce, e.g., N files and use them in round-robin.

test/e2e/pkg/infra/digitalocean/digitalocean.go
sergio-mena: > making it more difficult to debug in case something goes wrong

Regarding this, check my comment above

Yes, I thought about using the template library, but in the end I decided to go simple. As we add more logic from `qa-infra`, we can evaluate whether to move over to templates.

thanethomson: Interestingly I would consider using the template library to be the simpler option from a coding/debugging perspective. My concern's not so much with the generated code as it is with the Go code and the strings within the Go code.

For instance, some of the string concatenations are performed with strings with indentation where that indentation has meaning. It's usually far easier to reason about and extend when the whole template's visible, like with the Docker compose template: https://github.com/cometbft/cometbft/blob/dc28d6bd982b48a446ef9299cfc241d0da075c78/test/e2e/pkg/infra/docker/docker.go#L53-L117

For the current code, we need to now know that some strings require indentation with 4 spaces, others with 2 spaces, etc., making it more difficult to debug/extend since you can't see those strings within their intended context.

thanethomson (approved):
We don't need to block on this, but I'd highly recommend reworking the Ansible script generation to use Go plain text templates at some point soon.

test/e2e/pkg/infra/digitalocean/digitalocean.go
sergio-mena: You've got a point with the indentation. Current approach is not the best for that, I agree.
There's also another consideration that made me lean toward the current approach (for the time being). Playbooks have two parts:

* the initial part which is kind of static, so it renders itself well for templating
* the tasks, which are additive, so it's less obvious to me how a single template would help (unless you duplicate contents in many templates, as we have it in `qa-infra`)

Maybe, a solution that represents a best balance would be to have a template for the beginning, and then other templates, one for each of the different kind of tasks we support (shell, systemd, apt, etc).

However, I'd like to defer this to the next round of improvements.

Issue #843, opened 2023-05-16
qa: impact of length/presence of vote extensions


sergio-mena:
- [ ] define the test case (should have its section in RELEASES.md)
- [ ] define metrics we want to gather
- [ ] implementation (infra) - mostly qa-infra
- [ ] test execution (warning: first run, not beaten path)
- [ ] report section

PR #858, opened 2023-05-18
Adds `vote_extension_size` to manifests

lasarojc:
Contributes to #843

In order to evaluate the impact of vote extensions on systems' performance (#843) we need to test the system with different vote extension sizes. This PR addresses this need.

As a side effect, added another build target for tests (`docker-debug`) which builds a docker image with debug versions of CometBFT and the e2e kvstore app.
I will write a KB note on how to use it.

Changes in `common.mk`, `makefile` `Dockerfile.debug`  `entrypoint-delve` and `docker.go` could be back ported to add debug builds to previous versions.

---

#### PR checklist

- [ ] Tests written/updated


RELEASES.md
```diff
@@ -370,5 +370,17 @@ the total stake. The remaining 33% of the stake is configured to belong to
 a validator that is never actually run in the test network. The network is run
 for multiple days, ensuring that it is able to produce blocks without issue.
 
+#### Vote-extension Testnet
```
sergio-mena, line 373: Can we move this one before `Network Partition Testnet` (so we leave those never run at the end)?

RELEASES.md
```diff
@@ -370,5 +370,17 @@ the total stake. The remaining 33% of the stake is configured to belong to
...
+
+CometBFT v0.38.0 introduced **vote-extensions**, which are added as the name suggests, to votes sent by validators.
```
sergio-mena, line 375:
```suggestion
CometBFT v0.38.0 introduced **vote-extensions**, which are added as the name suggests, to precommit votes sent by validators.
```

RELEASES.md
```diff
@@ -370,5 +370,17 @@ the total stake. The remaining 33% of the stake is configured to belong to
...
+The Vote-extension Testnet is used to determine how vote-extensions affect the performance of CometBFT, under various settings.
+The testnet consists of 175 validators, 20 non-validator full-nodes, and 5 seed nodes.
+All 195 full-nodes begin by dialing a subset of the seed nodes to discover peers.
+Once all full-nodes are started, a 5 minute period is waited while load-generators are started and before any load is introduced in the system.
+Once the startup period is over, the load generators issue requests at a constant rate during 150 seconds.
+The load generation step is repeated 5 times with intervals of 60 seconds in between to allow the system to quiesce.
+A wide set of metrics is captured during the whole experiment but, unless abnormalities are observed, the first and last 30 seconds of each 150 interval are discarded; the remainder 90 seconds are used in computing aggregated metrics.
+The application used in the experiment is the same used on the (#200-node-testnet), so the results are comparable.
+
```
sergio-mena, line 384: Should we describe how the length of vote extensions is going to vary during the test execution?

common.mk
```diff
@@ -12,6 +12,9 @@ LD_FLAGS += $(LDFLAGS)
 ifeq (,$(findstring nostrip,$(COMETBFT_BUILD_OPTIONS)))
   BUILD_FLAGS += -trimpath
   LD_FLAGS += -s -w
+else
+  #prepare for delve
+  BUILD_FLAGS+= -gcflags "all=-N -l"
```
sergio-mena, line 17: I'm not sure I get why this is an `else` and not an independed "if" with its own option.
The "else" here means we want to strip the binary... but usually, we want to do the opposite for debugging

test/e2e/app/app.go
```diff
@@ -439,18 +446,28 @@ func (app *Application) ExtendVote(_ context.Context, req *abci.RequestExtendVot
...
+	var ext []byte
+	var extLen int
+	if app.cfg.VoteExtensionSize != 0 {
+		ext = make([]byte, app.cfg.VoteExtensionSize)
+		if _, err := rand.Read(ext); err != nil {
+			panic(fmt.Errorf("could not extend vote. Len:%d", len(ext)))
+		}
+		extLen = len(ext)
+	} else {
+		ext = make([]byte, 8)
```
sergio-mena, line 462:
```suggestion
		ext = make([]byte, binary.MaxVarintLen64)
```

test/e2e/app/app.go
```diff
@@ -439,18 +446,28 @@ func (app *Application) ExtendVote(_ context.Context, req *abci.RequestExtendVot
...
+		if num, err := rand.Int(rand.Reader, big.NewInt(voteExtensionMaxVal)); err != nil {
+			panic(fmt.Errorf("could not extend vote. Len:%d", len(ext)))
+		} else {
+			extLen = binary.PutVarint(ext, num.Int64())
```
sergio-mena, line 466: If you're converting to int64, do you really need the `big` library in line 463?

More generally, wouldn't it be simpler to keep the old code inside this "else" (as you did in `VerifyVoteExtension`?

sergio-mena (commented):
Overall looks great. Some comments

test/e2e/app/app.go
```diff
@@ -472,9 +489,9 @@ func (app *Application) VerifyVoteExtension(_ context.Context, req *abci.Request
 		}, nil
 	}
 
-	num, err := parseVoteExtension(req.VoteExtension)
+	num, err := parseVoteExtension(app.cfg, req.VoteExtension)
 	if err != nil {
-		app.logger.Error("failed to parse vote extension", "vote_extension", req.VoteExtension, "err", err)
+		app.logger.Error("failed to parse vote extension", "vote_extension", fmt.Sprintf("%x", req.VoteExtension[:4]), "err", err)
```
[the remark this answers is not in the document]
lasarojc: Just to have some information here, a "hash" of the extension.

test/e2e/docker/entrypoint-delve
```diff
@@ -0,0 +1,12 @@
...
+# Forcibly remove any stray UNIX sockets left behind from previous runs
+rm -rf /var/run/privval.sock /var/run/app.sock
+
+# dlv won't run the app until you connect to it with a client.
+# Once the app is run, the signer will try only a few times before stopping, so don't take long to let commet run as well.
+dlv --headless --listen=:2345 --log --log-output=debugger,debuglineerr,gdbwire,lldbout,rpc --accept-multiclient --api-version=2 exec /usr/bin/app -- /cometbft/config/app.toml &
+
+sleep 30
+
+dlv --headless --listen=:2346 --log --log-output=debugger,debuglineerr,gdbwire,lldbout,rpc --accept-multiclient --api-version=2 exec /usr/bin/cometbft -- "$@"
```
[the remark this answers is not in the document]
lasarojc: You are right.
Added another entry point.

RELEASES.md
lasarojc: Updated the text. Pls have a look.

test/e2e/app/app.go
lasarojc: The change was because I updated the code to use `crypto/rand`, which returns a `big.Int`.
It was laborious way of getting rid of a lint :)

common.mk
lasarojc: Actually, we strip in the if, because the comparison is with an empty string, so in the else we do not strip and add the extra flags.
But given that this is the only case in which we compare with the empty string and that now we have an else, I will invert the two cases. Thanks for raising it.

RELEASES.md
```diff
@@ -345,6 +345,21 @@ to blocksync to the head of the chain and begins producing blocks using
...
 takes its place. This network is run for several days.
 
+#### Vote-extension Testnet
+
+CometBFT v0.38.0 introduced **vote-extensions**, which are added as the name suggests, to precommit votes sent by validators.
+The Vote-extension Testnet is used to determine how vote-extensions affect the performance of CometBFT, under various settings.
+The application used in the experiment is the same used on the (#200-node-testnet), but is configured differently to gauge de effects of varying vote extension sizes.
+In the (#200-node-testnet) the application extends pre-commit votes with a 64 bit number encoded with variable compression.
+In the Vote-extension Testnet, pre-commit votes are extended with a non-compressed extension of configurable size.
+Experiments are run with multiple sizes to determine their impact and, for comparison sake, we include a run with the same with the same settings as in the (#200-node-testnet).
```
sergio-mena, line 355:
```suggestion
Experiments are run with multiple sizes to determine their impact and, for comparison sake, we include a run with the same settings as in the (#200-node-testnet).
```

test/docker/config-template.toml
```diff
@@ -1,2 +1,10 @@
+vote_extensions_enable_height = 0
+check_tx_delay = "1000s"
+finalize_block_delay = "1000ms"
+prepare_proposal_delay = "1000ms"
+process_proposal_delay = "1000ms"
+vote_extension_delay = "20ms"
+vote_extension_size = 8192
+
```
sergio-mena, line 8: [not 100% sure] These are parameters for the e2e app, but this file is used to build CometBFT Docker images without the e2e app.
Could you double-check where this file is used? (if the e2e app is not involved, please revert this hunk)

sergio-mena (approved):
LGTM

lasarojc:
Thanks for the review.
I had to add another commit. Could you review this last bit?
Let me explain what the change is for.  Because the VE are used to generate a new transaction, and the transaction changes the app state, the state never stops changing, even if there is no load in the system.
Hence app was updated so 

- the value stored in the kv-store does not include the VE hashes
- the appHash does not to include the height
- the sum of values in the VE always returns 0 when VE padding is enabled so after the load is removed the so the app state stops changing and the appHash stops changing and empty blocks eventually stop being proposed.

PR #861, opened 2023-05-19
Adds the Gossip CRDT 

lasarojc:
This PR adds the `query` functionality to all example CRDT.
Also adds the skeleton for the Gossip CRDT.

Some comments in #822 are addressed in this PR

---

#### PR checklist

- [ ] Tests written/updated


consensus/state.go
```diff
@@ -387,7 +387,7 @@ func (cs *State) OnStart() error {
 }
 
 // timeoutRoutine: receive requests for timeouts on tickChan and fire timeouts on tockChan
-// receiveRoutine: serializes processing of proposoals, block parts, votes; coordinates state transitions
+// receiveRoutine: serializes processing of proposals, block parts, votes; coordinates state transitions
```
lasarojc, line 390:
```suggestion
// receiveRoutine: serializes processing of proposoals, block parts, votes; coordinates state transitions
```

consensus/state.go
```diff
@@ -2001,7 +2001,7 @@ func (cs *State) handleCompleteProposal(blockHeight int64) {
 	if cs.Step <= cstypes.RoundStepPropose && cs.isProposalComplete() {
 		// Move onto the next step
 		cs.enterPrevote(blockHeight, cs.Round)
-		if hasTwoThirds { // this is optimisation as this will be triggered when prevote is added
+		if hasTwoThirds { // this is an optimization as this will be triggered when prevote is added
```
lasarojc, line 2004:
```suggestion
		if hasTwoThirds { // this is optimization as this will be triggered when prevote is added
```

faddat:
This is very interesting. Is it still on the roadmap?

melekes:
CRDT as a conflict-free replicated data type?

faddat:
@melekes yeah so, to expand on what I found interesting -- I had not thought of gossip this way, but after thinking about it some CRDT does seem to encapsulate the ideal of what we'd do with gossip -- both block gossip and tx/mempool gossip, even. 

It should also be able to avoid a great deal of the noise that we are currently dealing with in gossip (I think).

lasarojc:
We haven't allocated time for it but this is something I would very much like to see continued.

faddat:
Yeah, I guess my thoughts/comment on this concept is that it seems elegant and efficent.

lasarojc:
I am merging this into the base branch and opening a PR from there, so that the whole work is captured in a single place.

melekes (approved):
👍

sergio-mena:
👍 here is is :-)

PR #890, opened 2023-05-26
Prevent a transaction to appear twice in the mempool

otrack:
**Summary**
A transaction might appear twice in the mempool, i.e., it is present more than once in `txs`, while `txsMap` only links a single of these occurrences.  The bug happens when the cache overflows. 

**Context**
The bug is related to issues [#3865](https://github.com/tendermint/tendermint/issues/3865) and [#5281](https://github.com/tendermint/tendermint/issues/5281) in Tendermint. Both issues are closed with this fix. One may reproduce them easily in e2e, e.g.,

```
./build/generator -m v0.37.1:1 -d networks/generated/
./build/runner -f networks/generated/gen-0003.toml setup
find networks/generated/gen-0003/ -iname "config.toml" | xargs sed -i s,"cache_size = 10000","cache_size = 100", 
find networks/generated/gen-0003/ -iname "config.toml" | xargs sed -i s,"prepare_proposal_delay = 200ms","prepare_proposal_delay = 3000ms",
./build/runner -f networks/generated/gen-0003.toml start
curl -s http://localhost:5701/broadcast_tx_sync?tx=\"apple=100\"; for i in $(seq 1 1001); do curl -s http://localhost:5701/broadcast_tx_sync?tx=\"orange=${i}\"; done; curl -s http://localhost:5701/broadcast_tx_sync?tx=\"apple=100\"
```

**Remark**
The present fix works because the mempool receives _in order and sequentially_ all the asynchronous ABCI callbacks. Notice that in the past, this was not the case for the grpc client, as outlined [here](https://github.com/tendermint/tendermint/pull/5520). Sidestepping such a constraint is feasible, yet presently [open](https://github.com/tendermint/tendermint/issues/5519).

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -399,7 +399,7 @@ func (mem *CListMempool) resCbFirstTime(
 			// Check transaction not already in the mempool
 			if e, ok := mem.txsMap.Load(types.Tx(tx).Key()); ok {
 				memTx := e.(*clist.CElement).Value.(*mempoolTx)
-				memTx.senders.LoadOrStore(peerID, true)
+				memTx.addSender(txInfo.SenderID)
 				mem.logger.Debug(
 					"transaction already there, not adding it again",
 					"tx", types.Tx(tx).Hash(),
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -396,6 +396,18 @@ func (mem *CListMempool) resCbFirstTime(
 				return
 			}
 
+			// Check transaction not already in the mempool
+			if _, ok := mem.txsMap.Load(types.Tx(tx).Key()); ok {
+				mem.logger.Debug(
+					"transaction already there",
+					"tx", types.Tx(tx).Hash(),
+					"res", r,
+					"height", mem.height,
+					"total", mem.Size(),
+				)
+				return
+			}
+
 			memTx := &mempoolTx{
 				height:    mem.height,
 				gasWanted: r.CheckTx.GasWanted,
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -397,7 +397,9 @@ func (mem *CListMempool) resCbFirstTime(
 			}
 
 			// Check transaction not already in the mempool
-			if _, ok := mem.txsMap.Load(types.Tx(tx).Key()); ok {
+			if e, ok := mem.txsMap.Load(types.Tx(tx).Key()); ok {
+				memTx := e.(*clist.CElement).Value.(*mempoolTx)
+				memTx.senders.LoadOrStore(peerID, true)
 				mem.logger.Debug(
 					"transaction already there, not adding it again",
 					"tx", types.Tx(tx).Hash(),
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -399,7 +399,7 @@ func (mem *CListMempool) resCbFirstTime(
 			// Check transaction not already in the mempool
 			if _, ok := mem.txsMap.Load(types.Tx(tx).Key()); ok {
 				mem.logger.Debug(
-					"transaction already there",
+					"transaction already there, not adding it again",
 					"tx", types.Tx(tx).Hash(),
 					"res", r,
 					"height", mem.height,
--- a/mempool/clist_mempool_test.go
+++ b/mempool/clist_mempool_test.go
@@ -632,6 +632,51 @@ func TestMempoolTxsBytes(t *testing.T) {
 	assert.EqualValues(t, 10, mp.SizeBytes())
 }
 
+func TestMempoolNoCacheOverflow(t *testing.T) {
+	sockPath := fmt.Sprintf("unix:///tmp/echo_%v.sock", cmtrand.Str(6))
+	app := kvstore.NewInMemoryApplication()
+	_, server := newRemoteApp(t, sockPath, app)
+	t.Cleanup(func() {
+		if err := server.Stop(); err != nil {
+			t.Error(err)
+		}
+	})
+	cfg := test.ResetTestRoot("mempool_test")
+	mp, cleanup := newMempoolWithAppAndConfig(proxy.NewRemoteClientCreator(sockPath, "socket", true), cfg)
+	defer cleanup()
+
+	// add tx0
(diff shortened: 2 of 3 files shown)


mempool/clist_mempool_test.go
```diff
@@ -632,6 +632,50 @@ func TestMempoolTxsBytes(t *testing.T) {
...
+	})
+	cfg := test.ResetTestRoot("mempool_test")
+	mp, cleanup := newMempoolWithAppAndConfig(proxy.NewRemoteClientCreator(sockPath, "socket", true), cfg)
+	defer cleanup()
+
+	// add tx0
+	var tx0 = kvstore.NewTxFromID(0)
+	err := mp.CheckTx(tx0, nil, TxInfo{})
+	require.NoError(t, err)
+	mp.FlushAppConn()
```
cason: You probably have to check the return of this call.

mempool/clist_mempool.go
```diff
@@ -389,6 +389,18 @@ func (mem *CListMempool) resCbFirstTime(
 				return
 			}
 
+			// Check transaction not already in the mempool
```
cason, line 392: It would be helpful to explain why would the transaction be already in the mempool. Maybe, once we have an issue detailing the bug, add the link to the issue here.

mempool/clist_mempool.go
```diff
@@ -389,6 +389,18 @@ func (mem *CListMempool) resCbFirstTime(
...
+			if _, ok := mem.txsMap.Load(types.Tx(tx).Key()); ok {
+				mem.logger.Debug(
+					"transaction already there",
```
cason, line 395:
```suggestion
					"transaction already there, not adding it again",
```

cason (approved):
Great finding :)

We definitely need a specification for this interaction that takes into consideration these corner cases (e.g., cache eviction, out-of-order ABCI responses, etc.).

mempool/clist_mempool.go
```diff
@@ -389,6 +389,18 @@ func (mem *CListMempool) resCbFirstTime(
 				return
 			}
 
+			// Check transaction not already in the mempool
+			if _, ok := mem.txsMap.Load(types.Tx(tx).Key()); ok {
```
hvanz, line 393: What would happen if the new request was checked at a different height, or it wants different amount of gas, or the sender is not the same as recorded in memTx? I think these values need to be updated in the mempool.

cason: Notice that this code is called as the callback for `CheckTx` and it is intended to add a new transaction to the mempool. It is not called by `CheckTx` in recheck mode, for instance, when the txs should be already in the mempool.

cason: PLus, updating the sender is handler elsewhere, namely here: https://github.com/cometbft/cometbft/blob/5f4a27eadb140825b260e8ddf22b5c3fb1820007/mempool/clist_mempool.go#L243

hvanz: You're right that the sender is recorded before sending the CheckTx request, so there's no need to update it here.

hvanz: But what about the height or the gas? 
Could it be that the second request was requested at a different height? In this case it would be better to update the height in `mempoolTx`. 
Is the gas the same even when rechecked at the same height? That is, should the gas calculation be deterministic?

jmalicevic: The senders are added where @cason pointed if the tx is in cache. But what if it is not in cache, because the cache overflows and thus this update  never happens? Then we need to retrieve the tx from the mempool at this point and update the senders here?

About the other concerns of @hvanz , I am not 100% sure how to handle it, but we definitely should either update it when retrieving the tx to update the senders, or ignore those values. But whichever we do, we need to document this somewhere so users are aware.

jmalicevic: Yes, I agree with you. But the  senders in particular are something we should update because this will prevent the node to gossip the transaction back to the sender. The mempool remembers all senders of a transaction (duplicate or not) . If the cache did not overflow, the sender would be added in the code Daniel posted. If it is not in the cache it would be added below the fix. RIght now , we should somehow add the sender within the fix as well (my opinion).

otrack: Regarding the problem of not tracking the sender of a transaction properly, this is already the case in the initial codebase: if a transaction is received from some peer `p` before `resCbFirstTime` executes, the information about `p` is lost.

cason: After some discussion with Jasmina, I now agree that we should add senders to the existing entry here, as we do when the cache prevents us from checking a transaction again.

hvanz: I agree with updating the senders. In the current codebase, the sender is actually tracked when it is received either from a peer or via RPC, and before `resCbFirstTime`: https://github.com/cometbft/cometbft/blob/main/mempool/clist_mempool.go#L261

otrack: Alright, I did so.

sergio-mena (approved):
Excellent find!
I guess we need to backport to all active branches. Let me know if you need assistance on that

sergio-mena:
Could you please add a changelog entry?
You get inspiration from #383 on how to do it

otrack:
> Could you please add a changelog entry? You can get inspiration from #383 on how to do it

Done.

otrack:
> Could you please add a changelog entry? You can get inspiration from #383 on how to do it

I guess so. What should the backport tagert? (e.g., v0.34.x to v0.38x)

cason:
> > Could you please add a changelog entry? You can get inspiration from #383 on how to do it
> 
> I guess so. What should the backport tagert? (e.g., v0.34.x to v0.38x)

Yes. I am just wondering whether we can just use the bot to do so, as the organization of files has changed a lot from [v0.37.x](https://github.com/cometbft/cometbft/tree/v0.37.x/mempool) and [v0.34.x](https://github.com/cometbft/cometbft/tree/v0.34.x/mempool).

mzabaluev:
There are some newly appeared lints, but they are not caused by any changed lines, so should be addressed in a different PR.

cason:
> I rebased it and added the appropriate backport tags.

I would say to go ahead, squash and merge it. 

I would only remove this `FIX` from the commit message, also from the PR name.

PR #895, opened 2023-05-30
mempool: Handle concurrent requests in recheck callback

hvanz:
Addresses https://github.com/tendermint/tendermint/issues/5519

Also, this PR simplifies the implementation of the recheck callback by eliminating the global pointers `recheckCursor` and `recheckEnd`, which were used to traverse the list of transactions while processing recheck requests. These variables were also used in other parts of the code for deciding if there were still unprocessed recheck requests. This PR replaces them by the existing `txsMap`, which maps tx keys to `txs` entries, and by a new counter variable to keep track of how many requests are still to be processed.

Before it was assumed that re-CheckTx requests were processed and handled sequentially, but this is not true for gRPC ABCI clients (see comments in https://github.com/tendermint/tendermint/issues/5519).

Built on top of https://github.com/cometbft/cometbft/pull/894

---

#### PR checklist

- [ ] Tests written/updated


otrack:
Handling in order ABCI calls at the client is mandatory. If not then a `FlushAppConn` operation may return earlier than a prior `resCbFirstTime` or `resCbRecheck` call. It follows that `Update` may concurrently execute with such a (re-)check operation. This breaks some base invariants in `clist_mempool.go` and/or may raise exceptions (e.g., removing twice the same element in `CList`.)

In my view, the gRPC ABCI client does provide the desired in-order semantics. 

I would not add more concurrency to `CListMempool`, because it seems that there is no need for it performance wise.
In fact, I would do the opposite, that is additional locking, typically in the callbacks. In its current state, this class works because it relies heavily on (partly missing) assumptions wrt. consensus and the client application.

Nevertheless, the idea of removing the cursor is I believe a nice one because it does simplify the codebase :)

hvanz:
It seems that the ABCI Flush method is not relevant at all (see https://github.com/tendermint/tendermint/issues/6994). While simplifying the client interface for v0.36, it was proposed to be removed (see comments in https://github.com/tendermint/tendermint/pull/7607), which finally never happened. Even the SDK code has [a comment](https://github.com/CometBFT/cometbft/blob/4790ea3e46475064d5475c787427ae926c5a9e94/abci/client/client.go#L31) saying that this method should be removed as it is not implemented.

hvanz:
> I feel that we shouldn't backport this to `v0.38.x`:

You're right. I removed the label.

hvanz:
> * the "official" reason for having the `Flush` ABCI call is in [this section](https://github.com/cometbft/cometbft/blob/main/spec/abci/abci%2B%2B_app_requirements.md#commit) of the spec.

This is what the spec says:

> Before invoking Commit, CometBFT locks the mempool and flushes the mempool connection. This ensures that no new messages will be received on the mempool connection during this processing step, providing an opportunity to safely update all four connection states to the latest committed state at the same time.

I would say that only the locking part is needed to ensure that no new messages will be received. Is this correct?

otrack:
A bad interleaving that explains why we need `flush` (at least for the moment) is [here](https://github.com/cometbft/cometbft/compare/hernan/mempool-concurrent-rechecks...otrack:cometbft:hernan/mempool-concurrent-rechecks).

otrack:
Just a few more thoughts about improving the parallelism by allowing concurrent checkTx.

It seems that there is no easy way to do this. The knot of the problem is [this](https://github.com/cometbft/cometbft/blob/73a6c58259f60979fd8cd63a0c560d6f9f42a2cd/mempool/clist_mempool.go#L269) strange pattern in the mempool where one passes a callback to another callback. 

Maybe the best approach would be to have a re-entrant read/write lock for the `CListMempool`. Another idea is to ask `AppConnMempool::CheckTxAsync` to take as a parameter the callback to invoke once the application answers. 

All in all, I am not sure this is a priority, given the [speed](https://github.com/cometbft/cometbft/pull/934) of the current implementation: around 60us to check a Tx with the remote client.

sergio-mena:
> I would say that only the locking part is needed to ensure that no new messages will be received. Is this correct?

This has never been thoroughly discussed. Maybe we can include it in today's meeting?

mempool/clist_mempool.go
```diff
@@ -674,3 +650,20 @@ func (mem *CListMempool) recheckTxs() {
 	// all pending messages to the app. There doesn't seem to be any need here as the buffer
 	// will get flushed regularly or when filled.
 }
+
+func (mem *CListMempool) noRecheckingRequests() bool {
+	return atomic.LoadUint32(&mem.numRecheckingRequests) == 0
+}
+
+func (mem *CListMempool) increaseNumRecheckingRequests() uint32 {
+	mem.logger.Info("increaseNumRecheckingRequests")
```
sergio-mena, line 659: Can we make this log more readable?

hvanz: This log was not supposed to be here. Removed.

melekes:
@hvanz very interesting 👍 any plans to resume this?

hvanz:
Closing as it is currently not an option to break the FIFO ordering when checking transactions.

Issue #26, opened 2022-12-23
Pinpoint inefficiencies in block, vote and transaction propagation


lasarojc:
Was https://github.com/tendermint/tendermint/issues/9922

It has been identified that votes, block parts and transaction propagation in the mempool use more data than what should be needed to reach a decision (https://github.com/tendermint/tendermint/pull/9706)

In this task we determine by which factor this inefficiencies happen on each kind of message (which will let us prioritize optimization of each of them) and what are the sources of the inefficiencies (which will point the direction to the fixes).

Some questions to be answered

- do nodes forget having send messages and send them again?
   - they don't forget having sent (except for a case that has already been fixed), but sometimes require getting the message from the other node back, before stopping sending it.
- do all nodes needlessly send the same messages to the same nodes?
   - given the unstructured nature of the network, yes, the same message is received multiple times from multiple sources.
- is the "has votes" message effective?
   - yes, as long as it is delivered before the votes itself, which is not normally the case. This has been addressed in #904. 

Tasks:
- [x]  Add metrics to track how many times a node receives duplicate votes (#896)
- [x]  Add metric to track how many times a node receives duplicate block parts (#896)
- [x]  Add metric to track how many times a node receives duplicate transactions (present in mempool cache) (#637)
- [x]  Compile the results (see the discussion in #904)
- [ ]  Add logs to identify the sources of duplicate votes (metrics and code analysis were enough)
- [ ]  Add logs to identify the sources of duplicate block parts (metrics and code analysis were enough)
- [ ]  Add logs to identify the sources of duplicate duplicate transactions (present in mempool cache) (metrics and code analysis were enough)

DoD:

- [x] We identified why duplication happens.
- [x] This information serves as input to optimizing the message exchange: https://github.com/cometbft/cometbft/issues/30

PR #896, opened 2023-05-31
New metrics to track duplicate votes and block parts 

ebuchman:
Related to https://github.com/cometbft/cometbft/issues/26

Add metrics to track duplicate votes and block parts in preparation for changes to reduce bandwidth :) 

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -1940,6 +1940,11 @@ func (cs *State) addProposalBlockPart(msg *BlockPartMessage, peerID p2p.ID) (add
 	}
 
 	cs.metrics.BlockGossipPartsReceived.With("matches_current", "true").Add(1)
+	if !added {
+		// NOTE: we are disregarding possible duplicates above where heights dont match or we're not expecting block parts yet
+		// but between the matches_current = true and false, we have all the info.
+		cs.metrics.DuplicateBlockPart.Add(1)
+	}
 
 	if cs.ProposalBlockParts.ByteSize() > cs.state.ConsensusParams.Block.MaxBytes {
 		return added, fmt.Errorf("total size of proposal block parts exceeds maximum block bytes (%d > %d)",
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -2013,6 +2013,8 @@ func (cs *State) handleCompleteProposal(blockHeight int64) {
 // Attempt to add the vote. if its a duplicate signature, dupeout the validator
 func (cs *State) tryAddVote(vote *types.Vote, peerID p2p.ID) (bool, error) {
 	added, err := cs.addVote(vote, peerID)
+
+	// NOTE: some of these errors are swallowed here
 	if err != nil {
 		// If the vote height is off, we'll just ignore it,
 		// But if it's a conflicting sig, add it to the cs.evpool.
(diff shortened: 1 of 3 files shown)


lasarojc:
You missed running `make metrics` to update `metrics.gen.go`

types/part_set.go
```diff
@@ -270,9 +270,12 @@ func (ps *PartSet) Total() uint32 {
 }
 
 func (ps *PartSet) AddPart(part *Part) (bool, error) {
+	// TODO: remove this? would be preferable if this only returned (false, nil)
```
cason, line 273: Essentially all `PartSet` methods handle this condition in which the block set is not initialized. This might make the consensus logic simpler, but it is not really easy to read.

types/part_set.go
```diff
@@ -270,9 +270,12 @@ func (ps *PartSet) Total() uint32 {
...
+	// when its a duplicate block part
 	if ps == nil {
 		return false, nil
 	}
+
```
cason, line 278:
```suggestion

```

cason (approved):
Nice to have. I've left some minor comments that shouldn't affect the feature.

My more general comment is that both metrics are more helpful when they reflect the portion of votes or block parts that are duplicated. Of course, the operator can "plot" the ratio of different metrics to get this information, but I wonder whether a different metric abstraction would be more helpful.

types/part_set.go
```diff
@@ -270,9 +270,12 @@ func (ps *PartSet) Total() uint32 {
 }
 
 func (ps *PartSet) AddPart(part *Part) (bool, error) {
+	// TODO: remove this? would be preferable if this only returned (false, nil)
+	// when its a duplicate block part
 	if ps == nil {
 		return false, nil
```
sergio-mena, line 276: I checked the code and this function is only called from `State`'s `addProposalBlockPart` function (ignoring test code here).
If you look at `addProposalBlockPart`'s code immediately following the call to `AddPart`, you see that it is acting as though the part had indeed been added. So, returning `false, nil` in this line is a BUG. We must return `false` and an error, such as `fmt.Errorf("nil PartSet")`

ebuchman:
> My more general comment is that both metrics are more helpful when they reflect the portion of votes or block parts that are duplicated. Of course, the operator can "plot" the ratio of different metrics to get this information, but I wonder whether a different metric abstraction would be more helpful.

@cason what did you have in mind? my understanding is we should keep the metrics simple/raw since it's always easy to manipulate the raw data

consensus/state.go
```diff
@@ -1940,6 +1940,11 @@ func (cs *State) addProposalBlockPart(msg *BlockPartMessage, peerID p2p.ID) (add
 	}
 
 	cs.metrics.BlockGossipPartsReceived.With("matches_current", "true").Add(1)
+	if !added {
+		// NOTE: we are disregarding possible duplicates above where heights dont match or we're not expecting block parts yet
```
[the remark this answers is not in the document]
ebuchman: will update comment

types/part_set.go
ebuchman: I believe AddPart is only called in one place in line 1934 in consensus/state and we check right before if the parts are nil, so it should be totally redundant to have the check in here, but we thought it might break tests or something

types/part_set.go
ebuchman: not sure I understand this, my argument was that this check is redundant since before we call AddPart we already check if the part set is nil!

sergio-mena:
If we can get the data we want via prometheus queries (by using functions and/or combining basic metrics), I am in favour of keeping the ones in the code (which can't be changed easily) as simple as possible

cason:
So, I was thinking about a metric keeping a ratio, like for instance the load metric for the CPU. But this is a nice to have, not an actual requirement, in particular because I don't know whether Prometheus actually supports this behavior.

PR #900, opened 2023-06-01
Fixed Unintended leak of Proxy-Authorization header in requests

wica-sufatmawati:
Since Requests v2.3.0, Requests has been vulnerable to potentially leaking Proxy-Authorization headers to destination servers, specifically during redirects to an HTTPS origin. This is a product of how rebuild_proxies is used to recompute and reattach the Proxy-Authorization header to requests when redirected. Note this behavior has only been observed to affect proxied requests when credentials are supplied in the URL user information component (e.g. https://username:password@proxy:8080).

**Current vulnerable behavior(s):**
 1. HTTP → HTTPS: leak
 1. HTTPS → HTTP: no leak
 1. HTTPS → HTTPS: leak
 1. HTTP → HTTP: no leak

For HTTP connections sent through the proxy, the proxy will identify the header in the request itself and remove it prior to forwarding to the destination server. However when sent over HTTPS, the Proxy-Authorization header must be sent in the CONNECT request as the proxy has no visibility into further tunneled requests. This results in Requests forwarding the header to the destination server unintentionally, allowing a malicious actor to potentially exfiltrate those credentials.

The reason this currently works for HTTPS connections in Requests is the Proxy-Authorization header is also handled by urllib3 with our usage of the ProxyManager in adapters.py with proxy_manager_for. This will compute the required proxy headers in proxy_headers and pass them to the Proxy Manager, avoiding attaching them directly to the Request object. This will be our preferred option going forward for default usage.

## Impact
CVE-2023-32681
CWE-200
CVSS:3.1/AV:N/AC:H/PR:N/UI:R/S:C/C:H/I:N/A:N

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/scripts/qa/reporting/requirements.txt
+++ b/scripts/qa/reporting/requirements.txt
@@ -11,4 +11,4 @@ python-dateutil==2.8.2
 six==1.16.0
 pandas==1.5.3
 prometheus-pandas==0.3.2
-requests==2.28.2
+requests==2.31.0


adizere (commented):
Thanks!

This change concerns code that is intended to run manually. I don't think the vulnerability described in the CVE applies. From what I can tell, this change is inconsequential.

We'll upgrade the requirements for `requests` the next time we do QA. cc @lasarojc re: your branch `lasaro/843-b`.

lasarojc:
Thanks for catching this!
Indeed it is used only for some testing, but I will update the requirements nonetheless.

Issue #353, opened 2023-02-17
Design of e2e tests that should verify CometBFT's behaviour regarding ABCI++. 


nenadmilosevic95:
# Problem definition
We want to enhance the e2e tests with the architecture that will enable us to test whether CometBFT respects the ABCI++ grammar. The idea is to run the testnet, log the ABCI calls, and then do the post-mortum testing of the logs. 

## DoD
- [x] ADR explaining the design choices: #902 
- [ ] Implementation: #930

PR #902, opened 2023-06-01
ADR 108: ADR for extending E2E infrastructure so that we can check CometBFT's behaviour with respect to ABCI++ grammar.

nenadmilosevic95:
---

Related to #353.

Adding an ADR explaining changes made so we can check whether CometBFT respects ABCI++ grammar. 

#### PR checklist

- [ ] Tests written/updated


docs/architecture/adr-110-e2e-abci++.md
```diff
@@ -0,0 +1,267 @@
+# ADR 82: E2E tests for CometBFT's behaviour in respect to ABCI++.
```
sergio-mena, line 1: Please check what we right number is (CometBFT started at 100) by looking at `docs/architecture/README.md`

sergio-mena: Also, rename the file accordingly

docs/architecture/adr-110-e2e-abci++.md
```diff
@@ -0,0 +1,267 @@
...
+    
+	...
+}
+```
+Notice here that we create an empty `abci.RequestInitChain` object while we can also use the one passed to the `InitChain` function. The reason behind this is that, at the moment, we do not need specific fields of the request; we just need to be able to extract the information about the request type. For this, an empty object of a particular type is enough. 
+
+`app.logRequest(r)` function is a new function implemented in the same file (`test/e2e/app/app.go`). Its implementation is the following: 
+
+```go
+func (app *Application) logRequest(req *abci.Request) {
```
sergio-mena, line 42: I'd call it logAbciRequest, but up to you

docs/architecture/adr-110-e2e-abci++.md
```diff
@@ -0,0 +1,267 @@
...
+	s, err := GetABCIRequestString(req)
+	if err != nil {
+		panic(err)
```
sergio-mena, line 45: Panicking here is OK. But you'd get simpler code by just returning the error, and propagating it in `InitChain`.
If an ABCI call returns non-`nil` error CometBFT will panic itself.

docs/architecture/adr-110-e2e-abci++.md
```diff
@@ -0,0 +1,267 @@
...
+
+`GetABCIRequestString(req)` is a new method that receives a request and returns its string representation. The implementation and tests for this function and the opposite function `GetABCIRequestFromString(req)`
+that returns `abci.Request` from the string are provided in files `test/e2e/app/log.go` and `test/e2e/app/log_test.go`, respectively. To create a string representation of a request, we first marshal the request via `proto.Marshal()` method and then convert received bytes in the string using `base64.StdEncoding.EncodeToString()` method. The code of this method is below: 
+
+```go
+func GetABCIRequestString(req *abci.Request) (string, error) {
+	b, err := proto.Marshal(req)
+	if err != nil {
+		return "", err
+	}
```
sergio-mena, line 59: As this will go into a log line, consider compressing it before converting to base64 (although not sure what the gains would be... maybe later on when we log more info)

docs/architecture/adr-110-e2e-abci++.md
```diff
@@ -0,0 +1,267 @@
...
+
+`app.logRequest(r)` function is a new function implemented in the same file (`test/e2e/app/app.go`). Its implementation is the following: 
+
+```go
+func (app *Application) logRequest(req *abci.Request) {
+	s, err := GetABCIRequestString(req)
+	if err != nil {
+		panic(err)
+	}
+	app.logger.Debug(s)
```
sergio-mena, line 47: Debug is an option, but, reflecting on it, wouldn't it be more convenient to turn them into "info", and have a configuration option in the e2e app that enables disables these logs? (it would just mean an extra if here).

The problem I see with debug is that the debug output is unbearable ATM if we enable it for all modules.

docs/architecture/adr-110-e2e-abci++.md
```diff
@@ -0,0 +1,267 @@
...
+
+ 
+
+### 3) ABCI++ grammar checker
+The idea here was to find a library that automatically verifies whether a specific execution respects the prescribed grammar. 
+
+<strong>Implementation</strong>
+
+We found the following library - https://github.com/goccmack/gogll. It generates a GLL or LR(1) parser and FSA-based lexer for any context-free grammar. What we needed to do is to rewrite ABCI++ grammar ([CometBFT's expected behaviour](../../spec/abci/abci%2B%2B_comet_expected_behavior.md#valid-method-call-sequences))
+using the synthax that the library understand.
```
sergio-mena, line 89:
```suggestion
using the synthax that the library understands. 
```

docs/architecture/adr-110-e2e-abci++.md
```diff
@@ -0,0 +1,267 @@
...
+### 3) ABCI++ grammar checker
+The idea here was to find a library that automatically verifies whether a specific execution respects the prescribed grammar. 
+
+<strong>Implementation</strong>
+
+We found the following library - https://github.com/goccmack/gogll. It generates a GLL or LR(1) parser and FSA-based lexer for any context-free grammar. What we needed to do is to rewrite ABCI++ grammar ([CometBFT's expected behaviour](../../spec/abci/abci%2B%2B_comet_expected_behavior.md#valid-method-call-sequences))
+using the synthax that the library understand. 
+The new grammar is below and can be found inside `test/e2e/pkg/grammar/abci_grammar.md` file.
+
+```abnf
```
sergio-mena, line 92: We've made small changes lately to the original grammar in the spec. Could you double-check this one is still aligned?

docs/architecture/adr-110-e2e-abci++.md
```diff
@@ -0,0 +1,267 @@
...
+specific execution (list of ABCI++ calls logged by the Application while the
+testnet was running) respects the ABCI++ grammar. The implementation and tests 
+for it are inside `test/e2e/pkg/grammar/checker.go` and 
+`test/e2e/pkg/grammar/checker_test.go`, respectively. 
+
+How the `GrammarChecker` works is demonstrated with the test `TestABCIGrammar`
+implemented in `test/e2e/tests/abci_test.go` file. 
+
+```go
+func TestABCIGrammar(t *testing.T) {
```
sergio-mena, line 155: Please add somewhere in this section a couple of sentences to clarify that the grammar represents **one node**, and so it applied to the logs from each node successively.

This can be gleaned from this code, but it's not at all obvious

docs/architecture/adr-110-e2e-abci++.md
```diff
@@ -0,0 +1,267 @@
...
+	m := fetchABCIRequestsByNodeName(t)
+	checker := grammar.NewGrammarChecker(grammar.DefaultConfig())
+	testNode(t, func(t *testing.T, node e2e.Node) {
+		reqs := m[node.Name]
+		_, err := checker.Verify(reqs)
+		if err != nil {
+			t.Error(err)
```
sergio-mena, line 162:
```suggestion
			t.Error(fmt.Errorf("ABCI grammar verification failed: %w", err))
```

docs/architecture/adr-110-e2e-abci++.md
sergio-mena: This could also be the same bool config to enable/disable this feature when running e2e tests

docs/architecture/adr-110-e2e-abci++.md
```diff
@@ -0,0 +1,267 @@
...
+}
+```
+
+Specifically, the test first fetches all ABCI++ requests and creates a `GrammarChecker` object. Then for each
+node in the testnet it checks if a specific set of requests respects the ABCI++ 
+grammar by calling `checker.Verify(reqs)` method. If this method returns an error, the specific execution does not respect the grammar. 
+
+The `Verify()` method is shown below. 
+```go
+
```
sergio-mena, line 174:
```suggestion
```

docs/architecture/adr-110-e2e-abci++.md
```diff
@@ -0,0 +1,267 @@
...
+	s, _ := g.GetExecutionString(r)
+	return g.VerifyExecution(s)
+}
+```
+
+It takes a list of requests and does the following things.
+- Filter the last height. Basically, it removes all ABCI++ requests after the 
+last `Commit`. Function `fetchABCIRequestsByNodeName()` can be called in the middle of the height. As a result, the last height may be incomplete, and 
+the parser may return an error. The simple example here is that the last 
+request fetched via `fetchABCIRequestsByNodeName()` is `PrepareProposal`; however, `ProcessProposal` happens after
```
sergio-mena, line 194: Careful with this example. We changed this recently: #1033

docs/architecture/adr-110-e2e-abci++.md
```diff
@@ -0,0 +1,267 @@
...
+last `Commit`. Function `fetchABCIRequestsByNodeName()` can be called in the middle of the height. As a result, the last height may be incomplete, and 
+the parser may return an error. The simple example here is that the last 
+request fetched via `fetchABCIRequestsByNodeName()` is `PrepareProposal`; however, `ProcessProposal` happens after 
+`fetchABCIRequestsByNodeName()` was invoked. Consequently, the parser
+will return an error that `ProcessProposal` is missing, even though the `ProcessProposal` 
+will happen after.  
+- Generates an execution string by replacing `abci.Request` with the 
+corresponding terminal from the grammar. This logic is implemented in
+`GetExecutionString()` function. This function receives a list of `abci.Request` and generates a string where every request the grammar covers 
+will be replaced with a corresponding terminal. For example, `abci.Request_PrepareProposal` is replaced with `<PrepareProposal>`. If the request is not covered
```
sergio-mena, line 201: Wouldn't it be simpler to declare the terminal in the grammar as `abci.Request_PrepareProposal` (with a comment on the grammar that this represents `<PrepareProposal>`. That would save you one layer of "translation"

docs/architecture/adr-110-e2e-abci++.md
```diff
@@ -0,0 +1,267 @@
...
+the parser may return an error. The simple example here is that the last 
+request fetched via `fetchABCIRequestsByNodeName()` is `PrepareProposal`; however, `ProcessProposal` happens after 
+`fetchABCIRequestsByNodeName()` was invoked. Consequently, the parser
+will return an error that `ProcessProposal` is missing, even though the `ProcessProposal` 
+will happen after.  
+- Generates an execution string by replacing `abci.Request` with the 
+corresponding terminal from the grammar. This logic is implemented in
+`GetExecutionString()` function. This function receives a list of `abci.Request` and generates a string where every request the grammar covers 
+will be replaced with a corresponding terminal. For example, `abci.Request_PrepareProposal` is replaced with `<PrepareProposal>`. If the request is not covered 
+by the grammar, it will be ignored.
```
sergio-mena, line 202: > If the request is not covered by the grammar, it will be ignored.

I guess this is because of `ExtendVote` and `VerifyVoteExtension`. Are you planning to reject any terminal not defined in the grammar once you upgrade the grammar (and the ADR, and the code) to support vote extensions?

docs/architecture/adr-110-e2e-abci++.md
```diff
@@ -0,0 +1,267 @@
...
+moment, we are printing the last 10 errors; however, this is part of the configuration and can be changed. 
+
+### Suporting additional ABCI++ requests
+
+Here we present all the steps we need to do if we want to support other 
+ABCI++ requests in the future: 
+
+- The application needs to log the new request in the same way as we do now.
+- We should include the new request to the grammar and generate a new parser and lexer.  
+- We should modify `getRequestTerminal()` method inside `test/e2e/pkg/grammar/checker.go` to return a grammar terminal for the new ABCI++ request.
```
sergio-mena, line 251: 👍

docs/architecture/adr-110-e2e-abci++.md
```diff
@@ -0,0 +1,267 @@
...
+- We should include the new request to the grammar and generate a new parser and lexer.  
+- We should modify `getRequestTerminal()` method inside `test/e2e/pkg/grammar/checker.go` to return a grammar terminal for the new ABCI++ request. 
+
+
+## Status
+
+Implemented.
+
+To-do list:
+- integrating the generation of parser/lexer into the codebase.
```
sergio-mena, line 259: How are you planning to do this? (I think it should be explained in this ADR)

My advice is to be as similar as possible to other generated code we have (e.g., protos):
* you have "make proto" to generate the code (so you could have "make grammar")
* the autogenerated code is committed to the repo (so people don't need to do "make grammar" if they don't change it)
* there is a CI workflow that checks for every PR that the author didn't forget to run "make proto" (so we could think of something similar here)

sergio-mena:
Once you add the part on managing the auto generated code, please change the status of the PR to "Ready for review". Thanks

docs/architecture/adr-110-e2e-abci++.md
```diff
@@ -0,0 +1,276 @@
...
+Issue: [353](https://github.com/cometbft/cometbft/issues/353).
+
+
+## Decision
+
+### 1) ABCI++ requests logging
+The idea was to do this at the Application side. Every time the Application 
+receives a request, it logs it.
+
+<strong>Implementation</strong>
```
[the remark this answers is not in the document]
nenadmilosevic95: Ok, changed it!

docs/architecture/adr-110-e2e-abci++.md
nenadmilosevic95: I am not sure I fully understand how would you do this. Would you immediately check this after the `logReguest` call in `InitChain` and return if there is an error?

docs/architecture/adr-110-e2e-abci++.md
nenadmilosevic95: At the moment, this will not make much of a difference since objects are empty and hence small. However, I can do it if you think we should do it anyway for the future.

docs/architecture/adr-110-e2e-abci++.md
nenadmilosevic95: I have no specific reasons why I am using `Debug` here, I think we discussed it a long time ago, and we considered it to be the `Debug`.  Anyway, I will do what you suggest here.

docs/architecture/adr-110-e2e-abci++.md
nenadmilosevic95: Actually, you changed that now in case of a round with the `proposer` `ProcessProposal` is optional. From my understanding, you did this because of the process crashing scenario. Did you apply the same reasoning for the whole grammar?

docs/architecture/adr-110-e2e-abci++.md
nenadmilosevic95: I will do it!

nenadmilosevic95: I added sentences about this.

docs/architecture/adr-110-e2e-abci++.md
nenadmilosevic95: Yes, for example, we will ignore the `Info` abci request.

docs/architecture/adr-110-e2e-abci++.md
nenadmilosevic95: I did some experiments to evaluate how big is a single log so we can make a decision around this. And I think at the moment there is not much point in compressing since the size of a single log containing abci.Request is at most 24B. 
However, if we decide to log the full request instead of the empty one then maybe it would make sense to do the compressing since when I changed the logging to log the full request I observed that individual request took up to 7KB.

docs/architecture/adr-110-e2e-abci++.md
sergio-mena: I think so

docs/architecture/adr-107-e2e-abci++.md
```diff
@@ -62,21 +69,22 @@ func GetABCIRequestString(req *abci.Request) (string, error) {
...
+- Takes the output of a specific node in the testnet from the moment we launched the testnet until the function is called. The node name is passed as a function parameter. It uses the `docker-compose logs` and `grep nodeName` commands. 
+- Parses the logs line by line and extracts the  `abci.Request`, if one exists. The request is received by forwarding each line to the `app.GetABCIRequestFromString(req)` method.
+- Returns the array of slices where each slice contains the set of `abci.Request`s logged on that node. Every time the crash happens, a new array element (new slice `[]*abci.Request`) will be created. We know a crash has happened because we log "Application started" every time the application starts. Specifically, we added this log inside `NewApplication()` function in `test/e2e/app/app.go` file. In the end, the function will return just one slice if the node did not experience any crashes and $n+1$ slices if there were crashes, $n$ being the number of crashes. The benefit of logging the requests in the previously described way is that now we can use `[]*abci.Request` to store ABCI++ requests of any type.
```
sergio-mena:
```suggestion
- Returns the array of slices where each slice contains the set of `abci.Request`s logged on that node. Every time a crash happens, a new array element (new slice `[]*abci.Request`) will be created. We know a crash has happened because we log "Application started" every time the application starts. Specifically, we added this log inside `NewApplication()` function in `test/e2e/app/app.go` file. In the end, the function will return just one slice if the node did not experience any crashes and $n+1$ slices if there were crashes, $n$ being the number of crashes. The benefit of logging the requests in the previously described way is that now we can use `[]*abci.Request` to store ABCI++ requests of any type.
```

docs/architecture/adr-107-e2e-abci++.md
```diff
@@ -153,53 +162,65 @@ implemented in `test/e2e/tests/abci_test.go` file.
...
 	testNode(t, func(t *testing.T, node e2e.Node) {
-		reqs := m[node.Name]
-		_, err := checker.Verify(reqs)
+		if !node.Testnet.ABCITestsEnabled {
+			return
+		}
+		reqs, err := fetchABCIRequests(t, node.Name)
 		if err != nil {
-			t.Error(err)
+			t.Error(fmt.Errorf("Collecting of ABCI requests failed: %w", err))
```
sergio-mena:
```suggestion
			t.Error(fmt.Errorf("collecting of ABCI requests failed: %w", err))
```

docs/architecture/adr-107-e2e-abci++.md
```diff
@@ -153,53 +162,65 @@ implemented in `test/e2e/tests/abci_test.go` file.
...
-	var r []*abci.Request
-	var n int
-	if g.cfg.FilterLastHeight {
-		r, n = g.filterLastHeight(reqs)
-		if n != 0 {
-			debugMsg := fmt.Sprintf("Last height filtered, removed last %v abci calls out of %v.\n", n, len(reqs))
-			g.logger.Debug(debugMsg)
-		}
+func (g *GrammarChecker) Verify(reqs []*abci.Request, isCleanStart bool) (bool, error) {
+	r := g.filterRequests(reqs)
```
sergio-mena: Here, you are using `reqs` and validating its contents just after. Shouldn't you move this line after 196? (so that you first validate `reqs` and then use it)

sergio-mena: .. or should line 194 use `r` rather than `reqs`?

docs/architecture/adr-107-e2e-abci++.md
```diff
@@ -153,53 +162,65 @@ implemented in `test/e2e/tests/abci_test.go` file.
...
-last `Commit`. Function `fetchABCIRequestsByNodeName()` can be called in the middle of the height. As a result, the last height may be incomplete, and 
-the parser may return an error. The simple example here is that the last 
-request fetched via `fetchABCIRequestsByNodeName()` is `PrepareProposal`; however, `ProcessProposal` happens after 
-`fetchABCIRequestsByNodeName()` was invoked. Consequently, the parser
-will return an error that `ProcessProposal` is missing, even though the `ProcessProposal` 
-will happen after.  
+The method `Verify()` first calls method `filterRequests()` that is going to remove all the requests from the set that are not supported by the current version of the grammar. In addition, it will filter the last height by removing all ABCI++ requests after the 
+last `Commit`. The function `fetchABCIRequests()` can be called in the middle of the height. As a result, the last height may be incomplete, and 
+classified as invalid even if that is not the reality. The simple example here is that the last 
+request fetched via `fetchABCIRequests()` is `Decide`; however, `Commit` happens after
```
sergio-mena: `Decide`, did you mean `EndBlock` (or `FinalizeBlock` in `v0.38.x`) ?

docs/architecture/adr-107-e2e-abci++.md
```diff
@@ -153,53 +162,65 @@ implemented in `test/e2e/tests/abci_test.go` file.
...
-will happen after.  
+The method `Verify()` first calls method `filterRequests()` that is going to remove all the requests from the set that are not supported by the current version of the grammar. In addition, it will filter the last height by removing all ABCI++ requests after the 
+last `Commit`. The function `fetchABCIRequests()` can be called in the middle of the height. As a result, the last height may be incomplete, and 
+classified as invalid even if that is not the reality. The simple example here is that the last 
+request fetched via `fetchABCIRequests()` is `Decide`; however, `Commit` happens after 
+`fetchABCIRequests()` was invoked. Consequently, the execution
+will be considered as faulty because `Commit` is missing, even though the `Commit` 
+will happen after. 
+After filtering the requests `Verify()` checks if the remaining set of requests respect the grammar. It does that by calling two methods: `verifySpecific()` and `verifyGeneric()`. 
+Former should always be called first and is responsible of doing some specific checks that the parser cannot do. For example, at the moment it is checking if
```
sergio-mena: Incomplete sentence?

docs/architecture/adr-108-e2e-abci++.md
```diff
@@ -143,15 +141,22 @@ it totally from the new grammar. The Application is still logging the `Info`
 call, but a specific test would need to be written to check whether it happens
 in the right moment. 
 
-The `gogll` library receives the file with the grammar as input, and it generates the corresponding parser and lexer. The actual command is integrated into `test/e2e/Makefile` and is executed when `make grammar` is invoked. 
+Moreover, both grammars, the original and the new, represent the node's expected behaviour from the fresh beginning (`CleanStart`) or after the crash (`Recovery`).
```
sergio-mena, line 144:
```suggestion
Moreover, both grammars, the original and the new, represent the node's expected behaviour from the fresh beginning (`CleanStart`) or after a crash (`Recovery`).
```

docs/architecture/adr-108-e2e-abci++.md
```diff
@@ -194,85 +209,56 @@ func (g *GrammarChecker) Verify(reqs []*abci.Request, isCleanStart bool) (bool,
...
-            5: process_proposal finalize_block commit
 			...
 ```
-The parse error shown above represents an error that happened at height 0. The `ABCI requests` part represents 
-requests observed in this height, while `Unexpected request` and `Expected one of` represent the request which 
-should not happen, and the requests that should have happened instead, respectively. 
-Lastly, after the errors the full execution, one height per line, is printed. This can be turned off with the config flag. 
-Notice here that the parser can return many errors because the parser returns an error at every point at which the parser fails to parse
+The error shown above reports an invalid `CleanStart` execution. Moreover, it says why it is considered invalid (`init_chain` was missing) and the height of the error. Notice here that the height in the case of `CleanStart` execution corresponds to the actual consensus height, while for the `Recovery` execution, height 0 represents the first height after the crash. Lastly, after the error, the full execution, one height per line, is printed. This part may be optional and handled with the config flag, but we left it like this for now.
```
sergio-mena, line 253:
```suggestion
The error shown above reports an invalid `CleanStart` execution. Moreover, it says why it is considered invalid (`init_chain` was missing) and the height of the error. Notice here that the height in the case of `CleanStart` execution corresponds to the actual consensus height, while for the `Recovery` execution, height 0 represents the first height after the crash. Lastly, after the error, the full execution, one height per line, is printed. This part may be optional and handled with a config flag, but we left it like this for now. 
```

docs/architecture/adr-108-e2e-abci++.md
```diff
@@ -0,0 +1,296 @@
...
+```go
+func TestABCIGrammar(t *testing.T) {
+	checker := grammar.NewGrammarChecker(grammar.DefaultConfig())
+	testNode(t, func(t *testing.T, node e2e.Node) {
+		if !node.Testnet.ABCITestsEnabled {
+			return
+		}
+		reqs, err := fetchABCIRequests(t, node.Name)
+		if err != nil {
+			t.Error(fmt.Errorf("Collecting of ABCI requests failed: %w", err))
```
sergio-mena, line 177:
```suggestion
			t.Error(fmt.Errorf("collecting of ABCI requests failed: %w", err))
```

docs/architecture/adr-108-e2e-abci++.md
```diff
@@ -0,0 +1,296 @@
...
+true` in the manifest file of a particular testnet (e.g. `networks/ci.toml`). This will automatically activate logging on the 
+application side. 
+
+The `Verify()` method is shown below. 
+```go
+func (g *GrammarChecker) Verify(reqs []*abci.Request, isCleanStart bool) (bool, error) {
+	r := g.filterRequests(reqs)
+	// This should not happen in our tests.
+	if len(reqs) == 0 {
+		return false, fmt.Errorf("Execution with no ABCI calls.")
```
sergio-mena, line 210:
```suggestion
		return false, fmt.Errorf("execution with no ABCI calls.")
```

docs/architecture/adr-108-e2e-abci++.md
```diff
@@ -0,0 +1,296 @@
...
+	if err != nil {
+		return "", err
+	}
+	reqStr := base64.StdEncoding.EncodeToString(b)
+	s := ABCI_REQ + reqStr + ABCI_REQ
+	return s, nil
+}
+```
+
+*Note:* At the moment, we are not compressing the marshalled request before converting it to `base64` `string` because we are logging the empty requests that take at most 24B. However, if we decide to log the actual requests in the future, we might want to compress them. Based on a few tests, we observed that the size of a request can go up to 7KB.
```
sergio-mena, line 73:
```suggestion
*Note:* At the moment, we are not compressing the marshalled request before converting it to `base64` `string` because we are logging the empty requests that take at most 24 bytes. However, if we decide to log the actual requests in the future, we might want to compress them. Based on a few tests, we observed that the size of a request can go up to 7KB.  
```

docs/architecture/adr-108-e2e-abci++.md
```diff
@@ -0,0 +1,296 @@
...
+- Takes the output of a specific node in the testnet from the moment we launched the testnet until the function is called. The node name is passed as a function parameter. It uses the `docker-compose logs` and `grep nodeName` commands. 
+- Parses the logs line by line and extracts the  `abci.Request`, if one exists. The request is received by forwarding each line to the `app.GetABCIRequestFromString(req)` method.
+- Returns the array of slices where each slice contains the set of `abci.Request`s logged on that node. Every time the crash happens, a new array element (new slice `[]*abci.Request`) will be created. We know a crash has happened because we log "Application started" every time the application starts. Specifically, we added this log inside `NewApplication()` function in `test/e2e/app/app.go` file. In the end, `fetchABCIRequests()` will return just one slice if the node did not experience any crashes and $n+1$ slices if there were crashes, $n$ being the number of crashes. The benefit of logging the requests in the previously described way is that now we can use `[]*abci.Request` to store ABCI++ requests of any type.
```
sergio-mena, line 87:
```suggestion
- Returns the array of slices where each slice contains the set of `abci.Request`s logged on that node. Every time a crash happens, a new array element (new slice `[]*abci.Request`) will be created. We know a crash has happened because we log "Application started" every time the application starts. Specifically, we added this log inside `NewApplication()` function in `test/e2e/app/app.go` file. In the end, `fetchABCIRequests()` will return just one slice if the node did not experience any crashes and $n+1$ slices if there were crashes, $n$ being the number of crashes. The benefit of logging the requests in the previously described way is that now we can use `[]*abci.Request` to store ABCI++ requests of any type.
```

docs/architecture/adr-108-e2e-abci++.md
```diff
@@ -0,0 +1,296 @@
...
+
+The tests are executed only if `ABCITestsEnabled` is set to `true`. This is done through the manifest file. Namely, if we 
+want to test whether CometBFT respects ABCI++ grammar, we would need to enable these tests by adding `abci_tests_enabled = 
+true` in the manifest file of a particular testnet (e.g. `networks/ci.toml`). This will automatically activate logging on the 
+application side. 
+
+The `Verify()` method is shown below. 
+```go
+func (g *GrammarChecker) Verify(reqs []*abci.Request, isCleanStart bool) (bool, error) {
+	r := g.filterRequests(reqs)
```
sergio-mena, line 207: Here, you are using `reqs` and validating its contents just after. Shouldn't you move this line after 196? (so that you first validate `reqs` and then use it)

sergio-mena: .. or should line 194 use `r` rather than `reqs`?

docs/architecture/adr-108-e2e-abci++.md
```diff
@@ -0,0 +1,296 @@
...
+	}
+	return false, fmt.Errorf("%v\nFull execution:\n%v", g.combineErrors(errors, g.cfg.NumberOfErrorsToShow), g.addHeightNumbersToTheExecution(execution))
+}
+```
+
+It receives a set of ABCI++ requests and a flag saying whether they represent a `CleanStart` execution or not and does the following things:
+- Filter the requests by calling the method `filterRequests()`. This method will remove all the requests from the set that are not supported by the current version of the grammar. In addition, it will filter the last height by removing all ABCI++ requests after the 
+last `Commit`. The function `fetchABCIRequests()` can be called in the middle of the height. As a result, the last height may be incomplete and 
+classified as invalid, even if that is not the reality. The simple example here is that the last 
+request fetched via `fetchABCIRequests()` is `Decide`; however, `Commit` happens after
```
sergio-mena, line 230: By `Decide`, did you mean `EndBlock` (or `FinalizeBlock` in `v0.38.x`) ?

sergio-mena (commented):
Thanks the @nenadmilosevic95 for the many improvements. I think this PR is now ready for "official" review. Can you please move it away from draft status?

docs/architecture/adr-108-e2e-abci++.md
nenadmilosevic95: Decide refers to FinalizeBlock: the terminal name in the grammar is Decide, while the value is FinalizeBlock 
(`decide = %s"<FinalizeBlock>"`). This may need to be changed, but it is like that also in the grammar on main and on v0.38.x branch.

sergio-mena: I see. In `decide = %s"<FinalizeBlock>"`, technically, `decide` is a non-terminal that gets replaced by terminals "<FinalizeBlock>".
If you think this causes confusion, since we are not planning to backport this project (at least not beyond `v0.38.x`) maybe we can replace `decide` by `finalize`or `finalize-block` or something similar.

nenadmilosevic95: Yes, we can, just we need to rename it everywhere I guess. In my ADR it is not the problem but not sure where else.

sergio-mena: I don't think we need to replace all occurrences of the work decide (the concept is clear and understood), just the occurrences of the grammar's `decide` symbol. I quickly checked, and it looks like it's only 4 occurrences.

docs/architecture/adr-108-e2e-abci++.md
```diff
@@ -0,0 +1,304 @@
+# ADR 108: E2E tests for CometBFT's behaviour in respect to ABCI++.
+
+## Changelog
+- 2023-08-08: Initial version (@nenadmilosevic95)
+
+
+## Context
+
+We want to be able to test whether CommetBFT respects the ABCI++ grammar. To do this, we need to enhance the e2e tests infrastructure. Specifically, we plan to do three things:
```
josef-widder, line 9:
```suggestion
ABCI++ defines the interface between the application and CometBFT. A part of the specification is the ABCI++ grammar that describes the sequences of calls that the application can expect from CometBFT.
In order to demonstrate that CometBFT behaves as expected from the viewpoint of the application, we need to test whether CometBFT respects this ABCI++ grammar. To do this, we need to enhance the e2e tests infrastructure. Specifically, we plan to do three things:
```

josef-widder (commented):
Looks great. I just proposed some text in the context section. Feel free to ignore if it is too much ;-)

docs/architecture/README.md
```diff
@@ -52,6 +52,8 @@ numbering our ADRs from 100 onwards.
 
 ### Implemented
 
+- [ADR-108: E2E tests for CometBFT's behaviour in respect to ABCI++.](./adr-108-e2e-abci++.md)
```
sergio-mena, line 55:
```suggestion
- [ADR-108: E2E tests for CometBFT's behaviour in respect to ABCI 1.0.](./adr-108-e2e-abci++.md)
```

docs/architecture/adr-108-e2e-abci++.md
```diff
@@ -0,0 +1,305 @@
+# ADR 108: E2E tests for CometBFT's behaviour in respect to ABCI++.
```
sergio-mena, line 1:
```suggestion
# ADR 108: E2E tests for CometBFT's behaviour in respect to ABCI 1.0.
```

docs/architecture/adr-108-e2e-abci++.md
```diff
@@ -0,0 +1,305 @@
...
+
+## Changelog
+- 2023-08-08: Initial version (@nenadmilosevic95)
+
+
+## Context
+
+ABCI++ defines the interface between the application and CometBFT. A part of the specification is the ABCI++ grammar that describes the sequences of calls that the application can expect from CometBFT.
```
sergio-mena, line 9:
```suggestion
ABCI 1.0 defines the interface between the application and CometBFT. A part of the specification is the ABCI++ grammar that describes the sequences of calls that the application can expect from CometBFT.
```
Please, change ABCI++ to ABCI 1.0 all over. (ABCI 1.0 -> `PrepProp` and `ProcProp` but no vote extensions).
Whereas ABCI++ tends to be identifed with the whole lot (a.k.a. ABCI 2.0)

sergio-mena, line 9: Please add a link to the grammar's section in the spec

docs/architecture/adr-108-e2e-abci++.md
```diff
@@ -0,0 +1,305 @@
...
+In order to demonstrate that CometBFT behaves as expected from the viewpoint of the application, we need to test whether CometBFT respects this ABCI++ grammar. To do this, we need to enhance the e2e tests infrastructure. Specifically, we plan to do three things:
+- Log every CometBFT's ABCI++ request during the execution.
+- Parse the logs post-mortem and extract all ABCI++ requests.
+- Check if the set of observed requests respects the ABCI++ grammar.
+
+
+Issue: [353](https://github.com/cometbft/cometbft/issues/353).
+
+Current version does not support vote extensions. However, this is the next step.
```
sergio-mena, line 18:
```suggestion
Current version, ABCI 1.0, does not support vote extensions (ABCI 2.0). However, this is the next step. 
```

docs/architecture/adr-108-e2e-abci++.md
```diff
@@ -0,0 +1,305 @@
...
+- Log every CometBFT's ABCI++ request during the execution.
+- Parse the logs post-mortem and extract all ABCI++ requests.
+- Check if the set of observed requests respects the ABCI++ grammar.
+
+
+Issue: [353](https://github.com/cometbft/cometbft/issues/353).
+
+Current version does not support vote extensions. However, this is the next step. 
+
+
```
sergio-mena, line 20:
```suggestion
```

docs/architecture/adr-108-e2e-abci++.md
```diff
@@ -0,0 +1,305 @@
...
+- Takes the output of a specific node in the testnet from the moment we launched the testnet until the function is called. The node name is passed as a function parameter. It uses the `docker-compose logs` and `grep nodeName` commands. 
+- Parses the logs line by line and extracts the  `abci.Request`, if one exists. The request is received by forwarding each line to the `app.GetABCIRequestFromString(req)` method.
+- Returns the array of slices where each slice contains the set of `abci.Request`s logged on that node. Every time a crash happens, a new array element (new slice `[]*abci.Request`) will be created. We know a crash has happened because we log "Application started" every time the application starts. Specifically, we added this log inside `NewApplication()` function in `test/e2e/app/app.go` file. In the end, `fetchABCIRequests()` will return just one slice if the node did not experience any crashes and $n+1$ slices if there were crashes, $n$ being the number of crashes. The benefit of logging the requests in the previously described way is that now we can use `[]*abci.Request` to store ABCI++ requests of any type.
```
sergio-mena, line 92:
```suggestion
- Returns the array of slices where each slice contains the set of `abci.Request`s logged on that node. Every time a crash happens, a new array element (new slice `[]*abci.Request`) will be created. We know a crash has happened because we log "Application started" every time the application starts. Specifically, we added this log inside `NewApplication()` function in `test/e2e/app/app.go` file. In the end, `fetchABCIRequests()` will return just one slice if the node did not experience any crashes and $n+1$ slices if there were $n$ crashes. The benefit of logging the requests in the previously described way is that now we can use `[]*abci.Request` to store ABCI++ requests of any type.
```
it's just a suggestion

docs/architecture/adr-108-e2e-abci++.md
```diff
@@ -0,0 +1,305 @@
...
+If you compare this grammar with the original one, you will notice that, in addition to vote extensions,  
+`Info` is removed. The reason is that, as explained in the section [CometBFT's expected behaviour](../../spec/abci/abci%2B%2B_comet_expected_behavior.md#valid-method-call-sequences), one of the 
+purposes of the `Info` method is being part of the RPC handling from an external 
+client. Since this can happen at any time, it complicates the 
+grammar.  
+This is not true in other cases, but since the Application does 
+not know why the `Info` is called, we removed 
+it totally from the new grammar. The Application is still logging the `Info` 
+call, but a specific test would need to be written to check whether it happens
+in the right moment.
```
sergio-mena, line 147:
```suggestion
at the right moment. 
```

docs/architecture/adr-108-e2e-abci++.md
```diff
@@ -0,0 +1,305 @@
...
+
+It receives a set of ABCI++ requests and a flag saying whether they represent a `CleanStart` execution or not and does the following things:
+- Checks if the execution is an empty execution. 
+- Filter the requests by calling the method `filterRequests()`. This method will remove all the requests from the set that are not supported by the current version of the grammar. In addition, it will filter the last height by removing all ABCI++ requests after the 
+last `Commit`. The function `fetchABCIRequests()` can be called in the middle of the height. As a result, the last height may be incomplete and 
+classified as invalid, even if that is not the reality. The simple example here is that the last 
+request fetched via `fetchABCIRequests()` is `FinalizeBlock`; however, `Commit` happens after 
+`fetchABCIRequests()` was invoked. Consequently, the execution
+will be considered as faulty because `Commit` is missing, even though the `Commit` 
+will happen after. This is why if the execution consists of only one incomplete height and function `filterRequests()` returns an empty set of requests, the `Verify()` method considers this execution as valid and returns `true`.
```
sergio-mena, line 242: If could be good to mention here that this mechanism effectively turns the original grammar, which is infinite (omega grammar) into a finite one. In terms of the grammar, we change multiplier `(inf)` by `*`.

Anyway, this is a nice-to-have. Up to you if you want to explain it here.

docs/architecture/adr-108-e2e-abci++.md
```diff
@@ -0,0 +1,305 @@
...
+```bash 
+make grammar
+``` 
+Notice here that you need to have `gogll` installed 
+on your machine to run the make successfully. If this is not the case, you can install it with the following command: 
+
+```bash 
+go get github.com/goccmack/gogll/v3
+```  
+
```
sergio-mena, line 282:
```suggestion
Make sure you commit any changes to the auto-generated code together with the changes to the grammar.

```

docs/architecture/adr-108-e2e-abci++.md
```diff
@@ -0,0 +1,305 @@
...
+- The application needs to log the new request in the same way as we do now.
+- We should include the new request to the grammar and generate a new parser and lexer.  
+- We should add new requests to the list of supported requests. Namely, we should modify the function `isSupportedByGrammar()` in `test/e2e/pkg/grammar/checker.go` to return `true` for the new type of requests.
+
+## Status
+
+Implemented.
+
+To-do list:
+- adding the CI workflow to check if make grammar is executed.
```
sergio-mena, line 297:
```suggestion
- adding the CI workflow to check if make grammar is executed. 
- extend this ADR (and implementation) to support ABCI 2.0 (i.e., ABCI calls related to vote extensions)
- in the future, we might consider whether the logging (actually, tracing) should be done on the e2e application side, or on CometBFT side, so this infra can be reused for MBT-like activities)
```

sergio-mena (approved):
LGTM.
Some more comments but none is blocking the merge.
Thanks a lot for this work @nenadmilosevic95 !

docs/architecture/adr-108-e2e-abci++.md
nenadmilosevic95: Added.

docs/architecture/adr-108-e2e-abci++.md
nenadmilosevic95: Addressed. Not precisely here, but in the part where we explain the difference between the new and original grammar.

docs/architecture/README.md
```diff
@@ -52,6 +52,8 @@ numbering our ADRs from 100 onwards.
 
 ### Implemented
 
+- [ADR-108: E2E tests for CometBFT's behaviour in respect to ABCI 1.0.](./adr-108-e2e-abci++.md)
```
thanethomson, line 55: Has this ADR actually been implemented yet?

nenadmilosevic95: It is implemented here - https://github.com/cometbft/cometbft/pull/930. However, it still needs to be reviewed.

sergio-mena: > we also need a tracking issue for its implementation

It's #353, I just edited its DoD to refer to the actual PRs

PR #904, opened 2023-06-01
consensus: optimize vote and block part gossip with HasProposalBlockPartMessage and random sleeps

ebuchman:
Addresses https://github.com/tendermint/tendermint/issues/627 (finally!) 

This PR is based off #896 which added new metrics to measure duplicate votes and block parts received from peers.

## Summary

This PR introduces two optimizations to the consensus gossip: HasProposalBlockPartMessage and random sleeps in the fast loop in the gossip routines.

Control for both were added to the config file for testing purposes, but at least the HasProposalBlockPartMessage should probably not be in the config, it should just be hardcoded active once we're happy with it.

On another note, we may want to consider some changes so this can be rolled out in a non-breaking way. If I'm not mistaken since the message was added to an existing p2p channel it will be interpreted by peers on older versions as malicious. My understanding is we could just move it to a new channel to solve this so the change would be backwards compatible and nodes can roll it out progressively. 

### HasProposalBlockPartMessage

The first is a new HasProposalBlockPartMessage. This follows identical logic to the HasVoteMessage. When a node receives a ProposalBlockPart, it broadcasts the HasProposalBlockPartMessage to all its peers. This informs them that they don't need to send the node the part, thus reducing the amount of redundant block parts gossiped over the network.

### Gossip Sleeps

The second change is the introduction of a random sleep in the fast loops of the gossipVote and gossipData routines. We run these routines for each peer, and when there are things to send, the routines run in a tight loop picking and sending votes or parts to send to the peer. This reduces the effectiveness of the "HasXXXMessage" mechanism since there is minimal opportunity for the peer state to be updated with when the HasXXXMessages are received before the new votes/parts are selected to be sent. Adding a small random sleep gives the reactor a chance to breath and process inbound HasXXXMessages before picking new votes/parts to send. Randomizing the sleep ensures the peers don't all sleep the same amount, which could negate its effect.

## Results

Combining the new HasBlockPartMessage with the random sleep in gossipData routine results in a ~20% bandwidth reduction on BlockPart messages in 4 node and 7 node networks under load as measured in a local testnet on my laptop, as well as 4 node networks in the cloud. Larger tests are still to be conducted, but these initial results are very promising.

The random sleep in gossipVote routine also results in a ~15% bandwidth reduction on Vote messages in 7 node networks as measured in a local testnet on my laptop. Again, promising, but more tests to be done.

The largest consumers of bandwidth in the comet system are mempool txs, votes, and block parts. While mempool txs are not addressed here, the latter two are, and should lead to significant reductions in bandwidth usage overall.

Note the addition of a new HasProposalBlockPartMessage adds minimal additional bandwidth itself as each message contains only 3 integers (one 64 bit, two 32 bit, so 16 bytes total) compared to hundreds of bytes in a block part message. This was also confirmed experimentally. 

Some teaser screenshots below on small local networks using the cometbft `e2e` testing system on my laptop.

### Votes

With and without the random sleep (upper bounded at 50ms) on a 7 node network, with no load:

image

The main lines visible are the bandwidth from the Vote (the larger one) and BlockPart messages. Note how both are reduced ~15% from adding the sleep (the left). The effect of the sleep on block times is negligible (they rise from 1.6 to 1.7 seconds per block in this experiment)

### Block Parts

On a 4 node network, with and without the new HasProposalBlockPartMessage and with and without the random sleep on the gossipDataRoutine:

image

Ignore the spikes on the right for now, this is the system acting up under load running all 4 nodes on my laptop ... The image shows 4 experiments. The purple line is bandwidth from mempool txs (roughly the same in all experiments), the green is bandwidth from block part messages (what we're focusing on). The experiments, and the approximate block part bandwidth when it first flattens out, are:

1. has block parts, sleep (680k)
2. no has block parts, sleep (830k)
3. has block parts, no sleep (730k)
4. no has block parts, no sleep (870k)

The last experiment is the current baseline. Adding the sleep alone without the new message type (2nd experiment) has some minimal impact (870 -> 830). Adding just the new message type without the sleep (3rd experiment) has more significant impact (870 -> 730). But adding both together (1st experiiment) has the most significant impact (870->680), a reduction of ~20%. 

While we've done some preliminary experiments here, we still need to do larger experiments on cloud deployments, as well as writing some actual tests :) 

---

#### PR checklist

- [ ] Tests written/updated


consensus/reactor.go
```diff
@@ -430,6 +433,16 @@ func (conR *Reactor) subscribeToBroadcastEvents() {
 		}); err != nil {
 		conR.Logger.Error("Error adding listener for events", "err", err)
 	}
+
+	if conR.conS.config.HasBlockPart {
```
ebuchman, line 437: I think we might want the if statement inside the callback since the event will always be fired, not sure if we need to ensure theres a listener for it even if its a noop

consensus/reactor.go
```diff
@@ -545,6 +571,17 @@ OUTER_LOOP:
 		if !peer.IsRunning() || !conR.IsRunning() {
 			return
 		}
+
+		// sleep random amount to give reactor a chance to receive HasProposalBlockPart messages
```
ebuchman, line 575: these sleeps are at the top of the loop, so effect everything, including catchup logic, which is probably not optimal, but maybe ok.

sergio-mena:
# Results for a 200-node testnet

The goal of these tests is to validate @ebuchman's results obtained on 4 and 7 node testnets (local to his laptop). These tests are run a 200-node testnet on a data center: 175 validators, 20 non-validating full nodes, 5 seed nodes.

The baseline is `v0.38.0-rc.1`. Vote extensions are active, but the code changes do not deal with them in any way.

## In a nutshell

In case you want to skip the details, here are the three key takeaways:

* With `HasProposalBlockPartMessage` and  the gossip sleeps, the bandwidth savings are not only confirmed, but more pronounced in a 200-node testnet.
  * _**when no TXs: the BW improvements is 31%**_
  * _**with high load the BW improvements is 49%**_
* Changing the `flush_throttle_timeout` p2p parameter has no clear incidence in the results (so we better leave it alone)
* The gains reported below are probably more pronounced in this testnet than in a real deployment, as all our nodes are in the same data center, so latencies are lower than typically found on deployed blockchains. Yet, they are _substantial_.

## Experiments

### Experiment description

This is what was tested for each experiment in the order the appear in the graph

1. baseline
2. flush_throttle_timeout: 50ms + gossip_sleep: 50ms
3. flush_throttle_timeout: 50ms + _________________ + has_block_part
4. flush_throttle_timeout: 50ms + gossip_sleep: 50ms + has_block_part
5. _________________________ + gossip_sleep: 50ms + has_block_part

### Results

#### Idle system (no TXs submitted , empty blocks)

The _from --> to_ column shows which two experiments we are comparing.
The measurements are the total bandwidth consumed by all nodes for that particular message type.


| From --> to | Votes                              | Block Parts                    |
|-------------|------------------------|------------------------|
| 1 --> 2	       | 255M -> 80M  = -68% | 260M -> 255M = -2%  |
| 1 --> 3	       | 255M -> 254M = 0%    | 260M -> 243M = -7%  |
| 1 --> 5          | 255M -> 76M  = -70% | 260M -> 180M = _**-31%**_ |
| 5 --> 4         | 76M -> 77M   = +1%     | 180M -> 180M = 0%    |

#### System under high load (400 1-kB-long TXs submitted per second -- close to baseline's saturation point)

| From --> to | Votes                              | Block Parts                    |
|-------------|------------------------|------------------------|
| 1 --> 2         | 59M -> 25M = -58%     | 3.7G -> 2.8G = -24%    |
| 1 --> 3         | 59M -> 49M = -17%     | 3.7G -> 3.5G = -5.4%   |
| 1 --> 5         | 59M -> 20M = -66%     | 3.7G -> 1.9G = _**-49%**_ |
| 5 --> 4        | 20M -> 25M = +25%     | 1.9G -> 1.9G = 0%        |

As we can see for both tables, the gains by `gossip_sleep` and `has_block_part` are noticeable in isolation, but when combined they are even bigger.

### Graph

The metrics shown here are the same as for @ebuchman's experiments. For each experiment, the chain is allowed to stabilize before 400 TXs/s (1-kb-long) are generated during 300 s. Then the experiment is continued until the main's system metrics return to what they were before the TX load was started.

There are three lines:
* The greenish-yellow plot, which is the total bandwidth taken by block parts.
* The plain green plot: bandwidth taken by TXs, which doesn't change as the code changes we are testing do not deal with them
* The lighter greenish-yellow plot (the lowest of the three plots), which is the total bandwidth taken by votes.

The X axis is experiments' wall-clock time. The measurements presented in the tables in the previous section are extracted from these plots

Screenshot 2023-06-05 at 18 45 12

sergio-mena:
Rebased to `main`

config/config.go
```diff
@@ -957,6 +958,7 @@ func DefaultConsensusConfig() *ConsensusConfig {
 		CreateEmptyBlocksInterval:   0 * time.Second,
 		PeerGossipSleepDuration:     100 * time.Millisecond,
 		PeerQueryMaj23SleepDuration: 2000 * time.Millisecond,
+		PeerGossipFastSleepDuration: 50 * time.Millisecond,
```
sergio-mena, line 961:
```suggestion
		PeerGossipFastSleepDuration: 0 * time.Second,
```

consensus/reactor.go
sergio-mena: Scale tests show overall improvement, so let's keep it this way for the time being.

However, I've changed the default value in config to 0, as it's still experimental. Folks can try raising it and measuring its impact in their own networks.

consensus/reactor.go
```diff
@@ -1809,3 +1879,29 @@ func (m *VoteSetBitsMessage) String() string {
...
+type HasProposalBlockPartMessage struct {
+	Height int64
+	Round  int32
+	Index  int32
+}
+
+// ValidateBasic performs basic validation.
+func (m *HasProposalBlockPartMessage) ValidateBasic() error {
+	if m.Height < 0 {
+		return errors.New("negative Height")
```
sergio-mena, line 1893:
```suggestion
	if m.Height < 1 {
		return errors.New("invalid Height (< 1)")
```

consensus/state.go
```diff
@@ -1891,6 +1891,7 @@ func (cs *State) defaultSetProposal(proposal *types.Proposal) error {
 	}
 
 	proposal.Signature = p.Signature
+
```
sergio-mena, line 1894:
```suggestion
```

.changelog/unreleased/improvements/896-consensus-metric-duplicates.md
```diff
@@ -0,0 +1,2 @@
+- `[mempool]` New metrics (counters) to track duplicate votes and block parts.
+  ([\#896](https://github.com/cometbft/cometbft/pull/896))
```
sergio-mena, line 2: Forgot to add this when merging #896

.changelog/unreleased/improvements/896-consensus-metric-duplicates.md
```diff
@@ -0,0 +1,2 @@
+- `[mempool]` New metrics (counters) to track duplicate votes and block parts.
```
sergio-mena, line 1:
```suggestion
- `[consensus]` New metrics (counters) to track duplicate votes and block parts.
```

.changelog/unreleased/improvements/896-consensus-metric-duplicates.md
```diff
@@ -0,0 +1,2 @@
+- `[consensus]` New metrics (counters) to track duplicate votes and block parts.
+  ([\#896](https://github.com/cometbft/cometbft/pull/896))
```
sergio-mena, line 2: Forgot to add this when merging #896

consensus/reactor.go
```diff
@@ -417,28 +417,28 @@ func (conR *Reactor) subscribeToBroadcastEvents() {
 		func(data cmtevents.EventData) {
 			conR.broadcastNewRoundStepMessage(data.(*cstypes.RoundState))
 		}); err != nil {
-		conR.Logger.Error("Error adding listener for events", "err", err)
+		conR.Logger.Error("Error adding listener for events (NewRoundStep)", "err", err)
```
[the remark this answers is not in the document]
sergio-mena: Submitted #1002, since I'd like to tackle this problem separately.

sergio-mena (approved):
This approval refers to the base work done by @ebuchman , since approving my own changes doesn't make sense :-)

cmwaters:
Is this a breaking change to the p2p network? It looks like it would be because peers on an older version would not recognise the `HasProposalBlockPart ` message and boot the peer.

sergio-mena:
@cmwaters Indeed, that's one of main reasons why we are not backporting this

cmwaters:
Thanks @sergio-mena. Just wanted to make sure as we're interested in porting it across. Might want to consider bumping the `P2PVersion` [here](https://github.com/cometbft/cometbft/blob/86c07f20f163c17a0f5b7149cbd441a242b88557/version/version.go#L12)

sergio-mena:
Thanks @cmwaters, good point.
We usually do it as part of the process of cutting a release ([here](https://github.com/cometbft/cometbft/blob/86c07f20f/RELEASES.md?plain=1#L205)). But I guess I'll open a PR now in case we forget at that time.

PR #905, opened 2023-06-01
New metrics to track duplicate votes and block parts  (#896) -- Forward port

sergio-mena:
In order to be able to measure performance with a system close to the QA results for `v0.38.x`, the original patch (#896) was created on this branch rather than `main`.

This is a forward port of #896 to `main`.

---

#### PR checklist

- [ ] Tests written/updated


consensus/state.go
```diff
@@ -2013,6 +2018,8 @@ func (cs *State) handleCompleteProposal(blockHeight int64) {
 // Attempt to add the vote. if its a duplicate signature, dupeout the validator
 func (cs *State) tryAddVote(vote *types.Vote, peerID p2p.ID) (bool, error) {
 	added, err := cs.addVote(vote, peerID)
+
+	// NOTE: some of these errors are swallowed here
```
mzabaluev, line 2022: Can you clarify? This comment looks a bit cryptic.

consensus/state.go
```diff
@@ -2161,6 +2172,11 @@ func (cs *State) addVote(vote *types.Vote, peerID p2p.ID) (added bool, err error
 	added, err = cs.Votes.AddVote(vote, peerID, extEnabled)
 	if !added {
 		// Either duplicate, or error upon cs.Votes.AddByIndex()
+
+		// If the vote wasnt added but there's no error, its a duplicate vote
```
mzabaluev, line 2176: It's hard to follow the logic in this monster function. It's not a fault of this PR in particular, though.

types/part_set.go
```diff
@@ -270,9 +270,12 @@ func (ps *PartSet) Total() uint32 {
 }
 
 func (ps *PartSet) AddPart(part *Part) (bool, error) {
+	// TODO: remove this? would be preferable if this only returned (false, nil)
+	// when its a duplicate block part
 	if ps == nil {
 		return false, nil
 	}
```
mzabaluev, line 277: Does leaving this check here mean that a metric can be incremented on the wrong condition, which also "should not occur" in correct code and so it's hard to debug when it does occur?

consensus/state.go
sergio-mena: We can improve the message, but not in this PR: I'd like it to stay a forward-port (so as close a possible to the original one)

types/part_set.go
sergio-mena: I added a similar comment in the original PR. I think we should change line 276 to return a non-nil error. But it is orthogonal to the original PR (and to this one)

mzabaluev: Filed as #920.

mzabaluev (approved):

PR #919, opened 2023-06-05
Introduce Signature scheme BN254

blasrodri:
This PR introduces a new signature scheme (i.e.: bn254) into CometBFT.

#### PR checklist

- [ ] Tests written/updated

Credit to: https://github.com/hussein-aitlahcen/cometbft/commit/a88397bcdaa0a5e2354634c66043392090d44f18


zmanian:
I'd like to better understand the conditions where this signature scheme might be useful.

Right now each precommit vote is a unique value, these values can still be aggregated but verification requires a pairing operation per message which is very slow.

this could be viewed as an incremental step forward towards being BLS signature friendly in Comet BFT but I can't really see a situation where someone would find this key type instantly useful

PR #908, opened 2023-06-04
fix: lint fail when golangci-lint bump from v1.52.2 to v1.53.x

mmsqe:
add config for depguard and fix nakedret error, for more [info](https://github.com/cometbft/cometbft/pull/865#discussion_r1215620869)

---

#### PR checklist

- [ ] Tests written/updated
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments

PR #922, opened 2023-06-05
backport: Fix lint failures on `v0.38.x`

thanethomson:
Backport of #908 to `v0.38.x`

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -42,3 +42,55 @@ linters-settings:
     suggest-new: true
   misspell:
     locale: US
+  depguard:
+    rules:
+      main:
+        files:
+          - $all
+          - "!$test"
+        allow:
+          - $gostd
+          - github.com/cometbft
+          - github.com/cosmos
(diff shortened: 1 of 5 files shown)


.golangci.yml
```diff
@@ -42,3 +42,55 @@ linters-settings:
     suggest-new: true
   misspell:
     locale: US
+  depguard:
+    rules:
```
mzabaluev, line 46: Can't claim I understand the logic behind these exceptions, other than
"grab and override whatever the linter is currently complaining about"...

mzabaluev (approved):
...but then, you have already approved and merged the PR with the same changes against main, so let's not diverge here.

The coding style changes look good.

.golangci.yml
thanethomson: Afaik `depguard` is just one of the linters that golangci-lint runs, and this provides explicit configuration for it. See https://github.com/OpenPeeDeeP/depguard

Issue #615, opened 2023-03-30
spec/mempool: Define safety and liveness properties


hvanz:

hvanz:
[This document](https://github.com/cometbft/knowledge-base/blob/main/protocols/mempool-overview.md) in knowledge-base repo defines the main properties of the mempool.

Issue #24, opened 2022-12-23
ABCI 2.0: Spec work


sergio-mena:
Tasks:

  - [x] #444
  - [x] #483
  - [x] #403
  - [x] #2760
  - [x] Follow tendermint/tendermint#8796 and make sure the changes are present in the spec 
      - Move `data` from `Commit` to `FinalizeBlock` (as `AppHash`)
      - Rename `Misbehaviour` to `ByzantineValidators` (done already?)
  - [x] #203 
  - [x] #2698
  - [x] #779
  - [x] #2679
  - [x] Final read of spec for overall coherence

Issues identified:
- [x] #1326

DoD:

  - Out of all solutions proposed in RFC017, the one implemented (sub-optimal storage) is now part of the spec
  - Spec is updated to ABCI 2.0 (very similar to the one in `v0.36.x`)
  - Spec is complete and proof-read
  - Spec methods and data structures match the corresponding protobufs

Original issue: tendermint/tendermint#9917

sergio-mena:
Moved to Q3 as there are some final tasks we didn't manage to do in Q2 (they are all non-blocking for the `v0.38.0` release)

sergio-mena:
Thanks @andynog for driving this to completion

PR #928, opened 2023-06-06
Add requirement for `CheckTx` in ABCI spec

sergio-mena:
Contributes to #615 and #24

Following today's discussions with @otrack and @hvanz, we agreed I'd give a try at adding an additional requirement in ABCI spec to restrict the possible app implementations of `CheckTx`.

This might help simplify the mempool spec, which can refer to this ABCI requirement to become simpler.

The advantage of this approach is that we add this requirement to those we already require all ABCI applications to abide by.

---

#### PR checklist

- [ ] Tests written/updated


spec/abci/abci++_app_requirements.md
```diff
@@ -188,6 +188,31 @@ Likewise, `ExtendVote` can also be non-deterministic:
...
+
+
+* Requirement 13 [`CheckTx`, non-oscillation]: For any correct process *p*
+  and any transaction *tx*,
+  (a) either *OK(CheckTxCode<sub>tx,p,t<sub>0</sub></sub>) = OK(CheckTxCode<sub>tx,p,t<sub>1</sub></sub>)*
+  for any two times *t<sub>0</sub>* and *t<sub>1</sub>* in *p*'s execution, or
+  (b) there exists a time *t<sub>stable</sub>* in *p*'s execution
+  such that *&#172; OK(CheckTxCode<sub>tx,p,t</sub>)* for all *t > t<sub>stable</sub>*.
+
+Requirement 13 has two main conditions. The implementation of `CheckTx` must fullfill at least one of them.
```
jmalicevic, line 206:
```suggestion
Requirement 13 has two main conditions. The implementation of `CheckTx` must fulfill at least one of them.
```

spec/abci/abci++_app_requirements.md
```diff
@@ -188,6 +188,31 @@ Likewise, `ExtendVote` can also be non-deterministic:
...
+Condition (a) mandates that `CheckTx` on a transaction never changes its result,
+in other words, that the checks implemented are stateless (e.g., well-formedness, signature verification).
+Condition (b) states that, if the transaction stays in the mempool long enough, if will eventually fail
```
jmalicevic, line 209:
```suggestion
Condition (b) states that, if the transaction stays in the mempool long enough, it will eventually fail
```

spec/abci/abci++_app_requirements.md
```diff
@@ -188,6 +188,31 @@ Likewise, `ExtendVote` can also be non-deterministic:
...
+`CheckTx` forever. Finally, note that Requirement 13 is local to process *p*, i.e.,
```
jmalicevic, line 210: I assume the point of (b) is to show that a transaction will eventually leave the mempool one way or the other? I did not read the rest of the text in the file, and can't remember whether it is written down, but if it is not, it might be useful to emphasize that here.

spec/abci/abci++_app_requirements.md
```diff
@@ -188,6 +188,31 @@ Likewise, `ExtendVote` can also be non-deterministic:
...
+to CometBFT's `RequestCheckTx` call occurring at time *t* and having transaction *tx* as parameter.
+Let predicate *OK(CheckTxCode)* denote whether *CheckTxCode* is `SUCCESS`.
+
+
+* Requirement 13 [`CheckTx`, non-oscillation]: For any correct process *p*
+  and any transaction *tx*,
+  (a) either *OK(CheckTxCode<sub>tx,p,t<sub>0</sub></sub>) = OK(CheckTxCode<sub>tx,p,t<sub>1</sub></sub>)*
+  for any two times *t<sub>0</sub>* and *t<sub>1</sub>* in *p*'s execution, or
+  (b) there exists a time *t<sub>stable</sub>* in *p*'s execution
+  such that *&#172; OK(CheckTxCode<sub>tx,p,t</sub>)* for all *t > t<sub>stable</sub>*.
```
jmalicevic, line 204: Where for all times *t <= t<sub>stable</sub>* , the value was *OK(CheckTxCode<sub>tx,p,t</sub>)* ?

spec/abci/abci++_app_requirements.md
```diff
@@ -188,6 +188,31 @@ Likewise, `ExtendVote` can also be non-deterministic:
 * *w<sup>r</sup><sub>p</sub> = w<sup>r</sup><sub>q</sub> &#8655;
   e<sup>r</sup><sub>p</sub> = e<sup>r</sup><sub>q</sub>*
 
+Unlike all requirements described above,
+which affect the ABCI methods of the [Consensus Connection](#consensus-connection),
+the following requirement is on the [Mempool Connection](#mempool-connection).
+Let *CheckTxCode<sub>tx,p,t</sub>* denote the error code returned by *p*'s Application, via `ReponseCheckTx`,
```
hvanz, line 194:
```suggestion
Let *CheckTxCode<sub>tx,p,t</sub>* denote the error code returned by *p*'s Application, via `ResponseCheckTx`,
```

spec/abci/abci++_app_requirements.md
lasarojc, line 204:
```suggestion
  and any transaction *tx*,
  
  - (a) either *OK(CheckTxCode<sub>tx,p,t<sub>0</sub></sub>) = OK(CheckTxCode<sub>tx,p,t<sub>1</sub></sub>)*
  for any two times *t<sub>0</sub>* and *t<sub>1</sub>* in *p*'s execution, or
  - (b) there exists a time *t<sub>stable</sub>* in *p*'s execution
  such that *&#172; OK(CheckTxCode<sub>tx,p,t</sub>)* for all *t > t<sub>stable</sub>*.
```

lasarojc: Good point. Otherwise there will be no stability until `stable` and if `stable` can take very long to arrive, we are back to square one in practice.

spec/abci/abci++_app_requirements.md
lasarojc, line 209: > if the transaction stays in the mempool long enough

Does it have to stay in the mempool or could it be added and removed repeatedly before the stabilization?

spec/abci/abci++_app_requirements.md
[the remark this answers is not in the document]
sergio-mena: I thought about it. I know the mempool gets height info from Consensus, which is a dependency between the two modules. I used time here in the hope that, one day, we can get rid of height info in the mempool. If you have a strong opinion on this, we can change time for height.

Also, now that I think of it, if we use height, then we need an additional (previous) requirement that the result does not change within one height if executed several times. Actually, I like this two-requirement solution. If you also like it more than what's there I'll change it.

spec/abci/abci++_app_requirements.md
sergio-mena: I would let @otrack and @hvanz chime in here, but I have the impression we cannot require anything from the app before  t<sub>stable</sub>. This condition is just making sure that we don't oscillate forever: which app can easily fulfill with e.g., nonces. But remember that CheckTx cannot offer strong guarantees when just looking at a particular instance. BTW, I'm adding a sentence at @jmalicevic's suggestion that hopefully clarifies this point. I will extend that sentence with (parts of) this explanation.

spec/abci/abci++_app_requirements.md
sergio-mena: Good point. Let me try to clarify that

spec/abci/abci++_app_requirements.md
sergio-mena: That's what I understand we're trying to avoid: some nodes saying it's valid, gossiping it, then saying it is invalid... and the TX circulating around the p2p network but always disappearing them moment it is to be proposed.
Check the latest version, where I try to clarify that.

spec/abci/abci++_app_requirements.md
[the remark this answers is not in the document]
hvanz: I thought that it was a requirement that handling CheckTx in the application should be deterministic when validated against the same height. In the section "Specifics of `ResponseCheckTx`" in this document, it says that the `Data` field does not need to be deterministic (which btw is not mentioned in [proto/tendermint/abci/types.proto](https://github.com/cometbft/cometbft/blob/main/proto/tendermint/abci/types.proto#L256)). But it doesn't say anything about the `code` field of the response.

spec/abci/abci++_app_requirements.md
hvanz: If the transaction is removed and added again, it will need to pass through CheckTx, with the possibility to be rejected the second time. If it stays "long enough" in the mempool, without being rechecked, most probably it will eventually become invalid, while in the mempool.

spec/abci/abci++_app_requirements.md
lasarojc: I see. Maybe a better name for the property would be "eventual stability" or "eventual non-oscillation", given that it can oscillate for a while.

otrack:
I also believe that it should be about the height and not the time at which the call is made. Moreover, in my view the two requirements can be merged into a single one: 

> In every run, there is a height after which `CheckTx` always returns the same value.

This invariant ensures that INV4 in the mempool specification holds. INV4 is necessary for a transaction to eventually leave the mempool, while doing something useful, that is either committed, or forever invalid. This [PR](https://github.com/cometbft/knowledge-base/pull/9) on the knowledge-base is detailing such a point.

lasarojc:
wrt to 
> In every run, there is a height after which CheckTx always returns the same value.

what does "after" mean?  A possibility would be to consider the height "done" once recheckTx has returned after the agreeing of the corresponding block. 

Also, on something that @hvanz asked earlier, during the same height (i.e. in between two recheckTx), does checkTx need to return the same value (supposed the height takes long enough for the tx to be expunged and received again)? That is, does the check change its result based only on the agreed blocks or could it depend on some other condition, such as the size of the mempool? If the first, how could a mempool size cap be implemented while not violating the requirements?

While I see why we would want to not use time, I am not sure using height is the best option since it strengthens the dependency of the mempool on the consensus and I would rather go the other way.
We could instead introduce some logical time, monotonically increasing, which could be implemented using wall clock or height, but doesn't have to be either. Or we could phrase the requirements in terms of invocations of checkTx: 
> eventually every two consecutive invocations of checkTx return the same value.

otrack:
> what does "after" mean? A possibility would be to consider the height "done" once recheckTx has returned after the agreeing of the corresponding block.

The term "after" refers to heights which are higher than the limit one.
Formally, one would write things this way for some process $p$:  
$\exists b \in \\{true,false\\}. \exists h_0 \in \mathbb. \square(p.height > h_0 \implies p.app.checkTx(tx) = b)$

The height corresponds to the value in the consensus module.
A cached version of this value (refreshed often enough) works also.

> Also, on something that @hvanz asked earlier, during the same height (i.e. in between two recheckTx), does checkTx need to return the same value (supposed the height takes long enough for the tx to be expunged and received again)? That is, does the check change its result based only on the agreed blocks or could it depend on some other condition, such as the size of the mempool? If the first, how could a mempool size cap be implemented while not violating the requirements?

There is no need that a call to $checkTx$ returns twice the same value.
After all, such a call is _optimistic_ and might reflect some copy of the application state on which the verification is made.
For instance, if $tx1$ enables $tx0$ then checking $tx0$ when $tx1$ is not present returns first $false$.
Once $tx1$ is added, $tx0$ can be checked again (e.g., it is received a second time and the cache overflown).
This second verification returns $true$.

Stronger requirements are also possible, but in my view applications might be interested with this freedom.
For instance, they can try to optimize incoming transactions for better throughput.
Such a things would not be possible if we ask that $checkTx$ is deterministic per height.

> While I see why we would want to not use time, I am not sure using height is the best option since it strengthens the dependency of the mempool on the consensus and I would rather go the other way. We could instead introduce some logical time, monotonically increasing, which could be implemented using wall clock or height, but doesn't have to be either. Or we could phrase the requirements in terms of invocations of checkTx:

The height captures the progress on the application:
the higher it is, the more the application progresses.
It is also possible to state things using time, as @Sergio has written.
In my view, because the client application "ticks" wrt. the height, it is more accurate to refer to this logical time.

spec/abci/abci++_app_requirements.md
```diff
@@ -188,6 +188,37 @@ Likewise, `ExtendVote` can also be non-deterministic:
...
+  and any transaction *tx*,
+  
+    - (a) either *OK(CheckTxCode<sub>tx,p,t<sub>0</sub></sub>) = OK(CheckTxCode<sub>tx,p,t<sub>1</sub></sub>)*
+      for any two times *t<sub>0</sub>* and *t<sub>1</sub>* in *p*'s execution, or
+    - (b) there exists a time *t<sub>stable</sub>* in *p*'s execution
+      such that *&#172; OK(CheckTxCode<sub>tx,p,t</sub>)* for all *t > t<sub>stable</sub>*.
+
+Requirement 13 has two main conditions. The implementation of `CheckTx` must fulfill at least one of them.
+Condition (a) mandates that `CheckTx` on a transaction never changes its result,
+in other words, that the checks implemented are stateless (e.g., well-formedness, signature verification).
```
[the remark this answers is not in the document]
jmalicevic: ` My understanding is that it should be disabled for app that do stateless CheckTx.`

If this is indeed the rule, I believe it should be written down. Maybe it is and I just have not read it. Beacuse otherwise you end up with questions such as mine :) 

Then if `RecheckTx` is enabled, I feel that the check is stateless within a height (because you are not supposed to change the state in the first place), but then the requirement a) should be relaxed such that it is allowed to not succeed only if a block was committed.

spec/abci/abci++_app_requirements.md
```diff
@@ -188,6 +188,37 @@ Likewise, `ExtendVote` can also be non-deterministic:
 * *w<sup>r</sup><sub>p</sub> = w<sup>r</sup><sub>q</sub> &#8655;
   e<sup>r</sup><sub>p</sub> = e<sup>r</sup><sub>q</sub>*
 
+Unlike all requirements described above,
+which affect the ABCI methods of the [Consensus Connection](#consensus-connection),
+the following requirement is on the [Mempool Connection](#mempool-connection).
+Let *CheckTxCode<sub>tx,p,t</sub>* denote the error code returned by *p*'s Application, via `ResponseCheckTx`,
```
cason, line 194: The return value, or return code, not necessarily an error.

spec/abci/abci++_app_requirements.md
```diff
@@ -188,6 +188,37 @@ Likewise, `ExtendVote` can also be non-deterministic:
...
+
+* Requirement 13 [`CheckTx`, finite-oscillation]: For any correct process *p*
+  and any transaction *tx*,
+  
+    - (a) either *OK(CheckTxCode<sub>tx,p,t<sub>0</sub></sub>) = OK(CheckTxCode<sub>tx,p,t<sub>1</sub></sub>)*
+      for any two times *t<sub>0</sub>* and *t<sub>1</sub>* in *p*'s execution, or
+    - (b) there exists a time *t<sub>stable</sub>* in *p*'s execution
+      such that *&#172; OK(CheckTxCode<sub>tx,p,t</sub>)* for all *t > t<sub>stable</sub>*.
+
+Requirement 13 has two main conditions. The implementation of `CheckTx` must fulfill at least one of them.
```
cason, line 207:
```suggestion
Requirement 13 has two conditions. The implementation of `CheckTx` must fulfill at least one of them.
```

spec/abci/abci++_app_requirements.md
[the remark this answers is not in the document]
cason: > should the spec consider CheckTx and RecheckTx as different? 

Yes, in my opinion.

> If yes, is there any requirement we want to impose in how they must relate?

I agree with Jasmina's comments below. If the state of the application hasn't change between `CheckTx` and `ReCheckTx` for the same transaction, their output should be the same. In other words, in the absence of state changes (probably we can refer this using heights to indicate state changes), `ReCheckTx` should be handled as it was a duplicated/second call to `CheckTx`.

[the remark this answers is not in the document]
cason: Regarding the text specifically, I think that condition (a) prevents the application to implement replay protection mechanisms. And our assumption is kind of that is _mandatory_ the application to implement such mechanisms.

I wonder how we can state condition (a) in a way it still allows the application to reject `tx` in the case the same `tx` has already been included in a committed block.

spec/abci/abci++_app_requirements.md
```diff
@@ -188,6 +188,37 @@ Likewise, `ExtendVote` can also be non-deterministic:
...
+Condition (a) mandates that `CheckTx` on a transaction never changes its result,
+in other words, that the checks implemented are stateless (e.g., well-formedness, signature verification).
+Condition (b) states that, if the transaction stays in the mempool long enough, it will eventually fail
+`CheckTx` forever.
+Condition (b) is aiming at `CheckTx` implementations that read the Application's state,
+whose result can oscillate as the tx might be validated against different states in successive calls to `CheckTx`.
+Thus, condition (b), while being local to a process *p*, will make sure that the transaction will eventually
+leave the mempool of *all* full nodes.
```
cason, line 215: Is that including the possibility of `tx` been included in a committed block? This should be, in my opinion, provided as the main example of when/how a transaction becomes "forever" invalid.

spec/abci/abci++_app_requirements.md
```diff
@@ -188,6 +188,37 @@ Likewise, `ExtendVote` can also be non-deterministic:
...
+Finally, note that Requirement 13 is local to process *p*, i.e.,
+no restrictions are made across different correct processes.
```
cason, line 217: I didn't fully get the point here. The requirement applies to all processes, but the time at which a transaction becomes invalid may change over processes, right?

Having saying so, my opinion here is that we should replace (clock) times by heights, which represent logical times. This would in particular enable removing this observation. While in fact different processes reach the same height at different times, from that height they should "forever" reject the transaction.

cason (commented):
I have left some comments, also addressed by other reviewers.

I think we should add a subsection, as this requirement has a different purpose than the other requirements. Essentially, it is not a _safety_ requirement, but more a _liveness_ or performance requirement. I am not really sure where it should fit in the current spec organization, but I am pretty sure we shouldn't mix it with the existent requirements.

spec/abci/abci++_app_requirements.md
sergio-mena: The PR is trying to provide requirements to the mempool without having to link it to consensus. If that's possible, I prefer it that way (more modular)

spec/abci/abci++_app_requirements.md
sergio-mena: > but the time at which a transaction becomes invalid may change over processes, right?

The current version does so. It is a property local to a process, as a first version. If a local requirement is enough, we shouldn't complicate it with a global version. A local requirement is easier to understand (esp, for people less familiar with distributed systems), and check or prove.

The question (to @hvanz and @otrack) is, is the requirement expressed in local terms enough?

sergio-mena:
> In every run, there is a height after which CheckTx always returns the same value.

@otrack this is slightly different (actually, weaker) than what the text of this PR is currently stating. But I trust you folks on what we need to require from the App, which I believe will be an assumption in the mempool spec. So, will adapt the text accordingly.

Will also review [the new text](cometbft/knowledge-base#9) shortly

spec/abci/abci++_app_requirements.md
```diff
@@ -188,36 +192,33 @@ Likewise, `ExtendVote` can also be non-deterministic:
...
+
+Requirement 13 allows us to use heights rather than time in our notation.
+Let *CheckTxCode<sub>tx,p,h</sub>* denote the result  code returned by *p*'s Application, via `ResponseCheckTx`,
+to CometBFT's `RequestCheckTx` call occurring while the Application is at height *h*
+and having transaction *tx* as parameter.
+Let predicate *OK(CheckTxCode)* denote whether *CheckTxCode* is `SUCCESS`.
+
+* Requirement 14 [`CheckTx`, eventual non-oscillation]: For any correct process *p*
+  and any transaction *tx*, there exists a height *h<sub>p,stable</sub>*
+  such that *OK(CheckTxCode<sub>tx,p,h<sub>p,stable</sub></sub>) = OK(CheckTxCode<sub>tx,p,h</sub>)*
```
cason, line 212: I think here is not *OK(CheckTxCode)* but only *CheckTxCode<sub>tx,p,h</sub>*.

spec/abci/abci++_app_requirements.md
```diff
@@ -188,6 +192,34 @@ Likewise, `ExtendVote` can also be non-deterministic:
...
+  any transaction *tx*, and any height *h<sub>p</sub>*,
+  `ResponseCheckTx` returns the same result code while *p* is at *h<sub>p</sub>*.
+
+Requirement 13 allows us to use heights rather than time in our notation.
+Let *CheckTxCode<sub>tx,p,h</sub>* denote the result  code returned by *p*'s Application, via `ResponseCheckTx`,
+to CometBFT's `RequestCheckTx` call occurring while the Application is at height *h*
+and having transaction *tx* as parameter.
+Let predicate *OK(CheckTxCode)* denote whether *CheckTxCode* is `SUCCESS`.
+
+* Requirement 14 [`CheckTx`, eventual non-oscillation]: For any correct process *p*
```
hvanz, line 210: One of the allowed cases here is that the transaction will become valid forever, which is a bit strange if it was invalid at some point before. I mean, in a trivial application such as a kv-store, we have transactions whose validity never changes; they are valid (or invalid) all the time. In a non-trivial application, a transaction is valid possibly during one or multiple periods of time/height. First the transaction is invalid, at some height it will become valid, then at a later height invalid again, then it's valid and committed to a block, at which point is invalid forever.

spec/abci/abci++_app_requirements.md
```diff
@@ -188,6 +192,34 @@ Likewise, `ExtendVote` can also be non-deterministic:
...
+  and any transaction *tx*, there exists a height *h<sub>p,stable</sub>*
+  such that *OK(CheckTxCode<sub>tx,p,h<sub>p,stable</sub></sub>) = OK(CheckTxCode<sub>tx,p,h</sub>)*
+  for any height *h > h<sub>p,stable</sub>*.
+
+Requirement 14, local to process *p*, ensures that
+a transaction will eventually stop oscillating between `CheckTx` success and failure
+if it stays in *p's* mempool for long enough.
+Thus, whilst being a requirement local to *p*, it ensures that
```
hvanz, line 218: > it ensures that a transaction will leave the mempool of *all* full nodes

We cannot ensure that from Requirement 14. This requirement allows two different processes to non-oscillate at different validity values. One process may stabilize on valid while the other on invalid, and not necessarily at the same time/height. To be able to talk about a global property, I think we cannot avoid talking about transactions committed to the chain, as in [INV4](https://github.com/cometbft/knowledge-base/blob/main/protocols/mempool-overview.md).

spec/abci/abci++_app_requirements.md
hvanz: Having said that, I could have missed some scenarios where the transaction becomes valid forever after having been invalid.

cason: > One of the allowed cases here is that the transaction will become valid forever, which is a bit strange if it was invalid at some point before. 

I don't think it is that strange. Assume that validity checks the existence of an account used by the transaction. While the account doesn't exist, the transaction is invalid. Once the account is created by another transaction, the transaction becomes valid. A process may receive the transaction while it is lagging behind in the blockchain, so that the transaction creating the account was not yet processed (e.g. delivered) to this process.

cason: >  I mean, in a trivial application such as a kv-store, we have transactions whose validity never changes; they are valid (or invalid) all the time.

In the case of a stateless verification, the validity of a transaction never changes. This means that the property above is always valid, as the verification is timeless/stateless.

spec/abci/abci++_app_requirements.md
cason: >  One process may stabilize on valid while the other on invalid, and not necessarily at the same time/height.

Why? Since we are using heights and the application is deterministic, the output at the beginning of a given height should be exactly the same.

The problem I see in those properties is that a statefull validation is dependent on the order at which transactions are verified. In particular, Req. 13 is hard to fulfill given this assumption. The most simpler examples regards duplication: if `tx` is checked twice against the application, for instance because the mempool fails preventing duplication, a stateful application should reject the second "copy" of the transaction.

cason: > To be able to talk about a global property, I think we cannot avoid talking about transactions committed to the chain

Agreed.

spec/abci/abci++_app_requirements.md
```diff
@@ -188,6 +192,34 @@ Likewise, `ExtendVote` can also be non-deterministic:
...
   e<sup>r</sup><sub>p</sub> = e<sup>r</sup><sub>q</sub>*
 
+### Mempool Connection Requirements
+
+Before defining our notation, we need to introduce a requirement for `CheckTx` to always
+yield the same result while *p*'s Application is at a given height.
+
+* Requirement 13 [`CheckTx`, same-height immutability]: For any correct process *p*,
+  any transaction *tx*, and any height *h<sub>p</sub>*,
+  `ResponseCheckTx` returns the same result code while *p* is at *h<sub>p</sub>*.
```
cason, line 202: This requirement is somehow preventing the application from implementing replay protection, right?

spec/abci/abci++_app_requirements.md
```diff
@@ -188,6 +192,34 @@ Likewise, `ExtendVote` can also be non-deterministic:
...
+
+Requirement 13 allows us to use heights rather than time in our notation.
+Let *CheckTxCode<sub>tx,p,h</sub>* denote the result  code returned by *p*'s Application, via `ResponseCheckTx`,
+to CometBFT's `RequestCheckTx` call occurring while the Application is at height *h*
+and having transaction *tx* as parameter.
+Let predicate *OK(CheckTxCode)* denote whether *CheckTxCode* is `SUCCESS`.
```
cason, line 208: This definition of OK is not used in this version of the document and could be removed.

spec/abci/abci++_app_requirements.md
```diff
@@ -188,6 +192,34 @@ Likewise, `ExtendVote` can also be non-deterministic:
...
+
+* Requirement 14 [`CheckTx`, eventual non-oscillation]: For any correct process *p*
+  and any transaction *tx*, there exists a height *h<sub>p,stable</sub>*
```
cason, line 211: Why the height is associated to a particular process? The advantage of using heights instead of time is to make the property global. In fact, since applications are deterministic, the outputs produced by different applications at the same height (which is a global parameter) should be the same.

spec/abci/abci++_app_requirements.md
```diff
@@ -188,6 +192,34 @@ Likewise, `ExtendVote` can also be non-deterministic:
...
+  such that *OK(CheckTxCode<sub>tx,p,h<sub>p,stable</sub></sub>) = OK(CheckTxCode<sub>tx,p,h</sub>)*
+  for any height *h > h<sub>p,stable</sub>*.
+
+Requirement 14, local to process *p*, ensures that
+a transaction will eventually stop oscillating between `CheckTx` success and failure
+if it stays in *p's* mempool for long enough.
+Thus, whilst being a requirement local to *p*, it ensures that
+a transaction will leave the mempool of *all* full nodes,
+either because it is expunged everywhere due to failing `CheckTx` calls,
+or because it stays valid long enough to be gossipped, proposed and decided.
```
cason, line 221: Echoing Hernán above, if the transaction is decided an application, by implementing replay protection, will start rejecting the same transaction.

cason (commented):
I am afraid we are walking in circles regarding the wording of these additional requirements. Should we try to schedule a synchronous meeting to discuss them?

spec/abci/abci++_app_requirements.md
sergio-mena: What you propose is stronger. I prefer the weakest form that is useful for the spec

spec/abci/abci++_app_requirements.md
sergio-mena: The sentence in the text going from a local property (the requirement) to a global one was to explain that we only need a local property to obtain a global one. Of course as you rightly point out in your comments, the global property does not _trivially_ follow from the local one. You need other things (such as consensus agreement) to _prove_ it, but I left that of of the text... maybe by trying to be concise, I ended up being unclear

spec/abci/abci++_app_requirements.md
sergio-mena: I don't think so. Replay protection involves different heights

spec/abci/abci++_app_requirements.md
sergio-mena: It is

spec/abci/abci++_app_requirements.md
sergio-mena: Yes, but I didn't want to make this requirement global. This is aimed at the app devs. I want the requirements addressed to them to be as imple and understandable as possible.
Then, assuming they respect those, we can deduce that a global property emerges (thanks to consensus agreement)... but we shouldn't confuse app devs by introducing all that "complication" here

spec/abci/abci++_app_requirements.md
sergio-mena: Yes, but that' beyond (my understanding) of the requirement to the app. The requirements should tell the app _how_ to do stuff, but rather _what_ they can and cannot do.

A similar example is the requirement on PrepProp and ProcProp, we don't says "you should of shouldn't used the past state for your checks"... that's up to the app devs to decide. We just say what the "limits" are to what they can and cannot do. To me replay protection is a similar thing. It's up to the app to decided whether, or how to implement it. We just need to set the limits, so that the mempool functions correctly.

spec/abci/abci++_app_requirements.md
hvanz:
```suggestion
Let predicate *OK(CheckTxCode<sub>tx,p,h</sub>)* denote whether *CheckTxCode<sub>tx,p,h</sub>* is `SUCCESS`.
```

hvanz (approved):

spec/abci/abci++_app_requirements.md
```diff
@@ -188,6 +192,39 @@ Likewise, `ExtendVote` can also be non-deterministic:
...
+If *CheckTxCodes<sub>tx,p,h</sub>* is a singleton set, i.e. the Application always returned
+the same result code in `ResponseCheckTx` while at height *h*,
+we define *CheckTxCode<sub>tx,p,h</sub>* as the singleton value of *CheckTxCodes<sub>tx,p,h</sub>*.
+If *CheckTxCodes<sub>tx,p,h</sub>* is not a singleton set, *CheckTxCode<sub>tx,p,h</sub>* is undefined.
+Let predicate *OK(CheckTxCode<sub>tx,p,h</sub>)* denote whether *CheckTxCode<sub>tx,p,h</sub>* is `SUCCESS`.
+
+* Requirement 13 [`CheckTx`, eventual non-oscillation]: There exists a boolean value *b*,
+  and a height *h<sub>stable</sub>*  such that
+  for any correct process *p*
+  and any transaction *tx*,
```
josef-widder, line 212: I think the order is not what is intended here. From the requirement as written here, all `tx` must have the same `b`. From the English text below it seems that you want: 
for all `tx`, exists `b`, exists `h`, forall `p`
This also would mean that from some height on, all processes agree on the outcome. This should also be highlighted.

sergio-mena: Fixed. Please take a look

otrack (approved):

spec/abci/abci++_app_requirements.md
josef-widder: Looks good. I still would add a comment on that all processes must agree on that stable value of `b`.

sergio-mena: @otrack Yes, I tried (1) first but, unlike in the notation you are using, in a notation consistent with the rest of the ABCI spec (all requirements) above I found it very hard to understand... so in the end I went for (2), and explain that, in practice, App devs can just enforce (1)... since as you say (1) implies (2).

We could adapt the whole spec to use a more advanced notation as the one you are using, but I'm not sure now's the moment to do that.

sergio-mena: > Looks good. I still would add a comment on that all processes must agree on that stable value of b.

I added sentence at the end

sergio-mena:
> As a general note, is there some documentation about the ABCI requirements? They are not easy to parse (for instance, this one is clearly difficult) and some documentation / base examples would help the programmer.

All ABCI requirement we have come up with so far are in this page. @nenadmilosevic95 also wrote some [extreme cases](https://github.com/cometbft/cometbft/blob/main/spec/abci/abci%2B%2B_example_scenarios.md) (from the point of view of the App) to help app developers understand "uncommon runs" their code needs to support

spec/abci/abci++_app_requirements.md
josef-widder: Thanks! The new sentence looks good. I think it makes sense that you didn't use the term agree, for the reasons @otrack mentioned above.

cason (approved):

PR #930, opened 2023-06-07
Implementation of ADR-108

nenadmilosevic95:
---

Related to https://github.com/cometbft/cometbft/issues/353.

The implementation (explained in https://github.com/cometbft/cometbft/pull/902) that is enhancing e2e tests so that we can check whether CometBFT respects ABCI++ grammar.
This version is not supporting vote extensions.

#### PR checklist

- [ ] Tests written/updated


test/e2e/tests/e2e_test.go
```diff
@@ -155,3 +161,68 @@ func fetchBlockChain(t *testing.T) []*types.Block {
...
+		}
+		reqs[i] = append(reqs[i], r)
+	}
+	return reqs, nil
+}
+
+func fetchNodeLogs(testnet e2e.Testnet, nodeName string) ([]byte, error) {
+	dir := filepath.Join(testnet.Dir, "docker-compose.yml")
+	c1 := exec.Command("docker-compose", "-f", dir, "logs")
+	c2 := exec.Command("grep", nodeName)
```
sergio-mena, line 197: It's probably easier to do the "grep" part internally with `string.Contains`

test/e2e/tests/e2e_test.go
```diff
@@ -155,3 +161,68 @@ func fetchBlockChain(t *testing.T) []*types.Block {
...
+			reqs = append(reqs, make([]*abci.Request, 0))
+			continue
+		}
+		r, err := app.GetABCIRequestFromString(line)
+		require.NoError(t, err)
+		// Ship the lines that does not contain abci request.
+		if r == nil {
+			continue
+		}
+		reqs[i] = append(reqs[i], r)
```
sergio-mena, line 189:
```suggestion
		reqs[len(reqs)-1] = append(reqs[len(reqs)-1], r)
```
This allows you to get rid of `i` and is more readable IMO

sergio-mena (commented):
This is my first pass at this PR. I haven't yet reviewed anything inside `test/e2e/pkg/grammar/`. I'll do it as part of the next pass.
Everything I got to looks good overall, no major problems seen.

go.mod
```diff
@@ -48,6 +48,7 @@ require (
 	github.com/cometbft/cometbft-db v0.7.0
 	github.com/cosmos/gogoproto v1.4.10
 	github.com/go-git/go-git/v5 v5.8.1
+	github.com/goccmack/goutil v1.2.3
```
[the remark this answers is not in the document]
nenadmilosevic95: This package is actually added by the `gogll` as a dependency of the autogenerated code. I'm not sure how we can get rid of it. Any idea?

test/e2e/Makefile
```diff
@@ -38,4 +38,8 @@ generator:
 runner:
 	go build -o build/runner ./runner
 
-.PHONY: all node docker generator runner 
+grammar:
```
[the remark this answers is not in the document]
nenadmilosevic95: Yes sure.

test/e2e/Makefile
```diff
@@ -38,4 +38,8 @@ generator:
...
+	gogll pkg/grammar/clean-start/abci_grammar_clean_start.md
```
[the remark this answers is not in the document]
nenadmilosevic95: Yes, it assumes this. I have explained in the ADR how one can install it  (`go get github.com/goccmack/gogll/v3`), however, we can also incorporate it inside the make if you think so.

test/e2e/Makefile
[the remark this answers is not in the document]
nenadmilosevic95: We need to change it in the ADR also.

test/e2e/tests/e2e_test.go
nenadmilosevic95: True! Thanks!

test/e2e/tests/e2e_test.go
nenadmilosevic95: True. I had a reason why I did like this instead but I forgot it. I will change it to `string.Contains` and maybe the reason will appear otherwise I will keep `string.Contains` :D

test/e2e/Makefile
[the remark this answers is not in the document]
sergio-mena: We can do `go run ....` so the use doesn't need to worry about installation

go.mod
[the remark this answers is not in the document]
sergio-mena: In that case, let's leave it there for the time being

test/e2e/pkg/grammar/checker.go
```diff
@@ -0,0 +1,212 @@
...
+		errors = g.verifyRecovery(execution)
+	}
+	if errors == nil {
+		return true, nil
+	}
+	return false, fmt.Errorf("%v\nFull execution:\n%v", g.combineErrors(errors, g.cfg.NumberOfErrorsToShow), g.addHeightNumbersToTheExecution(execution))
+}
+
+// verifyCleanStart verifies if a specific execution is a valid clean-start execution.
+func (g *GrammarChecker) verifyCleanStart(execution string) []*Error {
```
sergio-mena, line 149: This function and the following are almost identical, IIUC, only differences are:
* the lexer you create
* the parser you call
* part of the error message

Can we factor out these two functions into an internal one accepting a lexer creation functor, a `Parse` functor, and a string "clean-start"/"recovery" to add to the error message?

test/e2e/pkg/grammar/checker.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	return errors
+}
+
+// addHeightNumbersToTheExecution adds height numbers to the execution. This is used just when printing the execution so we can find the height with error more easily.
+func (g *GrammarChecker) addHeightNumbersToTheExecution(execution string) string {
+	heights := strings.Split(execution, "\n")
+	s := ""
+	for i, l := range heights {
+		if l != "" {
```
sergio-mena, line 195: If this condition is true, won't we see a gap in heights? Is that what we intend?

test/e2e/pkg/grammar/checker.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// combineErrors combines at most n errors in one.
+func (g *GrammarChecker) combineErrors(errors []*Error, n int) error {
+	s := ""
+	for i, e := range errors {
+		if i == n {
+			break
+		}
+		s = fmt.Sprintf("%v%v\n", s, e)
```
sergio-mena, line 209:
```suggestion
		s += e.String() + "\n"
```
Just a suggestion

sergio-mena: Another option: `s += fmt.Sprintln(e)`

test/e2e/pkg/grammar/checker_test.go
```diff
@@ -0,0 +1,103 @@
...
+	processProposal = &abci.Request{Value: &abci.Request_ProcessProposal{ProcessProposal: &abci.RequestProcessProposal{}}}
+)
+
+const (
+	CLEAN_START = true
+	PASS        = true
+	FAIL        = false
+)
+
+var tests = []Test{
```
sergio-mena, line 33: This is OK, although it tends to be defined inside the test case that uses them, for readability

test/e2e/pkg/grammar/checker_test.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+}
+
+func TestFilterLastHeight(t *testing.T) {
+	reqs := []*abci.Request{initChain, finalizeBlock, commit}
+	checker := NewGrammarChecker(DefaultConfig())
+	rr, n := checker.filterLastHeight(reqs)
+	if len(reqs) != len(rr) || n != 0 {
+		t.Errorf("FilterLastHeight filtered %v ABCI calls, expected %v\n", n, 0)
+	}
```
sergio-mena, line 96:
```suggestion
	require.Equal(t, len(reqs),len(rr), "FilterLastHeight check failed with filtered ABCI calls")
	require.Zero(t, n, "Check failed with filtered ABCI calls")
```
Please use this pattern all over your unit tests (actually, we usually don't need the message parameter, as the framework will always show the line where the failing check was, and the values that failed the check)

test/e2e/pkg/grammar/clean-start/abci_grammar_clean_start.md
```diff
@@ -0,0 +1,33 @@
+```
+package "github.com/cometbft/cometbft/test/e2e/pkg/grammar/clean-start"
+
```
sergio-mena, line 3: Is it possible to add a comment with a link to the original grammar? (same for the recovery grammar)

test/e2e/pkg/grammar/clean-start/abci_grammar_clean_start.md
```diff
@@ -0,0 +1,33 @@
...
+
+Start : CleanStart ;
```
sergio-mena, line 5: Open-ended question: since you had to split the original grammar into two, isn't that a sign that the original grammar (in the spec) should be similarly split? (something to reflect on)

test/e2e/pkg/grammar/clean-start/lexer/lexer.go
```diff
@@ -0,0 +1,1141 @@
+
```
sergio-mena, line 1: The `lexer`, `parser`, `sppf`, and `token` libraries all contain _only_ auto-generated code, which should be opaque to the user. Is this true and accurate?

If yes, could we:
1) Create a directory `grammar-auto` at this level and move all four dirs inside? (the intention is to show a user navigating files that they shouldn't inspect inside as they'll only find auto-generated code.
2) Add a small README.md at this level, explaining the only methods that are relevant outside this directory (the ones you use in `checker.go`

I'm really not sure how to go about 1). Other options would be:

* just move `abci_grammar_clean_start.md` up one level, and rename directory `clean-start` to `abci_grammar_clean_start`
* or the other way around: move `abci_grammar_clean_start.md` up one level and rename it to `clean-start`

The point is for the directory structure to give the user a hint of "stop opening subfolders, this is all generated code"

sergio-mena (commented):
Reviewed the grammar part.

test/e2e/pkg/grammar/checker.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	return reqs[:pos+1], cnt
+}
+
+// getRequestTerminal returns a value of a corresponding terminal in the ABCI grammar for a specific request.
+func (g *GrammarChecker) getRequestTerminal(req *abci.Request) string {
+	// req.String() produces an output like this "init_chain:<time:<seconds:-62135596800 > >"
+	// we take just the part before the ":" (init_chain, in previous example) for each request
+	s := req.String()
+	t := strings.Split(s, ":")[0]
```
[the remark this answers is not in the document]
nenadmilosevic95: How should we fail?

test/e2e/pkg/grammar/checker.go
```diff
@@ -103,9 +103,11 @@ func (g *GrammarChecker) filterLastHeight(reqs []*abci.Request) ([]*abci.Request
...
 	// req.String() produces an output like this "init_chain:<time:<seconds:-62135596800 > >"
 	// we take just the part before the ":" (init_chain, in previous example) for each request
-	s := req.String()
-	t := strings.Split(s, ":")[0]
-	return t
+	parts := strings.Split(req.String(), ":")
+	if len(parts) >= 2 && len(parts[0]) > 0 {
+		return parts[0]
+	}
+	panic(fmt.Errorf("abci.Request doesn't have the expected string format: %v", req.String()))
```
sergio-mena, line 110:
```suggestion
	if len(parts) < 2 || len(parts[0]) == 0 {
		panic(fmt.Errorf("abci.Request doesn't have the expected string format: %v", req.String()))
	}
	return parts[0]
```
Usually, the un-happy paths tend to be inside sanity checks, and the happy path is when those checks don't hold. Makes the code more readable by others

test/e2e/app/log.go
```diff
@@ -0,0 +1,45 @@
+package app
```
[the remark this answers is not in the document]
sergio-mena: Do you agree with the name change suggested above?

test/e2e/app/log.go
```diff
@@ -0,0 +1,45 @@
...
+	b, err := proto.Marshal(req)
+	if err != nil {
+		return "", err
+	}
+	reqStr := base64.StdEncoding.EncodeToString(b)
+	s := ABCI_REQ + reqStr + ABCI_REQ
+	return s, nil
+}
+
+// GetABCIRequestFromString parse string and try to get a string of a Request created by GetRequestString.
```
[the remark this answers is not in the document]
sergio-mena: Do you agree with this suggestion?

test/e2e/app/log.go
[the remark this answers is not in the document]
nenadmilosevic95: Yes sure. I will do this.

test/e2e/app/log.go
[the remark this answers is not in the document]
nenadmilosevic95: I put "try to get" because it may return an error also.

test/e2e/app/log.go
```diff
@@ -0,0 +1,45 @@
...
+func GetABCIRequestFromString(s string) (*abci.Request, error) {
+	if !strings.Contains(s, ABCI_REQ) {
+		return nil, nil
+	}
+	parts := strings.Split(s, ABCI_REQ)
+	if len(parts) != 3 {
+		return nil, fmt.Errorf("String %v passed to GetRequestFromString does not have a good format!\n", s)
+	}
+	req := &abci.Request{}
```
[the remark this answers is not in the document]
nenadmilosevic95: I will add condition " len(parts[1] == 0 ".

test/e2e/app/log.go
```diff
@@ -0,0 +1,45 @@
...
+	reqStr := parts[1]
+	b, err := base64.StdEncoding.DecodeString(reqStr)
+	if err != nil {
+		return nil, fmt.Errorf("String %v cannot be decoded to bytes.\n Error: %v\n ", s, err.Error())
```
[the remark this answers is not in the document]
nenadmilosevic95: I will do the first

test/e2e/networks/simple.toml
```diff
@@ -1,3 +1,5 @@
+abci_tests_enabled = true
```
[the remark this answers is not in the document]
nenadmilosevic95: ok

test/e2e/pkg/testnet.go
```diff
@@ -206,6 +208,7 @@ func NewTestnetFromManifest(manifest Manifest, file string, ifd InfrastructureDa
 		if v == "" {
 			v = localVersion
 		}
+		//abciCalls := loadAbciCalls(testnet.Dir)
```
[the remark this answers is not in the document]
nenadmilosevic95: Removed!

test/e2e/app/log_abci.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// GetABCIRequestFromString parse string and try to get a string of a Request created by GetRequestString.
+func GetABCIRequestFromString(s string) (*abci.Request, error) {
+	if !strings.Contains(s, ABCI_REQ) {
+		return nil, nil
+	}
+	parts := strings.Split(s, ABCI_REQ)
+	if len(parts) != 3 || len(parts[1]) == 0 {
+		return nil, fmt.Errorf("String %v passed to GetRequestFromString does not have a good format!\n", s)
```
sergio-mena, line 31:
```suggestion
		return nil, fmt.Errorf("String %q passed to GetRequestFromString does not have a good format!\n", s)
```

test/e2e/pkg/grammar/README.md
```diff
@@ -0,0 +1,14 @@
+# The ``grammar`` folder structure
+
+This folder has two subdirectories (`clean-start` and `recovery`), `checker.go` and `checker_test.go` files. 
+
+Both `clean-start` and `recovery` folders have the same structure. They contain files with the context-free grammars 
+`abci_grammar_clean_start.md` and `abci_grammar_recovery.md`, respectively. In addition, they have a subdirectory, 
+`grammar-auto`, where the auto-generated code is stored. Namely, running `make grammar-gen` from the `e2e` directory will 
+invoke the `gogll` program that will use `abci_grammar_clean_start.md` and `abci_grammar_recovery.md` grammars as 
+input and produce lexer and parser for them. All the code `gogll` generates, is stored in the respected `grammar-auto` folder.
```
sergio-mena, line 9:
```suggestion
input and produce lexer and parser for them. All the code `gogll` generates, is stored in the mentioned `grammar-auto` folder. 
```

test/e2e/pkg/grammar/checker.go
nenadmilosevic95: This condition was actually added because the last element in "heights" is "".

test/e2e/pkg/grammar/checker.go
```diff
@@ -0,0 +1,215 @@
...
+
+// addHeightNumbersToTheExecution adds height numbers to the execution. This is used just when printing the execution so we can find the height with error more easily.
+func (g *GrammarChecker) addHeightNumbersToTheExecution(execution string) string {
+	heights := strings.Split(execution, "\n")
+	s := ""
+	for i, l := range heights {
+		if l != "" {
+			s = fmt.Sprintf("%v%v: %v\n", s, i, l)
+		}
+	}
```
sergio-mena, line 201:
```suggestion
	// assert that last element of heights == "" (if with error) 
	for i, l := range heights[:len(heights)-1] {
		s = fmt.Sprintf("%v%v: %v\n", s, i, l)
	}
```

test/e2e/pkg/grammar/checker.go
sergio-mena: After exploring some solutions here, we've decided to leave this code as it is... as the alternatives end up adding unjustified complexity

test/e2e/pkg/grammar/checker.go
```diff
@@ -0,0 +1,215 @@
...
+		errors = append(errors, e)
+	}
+	return errors
+}
+
+// addHeightNumbersToTheExecution adds height numbers to the execution. This is used just when printing the execution so we can find the height with error more easily.
+func (g *GrammarChecker) addHeightNumbersToTheExecution(execution string) string {
+	heights := strings.Split(execution, "\n")
+	s := ""
+	// assert that last element of heights == "" (if with error)
```
sergio-mena, line 197: Please change this comment by an "if" that handles the case where the last element is not ""

sergio-mena (approved):
Thanks @nenadmilosevic95 for your efforts. This is ready for prime time :-)

PR #931, opened 2023-06-07
Reduces log clutter

### LINKED ISSUE #907: Consensus, 253 logs

lasarojc:
Closes #907
This PR downgrades some `Info` logs that are relevant to developers but shouldn't matter to validators to `Debug` logs.
Most of these logs are on the `consensus` package.

On the positive side, there is less information overhead to operators and less storage is needed for saving logs during normal operation.
On the negative side, if debugging is needed, the log level must be changed after the need is detected, which may be too late to capture rare occurrences of bugs.

During your reviews, please let me know if any of the demoted logs have been useful in previous debug sessions and if you think they should not be demoted. 

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/state/execution.go
+++ b/state/execution.go
@@ -233,7 +233,7 @@ func (blockExec *BlockExecutor) ApplyBlock(
 		return state, fmt.Errorf("expected tx results length to match size of transactions in block. Expected %d, got %d", len(block.Data.Txs), len(abciResponse.TxResults))
 	}
 
-	blockExec.logger.Debug("executed block", "height", block.Height, "app_hash", abciResponse.AppHash)
+	blockExec.logger.Info("executed block", "height", block.Height, "app_hash", abciResponse.AppHash)
 
 	fail.Fail() // XXX
 
--- a/state/txindex/indexer_service.go
+++ b/state/txindex/indexer_service.go
@@ -99,7 +99,7 @@ func (is *IndexerService) OnStart() error {
 						return
 					}
 				} else {
-					is.Logger.Info("indexed block events", "height", height)
+					is.Logger.Debug("indexed block events", "height", height)
 				}
 
 				if err = is.txIdxr.AddBatch(batch); err != nil {


consensus/reactor.go
```diff
@@ -72,7 +72,8 @@ func NewReactor(consensusState *State, waitSync bool, options ...ReactorOption)
 // OnStart implements BaseService by subscribing to events, which later will be
 // broadcasted to other peers and starting state if we're not in block sync.
 func (conR *Reactor) OnStart() error {
-	conR.Logger.Info("Reactor ", "waitSync", conR.WaitSync())
+	conR.Logger.Debug("Consensus OnStart")
+	conR.Logger.Debug("waitSync", conR.WaitSync())
```
thanethomson, line 76: Seems like it should still be a single log line, should it not?

consensus/reactor.go
```diff
@@ -105,7 +107,7 @@ func (conR *Reactor) OnStop() {
 // SwitchToConsensus switches from block_sync mode to consensus mode.
 // It resets the state, turns off block_sync, and starts the consensus state-machine
 func (conR *Reactor) SwitchToConsensus(state sm.State, skipWAL bool) {
-	conR.Logger.Info("SwitchToConsensus")
+	conR.Logger.Debug("Consensus SwitchToConsensus")
```
thanethomson, line 110: This seems like it should still be at info level to assist with debugging (assuming that most operators will always have their logs at info level and above, and never at debug level).

thanethomson (commented):
While I agree with many of these changes (especially the ones that've been promoted from info to error) I'm not 100% sure whether operators rely on some of the logs that're being dropped to debug level. Also, operators have asked us for _more_ information at info level from consensus (and much less from P2P). It may be a good idea to chat with some of them to get an idea of what they want from consensus (perhaps in a follow-up PR?).

consensus/replay.go
```diff
@@ -489,6 +488,10 @@ func (h *Handshaker) replayBlocks(
 			return nil, err
 		}
 
+		if i == firstBlock || i%1000 == 0 || i == finalBlock {
+			h.logger.Info("executed blocks up to", "height", block.Height, "app_hash", appHash)
```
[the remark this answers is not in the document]
mzabaluev: In Rust frameworks like tracing and in general (e.g. in journald), the convention is not to have the message capitalized. In commonly used log printout formats, the message is prefixed with ancillary information (timestamp, severity, etc.), so the capitalization will stick out.

consensus/reactor.go
cason: Agreed. This message should be printed only once, which for me is a clear case for Info level.

consensus/reactor.go
```diff
@@ -650,12 +654,12 @@ func (conR *Reactor) gossipDataForCatchup(logger log.Logger, rs *cstypes.RoundSt
...
 		blockMeta := conR.conS.blockStore.LoadBlockMeta(prs.Height)
 		if blockMeta == nil {
-			logger.Error("Failed to load block meta", "ourHeight", rs.Height,
+			logger.Error("Failed to load block meta", "rs.height", rs.Height,
 				"blockstoreBase", conR.conS.blockStore.Base(), "blockstoreHeight", conR.conS.blockStore.Height())
 			time.Sleep(conR.conS.config.PeerGossipSleepDuration)
 			return
 		} else if !blockMeta.BlockID.PartSetHeader.Equals(prs.ProposalBlockPartSetHeader) {
-			logger.Info("Peer ProposalBlockPartSetHeader mismatch, sleeping",
+			logger.Error("Peer ProposalBlockPartSetHeader mismatch, sleeping",
```
cason, line 662: I would not put this an error. By the way we set those fields referring to a peer (they are hints), we cannot say that the peer is misbehaving when sending us the wrong block.

consensus/replay.go
```diff
@@ -137,7 +137,7 @@ func (cs *State) catchupReplay(csHeight int64) error {
 	}
 	defer gr.Close()
 
-	cs.Logger.Info("Catchup by replaying consensus messages", "height", csHeight)
+	cs.Logger.Debug("Catchup by replaying consensus messages", "height", csHeight)
```
cason, line 140: This again is a message that should be printed only once, Info level looks more appropriate.

consensus/state.go
```diff
@@ -1603,7 +1603,7 @@ func (cs *State) enterCommit(height int64, commitRound int32) {
 	// If we don't have the block being committed, set up to get it.
 	if !cs.ProposalBlock.HashesTo(blockID.Hash) {
 		if !cs.ProposalBlockParts.HasHeader(blockID.PartSetHeader) {
-			logger.Info(
+			logger.Debug(
```
cason, line 1606: This looks to me an unlikely situation. I would preserve info, even warning if we have that level.

consensus/ticker.go
```diff
@@ -121,7 +121,7 @@ func (t *timeoutTicker) timeoutRoutine() {
 			t.timer.Reset(ti.Duration)
 			t.Logger.Debug("Scheduled timeout", "dur", ti.Duration, "height", ti.Height, "round", ti.Round, "step", ti.Step)
 		case <-t.timer.C:
-			t.Logger.Info("Timed out", "dur", ti.Duration, "height", ti.Height, "round", ti.Round, "step", ti.Step)
+			t.Logger.Debug("Timed out", "dur", ti.Duration, "height", ti.Height, "round", ti.Round, "step", ti.Step)
```
cason, line 124: We need to make timeouts logged when they are processed, not when they are produced. So while I agree with this change of level, we need to turn into Info when a timeout produce an operation state change.

state/execution.go
```diff
@@ -696,7 +696,7 @@ func ExecCommitBlock(
 		return nil, fmt.Errorf("expected tx results length to match size of transactions in block. Expected %d, got %d", len(block.Data.Txs), len(resp.TxResults))
 	}
 
-	logger.Info("executed block", "height", block.Height, "app_hash", resp.AppHash)
+	logger.Debug("executed block", "height", block.Height, "app_hash", resp.AppHash)
```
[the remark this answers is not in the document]
cason: Agreed.

cason (commented):
While this is a relevant issue to address, I wonder whether we need some more information in order to decide which logs and levels are relevant.

For instance, we should have a document displaying all log messages produced in Info and Debug level for a successful ("good") height, then when something odd happens (there are several situations to cover). With this overall output or representation we could discuss which logs are redundant, unclear, and which information is indeed missing.

Finally, I would rename the PR to make clear it refers to the consensus module/reactor.

consensus/replay.go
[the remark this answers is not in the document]
adizere: > the convention is not to have the message capitalized. In commonly used log printout formats, the message is prefixed with ancillary information (timestamp, severity, etc.), so the capitalization will stick out.

+1

lasarojc:
Relates to tendermint/tendermint#6987

consensus/state.go
```diff
@@ -1899,7 +1899,7 @@ func (cs *State) defaultSetProposal(proposal *types.Proposal) error {
 		cs.ProposalBlockParts = types.NewPartSetFromHeader(proposal.BlockID.PartSetHeader)
 	}
 
-	cs.Logger.Info("received proposal", "proposal", proposal)
+	cs.Logger.Debug("received proposal", "proposal", proposal)
```
sergio-mena, line 1902: This turned out to be crucial to day, so please leave it in Info

state/execution.go
```diff
@@ -220,7 +220,7 @@ func (blockExec *BlockExecutor) ApplyBlock(
 		return state, err
 	}
 
-	blockExec.logger.Info(
+	blockExec.logger.Debug(
```
sergio-mena, line 223: This one also was useful today (well, actually its `v0.34.x` counterpart: "executed block") to delimit the logs that were coming from the block execution itself. I think it should stay in Info

state/execution.go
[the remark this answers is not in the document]
sergio-mena: I would change the message to be similar to "finalized block" above (it is its "handshake" counterpart), something like:
```
	blockExec.logger.Info(
		"finalized block (replay blocks)",
		"height", block.Height,
		"num_txs_res", len(resp.TxResults),
		"num_val_updates", len(resp.ValidatorUpdates),
		"block_app_hash", fmt.Sprintf("%X", resp.AppHash),
	)
```

state/txindex/indexer_service.go
```diff
@@ -99,7 +99,7 @@ func (is *IndexerService) OnStart() error {
 						return
 					}
 				} else {
-					is.Logger.Info("indexed block events", "height", height)
+					is.Logger.Debug("indexed block events", "height", height)
```
sergio-mena, line 102: This one I saw it today quite often in the logs but I didn't really find it useful. I'd agree on "demoting" it to Debug

sergio-mena (commented):
We've had a troubleshooting session today, trying to find out why, and under which conditions, there were frequent heights decided in round 1.
I've been exposed to the consensus logs for that troubleshooting session.
Since I have the logs I found (and didn't find) useful fresh in my mind, I decided to come back to this PR and provide first-hand feedback.

lasarojc:
Thanks, @sergio-mena 
I should resume working on this PR soon.

melekes:
@lasarojc any plans to resume this?

melekes:
Let's revisit this PR when we have them then.

PR #933, opened 2023-06-08
e2e tests are being skipped; `INFRASTRUCTURE-*` bug; portGen inconsistency bug

### LINKED ISSUE #929: e2e tests are being skipped and not executed due to the bug.
## Bug Report

The e2e tests are being skipped and not executed due to the bug. Basically, added logic that should differentiate between different environments (docker, digital ocean, etc.) has a bug that causes the tests to be skipped. 

### Setup

**CometBFT version** (use `cometbft version` or `git rev-parse --verify HEAD` if installed from source): main a09f5d33ecd8846369b93cae9063291eb8abc3a0

**ABCI app** (name for built-in, URL for self-written if it's publicly available): default

**Environment**:
- **OS** (e.g. from /etc/os-release): N/A
- **Install tools**: N/A
- **Others**: N/A

**node command runtime flags**: N/A

### What happened?

While trying to run new abci tests, I realised that the tests were not being executed. 

The test I was expecting to fail was passing. After a closer look, I realised that the tests were skipped. However, the `runner` 
shows the output as if all tests are being executed and all pass.  

### How to reproduce it

Just add `t.Error("Hello")` in any test and start the runner with any testnet. Apart from `Hello` error message, it will show also 
`INFRASTRUCTURE_DATA not set and INFRASTRUCTURE_TYPE is not docker` message. If you see where in the code this is printed you will find this part of the code in the `test/e2e/tests/e2e_test.go` lines 76-80. 

```go

ifdType := os.Getenv("INFRASTRUCTURE_DATA")
ifdFile := os.Getenv("INFRASTRUCTURE_FILE")
if ifdFile == "" && ifdType != "docker" {
	t.Skip("INFRASTRUCTURE_DATA not set and INFRASTRUCTURE_TYPE is not docker")
}
 
```

The setting of variables and conditions needs to be revisited! In addition, to this specific part of the code, we should recheck all the places in the code related to the variables `INFRASTRUCTURE_DATA`, `INFRASTRUCTURE_FILE` and `INFRASTRUCTURE_TYPE`.

sergio-mena:
Closes #929

This PR fixes three problems:

* When there is a problem in env variables `INFRASTRUCTURE_FILE` and `INFRASTRUCTURE_TYPE`, we shouldn't _skip_ but fail. Otherwise we just hide a problem we should fix
* The code introducing `INFRASTRUCTURE_FILE` and `INFRASTRUCTURE_TYPE` (#796) had a couple of typos and so, didn't work (we didn't see this because of the first bullet, and so, e2e tests were not run)
* There was an ordering inconsistency problem in manifest- and testnet-related code, which caused the e2e tests to fail

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/test/e2e/pkg/infrastructure.go
+++ b/test/e2e/pkg/infrastructure.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"sort"
 )
 
 const (
@@ -43,6 +44,16 @@ type InstanceData struct {
 	Port         uint32 `json:"port"`
 }
 
+func sortNodeNames(m Manifest) []string {
+	// Set up nodes, in alphabetical order (IPs and ports get same order).
+	nodeNames := []string{}
+	for name := range m.Nodes {
+		nodeNames = append(nodeNames, name)
+	}
+	sort.Strings(nodeNames)
+	return nodeNames
+}
+
 func NewDockerInfrastructureData(m Manifest) (InfrastructureData, error) {
 	netAddress := dockerIPv4CIDR
 	if m.IPv6 {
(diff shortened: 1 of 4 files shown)


thanethomson (approved):
Good catch!

PR #974, opened 2023-06-14
changelog: Import entries for v0.34.28, v0.34.29, v0.37.1 and v0.37.2 to `main`

thanethomson:
Brings all the changelog entries from the latest v0.34.x and v0.37.x entries to `main` for ease of reference.

:book: [CHANGELOG rendered](https://github.com/cometbft/cometbft/blob/thane/update-changelog/CHANGELOG.md)

I've noticed that unclog builds all of the v0.34.x entries after all of the v0.37.x entries, which may or may not be the behaviour that we actually want. Do we want changelog entries sorted by semantic version, or by release date? If we need to sort by release date, we'll need to modify unclog accordingly.

---

#### PR checklist

- [ ] Tests written/updated


sergio-mena:
> Do we want changelog entries sorted by semantic version, or by release date?

A priori, I'd say semantic version is more meaningful to the user, as it follows the logic of the code itself, dates don't have that feature (as we maintain -- and release from -- several branches).
Anyway not 100% sure I'm not missing something here.

sergio-mena (approved):
LGTM

PR #975, opened 2023-06-14
Rename ABCI enums and values to satisfy buf guidelines

### LINKED ISSUE #736: proto: enum definitions should follow the `*_UNSPECIFIED` guideline

mzabaluev:
Resolves #736

Additional sympathetic changes:

* The `ENUM_VALUE_PREFIX` lint is satisfied as well.
* Moved enum definitions out of message definitions because nesting is not recommended by buf style and the generated names in Go were messy.

---

#### PR checklist

- [x] Tests written/updated


test/e2e/docker/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # We need to build in a Linux environment to support C libraries, e.g. RocksDB.
 # We use Debian instead of Alpine, so that we can use binary database packages
 # instead of spending time compiling them.
-FROM golang:1.20
+FROM golang:1.20-bullseye
```
sergio-mena, line 4: Do we want to leave this change in the branch?

sergio-mena (approved):
LGTM, thanks for this!
We still need to decide what we'll do with `CheckTx` enums

test/e2e/docker/Dockerfile
mzabaluev: I will rebase it away.

proto/cometbft/abci/v4/types.proto
```diff
@@ -0,0 +1,233 @@
...
+
+enum CheckTxType {
+  option (gogoproto.goproto_enum_prefix) = false;
+
+  // The non-conventional name for the zero value is preserved
+  // due to history of use.
+  // buf:lint:ignore ENUM_ZERO_VALUE_SUFFIX
+  CHECK_TX_TYPE_NEW = 0;
+  CHECK_TX_TYPE_RECHECK = 1;
+}
```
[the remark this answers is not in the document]
mzabaluev: This would be a clean, if disruptive, way to fix the forward compatibility issue.
With domain types e.g. in tendermint-rs, we can also convert seamlessly between the old and the new encodings.

api/cometbft/abci/v4/types.pb.go
```diff
@@ -0,0 +1,7003 @@
...
+		FullMethod: "/cometbft.abci.v4.ABCI/FinalizeBlock",
+	}
+	handler := func(ctx context.Context, req interface{}) (interface{}, error) {
+		return srv.(ABCIServer).FinalizeBlock(ctx, req.(*RequestFinalizeBlock))
+	}
+	return interceptor(ctx, in, info, handler)
+}
+
+var _ABCI_serviceDesc = grpc.ServiceDesc{
+	ServiceName: "cometbft.abci.v4.ABCI",
```
mzabaluev, line 2107: As can be seen here again, the new package introduces a new gRPC service path, so wire compatibility is not even necessary to maintain for the gRPC service.

On the other hand, maybe it's worth doing what @thanethomson did in #818 and put the service definition into a dedicated package namespace, so that it could be versioned separately? We don't really need the new service path as long as all request and response messages valid with the old service can still be compatibly represented.

test/e2e/docker/Dockerfile
sergio-mena: Nevermind, check #984

test/e2e/app/app.go
```diff
@@ -503,7 +503,7 @@ func (app *Application) VerifyVoteExtension(_ context.Context, req *abci.Request
 
 	app.logger.Info("verified vote extension value", "height", req.Height, "vote_extension", fmt.Sprintf("%x", req.VoteExtension[:4]), "num", num)
 	return &abci.ResponseVerifyVoteExtension{
-		Status: abci.VERIFY_VOTE_EXTENSION_STATUS_REJECT,
```
sergio-mena: Good catch!

sergio-mena:
As discussed, waiting for the `CheckTx` result code changes before re-approving

proto/cometbft/abci/v4/types.proto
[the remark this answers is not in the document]
mzabaluev: I have regularized `CheckTxType` in the proposed way, and changed the fields of `RequestCheckTx` to ensure the clean break.

mzabaluev:
Note that I have not changed the enum names or definitions in the `tendermint.*` packages, which were following the latest protocol changes up to this point. I'm not sure if we want to break those, and how do we reconcile the new `CheckTxType` with those protobufs.

proto/cometbft/abci/v4/types.proto
```diff
@@ -53,19 +53,23 @@ message Request {
...
   option (gogoproto.goproto_enum_prefix) = false;
 
-  // The non-conventional name for the zero value is preserved
-  // due to history of use.
-  // buf:lint:ignore ENUM_ZERO_VALUE_SUFFIX
-  CHECK_TX_TYPE_NEW = 0;
-  CHECK_TX_TYPE_RECHECK = 1;
+  CHECK_TX_TYPE_UNKNOWN = 0;
+  CHECK_TX_TYPE_NEW = 1;
+  CHECK_TX_TYPE_RECHECK = 2;
```
sergio-mena, line 66: Is it possible to:
* keep RECHECK with the same number as before (still breaking change, but one less element for devs to take care of when adapting their code)
* Rename NEW to something like CHECK, so reduce the possibility of confusion when referring to either (or both) in discussions

mzabaluev: > Would it be possible to:
> 
>     * keep RECHECK with the same number as before (still breaking change, but one less element for devs to take care of when adapting their code)

I don't think it will matter when adapting code, unless someone's been using literal numbers to represent these codes. The symbolic constants are renamed anyway. But the order of numbers does not matter, so why not.

>     * Rename NEW to something like CHECK, to reduce the possibility of confusion when referring to either (or both) in discussions

Good idea, I overlooked it from your proposal. Will do.

proto/cometbft/abci/v4/types.proto
```diff
@@ -62,8 +62,8 @@ enum CheckTxType {
   option (gogoproto.goproto_enum_prefix) = false;
 
   CHECK_TX_TYPE_UNKNOWN = 0;
-  CHECK_TX_TYPE_NEW = 1;
-  CHECK_TX_TYPE_RECHECK = 2;
+  CHECK_TX_TYPE_CHECK = 2;
+  CHECK_TX_TYPE_RECHECK = 1;
```
sergio-mena, line 66:
```suggestion
  CHECK_TX_TYPE_RECHECK = 1;
  CHECK_TX_TYPE_CHECK = 2;
```

proto/cometbft/abci/v4/types.proto
mzabaluev: Changed in cbf1e802b5b9b13ad68963dae3266d810c87c447

mzabaluev:
The tests pass locally for me. Seems like a flaky test?

sergio-mena (approved):
Re-approving

test/e2e/docker/Dockerfile
mzabaluev: I'm leaving this in so as to not go through another merge of main into `feature/proto-upgrade` just for this.

PR #983, opened 2023-06-16
Add `CMT_HOME` (or remove it?)

### LINKED ISSUE #982: `CMT_HOME` is not picked up by `cometbft init`
### Setup

**CometBFT version** : `main`

**ABCI app** (name for built-in, URL for self-written if it's publicly available): N/A

**Environment**:
- **OS** (e.g. from /etc/os-release): OSX
- **Install tools**: N/A
- **Others**: N/A

### Config

N/A

### What happened?

Our docs state that both `CMTHOME` and `CMT_HOME` are valid variable names to set the root directory where `data` and `config` will live.

Command `init` does not pick up `CMT_HOME`:

```
% echo $CMT_HOME 

% export CMT_HOME=/tmp/my_cmt
% echo $CMT_HOME             
/tmp/my_cmt
% mkdir $CMT_HOME
% build/cometbft init validator
I[2023-06-16|17:02:35.885] Generated private validator                  module=main keyFile=/Users/sergio/.cometbft/config/priv_validator_key.json stateFile=/Users/sergio/.cometbft/data/priv_validator_state.json
I[2023-06-16|17:02:35.885] Generated node key                           module=main path=/Users/sergio/.cometbft/config/node_key.json
I[2023-06-16|17:02:35.885] Generated genesis file                       module=main path=/Users/sergio/.cometbft/config/genesis.json
% 
```

Check the logs: the root directory is the default one (`.cometbft` in my home dir)

`init` command should pick up the value of `CMT_HOME` (`/tmp/my_cmt` in the example above)

### How to reproduce it

See above

### Logs

See above

N/A

No

sergio-mena:
Closes #982

Added `CMT_HOME` everywhere `CMTHOME` is used.

### Notes to reviewers

This could be fixed the opposite way, by removing the only reference to `CMT_HOME` in the code, and also the reference in `UPGRADING.md` (two lines of code).

However, the reference in `UPGRADING.md`, which is part of our documentation, is already present in `v0.34.x` (not in `v0.37.x` though!). That's why this PR introduces `CMT_HOME` to work in equal conditions as `CMTHOME`.

If reviewers lean toward removing `CMT_HOME` from the doc in `v0.34.x` (and unreleased `v0.38.x` and `main`), I can do it easily.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -117,7 +117,7 @@ on instead of `~/.tendermint`.
 ### Environment variables
 
 The environment variable prefixes have now changed from `TM` to `CMT`. For
-example, `TMHOME` or `TM_HOME` become `CMTHOME`.
+example, `TMHOME` becomes `CMTHOME`.
 
 We have implemented a fallback check in case `TMHOME` is still set and `CMTHOME`
 is not, but you will start to see a warning message in the logs if the old
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -117,7 +117,7 @@ on instead of `~/.tendermint`.
 ### Environment variables
 
 The environment variable prefixes have now changed from `TM` to `CMT`. For
-example, `TMHOME` or `TM_HOME` become `CMTHOME` or `CMT_HOME`.
+example, `TMHOME` or `TM_HOME` become `CMTHOME`.
 
 We have implemented a fallback check in case `TMHOME` is still set and `CMTHOME`
 is not, but you will start to see a warning message in the logs if the old
--- a/abci/types/types.go
+++ b/abci/types/types.go
@@ -138,9 +138,9 @@ var _ jsonRoundTripper = (*ResponseCheckTx)(nil)
 
 var _ jsonRoundTripper = (*EventAttribute)(nil)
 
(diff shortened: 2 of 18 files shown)


lasarojc:
Good catch!
Even if `CMT_HOME` is in the 0.34 documentation, nobody would be using it since does not work, so I would be leaning towards the alternative fix for this, which also keeps the software simpler.

sergio-mena:
> so I would be leaning towards the alternative fix for this, which also keeps the software simpler.

So be it :-D

UPGRADING.md
```diff
@@ -117,7 +117,7 @@ on instead of `~/.tendermint`.
 ### Environment variables
 
 The environment variable prefixes have now changed from `TM` to `CMT`. For
-example, `TMHOME` or `TM_HOME` become `CMTHOME` or `CMT_HOME`.
+example, `TMHOME` or `TM_HOME` become `CMTHOME`.
```
lasarojc, line 120:
```suggestion
example, `TMHOME` becomes `CMTHOME`.
```

lasarojc (approved):

PR #1009, opened 2023-06-20
deps: Manually bump github.com/vektra/mockery/v2 from 2.29.0 to 2.30.1

thanethomson:
Supersedes #995.

We have to do this manually because the code that the new version of mockery generates is subtly (but not meaningfully, afaict) different.

---

#### PR checklist

- [ ] Tests written/updated


mzabaluev (approved):
Mechanical changes, look benign.

PR #934, opened 2023-06-08
mempool: Fix the benchmarks

otrack:
The mempool [benchmarks](https://github.com/cometbft/cometbft/blob/main/mempool/bench_test.go) were invalid due to an incorrect transaction payload.  To see this, one may run `go test -v -count 1 -bench ^BenchmarkCheckTx$ -test.run ^$` against `main` from the `mempool` directory. These modificationss fix such a problem and add a new benchmark to exercice`mempool:update`.

---

#### PR checklist

- [ ] Tests written/updated
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments

PR #1010, opened 2023-06-20
mempool: Keep track of senders in reactor instead of implementation

hvanz:
Implements [ADR-105](https://github.com/cometbft/cometbft/pull/1032)

Currently we keep track of the list of peer IDs that send each transaction in the mempool itself, that is, in `mempoolTx`, the entries of the concurrent list `txs`. This PR moves the senders in `mempoolTx` to a new map `txSenders` from a tx key to a set of peer IDs located in the reactor. The benefits of this change are:
- More clearly separate the transaction propagation protocol in the mempool reactor and the mempool implementation in `CListMempool`.
- Get rid of a complex design pattern in the mempool where the callbacks for `CheckTx` responses of type `CheckTxType_New` are passed to another callback. 
- Allow future improvements to the concurrent handling of ABCI requests.

---

#### PR checklist

- [x] Tests written/updated


sergio-mena:
Since this is am improvement, do we really need to backport this to `v0.38.x`?

consensus/mempool_test.go
```diff
@@ -166,17 +166,16 @@ func TestMempoolRmBadTx(t *testing.T) {
...
-				return
-			}
-			checkTxRespCh <- struct{}{}
-		}, mempl.TxInfo{})
+		reqRes, err := assertMempool(cs.txNotifier).CheckTx(invalidTx)
 		if err != nil {
 			t.Errorf("error after CheckTx: %v", err)
 			return
 		}
+		if reqRes.Response.GetCheckTx().Code != kvstore.CodeTypeInvalidTxFormat {
```
otrack, line 174: Even if the app is in-memory, I believe the verification should be async. (As in the original code base.)

consensus/reactor_test.go
```diff
@@ -232,11 +232,11 @@ func TestReactorCreatesBlockWhenEmptyBlocksFalse(t *testing.T) {
...
 
 	// send a tx
-	if err := assertMempool(css[3].txNotifier).CheckTx(kvstore.NewTxFromID(1), func(resp *abci.ResponseCheckTx) {
-		require.False(t, resp.IsErr())
-	}, mempl.TxInfo{}); err != nil {
+	reqRes, err := assertMempool(css[3].txNotifier).CheckTx(kvstore.NewTxFromID(1))
+	if err != nil {
 		t.Error(err)
 	}
+	require.False(t, reqRes.Response.GetCheckTx().IsErr())
```
otrack, line 239: Same remark as above here.

consensus/reactor_test.go
```diff
@@ -669,11 +669,9 @@ func waitForAndValidateBlock(
 
 		// optionally add transactions for the next block
 		for _, tx := range txs {
-			err := assertMempool(css[j].txNotifier).CheckTx(tx, func(resp *abci.ResponseCheckTx) {
-				require.False(t, resp.IsErr())
-				fmt.Println(resp)
-			}, mempl.TxInfo{})
+			reqRes, err := assertMempool(css[j].txNotifier).CheckTx(tx)
 			require.NoError(t, err)
+			require.False(t, reqRes.Response.GetCheckTx().IsErr())
```
otrack, line 674: Again/

consensus/replay_test.go
```diff
@@ -117,13 +117,14 @@ func sendTxs(ctx context.Context, cs *State) {
...
-				if resp.Code != 0 {
-					panic(fmt.Sprintf("Unexpected code: %d, log: %s", resp.Code, resp.Log))
-				}
-			}, mempool.TxInfo{}); err != nil {
+			reqRes, err := assertMempool(cs.txNotifier).CheckTx(tx)
+			if err != nil {
 				panic(err)
 			}
+			resp := reqRes.Response.GetCheckTx()
+			if resp.Code != 0 {
```
otrack, line 125: Same remark.

mempool/cache_test.go
```diff
@@ -64,10 +64,9 @@ func TestCacheAfterUpdate(t *testing.T) {
 	for tcIndex, tc := range tests {
 		for i := 0; i < tc.numTxsToCreate; i++ {
 			tx := kvstore.NewTx(fmt.Sprintf("%d", i), "value")
-			err := mp.CheckTx(tx, func(resp *abci.ResponseCheckTx) {
-				require.False(t, resp.IsErr())
-			}, TxInfo{})
+			reqRes, err := mp.CheckTx(tx)
 			require.NoError(t, err)
+			require.False(t, reqRes.Response.GetCheckTx().IsErr())
```
otrack, line 69: Again.

mempool/cache_test.go
```diff
@@ -80,9 +79,10 @@ func TestCacheAfterUpdate(t *testing.T) {
 
 		for _, v := range tc.reAddIndices {
 			tx := kvstore.NewTx(fmt.Sprintf("%d", v), "value")
-			_ = mp.CheckTx(tx, func(resp *abci.ResponseCheckTx) {
-				require.False(t, resp.IsErr())
-			}, TxInfo{})
+			reqRes, err := mp.CheckTx(tx)
+			if err == nil {
+				require.False(t, reqRes.Response.GetCheckTx().IsErr())
```
otrack, line 84: Again.

mempool/reactor.go
```diff
@@ -100,19 +115,22 @@ func (memR *Reactor) Receive(e p2p.Envelope) {
...
-			ntx := types.Tx(tx)
-			err = memR.mempool.CheckTx(ntx, nil, txInfo)
+		for _, txBytes := range protoTxs {
+			tx := types.Tx(txBytes)
+
+			// Record the sender of every transaction received.
+			// Senders are stored until the transaction is removed from the mempool.
+			// Note that it's possible a tx is still in the cache but no longer in the mempool.
+			// For example, after committing a block, txs are removed from mempool but not the cache.
+			memR.addSender(tx.Key(), memR.ids.GetForPeer(e.Src))
```
otrack, line 127: I think we should try to maintain the following invariant: 

  $tx \in memR.mempool.txs \iff tx \in memR.txSenders.keys()$. 

The rationale is to keep memory usage bounded. However, the code here is breaking such an invariant, because `tx` might not be added to the mempool.

otrack (approved):
Great work :smiley: I believe that such a rewriting was necessary. The proposed pattern is way cleaner!

As a bonus, this change would permit to use locks in the two callbacks. This will permit to (i) remove the `flush` operation at the ABCI interface, and (ii) drop the need for in order processing of `checkTx` requests in the app.  Hence, this change may lead to nice future performance improvements in the mempool.

Regarding the current PR, I just have a handful of remarks (see inlined comments).

mempool/reactor.go
hvanz: I agree and I have changed this code to record the sender only when the transaction is valid (and added to the mempool). Another problem with my previous code was that, if the received transaction was invalid, it would store the sender and never be removed.

hvanz:
> Since this is am improvement, do we really need to backport this to `v0.38.x`?

You're right. We don't need to backport it.

consensus/reactor_test.go
hvanz: Same answer as above.

mempool/cache_test.go
hvanz: Same answer as above.

config/config.go
```diff
@@ -745,7 +745,11 @@ func DefaultMempoolConfig() *MempoolConfig {
 // TestMempoolConfig returns a configuration for testing the CometBFT mempool
 func TestMempoolConfig() *MempoolConfig {
 	cfg := DefaultMempoolConfig()
-	cfg.CacheSize = 1000
+	cfg.Size = 1000
+	cfg.CacheSize = 2000
+	if cfg.Size > cfg.CacheSize {
```
otrack, line 750: unclear why

mempool/clist_mempool.go
```diff
@@ -31,6 +32,10 @@ type CListMempool struct {
 	notifiedTxsAvailable bool
 	txsAvailable         chan struct{} // fires once for each height, when the mempool is not empty
 
+	// Function set by the reactor to be called when a transaction is removed
+	// from the mempool.
+	removeTxOnReactor func(txKey types.TxKey)
```
otrack, line 37: rename  removeTxOnReactorCb instead?

mempool/clist_mempool.go
```diff
@@ -130,6 +149,16 @@ func (mem *CListMempool) EnableTxsAvailable() {
 	mem.txsAvailable = make(chan struct{}, 1)
 }
 
+func (mem *CListMempool) SetTxRemovedCallback(cb func(txKey types.TxKey)) {
+	mem.removeTxOnReactor = cb
+}
+
+func (mem *CListMempool) invokeRemoveTxOnReactor(txKey types.TxKey) {
+	if mem.removeTxOnReactor != nil {
```
otrack, line 157: I would report an error if the callback is null here. This should never happen.

mempool/clist_mempool.go
```diff
@@ -106,11 +112,23 @@ func (mem *CListMempool) getCElement(txKey types.TxKey) (*clist.CElement, bool)
...
+
+func (mem *CListMempool) addToCache(tx types.Tx) bool {
+	return mem.cache.Push(tx)
+}
+
+func (mem *CListMempool) forceRemoveFromCache(tx types.Tx) {
+	mem.cache.Remove(tx)
+}
+
+func (mem *CListMempool) removeFromCache(tx types.Tx) {
```
otrack, line 128: rename tryRemoveInvalidTxFromCache instead?

mempool/clist_mempool.go
```diff
@@ -496,10 +476,7 @@ func (mem *CListMempool) resCbRecheck(req *abci.Request, res *abci.Response) {
 			if err := mem.RemoveTxByKey(memTx.tx.Key()); err != nil {
 				mem.logger.Debug("Transaction could not be removed from mempool", "err", err)
 			}
-			// We remove the invalid tx from the cache because it might be good later
-			if !mem.config.KeepInvalidTxsInCache {
-				mem.cache.Remove(tx)
-			}
+			mem.removeFromCache(tx) // it might be valid later
```
otrack, line 479: this comment should be inside the removeFromCache function

mempool/clist_mempool.go
```diff
@@ -130,6 +149,16 @@ func (mem *CListMempool) EnableTxsAvailable() {
 	mem.txsAvailable = make(chan struct{}, 1)
 }
 
+func (mem *CListMempool) SetTxRemovedCallback(cb func(txKey types.TxKey)) {
+	mem.removeTxOnReactor = cb
+}
```
otrack, line 154: check it is called once? (defensive programming)

mempool/clist_mempool.go
```diff
@@ -209,122 +238,80 @@ func (mem *CListMempool) TxsWaitChan() <-chan struct{} {
...
-) func(res *abci.Response) {
-	return func(res *abci.Response) {
-		if mem.recheckCursor != nil {
-			// this should never happen
-			panic("recheck cursor is not nil in reqResCb")
+		case abci.CheckTxType_Recheck:
+			if mem.recheckCursor == nil {
+				return
+			}
+			mem.metrics.RecheckTimes.Add(1)
```
otrack, line 306: it is better positioned here indeed

mempool/clist_mempool.go
```diff
@@ -342,6 +329,9 @@ func (mem *CListMempool) addTx(memTx *mempoolTx) {
 //   - Update (lock held) if tx was committed
 //   - resCbRecheck (lock not held) if tx was invalidated
 func (mem *CListMempool) RemoveTxByKey(txKey types.TxKey) error {
+	// The transaction should be removed in the reactor, even if it cannot be
```
otrack, line 332: in -> from

I agree with this pattern.

mempool/clist_mempool.go
otrack, line 1: func (mem *CListMempool) addTx(memTx *mempoolTx)  

rename addMemTx instead?

mempool/clist_mempool.go
```diff
@@ -410,36 +397,29 @@ func (mem *CListMempool) resCbFirstTime(
...
 				height:    mem.height,
 				gasWanted: r.CheckTx.GasWanted,
 				tx:        tx,
-			}
-			memTx.addSender(txInfo.SenderID)
-			mem.addTx(memTx)
+			})
 			mem.logger.Debug(
-				"added good transaction",
+				"added valid transaction",
```
otrack, line 406: +1

in general, it would be nice to use valid/invalid 
instead of their synonyms everywhere

mempool/clist_mempool.go
```diff
@@ -598,6 +575,7 @@ func (mem *CListMempool) ReapMaxTxs(max int) types.Txs {
 }
 
 // Lock() must be help by the caller during execution.
+// TODO: always returns nil
```
otrack, line 578: Well, it may return an error. Strictly speaking the function is void.

mempool/bench_test.go
otrack, line 1: In `BenchmarkReap` , the line `binary.BigEndian.PutUint64(tx, uint64(i))` should be removed.
This was mistakenly forgotten in #934.
Can you please take care of this?

otrack (approved):
This PR has a slightly positive performance impact (according to the benchmarks in `/mempool/bench_test.go`).  Overall, I am quite fine with these changes, which are all carefully written. Great work!

mempool/clist_mempool.go
hvanz: I added a comment.

mempool/clist_mempool.go
hvanz: I renamed it to `tryRemoveFromCache`. The function takes any kind of transaction; it doesn't know that is invalid.

mempool/clist_mempool.go
hvanz: Yes, what I wanted to say is that, because `Update`  always returns nil, we should remove the returning error. But I would leave that for another PR.

mempool/bench_test.go
hvanz: Cool. It was failing and I didn't know why. I removed it.

mempool/clist_mempool.go
hvanz: Not sure if it's needed. This function is only called in the reactor constructor.

cason (commented):
This PR is API breaking. It changes the signature (parameters and return type) of main API public method offered by the mempool, namely `CheckTx`.

I wonder whether this is really necessary to implement the changes proposed in this PR, namely, store the senders of a transaction in the reactor instead of in the mempool.

In general, I think that the interface between the mempool (list of pending transactions) and the mempool protocol (which disseminates transactions) should be discussed in more detail.

I would suggest coming up with a document (e.g., an ADR, but it can also be an issue) that lists and discusses all the proposed changes, so that we feel more comfortable in breaking and re-proposing that public API.

.changelog/unreleased/breaking-changes/1010-mempool-interface.md
lasarojc, line 1: Pls also update the upgrading guide, showing how to use the updated API.

lasarojc (approved):

.changelog/unreleased/breaking-changes/1010-mempool-interface.md
hvanz: Done.

hvanz:
> This PR is API breaking. It changes the signature (parameters and return type) of main API public method offered by the mempool, namely CheckTx.

The `Mempool` interface is an internal API, not a public one. Even when a few users may want to access its methods directly, it is was agreed that the change is justified in this case.

PR #857, opened 2023-05-18
feat: make handshake cancelable

yihuang:
it'll make the handshake work with graceful shutdown.

see: https://github.com/cosmos/cosmos-sdk/issues/16202

---

#### PR checklist

- [ ] Tests written/updated
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments

PR #1011, opened 2023-06-21

yihuang:

this one can be backport to 0.37 and 0.34.

---

#### PR checklist

- [ ] Tests written/updated

---

#### PR checklist

- [ ] Tests written/updated


mzabaluev (approved):
Looks good as a backport to me.

sergio-mena (approved):
We might need to refactor this in a breaking way (to simplify) in a future release. But this is good for a backport

yihuang:
> We might need to refactor this in a breaking way (to simplify) in a future release. But this is good for a backport

the breaking version is already in main branch, is that what you mean?

sergio-mena:
> is that what you mean?

Oh yes, I had missed that. All good

PR #1013, opened 2023-06-21

yihuang:

---

- [ ] Tests written/updated

---

#### PR checklist

- [ ] Tests written/updated


consensus/replay.go
```diff
@@ -265,7 +271,7 @@ func (h *Handshaker) Handshake(proxyApp proxy.AppConns) error {
 	}
 
 	// Replay blocks up to the latest in the blockstore.
-	_, err = h.ReplayBlocks(h.initialState, appHash, blockHeight, proxyApp)
+	appHash, err = h.ReplayBlocksWithContext(ctx, h.initialState, appHash, blockHeight, proxyApp)
```
sergio-mena, line 274: I know in future branches the original here re-writes `appHash` but not here. I'd leave it as is unless we understand why we are making that change.

sergio-mena: @yihuang any thoughts?

yihuang: I think it's kind of bug fix, previously the log don't output latest hash, but I'm ok that we change that in a separate PR

yihuang: hi @sergio-mena , do you want me to revert this line?

sergio-mena: Hi @yihuang sorry for delay on coming back to this (constant flow of kinda-more-urgent things)

I took a look at this again, and when comparing this with the backport on v0.34.x (#1012) I realise this change is present there as well, so I'm OK with it. Approving...

sergio-mena (approved):

PR #1017, opened 2023-06-22
Remove genesis persistence in state db

drklee3:
This avoids saving the genesisDoc to database. When using goleveldb and ~4GiB+ genesis files, it causes a panic during snappy encoding (`panic: snappy: decoded block is too large`). We run into this with a large Ethermint state in our mainnet exports.

The main file to review is `node/setup.go`, with changes mostly from https://github.com/tendermint/tendermint/pull/6554

Contributes to #1037

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/node/setup.go
+++ b/node/setup.go
@@ -3,7 +3,6 @@ package node
 import (
 	"bytes"
 	"context"
-	"errors"
 	"fmt"
 	"net"
 	"strings"
@@ -21,7 +20,6 @@ import (
 	"github.com/cometbft/cometbft/evidence"
 	"github.com/cometbft/cometbft/statesync"
 
-	cmtjson "github.com/cometbft/cometbft/libs/json"
 	"github.com/cometbft/cometbft/libs/log"
 	"github.com/cometbft/cometbft/light"
 	mempl "github.com/cometbft/cometbft/mempool"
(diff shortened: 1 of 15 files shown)


node/setup.go
```diff
@@ -528,67 +526,32 @@ func startStateSync(
...
+// loadStateFromDBOrGenesisDocProvider attempts to load the state from the
 // database, or creates one using the given genesisDocProvider. On success this also
 // returns the genesis doc loaded through the given provider.
-func LoadStateFromDBOrGenesisDocProvider(
-	stateDB dbm.DB,
-	genesisDocProvider GenesisDocProvider,
-) (sm.State, *types.GenesisDoc, error) {
-	// Get genesis doc
-	genDoc, err := loadGenesisDoc(stateDB)
+func loadStateFromDBOrGenesisDocProvider(stateStore sm.Store, genDoc *types.GenesisDoc) (sm.State, error) {
```
mzabaluev, line 532: How about shortening this to `loadStateFromDbOrGenesis`?

mzabaluev (commented):
Good quality code.
One concern voiced about this change is, we'd like to keep a hash of the genesis doc in the store for a sanity check.

node/setup.go
```diff
@@ -528,67 +526,32 @@ func startStateSync(
...
-		err = genDoc.ValidateAndComplete()
+	if state.IsEmpty() {
+		// 2. If it's not there, derive it from the genesis doc
+		state, err = sm.MakeGenesisState(genDoc)
 		if err != nil {
-			return sm.State{}, nil, fmt.Errorf("error in genesis doc: %w", err)
+			return sm.State{}, err
 		}
-		// save genesis doc to prevent a certain class of user errors (e.g. when it
-		// was changed, accidentally or not). Also good for audit trail.
```
mzabaluev, line 553: This is what I meant in https://github.com/cometbft/cometbft/pull/1017#pullrequestreview-1512645878
The hash should be stored, and checked against the genesis doc at the beginning of this function.

jmalicevic:
Hey @drklee3 , thanks a lot for doing this. As @mzabaluev  above said  there are a few more options we are considering adding to this solution (like the hash of the doc in the store etc.) , its impact on the retrieval of the genesis doc via RPC and whether it can be improved when we have a solution for storing the file like this. We will explore them and build on the solution this quarter though : see https://github.com/cometbft/cometbft/issues/1037 .

node/setup.go
drklee3: updated to check and store the hash !

alpha-omega-labs:
We need that fix to be able to import large (~10GB) genesis.json is kava or cronos fork for GenesisL1 blockchain.
Thank you for addressing that issue, its very helpful!

jmalicevic:
Re-opening this because it is waiting for us to take a look at it once ADR-101 PRs are merged.

node/setup.go
```diff
@@ -528,67 +528,60 @@ func startStateSync(
...
-		return sm.State{}, nil, err
+		return sm.State{}, err
 	}
-	return state, genDoc, nil
-}
 
-// panics if failed to unmarshal bytes
-func loadGenesisDoc(db dbm.DB) (*types.GenesisDoc, error) {
-	b, err := db.Get(genesisDocKey)
+	genDocBytes, err := cmtjson.Marshal(genDoc)
```
mzabaluev, line 547: I wonder if we can plug in some streaming output backend to `Marshal` that would only compute the hash from the data, instead of populating a potentially huge buffer.

JSON is not a canonical data representation, so any semantically insignificant variation, e.g. added whitespace, in the genesis file would result in a hash mismatch, if not for this approach of hashing the sanitized marshalling output. Is this intentional? In any case it adds some robustness. Maybe it's worth explaining somewhere that the hash we store in the database is not actually the hash of the file content.

mzabaluev (commented):
This looks like what we want in the first phase now, except for some concerns with the method and performance of obtaining the genesis doc hash.

node/setup.go
```diff
@@ -528,67 +528,60 @@ func startStateSync(
 
 //------------------------------------------------------------------------------
 
-var genesisDocKey = []byte("genesisDoc")
+var genesisDocHashKey = []byte("genesisDocHash")
 
-// LoadStateFromDBOrGenesisDocProvider attempts to load the state from the
+// loadStateFromDbOrGenesisDoc attempts to load the state from the
 // database, or creates one using the given genesisDocProvider. On success this also
 // returns the genesis doc loaded through the given provider.
```
sergio-mena, line 535:
```suggestion
// database, or creates one using the given genesisDocProvider.
```

node/setup.go
sergio-mena: Agree that hashing based on JSON serialization is brittle. Maybe it would be better to use what we use for (signed) votes and proposals (but without signing). Please find an example [here](https://github.com/cometbft/cometbft/blob/8e6e6ca0b/test/e2e/app/app.go#L664).

This won't totally remove json from the equation, as we still have field `AppState`, which contains the app's state in serialized json format. But that one has been like that for years (battle-tested). Although, if we want to improve on that as well, the moment is now (since we'll need an upgrade strategy anyway).

andynog: I think the concern with the `AppState` is not great if they use the `cometbft` `genesis` type which they do. So in the SDK they use the `genesis.saveAs` from cometbft, as long as that doesn't change I'd think should be OK.

For example, another thing to keep in mind is that `saveAs` function uses a cometbft `MarshalIdent` function so the exported json is in a format that is very specific. If there's any "outside" process that formats this file (e.g. remove indentation) then every might break
https://github.com/cometbft/cometbft/blob/8e6e6ca0bcc732747e10020be2c5e9f3779e6b0c/types/genesis.go#L50

andynog (approved):
We can approve and merge this PR and address the concerns in a future PR. The concerns will be addressed in #1287

andynog:
Hi @drklee3, we will merge this PR and address some of the concerns we mentioned in a later PR, we'll be using issue #1287 to track that. But for now the check on [Golang Linter is failing](https://github.com/cometbft/cometbft/actions/runs/5712621277/job/15476459020?pr=1017). Could you please fix the lint problems and push it again so we can merge ? Thanks !

sergio-mena (approved):
See @andynog's last comment

PR #1019, opened 2023-06-23
[backport to v0.37.x] config: add bootstrap peers (tendermint#9680)

### LINKED ISSUE #1016: Backport "config: add bootstrap peers" (#9680) to v0.37.x
## Summary

I would come real handy to have [9680](https://github.com/tendermint/tendermint/pull/9680) backported to 0.37.x.

## Problem Definition

Having only `seeds` and `persistent_peers` is kinda annoying for some use cases.

## Proposal

Just backport [9680](https://github.com/tendermint/tendermint/pull/9680) 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

cason:
Closes #1016.

Reviewers, please help with the `.changelog/` entry. The remain is just a simple backport from main.

### DIFF
--- a/.changelog/unreleased/features/9680-config-bootstrap-ppers.md
+++ b/.changelog/unreleased/features/9680-config-bootstrap-ppers.md
@@ -1,0 +1,3 @@
+- `[config]` Introduce `BootstrapPeers` to the config to allow
+  nodes to list peers to be added to the addressbook upon start up.
+  ([\#9680](https://github.com/tendermint/tendermint/pull/9680))
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -20,6 +20,15 @@ coordinated upgrade.
 When upgrading from the v0.34 release series, please note that the Go module has
 now changed to `github.com/cometbft/cometbft`.
 
+## Config Changes
+
+* A new config field, `BootstrapPeers` has been introduced as a means of
+  adding a list of addresses to the addressbook upon initializing a node. This is an
+  alternative to `PersistentPeers`. `PersistentPeers` shold be only used for
+  nodes that you want to keep a constant connection with i.e. sentry nodes
+
+----
+
 ### ABCI Changes
 
 * The `ABCIVersion` is now `1.0.0`.
--- a/config/config.go
+++ b/config/config.go
@@ -558,6 +558,11 @@ type P2PConfig struct { //nolint: maligned
 	// We only use these if we can’t connect to peers in the addrbook
 	Seeds string `mapstructure:"seeds"`
 
+	// Comma separated list of peers to be added to the peer store
+	// on startup. Either BootstrapPeers or PersistentPeers are
+	// needed for peer discovery
+	BootstrapPeers string `mapstructure:"bootstrap_peers"`
+
 	// Comma separated list of nodes to keep persistent connections to
 	PersistentPeers string `mapstructure:"persistent_peers"`
 
--- a/config/toml.go
+++ b/config/toml.go
@@ -284,6 +284,11 @@ external_address = "{{ .P2P.ExternalAddress }}"
 # Comma separated list of seed nodes to connect to
 seeds = "{{ .P2P.Seeds }}"
 
+# Comma separated list of peers to be added to the peer store
+# on startup. Either BootstrapPeers or PersistentPeers are
+# needed for peer discovery
+bootstrap_peers = "{{ .P2P.BootstrapPeers }}"
+
 # Comma separated list of nodes to keep persistent connections to
 persistent_peers = "{{ .P2P.PersistentPeers }}"
 
--- a/node/node.go
+++ b/node/node.go
@@ -873,6 +873,17 @@ func NewNode(config *cfg.Config,
 		return nil, fmt.Errorf("could not create addrbook: %w", err)
 	}
 
+	for _, addr := range splitAndTrimEmpty(config.P2P.BootstrapPeers, ",", " ") {
+		netAddrs, err := p2p.NewNetAddressString(addr)
+		if err != nil {
+			return nil, fmt.Errorf("invalid bootstrap peer address: %w", err)
+		}
+		err = addrBook.AddAddress(netAddrs, netAddrs)
+		if err != nil {
+			return nil, fmt.Errorf("adding bootstrap address to addressbook: %w", err)
+		}
+	}
+
 	// Optionally, start the pex reactor
 	//
 	// TODO:
--- a/cmd/cometbft/commands/run_node.go
+++ b/cmd/cometbft/commands/run_node.go
@@ -66,6 +66,7 @@ func AddNodeFlags(cmd *cobra.Command) {
 	cmd.Flags().String("p2p.external-address", config.P2P.ExternalAddress, "ip:port address to advertise to peers for them to dial")
 	cmd.Flags().String("p2p.seeds", config.P2P.Seeds, "comma-delimited ID@host:port seed nodes")
 	cmd.Flags().String("p2p.persistent_peers", config.P2P.PersistentPeers, "comma-delimited ID@host:port persistent peers")
+	cmd.Flags().String("p2p.bootstrap_peers", config.P2P.BootstrapPeers, "comma-delimited ID@host:port peers to be added to the addressbook on startup")
 	cmd.Flags().String("p2p.unconditional_peer_ids",
 		config.P2P.UnconditionalPeerIDs, "comma-delimited IDs of unconditional peers")
 	cmd.Flags().Bool("p2p.upnp", config.P2P.UPNP, "enable/disable UPNP port forwarding")
--- a/spec/p2p/config.md
+++ b/spec/p2p/config.md
@@ -17,6 +17,14 @@ and upon incoming connection shares some peers and disconnects.
 Dials these seeds when we need more peers. They should return a list of peers and then disconnect.
 If we already have enough peers in the address book, we may never need to dial them.
 
+## Bootstrap Peers
+
+`--p2p.bootstrap_peers “id100000000000000000000000000000000@1.2.3.4:26656,id200000000000000000000000000000000@2.3.4.5:26656”`
+
+A list of peers to be added to the addressbook upon startup to ensure that the node has some peers to initially dial.
+Unlike persistent peers, these addresses don't have any extra privileges. The node may not necessarily connect on redial
+these peers.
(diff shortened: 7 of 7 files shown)


config/toml.go
```diff
@@ -284,6 +284,11 @@ external_address = "{{ .P2P.ExternalAddress }}"
 # Comma separated list of seed nodes to connect to
 seeds = "{{ .P2P.Seeds }}"
 
+# Comma separated list of peers to be added to the peer store
+# on startup. Either BootstrapPeers or PersistentPeers are
+# needed for peer discovery
```
jmalicevic, line 289: Maybe write down what the difference is between them although most people already know.

jmalicevic (approved):
Minor comment but LGTM.

node/node.go
```diff
@@ -873,6 +873,17 @@ func NewNode(config *cfg.Config,
 		return nil, fmt.Errorf("could not create addrbook: %w", err)
 	}
 
+	for _, addr := range splitAndTrimEmpty(config.P2P.BootstrapPeers, ",", " ") {
```
sergio-mena, line 876: As this is hard to unit-test (code in Node usually is), would it make sense to add support for this in the e2e manifest (no need to touch the generator) so we can test this in, e.g., `networks/ci.toml`?

config/toml.go
cason: The affirmation is not even true. One node can discovery peers by only having `SeedNodes` config parameter set.

config/config.go
```diff
@@ -558,6 +558,11 @@ type P2PConfig struct { //nolint: maligned
 	// We only use these if we can’t connect to peers in the addrbook
 	Seeds string `mapstructure:"seeds"`
 
+	// Comma separated list of peers to be added to the peer store
```
cason, line 561:
```suggestion
	// Comma separated list of peers to be added to the address book
```

config/toml.go
cason: So, here there are several points which are not clear:

1. A node can discovery peers provided it has any address on its address book (`BootrstapPeers` flag ensures this), any peer configured as persistent, or any address configured as seed node
2. Persistent peers is not meant for peer discovery. It is mostly used by validator nodes that don't run the PEX reactor, namely, don't participate on the peer discovery protocol
3. There is no peer store. There is an address book with addresses of _potential_ peers

node/node.go
cason: I didn't get the point here.

cason: The addresses here are directly passed to the address book, the address book unit tests should cover this. If not, the PEX reactor unit test, which requires setting up the address book, should also cover this.

sergio-mena: > I didn't get the point here.

As a general rule, when we add production code, we need to add some (unit) test code that hits the new code.
Here, I think unit testing would be cumbersome and pointless (but I might be wrong, as I didn't look around into UTs that test neighboring code... it may be simpler), so I was suggesting to add the config option as part of `networks/ci.toml` so that it's part of e2e

> The addresses here are directly passed to the address book, the address book unit tests should cover this. If not, the PEX reactor unit test, which requires setting up the address book, should also cover this.

Can you point at the test cases that are testing this new code?

cason: There is no test unit covering this flag specifically. In particular because this is not a flag that is considered by the p2p layer.

What I meant is that there are test cases covering the `addrBook.AddAddress()` method, which is a very basic method of the PEX implementation, that is invoked for all addresses parsed from this config flag.

cason: > so I was suggesting to add the config option as part of networks/ci.toml so that it's part of e2e

Ok, I will take a look on this file.

cason: Ok, found it, but this is a specific config triggering some scenarios. I don't see the use of this flag in the network setup/topology enforced in those runs. Actually, those runs do not cover the user case of this new flag, as node rely on persistent peers and a seed node for establishing connections. The option there would be to replace a persistent peer in a given node by a bootstrap peer, but this would mean changing the semantics of the setup, as the node will not necessarily connect to the configured bootstrap peer, as it can retrieve other addresses from the nodes configured as persistent or seeds beforehand.

cason:
So, I am ok to apply the suggested changes provided they are not part of this PR. This PR is meant to backport the code and operation present on `main` and `v0.38.x` branches to the `v0.37.x` branches.

I can take the opportunity to update the documentation (this includes comments on soruce or documentation files) so that they are more precise and clear. But I would rather do that as part of a different issue/PR, and first on `main`, then on release branches.

cason:
I've created this new issue to handle updates on the p2p parameters: https://github.com/cometbft/cometbft/issues/1038

.changelog/unreleased/features/9680-config-bootstrap-ppers.md
```diff
@@ -0,0 +1,3 @@
+- `[config]` Introduce `BootstrapPeers` to the config to allow
```
thanethomson, line 1:
```suggestion
- `[config]` Introduce `p2p.bootstrap_peers` to the config to allow
```

Operators don't see the `BootstrapPeers` variable - they see the configuration file entry.

UPGRADING.md
```diff
@@ -20,6 +20,15 @@ coordinated upgrade.
 When upgrading from the v0.34 release series, please note that the Go module has
 now changed to `github.com/cometbft/cometbft`.
 
+## Config Changes
+
+* A new config field, `BootstrapPeers` has been introduced as a means of
+  adding a list of addresses to the addressbook upon initializing a node. This is an
+  alternative to `PersistentPeers`. `PersistentPeers` shold be only used for
+  nodes that you want to keep a constant connection with i.e. sentry nodes
```
thanethomson, line 28:
```suggestion
* A new config field, `p2p.bootstrap_peers` has been introduced as a means of
  adding a list of addresses to the addressbook upon initializing a node. This is an
  alternative to `p2p.persistent_peers`. `p2p.persistent_peers` should be only used for
  nodes that you want to keep a constant connection with i.e. sentry nodes.
```

config/toml.go
```diff
@@ -284,6 +284,11 @@ external_address = "{{ .P2P.ExternalAddress }}"
 # Comma separated list of seed nodes to connect to
 seeds = "{{ .P2P.Seeds }}"
 
+# Comma separated list of peers to be added to the peer store
+# on startup. Either BootstrapPeers or PersistentPeers are
```
thanethomson, line 288:
```suggestion
# on startup. Either bootstrap_peers or persistent_peers are
```

The configuration file has no notion of `BootstrapPeers` - this is the Go code field name. The representation in the config file is `bootstrap_peers`.

node/node.go
sergio-mena: OK, if you think it makes no sense to test this using e2e, then we're back with unit tests.

I took a closer look at relevant existing UTs. There is `TestNodeNewNodeCustomReactors` which is testing the `NewNode` function where your new code is. So I see several options:

* You can create a similar test case, next to it, where you modify the config before passing it to `NewNode` so that in contains relevant bootstrap peer config. The easiest is to feed incorrect config, to exercise the error paths introduced by the new code (where you are returning an error).
* For testing the happy path, I think it's best to modify `TestP2PConfig` to include correct bootstrap config, and then re-running all UTs and make sure they all pass. Alternatively, you can just test the happy path as part of the new UT where you're testing the error paths

sergio-mena:
> This PR is meant to backport the code and operation present on main and `v0.38.x` branches to the `v0.37.x` branches.

I see. Then the UTs were missing on the initial PR. Please add the UTs I am requesting on `main` (in a different PR)

node/node.go
```diff
@@ -873,6 +873,17 @@ func NewNode(config *cfg.Config,
...
+		netAddrs, err := p2p.NewNetAddressString(addr)
+		if err != nil {
+			return nil, fmt.Errorf("invalid bootstrap peer address: %w", err)
+		}
+		err = addrBook.AddAddress(netAddrs, netAddrs)
```
cason, line 881: I've just realize that the source address (second parameter) should not be that one.

In fact, when we manually add addresses to the address book, the source should be the node itself. Refer to the implementation [spec](https://github.com/cometbft/cometbft/blob/main/spec/p2p/implementation/addressbook.md#adding-addresses) (point 3. referring to the source address) and to the way we add addresses of configured persistent peers on the [switch](https://github.com/cometbft/cometbft/blob/main/p2p/switch.go#L503).

cason:
I think we should block (if not drop) this PR, since the implementation of this feature in the `main` branch has a lot of (minor) problems, listed by the reviewers.

For instance:
- The documentation of the new configuration parameter is imprecise and it is not following the correct naming for config/toml file parameters
- The way we are adding the parsed addresses to the address book is incorrect (improper source address)
- We don't have tests checking the operation of the introduced configuration parameter

While none of those problems are critical, it would be good to fix them on `main` to then backport a cleaner and safer implementation to the feature branches.

sergio-mena:
@cason I think I agree with you overall. Here's the next steps I suggest

* As this is a backport (I hadn't realized until yesterday, sorry),
  * either we backport it as-is (with no changes w.r.t. the original patch),
  * or we revert the original on `main` (and any other branch that may contain it)
*  and then, just after, we open a PR that addresses the problems the original patch has (on `main` then backport)

Also, this PR was in response to a user's request wasn't it?

cason:
> Also, this PR was in response to a user's request wasn't it?

Yes, #1016

sergio-mena:
Then, whichever option we take, we need to drive it to completion, so #1016 can be considered done

cason:
I don't know what is the best solution here.

Fixing the documentation of the p2p parameters is an issue I plan to tackle from next week, as part of #1038. This issue is more extensive, but in particular addresses the documentation of the parameter introduced by this PR.

Reverting the original PR on `main` and `v0.38.x` and replace for a correct and minimally unit-tested solution is also on my plans, with support of @glnro.

sergio-mena:
> Reverting the original PR on main and v0.38.x and replace for a correct and minimally unit-tested solution is also on my plans, with support of @glnro.

Yes, please, go ahead and drive it to completion. I don't think it's a lot of work to fix the doc messages on the original patch, and add a UT to test the added code.

cason:
Rejected. Please refer to https://github.com/cometbft/cometbft/pull/1109 for more detailed reasons.

PR #1023, opened 2023-06-24
Update codeowners to include DevRel team

thanethomson:
I noticed that we hadn't added the DevRel team to our CODEOWNERS file yet. This PR updates it.

Also, our org name was changed from "CometBFT" to "cometbft" some time ago, and while this is a purely cosmetic change it probably makes sense to keep it consistent.

---

#### PR checklist

- [ ] Tests written/updated


sergio-mena (approved):
LGTM

PR #1032, opened 2023-06-26
mempool: ADR for refactoring list of senders 

hvanz:
Triggered by this comment https://github.com/cometbft/cometbft/pull/895#issuecomment-1584948704

### DIFF
--- a/docs/architecture/adr-105-refactor-mempool-senders.md
+++ b/docs/architecture/adr-105-refactor-mempool-senders.md
@@ -1,0 +1,148 @@
+# ADR 105: Refactor list of senders in mempool
+
+## Changelog
+
+- 2023-06-26: Initial draft (@hvanz)
+
+## Status
+
+Proposed
+
+## Context
+
+Before adding a transaction to the mempool, we need to send a `CheckTx` message
+to the application for validating the transaction. There are two variants of
+this message, distinghished by the value in message field `type`:
+- `CheckTxType_New` is for transactions that need to be validated before adding
+it to the mempool for the first time.
+- `CheckTxType_Recheck` is for transactions that are in the mempool and need to
+be re-validated after committing a block and advancing to the next height.
+
+The mempool communicates with the ABCI server (that is, the application) by
+sending `abci.Request`s through the proxy `AppConnMempool`. The proxy provides a
+callback mechanism for handling `abci.Response`s. The current mempool
+implemention `CListMempool` (also called v0) utilizes this mechanism for
+`Recheck` transactions but not for `New` transactions. Instead `New`
+transactions require an ad-hoc mechanism for each request. 
+
+The reason behind this difference is that for `New` transactions we need to
+record the ID of the peer that sent the transaction. However, this information
+is not included in `RequestCheckTx` messages. Recording the sender's ID is
+necessary for the transaction propagation protocol, which uses the recorded list
+of senders to prevent sending the transaction back to these peers, thus avoiding
+duplicated messages. More importantly, this mechanism serves as the only means
+to stop propagating transactions.
+
+There are two design problems with this implementation. First, there is a
+complex pattern for handling callbacks on `New` requests. The following code
+snippet at the end of the `CheckTx` method, where transactions received for the
+first time are processed, demonstrates the issue:
+``` golang
+	reqRes, err := mem.proxyAppConn.CheckTxAsync(context.TODO(), &abci.RequestCheckTx{Tx: tx})
+	reqRes.SetCallback(mem.reqResCb(tx, txInfo, cb))
+```
+When we send a request for validating a transaction via `CheckTxAsync`, it
+returns a `ReqRes` object. To handle the response asynchronously, we set an
+ad-hoc callback `reqResCb` on `ReqRes`. This callback is different for each
+transaction `tx` because it's parameterized by `tx`, `txInfo` (which essentially
+contains the sender ID), and another callback function `cb` provided by the
+caller of `CheckTx` to be applied on the response.
+
+Secondly, the list of senders for each transaction is recorded directly in the
+mempool's `txs` data structure. However, the list of senders is an important
+component of the propagation protocol and it should be part of the reactor,
+while `txs` is part of the implementation of this specific version (v0) of the
+mempool.
+
+In this document, we propose a solution that involves moving the list of senders
+from the mempool implementation to the reactor. This change will simplify the
+code, establish a more clear separation of the propagation protocol and the data
+structure, and allow for future improvements to the mempool as a whole.
+
+## Alternative Approaches
+
+> This section contains information around alternative options that are considered
+> before making a decision. It should contain a explanation on why the alternative
+> approach(es) were not chosen.
+
+## Decision
+
+> This section records the decision that was made.
+> It is best to record as much info as possible from the discussion that happened.
+> This aids in not having to go back to the Pull Request to get the needed information.
+
+## Detailed Design
+ 
+We propose the following changes to the mempool's reactor and data structure.
+
+- In the `Mempool` interface, change the signature of `CheckTx` from
+    ``` golang
+    CheckTx(tx types.Tx, cb func(*abci.ResponseCheckTx), txInfo TxInfo) error
+    ```
+    to
+    ``` golang
+    CheckTx(tx types.Tx) (abcicli.ReqRes, error)
+    ```
+  - The returning `ReqRes` object can be used to set and invoke a callback to
+    handle the response, if needed.
+  - The callback parameter `cb` is no longer needed. Currently, this is mainly
+    used by the RPC endpoints `broadcast_tx_sync` and `broadcast_tx_commit`, but
+    the same functionality can be obtained with `ReqRes`.
+  - `txInfo` contains information about the sender, which is also no longer
+    needed, as justified by the next point.
+- The list of senders for each transaction is stored in `mempoolTx`, the entries
+  for `txs`. Move the senders out of `mempoolTx` to a new map `txSenders` of
+  type `map[types.TxKey]map[uint16]bool` in the mempool reactor. `txSenders`
+  would map transaction keys to a set (a map to booleans) of peer ids. Add also
+  a `cmtsync.RWMutex` lock to handle concurrent accesses to the map.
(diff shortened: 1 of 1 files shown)


docs/architecture/adr-105-refactor-mempool-senders.md
```diff
@@ -0,0 +1,141 @@
...
+`Recheck` transactions but not for `New` transactions. Instead `New`
+transactions require an ad-hoc mechanism for each request. 
+
+The reason behind this difference is that for `New` transactions we need to
+record the ID of the peer that sent the transaction. However, this information
+is not included in `RequestCheckTx` messages. Recording the sender's ID is
+necessary for the transaction propagation protocol, which uses the recorded list
+of senders to prevent sending the transaction back to these peers, thus avoiding
+duplicated messages. More importantly, this mechanism serves as the only means
+to stop propagating transactions.
```
sergio-mena, line 34: Are we sure we'll still need this info, in this way, by the end of Q3?

docs/architecture/adr-105-refactor-mempool-senders.md
```diff
@@ -0,0 +1,141 @@
...
+    When a transaction is removed from the mempool, notify the reactor to remove
+    the list of senders for that transaction, with the channel described below.
+- In `CListMempool`, introduce a new channel `txsRemoved` of type `chan
+  types.TxKey` to notify the mempool reactor that a transaction was removed from
+  the mempool.
+  - This is the same mechanism used to notify the consensus reactor when there
+  are transactions available to include in a block.
+- In the mempool reactor, introduce a new goroutine to handle incoming
+  transaction keys from the `txsRemoved` channel. For each key received, update
+  `txSenders`.
```
sergio-mena, line 102: Are we sure there's no possibility here (even if we're using locks) that the info flowing on the `txsRemoved` channel "conflicts" with info flowing in the other direction? What I'm thinking of (not 100% clear in my mind) is the following.
More or less at the same time:
* A TX (tx1) is removed and placed in the channel
* A newly received TX (tx2) passes `CheckTx` and is added to the mempool
* tx1 and tx2 have the same sender

I don't know, I have the impression that, rather than a channel and a goroutine, you want a callback set by the reactor, and called by the mempool implementation. But I'm not 100% sure.

docs/architecture/adr-105-refactor-mempool-senders.md
```diff
@@ -0,0 +1,141 @@
...
+stored internally.
+
+### Positive
+
+- Get rid of the complex design pattern of callbacks for handling
+  `CheckTxType_New` responses.
+- Clear separation of propagation protocol in the reactor and the mempool
+  implementation.
+- Allow for future improvements to both the propagation protocol and the mempool
+  implementation.
```
sergio-mena, line 121: ❤️

docs/architecture/adr-105-refactor-mempool-senders.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Negative
+
+None
```
sergio-mena, line 125: How confident are we that we are well covered by the current set of unit tests that are testing this code?
IOW, do we have a rough idea of:
* how costly it would be to fix the existing ones
* how many new ones we would have to write
* how likely it is that the (adapted) old ones and the new ones will catch problems in the refactoring

sergio-mena (approved):
Some comments to consider but overall I think:

* this is a very nice refactoring of the most intricate parts of the best understood reactor
* I can see the benefits of this when we start trying out improvements in Q3

I'm approving the PR, but also vote for accepting the ADR itself

docs/architecture/adr-105-refactor-mempool-senders.md
hvanz: I'm mostly certain that we will still use a mechanism like this. Now we keep track of the senders of each transaction. Probably we will expand this set to include any node that is known to have the transactions. In any case, this ADR can be superseded by another ADR.

docs/architecture/adr-105-refactor-mempool-senders.md
hvanz: Each transaction has its own list of sender IDs. So in your scenario there are no conflicts: tx1 will lose a sender, and tx2 will gain one, even when they are the sender. 

I didn't considered a callback instead of a channel and a goroutine. Probably it has its benefits or it's more idiomatic, but I'm not 100% sure either.

docs/architecture/adr-105-refactor-mempool-senders.md
hvanz: In the draft PR (#1010), I already fixed the existing tests. There were many changes, but not difficult. We'll certainly have to extend the test scenarios, at least to cover the new data structure `txSenders`, and probably a bit more. I still have to check what the tests are not covering.

In any case, I think all these are implementation details, not related to the architecture or design. What we could do is to add to the Neutral section something like "We need to extend the existing tests to cover new scenarios related to the new data structures and some potential concurrent issues".

sergio-mena: 💯

docs/architecture/adr-105-refactor-mempool-senders.md
```diff
@@ -0,0 +1,143 @@
...
+  peer ids (of type `uint16`). Add also a `cmtsync.RWMutex` lock to handle
+  concurrent accesses to the map.
+  - This refactoring should not change the fact that the list of senders live as
+    long as the transactions are in the mempool. When a transaction is received
+    by the reactor (either via RPC or P2P), we call `CheckTx`. We know whether a
+    transaction is valid and was included in the mempool by reading the `ReqRes`
+    response. If this is the case, record the list of senders in `txSenders`.
+    When a transaction is removed from the mempool, notify the reactor to remove
+    the list of senders for that transaction, with the channel described below.
+- In `CListMempool`, introduce a new channel `txsRemoved` of type `chan
```
lasarojc, line 95: Should add the change to the mempool struct, in terms of methods added, since txsRemoved will be internal to an implementation.

docs/architecture/adr-105-refactor-mempool-senders.md
```diff
@@ -0,0 +1,143 @@
...
+    handle the response, if needed.
+  - The callback parameter `cb` is no longer needed. Currently, this is mainly
+    used by the RPC endpoints `broadcast_tx_sync` and `broadcast_tx_commit`, and
+    in tests for checking that the response is valid. However, the same
+    functionality can be obtained with `ReqRes` response.
+  - `txInfo` contains information about the sender, which is also no longer
+    needed, as justified by the next point.
+- The list of senders for each transaction is currently stored in `mempoolTx`,
+  the data structure for the entries of `txs`. Move the senders out of
+  `mempoolTx` to a new map `txSenders` of type `map[types.TxKey]map[uint16]bool`
```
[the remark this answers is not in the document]
hvanz: In Go, the way to implement a set of elements of type X is with a `map[X]bool`.

docs/architecture/adr-105-refactor-mempool-senders.md
hvanz: That's right. I added it to the design.

docs/architecture/adr-105-refactor-mempool-senders.md
hvanz: I added the callback option as an alternative in the ADR. After playing with both implementations, now I think a callback will be the better option.

docs/architecture/adr-105-refactor-mempool-senders.md
```diff
@@ -0,0 +1,173 @@
...
+  type `CheckTxType_New`. Instead of setting it as an ad-hoc callback on each
+  transaction, we could now call it directly from `globalCb`, where responses of
+  type `CheckTxType_Recheck` are already being handled.
+
+### Communicating that a transaction was removed from the mempool
+
+We have identified two approaches for communicating the removal of a transaction
+from the mempool to the reactor.
+
+1. With a channel and an infinite loop in a gouroutine.
```
thanethomson, line 107:
```suggestion
1. With a channel and an infinite loop in a goroutine.
```

docs/architecture/adr-105-refactor-mempool-senders.md
```diff
@@ -0,0 +1,173 @@
...
+- We would need to extend the existing tests to cover new scenarios related to
+  the new data structures and some potential concurrent issues.
+
+## References
+
+The trigger for this refactoring was [this comment][comment], where we discussed
+improvements to the concurrency in the mempool.
+
+
+[CheckTxType]: https://github.com/cometbft/cometbft/blob/main/proto/tendermint/abci/types.proto#L94-L97
```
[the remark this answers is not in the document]
hvanz: Fixed.

hvanz:
> I'm a bit concerned about the idea of introducing goroutines into a struct that's supposed to represent a form of state for the mempool reactor. Ideally we should be progressively refactoring the codebase to remove all spawning of goroutines from any state-related structs, and pushing all such goroutine spawning and management exclusively to reactors.
> 
> Is that at all possible here?

Yes, in the latest version of the document there is the option of using callbacks. I think we will finally choose this option.

docs/architecture/adr-105-refactor-mempool-senders.md
```diff
@@ -0,0 +1,183 @@
...
+  `mempoolTx` to a new map `txSenders` of type `map[types.TxKey]map[uint16]bool`
+  in the mempool reactor. `txSenders` would map transaction keys to a set of
+  peer ids (of type `uint16`). Add also a `cmtsync.RWMutex` lock to handle
+  concurrent accesses to the map.
+  - This refactoring should not change the fact that the list of senders live as
+    long as the transactions are in the mempool. When a transaction is received
+    by the reactor (either via RPC or P2P), we call `CheckTx`. We know whether a
+    transaction is valid and was included in the mempool by reading the `ReqRes`
+    response. If this is the case, record the list of senders in `txSenders`.
+    When a transaction is removed from the mempool, notify the reactor to remove
```
lasarojc, line 96: Should remove ~, with the channel described below.~ since the channel is one of the options.

lasarojc (commented):
Just small fix on the text of this already approved ADR

PR #1033, opened 2023-06-26
Clarifies that processProposal may be called for set of transactions different from the one returned in the preceding prepareProposal

lasarojc:
If a proposer fails after calling prepareProposal and before calling processProposal, then the following may happen upon restarting:
- if failed before signing another message, then will invoke prepareProposal again, sign a new block, probably empty, and propose it;
- if failed after signing a proposal but before writing the  proposal message into the WAL, then will invoke prepareProposal and produce a new, probably empty block, [fail to sign it](https://github.com/cometbft/cometbft/blob/2789a59a9cc61c6ea56a6b266eeadf0f26ca2456/consensus/state.go#L1221), and not invoke processProposal; prevote timeouts will ensure the CometBFT is not stuck;
- if failed after writing the proposal message to the WAL, then will invoke prepareProposal, produce a new, probably empty block, fail to sign it, and invoke processProposal with the block signed before crashing.

---

#### PR checklist

- [ ] Tests written/updated


spec/abci/abci++_basic_concepts.md
```diff
@@ -200,25 +204,27 @@ and an [**Echo**](./abci++_methods.md#echo) method that is used for debugging.
...
-executing the block, the default value of *TimeoutPropose* might not be sufficient
-to accommodate the long block execution time and non-proposer nodes might time
-out and prevote `nil`. The proposal, in this case, will probably be rejected and a new round will be necessary.
+`PrepareProposal` and `ProcessProposal` stand on the consensus algorithm critical path,
+i.e., CometBFT cannot make progress while these methods are being executed.
+Hence, if the Application takes a long time preparing a proposal or in verifying it,
+the default values of *TimeoutPropose* and *TimeoutPrevote* might not be sufficient
+to accommodate these methods' executions and validator nodes might time out and prevote `nil`.
+The proposal, in this case, will probably be rejected and a new round will be necessary.
+If this happens successively, the liveness of CometBFT might be compromised.
```
sergio-mena, line 215: If the implementation of `PrepProp` and `ProcProp` are bounded in time, we don't introduce liveness problems (the algorithm is guaranteed to terminate), but performance problems (unwanted rounds).

spec/abci/abci++_basic_concepts.md
```diff
@@ -200,25 +204,27 @@ and an [**Echo**](./abci++_methods.md#echo) method that is used for debugging.
...
+## Proposal related timeouts
 
-Immediate execution requires the Application to fully execute the prepared block
-before returning from `PrepareProposal`, this means that CometBFT cannot make progress
-during the block execution.
-This stands on the consensus algorithm critical path: if the Application takes a long time
-executing the block, the default value of *TimeoutPropose* might not be sufficient
-to accommodate the long block execution time and non-proposer nodes might time
-out and prevote `nil`. The proposal, in this case, will probably be rejected and a new round will be necessary.
+`PrepareProposal` and `ProcessProposal` stand on the consensus algorithm critical path,
```
sergio-mena, line 209: Overall, I like this write-up more than the previous one.

Just on thing: you're introducing `ProcProp` in the reasoning, but the problem is way less serious with `ProcProp` than with `PrepProp`:

* At the beginning of a round, all non-proposers will schedule `TimeoutPropose` immediately
* however, with `ProcProp` the `TimeoutPropose` timeout is only scheduled after having received 2/3 of all prevotes. In practice, since the block tends to take a similar time to run at all validators, when a validator schedules `TimeoutPropose` it is likely to be about to finish its own `ProcProp` execution.

IOW, the _real_ problem is `PrepProp` and `TimeoutPropose`

spec/abci/abci++_basic_concepts.md
```diff
@@ -233,9 +239,12 @@ from block execution (`FinalizeBlock` calls), and not through
 any other kind of request. This is the only way to ensure all nodes see the same
 transactions and compute the same results.
 
-Some Applications may choose to implement immediate execution, which entails executing the blocks
+Some Applications may choose to execute blocks asynchronously to optimistically speed
```
sergio-mena, line 242: Did you mean synchronously?
If you execute blocks asynchronously, then you can't reject the invalid blocks.
Or maybe I'm not fully understanding

sergio-mena: Remember, what we achieve with immediate execution is the 100% guarantee that no TX included in a block will fail.
The app may choose to execute the block in parallel by replying `ACCEPT` right away, but then it cannot reject blocks based on its execution (in the current version of ABCI).

spec/abci/abci++_methods.md
```diff
@@ -444,6 +443,7 @@ the consensus algorithm will use it as proposal and will not call `RequestPrepar
       inform the Application of the block header's hash, which cannot be done at `PrepareProposal`
       time. In this case, the call to `RequestProcessProposal` occurs right after the call to
       `RequestPrepareProposal`.
+      However, there is no guarantee that the values of `RequestProcessProposal.txs` and `RequestProcessProposal.hash` will be the same as returned in the preceding `RequestPrepareProposal`.
```
sergio-mena, line 446: Excellent!

spec/abci/abci++_comet_expected_behavior.md
```diff
@@ -155,9 +155,9 @@ Let us now examine the grammar line by line, providing further details.
 >consensus-round     = proposer / non-proposer
 >```
 
-* For every round, if the local process is the proposer of the current round, CometBFT calls `PrepareProposal`, followed by `ProcessProposal`. 
+* For every round, if the local process is the proposer of the current round, CometBFT calls `PrepareProposal`, followed by `ProcessProposal`.
 These two always come together because they reflect the same proposal that the process
-also delivers to itself. 
+also delivers to itself.
```
sergio-mena, line 160: I think we need to adapt this as well

sergio-mena: Also, in the end, we agreed not to touch line 59 above, didn't we?
I think the final decision was to leave that line untouched (but not 100% sure, hence double-checking)

spec/abci/abci++_basic_concepts.md
lasarojc: Good point, @sergio-mena. Let me rephrase it to remove ProcProp.

spec/abci/abci++_basic_concepts.md
lasarojc: I really meant asynchronouusly, by replying ACCEPT right away. I modified the previous change to explicitly call out immediate execution (maybe a better name would be early execution) and the following change to discuss both immediate execution and optimistic/asynchronous/parallel execution.

spec/abci/abci++_comet_expected_behavior.md
lasarojc: I didn't change anything in this file because the grammar specifies the behavior in correct processes (that's why we agreed not to change line 59).
The main change hapened in the `example_scenarios` file, but I have now updated the text here. Let me know if I should revert.

spec/abci/abci++_basic_concepts.md
```diff
@@ -206,19 +206,27 @@ More details on managing state across connections can be found in the section on
...
-This is particularly important if applications need to execute the block to validate it.
-Then they would have to execute it in `PrepareProposal` and `ProcessProposal`.
 
-Operators will need to adjust the default values of *TimeoutPropose* and *TimeoutPrevote* in CometBFT's
-configuration file, in order to suit the needs of the particular application being deployed.
+Operators will need to adjust the default value of *TimeoutPropose* in CometBFT's configuration file,
+in order to suit the needs of the particular application being deployed.
+
+This is particularly important if applications implement *immediate execution*, a technique in
+which applications require the proposer to execute a block in `PrepareProposal` (and all validators to execute it in `ProcessProposal`) to ensure that all transactions are executable in case the block is decided.
```
cason, line 222: Can you break this line please?

cason: Actually, the validators have to verify the block execution in `ProcessProposal`.

cason: I didn't understand the second part of the sentence, starting from "to ensure that..."

spec/abci/abci++_basic_concepts.md
```diff
@@ -239,14 +247,13 @@ from block execution (`FinalizeBlock` calls), and not through
...
-up the execution of `FinalizeBlock` and `Commit`, in what is sometimes referred as *immediate execution*.
-Applications do so by executing the blocks
-that are about to be proposed (via `PrepareProposal`), and those that the Application is asked to
-validate (via `ProcessProposal`).
+Applications that implement immediate execution (execute the blocks
+that are about to be proposed, in `PrepareProposal`, or that require validation, in `ProcessProposal`) produce a new state before a block is decided.
 The state changes caused by processing those
 proposed blocks must never replace the previous state until `FinalizeBlock` confirms
-the block decided.
+the block decided and `Commit` is invoked in it.
```
cason, line 254:
```suggestion
that the proposed block was decided and `Commit` is invoked for it.
```

spec/abci/abci++_comet_expected_behavior.md
```diff
@@ -155,9 +155,8 @@ Let us now examine the grammar line by line, providing further details.
 >consensus-round     = proposer / non-proposer
 >```
 
-* For every round, if the local process is the proposer of the current round, CometBFT calls `PrepareProposal`, followed by `ProcessProposal`.
-These two always come together because they reflect the same proposal that the process
-also delivers to itself.
+* For every round, if the local process is the proposer of the current round, CometBFT calls `PrepareProposal` and then `ProcessProposal`.
+These two always come together because once a proposal is prepared it is broadcast to all validators, including the proposer itself. In the absence of restarts, `ProcessProposal` is invoked with the proposal returned by `PrepareProposal`.
```
cason, line 159: Sorry, it is not this part that we have to change?

cason: If the node is the proposer of the current round, we call `PrepareProposal`.

If the proposer doesn't crash, `ProcessProposal` is called for the same (prepared) proposal.

If the proposer crashes and then recovers, there are some possibilities, we cannot ensure in the current implementation this match of prepare and process the same proposal.

spec/abci/abci++_basic_concepts.md
sergio-mena: > all non-proposers will schedule TimeoutPropose immediately

Actually, that's wrong as we saw on Fri (it's true in the arXiv paper, but not in the code)

cason (commented):
I don't think this new redaction is really clarifying the situation mentioned in the title of this PR.

spec/abci/abci++_basic_concepts.md
sergio-mena: Ok I'll re-read in the next pass

spec/abci/abci++_comet_expected_behavior.md
lasarojc: It was changed. Let me try to make it more clear.

cason: Sorry, we had a concurrency problem here probably. Now I've read you preferred not update the wording here because the grammar says otherwise.

spec/abci/abci++_basic_concepts.md
lasarojc: We need to agree on how to break lines, if at the 80 or 100 char mark or something else.
I prefer breaking into periods, even if it makes lines longer.

I didn't catch this comment.
> Actually, the validators have to verify the block execution in ProcessProposal.

Do you mean that I should say validate instead of execute?
I guess I just expanded "validate" to "execute a block (...) to ensure all transactions are executable in case the block is decided."

cason: We can have two cases here:
1. Everyone executes the block in `ProcessProposal`, as a pre-execution that will actually be confirmed when the block is finally committed
2. The proposes execute the block on `PrepareProposal`, encodes somehow the results of the execution on the block itself, and non-proposers have to check whether the proposer's execution is valid, in `ProcessProposal`

I think we should discourage option 2, because it adds a lot of complexity and corner cases. My suggestion can be just to mention case 1 here.

lasarojc: We need to agree on how to break lines, if at the 80 or 100 char mark or something else.
I prefer breaking into periods, even if it makes lines longer.

I didn't catch this comment.
> Actually, the validators have to verify the block execution in ProcessProposal.

Do you mean that I should say validate instead of execute?
I guess I just expanded "validate" to "execute a block (...) to ensure all transactions are executable in case the block is decided."

cason: > to ensure that all transactions are executable in case the block is decided.

This part I really don't get: are we executing txs to ensure they can be executed? If we are executing them now, we don't execute them again upon Commit. Upon commit we only check whether the block that we previously executed is the same block that has been committed. If this is not the case, we execute the block on Commit replacing the (optimistic) execution we performed upon `ProcessProposal`.

spec/abci/abci++_basic_concepts.md
sergio-mena: > Ah, this is a very important information. What happens if the application doesn't reply on time? What is proposed?

CometBFT blocks until the application replies (similarly to `FinalizeBlock`). It's a **sync** ABCI call

spec/abci/abci++_basic_concepts.md
```diff
@@ -233,11 +246,13 @@ from block execution (`FinalizeBlock` calls), and not through
...
-that are about to be proposed (via `PrepareProposal`), and those that the Application is asked to
-validate (via `ProcessProposal`). However, the state changes caused by processing those
+Applications that implement immediate execution (execute the blocks
+that are about to be proposed, in `PrepareProposal`, or that require validation, in `ProcessProposal`) produce a new state before a block is decided.
+The state changes caused by processing those
 proposed blocks must never replace the previous state until `FinalizeBlock` confirms
-the block decided.
+that the proposed block was decided and `Commit` is invoked for it.
+
+The same is true to Applications that quickly accept blocks and execute the blocks optimistically in parallel with the remainder consensus steps to save time during `FinalizeBlock`; they must only apply state changes in `Commit`.
```
sergio-mena, line 255:
```suggestion
The same is true to Applications that quickly accept blocks and execute the blocks optimistically in parallel with the remaining consensus steps to save time during `FinalizeBlock`; they must only apply state changes in `Commit`.
```

Thanks for improving the wording here

spec/abci/abci++_basic_concepts.md
cason: So, timeout prevote does not apply here, as it is only triggered when the node receives 2/3+ votes that do not form a quorum.

spec/abci/abci++_comet_expected_behavior.md
```diff
@@ -155,9 +155,8 @@ Let us now examine the grammar line by line, providing further details.
 >consensus-round     = proposer / non-proposer
 >```
 
-* For every round, if the local process is the proposer of the current round, CometBFT calls `PrepareProposal`, followed by `ProcessProposal`. 
-These two always come together because they reflect the same proposal that the process
-also delivers to itself. 
+* For every round, if the local process is the proposer of the current round, CometBFT calls `PrepareProposal` and then `ProcessProposal`.
+These two always come together because once a proposal is prepared it is broadcast to all validators, including the proposer itself. In the absence of restarts, `ProcessProposal` is invoked with the proposal returned by `PrepareProposal`.
```
sergio-mena, line 159:
```suggestion
These two always come together because once a proposal is prepared it is broadcast to all validators, including the proposer itself. In the absence of crashes, `ProcessProposal` is invoked with the proposal returned by `PrepareProposal`.
```

spec/abci/abci++_basic_concepts.md
cason: > I still feel we shouldn't say that liveness is compromised (but performance is degraded). Thoughts?

I think it affects liveness somehow. The point is that timeouts are increasing, so eventually they become big enough so that to include the processing time involved on these calls.

lasarojc: > > I still feel we shouldn't say that liveness is compromised (but performance is degraded). Thoughts?
> 
> I think it affects liveness somehow. The point is that timeouts are increasing, so eventually they become big enough so that to include the processing time involved on these calls.

Good point. At some point then proposals will be made before the other validators prevote nil and eventually we decide. I agree with you that it is not a liveness issue, but a performance one. A potentially very painful one, since the timeout values are reset for each new round.

spec/abci/abci++_methods.md
sergio-mena, line 446:
```
      However, there is no guarantee that the values of `RequestProcessProposal.txs` and `RequestProcessProposal.hash` will be the same as returned in the preceding call to `PrepareProposal`.
```suggestion

Also, we may need to clarify that `RequestProcessProposal.hash` is not returned by `RequestPrepareProposal`, but calculated by CometBFT just after `PrepareProposal` returns.

**Review by sergio-mena — APPROVED** [review]
Thanks for this!

**Comment by sergio-mena** [comment]
Shall we backport this to `v0.38.x` and `v0.37.x`?

File: spec/abci/abci++_methods.md [hunk 11]
```diff
@@ -441,9 +440,9 @@ the consensus algorithm will use it as proposal and will not call `RequestPrepar
         * However, any resulting state changes must be kept as _candidate state_,
           and the Application should be ready to discard it in case another block is decided.
     * `RequestProcessProposal` is also called at the proposer of a round. The reason for this is to
-      inform the Application of the block header's hash, which cannot be done at `PrepareProposal`
-      time. In this case, the call to `RequestProcessProposal` occurs right after the call to
-      `RequestPrepareProposal`.
+      inform the Application of the block header's hash, which is only calculated by CometBFT after the execution of `PrepareProposal`.
+      The call to `RequestProcessProposal` occurs right after the call to `RequestPrepareProposal`.
+      However, there is no guarantee that the value of `RequestProcessProposal.txs` will be the same returned by the preceding `RequestPrepareProposal`.
```
lasarojc, line 445: Considering the second bullet in the description, the following may be a better explanation.

```suggestion
    * `RequestProcessProposal` is also called at the proposer of a round. The reason for this is to
      inform the Application of the block header's hash, which is only calculated by CometBFT after the execution of `PrepareProposal`.
      Normally the call to `RequestProcessProposal` occurs right after the call to `RequestPrepareProposal` and
      `RequestProcessProposal` matches the block produced based on `ResponsePrepareProposal` (i.e.,
      `RequestPrepareProposal` equals `RequestProcessProposal.txs`).
      However, no such guarantee is made since, in the presence of failures, `RequestProcessProposal` may match
      `ResponsePrepareProposal` from an earlier invocation or `ProcessProposal` may not be invoked at all.
```

sergio-mena: I would remove the sentence "the reason for this is to inform the Application of the block header's hash, which is only calculated by CometBFT after the execution of `PrepareProposal`."

Given the current situation, it only adds confusion IMO

lasarojc:
> Shall we backport this to `v0.38.x` and `v0.37.x`?

I think we should. I will wait for @cason's opinion and tag accordingly before merging.

cason (approved):
I am not sure whether the new grammar is precise, in this case, as it allows for individual calls to the pair Prepare/ProcessProposal, which I am not sure that is the case, even in replay mode.

In any case, I believe that the changes made do, in particular, make clear that there is no a necessary bond between the proposal prepared and processed by a proposer, as it can process at most a proposal per round of consensus (despite crashes), but it may prepare more than one proposal per round.

I have left some comments that should not block merging this PR.

spec/abci/abci++_comet_expected_behavior.md
```diff
@@ -247,22 +257,18 @@ of the usage of `retain_height` stay the same.
...
 historical commits and potential optimizations, are discussed in detail in [RFC-100](./../../docs/rfc/rfc-100-abci-vote-extension-propag.md#current-limitations-and-possible-implementations)
 
-## Handling upgrades to ABCI 2.0 
+## Handling upgrades to ABCI 2.0
 
-If applications upgrade to ABCI 2.0, CometBFT internally ensures that the [application setup](./abci%2B%2B_app_requirements.md#application-configuration-required-to-switch-to-abci-20) is reflected in its operation. 
-CometBFT retrieves from the application configuration the value of `VoteExtensionsEnableHeight`( *h<sub>e</sub>*,), 
-the height at which vote extensions are required for consensus to proceed, and uses it to determine the data it stores and data it sends to a peer 
-that is catching up.
+If applications upgrade to ABCI 2.0, CometBFT internally ensures that the [application setup](./abci%2B%2B_app_requirements.md#application-configuration-required-to-switch-to-abci-20) is reflected in its operation.
```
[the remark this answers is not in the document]
lasarojc: I agree it is confusing sometimes. If using GitHub's web interface, then it highlights where the differences are, but it is not perfect. In this case, there is really no change.

spec/abci/abci++_methods.md
```diff
@@ -441,9 +440,12 @@ the consensus algorithm will use it as proposal and will not call `RequestPrepar
         * However, any resulting state changes must be kept as _candidate state_,
           and the Application should be ready to discard it in case another block is decided.
     * `RequestProcessProposal` is also called at the proposer of a round. The reason for this is to
-      inform the Application of the block header's hash, which cannot be done at `PrepareProposal`
-      time. In this case, the call to `RequestProcessProposal` occurs right after the call to
-      `RequestPrepareProposal`.
+    * `RequestProcessProposal` is also called at the proposer of a round.
```
lasarojc, line 443:
```suggestion
```

spec/abci/abci++_comet_expected_behavior.md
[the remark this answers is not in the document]
cason: I was looking for a plugin for mardown diff, but I didn't find anything that can be loaded into GitHub...

spec/abci/abci++_comet_expected_behavior.md
```diff
@@ -56,7 +56,7 @@ consensus-exec      = (inf)consensus-height
 consensus-height    = *consensus-round decide commit
 consensus-round     = proposer / non-proposer
 
-proposer            = *got-vote [prepare-proposal process-proposal] [extend]
+proposer            = *got-vote [prepare-proposal] [process-proposal] [extend]
```
lasarojc, line 59:
```suggestion
proposer            = *got-vote [prepare-proposal [process-proposal]] [extend]
```

spec/abci/abci++_comet_expected_behavior.md
```diff
@@ -155,17 +156,26 @@ Let us now examine the grammar line by line, providing further details.
...
+  Receiving a proposal block triggers `ProcessProposal` with such a block.
 
   Then, optionally, the Application is
   asked to extend its vote for that round. Calls to `VerifyVoteExtension` can come at any time: the
   local process may be slightly late in the current round, or votes may come from a future round
   of this height.
 
 >```abnf
->proposer            = *got-vote [prepare-proposal process-proposal] [extend]
+>proposer            = *got-vote [prepare-proposal] [process-proposal] [extend]
```
lasarojc, line 178:
```suggestion
>proposer            = *got-vote [prepare-proposal [process-proposal]] [extend]
```

lasarojc:
Thanks for the reviews. I've fixed the grammar to have `ProcessProposal` follow `PrepareProposal`.
Merging the changes.

PR #1057, opened 2023-06-30
node/state:bootstrap state api

### LINKED ISSUE #884: Add Go API support to bootstrap node after it was state synced
### Context

The SDK recently added support to allow node operators to restore app state from a local snapshot (and additionally manipulate app snapshots).  For a more thorough context, see section "Application-implemented local state sync" from ADR 104 in #801.

As part of this work, the [`bootstrap-state`](https://github.com/cosmos/cosmos-sdk/pull/16061/files#diff-f946934088bd1a5b75846b7261731373d0b1f00ea507568ceb30fa8d624c6b4aR262) command in the SDK handles the bootstrapping of a fresh comet-based node. 

### Problem

This SDK `bootstrap-state` command comprises initializing the block store and state store, as well as light client verification. All of these steps rely on internal interfaces of CometBFT. 

### Proposal

Ideally, the SDK should be provided with a simpler and direct API from Comet that handles this bootstrapping functionality. We should add a public Go API method for this purpose.

jmalicevic:
---
This PR Closes #884  and is based on the work done by @yihuang  in #803 . We decided to accept their suggestion but provide it within an API rather than a separate command. 

The API provides users a way to synchronise with the application at a given height and set the stores to a proper initial state before starting up CometBFT. 

The `appHash` parameter is used for extra safety guarantees, if provided, we can check whether the application provided `appHash` matches the `appHash` retrieved via the light client **before** the node runs. This way users can re-execute statesync if it has been done offline or check the application state without having the node running. 

It can though be `nil`, and checked during regular startup. Note that in the case there is a mismatch, the node will not be able to start. 

Note that after bootstrapping, the state store has a height ahead of the blockstore. Which is invalid in regular consensus. That is why we add a new DB key that stores the last height at which statesync was performed. This is needed for the blocksync reactor to start syncing blocks from that height on and not from the beginning. 

Also it would panic due to the height mismatch otherwise. 

Instead of introducing a new keyword to mark this, we could read the height of the `seenCommit` which is stored in the blockstore after statesync. However, this might have unwanted implications for do not expect this value to be used in this way and they are not using offline statesync. 

Allowing a new key in the state store requires two additional functions and is thus API breaking. 

#### PR checklist

- [x] Tests written/updated. Tests are done manually in : https://github.com/cometbft/cometbft/tree/jasmina/884-bootstrap-state_test . To run the offline statesync scenario run `run_bootstrap_state_test.sh` in `test/e2e`. 


node/node.go
```diff
@@ -130,6 +133,82 @@ func StateProvider(stateProvider statesync.StateProvider) Option {
 	}
 }
 
+// This function is expected to synchronize CometBFTs
+// stores with the application after statesync has been performed offline
+// It is expected that the blockstore and statestore are empty at
+// the time the function is called
```
sergio-mena, line 139: What happens if the stores are not empty when this function is called?

node/node.go
```diff
@@ -130,6 +133,82 @@ func StateProvider(stateProvider statesync.StateProvider) Option {
...
+func BootstrapState(height uint64, appHash []byte, config *cfg.Config) error {
+	ctx := context.Background()
```
sergio-mena, line 141: Consider accepting the context from the caller (an extra param)

node/node.go
```diff
@@ -130,6 +133,82 @@ func StateProvider(stateProvider statesync.StateProvider) Option {
...
+	logger := log.NewTMLogger(log.NewSyncWriter(os.Stdout))
+	if config == nil {
+		config = cfg.DefaultConfig()
```
sergio-mena, line 144: Can we log a warning message in this case?

node/node.go
```diff
@@ -130,6 +133,82 @@ func StateProvider(stateProvider statesync.StateProvider) Option {
...
+	if err != nil {
+		return err
+	}
+
+	stateStore := sm.NewStore(stateDB, sm.StoreOptions{
+		DiscardABCIResponses: config.Storage.DiscardABCIResponses,
+	})
+
+	if !blockStore.IsEmpty() {
+		logger.Error("overwriting non empty block store")
```
sergio-mena, line 157: Can we bail out instead? and have the operator delete it (at their own risk)?

sergio-mena: Also, can we do the same check with the state store?

node/node.go
```diff
@@ -130,6 +133,82 @@ func StateProvider(stateProvider statesync.StateProvider) Option {
...
+// stores with the application after statesync has been performed offline
+// It is expected that the blockstore and statestore are empty at
+// the time the function is called
+func BootstrapState(height uint64, appHash []byte, config *cfg.Config) error {
+	ctx := context.Background()
+	logger := log.NewTMLogger(log.NewSyncWriter(os.Stdout))
+	if config == nil {
+		config = cfg.DefaultConfig()
+	}
+	blockStore, stateDB, err := initDBs(config, cfg.DefaultDBProvider)
```
sergio-mena, line 146: Are the SDK folks OK with `cfg.DefaultDBProvider` hardcoded here, or do they prefer to pass it as a parameter?

node/node.go
```diff
@@ -130,6 +133,82 @@ func StateProvider(stateProvider statesync.StateProvider) Option {
...
+	err = blockStore.SaveSeenCommit(state.LastBlockHeight, commit)
+	if err != nil {
+		return err
+	}
+
+	if appHash == nil {
+		logger.Info("warning: cannot verify appHash. Verification will happen when node boots up!")
+	} else {
+		if !bytes.Equal(appHash, state.AppHash) {
+			logger.Error("the app hash returned by the light client does not match the provided appHash, expected %X, got %X", state.AppHash, appHash)
```
sergio-mena, line 205: I would return an error in this case

node/node.go
```diff
@@ -130,6 +133,82 @@ func StateProvider(stateProvider statesync.StateProvider) Option {
...
+	commit, err := stateProvider.Commit(ctx, height)
+	if err != nil {
+		return err
+	}
+
+	if err := stateStore.Bootstrap(state); err != nil {
+		return err
+	}
+
+	// store.SaveBlockStoreState(&cmtstore.BlockStoreState{
```
yihuang, line 190: Does it work without this line?

jmalicevic: From checking the statesync code atm, I did not find a place where `base` and `height` are higher than 0 after statesync. I tested this locally in one of our e2e tests but am writing a test to commit as well. Even in the initial handshake if `store.LastBlockHeight == 0` we just check the `appHash` and do not return an error.  But I am writing a test to confirm this and to confirm that bootstrapping like this after statesync will work (I want to run statesync, crash the node, delete the stores and run this).

jmalicevic: But this is only correct if the store is empty beforehand (which is stated as expectation in the spec), and that is why Sergio is right above, that non empty stores should lead to an error indicating to the operator that they have to make sure this condition is met.

jmalicevic: @yihuang , you are in fact right, due to the way we boot up the Blocksync reactor [here](https://github.com/cometbft/cometbft/blob/70d0a65993c622adeb1cb6bfe5b48d9acd117458/blocksync/reactor.go#L72C15-L72C15), there will be an error if we do not call the code I commented out above. 

You actually said this in your other PR, that if a crash happens even in regular state-sync before the block is synced, the node will not start-up. I reproduced the case and now understand what you meant, apologies. But by definition this should not happen and we should have a more graceful handling of this. Setting the height and base like this violates their definition :"Height of first and last **block** in the block store" .

This is a bug and we should address it.

node/node.go
jmalicevic: I added the check to confimr that the state is indeed empty.

node/node.go
jmalicevic: Added it as a parameter. Although currently both in our code and in the SDK code it is the DefaultDBProvider.

cason (commented):
Legit.

At the end, implementing this properly has revealed much more complex than expected. I wonder, however, whether we should add some unit tests.

yihuang:
Will it be easier if we make a special case in block store state, instead of create a new kv entry?

node/node.go
```diff
@@ -130,6 +133,109 @@ func StateProvider(stateProvider statesync.StateProvider) Option {
 	}
 }
 
+// This function is expected to synchronize CometBFTs
+// stores with the application after statesync has been performed offline
+// It is expected that the blockstore and statestore are empty at
+// the time the function is called.
+// In case the blockstore is not empty, the function returns an error
```
thanethomson, line 140:
```suggestion
// BootstrapState synchronizes the stores with the application after state sync
// has been performed offline. It is expected that the block store and state
// store are empty at the time the function is called.
//
// If the block store is not empty, the function returns an error.
```

Publicly exported functions' doc strings must start with the name of the function (`BootstrapState`, in this case).

.changelog/unreleased/breaking-changes/1057-bootstrap-state-api.md
```diff
@@ -0,0 +1,4 @@
+- `[node/state]` Add API to bootstrap blockstore and statestore to a height. Make sure blocksync starts syncing from bootstrapped height. 
+  ([\#1057](https://github.com/tendermint/tendermint/pull/#1057)) (@yihuang)
+- `[state/store]` (api-breaking) Added functions to save height at which offline statesync is performed.
```
thanethomson, line 3:
```suggestion
- `[state/store]` Added functions to save height at which offline state sync is performed. 
```

.changelog/unreleased/breaking-changes/1057-bootstrap-state-api.md
```diff
@@ -0,0 +1,4 @@
+- `[node/state]` Add API to bootstrap blockstore and statestore to a height. Make sure blocksync starts syncing from bootstrapped height. 
+  ([\#1057](https://github.com/tendermint/tendermint/pull/#1057)) (@yihuang)
```
thanethomson, line 2:
```suggestion
- `[node/state]` Add API to bootstrap block store and state store to a height. Make sure block sync starts syncing from bootstrapped height. 
  ([\#1057](https://github.com/tendermint/tendermint/pull/#1057)) (@yihuang)
```

.changelog/unreleased/features/1057-bootstrap-state-api.md
```diff
@@ -0,0 +1,2 @@
+- `[node/state]` Add API to bootstrap blockstore and statestore to a height
```
thanethomson, line 1:
```suggestion
- `[node/state]` Add API to bootstrap block store and state store to a height
```

node/node.go
```diff
@@ -130,6 +133,109 @@ func StateProvider(stateProvider statesync.StateProvider) Option {
...
+func BootstrapState(ctx context.Context, config *cfg.Config, dbProvider cfg.DBProvider, height uint64, appHash []byte) error {
```
thanethomson, line 141: How do we anticipate this functionality will eventually make its way to operators so they can actually use it? Is the assumption that the SDK will make use of it? And if so, what about non-SDK users?

thanethomson (commented):
We definitely need tests for this (even if it's just E2E tests).

jmalicevic:
> We definitely need tests for this (even if it's just E2E tests).

For what it's worth I tested this thoroughly locally, and am trying to figure out whether we can automate this testing somehow. 

Basically here is what I did:
- I run an e2e test where one node is statesyncing and running local code, the other nodes run the latest e2e-node. 
- Edit the code to exit after statesync is done and introduce a config flag `offlineStateSyncHeight`
- Delete the statestore and blockstore of that node
- Boot the node again where the code is setup to call `BootstrapSate (offlineStateSyncHeight)` if `OfflineStateSyncHeight == 0`. Node will shut down after this
---- The above procedure simulates offline statesync and bootstraping the nodes. 
- Start up the node . It should run as any other node. I tried also a variation with explicit disabling of blocksync to make sure that the node is in proper state for consensus in case blocksync times out. 

I think I could put this code in a separate branch and create an e2e manifest with a script that can repeat the behaviour I just described above.

jmalicevic:
@yihuang indeed, we considered this because the end result is the same. If the blockstore Base and HEight are 0 and the lastCommit height is equal to state store height we could accept that as a valid condition. We were just afraid that this might open up a corner case that we are not aware of and break something , allowing the blocksync reactor be created if the state of the store is like this but there was no offline statesync. Although there is no correct scenario where this can happen.

blocksync/reactor.go
```diff
@@ -68,17 +68,35 @@ type Reactor struct {
...
+	if storeHeight == 0 {
+		// If state sync was performed offline and the stores were bootstrapped to height H
+		// the state stores lastHeight will be H while blockstores Height and Base are still 0
+		// 1. This scenario should not lead to a panic in this case, which is indicated by
+		// having a OfflineStateSyncHeight > 0
+		// 2. We need to instruct the blocksync reactor to start fetching blocks from H+1
+		// instead of 0.
+		storeHeight, err = blockExec.Store().GetOfflineStateSyncHeight()
+		if err != nil && err.Error() != "value empty" {
+			panic(fmt.Sprintf("failed to retrieve statesynced height from store %s", err))
```
sergio-mena, line 83: Could you also add `state.LastBlockHeight` somehow in the panic message (since it was appearing before)

blocksync/reactor.go
```diff
@@ -68,17 +68,35 @@ type Reactor struct {
...
+		}
+		err = blockExec.Store().SetOfflineStateSyncHeight(0)
+		if err != nil {
+			panic("failed to reset the offline state sync height ")
+		}
```
sergio-mena, line 88: For context, @jmalicevic and I discussed the case when the process crashes precisely here. In that case, when restarting the process, it will fail at line 83 above. However, we decided not to try to solved this here because:

1. This problem also exists (it always has, since introduction of state-sync) with the regular (online) bootstrapping
2. This would occur when an operator tries to start a new node from scratch (bootstrapping either online or via a snapshot from the SDK). So it would happen immediately after the operator types the command. Therefore, a good error message should suffice

blocksync/reactor.go
sergio-mena: Following my comment below. Do you think we can write a more expressive message to make sure the operator know what to do if they see this message? (likely, wipe out the data dir and start over)

node/node.go
sergio-mena: In my view, if non-SDK users come to us asking for offline state sync we will need to prioritize the full Comet-based solution described as the main solution in [ADR-104](https://github.com/cometbft/cometbft/blob/main/docs/architecture/adr-104-out-of-band-state-sync.md).

node/node.go
```diff
@@ -130,6 +133,114 @@ func StateProvider(stateProvider statesync.StateProvider) Option {
...
+
+	// Once the stores are bootstrapped, we need to set the height at which the node has finished
+	// statesyncing. This will allow the blocksync reactor to fetch blocks at a proper height.
+	// In case this operation fails, it is equivalent to a failure in  online state sync where the operator
+	// needs to manually delete the state and blockstores and rerun the bootstrapping process.
+	err = stateStore.SetOfflineStateSyncHeight(state.LastBlockHeight)
+	if err != nil {
+		return fmt.Errorf("failed to set synced height")
+	}
+	if err := blockStore.Close(); err != nil {
```
sergio-mena, line 234:
```suggestion
	if err = blockStore.Close(); err != nil {
```

node/node.go
```diff
@@ -130,6 +133,114 @@ func StateProvider(stateProvider statesync.StateProvider) Option {
...
+		logger.Error("failed to close blockstore")
+	}
+	if err := stateStore.Close(); err != nil {
+		logger.Error("failed to close statestore")
+	}
+
+	return nil
```
sergio-mena, line 241:
```suggestion
	if err == nil {
		return err2
	}
	return err
```

state/store.go
```diff
@@ -73,6 +75,10 @@ type Store interface {
 	Bootstrap(State) error
 	// PruneStates takes the height from which to start pruning and which height stop at
 	PruneStates(int64, int64, int64) error
+	// Saves the height at which the store is bootstrapped after out of band statesync
```
sergio-mena, line 78:
```suggestion
	// Saves the height at which the store is bootstrapped after out of band statesync, a value of zero clears the data
```

state/store.go
```diff
@@ -41,6 +42,7 @@ func calcABCIResponsesKey(height int64) []byte {
 //----------------------
 
 var lastABCIResponseKey = []byte("lastABCIResponseKey")
+var offlineStateSyncHeight = []byte("offlineStateSyncH")
```
sergio-mena, line 45: Consider renaming to `offlineStateSyncHeightKey`

state/store.go
```diff
@@ -683,6 +689,31 @@ func (store dbStore) saveConsensusParamsInfo(nextHeight, changeHeight int64, par
...
+	if err != nil {
+		return 0, err
+	}
+
+	if len(buf) == 0 {
+		return 0, errors.New("value empty")
+	}
+
+	height := int64FromBytes(buf)
+	return height, nil
```
sergio-mena, line 714: I would do an extra sanity check that it is not negative (and maybe also at the beginning of `SetOfflineStateSyncHeight`)... with a panic, of course

sergio-mena (approved):
Awesome job, @jmalicevic !

state/store.go
jmalicevic: Changed the code but I did not add the panic here, it will panic from where it is called once the error is returned.

node/node.go
```diff
@@ -130,6 +133,114 @@ func StateProvider(stateProvider statesync.StateProvider) Option {
...
+	}
+
+	if err := stateStore.Bootstrap(state); err != nil {
+		return err
+	}
+
+	err = blockStore.SaveSeenCommit(state.LastBlockHeight, commit)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
jmalicevic: You were right to suggest this. I played a bit with the various heights at which vote extensions are enabled. The test fails if `vote_extensions_enable_height` is `0` or  below the height at which the node is switching to blocksync and consensus. The reason is the check here:https://github.com/cometbft/cometbft/blob/868b71e66addbf5721e12fa741f8c70ed5b35a34/consensus/state.go#L183 . This condition is normally never true unless there was no crash during statesync and consensus was running. I will think a bit in the morning how to address this. So if it were to get to blocksync,, it would work, but it does not get to that point.

node/node.go
```diff
@@ -130,6 +133,116 @@ func StateProvider(stateProvider statesync.StateProvider) Option {
...
+	// In case this operation fails, it is equivalent to a failure in  online state sync where the operator
+	// needs to manually delete the state and blockstores and rerun the bootstrapping process.
+	err = stateStore.SetOfflineStateSyncHeight(state.LastBlockHeight)
+	if err != nil {
+		return fmt.Errorf("failed to set synced height")
+	}
+	if err = blockStore.Close(); err != nil {
+		logger.Error("failed to close blockstore")
+	}
+	var err2 error
```
cason, line 237: Do we really need a different error variable here? We can define a new local `err` inside the `if` clause.

node/node.go
```diff
@@ -144,11 +257,11 @@ func NewNode(ctx context.Context,
 	logger log.Logger,
 	options ...Option,
 ) (*Node, error) {
+
```
cason, line 260: Not really important, but can you revert the newline changes in the method, which have absolute no practical effect but will put this PR on the blame for those lines?

state/store.go
```diff
@@ -706,3 +748,14 @@ func responseFinalizeBlockFromLegacy(legacyResp *cmtstate.LegacyABCIResponses) *
 		// be caught and filled in consensus/replay.go
 	}
 }
+
+func int64FromBytes(bz []byte) int64 {
```
cason, line 752: I wonder whether a test unit could be added to verify whether saved and loaded `int64` values are the same.

cason (approved):
Legit, good work Jasmina. 

I think this solution covers several corner cases not considered in the original PR. While not a perfect solution, it does not adds additional corner cases not already present when using the original (online) state sync. Addressing the corner cases of both forms of state sync is a more complex task, that doesn't need to be addressed by this PR.

I've left some comments, mostly cosmetic. In addition to them, I find it useful that the description of this PR contains a pointer to the testing branch that was used to "manually" verify the proper operation of the introduced API.

jmalicevic:
Converting it to draft until we resolve the corner case regarding vote extensions and are sure about the safety of the proposed approach.

consensus/state.go
```diff
@@ -180,7 +180,24 @@ func NewState(
...
+		// it will try to reconstruct the extended commit here.
+		storeHeight, err := cs.blockExec.Store().GetOfflineStateSyncHeight()
+		if err != nil && err.Error() != "value empty" {
+			panic(fmt.Sprintf("failed to retrieve statesynced height from store %s", err))
+		}
+		if storeHeight != 0 {
+			cs.reconstructSeenCommit(state)
+			err = cs.blockExec.Store().SetOfflineStateSyncHeight(0)
+			if err != nil {
+				panic("failed to reset the offline state sync height ")
```
sergio-mena, line 196: Please include the error in the panic message

consensus/state.go
```diff
@@ -560,21 +577,29 @@ func (cs *State) sendInternalMessage(mi msgInfo) {
 	}
 }
 
+// ReconstructSeenCommit reconstructs the seen commit
+// This function is meant to be called only after statesync
```
sergio-mena, line 581:
```suggestion
// This function is meant to be called after statesync
```

sergio-mena (approved):
Just went through the whole PR again. All my concerns are gone :-)
Excellent work!

blocksync/reactor.go
```diff
@@ -66,22 +66,18 @@ type Reactor struct {
...
 // NewReactor returns new reactor instance.
 func NewReactor(state sm.State, blockExec *sm.BlockExecutor, store *store.BlockStore,
-	blockSync bool, metrics *Metrics,
+	blockSync bool, metrics *Metrics, offlineStateSyncHeight int64,
 ) *Reactor {
 
 	storeHeight := store.Height()
-	var err error
-	if storeHeight == 0 {
+	if offlineStateSyncHeight != 0 {
```
sergio-mena, line 73:
```suggestion
	if storeHeight == 0 {
```

node/node.go
```diff
@@ -248,18 +361,29 @@ func NewNode(ctx context.Context,
 		sm.BlockExecutorWithMetrics(smMetrics),
 	)
 
+	offlineStateSyncHeight := int64(0)
+	if blockStore.Base() == 0 {
```
jmalicevic, line 365:
```suggestion
	if blockStore.Height() == 0 {
```

blocksync/reactor.go
jmalicevic: I just added a fix for the code in `node.go` which should already take care of this check: https://github.com/cometbft/cometbft/pull/1057/files#r1285792914 

If storeHeight is not 0 `offlineStateSyncHeight` will be 0. But I don't mind repeating the check here.

consensus/state.go
```diff
@@ -560,21 +577,29 @@ func (cs *State) sendInternalMessage(mi msgInfo) {
...
+	cs.LastCommit = votes
+}
+
 // Reconstruct the LastCommit from either SeenCommit or the ExtendedCommit. SeenCommit
 // and ExtendedCommit are saved along with the block. If VoteExtensions are required
 // the method will panic on an absent ExtendedCommit or an ExtendedCommit without
 // extension data.
 func (cs *State) reconstructLastCommit(state sm.State) {
 	extensionsEnabled := state.ConsensusParams.ABCI.VoteExtensionsEnabled(state.LastBlockHeight)
+
```
jmalicevic, line 598:
```suggestion
```

node/node.go
```diff
@@ -146,11 +259,11 @@ func NewNode(ctx context.Context,
 	logger log.Logger,
 	options ...Option,
 ) (*Node, error) {
+
```
jmalicevic, line 262:
```suggestion
```

node/node.go
```diff
@@ -146,11 +259,11 @@ func NewNode(ctx context.Context,
...
 	blockStore, stateDB, err := initDBs(config, dbProvider)
 	if err != nil {
 		return nil, err
 	}
-
 	stateStore := sm.NewStore(stateDB, sm.StoreOptions{
```
jmalicevic, line 267:
```suggestion
	
	stateStore := sm.NewStore(stateDB, sm.StoreOptions{
```

state/store.go
jmalicevic: Done!

.changelog/unreleased/breaking-changes/1057-bootstrap-state-api.md
```diff
@@ -0,0 +1,4 @@
+- `[node/state]` Add API to bootstrap block store and state store to a height. Make sure block sync starts syncing from bootstrapped height. 
+  ([\#1057](https://github.com/tendermint/tendermint/pull/#1057)) (@yihuang)
+- `[state/store]` Added functions to save height at which offline state sync is performed. 
+  ([\#1057](https://github.com/tendermint/tendermint/pull/#1057)) (@jmalicevic)
```
thanethomson, line 4:
```suggestion
- `[node/state]` Add Go API to bootstrap block store and state store to a height. Make sure block sync starts syncing from bootstrapped height. 
  ([\#1057](https://github.com/tendermint/tendermint/pull/#1057)) (@yihuang)
- `[state/store]` Added Go functions to save height at which offline state sync is performed. 
  ([\#1057](https://github.com/tendermint/tendermint/pull/#1057)) (@jmalicevic)
```

.changelog/unreleased/features/1057-bootstrap-state-api.md
```diff
@@ -0,0 +1,2 @@
+- `[node/state]` Add API to bootstrap block store and state store to a height
+  ([\#1057](https://github.com/tendermint/tendermint/pull/#1057)) (@yihuang)
```
thanethomson, line 2:
```suggestion
- `[node/state]` Add Go API to bootstrap block store and state store to a height
  ([\#1057](https://github.com/tendermint/tendermint/pull/#1057)) (@yihuang)
```

node/node.go
```diff
@@ -132,6 +135,116 @@ func StateProvider(stateProvider statesync.StateProvider) Option {
...
+	if config == nil {
+		logger.Info("no config provided, using default configuration")
+		config = cfg.DefaultConfig()
+	}
+
+	if dbProvider == nil {
+		dbProvider = cfg.DefaultDBProvider
+	}
+	blockStore, stateDB, err := initDBs(config, dbProvider)
+
```
thanethomson, line 158: Is it not safer to do:

```suggestion
	blockStore, stateDB, err := initDBs(config, dbProvider)
	defer blockStore.Close()
```

node/node.go
```diff
@@ -132,6 +135,116 @@ func StateProvider(stateProvider statesync.StateProvider) Option {
...
+	if err != nil {
+		return err
+	}
+	if !blockStore.IsEmpty() {
+		return fmt.Errorf("blockstore not empty, trying to initialize non empty state")
+	}
+
+	stateStore := sm.NewStore(stateDB, sm.StoreOptions{
+		DiscardABCIResponses: config.Storage.DiscardABCIResponses,
+	})
```
thanethomson, line 168: Similarly to my other comment, is it not safer to:

```suggestion
	stateStore := sm.NewStore(stateDB, sm.StoreOptions{
		DiscardABCIResponses: config.Storage.DiscardABCIResponses,
	})
	defer stateStore.Close()
```

node/node.go
```diff
@@ -132,6 +135,116 @@ func StateProvider(stateProvider statesync.StateProvider) Option {
...
+		return fmt.Errorf("failed to set synced height")
+	}
+	if err = blockStore.Close(); err != nil {
+		logger.Error("failed to close blockstore")
+	}
+	var err2 error
+	if err2 = stateStore.Close(); err2 != nil {
+		logger.Error("failed to close statestore")
+		return err2
+	}
```
thanethomson, line 243: It looks to me as though all these manual close operations spread throughout this function can simply be replaced by the `defer` statements in my other two comments.

In general it's better to do this sort of thing in a `defer` because, even if the function panics somewhere, the `defer` statements will still be called. In the current version of this function, that won't happen, and it's not clear to me what the impact on the stores would be as a result.

jmalicevic: Yes, if they are not closed properly then re-opening them will be a problem. The only problem is what if close fails. But we can have another function that tries to close them, returning an error and call this function with defer. 

`defer func() (error) {

if err = blockStore.Close(); err != nil {
		logger.Error("failed to close blockstore")
	}
var err2 error
if err2 = stateStore.Close(); err2 != nil {
		logger.Error("failed to close statestore")
		return err2
}
return err
}
`

thanethomson: It's always best to separate each close operation into separate `defer` statements. Also, the `(error)` return value for the function would end up being ignored. If the enclosing function doesn't need to know about the errors, one could simply:

```go
defer func() {
	if err := blockStore.Close(); err != nil {
		logger.Error("Failed to close blockstore", "err", err)
	}
}()
// ...
defer func() {
	if err := stateStore.Close(); err != nil {
		logger.Error("Failed to close statestore", "err", err)
	}
}()
```

If the enclosing function _does_, however, need to return any errors produced from within the `defer` statements, you'd have to name the return value in the function signature and use it directly in the `defer` statements:

```go
func BootstrapState(
	ctx context.Context,
	config *cfg.Config,
	dbProvider cfg.DBProvider,
	height uint64,
	appHash []byte,
) (err error) {
	// ...
	defer func() {
		if derr := blockStore.Close(); derr != nil {
			logger.Error("Failed to close blockstore", "err", derr)
			// Set the return value
			err = derr
		}
	}()

	// ...
	defer func() {
		if derr := stateStore.Close(); derr != nil {
			logger.Error("Failed to close statestore", "err", derr)
			err = derr
		}
	}()

	// ...
}
```

Both `defer` statements will be executed, and whichever of them errors last will influence the actual function result.

jmalicevic: Thanks, will refactor this then. I think we should return errors so I'll take the second approach.

jmalicevic: @thanethomson is the current way of closing the stores ok? So we can merge this PR:)

jmalicevic:
As there have been no recent comments, and this is needed for cutting the 0.38 release, this will be merged, any potential issues with the PR will be solved in follow up PRs.

node/node.go
```diff
@@ -132,6 +135,124 @@ func StateProvider(stateProvider statesync.StateProvider) Option {
...
+		if !bytes.Equal(appHash, state.AppHash) {
+			if err := blockStore.Close(); err != nil {
+				logger.Error("failed to close blockstore: %w", err)
+			}
+			if err := stateStore.Close(); err != nil {
+				logger.Error("failed to close statestore: %w", err)
+			}
+			return fmt.Errorf("the app hash returned by the light client does not match the provided appHash, expected %X, got %X", state.AppHash, appHash)
+		}
+	}
```
tac0turtle, line 228: why does this need to check the apphash?

jmalicevic: It is not required that this check is performed. It is an additional safety mechanism which can be used by the app, who should know the apphash to choose to provide the appHash to the function. If the apphash is provided then we check it matches the hash obtained by the light client. If no appHash is provided to the function (L217), we just log an error. 

This extra check can let the operator know something's wrong without having to start comet. The appHash is checked anyways when comet starts and if there is a mismatch it won't start.

PR #1073, opened 2023-07-03
Fix build on mac

Taction:
Run `make build` on Mac encountered the following error:
```
CGO_ENABLED=0 go build -mod=readonly -ldflags "-X github.com/cometbft/cometbft/version.TMGitCommitHash=b23ef56f8 -L/opt/homebrew/opt/llvm@15/lib -s -w" -trimpath -tags 'cometbft' -o /Users/zc/Desktop/Workspace/zhongtai/github/cometbft/cometbft/build/cometbft ./cmd/cometbft/
# github.com/cometbft/cometbft/cmd/cometbft
flag provided but not defined: -L/opt/homebrew/opt/llvm@15/lib
usage: link [options] main.o
```

Added a fix that may solve this.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -6,7 +6,9 @@ COMMIT_HASH := $(shell git rev-parse --short HEAD)
 LD_FLAGS = -X github.com/cometbft/cometbft/version.TMGitCommitHash=$(COMMIT_HASH)
 BUILD_FLAGS = -mod=readonly -ldflags "$(LD_FLAGS)"
 # allow users to pass additional flags via the conventional LDFLAGS variable
-LD_FLAGS += $(LDFLAGS)
+ifneq ($(strip $(LDFLAGS)),)
+	LD_FLAGS+=-extldflags=$(LDFLAGS)
+endif
 
 # handle nostrip
 ifeq (nostrip,$(findstring nostrip,$(COMETBFT_BUILD_OPTIONS)))


sergio-mena:
The flag `-L/opt/homebrew/opt/llvm@15/lib` is not added in the default setup for building CometBFT. When I run `make build` I don't see this flag (I do, however see it, and the error reported, if I set it in the `make` command).
Further, make variable `LD_FLAGS` is only supposed to pick thing up from the `make` command line, not from the user's env.
So I'd like to understand what element of your setup produces a build line that includes `-L/opt/homebrew/opt/llvm@15/lib`

Taction:
> The flag -L/opt/homebrew/opt/llvm@15/lib is not added in the default setup for building CometBFT.

Yes, you are right. I figured out that I manually set `LDFLAGS` to the machine's environment variable previously, which is not related to this project. 
In my case, there is some other project that needs llvm, so I installed it via homebrew and set `LDFLAGS=-L/opt/homebrew/opt/llvm@15/lib` to default environment variables. If `LDFLAGS` is set the makefile will use, and causes this issue.

sergio-mena:
I see.
Yes, I managed to repro the error by doing  `export LDFLAGS=-L/opt/homebrew/opt/llvm@15/lib`, then `make build`.

In that case, the simplest solution is to do `make LDFLAGS= build` for building `cometbft`.

Taction:
Sorry for the late reply.

Thanks and your solution definitely works! The Makefile appends `LDFLAGS` in L9 `LD_FLAGS += $(LDFLAGS)`. Anyone who has this environment variable would be affected. Delete this line or add a fix for it may reduce the friction for users to build it.

Issue #1074, opened 2023-07-03
rollback set incorrect `LastHeightValidatorsChanged`


yihuang:
### Setup

**CometBFT version** all versions that support rollback

**Have you tried the latest version**: yes

**ABCI app** (name for built-in, URL for self-written if it's publicly available):

**Environment**:
- **OS** (e.g. from /etc/os-release):
- **Install tools**:
- **Others**:

**node command runtime flags**:

thanethomson:
Please complete the template here, or at least provide us with some information on the bug and how to reproduce it.

arno01:
@yihuang (shot in the dark) could be missing https://github.com/cometbft/cometbft/pull/1610 maybe?

yihuang:
Totally forgot about this, let me finish the PR tmr

melekes:
Closing in favor of https://github.com/cometbft/cometbft/issues/2137 @yihuang let's coordinate there.

PR #1075, opened 2023-07-03
fix: rollback set incorrect `LastHeightValidatorsChanged`

yihuang:
Closes: #1074

Solution:
- persist the new validator set if changed.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/.changelog/unreleased/bug-fixes/1075-fix-rollback-validator-set.md
+++ b/.changelog/unreleased/bug-fixes/1075-fix-rollback-validator-set.md
@@ -1,0 +1,1 @@
+- `[state]` \#1075 When rollback, persist the new validator set if changed. (@yihuang)
--- a/state/rollback.go
+++ b/state/rollback.go
@@ -66,9 +66,9 @@ func Rollback(bs BlockStore, ss Store, removeBlock bool) (int64, []byte, error) 
 	}
 
 	valChangeHeight := invalidState.LastHeightValidatorsChanged
-	// this can only happen if the validator set changed since the last block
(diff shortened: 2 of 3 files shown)


yihuang:
Unstale, it's a bug

yihuang:
> Hey @yihuang do you know of any way to test this, or any way to reproduce the bug? That would help us a lot to feel confident about merging this PR.

when rollback, the just rollback-ed `LastHeightValidatorsChanged` field could be wrong, according to the code logic, validator set in the state could be skipped if not changed, it'll reference the height that actually stored the validator set in the field `LastHeightValidatorsChanged`, in current rollback logic, the rollback-ed `LastHeightValidatorsChanged` could be wrong, because it don't guarantee that height do stored the validator set, the patched version will force the rollback-ed version to store the validator set.

sergio-mena:
@yihuang, would you mind?
* Writing a UT that exposes the problem, and demonstrates that this PR fixes it (i.e., UT fails without the fix, and passes with it)
* Adding the info you provided in this PR in the isue itself (#1074) for traceability purposes? (Including [this comment](https://github.com/cometbft/cometbft/pull/1075#issuecomment-1654903582) in the issue's description should suffice)

mmsqe:
I try add a [case](https://github.com/mmsqe/cometbft/compare/1207a24c6c70a18950169501e86ed6872d2009a5...f1bcf8ec94414c9a30b4cb17c2dd54742f9d3bc5) for last height > 2, see if it helps to reproduce.

PR #1079, opened 2023-07-04
Log proposer's address when correctly accepting a proposal

sergio-mena:
Contributes to #907 

While troubleshooting today, a key info we needed was missing in our Info logs

Tested via e2e:
```
full02       | I[2023-07-04|14:27:28.059] received proposal                            module=consensus proposal="Proposal{1059/1 (2DF98F8362C45180CEF86031DA7C5CBA7AAA2DBD52FF889DC80C508DFEC5CDC7:1:77AEA1FD157
2, -1) 24AD3C5810DE @ 2023-07-04T14:27:27.804019764Z}" proposer=07EB27AE0D0607DB338F1F102B3C25049FE04C68
```

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -1899,7 +1899,7 @@ func (cs *State) defaultSetProposal(proposal *types.Proposal) error {
 		cs.ProposalBlockParts = types.NewPartSetFromHeader(proposal.BlockID.PartSetHeader)
 	}
 
-	cs.Logger.Info("received proposal", "proposal", proposal)
+	cs.Logger.Info("received proposal", "proposal", proposal, "proposer", cs.Validators.GetProposer().Address)
 	return nil
 }
 
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -1039,7 +1039,7 @@ func (cs *State) enterNewRound(height int64, round int32) {
 		logger.Debug("need to set a buffer and log message here for sanity", "start_time", cs.StartTime, "now", now)
 	}
 
-	logger.Debug("entering new round", "current", log.NewLazySprintf("%v/%v/%v", cs.Height, cs.Round, cs.Step))
+	prevHeight, prevRound, prevStep := cs.Height, cs.Round, cs.Step
 
 	// increment validators if necessary
 	validators := cs.Validators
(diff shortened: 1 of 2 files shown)


consensus/state.go
```diff
@@ -1899,7 +1899,7 @@ func (cs *State) defaultSetProposal(proposal *types.Proposal) error {
 		cs.ProposalBlockParts = types.NewPartSetFromHeader(proposal.BlockID.PartSetHeader)
 	}
 
-	cs.Logger.Info("received proposal", "proposal", proposal)
```
sergio-mena, line 1902: @ancazamfir , before merging this, if you have some spare cycles, could you please double-check your concern about the usage of `GetProposer()` here?

Let me explain why I think it is OK. Look at lines 1887-1891 above. We are using `GetProposer()` to obtain the public key that verifies the proposal's signature. So, if that verification succeeds, the validator returned by `GetProposer()` should be the one that proposed `proposal`. Do you agree?

cason (approved):
Legit.

Minor points:
1. For the commit, title of the PR, I would phrase it as "accepting" a proposal.
2. The address of a `Validator` is the address of its public key. An alternative is to re-use the public key already retrieved previously in the method.
3. While logging "proposer" can be practical, for all effects it is the sender of the proposal.

Notice that the proposer address is the the proposer's public key's address.

sergio-mena:
@cason 
1. Changed the PR's title
2. Did a quick test logging `cs.Validators.GetProposer().Address` and `cs.Validators.GetProposer().PubKey.Address()` side by side and they output the same value. So will use your suggestion of reusing the public key retrieved earlier in the method.
3. This PR is adding what we needed in our last troubleshooting session. We can always extend it to log the sender later

consensus/state.go
sergio-mena: That's a good point, @ancazamfir. I have followed your suggestion, but have moved the log to be hit after the proposer is updated (so we have the proposer corresponding to the round we're entering). I have also added an _Info_ log that will log the proposer in non-zero rounds.
Let me know what you think.

lasarojc (approved):

PR #1092, opened 2023-07-05
RFC 104: Internal messaging using the actor model

thanethomson:
:book: [Rendered](https://github.com/cometbft/cometbft/blob/thane/rfc-internal-messaging/docs/rfc/rfc-104-actor-model.md)

I'm submitting this as an RFC as part of a broader effort to consider different options in simplifying Comet's internal concurrency model (which is unnecessarily complex right now). The actor model is only one possible option, and I thought it would be worth capturing some of my past explorations on employing the model somewhere.

Again, this is an _RFC_ and not an ADR because it's merely exploring the solution space as opposed to proposing a concrete course of action. The conclusion of the RFC is that adopting an actor framework is _not_ desirable for a whole host of reasons, and neither is adopting a full-blown actor model. We can, however, extract some principles or practices from the model that could be beneficial.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/docs/rfc/README.md
+++ b/docs/rfc/README.md
@@ -46,3 +46,4 @@ relating to Tendermint Core prior to forking, please see
 - [RFC-101: Banning peers based on ResponseCheckTx](./rfc-101-p2p-bad-peers-checktx.md)
 - [RFC-102: Improve forward compatibility of proto-generated Rust code](./rfc-102-rust-gen-builders.md)
 - [RFC-103: Incoming transactions when node is catching up](./rfc-103-incoming-txs-when-catching-up.md)
+- [RFC-104: Internal messaging using the actor model](./rfc-104-actor-model.md)
--- a/docs/rfc/rfc-104-actor-model.md
+++ b/docs/rfc/rfc-104-actor-model.md
@@ -1,0 +1,193 @@
+# RFC 104: Internal Messaging Using the Actor Model
+
+## Changelog
+
+- 2023-06-29: First draft (@thanethomson)
+
+## Abstract
+
+At present, CometBFT is a collection of components that run concurrently within
+the same process and interact with each other via several different mechanisms:
+
+1. The [`EventBus`][event-bus]
+2. The [`Switch`][switch], where reactors can look up other reactors by name and
+   interact with them directly by calling their methods
+3. Go channels held by reactors
+
+This results in a number of problems:
+
+- Difficulty and complexity in testing individual reactors in isolation, which
+  results in brittle and unnecessarily complex tests.
+- Difficulty reasoning about overall system operation, which slows onboarding of
+  new developers and debugging.
+- Non-deterministic tests which are of dubious value; these also slow down
+  development efforts.
+
+This RFC explores the possibility and potential implications of making use of
+the [Actor] model internally to replace the reactor/service model, all while
+coalescing the three existing concurrent communication approaches into a single
+one.
+
+## Background
+
+The [Actor] model as an idea has been around since the 1970s and is most widely
+known to be employed in [Erlang][erlang]. It imposes an architecture on a system
+whereby all concurrent subprocesses/tasks are modelled as "actors", which are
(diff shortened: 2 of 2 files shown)


mzabaluev:
I'm wary about converting to a fully type-erased actor model. The disadvantages are as you have outlined, and without exhaustive integration test coverage, there will be a possibility programmer errors that are easily caught by strict type checking.

I'd suggest a hybrid approach: make each reactor testable in isolation with dependency services provided by handles, but define typed APIs for actor handles that can be used to send messages to a particular type of actor. We shouldn't have too many of them in CometBFT.

For the sake of programmer sanity, it would be good if the request-response model is be usable with goroutines. As in, to program an asyncronous flow involving a request roundtrip to another actor, one should not have to pass callbacks, manually rig up response matching at the local actor's message processing loop, or anything of the sort. It should be just sequential code. Unfortunately, this is not Rust and we can't have `async`/`await` to make the suspension points explicit.

thanethomson:
@mzabaluev I totally agree. That for me is one of the biggest reasons not to use the actor model (at least the strict version implemented by most actor frameworks).

That's also why I submitted this as an RFC as opposed to an ADR - it's just capturing some of the tradeoffs of the architectural approach. I, like you, believe that there's definitely a better way, but there are some subtle things we could learn from the actor model that would help address some of the tech debt in the system.

adizere (approved):
Thanks Thane for documenting this. The conclusion is sound and the discussion is very useful for future reference.

sergefdrv:
Following up your discussion, I'd like to add my 5 cents. In my opinion, one big disadvantage of the actor model is the lack of built-in flow control (a.k.a. backpressure) between single-threaded actors. What should an actor do if it's not ready to process some incoming message? How to ensure bounded memory usage if individual actors make progress at different pace?

Another concern is fragmentation of the logic in code. Actor's code would need to be implemented as a state machine reacting on incoming events (messages). If certain processes happen to be split across several rounds of communication between actors then the overall logic becomes segregated and harder to follow.

sergefdrv:
@thanethomson 

> Is there some kind of general approach that you'd recommend considering here? Given my understanding of our internal protocols, each one's approach to handling backpressure would possibly be different.

I'd recommend sticking to _pull_-based mechanisms, which I distinguish from _poll_- and _push_-based. In pull-based interaction the consumer connects to the producer and fetches the items provided by the producer if they are ready, or waits until a new item becomes available. As apposed to push-based approaches, the producer cannot force the consumer to have to deal with a new item if the consumer is not ready. Also, in contrast to poll-based approaches, the consumer doesn't have to repetitively check for new items.

I think one can relatively naturally build such mechanism on top of Go channels. A big advantage of using Go channels, comparing e.g. to blocking methods, is the possibility to combine several blocking operations in a single `select` statement.

> ... modelling protocol state machines separately to the code that mutates them, with state machines being thread-safe but single-/un-threaded (i.e. no goroutines or threads ever spawned from _within_ the state machine).

I suppose, by mutation, you mean restructuring the state machine. In any case, if we want to implement a distributed atomic broadcast mechanism (sequence consensus) then we have to merge inherently concurrent paths of execution into a single-threaded sequential execution at some point. I suspect the earlier we try to get rid of the concurrency in code the more we have to fragment otherwise cohesive protocol logic. In particular, I find sigle-threaded state machines are often not really straightforward to follow, especially if hand-coded. Async/await certainly helps a lot, but it's a quite low-level feature, which without additional structure easily results in a great mess. Regarding Go, I just came across [this article](https://research.swtch.com/coro) about coroutines in Go, which you might want to look at.

> Did you have any recommendations in mind though on this front as to good patterns to consider?

I think I'm yet to discover a decent approach to implementing distributed protocols; this is the whole point of the [Replica_IO](https://github.com/replica-io/replica-io) project, which I'm working on. I have some promising stuff to look at more closely on [this list](https://github.com/replica-io/replica-io/issues/7). Luckily, I'm starting from scratch and don't have to bother with any legacy burden at this point.

PR #1109, opened 2023-07-11
Revert "config: add bootstrap peers (#9680)"

cason:
This reverts commit f12588aab1f1b52ffa82ed143676d69e2fab7bf4.

This configuration parameter was first introduced in the `v0.35.x` branch (https://github.com/tendermint/tendermint/pull/6372), then forward-ported to `main` after the retraction of this release (https://github.com/tendermint/tendermint/pull/9680).

More recently, users have requested its inclusion on the `v0.37.x` release branch (https://github.com/cometbft/cometbft/issues/1016). However, when backporting this commit from main (https://github.com/cometbft/cometbft/pull/1019), the reviewers have observed a number of minor issues both in the code and in the documentation of the introduced parameter: 

- The documentation of the introduced configuration parameter is imprecise and it is not following the proper naming adopted for CometBFT's configuration TOML file
- The configured bootstrap peer addresses are added to the address book using a source address that is likely incorrect, as it differs from the source address adopted when manually adding other addresses (e.g. persistent peers) to the address book
- Contrarily to other p2p parameters, there is not method inside the `p2p` package responsible for configuring the provided peer addresses, which in particular renders testing more complex
- No tests were provided to check the operation of the introduced configuration parameter

Given the above summarized issues and provided that this configuration parameter, while useful and good to have, it is not required for a proper operation of a CometBFT node, the team has decided to block the backporting of https://github.com/tendermint/tendermint/pull/9680 to the `v0.37.x` release and to revert it from both `main` and `v0.38.x` branches.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -6,10 +6,6 @@ This guide provides instructions for upgrading to specific versions of CometBFT.
 
 ### Config Changes
 
-* A new config field, `BootstrapPeers` has been introduced as a means of
-  adding a list of addresses to the address book upon initializing a node. This is an
-  alternative to `PersistentPeers`. `PersistentPeers` should be only used for
-  nodes that you want to keep a constant connection with i.e. sentry nodes
 * The field `Version` in the mempool section has been removed. The priority
   mempool (what was called version `v1`) has been removed (see below), thus
   there is only one implementation of the mempool available (what was called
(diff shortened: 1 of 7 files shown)


mzabaluev (approved):
Looks good! Was there a .changelog entry to scrub as well?

cason:
> Looks good! Was there a .changelog entry to scrub as well?

Good point. We started using `unclog` in 2023, while the reverted commit is still from 2022. I've also checked all p2p occurrences under this directory, and none is related to this change.

thanethomson (approved):
Thanks @cason!

PR #1114, opened 2023-07-12
p2p: Remove UPnP functionality

### LINKED ISSUE #1113: p2p: Remove UPnP functionality
The UPnP functionality in Comet seems to have been introduced a long time ago in an effort to facilitate port forwarding, as per https://github.com/cometbft/cometbft/blob/868b71e66addbf5721e12fa741f8c70ed5b35a34/p2p/upnp/upnp.go#L1-L3

It is unclear whether this is really used by anyone any more, and there are other ways to facilitate port forwarding, so we are considering removing it entirely.

thanethomson:
Closes #1113.

I've marked this to be backported to v0.38.x, but should we also be backporting this further back to v0.37.x? It is a breaking change, but if nobody's actively using it then we could easily just make this breaking change in a patch release.

We would just need to verify that people's configuration files are still parsed successfully, ignoring the old UPnP configuration option which is automatically written to `config.toml` when generating it.

---

#### PR checklist

- [ ] Tests written/updated


thanethomson:
> If we can verify this easily, makes sense to port back to 0.37 also! From a ["programmer covenant"](https://emptysqua.re/blog/api-evolution-the-right-way/) perspective I get the impression that this could be perceived as a surprise (even though we checked with users thoroughly).

Agreed. But if it's a dead option that nobody uses, _and_ it's also a security liability to use it, it's probably worth it to break it in this case. It's definitely the exception though as opposed to the norm.

cason:
Shouldn't we deprecate this feature instead? At least on released versions?

thanethomson:
> Shouldn't we deprecate this feature instead? At least on released versions?

Sure, I could do that instead. I really wish our logging framework had a "warning" level - it'd really come in handy for notifying operators that they're using deprecated functionality.

docs/core/configuration.md
```diff
@@ -223,9 +221,6 @@ bootstrap_peers = ""
 # Comma separated list of nodes to keep persistent connections to
 persistent_peers = ""
 
-# UPNP port forwarding
```
cason, line 226: Is it my impression, or this configuration flag is currently useless? What is the effect of setting it to true here?

From the remaining of the review, it appears that no change was made on `p2p` package.

cason (approved):
I think this PR has two purposes.

The first is to remove the unused `upnp` boolean configuration flag. Except if not caught by us, it appears not to be used anywhere in the code base. Removing it from the configuration file and documentation is then legit and harmlesse.

The second is to remove the CLI client `probe-upnp`. This is an offline command, meaning that it does not require or use anything from the node or node configuration, intended, from what I understand, to test whether the UPNP is available on a network. I find a little hard to understand how could it be exploited, to be honest. At the same time, however, no one appears to be using this command and the code base is non-maintained.

docs/core/configuration.md
thanethomson: Actually yes you're right, it appears as though it's dead code. No idea where its usage was removed from the P2P layer. Another good reason to just remove it from v0.37 :slightly_smiling_face:

cason: But also removing the CLI command?

thanethomson: If there's no UPnP functionality in Comet v0.37+ (perhaps even v0.34+, still need to check), what's the point of offering that probe command?

cason: Ok, you have a point. Go ahead! :)

PR #1115, opened 2023-07-12
E2E: Add support for fast prototyping

otrack:
Addresses [this](https://github.com/orgs/cometbft/projects/1/views/4?filterQuery=band&pane=issue&itemId=32190131) issue.

Built atop #1043.

This PR adds a support for fast prototyping to the E2E test suite.
Please read the REAME.md located [here](https://github.com/otrack/cometbft/blob/pierre/mempool-performance/test/e2e/fast-prototyping/README.md)  to start using this code.

---

#### PR checklist

- [ ] Tests written/updated


cason (commented):
This is a huge PR that touches several packages. It is also API breaking, so I think we should consider splitting it into smaller parts easy to review and understand the impacts.

Moreover, it appears to double changes also introduced in another PR, which renders reviewing more complex.

I would suggest not merging this PR into main. Maybe we should consider a feature branch for the mempool work, where we can introduce all these changes with less risk?

otrack:
> This is a huge PR that touches several packages. It is also API breaking, so I think we should consider splitting it into smaller parts easy to review and understand the impacts.
> 
> Moreover, it appears to double changes also introduced in another PR, which renders reviewing more complex.
> 
> I would suggest not merging this PR into main. Maybe we should consider a feature branch for the mempool work, where we can introduce all these changes with less risk?

otrack:
Was replaced with #1139 .

Issue #785, opened 2023-05-03
mempool: Reject incoming txs while catching up


hvanz:
### Summary

Following the decision taken in [RFC 103](https://github.com/cometbft/cometbft/issues/734), a node that is catching up (that is, performing BlockSync or StateSync) should not accept incoming transactions until it switches to consensus mode.

See #734.

### Proposal

While the node is performing state sync or block sync:
- Disable the RPC endpoints `broadcast_txs_*` by returning an error.
- The mempool reactor should reject incoming transactions and should not send any transaction that may have in the mempool.

There was a proposal for implementing [node phases](https://github.com/tendermint/tendermint/issues/4644), but we won't do that here.

cason:
Possibly related, from Tendermint Core repo: [ mempool: tx gossip cause CheckTx error during the statesync stage #8243 ](https://github.com/tendermint/tendermint/issues/8243)

PR #1116, opened 2023-07-12
mempool: Fix mempool capacity when testing

hvanz:
The mempool configuration used for testing has some values that are not realistic: the mempool size is 5000 (the default value) while the cache is set to 1000. 

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -745,6 +745,7 @@ func DefaultMempoolConfig() *MempoolConfig {
 // TestMempoolConfig returns a configuration for testing the CometBFT mempool
 func TestMempoolConfig() *MempoolConfig {
 	cfg := DefaultMempoolConfig()
+	cfg.Size = 1000
 	cfg.CacheSize = 1000
 	return cfg
 }
@@ -773,6 +774,9 @@ func (cfg *MempoolConfig) ValidateBasic() error {
 	}
 	if cfg.MaxTxBytes < 0 {
 		return errors.New("max_tx_bytes can't be negative")
+	}
+	if cfg.Size > cfg.CacheSize {
+		return errors.New("mempool's cache cannot be smaller than mempool's capacity")
 	}
 	return nil
 }
--- a/mempool/clist_mempool_test.go
+++ b/mempool/clist_mempool_test.go
@@ -646,21 +646,28 @@ func TestMempoolNoCacheOverflow(t *testing.T) {
 	mp, cleanup := newMempoolWithAppAndConfig(proxy.NewRemoteClientCreator(sockPath, "socket", true), cfg)
 	defer cleanup()
 
+	// create N + 1 transactions, with N = the capacity of the cache
+	txs := newUniqueTxs(mp.config.CacheSize)
+	tx0 := kvstore.NewTxFromID(mp.config.CacheSize)
+
 	// add tx0
-	var tx0 = kvstore.NewTxFromID(0)
 	err := mp.CheckTx(tx0, nil, TxInfo{})
 	require.NoError(t, err)
 	err = mp.FlushAppConn()
 	require.NoError(t, err)
 
 	// saturate the cache to remove tx0
-	for i := 1; i <= mp.config.CacheSize; i++ {
-		err = mp.CheckTx(kvstore.NewTxFromID(i), nil, TxInfo{})
+	for _, tx := range txs {
+		err = mp.CheckTx(tx, nil, TxInfo{})
 		require.NoError(t, err)
 	}
 	err = mp.FlushAppConn()
 	require.NoError(t, err)
-	assert.False(t, mp.cache.Has(kvstore.NewTxFromID(0)))
+	assert.False(t, mp.cache.Has(tx0))
+
+	// update one transaction to make space for tx0 in the mempool
+	err = mp.Update(1, txs[:1], abciResponses(1, abci.CodeTypeOK), nil, nil)
+	require.NoError(t, err)
 
 	// add again tx0
 	err = mp.CheckTx(tx0, nil, TxInfo{})
@@ -676,6 +683,14 @@ func TestMempoolNoCacheOverflow(t *testing.T) {
 		}
 	}
 	assert.True(t, found == 1)
+}
+
+func newUniqueTxs(n int) types.Txs {
+	txs := make(types.Txs, n)
+	for i := 0; i < n; i++ {
+		txs[i] = kvstore.NewTxFromID(i)
+	}
+	return txs
 }
 
 // This will non-deterministically catch some concurrency failures like
@@ -748,14 +763,18 @@ func doCommit(t require.TestingT, mp Mempool, app abci.Application, txs types.Tx
 	for i, tx := range txs {
 		rfb.Txs[i] = tx
 	}
-	_, e := app.FinalizeBlock(context.Background(), rfb)
-	require.True(t, e == nil)
+	_, err := app.FinalizeBlock(context.Background(), rfb)
+	require.Nil(t, err)
+
 	mp.Lock()
-	e = mp.FlushAppConn()
-	require.True(t, e == nil)
-	_, e = app.Commit(context.Background(), &abci.RequestCommit{})
-	require.True(t, e == nil)
-	e = mp.Update(height, txs, abciResponses(txs.Len(), abci.CodeTypeOK), nil, nil)
-	require.True(t, e == nil)
-	mp.Unlock()
+	defer mp.Unlock()
+
+	err = mp.FlushAppConn()
+	require.Nil(t, err)
+
+	_, err = app.Commit(context.Background(), &abci.RequestCommit{})
+	require.Nil(t, err)
+
+	err = mp.Update(height, txs, abciResponses(txs.Len(), abci.CodeTypeOK), nil, nil)
+	require.Nil(t, err)
 }


mempool/clist_mempool.go
```diff
@@ -39,14 +39,15 @@ type CListMempool struct {
...
-	txs          *clist.CList // concurrent linked-list of good txs
+	txs          *clist.CList // concurrent linked-list of valid txs
 	proxyAppConn proxy.AppConnMempool
 
 	// Track whether we're rechecking txs.
 	// These are not protected by a mutex and are expected to be mutated in
 	// serial (ie. by abci responses which are called in serial).
 	recheckCursor *clist.CElement // next expected response
 	recheckEnd    *clist.CElement // re-checking stops here
+	recheckMtx    cmtsync.Mutex
```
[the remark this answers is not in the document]
hvanz: Yeah, it's not easy to reason about all the mutexes and I was not happy to add a new one. At least we know we plan to eliminate the variables `recheckCursor` and `recheckEnd` in #785, which will also remove the new `recheckMtx` mutex.

In the meantime, below is a list of all the places where these mutexes are locked. I still have to find a good way to visualize their interactions. After a first analysis I can say that `updateMtx` and `recheckMtx` could be both locked simultaneously during a brief time in the `recheckTxs` function, called at the end of `Update`. This can make `globalCb` to wait until `recheckMtx` is released, but this doesn't lead to a deadlock. `CheckTx` is already blocked for being executed during `Update` because `Update` has `updateMtx` taken. And I don't see how `localClient`'s mutex can interfere with the other two.

### `CListMempool.updateMtx`

`updateMtx` is used by the following `CListMempool` methods:
- `CheckTx`
  - `updateMtx` is locked during the whole execution
  -  calls `CheckTxAsync`, which locks `localClient` 's mutex
  - then calls `ReqRes.SetCallback`, which locks it's own mutex
- `Lock` and `Unlock`
    - These methods are only used to lock `updateMtx` during the whole execution of `state.execution.Commit`, which calls, in order:
        - `mempool.FlushAppConn`
        - `mempool.Update` (see below)

`updateMtx` is also locked during the whole execution of:
- `ReapMaxBytesMaxGas`
- `ReapMaxTxs`
- `Flush`

### `CListMempool.recheckMtx`

`recheckMtx` is locked during the whole execution of:
- `initRecheckCursors`, called at the beginning of `recheckTxs`, which is called at the end of `Update`.
- `resCbRecheck`, when processing `CheckTx` responses of type `Recheck`.
- `recheckCursorIsNil`, called:
  - at the beginning of `globalCb`
  - at the beginning of the callback function returned by `reqResCb`, though it's actually not really needed here.

### `localClient`'s `mtx`

There are two methods locking the mutex that are relevant for the mempool:
- `SetResponseCallback`
    - used to set `CListMempool.globalCb` in `CListMempool`'s constructor
- `CheckTxAsync`, which is called by:
        - `mempool.CheckTx`
        - `mempool.recheckTxs` (after calling `initRecheckCursors`)

config/config.go
```diff
@@ -774,6 +775,9 @@ func (cfg *MempoolConfig) ValidateBasic() error {
 	if cfg.MaxTxBytes < 0 {
 		return errors.New("max_tx_bytes can't be negative")
 	}
+	if cfg.CacheSize < cfg.Size {
```
cason, line 778: I am not sure that this must be a requirement. If someone doesn't want any caching?

mempool/clist_mempool.go
```diff
@@ -39,14 +39,15 @@ type CListMempool struct {
 	preCheck  PreCheckFunc
 	postCheck PostCheckFunc
 
-	txs          *clist.CList // concurrent linked-list of good txs
+	txs          *clist.CList // concurrent linked-list of valid txs
```
cason, line 42: I would keep the original version here. It might not be great, but we are not changing in this PR the semantics of the `txs` list.

mempool/clist_mempool.go
```diff
@@ -397,11 +405,11 @@ func (mem *CListMempool) resCbFirstTime(
 			}
 
 			// Check transaction not already in the mempool
-			if e, ok := mem.txsMap.Load(types.Tx(tx).Key()); ok {
-				memTx := e.(*clist.CElement).Value.(*mempoolTx)
+			if elem, ok := mem.txsMap.Load(types.Tx(tx).Key()); ok {
```
cason, line 408: Any real reason for renaming this variable?

mempool/clist_mempool.go
```diff
@@ -418,17 +426,16 @@ func (mem *CListMempool) resCbFirstTime(
 			memTx.addSender(txInfo.SenderID)
 			mem.addTx(memTx)
 			mem.logger.Debug(
-				"added good transaction",
+				"added valid transaction to mempool",
```
cason, line 429: Any real reason for changing this log message?

mempool/clist_mempool.go
```diff
@@ -418,17 +426,16 @@ func (mem *CListMempool) resCbFirstTime(
...
 				"res", r,
 				"height", memTx.height,
 				"total", mem.Size(),
 			)
 			mem.notifyTxsAvailable()
 		} else {
-			// ignore bad transaction
 			mem.logger.Debug(
-				"rejected bad transaction",
+				"rejected invalid transaction",
```
cason, line 438: Again, why to remove the comment and change the log message?

cason (commented):
I don't like using a PR to address two problems.

If there is a concurrency problem, I think a test should be added to catch the bug. Then a solution should be added to solve the problem. This is a PR by itself.

config/config.go
hvanz: I think the mempool is not prepared to not have caching as it doesn't behave as expected when `cache_size` is set to 0. 

I suspect that the implementation of `LRUTxCache` is incorrect. For example, while the capacity of the cache (the `size`) is set to 0 in the constructor and it never changes, the internal data structures `cacheMap` and `list` will grow when we push a tx to the cache.

mempool/clist_mempool.go
hvanz: We can keep the original comment. I thought to change it to keep the wording consistent across the code and the documentation. In fact, I don't think it is anywhere defined what is a "good" or "bad" transaction.

mempool/clist_mempool.go
hvanz: This is also to keep the naming consistent. The mempool holds `CElement`s and in other places we use the name `elem` for functions returning `CElement`s. But we can keep it as it is if you prefer.

config/config.go
cason: In this case, this method should reject size 0. In any case, we should provide to users a better explanation regarding the actual use of the cache.

mempool/clist_mempool.go
otrack: The terms good/bad correspond to valid/invalid. Maybe, having a PR to fix this, as well as others inconsistent comments in the  code, is better.

mempool/clist_mempool.go
hvanz: Same as above. Added "to mempool" to distinguish it from adding to cache.

mempool/clist_mempool.go
hvanz: I believe the comment is not needed here. The log message says exactly what is happening in this part of the code.

mempool/clist_mempool_test.go
```diff
@@ -748,14 +763,18 @@ func doCommit(t require.TestingT, mp Mempool, app abci.Application, txs types.Tx
...
+	defer mp.Unlock()
+
+	err = mp.FlushAppConn()
+	require.Nil(t, err)
+
+	_, err = app.Commit(context.Background(), &abci.RequestCommit{})
+	require.Nil(t, err)
+
+	err = mp.Update(height, txs, abciResponses(txs.Len(), abci.CodeTypeOK), nil, nil)
+	require.Nil(t, err)
```
otrack, line 779: I am not sure to understand the rationale of these changes. The test was written with `mempoolSize > cacheSize` in mind. Now, it works with  `mempoolSize >= cacheSize`, but still fails when`mempoolSize < cacheSize`.

otrack (commented):
I have a couple of concerns regarding these changes. First, as discussed in the team meeting, it would be nice to have some feedback from the community, before we enforce any invariant between `cacheSize` and `mempoolSize`. Second, if we need to handle concurrent re-checks, I suggest to use the mempool mutex instead of adding a new one.

mempool/clist_mempool.go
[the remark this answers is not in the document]
hvanz: I have tried only with `updateMtx` but it doesn't work. Btw, the recheck logic only changes the mempool with `RemoveTxByKey`, which is currently not guarded by any mutex. `RemoveTxByKey` is called by `Update`, which holds `updateMtx`, and by `resCbRecheck` that doesn't lock. We could just use `updateMtx` to guard `RemoveTxByKey` only if it is reentrant, as you said.

I added a new test `TestMempoolRecheckRace` that shows more clearly the problem. The test is: 
1. add a bunch of transactions to the mempool
2. update one transaction to force rechecking the rest
3. add again one transaction → this will fail with a data race for the variable `recheckCursor`, which is accessed from the functions `reqResCb` and `recheckTxs`.

(This new test actually a simplified version of `TestMempoolNoCacheOverflow`, so it's a bit redundant and we can merge them.)

mempool/clist_mempool.go
hvanz: Ok, I reverted all these small changes to the comments and logs.

mempool/clist_mempool.go
hvanz: Reverted.

mempool/clist_mempool.go
hvanz: Reverted.

mempool/clist_mempool_test.go
hvanz: These changes in `doCommit` are just a small refactoring to make the code consistent with the rest of the testing code. They shouldn't modify the result of the tests.

mempool/clist_mempool.go
hvanz: Reverted.

hvanz:
I reverted all the changes related to testing to focus the discussion on the concurrency problem on the recheck variables. So fixing the mempool capacity for testing is left for a separate issue #1144.

hvanz:
Closing as there is no agreement on the proposed solution. The bug is now reported in #1827.

PR #1119, opened 2023-07-13
Reject incoming txs while node is catching up

### LINKED ISSUE #785: mempool: Reject incoming txs while catching up

hvanz:
Closes #785 

I also change the type of `consensus.reactor.waitSync` from `bool` to `atomic.Bool`. Now the mutex `mtx` is only required by the round state `rs`, so I renamed `mtx` to `rsMtx`.

---

#### PR checklist

- [ ] Tests written/updated


rpc/core/mempool.go
```diff
@@ -20,6 +20,9 @@ import (
 // CheckTx nor transaction results.
 // More: https://docs.cometbft.com/main/rpc/#/Tx/broadcast_tx_async
 func (env *Environment) BroadcastTxAsync(_ *rpctypes.Context, tx types.Tx) (*ctypes.ResultBroadcastTx, error) {
+	if env.ConsensusReactor.WaitSync() {
```
cason, line 23: I think the calls should check against the mempool whether it is accepting transactions.

mempool/reactor.go
```diff
@@ -136,7 +155,7 @@ func (memR *Reactor) broadcastTxRoutine(peer p2p.Peer) {
 
 	for {
 		// In case of both next.NextWaitChan() and peer.Quit() are variable at the same time
-		if !memR.IsRunning() || !peer.IsRunning() {
+		if !memR.IsRunning() || !peer.IsRunning() || memR.WaitSync() {
```
cason, line 158: This addition causes this routine to be interrupted (and never restarted) if a peer is added while the node is still syncing. This will probably break the protocol.

mempool/reactor.go
```diff
@@ -19,17 +20,21 @@ import (
...
-	ids     *mempoolIDs
+	config   *cfg.MempoolConfig
+	mempool  *CListMempool
+	ids      *mempoolIDs
+	waitSync *atomic.Bool
 }
 
 // NewReactor returns a new Reactor with the given config and mempool.
-func NewReactor(config *cfg.MempoolConfig, mempool *CListMempool) *Reactor {
+func NewReactor(config *cfg.MempoolConfig, mempool *CListMempool, waitSync bool) *Reactor {
```
cason, line 30: Is `waitSync` ever initialized to true? I mean, does its initial value varies depending on the scenario?

consensus/reactor.go
cason, line 1: I wander whether this check should not be performed inside `CheckTx`. In this way, it would apply to any way transactions are added to the mempool.

mempool/reactor.go
```diff
@@ -95,6 +100,11 @@ func (memR *Reactor) Receive(e p2p.Envelope) {
 	memR.Logger.Debug("Receive", "src", e.Src, "chId", e.ChannelID, "msg", e.Message)
 	switch msg := e.Message.(type) {
 	case *protomem.Txs:
+		if memR.WaitSync() {
+			memR.Logger.Info("ignore message received during syncing", "msg", msg)
```
cason, line 104: If syncing takes a while, we may have a large number of such messages in the log, in INFO level.

rpc/core/mempool.go
hvanz: I think that would be better, but then we would need to add a `MempoolReactor` to the environment. Are we ok with that?

consensus/reactor.go
hvanz: `CheckTx` is not part of the reactor. If we put the check inside `CheckTx`, we would need to add it to all implementations of `Mempool`.

mempool/reactor.go
hvanz: Right. I will change it to debug level. The number of log messages will still be high, but I think they will be useful for debugging or testing.

mempool/reactor.go
hvanz: I'm not sure I understand the question. `waitSync` is initialized with the same value as the `waitSync` variable in the consensus reactor. So it's initially true when `stateSync || blockSync` is true, as defined in the node. In both reactors, if `waitSync` is initially true, it will be set to false by the methods `EnableIncomingTxs` or `SwitchToConsensus`.

rpc/core/mempool.go
hvanz: There are two options here. One is what it's currently implemented: we keep the information that the node is catching up (WaitSync) only in the consensus reactor, and the mempool RPC endpoints query the consensus reactor via the RPC environment. This creates a slight dependency in the mempool endpoints to the consensus reactor. 

The other option is to duplicate the WaitSync information in the mempool reactor and add to the RPC environment a pointer to the mempool reactor. The method SwitchToConsensus in consensus will have to update the waitSync info in consensus and mempool, creating a new dependency between the two reactors. I think we don't want to do this.

Ideally we should have a separate module that controls the different phases of the node, but that's for another discussion.

hvanz: In the end I changed ConsensusReactor by MempoolReactor. The waitSync information is already in the mempool reactor so I just needed to add MempoolReactor to the environment.

cason (commented):
My main comment regards the changes made in the mempool reactor to prevent the sending of transactions to peers during sync mode.

The adopted approach consists on not creating the per-peer send transaction routines while in sync mode, to then, when sync mode is concluded, start all of them together.

The limitation of this approach is the assumption that the transition between sync mode and no sync mode happens only once, and provided that we were in sync mode. While this is a reasonable assumption, the code behavior becomes unpredictable if, for any reason, this assumption is violated. Namely, if  `EnableInOutTxs` method happens to be called more than once, or when it is called while `waitSync` is false. In this case, multiple send routines will be started for the same peer, with unclear consequences.

mempool/reactor.go
```diff
@@ -21,8 +21,9 @@ import (
 // peers you received it from.
 type Reactor struct {
 	p2p.BaseReactor
-	config  *cfg.MempoolConfig
-	mempool *CListMempool
+	config   *cfg.MempoolConfig
+	mempool  *CListMempool
+	waitSync *atomic.Bool
```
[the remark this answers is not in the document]
hvanz: I agree. I changed the code.

mempool/reactor.go
```diff
@@ -52,10 +56,10 @@ func (memR *Reactor) SetLogger(l log.Logger) {
 
 // OnStart implements p2p.BaseReactor.
 func (memR *Reactor) OnStart() error {
+	memR.Logger.Info("Reactor starts", "waitSync", memR.WaitSync())
```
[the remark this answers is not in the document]
hvanz: Ok, I changed it to only log when waitSync is true, with a better message for node operators. Note that I copied this line from the consensus reactor, so I updated the line in the consensus reactor as well.

consensus/reactor.go
```diff
@@ -43,22 +44,24 @@ type Reactor struct {
 
 	conS *State
 
-	mtx      cmtsync.RWMutex
-	waitSync bool
+	waitSync *atomic.Bool
```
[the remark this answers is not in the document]
hvanz: Same as above.

node/node.go
```diff
@@ -64,7 +64,7 @@ type Node struct {
 	stateStore        sm.Store
 	blockStore        *store.BlockStore // store the blockchain to disk
 	bcReactor         p2p.Reactor       // for block-syncing
-	mempoolReactor    p2p.Reactor       // for gossipping transactions
+	mempoolReactor    *mempl.Reactor    // for gossipping transactions
```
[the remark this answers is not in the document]
hvanz: When we create the RPC environment ([here](https://github.com/cometbft/cometbft/blob/1b076af89d362d3535d854b7268e89c13f1335e8/node/node.go#L667)) we need to pass a struct with the method `WaitSync`. `p2p.Reactor` does not have this method.

consensus/reactor.go
```diff
@@ -43,22 +44,24 @@ type Reactor struct {
...
 	eventBus *types.EventBus
-	rs       *cstypes.RoundState
+
+	rs    *cstypes.RoundState
+	rsMtx cmtsync.Mutex
```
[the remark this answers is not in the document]
hvanz: This mutex was used by `waitSync` and `rs`. Now it's only used by `rs`. I think it is more clear to name it `rsMtx` because it's the companion mutex to `rs` only, and to not confuse it with the other variables in the Reactor. This practice is applied in other parts of the code. But we can leave it unchanged if you prefer to.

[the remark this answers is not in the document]
cason: it is up to you, as this has no practical effects. My worries are also about restricting source code diffs to when there is an actual change in behavior. Renaming is the type of source change change that has no effect at all in behavior.

consensus/reactor.go
```diff
@@ -43,26 +44,29 @@ type Reactor struct {
...
 	conS *State
 
-	mtx      cmtsync.RWMutex
-	waitSync bool
+	waitSync atomic.Bool
 	eventBus *types.EventBus
-	rs       *cstypes.RoundState
+
+	rs    *cstypes.RoundState
+	rsMtx cmtsync.Mutex
```
thanethomson, line 51:
```suggestion
	rsMtx cmtsync.Mutex
	rs    *cstypes.RoundState
```

Our convention is usually to list the mutex variable just above all the other variables it guards. I don't see any issues with renaming it.

mempool/reactor.go
```diff
@@ -129,6 +142,21 @@ func (memR *Reactor) Receive(e p2p.Envelope) {
 	// broadcasting happens from go routines per peer
 }
 
+func (memR *Reactor) EnableInOutTxs() {
```
thanethomson, line 145: Easier still, why not just:

```suggestion
func (memR *Reactor) EnableInOutTxs() {
	if !memR.WaitSync() {
		return
	}
```

cason: This works if we assume that we never transition back, right? I mean, wait sync is true to then become false forever.

Even in this case, what exactly happens if `AddPeer` and this method are called at the same time?

thanethomson (commented):
Two things:
1. This is a great example of where adopting a [typestate](http://www.cs.cmu.edu/~aldrich/papers/classic/tse12-typestate.pdf)-based approach to modelling the mempool reactor would be better imho. In other words, instead of maintaining different "modes" of operation of a single reactor (an approach which is prone to greater complexity and is more difficult to reason about), rather split each mode into its own type with functions that facilitate the destruction of the existing instance and simultaneous creation of the new instance (of the appropriate type).
2. Why use an atomic bool over the existing mutex-based approach?

mempool/reactor.go
hvanz: > This works if we assume that we never transition back, right?

That's the current assumption, isn't it? `SwitchToConsensus` in the consensus reactor works like this.

> what exactly happens if AddPeer and this method are called at the same time?

What about the following code?
```suggestion
func (memR *Reactor) EnableInOutTxs() {
	// If waitSync is not true, it means that txs are already enabled, then
	// return immediately; otherwise, set it to false and enable txs.
	if !memR.waitSync.CompareAndSwap(true, false) {
		return
	}
```

consensus/reactor.go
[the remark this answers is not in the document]
hvanz: So if this about code readability vs easiness of git archeology, I prefer the first one, as I believe having precise names is very important. But I may be missing something. For the latter there are tools and plugins.

hvanz:
> 2. Why use an atomic bool over the existing mutex-based approach?

Afaiu, mutexes are slower than atomic variables. Also, this mutex in particular was used for two different variables.

In some parts of the code, when we need an atomic bool we use a `uint32` and treat it as atomic (for [example](https://github.com/cometbft/cometbft/blob/main/libs/service/service.go#L100-L101)). When this code was written, Go standard libraries didn't have atomic booleans. `atomic.Bool` was introduced with Go 1.8 and the [underlying implementation](https://github.com/golang/go/blob/master/src/sync/atomic/type.go#L9-L36) uses a `uint32`, so I think they are equivalent.

cason:
I would add a more general question here regarding how we are trying to solve this problem.
 
The problem, if I understood correctly, is that the mempool should not accept transactions while the node is "sync mode". This is a problem because to add a transaction to the mempool, we need first to call the ABCI `CheckTx` method so that the application validates the transaction. While a node is in "sync mode", the state of the application is outdated, rendering the validation of transactions potentially invalid (based on an outdated state).

To solve this problem, I would make the mempool to reject transactions while the node is in "sync mode". The solution, however, does the following:
- Incoming transactions from all three possible endpoints (3 RPC + network) are rejected while the node is in "sync mode". So, we don't call `mempool.CheckTx` to add those transactions to the mempool. 
   - I wonder why is not this method that rejects transactions when the node is in "sync mode" and the mempool therefore is "locked".
- Outgoing transactions: the mempool reactor consumes the content of the mempool to send transactions to peers. This is performed by broadcast routines, one per peer. The proposed solution prevents starting those routines while in "sync mode".
    -  If the mempool is locked, not receiving transactions, why should it return transactions to the broadcast routines? 
    -  The broadcast routines wait for new transactions to be returned by the mempool. The broadcast routines will not do anything if the mempool does not provide them transactions to send. It seems thus simpler to prevent the mempool from providing transactions to the broadcast routines while the node is "sync mode" than blocking (actually, not starting) the broadcast routines on the mempool reactor.

mempool/reactor.go
```diff
@@ -93,6 +98,11 @@ func (memR *Reactor) Receive(e p2p.Envelope) {
 	memR.Logger.Debug("Receive", "src", e.Src, "chId", e.ChannelID, "msg", e.Message)
 	switch msg := e.Message.(type) {
 	case *protomem.Txs:
+		if memR.WaitSync() {
+			memR.Logger.Debug("ignore message received during syncing", "msg", msg)
```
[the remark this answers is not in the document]
sergio-mena: Log messages should start with lower case.
We should make an effort at not duplicating them. If a user comes with a problem and a log line, we should be able to know for sure which line produced the log.

hvanz:
Thanks a lot for the feedback @cason! I now see that there are many concurrency issues with the current solution. The main problem seems to be that `broadcastTxRoutine` could be spawned multiple times for the same peer, because it can be triggered from `AddPeer` and from `EnableInOutTxs` if bad interleaving happens. 

I propose another solution. We append each new peer to a queue (a channel) and we only start processing this queue in a separate goroutine only when `waitSync` is false, which can happen either in `OnStart` or in `EnableInOutTxs`. These two methods can't run concurrently, in contrast to `AddPeer` and `EnableInOutTxs`. Check out [this branch](https://github.com/cometbft/cometbft/pull/1381/files?diff=split&w=0) for an implementation of this solution. Also, there is no need to loop over `memR.Switch.Peers().List()`.

hvanz:
> I wonder why is not the mempool itself, through its CheckTx method, that rejects transactions when the node is in "sync mode" (the mempool in this case is kind of "locked").

iiuc, it's because the reactor can do more things than just calling CheckTx when a new tx is received from a peer. The current implementation actually only calls CheckTx and then registers the sender if the tx is valid. But other protocols may work differently or receive other types of messages.

hvanz:
> it is not up to the reactor implementation to decide whether a transaction should be propagated or not.  It is up to this implementation to decide how to propagate transactions that are in the mempool.

It is precisely the reactor (`mempool/reactor.go` currently) that takes care of tx propagation. The concern of the mempool implementation (`mempool/clist_mempool.go`) is how to validate and store txs.

> The actions taken by the reactor upon receiving a transaction are irrelevant when the mempool just reject the transaction, right?

The description of the `BaseReactor` interface starts like this: "Reactor is responsible for handling incoming messages on one or more Channel". So what I think is that, before the mempool implementation validates the transaction, the mempool reactor should decide whether to receive and process the transaction or not. And in the case of transactions received via RPC endpoints, we leave it to each RPC endpoint to reject transactions when in sync mode.

cason:
A proposal for solving the reported synchronization issues: #1386

mempool/reactor.go
```diff
@@ -129,6 +139,21 @@ func (memR *Reactor) Receive(e p2p.Envelope) {
 	// broadcasting happens from go routines per peer
 }
 
+func (memR *Reactor) EnableInOutTxs() {
```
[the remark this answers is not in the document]
hvanz: Comment already addressed with the line `if !memR.waitSync.CompareAndSwap(true, false) { ...`

mempool/reactor.go
hvanz: All these comments were addressed with the introduction of `waitSyncCh`.

lasarojc (approved):

cason (approved):
While we don't have a better solution to coordinate the startup of reactors, the proposed changes should prevent the mempool from processing transactions while the node is catching up. I still believe that mempool itself, not the mempool protocol or the RPC endpoints, should be responsible for this accepting or not transactions, but this was already discussed as part of this PR.

PR #1123, opened 2023-07-14
ci: Disable CodeQL check in merge queues

thanethomson:
Seems as though the CodeQL checks are failing on `merge_group` for some reason: https://github.com/cometbft/cometbft/actions/runs/5555159575/jobs/10145922428 - seems like it can't find the code it's supposed to check.

I'm not sure if that check's compatible with merge queues, so I'll disable this check for now. It still gets run on pull requests and pushes to `main` as it always was.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -13,7 +13,6 @@ name: "CodeQL"
 
 on:
   workflow_dispatch:
-  merge_group:
   push:
     branches: ["main"]
   pull_request:


cason (approved):
I had the same problem with a PR that I've added to the merge queue.

I didn't understand the reason of the failure, to be honest.

thanethomson:
> I didn't understand the reason of the failure, to be honest.

The check just couldn't find the specific commit it was supposed to check. I suspect it may be a bug in the action itself.

Issue #88, opened 2023-01-11
Introduce relaxed ABCI concurrency models


lasarojc:
With regards to concurrent calls to ABCI methods, the current specification states that
> In principle, each of the four ABCI++ connections operates concurrently with one another. This means applications need to ensure access to state is thread safe. Both the [default in-process ABCI client](https://github.com/tendermint/tendermint/blob/v0.34.x/abci/client/local_client.go#L18) and the [default Go ABCI server](https://github.com/tendermint/tendermint/blob/v0.34.x/abci/server/socket_server.go#L32) use a global lock to guard the handling of events across all connections, so they are not concurrent at all. This means whether your app is compiled in-process with Tendermint using the NewLocalClient, or run out-of-process using the SocketServer, ABCI messages from all connections are received in sequence, one at a time.
> 
> The existence of this global mutex means Go application developers can get thread safety for application state by routing all reads and writes through the ABCI system. Thus it may be unsafe to expose application state directly to an RPC interface, and unless explicit measures are taken, all queries should be routed through the ABCI Query method.

In the recalled [0.36 branch](https://github.com/tendermint/tendermint/blob/09ee6b1453a1d4a5019b399a1bda9437083625dc/spec/abci%2B%2B/abci%2B%2B_app_requirements.md), changes had been made to allow for more concurrency

> In principle, each of the four ABCI++ connections operates concurrently with one another. This means applications need to ensure access to state is thread safe. Up to v0.35.x, both the [default in-process ABCI client](https://github.com/tendermint/tendermint/blob/v0.35.x/abci/client/local_client.go#L18) and the [default Go ABCI server](https://github.com/tendermint/tendermint/blob/v0.35.x/abci/server/socket_server.go#L32) used a global lock to guard the handling of events across all connections, so they were not concurrent at all. This meant whether your app was compiled in-process with Tendermint using the NewLocalClient, or run out-of-process using the SocketServer, ABCI messages from all connections were received in sequence, one at a time. This is no longer the case starting from v0.36.0: the global locks have been removed and it is up to the Application to synchronize access to its state when handling ABCI++ methods on all connections. Nevertheless, as all ABCI calls are now synchronous, ABCI messages using the same connection are still received in sequence.

Also

> Before invoking Commit, Tendermint locks the mempool and flushes the mempool connection. This ensures that no new messages will be received on the mempool connection during this processing step, providing an opportunity to safely update all four connection states to the latest committed state at the same time.

and 

[...]

PR #1141, opened 2023-07-18
abci: Add relaxed local client synchronization models

thanethomson:
Relates to #88.

In this PR, the `proxy.ClientCreator` interface is expanded such that the client creator has more control over the concurrency model imposed from Comet's perspective for each "connection" (i.e. group of ABCI methods).

This PR also effectively renames the "unsynchronized" local client creator to the "connection-synchronized" local client creator to be more clear on its purpose.

As an example of a more advanced local client concurrency model, it introduces a "consensus-synchronized" local client creator. This only imposes synchronization (i.e. a mutex) on the consensus "connection" to ensure that Comet serializes all of its consensus-related ABCI calls, while all other connections' concurrency models are left up to the application to decide.

It also introduces a fully "unsynchronized" local client creator, which imposes no restrictions on _any_ of the ABCI calls. This is the most advanced kind of local ABCI client creator, which requires greater care on the application side when using it.

Should be able to be reviewed commit by commit.

---

#### PR checklist

- [ ] Tests written/updated


adizere (commented):
From what I understand, the present PR has 3 local client creators:
1. `localClientCreator`, which has a global mutex shared by all 4 connections  (consensus, mempool, query, snapshot), meaning full serialization across all 4 connection is enforced
2. `connSyncLocalClientCreator`, which has one mutex per each of the 4 connections, meaning that communication is serialized per-connection; across separate connections, communication can proceed in parallel
3. `consensusSyncLocalClientCreator` which has a mutex guarding the consensus connection, and no serialization mechanism guarding the other 3 connections

Should we also provide a fully unsynchronized local client creator, i.e., the `unsyncLocalClientCreator` from [old PR #9660](https://github.com/tendermint/tendermint/pull/9660/files#diff-ac53245fa7c37f12472538afe3b39784ae317e77df4e9a7624471b6e6bb8ca33R45-R61)?

From what I understood with discussions with Marko, Dev, and Luke, `unsyncLocalClientCreator` is likely what they'd want to use in the future.

thanethomson:
> Should we also provide a fully unsynchronized local client creator, i.e., the unsyncLocalClientCreator from [old PR #9660](https://github.com/tendermint/tendermint/pull/9660/files#diff-ac53245fa7c37f12472538afe3b39784ae317e77df4e9a7624471b6e6bb8ca33R45-R61)?

The question there is, do we want to leave the concurrency model up to the application when it comes to the consensus interaction? Or do we want to ensure serialization of all consensus-related requests to the application? My sense is that we'd want to actually enforce that, but just for consensus.

I can see how parallelizing vote extension verification, for instance, could be useful. But would we want to be able to parallelize Prepare/ProcessProposal and FinalizeBlock/Commit?

alexanderbez:
> I can see how parallelizing vote extension verification, for instance, could be useful. But would we want to be able to parallelize Prepare/ProcessProposal and FinalizeBlock/Commit?

This is going to tie into optimistic execution in the SDK I believe. It's hard to imagine what this might look like if CometBFT gives us complete control, but having full concurrent calls to ABCI consensus methods is a game changer for us IMO and something we should have support for.

thanethomson:
> It's hard to imagine what this might look like if CometBFT gives us complete control, but having full concurrent calls to ABCI consensus methods is a game changer for us IMO and something we should have support for.

Since it's Comet who ultimately calls the application for consensus-related requests, it's up to us to make sure that it's safe and correct to make those calls at the right time(s), in the right sequence, and that requests don't conflict with each other, which is why it seems to make sense to me for us to control the concurrency model just for the consensus-related requests. I have no concerns around offering fully concurrent calling for mempool, query and state sync-related requests (or at least allowing the application control over the concurrency there).

@alexanderbez do you have specific ideas in mind as to how you'd like/need the consensus-related calls to be parallelized?

adizere:
I think we should add the `unsyncLocalClient` and then get this merged.

The motivation for adding the fully unsynchronized local client is that this is being used in production ([ref1](https://github.com/cometbft/cometbft/issues/88#issuecomment-1557515465), [ref2](https://github.com/dydxprotocol/cometbft/pull/3)) already. The main disadvantages: (a) more code for Comet team to maintain, (b) more risk for application builders to misuse the unsynchronized API. We can mitigate risk (b) through documentation and code comments cc @andynog. We cannot mitigate risk (a), but arguably the fact that there exist forks with this custom modification is a bigger risk for the ecosystem.

thanethomson:
Sure thing @adizere, I've now added the unsynchronized local client creator :+1:

.changelog/unreleased/improvements/1141-abci-unsync-proxy.md
```diff
@@ -0,0 +1,3 @@
+- `[abci/client]` Add fully unsynchronized local client creator, which only
```
adizere, line 1:
```suggestion
- `[abci/client]` Add fully unsynchronized local client creator, which
```

adizere (approved):

ValarDragon:
Yay, glad this got in!

adizere:
SDK just mentioned they plan to have this tested before end of the year, and the change can be rolled out before the next release of SDK (because of separately versioned modules).

thanethomson:
Should we backport this to `v0.38.x`?

adizere:
Not sure. @tac0turtle do you see the value in backporting this to 0.38?

tac0turtle:
depends on when youre next release would be? if its before eoy or even january then it should be fine for us to wait as we will try to make serverv2 compatible with exisitng chains allowing chains to migrate to that sooner then an official release

adizere:
> depends on when youre next release would be? if its before eoy or even january then it should be fine for us to wait as we will try to make serverv2 compatible with exisitng chains allowing chains to migrate to that sooner then an official release

Great! Next release will be before EOY so then no need to backport. I might ask a couple of teams additionally if they need this backported. But by default we won't do it unless there's explicit need for backporting. Thanks!

Olshansk:
cc @red-0ne

PR #1142, opened 2023-07-18
Add gRPC block service

### LINKED ISSUE #1094: ADR-101: Implement gRPC `BlockService`
We need an implementation of the `BlockService` from ADR-101 on the `feature/adr101-pull-companion` branch.

andynog:
Closes #1094 

Implemented the gRPC [Block Service](https://github.com/cometbft/cometbft/blob/thane/adr-084-data-companion-pull-api/docs/architecture/adr-101-data-companion-pull-api.md#block-service) as defined in the gRPC API requirements in the ADR-101 Data Companion Pull API.

- Implemented both `GetByHeight` and `GetLatestHeight` (streaming) rpc endpoints
- Added protobuf files to support Block Service endpoint
- Implemented service server code for both endpoints
- Implemented client code for both endpoints
- Added e2e tests
- Add logic to enable / disable service in the config

---

#### PR checklist

- [X] Tests written/updated


rpc/grpc/server/services/blockservice/service.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	for {
+		select {
+		case msg := <-sub.Out():
+			switch eventType := msg.Data().(type) {
+			case types.EventDataNewBlock:
+				if err := stream.Send(&blocksvc.GetLatestHeightResponse{Height: eventType.Block.Height}); err != nil {
+					err := status.Error(codes.Unimplemented, "cannot send stream response")
+					return err
```
thanethomson, line 122:
```suggestion
					return status.Error(codes.Unavailable, "cannot send stream response")
```

The "unavailable" code seems like it makes more sense for this kind of problem: https://pkg.go.dev/google.golang.org/grpc@v1.56.2/codes#Code

I'd highly recommend logging internal errors encountered within this function at error level, along with the UUID, so that the errors aren't swallowed up and lost.

rpc/grpc/server/services/blockservice/service.go
```diff
@@ -0,0 +1,139 @@
...
+				}
+			}
+		case <-sub.Canceled():
+			if sub.Err() == cmtpubsub.ErrUnsubscribed {
+				err := status.Error(codes.Canceled, "client unsubscribed")
+				return err
```
thanethomson, line 128:
```suggestion
				return status.Error(codes.Canceled, "client unsubscribed")
```

rpc/grpc/server/services/blockservice/service.go
```diff
@@ -0,0 +1,139 @@
...
+		case msg := <-sub.Out():
+			switch eventType := msg.Data().(type) {
+			case types.EventDataNewBlock:
+				if err := stream.Send(&blocksvc.GetLatestHeightResponse{Height: eventType.Block.Height}); err != nil {
+					err := status.Error(codes.Unimplemented, "cannot send stream response")
+					return err
+				}
+			}
+		case <-sub.Canceled():
+			if sub.Err() == cmtpubsub.ErrUnsubscribed {
```
thanethomson, line 126: What about the case where the subscription's been cancelled, but the error isn't `cmtpubsub.ErrUnsubscribed` (and perhaps `sub.Err() == nil`)?

rpc/grpc/server/services/blockservice/service.go
```diff
@@ -0,0 +1,139 @@
...
+				return err
+			}
+		default:
+			continue
+		}
+		err := sub.Err()
+		if err != nil {
+			err := status.Error(codes.Internal, "error in new block subscription")
+			return err
+		}
```
thanethomson, line 137:
```suggestion
		if sub.Err() != nil {
			return status.Error(codes.Internal, "error in new block subscription")
		}
```

Simplify

test/e2e/tests/grpc_test.go
```diff
@@ -30,3 +30,94 @@ func TestGRPC_Version(t *testing.T) {
...
+			require.NotNil(t, lastBlock.BlockID)
+			require.Equal(t, lastBlock.Block.Height, last)
+		}
+	})
+}
+
+func TestGRPC_Block_GetLatestHeight(t *testing.T) {
+
+	testnet := loadTestnet(t)
+	node := testnet.RandomNode()
```
thanethomson, line 88: What if this returns a light client?

test/e2e/tests/grpc_test.go
```diff
@@ -30,3 +30,94 @@ func TestGRPC_Version(t *testing.T) {
...
+
+	client, err := node.Client()
+	require.NoError(t, err)
+	status, err := client.Status(ctx)
+	require.NoError(t, err)
+
+	ch := make(chan int64)
+	//gCtx, cancel := context.WithCancel(context.Background())
+	//defer cancel()
```
thanethomson, line 97: Still necessary?

test/e2e/tests/grpc_test.go
```diff
@@ -30,3 +30,94 @@ func TestGRPC_Version(t *testing.T) {
...
+	gCtx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+	//gCtx := context.Background()
```
thanethomson, line 100: Still necessary?

rpc/grpc/client/block_service.go
```diff
@@ -0,0 +1,96 @@
...
+	req := blocksvc.GetLatestHeightRequest{}
+
+	latestHeight, err := c.client.GetLatestHeight(ctx, &req)
+	if err != nil {
+		return err
+	}
+
+	go func() {
+		for {
+			response, _ := latestHeight.Recv()
```
thanethomson, line 73: Error handling?

rpc/grpc/client/block_service.go
```diff
@@ -0,0 +1,96 @@
...
+			ch <- response.Height
```
thanethomson, line 74: What if the channel's closed?

rpc/grpc/client/block_service.go
```diff
@@ -0,0 +1,96 @@
...
+func (c *blockServiceClient) GetBlockLatestHeight(ctx context.Context, ch chan<- int64) error {
+	req := blocksvc.GetLatestHeightRequest{}
+
+	latestHeight, err := c.client.GetLatestHeight(ctx, &req)
+	if err != nil {
+		return err
+	}
+
+	go func() {
+		for {
```
thanethomson, line 72: The supplied context (`ctx`) isn't respected in this goroutine.

test/e2e/tests/grpc_test.go
```diff
@@ -30,3 +30,94 @@ func TestGRPC_Version(t *testing.T) {
...
+	count := 0
+	for {
+		select {
+		case <-gCtx.Done():
+			return
+		default:
+			// received a new block height and test if it is greater than earliest block
+			h := <-ch
+			if count == 10 {
+				<-gCtx.Done()
```
thanethomson, line 116: Why not just break out of the `for` loop?

thanethomson (changes requested):
Looks like it's starting to take shape, thanks @andynog! Still some work to be done though.

rpc/grpc/server/services/blockservice/service.go
```diff
@@ -0,0 +1,129 @@
+package blockservice
```
mzabaluev, line 1: Perhaps this could be renamed to `block`?
I'm not a fan of repetitive naming conventions, and I'd like the package for the
pruning service be named simply `services/pruning`.

mzabaluev: Then again, there's already `versionservice`.

rpc/grpc/client/block_service.go
```diff
@@ -0,0 +1,100 @@
...
+			response, err := latestHeight.Recv()
+			if err != nil {
+				break
+			}
+			ch <- response.Height
```
thanethomson, line 79:
```suggestion
			select {
			case ch <- response.Height:
			default:
			}
```

As per my previous question, what's a good behaviour to implement if the `out` channel's full? We should check for this and probably just drop the publishing of the current height so as to not block this goroutine. They'll get the next height anyways once the channel has space again.

If we go this route we should probably document the behaviour in the docstring so callers are aware that they may get non-contiguous heights.

rpc/grpc/server/services/blockservice/service.go
thanethomson: The reason for the naming convention here is to facilitate human-readable construction like:

```golang
svr := blockservice.New(store, eventBus, logger)
```

If we rename the package to `block` that would become:

```golang
svr := block.New(store, eventBus, logger)
```

which doesn't make sense. I can't think of a constructor name off-hand right now that would make sense if we named the package `block`.

rpc/grpc/server/services/blockservice/service.go
```diff
@@ -0,0 +1,129 @@
...
+	latestHeight := s.store.Height()
+
+	// validate height parameter, if height is 0 or
+	// the request is nil, then use the latest height
+	if req.Height == 0 {
+		height = latestHeight
+	} else if req.Height < 0 {
+		errMsg := fmt.Sprintf("got negative height (%d), please specify a height >= 0", req.Height)
+		s.logger.Error("GetByHeight", "err", errMsg)
+		return nil, status.Error(codes.InvalidArgument, errMsg)
```
thanethomson, line 47: Please simplify all these sorts of error returns, as per my previous round of feedback.

rpc/grpc/server/services/blockservice/service.go
```diff
@@ -0,0 +1,129 @@
...
+	} else {
+		height = req.Height
+	}
+
+	// check if the height requested is not higher
+	// than the latest height in the store
+	if height > latestHeight {
+		errMsg := fmt.Sprintf("height requested (%d) is higher than the latest available height (%d)", height, latestHeight)
+		s.logger.Error("GetByHeight", "err", errMsg)
+		return nil, status.Errorf(codes.InvalidArgument, errMsg)
```
thanethomson, line 57: Simplify

rpc/grpc/server/services/blockservice/service.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	block := s.store.LoadBlock(height)
+	blockProto, err := block.ToProto()
+	if err != nil {
+		errMsg := fmt.Sprintf("block at height %d not found", height)
+		s.logger.Error("GetByHeight", "err", errMsg)
+		return nil, status.Errorf(codes.NotFound, errMsg)
```
thanethomson, line 65: Simplify

rpc/grpc/server/services/blockservice/service.go
```diff
@@ -0,0 +1,129 @@
...
+// GetLatestHeight implements v1.BlockServiceServer GetLatestHeight method
+func (s *blockServiceServer) GetLatestHeight(_ *blocksvc.GetLatestHeightRequest, stream blocksvc.BlockService_GetLatestHeightServer) error {
+	// Generate a unique subscriber ID using a UUID
+	// The subscriber needs to be unique across all clients
+	id, err := uuid.NewUUID()
+	if err != nil {
+		// cannot generate unique id
+		errMsg := "error generating a subscriber id, cannot subscribe to new block events"
+		s.logger.Error("GetLatestHeight", "err", errMsg)
+		return status.Error(codes.Internal, errMsg)
```
thanethomson, line 87: Simplify

rpc/grpc/server/services/blockservice/service.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	subscriber := fmt.Sprintf("subscriber_%s", id.String())
+
+	sub, err := s.eventBus.Subscribe(context.Background(), subscriber, types.QueryForEvent(types.EventNewBlock), 1)
+	if err != nil {
+		errMsg := "cannot subscribe to new block events"
+		s.logger.Error("GetLatestHeight", "err", errMsg)
+		return status.Error(codes.Internal, errMsg)
```
thanethomson, line 95: Simplify

rpc/grpc/server/services/blockservice/service.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	for {
+		select {
+		case msg := <-sub.Out():
+			switch eventType := msg.Data().(type) {
+			case types.EventDataNewBlock:
+				if err := stream.Send(&blocksvc.GetLatestHeightResponse{Height: eventType.Block.Height}); err != nil {
+					s.logger.Error("GetLatestHeight", "err", fmt.Sprintf("failed to stream a new block height %d to %s", eventType.Block.Height, subscriber))
```
thanethomson, line 104: Where'd the `err` go? It seems like it's just dropped/ignored here.

```suggestion
					s.logger.Error("Failed to stream new block height", "err", err, "height", eventType.Block.Height, "subscriber", subscriber)
```

rpc/grpc/server/services/blockservice/service.go
```diff
@@ -0,0 +1,129 @@
...
+					return status.Error(codes.Unavailable, "cannot send stream response")
+				}
+				s.logger.Debug("GetLatestHeight", "msg", fmt.Sprintf("streamed new block height %d", eventType.Block.Height))
+			}
+		case <-sub.Canceled():
+			switch sub.Err() {
+			case cmtpubsub.ErrUnsubscribed:
+				s.logger.Error("GetLatestHeight", "err", fmt.Sprintf("subscriber %s unsubscribed", subscriber))
```
thanethomson, line 112: Please stick to the usual convention for error logging, which is like:

```suggestion
				s.logger.Error("Internal subscriber terminated subscription", "subscriber", subscriber)
```

Same goes for all other logging operations introduced in this PR.

rpc/grpc/server/services/blockservice/service.go
```diff
@@ -0,0 +1,129 @@
...
+	// Generate a unique subscriber ID using a UUID
+	// The subscriber needs to be unique across all clients
+	id, err := uuid.NewUUID()
+	if err != nil {
+		// cannot generate unique id
+		errMsg := "error generating a subscriber id, cannot subscribe to new block events"
+		s.logger.Error("GetLatestHeight", "err", errMsg)
+		return status.Error(codes.Internal, errMsg)
+	}
+	subscriber := fmt.Sprintf("subscriber_%s", id.String())
```
thanethomson, line 89: Prefix is redundant.

```suggestion
	subscriber := id.String()
```

rpc/grpc/server/services/blockservice/service.go
```diff
@@ -0,0 +1,129 @@
...
+	blockIDProto := blockMeta.BlockID.ToProto()
+
+	return &blocksvc.GetByHeightResponse{
+		BlockId: &blockIDProto,
+		Block:   blockProto,
+	}, nil
+}
+
+// GetLatestHeight implements v1.BlockServiceServer GetLatestHeight method
+func (s *blockServiceServer) GetLatestHeight(_ *blocksvc.GetLatestHeightRequest, stream blocksvc.BlockService_GetLatestHeightServer) error {
```
thanethomson, line 79: If you'd like the endpoint name to appear in the context for all logging operations, simply just:

```golang
logger := s.logger.With("endpoint", "GetLatestHeight")
```

And then use `logger` throughout the rest of this method instead of `s.logger`.

test/e2e/tests/grpc_test.go
```diff
@@ -30,3 +30,88 @@ func TestGRPC_Version(t *testing.T) {
...
+	})
+}
+
+func TestGRPC_Block_GetLatestHeight(t *testing.T) {
+
+	testnet := loadTestnet(t)
+	node := testnet.RandomNode()
+	if node.Mode != e2e.ModeFull && node.Mode != e2e.ModeValidator {
+		return
+	}
```
thanethomson, line 91: This just skips the test entirely if the random node happens to be a light client - not sure that's super useful.

thanethomson (changes requested):

rpc/grpc/client/block_service.go
```diff
@@ -0,0 +1,103 @@
...
+
+	response := Block{
+		BlockID: *blockID,
+		Block:   block,
+	}
+	return &response, nil
+}
+
+// GetLatestHeight implements BlockServiceClient GetLatestHeight
+func (c *blockServiceClient) GetLatestHeight(ctx context.Context) (chan int64, chan error) {
```
thanethomson, line 66: I wouldn't recommend this approach of returning two channels.

Rather:
1. Only use a single channel with a result type that encapsulates both a height and an error ([example](https://stackoverflow.com/a/25142269/1156132)).
2. Allow the caller to supply a channel as you did previously. This allows the caller to control whether it's buffered/unbuffered and the channel size.

rpc/grpc/client/block_service.go
```diff
@@ -0,0 +1,123 @@
...
+// LatestHeightResult type used in GetLatestResult and send to the client
+// via a channel
+type LatestHeightResult struct {
+	Height int64
+	Error  error
+}
+
+// BlockServiceClient provides block information
+type BlockServiceClient interface {
+	GetBlockByHeight(ctx context.Context, height int64) (*Block, error)
```
glnro, line 27: Is is better for `GetBlockByHeight` on the client interface to take an `int64` height and then instantiate a `GetByHeightRequest` in method scope?

rpc/grpc/server/services/blockservice/service.go
```diff
@@ -0,0 +1,130 @@
...
+
+type blockServiceServer struct {
+	store    *store.BlockStore
+	eventBus *types.EventBus
+	logger   log.Logger
+}
+
+// New creates a new CometBFT version service server.
+func New(store *store.BlockStore, eventBus *types.EventBus, logger log.Logger) blocksvc.BlockServiceServer {
+	log := logger.With("module", "grpc-block-service")
```
thanethomson, line 25: The "module" field is already taken by the server: https://github.com/cometbft/cometbft/blob/902631a5f58acf0888c2ae0a98a63ff55f4a101f/rpc/grpc/server/server.go#L60

Rather:

```suggestion
	log := logger.With("service", "BlockService")
```

rpc/grpc/server/services/blockservice/service.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// retrieve the last height in the store
+	latestHeight := s.store.Height()
+
+	// validate height parameter, if height is 0 or
+	// the request is nil, then use the latest height
+	if req.Height == 0 {
+		height = latestHeight
+	} else if req.Height < 0 {
+		logger.Error("negative height request", "height", req.Height)
```
thanethomson, line 48: Please capitalize the first letter of all log messages:

```suggestion
		logger.Error("Negative height request", "height", req.Height)
```

Same goes for the rest of this file.

It's only when we're constructing `error`s that we should not capitalize the first letter of the string, since errors can be nested. Log messages aren't nested though.

rpc/grpc/client/block_service.go
thanethomson: Yes. This particular client wrapper interface should be as trivially ergonomic as possible. Just like in tendermint-rs, this involves "unwrapping" the underlying RPC request and response structures on behalf of the callers.

If callers want to use the underlying gRPC client directly, they're also welcome to do that.

thanethomson: Yes, this client wrapper interface should be as trivially ergonomic as possible. Just like in tendermint-rs' RPC client, this involves "unwrapping" key request struct members into function parameters (and similarly for responses).

Callers are always welcome to either generate their own gRPC clients, or use our generated one directly if they need access to them for some reason.

rpc/grpc/server/services/blockservice/service.go
```diff
@@ -0,0 +1,130 @@
...
+	sub, err := s.eventBus.Subscribe(context.Background(), subscriber, types.QueryForEvent(types.EventNewBlock), 1)
+	if err != nil {
+		logger.Error("cannot subscribe to new block events", "err", err)
+		return status.Error(codes.Internal, "Cannot subscribe to new block events")
+	}
+
+	for {
+		select {
+		case msg := <-sub.Out():
+			switch eventType := msg.Data().(type) {
```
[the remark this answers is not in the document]
thanethomson: Done here. Let me know whether you were thinking of something different?

docs/core/configuration.md
```diff
@@ -404,6 +438,7 @@ create_empty_blocks_interval = "0s"
 
 # Reactor sleep duration parameters
 peer_gossip_sleep_duration = "100ms"
+peer_gossip_intraloop_sleep_duration = "0s"
```
jmalicevic, line 441: Why is this change here?

rpc/grpc/server/services/blockservice/service.go
```diff
@@ -0,0 +1,127 @@
...
+		logger:   logger.With("service", "BlockService"),
+	}
+}
+
+// GetByHeight implements v1.BlockServiceServer GetByHeight method
+func (s *blockServiceServer) GetByHeight(_ context.Context, req *blocksvc.GetByHeightRequest) (*blocksvc.GetByHeightResponse, error) {
+	var height int64
+
+	// retrieve the last height in the store
+	latestHeight := s.store.Height()
```
jmalicevic, line 37: A check for the lowest height can also be added here:
`earliestHeight := s.store.Base()`

rpc/grpc/server/services/blockservice/service.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// validate height parameter, if height is 0 or
+	// the request is nil, then use the latest height
```
jmalicevic, line 40: How can the request be `nil` ?

rpc/grpc/server/services/blockservice/service.go
```diff
@@ -0,0 +1,127 @@
...
+		height = latestHeight
+	} else if req.Height < 0 {
+		return nil, status.Error(codes.InvalidArgument, "Negative height request")
+	} else {
+		height = req.Height
+	}
+
+	// check if the height requested is not higher
+	// than the latest height in the store
+	if height > latestHeight {
```
jmalicevic, line 51: I would simplify this with an internal function `checkHeightBounds` that should check whether the height conforms to the following conditions :
`height > 0 && height >= store.Base() && height <= store.height`

docs/core/configuration.md
thanethomson: When I added the sample config file to the docs, all I did was `cometbft init` locally and copy/paste the contents of `~/.cometbft/config/config.toml` into this doc. The fact that this was missing implies the doc was already out of date on `main`.

rpc/grpc/server/services/blockservice/service.go
thanethomson: It can't afaik (at least, if it's being called by the gRPC server, which is how this function's usually invoked), and there's no check here for `nil` anyways. I'll remove the comment.

docs/core/configuration.md
jmalicevic: Ok good, I was just double checking that some testing config did not make it into the PR. Sounds good!

thanethomson (approved):

PR #1150, opened 2023-07-19
ADR 101: Implement pruning mechanism

### LINKED ISSUE #1096: ADR-101: Implement block and block results pruning mechanism
Before we can implement #1097, we need to implement a pruning mechanism in the codebase to facilitate pruning of blocks and block results based on a combination of the application's retain height and the retain heights set by a data companion. The implementation must follow the logic outlined in ADR-101.

This can be implemented independently of #1097 on `main` and be backported to `v0.38.x` and `v0.37.x`.
```[tasklist]
### Tasks
- [ ] https://github.com/cometbft/cometbft/pull/1150
```

jmalicevic:
Closes #1096

This PR introduces a pruning mechanism for pruning the blockstore and state store, and separate pruning of the abci results. 

Whether to prune is indicated either by the application (for the blockstore and state store via the `retainHeight`  return parameter)  or a data companion attached via the data companion API (pruning both the stores and the ABCI responses). 

Still missing in the PR :

- configuration parameter for the data companion to indicate it's desired pruning height on start up. This height  will be read on boot-up and taken into consideration when the application requests blocks to be pruned
- stopping the pruning service gracefully when a node shuts down. 

---

#### PR checklist

- [x] Tests written/updated


state/prunning_service.go
```diff
@@ -0,0 +1,228 @@
...
+	case AppRequester:
+		currentRetainHeight, err = p.stateStore.GetApplicationRetainHeight()
+		if err != nil {
+			if err == ErrKeyNotFound {
+				err = p.stateStore.SaveApplicationRetainHeight(retainHeightInfo.Height)
+				return retainHeightInfo.Height, err
+			}
+			return 0, err
+		}
+		if currentRetainHeight > retainHeightInfo.Height {
```
jmalicevic, line 93: We should compare the existing data companion retain height as well and not accept the app retain height if a higher dc retain height has already been set.

state/prunning_service.go
```diff
@@ -0,0 +1,228 @@
...
+	var state State
+	state, err = p.stateStore.Load()
+	if err != nil {
+		p.logger.Error("failed to load state, cannot prune")
+		return
+	}
+	p.logger.Info(fmt.Sprintf("Received pruning request for %d. Accepted retain height is : %d", height, height))
+	pruned, evRetainHeight, err = p.bs.PruneBlocks(height, state)
+	if err != nil {
+		p.logger.Error(fmt.Sprintf("failed to prune blocks for retain height %d with error: %s", height, err))
```
jmalicevic, line 213: return err

state/prunning_service.go
```diff
@@ -0,0 +1,228 @@
...
+	Height    int64
+	Requester PruningRequester
+}
+
+func NewPruner(stateStore Store, bs BlockStore, logger log.Logger) *Pruner {
+	p := &Pruner{
+		bs:            bs,
+		stateStore:    stateStore,
+		logger:        logger,
+		pruningTicker: time.NewTicker(time.Second * 10),
```
jmalicevic, line 46: Frequency should be customizable.

state/prunning_service.go
```diff
@@ -0,0 +1,228 @@
...
+)
+
+type Pruner struct {
+	service.BaseService
+	logger log.Logger
+	// DB to which we save the retain heights
+	bs BlockStore
+	// State store to prune state from
+	stateStore    Store
+	pruningTicker *time.Ticker
```
thanethomson, line 33: What's the benefit of using a ticker here as opposed to just configuring an interval and doing a `time.Sleep` between pruning operations? Would they have a similar effect?

I suppose as long as you don't create a hot loop of pruning operations with the ticker-based approach in cases where pruning takes longer than the pruning ticker timeout (i.e. > 10s) it should be okay.

state/prunning_service.go
```diff
@@ -0,0 +1,228 @@
+package state
```
thanethomson, line 1: This file should probably be called `state/pruner.go` to coincide with the name of the primary struct provided.

state/prunning_service.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+func NewPruner(stateStore Store, bs BlockStore, logger log.Logger) *Pruner {
+	p := &Pruner{
+		bs:            bs,
+		stateStore:    stateStore,
+		logger:        logger,
+		pruningTicker: time.NewTicker(time.Second * 10),
+	}
+	p.BaseService = *service.NewBaseService(logger, "Prunning service", p)
```
thanethomson, line 48:
```suggestion
	p.BaseService = *service.NewBaseService(logger, "Pruner", p)
```

state/prunning_service.go
```diff
@@ -0,0 +1,228 @@
...
+	return p
+}
+
+// If OnStart() returns an error, it's returned by Start()
```
thanethomson, line 52:
```suggestion
```

state/prunning_service.go
```diff
@@ -0,0 +1,228 @@
...
+	if err := p.BaseService.OnStart(); err != nil {
+		return err
+	}
+	go p.pruningRoutine()
+	return nil
+}
+
+// Stop the service.
+// If it's already stopped, will return an error.
+// OnStop must never error.
```
thanethomson, line 63:
```suggestion
```

state/prunning_service.go
```diff
@@ -0,0 +1,228 @@
...
+	p.BaseService.OnStop()
+
+}
+
+// SetPruningHeight is called by either the application or the data companion to set the
+// retain height for blocks or ABCI block results.
+// It returns the height which will be used as the retain height and an error message in case
+// the retain height cannot be set to the desired value (if the data was already pruned, if a lower
+// retain height was already set)
+func (p *Pruner) SetPruningHeight(retainHeightInfo RetainHeightInfo) (retainHeight int64, err error) {
```
thanethomson, line 75: I'd recommend splitting this method into 3 different ones to reflect the business logic better at compile-time (the current approach effectively does this at runtime, making it more difficult to test):
1. `SetApplicationRetainHeight` to handle the `AppRequester` part of the `switch` statement
2. `SetCompanionBlockRetainHeight` (to be specific about the fact that it influences block retention) to handle the `DataCompanionRequester` part of the `switch` statement
3. `SetABCIResRetainHeight` to handle the `ABCIResRetainHeightRequester` part of the `switch` statement

A general rule of thumb is that each function should only do one thing (and ideally only 20-30 lines of code per function so it's quick and easy to read - not always possible, of course, but something to aim for).

state/prunning_service.go
```diff
@@ -0,0 +1,228 @@
...
+	if err != nil {
+		p.logger.Error(fmt.Sprintf("failed to prune the state store with error: %s ", err))
+	}
+	return pruned, evRetainHeight, err
+}
+
+func (p *Pruner) PruneABCIResponses(height int64) (pruned uint64, err error) {
+	pruned, err = p.stateStore.PruneABCIResponses(height)
+	return
+}
```
thanethomson, line 228:
```suggestion
func (p *Pruner) PruneABCIResponses(height int64) (uint64, error) {
	return p.stateStore.PruneABCIResponses(height)
}
```

Also, does this method really need to be public? If not I'd recommend getting rid of it altogether in favour of accessing `p.stateStore.PruneABCIResponses` directly.

state/prunning_service.go
```diff
@@ -0,0 +1,228 @@
...
+func (p *Pruner) pruningRoutine() {
+	lastHeightPruned := int64(0)
+	lastABCIResPrunedHeight := int64(0)
+	for {
+		select {
+		case <-p.pruningTicker.C:
+			retainHeight := p.findMinRetainHeight()
+			if retainHeight != lastHeightPruned {
+				pruned, evRetainHeight, err := p.pruneBlocks(retainHeight)
+				p.logger.Info(fmt.Sprintf("Pruned %d, evidence retain height is %d , err %s", pruned, evRetainHeight, err))
```
thanethomson, line 139: We should avoid using `fmt.Sprintf` in log messages. Rather:

```suggestion
				if err != nil {
					p.logger.Error("Failed to prune blocks", "err", err)
				} else {
					p.logger.Debug("Pruned block(s)", "height", pruned, "evidenceRetainHeight", evRetainHeight)
				}
```

The error should most likely be logged at a different level to the info regarding which heights were pruned. We should be very careful with using `Info` level logging, since we may end up polluting operators' logs.

If we log anything at `Info` level during pruning operations, it should exclusively be a summary and only when something was actually pruned (I imagine this would be useful for operators to see to know that pruning's working).

state/store.go
```diff
@@ -73,6 +79,20 @@ type Store interface {
...
 	// PruneStates takes the height from which to start pruning and which height stop at
 	PruneStates(int64, int64, int64) error
+	// PruneABCIResponses will prune all ABCI responses below the given height.
+	PruneABCIResponses(int64) (uint64, error)
+	// SaveApplicationRetainHeight persists the application retain height from the application
+	SaveApplicationRetainHeight(height int64) error
+	// GetApplicationRetainHeight returns the retain height set by the application
+	GetApplicationRetainHeight() (int64, error)
+	// SaveDataCompanionRetainHeight saves the retain height set by the data companion
+	SaveDataCompanionRetainHeight(height int64) error
```
thanethomson, line 89: Should probably be called `SaveCompanionBlockRetainHeight` to make it clear that this only influences block retention. (I dropped the `Data` part because otherwise the function name would be crazy long)

state/store.go
```diff
@@ -73,6 +79,20 @@ type Store interface {
...
+	// GetDataCompanionRetainHeight returns the retain height set by the data companion
+	GetDataCompanionRetainHeight() (int64, error)
```
thanethomson, line 91: Similarly to the previous comment, should probably be called something like `GetCompanionBlockRetainHeight`.

light/provider/http/http_test.go
```diff
@@ -70,7 +70,9 @@ func TestProvider(t *testing.T) {
 	lb, err = p.LightBlock(context.Background(), lower)
 	require.NoError(t, err)
 	assert.Equal(t, lower, lb.Height)
-
+	// Sleeping because pruning is now asynchrnous and might not yet
```
thanethomson, line 73:
```suggestion
	// Sleeping because pruning is now asynchronous and might not yet
```

state/execution.go
```diff
@@ -42,7 +42,8 @@ type BlockExecutor struct {
 
 	logger log.Logger
 
-	metrics *Metrics
+	metrics        *Metrics
+	pruningService *Pruner
```
thanethomson, line 46: Shouldn't this be called `pruner`, similar to the name of the struct?

state/prunning_service.go
```diff
@@ -0,0 +1,228 @@
...
+	ABCIResRetainHeightRequester PruningRequester = 2
+)
+
+var (
+	AppRetainHeightKey           = []byte("AppRetainHeightKey")
+	DataCompanionRetainHeightKey = []byte("DCRetainHeightKey")
+	ABCIResultsRetainHeightKey   = []byte("ABCIResRetainHeightKey")
+)
+
+type Pruner struct {
```
thanethomson, line 26: Needs a docstring to describe it.

state/prunning_service.go
thanethomson: Yeah it should just be a parameter supplied to the `NewPruner` constructor. You could also use the options pattern (like in the [gRPC server constructor](https://github.com/cometbft/cometbft/blob/902631a5f58acf0888c2ae0a98a63ff55f4a101f/rpc/grpc/server/server.go#L77) and elsewhere in the code) for all optional parameters past the `logger`.

That way, usage like the following would be possible:

```golang
// Create a pruner with the default pruning interval.
pruner := state.NewPruner(stateStore, blockStore, logger)

// Create a pruner, overriding the default pruning interval.
pruner := state.NewPruner(stateStore, blockStore, logger, state.PruneInterval(60 * time.Second))
```

Alternatively you could make it a compulsory parameter, but it seems like it'd be useful to make it optional in the constructor.

thanethomson (commented):
I did a first pass review here, and it looks like it's starting to take shape, thanks @jmalicevic!

state/prunning_service.go
jmalicevic: Hmm I did not think of that, that's  a good point. I don't think there is a way currently to prevent this behaviour. Epecially if we allow the operator to configure this parameter.  I initially did it with sleep but then chose this way because I found it cleaner to have a way to accept a signal to cancel the go routine and break the loop (which can be done even w/o a ticker).  But given your comment, it's much safer to just `time.Sleep`

state/prunning_service.go
jmalicevic: Actually this is dead code I did not delete. The pruning function is using the `stateStore.PruneABCIResponses` directly.

config/toml.go
```diff
@@ -550,4 +550,26 @@ max_open_connections = {{ .Instrumentation.MaxOpenConnections }}
 
 # Instrumentation namespace
 namespace = "{{ .Instrumentation.Namespace }}"
+
+#######################################################
+###       Data Companion Configuration Options          ###
+#######################################################
+[datacompanion]
+
+# Data Companion is enabled and we take the initial height as the official retain height:
```
adizere, line 559: I don't understand this comment. I suspect `enabled` likely enables the DC, but as a user I might find the comment confusing.

Can we describe what `enabled` and `initial_retain_height` respectively do?

thanethomson: I expanded on this in the ADR: https://github.com/cometbft/cometbft/blob/thane/adr-084-data-companion-pull-api/docs/architecture/adr-101-data-companion-pull-api.md#configuration

We'll need to update this configuration approach to match the ADR.

state/pruner.go
```diff
@@ -63,18 +63,10 @@ func NewPruner(stateStore Store, bs BlockStore, logger log.Logger, options ...Pr
 }
 
 func (p *Pruner) OnStart() error {
-	if err := p.BaseService.OnStart(); err != nil {
```
jmalicevic: Thanks Thane! Btw, we use this exact same code in the p2p service, is there a reason why its used there ? Or should we also refactor it there as well.

node/node.go
```diff
@@ -61,8 +62,9 @@ type Node struct {
 	eventBus          *types.EventBus // pub/sub for services
 	stateStore        sm.Store
 	blockStore        *store.BlockStore // store the blockchain to disk
-	bcReactor         p2p.Reactor       // for block-syncing
-	mempoolReactor    p2p.Reactor       // for gossipping transactions
+	pruner            *sm.Pruner
```
jmalicevic, line 65: 100% agree it should be here instead of the executor.

state/pruner.go
thanethomson: I wouldn't refactor the P2P code yet until we're actually changing something there. We should honestly refactor this whole `Service` model entirely at some point - it's trying to simulate some kind of object-oriented way of programming that isn't at all ergonomic or idiomatic in Go.

thanethomson (approved):

node/node.go
```diff
@@ -426,6 +445,11 @@ func (n *Node) OnStart() error {
 		}
 	}
 
+	// Start background pruning
+	if err := n.pruner.Start(); err != nil {
+		return fmt.Errorf("failed to start background pruning routine: %w", err)
```
glnro, line 450: Is it worth logging something here or no?

thanethomson: What would you log - the error? If so, that's output here when running Comet as a standalone server: https://github.com/cometbft/cometbft/blob/24ebdd9df78d15828aa758e16a21d82fadcbb95b/cmd/cometbft/commands/run_node.go#L106-L108

The SDK would get the error when calling `Node.Start()` and they could choose what they'd want to do with it.

state/store.go
```diff
@@ -366,6 +389,51 @@ func (store dbStore) PruneStates(from int64, to int64, evidenceThresholdHeight i
 	return nil
 }
 
+// PruneABCIResponses attempts to prune all ABCI responses up to, but not
+// including, the given height. On success, returns the height to which ABCI
+// responses were pruned.
+func (store dbStore) PruneABCIResponses(height int64) (uint64, error) {
```
thanethomson, line 395: @glnro and I were talking about this method today and realized we should refactor it a bit. There are two main changes I've implemented here that I'd appreciate your perspective on @jmalicevic:

1. The pruned height counting mechanism's been refactored.
2. We're now caching the last ABCI responses pruned height. I imagine that in chains at large heights (e.g. over 1 million), this would be substantially sub-optimal to run through all possible heights for every pruning operation. @glnro recommended supplying a `fromHeight` and `toHeight` parameter to this method instead of just `height`, but thinking about it now, under what circumstances would we want to be able to offer pruning of selective chunks of ABCI responses? Seems to me as though, if we want to prune up to a particular height, all of the relevant mechanisms for that pruning should be self-contained.

glnro: That being said there could be a use case for selectively pruning a range but I imagine this would be for extenuating circumstances (debugging). Conceivably we could pass a `fromHeight` and always call it with the latest pruned height.

state/pruner.go
```diff
@@ -0,0 +1,267 @@
...
+	currentCompanionRetainHeight, err := p.stateStore.GetCompanionBlockRetainHeight()
+	if err != nil {
+		if err == ErrKeyNotFound {
+			currentCompanionRetainHeight = height
+		} else {
+			return err
+		}
+	}
+	currentAppRetainHeight, err := p.stateStore.GetApplicationRetainHeight()
+	var noAppRetainHeight bool
```
glnro, line 129: Shouldn't this be initialized to false?

state/pruner.go
```diff
@@ -0,0 +1,267 @@
...
+// SetApplicationRetainHeight sets the application retain height with some
+// basic checks on the requested height.
+//
+// If a higher retain height is already set, we cannot accept the requested
+// height because the blocks might have been pruned.
+//
+// If the data companion has already set a retain height to a higher value we
+// also cannot accept the requested height as the blocks might have been
+// pruned.
+func (p *Pruner) SetApplicationRetainHeight(height int64) error {
```
glnro, line 80: This might be unecessary, but both `SetApplicationRetainHeight` and `SetCompanionRetainHeight` are near identical...could consolidate into a generalized `SetRetainHeight` and pass in the relevant function calls as args

```go
func (p *Pruner) SetRetainHeight(
	height int64,
	getServiceRetainHeight func() (int64, error),
	getAlternateRetainHeight func() (int64, error),
	saveServiceRetainHeight func(int642 int64) error,
) error {
	if height <= 0 || height < p.bs.Base() || height > p.bs.Height() {
		return ErrInvalidHeightValue
	}
	currentServiceRetainHeight, err := getServiceRetainHeight()
	if err != nil {
		if err == ErrKeyNotFound {
			currentServiceRetainHeight = height
		} else {
			return err
		}
	}
	currentAltRetainHeight, err := getAlternateRetainHeight()
	var noAltRetainHeight bool
	if err != nil {
		if err == ErrKeyNotFound {
			noAltRetainHeight = true
		} else {
			return err
		}
	}
	if currentServiceRetainHeight > height || (!noAltRetainHeight && currentAltRetainHeight > height) {
		return errors.New("cannot set a height lower than previously requested - blocks might have already been pruned")
	}
	err = saveServiceRetainHeight(height)
	return err
}
``` 

Then call in `execution.go`
```go
err := blockExec.pruner.SetRetainHeight(
			retainHeight,
			blockExec.pruner.stateStore.GetApplicationRetainHeight,
			blockExec.pruner.stateStore.GetCompanionBlockRetainHeight,
			blockExec.pruner.stateStore.SaveApplicationRetainHeight,
		)
```

state/pruner.go
jmalicevic: Is it not initialized with the default value for bool (false)?

state/pruner.go
```diff
@@ -0,0 +1,267 @@
...
+		if errors.Is(err, ErrKeyNotFound) {
+			return 0
+		}
+		p.logger.Error("Failed to get ABCI result retain height", "err", err)
+	} else if lastABCIResPrunedHeight != abciResRetainHeight {
+		prunedHeight, err := p.stateStore.PruneABCIResponses(abciResRetainHeight)
+		if err != nil {
+			p.logger.Error("Failed to prune ABCI responses", "err", err, "abciResRetainHeight", abciResRetainHeight)
+		}
+		p.logger.Info("Pruned ABCI responses", "height", prunedHeight)
```
jmalicevic, line 211: Here we miss setting `lastABCIPrunedHeight` to `abciResRetainHeight` as we are doing in the method above for `lastHeightPruned`

state/store.go
jmalicevic: Hey so here is my take:
1. I think it's a good way to improve the count and we should improve the same for counting the blocks pruned. 
2. We were always caching it but not in a persistent way. I just saw that the refactored code was not actually updating the `lastABCIResPrunedHeight`, but with that fixed, the pruner always remembers the height up to which it pruned the ABCI results and prunes again, only if the retain height is different from the last height up to which we pruned. And we always prune from height 1 to `retainHeight`. Which means that `lastABCIResPrunedHeight < retainHeight` for all correct cases. 
That being said, if pruning works correctly, line 414 will error once we hit a height (which is height 1 right now) that does not exist and we should indeed always be pruning backwards from `retainHeight` until either: 
 a) We hit an error , in which we should flush the batch up until that point (which we are not doing now)
 b) We pass `lastABCIResPrunedHeight` as a `fromHeight` and prune between that and the `retainHeight`. If this is not persistent though it is not the best way. 

I do not see a use case where we should partially delete the responses (between two heights). It should always be from 1 to `retainHeight` and then from there upwards. Otherwise we open ourselves for a bunch of corner cases. That is why I would not pass a second parameter to the function, but rather go with option `a)` and check for a `ErrKeyNotFound` on delete, and return. Or we persist the parameter and check for it internally, without passing it as a function parameter.

jmalicevic: To sum up, the idea was always to cache this but the old way of caching would actually not work because the `Delete` statement would always error after the initial pruning is done. So we either keep that way but prune backwards until we error or do it the way you are doing it now with persisting the info but not passing it as a `fromHeight` parameter ,rather reading it within the function. 
For block pruning, there is no reason to persist this because we implicitly do this by updating `store.Base()` on pruning which is always checked so we have this info.

thanethomson: The backwards pruning seems more optimal to me. If we go that route, and we get an error, is it a specific kind of error that we can detect like `ErrKeyNotFound`? That would then do away with the need for tracking the last retain height in the DB and we could revert those changes of mine.

jmalicevic: I believe yes but I can do a quick check. If that is the case, I'll push a version without persisting the last height in the DB.

state/pruner.go
thanethomson: I would argue it probably makes more sense for it to stay the way it is. Duplication often results in better readability and greater simplicity.

state/pruner.go
```diff
@@ -0,0 +1,274 @@
...
+	currentCompanionRetainHeight, err := p.stateStore.GetCompanionBlockRetainHeight()
+	if err != nil {
+		if err == ErrKeyNotFound {
+			currentCompanionRetainHeight = height
+		} else {
+			return err
+		}
+	}
+	currentAppRetainHeight, err := p.stateStore.GetApplicationRetainHeight()
+	var noAppRetainHeight bool
```
thanethomson, line 129:
```suggestion
	noAppRetainHeight := false
```

is clearer

state/pruner.go
```diff
@@ -0,0 +1,274 @@
...
+	if retainHeight != lastHeightPruned {
+		pruned, evRetainHeight, err := p.pruneBlocks(retainHeight)
+		if err != nil {
+			p.logger.Error("Failed to prune blocks", "err", err, "blockRetainHeight", retainHeight)
+		} else if pruned > 0 {
+			p.logger.Info("Pruned blocks", "count", pruned, "evidenceRetainHeight", evRetainHeight)
+		}
+		// The last height up until which we pruned is the current lowest point of the block store
+		// indicated by Base()
+		lastHeightPruned = p.bs.Base()
```
thanethomson, line 194: Shouldn't this be:

```suggestion
		lastHeightPruned = p.bs.Base() - 1
```

because the base height is the lowest height we _have_, not the height to which we've pruned? The height to which we've pruned (which I thought was what was being returned here) is just below the base height.

jmalicevic: Good point! You are totally right.

state/store.go
```diff
@@ -366,6 +389,51 @@ func (store dbStore) PruneStates(from int64, to int64, evidenceThresholdHeight i
...
+			batchPruned = 0
+			if err := store.setLastABCIResponsesPruneHeight(h); err != nil {
+				return pruned, lastPruneHeight + pruned, fmt.Errorf("failed to set last ABCI responses prune height: %w", err)
+			}
+
+			batch = store.db.NewBatch()
+			defer batch.Close()
+		}
+	}
+	return pruned + batchPruned, height, batch.WriteSync()
```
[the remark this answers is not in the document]
jmalicevic: Yes, its the same bug you found in the block pruning (where it was `Base()` instead of `Base() - 1`.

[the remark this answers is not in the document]
jmalicevic: Actually on second thought, this is compared to the `retainHeight` which is exclusive. So  in this case it should either stay the way it is or it should be compared with `abciResRetainHeight - 1`.

state/store_test.go
```diff
@@ -242,6 +245,204 @@ func sliceToMap(s []int64) map[int64]bool {
...
+	pruner := sm.NewPruner(stateStore, bs, log.TestingLogger())
+
+	retainHeight := int64(2)
+	err = stateStore.SaveABCIResRetainHeight(retainHeight)
+	require.NoError(t, err)
+	abciResRetainHeight, err := stateStore.GetABCIResRetainHeight()
+	require.NoError(t, err)
+	require.Equal(t, retainHeight, abciResRetainHeight)
+	pruned := pruner.PruneABCIResToRetainHeight(0)
+	require.Equal(t, retainHeight, pruned)
```
thanethomson, line 360: Should this not be:

```suggestion
	require.Equal(t, retainHeight - 1, pruned)
```

We should always be pruning up to, but just before, the retain height, should we not?

thanethomson: I suppose what I'm confused about is what we _should_ be returning from `pruner.pruneABCIResToRetainHeight` - should we be returning the new retain height, or the height to which we've pruned?

jmalicevic: The way it is currently implemented it returns the last height **up** to which we pruned (same as retain height). However, we can change this if it is more intuitive for you the other way. The only purpose of this is not to prune the same blocks twice, so this variable can mean whatever we want.

thanethomson: But we shouldn't ever be pruning the retain height - only just before the retain height, should we not?

So the return value is therefore the new retain height, and not the pruned height, since the pruned height should always be `retainHeight - 1`.

As such, do we need to even return the new retain height? All we need to know is whether it was successful or not in reaching the new retain height.

jmalicevic: So the naming of the variable is the source of the confusion. We do not prune the retain height. We prune up to but exclude the retain height. The variable returned here is purely for optimization purposes to avoid having the pruner pruning if the new retain height is the same as the one we already had before. In some cases we will not prune up until teh requested height (due to an error, or the evidence retain height in the blockstore). In that case we want to prune again if this 'lastHeightPruned' differs from the retain height in the database.

state/pruner.go
```diff
@@ -0,0 +1,277 @@
...
+			p.logger.Error("Failed to prune blocks", "err", err, "blockRetainHeight", retainHeight)
+		} else if pruned > 0 {
+			p.logger.Info("Pruned blocks", "count", pruned, "evidenceRetainHeight", evRetainHeight)
+		}
+		// The last height up until which we pruned is the current lowest point of the block store
+		// indicated by Base()
+		lastHeightPruned = p.bs.Base() - 1
+	}
+	return lastHeightPruned
+}
```
thanethomson, line 197:
```suggestion
func (p *Pruner) pruneBlocksToRetainHeight(lastRetainHeight int64) int64 {
	retainHeight := p.findMinRetainHeight()
	if retainHeight != lastRetainHeight {
		pruned, evRetainHeight, err := p.pruneBlocks(retainHeight)
		if err != nil {
			p.logger.Error("Failed to prune blocks", "err", err, "blockRetainHeight", retainHeight)
		} else if pruned > 0 {
			p.logger.Info("Pruned blocks", "count", pruned, "evidenceRetainHeight", evRetainHeight)
		}
		// The last height up until which we pruned is the current lowest point of the block store
		// indicated by Base()
		lastRetainHeight = p.bs.Base()
	}
	return lastRetainHeight
}
```

state/pruner.go
```diff
@@ -0,0 +1,277 @@
...
+	lastHeightPruned := int64(0)
+	lastABCIResPrunedHeight := int64(0)
+	for {
+		select {
+		case <-p.Quit():
+			return
+		default:
+			lastHeightPruned = p.pruneBlocksToRetainHeight(lastHeightPruned)
+			lastABCIResPrunedHeight = p.pruneABCIResToRetainHeight(lastABCIResPrunedHeight)
+			time.Sleep(p.interval)
```
thanethomson, line 178:
```suggestion
	lastRetainHeight := int64(0)
	lastABCIResRetainHeight := int64(0)
	for {
		select {
		case <-p.Quit():
			return
		default:
			lastRetainHeight = p.pruneBlocksToRetainHeight(lastRetainHeight)
			lastABCIResRetainHeight = p.pruneABCIResToRetainHeight(lastABCIResRetainHeight)
			time.Sleep(p.interval)
```

state/pruner.go
```diff
@@ -0,0 +1,277 @@
...
+	// lastPrunedHeight is the last height up until which we pruned successfully
+	// In case of an error it will be 0, but then it will also be ignored.
+	numPruned, lastPrunedHeight, err := p.stateStore.PruneABCIResponses(abciResRetainHeight)
+	if err != nil {
+		p.logger.Error("Failed to prune ABCI responses", "err", err, "abciResRetainHeight", abciResRetainHeight)
+		return lastABCIResPrunedHeight
+	}
+	p.logger.Info("Pruned ABCI responses", "height", numPruned)
+	return lastPrunedHeight
+}
```
thanethomson, line 224:
```suggestion
func (p *Pruner) pruneABCIResToRetainHeight(lastABCIResRetainHeight int64) int64 {
	abciResRetainHeight, err := p.stateStore.GetABCIResRetainHeight()
	if err != nil {
		// ABCI response retain height has not yet been set - do not log any
		// errors at this time.
		if errors.Is(err, ErrKeyNotFound) {
			return 0
		}
		p.logger.Error("Failed to get ABCI result retain height", "err", err)
		return lastABCIResRetainHeight
	}

	if lastABCIResRetainHeight == abciResRetainHeight {
		return lastABCIResRetainHeight
	}

	// newRetainHeight is the last height just before which we pruned successfully.
	// In case of an error it will be 0, but then it will also be ignored.
	numPruned, newRetainHeight, err := p.stateStore.PruneABCIResponses(abciResRetainHeight)
	if err != nil {
		p.logger.Error("Failed to prune ABCI responses", "err", err, "abciResRetainHeight", abciResRetainHeight)
		return lastABCIResRetainHeight
	}
	p.logger.Info("Pruned ABCI responses", "height", numPruned)
	return newRetainHeight
}
```

state/store.go
```diff
@@ -366,6 +389,51 @@ func (store dbStore) PruneStates(from int64, to int64, evidenceThresholdHeight i
...
+			pruned += batchPruned
+			batchPruned = 0
+			if err := store.setLastABCIResponsesPruneHeight(h); err != nil {
+				return pruned, lastPruneHeight + pruned, fmt.Errorf("failed to set last ABCI responses prune height: %w", err)
+			}
+
+			batch = store.db.NewBatch()
+			defer batch.Close()
+		}
+	}
```
thanethomson, line 433: I forgot to add this previously:

```suggestion
	}
	if err := store.setLastABCIResponsesPruneHeight(height); err != nil {
		return pruned, height, fmt.Errorf("failed to set last ABCI responses prune height: %w", err)
	}
```

state/pruner.go
```diff
@@ -0,0 +1,277 @@
...
+
+// SetCompanionRetainHeight sets the application retain height with some basic
+// checks on the requested height.
+//
+// If a higher retain height is already set, we cannot accept the requested
+// height because the blocks might have been pruned.
+//
+// If the application has already set a retain height to a higher value we also
+// cannot accept the requested height as the blocks might have been pruned.
+func (p *Pruner) SetCompanionRetainHeight(height int64) error {
```
mzabaluev, line 116: This API uses signed integers for height, whereas in ADR-101 protobufs, the height fields are specified as `uint64` and this type is emitted in the generated gRPC interface in #1154. To avoid tedious checking logic everywhere, can we agree on a single integer type? Cc @thanethomson

This might be too late to ask, as I see both `int64` and `uint64` used for height throughout the various proto packages. Sometimes both in the same package!

state/export_test.go
```diff
@@ -44,3 +44,13 @@ func SaveValidatorsInfo(db dbm.DB, height, lastHeightChanged int64, valSet *type
...
+
+// FindMinRetainHeight is an alias for the private findMinRetainHeight method
+// in pruner.go, exported exclusively and expicitly for testing.
+func (p *Pruner) FindMinRetainHeight() int64 {
+	return p.findMinRetainHeight()
+}
+
+func (p *Pruner) PruneABCIResToRetainHeight(lastABCIResPrunedHeight int64) int64 {
+	return p.pruneABCIResToRetainHeight(lastABCIResPrunedHeight)
+}
```
thanethomson, line 56:
```suggestion
func (p *Pruner) PruneABCIResToRetainHeight(lastABCIResRetainHeight int64) int64 {
	return p.pruneABCIResToRetainHeight(lastABCIResRetainHeight)
}
```

state/store.go
```diff
@@ -366,6 +389,51 @@ func (store dbStore) PruneStates(from int64, to int64, evidenceThresholdHeight i
 	return nil
 }
 
+// PruneABCIResponses attempts to prune all ABCI responses up to, but not
+// including, the given height. On success, returns the height to which ABCI
+// responses were pruned.
```
thanethomson, line 394:
```suggestion
// including, the given height. On success, returns the number of heights
// pruned and the new retain height.
```

state/store.go
```diff
@@ -366,6 +389,51 @@ func (store dbStore) PruneStates(from int64, to int64, evidenceThresholdHeight i
...
+	if store.DiscardABCIResponses {
+		return 0, 0, errors.New("ABCI responses are discarded, nothing to prune")
+	}
+	lastPruneHeight, err := store.getLastABCIResponsesPruneHeight()
+	if err != nil {
+		return 0, 0, fmt.Errorf("failed to look up last ABCI responses prune height: %w", err)
+	}
+	if lastPruneHeight == 0 {
+		lastPruneHeight = 1
+	}
```
thanethomson, line 405:
```suggestion
	lastRetainHeight, err := store.getLastABCIResponsesRetainHeight()
	if err != nil {
		return 0, 0, fmt.Errorf("failed to look up last ABCI responses retain height: %w", err)
	}
	if lastRetainHeight == 0 {
		lastRetainHeight = 1
	}
```

state/store.go
```diff
@@ -366,6 +389,51 @@ func (store dbStore) PruneStates(from int64, to int64, evidenceThresholdHeight i
...
+			if err := batch.Write(); err != nil {
+				return pruned, lastPruneHeight + pruned, fmt.Errorf("failed to write ABCI responses deletion batch at height %d: %w", h, err)
+			}
+			batch.Close()
+
+			pruned += batchPruned
+			batchPruned = 0
+			if err := store.setLastABCIResponsesPruneHeight(h); err != nil {
+				return pruned, lastPruneHeight + pruned, fmt.Errorf("failed to set last ABCI responses prune height: %w", err)
+			}
```
thanethomson, line 428:
```suggestion
	for h := lastRetainHeight; h < height; h++ {
		if err := batch.Delete(calcABCIResponsesKey(h)); err != nil {
			return pruned, lastRetainHeight + pruned, fmt.Errorf("failed to delete ABCI responses at height %d: %w", h, err)
		}
		batchPruned++
		if batchPruned%1000 == 0 && batchPruned > 0 {
			if err := batch.Write(); err != nil {
				return pruned, lastRetainHeight + pruned, fmt.Errorf("failed to write ABCI responses deletion batch at height %d: %w", h, err)
			}
			batch.Close()

			pruned += batchPruned
			batchPruned = 0
			if err := store.setLastABCIResponsesRetainHeight(h); err != nil {
				return pruned, lastRetainHeight + pruned, fmt.Errorf("failed to set last ABCI responses retain height: %w", err)
			}
```

state/store.go
```diff
@@ -499,6 +567,91 @@ func (store dbStore) SaveFinalizeBlockResponse(height int64, resp *abci.Response
...
+
+	if height < 0 {
+		return 0, ErrInvalidHeightValue
+	}
+
+	return height, nil
+}
+
+func (store dbStore) getLastABCIResponsesPruneHeight() (int64, error) {
+	bz, err := store.getValue(lastABCIResponsesPruneHeightKey)
```
thanethomson, line 640:
```suggestion
func (store dbStore) getLastABCIResponsesRetainHeight() (int64, error) {
	bz, err := store.getValue(lastABCIResponsesRetainHeightKey)
```

state/store.go
```diff
@@ -499,6 +567,91 @@ func (store dbStore) SaveFinalizeBlockResponse(height int64, resp *abci.Response
...
+	}
+	height := int64FromBytes(bz)
+	if height < 0 {
+		return 0, ErrInvalidHeightValue
+	}
+	return height, nil
+}
+
+func (store dbStore) setLastABCIResponsesPruneHeight(height int64) error {
+	return store.db.SetSync(lastABCIResponsesPruneHeightKey, int64ToBytes(height))
```
thanethomson, line 652:
```suggestion
func (store dbStore) setLastABCIResponsesRetainHeight(height int64) error {
	return store.db.SetSync(lastABCIResponsesRetainHeightKey, int64ToBytes(height))
```

state/store.go
```diff
@@ -40,7 +46,10 @@ func calcABCIResponsesKey(height int64) []byte {
 
 //----------------------
 
-var lastABCIResponseKey = []byte("lastABCIResponseKey")
+var (
+	lastABCIResponseKey             = []byte("lastABCIResponseKey")
+	lastABCIResponsesPruneHeightKey = []byte("lastABCIResponsesPruneHeight")
```
thanethomson, line 51:
```suggestion
	lastABCIResponsesRetainHeightKey = []byte("lastABCIResponsesRetainHeight")
```

thanethomson (commented):
Sorry for the back and forth on this, but from what I can tell I was probably using the wrong terminology in various places here: we should be referring to retain heights in general as opposed to pruning heights, since there's an off-by-one difference between the two, and mixing them ends up being confusing.

I've therefore collapsed a bunch of recommendations, including updating this PR from the feature branch, into #1201. So the order of merging that I'd recommend would be:
1. Merge #1201 into this branch by way of a merge commit to preserve the history.
2. Merge #1183 into this branch by way of squash-merging, as that PR's history isn't as important.
3. Merge this PR into the feature branch.

thanethomson:
> We noticed that indexer pruning was interfering (slowing down) the block pruning. So [Anton separated](https://github.com/cometbft/cometbft/pull/1176/commits/188fcf503f0ac4b9660841e68078631d32c155b1) that into a different go routine. I wonder whether we should do this for the ABCI results pruning as well?

For interest's sake, how did you find out that the indexer pruning was slowing down block pruning? And do we have evidence that block results pruning is slowing down block pruning?

state/pruner.go
```diff
@@ -0,0 +1,331 @@
...
+// GetApplicationRetainHeight method of the underlying state store.
+func (p *Pruner) GetApplicationRetainHeight() (int64, error) {
+	return p.stateStore.GetApplicationRetainHeight()
+}
+
+// GetCompanionBlockRetainHeight is a convenience method for accessing the
+// GetCompanionBlockRetainHeight method of the underlying state store.
+func (p *Pruner) GetCompanionBlockRetainHeight() (int64, error) {
+	return p.stateStore.GetCompanionBlockRetainHeight()
+}
```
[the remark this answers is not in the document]
mzabaluev: > if we want a consistent view on these values

OK, perhaps this is too much as we don't have to guarantee any consistent order of modifying these values between the ABCI app and the companion service.

thanethomson:
> We currently do not have any evidence that pruning ABCI results is slowing down the block pruning. But I can see the same thing happening for ABCI results if for example someone decides to prune them at a very high height and the process takes long. THe block pruning would have to wait for this to be done. Do you see a drawback in having a separate go routine for each of the different pruning options?

I mean, more concurrency :sweat_smile: But with the current design being as simple as it is it shouldn't be a problem. Feel free to separate out the results pruning into a separate goroutine!

state/pruner.go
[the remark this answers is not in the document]
thanethomson: Well, only the companion has control over the companion retain height. It's just the application retain height that could change in the background outside of the companion's control.

jmalicevic:
@thanethomson I pushed a commit with that. Separating this causes a bit of confusion with regards to the pruner so please check out the last commit.

state/pruner.go
```diff
@@ -205,30 +207,42 @@ func (p *Pruner) GetABCIResRetainHeight() (int64, error) {
...
+func (p *Pruner) pruneABCIResRoutine() {
+	p.logger.Info("Started pruning ABCI results", "interval", p.interval.String())
+	lastABCIResRetainHeight := int64(0)
+	for {
+		select {
+		case <-p.Quit():
+			return
+		default:
+			newABCIResRetainHeight := p.pruneABCIResToRetainHeight(lastABCIResRetainHeight)
+			p.observer.PrunerPruned(&PrunedInfo{
```
thanethomson, line 219: The observer's `PrunerPruned` method would need to be split into two methods:
- `PrunerPrunedBlocks(*BlocksPrunedInfo)`
- `PrunerPrunedABCIResults(*ABCIResponsesPrunedInfo)`

If we split the pruning routines, it no longer makes sense to have a single event represent both.

jmalicevic: Done!

state/pruner.go
```diff
@@ -0,0 +1,341 @@
...
+				FromHeight: lastABCIResRetainHeight,
+				ToHeight:   newABCIResRetainHeight - 1,
+			})
+			lastABCIResRetainHeight = newABCIResRetainHeight
+			time.Sleep(p.interval)
+		}
+	}
+}
+func (p *Pruner) pruneBlocksRoutine() {
+	p.logger.Info("Pruner started", "interval", p.interval.String())
```
thanethomson, line 229:
```suggestion
	p.logger.Info("Started pruning blocks", "interval", p.interval.String())
```

state/pruner.go
```diff
@@ -0,0 +1,341 @@
...
+
+	// newRetainHeight is the height just after that which we have successfully
+	// pruned. In case of an error it will be 0, but then it will also be
+	// ignored.
+	numPruned, newRetainHeight, err := p.stateStore.PruneABCIResponses(targetRetainHeight)
+	if err != nil {
+		p.logger.Error("Failed to prune ABCI responses", "err", err, "targetRetainHeight", targetRetainHeight)
+		return lastRetainHeight
+	}
+	p.logger.Info("Pruned ABCI responses", "height", numPruned)
```
thanethomson, line 288:
```suggestion
	if numPruned > 0 {
		p.logger.Info("Pruned ABCI responses", "heights", numPruned, "newRetainHeight", newRetainHeight)
	}
```

So as to not spam operators.

thanethomson (approved):
Just a few minor nits, but otherwise this LGTM!

sergio-mena (commented):
As discussed yesterday, I took a look at this PR, even if it's already merged.

state/pruner.go
```diff
@@ -0,0 +1,343 @@
...
+		return lastRetainHeight
+	}
+	pruned, evRetainHeight, err := p.pruneBlocks(targetRetainHeight)
+	// The new retain height is the current lowest point of the block store
+	// indicated by Base()
+	newRetainHeight := p.bs.Base()
+	if err != nil {
+		p.logger.Error("Failed to prune blocks", "err", err, "targetRetainHeight", targetRetainHeight, "newRetainHeight", newRetainHeight)
+	} else if pruned > 0 {
+		p.logger.Info("Pruned blocks", "count", pruned, "evidenceRetainHeight", evRetainHeight, "newRetainHeight", newRetainHeight)
```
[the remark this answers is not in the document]
thanethomson: I suppose that, if we're exposing metrics for operators, we could downgrade this to `Debug` level. As long as there's some way of operators knowing whether it's actually working.

state/store.go
```diff
@@ -366,6 +389,51 @@ func (store dbStore) PruneStates(from int64, to int64, evidenceThresholdHeight i
...
+	for h := lastRetainHeight; h < targetRetainHeight; h++ {
+		if err := batch.Delete(calcABCIResponsesKey(h)); err != nil {
+			return pruned, lastRetainHeight + pruned, fmt.Errorf("failed to delete ABCI responses at height %d: %w", h, err)
+		}
+		batchPruned++
+		if batchPruned%1000 == 0 && batchPruned > 0 {
+			if err := batch.Write(); err != nil {
+				return pruned, lastRetainHeight + pruned, fmt.Errorf("failed to write ABCI responses deletion batch at height %d: %w", h, err)
+			}
+			batch.Close()
```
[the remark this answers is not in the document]
thanethomson: Yes. Technically it'll return an error if called twice, but it doesn't matter.

state/pruner.go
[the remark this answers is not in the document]
sergio-mena: > way of operators knowing whether it's actually working

Then maybe we could just Info-log every N (e.g., N=500) blocks to reduce verbosity, the rest being Debug-log. I'm convinced that the current verbosity will add noise to the Info logs.

I think we're using (something similar to) this technique elsewhere in this PR.

config/config.go
```diff
@@ -1167,13 +1172,16 @@ type StorageConfig struct {
 	// required for `/block_results` RPC queries, and to reindex events in the
 	// command-line tool.
 	DiscardABCIResponses bool `mapstructure:"discard_abci_responses"`
```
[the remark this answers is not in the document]
jmalicevic: It can still make sense, at least for a first release of this API. To indicate pruning of ABCI results atm, an operator has to set up a data companion to be attached to the API. And if they do not want to do it at the moment, they can leverage this config flag and avoid storing data they do not want.

config/config.go
```diff
@@ -1303,3 +1319,89 @@ func getDefaultMoniker() string {
...
+//-----------------------------------------------------------------------------
+// DataCompanionPruningConfig
+
+type DataCompanionPruningConfig struct {
+	// Whether automatic pruning respects values set by the data companion.
+	// Disabled by default. All other parameters in this section are ignored
+	// when this is disabled.
+	//
+	// If disabled, only the application retain height will influence block
+	// pruning (but not block results pruning). Only enabling this at a later
```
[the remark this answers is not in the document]
jmalicevic: If disabled, there is nothing to influence ABCI results pruning other than the `DiscardABCIResults` config parameter. At this point, the application has no way to indicate it wants to prune the ABCI results. For that we'd have to alter the response in `Commit` to include this retain height as well.

state/pruner.go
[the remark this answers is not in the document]
jmalicevic: We have metrics now so I think it's all good. Unless you think there is  still a point in logging this ever X blocks in case operators do not use prometheus to monitor the node.

state/pruner.go
```diff
@@ -0,0 +1,343 @@
...
+		}
+	}
+}
+
+func (p *Pruner) pruneBlocksToRetainHeight(lastRetainHeight int64) int64 {
+	targetRetainHeight := p.findMinRetainHeight()
+	if targetRetainHeight == lastRetainHeight {
+		return lastRetainHeight
+	}
+	pruned, evRetainHeight, err := p.pruneBlocks(targetRetainHeight)
```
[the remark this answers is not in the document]
jmalicevic: Hmm, the way things are now we are not using it. We could technically keep it and verify that the requested retain height is not below it. However,  I don't think that makes much sense if we keep the `newRetainHeight`. It won't tell us anything new and it makes no sense to compare agains an olg `evidenceRetainHeight` as it is calculated dynamically based on the state of the blockchain at the moment of pruning.

state/pruner.go
```diff
@@ -0,0 +1,343 @@
...
+	newRetainHeight := p.bs.Base()
+	if err != nil {
+		p.logger.Error("Failed to prune blocks", "err", err, "targetRetainHeight", targetRetainHeight, "newRetainHeight", newRetainHeight)
+	} else if pruned > 0 {
+		p.logger.Info("Pruned blocks", "count", pruned, "evidenceRetainHeight", evRetainHeight, "newRetainHeight", newRetainHeight)
+	}
+	return newRetainHeight
+}
+
+func (p *Pruner) pruneABCIResToRetainHeight(lastRetainHeight int64) int64 {
```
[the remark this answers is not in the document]
jmalicevic: It should be ok even if it was changed between then.. because changing it can only happen if the new height is higher than the previous one, which means we will still prune the previously requested results. Same for block results. 

For example the retain height is set to 5. 
We enter `pruneABCIResToRetaineHeight` at the same time someone tries to write 3. That someone won't be allowed to write 3 due to the check in 186.  If they try to set it to 6, and we read it just after that,it's still ok.

state/store.go
```diff
@@ -366,6 +389,51 @@ func (store dbStore) PruneStates(from int64, to int64, evidenceThresholdHeight i
 	return nil
 }
 
+// PruneABCIResponses attempts to prune all ABCI responses up to, but not
+// including, the given height. On success, returns the number of heights
+// pruned and the new retain height.
+func (store dbStore) PruneABCIResponses(targetRetainHeight int64) (int64, int64, error) {
+	if store.DiscardABCIResponses {
+		return 0, 0, errors.New("ABCI responses are discarded, nothing to prune")
```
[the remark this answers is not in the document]
jmalicevic: Hmm you might be right. The pruner will always end up here if a data companion tries to set a retain height for ABCI results. So we should just return 0,0,nil with a Debug message even. Info would still be too much. The operator can see with the metrics that nothing is pruned (that nothing even exists).

state/store.go
```diff
@@ -366,6 +389,51 @@ func (store dbStore) PruneStates(from int64, to int64, evidenceThresholdHeight i
...
+
+	pruned := int64(0)
+	batchPruned := int64(0)
+
+	for h := lastRetainHeight; h < targetRetainHeight; h++ {
+		if err := batch.Delete(calcABCIResponsesKey(h)); err != nil {
+			return pruned, lastRetainHeight + pruned, fmt.Errorf("failed to delete ABCI responses at height %d: %w", h, err)
+		}
+		batchPruned++
+		if batchPruned%1000 == 0 && batchPruned > 0 {
```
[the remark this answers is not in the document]
jmalicevic: Hmm, probably nothing, but it would not make sense to use `>=`, `==` would have the same effect as `%`

config/config.go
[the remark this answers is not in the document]
sergio-mena: I see. Makes sense.
Have you folks thought about the possibility of using this pruning mechanism with a noop companion? Think of validators, that would a good fit wouldn't it? Anyway, this is beyond the scope of this PR. But just something to think about.

config/config.go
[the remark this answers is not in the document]
sergio-mena: Thanks for explaining. I still think that, with the new mechanism, ABCI results should be pruned together with their corresponding block, but that's just my opinion. And probably doing sth like that in this next release is not the best moment. Better to give users the time to use (and get used to) a companion.

state/pruner.go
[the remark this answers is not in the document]
sergio-mena: With metrics, I think turning this into a Debug should be good enough

state/store.go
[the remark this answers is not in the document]
sergio-mena: `>='  or `==` are both good to me (I tend to use the former as it is more "defensive")

The point of my question was that the current if is not very "readable"... it makes you think that if will periodically be activated, whereas it is not the case (BTW, this is an "idiomatic" comment 😄)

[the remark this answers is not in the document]
sergio-mena: Yes, you're right. The choice of `>=` (defensive) is in case there's a bug we didn't catch inside the loop that would (e.g, skip the value where `==` becomes true, effectively turning it into an infinite loop).
But this is super-minor, you can keep the `==`, no problem with that.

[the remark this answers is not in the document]
jmalicevic: Hah that is a new level of defensive :D Thanks for sharing that, no worries, I can change it to `>=` :D

PR #1152, opened 2023-07-21
Store: batching writes to the block and state stores

### LINKED ISSUE #1040: Batch commits to the state and block store
Currently, state is committed to the block and state store using asynchronous calls followed by a synchronous call at the end. 

We need to investigate whether explicitly batching them improves performance. In particular, consider the work done in :
https://github.com/tendermint/tendermint/pull/6067
https://github.com/tendermint/tendermint/pull/6018

and issue: https://github.com/tendermint/tendermint/issues/5888 

The goal of this is two fold:
 - Increase atomicity guarantees
 - Improve performance
DoD:
* [ ] Understand, from a DB perspective whether atomicity is actually increased or not
* [ ] Evaluate whether there are corner cases where the batch becomes very big and understand how the DB handles this underneath (memory vs. disk)
* [x] State is saved in batches with an indication on performance improvement

werty144:
Closes #1040

---

#### PR checklist

- [x] Tests written/updated


state/store.go
```diff
@@ -598,7 +620,8 @@ func (store dbStore) saveValidatorsInfo(height, lastHeightChanged int64, valSet
 		return err
 	}
 
-	err = store.db.Set(calcValidatorsKey(height), bz)
+	//err = store.db.Set(calcValidatorsKey(height), bz)
```
adizere, line 623: Should we delete this comment ?

state/store.go
```diff
@@ -675,7 +698,8 @@ func (store dbStore) saveConsensusParamsInfo(nextHeight, changeHeight int64, par
 		return err
 	}
 
-	err = store.db.Set(calcConsensusParamsKey(nextHeight), bz)
+	//err = store.db.Set(calcConsensusParamsKey(nextHeight), bz)
```
adizere, line 701: Maybe delete this.

adizere (commented):
Great work Anton!

@thanethomson should we use `main` as the base branch for this PR? Unclear if there's a feature branch with storage improvements or not.

@werty144 Shall we also add a table with performance results so we can understand the impact of this batching mechanism?

werty144:
I've also performed some measurements using `state/state_test.go/BenchmarkStateSave` function. 
Here are the results: https://docs.google.com/spreadsheets/d/1aZDcb4F6_ZCAqFm8E-JpBT5OwOZZPi-0gDzgsR5tmjg/edit?usp=sharing

- Batching gives a minor improvement for CLevel, Rocks, Badger and Bolt DBs.
- Batching is actually worse with GoLevelDB.
- Interestingly BoltDB is ~2 times worse than others

state/store.go
werty144: Done

state/store.go
werty144: Done

sergio-mena:
I quickly went through this and looks pretty good to me. I would say we can merge this to `main` with very little risk, unless the storage squad has decided on a different policy regarding feature branches.

store/store.go
```diff
@@ -393,12 +393,26 @@ func (bs *BlockStore) SaveBlock(block *types.Block, blockParts *types.PartSet, s
...
 	}
-	if err := bs.saveBlockToBatch(block, blockParts, seenCommit); err != nil {
+
+	batch := bs.db.NewBatch()
+	defer func(batch dbm.Batch) {
+		err := batch.Close()
+		if err != nil {
+			panic(err)
+		}
+	}(batch)
```
thanethomson, line 403: I would extract this as a function:

```golang
func closeBatchOrPanic(batch dbm.Batch) {
	if err := batch.Close(); err != nil {
		panic(err)
	}
}
```

Then you could simply:

```suggestion
	defer closeBatchOrPanic(batch)
```

Makes the enclosing function a little easier to read. Similarly for this same type of pattern in other packages.

thanethomson:
> I quickly went through this and looks pretty good to me. I would say we can merge this to `main` with very little risk, unless the storage squad has decided on a different policy regarding feature branches.

If the performance worsens with GoLevelDB (the most popular underlying DB for Comet atm, afaict) then I'm not sure we'd want to merge this.

I still find it a very useful exercise to have the data. Might be worthwhile to transcribe the sheet into a table that can be added into a comment on this PR for posterity.

sergio-mena:
> If the performance worsens with GoLevelDB (the most popular underlying DB for Comet atm, afaict) then I'm not sure we'd want to merge this.

I see, I have to believe the measurements as that's all we have, so yes, it's probably better to wait.

However, I'm surprised that simply batching DB transactions incurs an extra performance cost, as batching is done precisely to have to sync less often. Also, remember that the IG team had implemented batching in `v0.35.x`/`v0.36.x` although I don't know if they measured performance when they did it.

p0mvn:
This might be related to why goleveldb is having issues: https://github.com/cosmos/iavl/issues/619

TLDR; There is an optimal batch size. Having the batch be too large causes garbage collector overhead since GoLevelDB simply has a bytes slice where it stores data until commit.

jmalicevic:
Reopening this because we are still to decide whether to go with this after we benchmark it further.

melekes:
Superseded by https://github.com/cometbft/cometbft/pull/1735

PR #1153, opened 2023-07-21
Adding a test for when peers are late

lasarojc:
This PR adds some extra testing to the base one.

### DIFF
--- a/mempool/reactor.go
+++ b/mempool/reactor.go
@@ -164,18 +164,21 @@ func (memR *Reactor) broadcastTxRoutine(peer p2p.Peer) {
 
 		// Wait until either: a mempool entry is available in the channel, or
 		// the peer was disconnected, or the reactor stopped.
-		select {
-		case <-iter.WaitNext():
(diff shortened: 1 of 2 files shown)


mempool/reactor.go
```diff
@@ -213,6 +216,7 @@ func (memR *Reactor) broadcastTxRoutine(peer p2p.Peer) {
 				time.Sleep(PeerCatchupSleepIntervalMS * time.Millisecond)
 				continue
 			}
+			entry = nil
```
hvanz, line 219: Shouldn't this line be after the closing bracket of `if !memR.isSender(...`? Otherwise, if the peer is a sender, it will keep looping with the same entry, instead of fetching the next one.

hvanz (approved):
Good catch! Please check that the last line `entry = nil` is in the right posiiton.

I think now the test `TestReactorTxSendersMultiNode` could be improved by testing that the transactions were received in order, but we could fix it on the main PR branch.

PR #1154, opened 2023-07-22
ADR-101: implement gRPC `PruningService`

### LINKED ISSUE #1097: ADR-101: Implement gRPC `PruningService`
We need a gRPC API on the `feature/adr101-pull-companion` branch to provide integrators with access to the pruning functionality specified in ADR-101.

mzabaluev:
Closes #1097

---

#### PR checklist

- [x] Tests written/updated


config/toml.go
```diff
@@ -276,6 +276,31 @@ laddr = "{{ .GRPC.ListenAddress }}"
...
+laddr = "{{ .GRPC.Privileged.ListenAddress }}"
+
+#
+# Configuration specifically for the gRPC pruning service, which is considered a
+# privileged service.
+#
+[grpc.privileged.pruning_service]
+
+# Only controls whether the pruning service is accessible via the gRPC API - not
+# whether a previously set pruning service retain height is honoured by the
```
mzabaluev, line 294: @thanethomson Your English is not acceptable by the linter :stuck_out_tongue_closed_eyes:

rpc/grpc/server/services/pruningservice/service.go
```diff
@@ -0,0 +1,89 @@
...
+func (s *pruningServiceServer) SetBlockRetainHeight(_ context.Context, req *v1.SetBlockRetainHeightRequest) (*v1.SetBlockRetainHeightResponse, error) {
+	height := req.Height
+	// Because we can't agree on a single type to represent block height.
+	if height > uint64(math.MaxInt64) {
+		return nil, status.Errorf(codes.InvalidArgument, fmt.Sprintf("Invalid height %d", height))
+	}
+	logger := s.logger.With("endpoint", "SetBlockRetainHeight")
+	err := s.pruner.SetCompanionRetainHeight(int64(height))
+	if err != nil {
+		logger.Error("Cannot set block retain height", "err", err)
```
mzabaluev, line 39: I can add tracing to the logs like in #1142 (noticed this is not uniformly implemented for every endpoint).

thanethomson: It's only valuable for cases where we don't want to leak internal server error information to the caller.

But if there are such cases where that's missing then we'll need to add the tracing in.

rpc/grpc/server/services/pruningservice/service.go
```diff
@@ -0,0 +1,89 @@
...
+		pruner: pruner,
+		logger: logger.With("service", "PruningService"),
+	}
+}
+
+// SetBlockRetainHeight implements v1.PruningServiceServer.
+func (s *pruningServiceServer) SetBlockRetainHeight(_ context.Context, req *v1.SetBlockRetainHeightRequest) (*v1.SetBlockRetainHeightResponse, error) {
+	height := req.Height
+	// Because we can't agree on a single type to represent block height.
+	if height > uint64(math.MaxInt64) {
```
jmalicevic, line 33: Height should also never be < 0 .

rpc/grpc/server/services/pruningservice/service.go
```diff
@@ -0,0 +1,89 @@
...
+		PruningServiceRetainHeight: uint64(svcHeight),
+		AppRetainHeight:            uint64(appHeight),
+	}, nil
+}
+
+// SetBlockResultsRetainHeight implements v1.PruningServiceServer.
+func (s *pruningServiceServer) SetBlockResultsRetainHeight(_ context.Context, req *v1.SetBlockResultsRetainHeightRequest) (*v1.SetBlockResultsRetainHeightResponse, error) {
+	height := req.Height
+	// Because we can't agree on a single type to represent block height.
+	if height > uint64(math.MaxInt64) {
```
jmalicevic, line 68: height has to be > 0

rpc/grpc/server/services/pruningservice/service.go
mzabaluev: It is by virtue of its `uint64` encoding in protobuf.

config/toml.go
```diff
@@ -285,6 +285,31 @@ enabled = {{ .GRPC.BlockService.Enabled }}
...
+#
+[grpc.privileged.pruning_service]
+
+# Only controls whether the pruning service is accessible via the gRPC API - not
+# whether a previously set pruning service retain height is honored by the
+# node.
+#
+# To disable the influence of previously set pruning service retain height(s) on
+# node pruning, this endpoint should be enabled and the relevant pruning service
+# retain heights should be set to 0.
```
jmalicevic, line 308: I am not 100% sure I understand this comment. I guess what you want to say is that one should try to set the retain height using this gRPC endpoint if he wants a different height to be considered? Because if one enables the companion and tries to set the height to 0 when it was previously set to some thing >0 by either the app or another data companion, this operation will fail.

mzabaluev: Hmm, I just copied this from the ADR-101 document. @thanethomson can we make this more clear?

rpc/grpc/server/services/pruningservice/service.go
mzabaluev: Now let's be precise: you mean it can never be 0 (it can't be negative as commented above)?

jmalicevic: What I commented above but you already pointed me to the answer so I resolved this :) All good!

config/config.go
```diff
@@ -607,6 +613,52 @@ func TestGRPCBlockServiceConfig() *GRPCBlockServiceConfig {
 	}
 }
 
+//-----------------------------------------------------------------------------
+// GRPCPrivilegedConfig
+
+// GRPCPrivilegedConfig defines the configuration for the CometBFT gRPC server
+// exposing privileged endpoints.
+
```
thanethomson, line 621:
```suggestion
```

config/config.go
```diff
@@ -607,6 +613,52 @@ func TestGRPCBlockServiceConfig() *GRPCBlockServiceConfig {
...
+type GRPCPrivilegedConfig struct {
+	// TCP or Unix socket address for the gRPC server for privileged clients
+	// to listen on. If empty, the gRPC server will be disabled.
```
thanethomson, line 624:
```suggestion
	// to listen on. If empty, the privileged gRPC server will be disabled.
```

config/toml.go
thanethomson, line 308:
```suggestion
# node. See the [storage.pruning] section for control over pruning.
```

rpc/grpc/client/privileged/pruning_service.go
```diff
@@ -0,0 +1,97 @@
...
+// set for block information to be retained by the ABCI application and the
+// pruning service.
+type RetainHeights struct {
+	App            uint64
+	PruningService uint64
+}
+
+type PruningServiceClient interface {
+	SetBlockRetainHeight(ctx context.Context, height uint64) error
+	GetBlockRetainHeight(ctx context.Context) (RetainHeights, error)
```
thanethomson, line 21: Should we not have two separate functions here:

```suggestion
	GetBlockRetainHeight(ctx context.Context) (uint64, error)
	GetRetainHeights(ctx context.Context) (RetainHeights, error)
```

rpc/grpc/server/services/pruningservice/service.go
```diff
@@ -0,0 +1,89 @@
...
+// SetBlockRetainHeight implements v1.PruningServiceServer.
+func (s *pruningServiceServer) SetBlockRetainHeight(_ context.Context, req *v1.SetBlockRetainHeightRequest) (*v1.SetBlockRetainHeightResponse, error) {
+	height := req.Height
+	// Because we can't agree on a single type to represent block height.
+	if height > uint64(math.MaxInt64) {
+		return nil, status.Errorf(codes.InvalidArgument, fmt.Sprintf("Invalid height %d", height))
+	}
+	logger := s.logger.With("endpoint", "SetBlockRetainHeight")
+	err := s.pruner.SetCompanionRetainHeight(int64(height))
+	if err != nil {
```
thanethomson, line 38:
```suggestion
	if err := s.pruner.SetCompanionRetainHeight(int64(height)); err != nil {
```

rpc/grpc/server/services/pruningservice/service.go
```diff
@@ -0,0 +1,89 @@
...
+		logger.Error("Cannot set block retain height", "err", err)
+		return nil, status.Errorf(codes.Internal, "Failed to set block retain height")
```
thanethomson, line 40: For cases like this where we're swallowing the error from the client's perspective, it makes sense to use the tracing such that an operator can at least check why a specific request failed.

rpc/grpc/server/services/pruningservice/service.go
```diff
@@ -0,0 +1,89 @@
...
+	svcHeight, err := s.pruner.GetCompanionBlockRetainHeight()
+	if err != nil {
+		logger.Error("Cannot get block retain height stored by companion", "err", err)
+		return nil, status.Errorf(codes.Internal, "Failed to get companion block retain height")
+	}
+	appHeight, err := s.pruner.GetApplicationRetainHeight()
+	if err != nil {
+		logger.Error("Cannot get block retain height specified by application", "err", err)
+		return nil, status.Errorf(codes.Internal, "Failed to get app block retain height")
+	}
```
thanethomson, line 57: Same as my other comment re: error tracing.

rpc/grpc/server/services/pruningservice/service.go
```diff
@@ -0,0 +1,89 @@
...
+	height := req.Height
+	// Because we can't agree on a single type to represent block height.
+	if height > uint64(math.MaxInt64) {
+		return nil, status.Errorf(codes.InvalidArgument, fmt.Sprintf("Invalid height %d", height))
+	}
+	logger := s.logger.With("endpoint", "SetBlockResultsRetainHeight")
+	err := s.pruner.SetABCIResRetainHeight(int64(height))
+	if err != nil {
+		logger.Error("Cannot set block results retain height", "err", err)
+		return nil, status.Errorf(codes.Internal, "Failed to set block results retain height")
```
thanethomson, line 75: Same as my other comment re: error tracing.

rpc/grpc/server/services/pruningservice/service.go
```diff
@@ -0,0 +1,89 @@
...
+	return &v1.SetBlockResultsRetainHeightResponse{}, nil
+}
+
+// GetBlockResultsRetainHeight implements v1.PruningServiceServer.
+func (s *pruningServiceServer) GetBlockResultsRetainHeight(_ context.Context, _ *v1.GetBlockResultsRetainHeightRequest) (*v1.GetBlockResultsRetainHeightResponse, error) {
+	logger := s.logger.With("endpoint", "GetBlockResultsRetainHeight")
+	height, err := s.pruner.GetABCIResRetainHeight()
+	if err != nil {
+		logger.Error("Cannot get block results retain height", "err", err)
+		return nil, status.Errorf(codes.Internal, "Failed to get block results retain height")
```
thanethomson, line 86: Same as my other comment re: error tracing.

thanethomson (commented):
Looking great so far @mzabaluev, thanks! Just a couple of minor suggestions.

rpc/grpc/client/privileged/pruning_service.go
thanethomson: Ah yes, we want to return both here. Nvm!

thanethomson (approved):
Nice, LGTM! Thanks @mzabaluev!

rpc/grpc/client/privileged/privileged.go
```diff
@@ -0,0 +1,104 @@
...
+	builder := newClientBuilder()
+	for _, opt := range opts {
+		opt(builder)
+	}
+	conn, err := ggrpc.DialContext(ctx, addr, builder.grpcOpts...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to dial %s: %w", addr, err)
+	}
+	pruningServiceClient := newDisabledPruningServiceClient()
+	if builder.pruningServiceEnabled {
```
jmalicevic, line 96: Should we not have here an `if ... else` rather then setting it and immediately changing it if it is enabled?

rpc/grpc/client/privileged/pruning_service.go
```diff
@@ -0,0 +1,97 @@
...
+// GetBlockResultsRetainHeight implements PruningServiceClient.
+func (c *pruningServiceClient) GetBlockResultsRetainHeight(ctx context.Context) (uint64, error) {
+	res, err := c.inner.GetBlockResultsRetainHeight(ctx, &v1.GetBlockResultsRetainHeightRequest{})
+	if err != nil {
+		return 0, err
+	}
+	return res.PruningServiceRetainHeight, nil
+}
+
+type disabledPruningServiceClient struct{}
```
jmalicevic, line 73: Why do we need a different struct for a disabled pruning service client? Is it because of testing? Or because we have places where we need to pass on a client even though it is disabled?

config/toml.go
thanethomson: I've removed this from the ADR now. It's no longer relevant.

rpc/grpc/client/privileged/privileged.go
thanethomson: Either pattern's fine. This is just more compact. The alternative definition involves more lines of code:

```go
var pruningServiceClient PruningServiceClient
if builder.pruningServiceEnabled {
    pruningServiceClient = newPruningServiceClient(conn)
} else {
    pruningServiceClient = newDisabledPruningServiceClient()
}
```

The computational cost of constructing a disabled pruning service client is entirely negligible.

rpc/grpc/client/privileged/pruning_service.go
thanethomson: The assumption is that the privileged client may potentially be expanded over time and some functionality (like the pruning service client) may not be enabled. It's not very useful at the moment.

jmalicevic: Ah I see, thanks for clarifying!

rpc/grpc/client/privileged/pruning_service.go
```diff
@@ -0,0 +1,97 @@
...
+
+func newDisabledPruningServiceClient() PruningServiceClient {
+	return &disabledPruningServiceClient{}
+}
+
+// SetBlockRetainHeight implements PruningServiceClient.
+func (*disabledPruningServiceClient) SetBlockRetainHeight(context.Context, uint64) error {
+	panic("pruning service client is disabled")
```
thanethomson, line 81: I was thinking whether we should perhaps consider returning errors here as opposed to panicking, given how this may end up being used. Someone instantiating a client may end up wanting to, for whatever, reason, disable certain client facets that they give to their users, and a more graceful way of communicating to users that that facet's disabled is to return an error.

jmalicevic (approved):
LGTM, thanks Mikhail!

rpc/grpc/client/privileged/pruning_service.go
jmalicevic: It might make sense, with a defined error of `ErrServiceDisabled` that could be potentially reused for other services down the road.

mzabaluev: Good idea, as long as we want to have this disabled-on-the-client functionality at all. Let's rework all services in a follow-up PR.

PR #1156, opened 2023-07-24
Fix operator priority bug

dudong2:
The condition for marking `metrics.MarkRound` in the updateRoundStep function is when it has moved on to the `next round`. 
For `A || B && C`, I think the intended behavior is `(A || B) && C`. However, the actual behavior is `A || (B && C)`.
```
a := 1
b := 1
c := 1
if a == 1 || b == 0 && c == 0 {
	fmt.Println("bug!!")
} else {
	fmt.Println("ok")
}
```
In the code above, I expected `ok` to be printed, but `bug!!` was printed.
Even if the intended behavior is `A || (B && C)`, the original code is not readable and needs to be fixed.

---

#### PR checklist

- [ ] Tests written/updated


lasarojc:
I believe your analysis is correct, @adizere, and agree with both that adding the `()` will make the intent more explicit, even though the code is correct.

dudong2:
> Hi @dudong2, thanks for your contribution!
> 
> I suspect the original code is actually correct. But I'd like to double-check with a colleague that has more technical expertise than me to confirm. @cason any thoughts on this change?
> 
> In case my colleague confirms the original code is correct, it might still be worthwhile to explicitly add the `( )` to make it clear what is the intended behavior.
> 
> ### My current understanding
> From what I can tell, method `updateRoundStep` registers metrics concerning `round`s and `step`s, such as how much time a `round` took, or how long did the local validator spend during a specific `step`. There's 2 types of cases when this method is called:
> 
> 1. when everything goes fine (no failures nor network asynchrony), the method typically gets called in this sequence:
>    
>    * `updateRoundStep(round = 0 , step = cstypes.RoundStepNewHeight)` <- new block being produced
>    * `updateRoundStep(round = 0 , step = cstypes.RoundStepNewRound)`
>    * `updateRoundStep(round = 0 , step = cstypes.RoundStepPropose)`
>    * `updateRoundStep(round = 0 , step = cstypes.RoundStepPrevote)`
>    * `updateRoundStep(round = 0 , step = cstypes.RoundStepPrevoteWait)`
>    * `updateRoundStep(round = 0 , step = cstypes.RoundStepPrecommit)`
>    * `updateRoundStep(round = 0 , step = cstypes.RoundStepCommit)` <- block finishes
>    * `updateRoundStep(round = 0 , step = cstypes.RoundStepNewHeight)` <- new block being produced
>    * `updateRoundStep(round = 0 , step = cstypes.RoundStepNewRound)` ...
>    * ... with the same `round` parameter. This means that the validator is progressing at increasingly higher heights, and always restarting a new height at `round = 0`.
> 2. when a problem (eg proposer failure or network asynchrony causing timeouts) occurs, then this method is usually called in this pattern:
>    
>    * `updateRoundStep(round = 0 , step = cstypes.RoundStepNewRound)`
>    * ...
>    * `updateRoundStep(round = 1 , step = cstypes.RoundStepNewRound)`
>    * ..
> 
> The `if` branch guarded by `round != cs.Round` is triggered when we are in case 2. And I think the `if` branch guarded by `(round == 0 && step == cstypes.RoundStepNewRound)` covers case 1. Curious if I'm missing anything here!

Hi @adizere.
As I understand it, if a round is a change, we call `MarkRound` to record in the metrics when the round was actually changed. The part you mention is recording when `step` becomes a change, and this is done by calling `MarkStep`, so I don't think the original code is the intended behavior. If I'm wrong, please let me know! If the original code is correct, I will modify the parentheses so that the original intent is clearly expressed.

lasarojc:
I am closing this PR for now and on a second thought I don't think we should add the parentesis.
Please let us know if you still think we should invert the evaluation order and we can resume the discussion.

Issue #1055, opened 2023-06-29
ADR 108: Internal messaging simplification


thanethomson:
At present, we have 4 or 5 different mechanisms by which we facilitate communication between different parts of the system (e.g. between the reactors). This results in:
1. Difficulty in testing those components independently
2. Difficulty in reasoning about the operation of different parts of the system
3. Non-deterministic tests that are, because they are non-deterministic, of dubious value; and they slow down development effort

We should converge on a single, much simpler approach to communication between different parts of the system that need to operate concurrently and communicate with each other.

- [ ] Define new internal messaging approach in ADR
- [ ] Specify next steps after ADR has been developed

PR #1164, opened 2023-07-24
RFC 107: Internal signalling using event observers

thanethomson:
Relates to #1055.

:book: [Rendered](https://github.com/cometbft/cometbft/blob/thane/1055-rfc-event-observer/docs/rfc/rfc-107-event-observer.md)

---

#### PR checklist

- [ ] Tests written/updated


docs/rfc/rfc-105-event-observer.md
```diff
@@ -0,0 +1,195 @@
...
+        obs.ConsensusStateNewRoundStep(ev)
+    }
+}
+
+// ...
+```
+
+### Testing
+
+Many tests in the CometBFT codebase rely heavily on subscribing to specific
```
adizere, line 143: The RFC covers very well the _publisher_ side of coupled components, but not so much the _subscriber_ side, apart from the discussion here about testing. 

Looking through the codebase, I see these two types of subscriptions possible:
- [Subscribe](https://github.com/cometbft/cometbft/blob/v0.38.x/types/event_bus.go#L75)
- [SubscribeUnbuffered](https://github.com/cometbft/cometbft/blob/v0.38.x/types/event_bus.go#L86)

So I have two remarks/suggestions:

1. It might help to clarify if both possible of subscriptions are possible with event observers (I understood they are).
2. [Nit] It seems like the heaviest users of subscription are test modules: more than half of references I found to the above subscribe methods originate from _test.go files, which was surprising to me. It seems useful to clarify that part of the motivation of this change is to simplify testing (which we do at the end of the RFC but not as motivation at the outset).

adizere (commented):
Great writeup Thane, thank you, seems like a step in the right direction!

docs/rfc/rfc-105-event-observer.md
thanethomson: I've updated the RFC. The two subscribe mechanisms are pretty much the same - the only difference is the channel capacity.

lasarojc:
Thanks for starting the discussion, Thane.
It is not clear to me why it would be useful to have single observer instances and fan-out observers and not just the latter. Or was the single observer just an example?

thanethomson:
> Thanks for starting the discussion, Thane.
> 
> It is not clear to me why it would be useful to have single observer instances and fan-out observers and not just the latter. Or was the single observer just an example?

Both are examples. The important things are the pattern employed by the interface and the places from which that interface's methods are called.

The implementations of observers are use case-specific, and consensus engine developers (us or other teams) are free to choose how to implement that interface.

thanethomson:
See #1183 for a trivial example, including usage from a test.

docs/rfc/rfc-105-event-observer.md
```diff
@@ -0,0 +1,240 @@
...
+// caller or the callee to decide how to handle the concurrency of certain
+// events. The event bus approach, by contrast, is always concurrent and relies
+// on Go channels, which could end up filling up and causing back-pressure into
+// the caller (already observed in slow WebSocket subscribers).
+type StateObserver interface {
+    ConsensusStateNewRoundStep(ev EventDataRoundState)
+    ConsensusStateTimeoutPropose(ev EventDataRoundState)
+    ConsensusStateTimeoutWait(ev EventDataRoundState)
+    // ...
+}
```
sergio-mena, line 91: If we call it `ConsensusStateObserver`, and remove the prefix from the methods, would it cause name clash with other `XxxObserver`, `YyyObersver`, interfaces?

Actually, since the interface is part of the Consensus package, we wouldn't even need the `Consensus...`, since every time it's used outside the package, we'd need to prefix with the package name.

docs/rfc/rfc-105-event-observer.md
```diff
@@ -0,0 +1,240 @@
...
+following:
+
+```diff
+-        if err := cs.eventBus.PublishEventNewRoundStep(rs); err != nil {
+-            cs.Logger.Error("failed publishing new round step", "err", err)
+-        }
++        // Notify the observer
++        cs.observer.ConsensusStateNewRoundStep(rs)
+```
+
```
sergio-mena, line 115: I'm not sure where, but I would add somewhere in the text that this pattern will also be able to do away with cumbersome interfaces like [SwitchToConsensus](https://github.com/cometbft/cometbft/blob/3c88a5ec/blocksync/reactor.go#L35) and its awkward (and ugly) [callsite](https://github.com/cometbft/cometbft/blob/3c88a5ec/blocksync/reactor.go#L383-L385)

https://github.com/cometbft/cometbft/blob/3c88a5ec0b12f28c1646268ba39950f2a7cc4013/blocksync/reactor.go#L385

docs/rfc/rfc-105-event-observer.md
```diff
@@ -0,0 +1,240 @@
...
+- [`Subscribe`]
+- [`SubscribeUnbuffered`]
+
+The observer pattern proposed in this RFC can facilitate both patterns and more,
+given that one has the option of intervening synchronously in a blocking way
+when the "publisher" calls the observer event handler. Using the proposed
+observer pattern, it is up to either the publisher or the observer to implement
+their own concurrency.
+
```
sergio-mena, line 129: Another great benefit of this approach with respect to a go-channel-based one is the following.
In the near future, we may want to "standardise" reactor boundaries to improve the modularity of the system... allowing e.g., reactors in different processes, or even different programming languages for some reactors. The fact that the reactors interact via go channels, would complicate that effort, as go channels are particular to go and hard to handle in e.g., Protobufs, FFI, etc.

I think we can also include this argument in favor of this approach.

docs/rfc/rfc-105-event-observer.md
```diff
@@ -0,0 +1,240 @@
...
+- Greater compile-time correctness guarantees
+- Code becomes easier to reason about, since one can easily follow the call
+  chain for certain events using one's IDE instead of needing to search the
+  codebase for subscriptions to certain events
+- Easier to test
+- Does away with needing to access internal/private `eventBus` variables within
+  reactors/state from tests ([example][test-eventbus-access])
+- Splits event generation and handling out into a per-package responsibility,
+  more cleanly separating and modularizing the codebase
+
```
sergio-mena, line 223: A "conclusion" bullet here would be the KISS principle. When a simpler solution fits your needs, if you go for the complex one you're over-engineering .

sergio-mena (approved):
I like this proposal. This is (I think) a step in the right direction to disentangle the (let's call it) suboptimal way in which the components of the system interact today.

There is another big thing we need to tackle modularity-wise: the key-value store present in `Switch`. My question is: is the pattern proposed here also able to do away with that key-value store? (I haven't given it a lot of thought TBH)

docs/rfc/rfc-105-event-observer.md
thanethomson: This was a conscious decision on my part to name the functions as such - perhaps I should elaborate on that in the RFC. This allows for observer composition. For instance, if we have two observer interfaces, the `StateObserver` one and another interface:

```go
type OtherObserver interface {
    OtherTimeoutWait(params OtherTimeoutParams)
}
```

We could imagine an observer that implements both interfaces because it's interested in both events:

```go
type CombinedObserver interface {
    StateObserver
    OtherObserver
}
```

If we strip the `ConsensusState` and `Other` prefixes from the function names, the two would clash and there would be no way for anyone who doesn't control the interface definitions (e.g. external consensus engine developers) to implement a single observer for both sets of events.

Therefore we need to make sure that all observers' function names are globally unique throughout our codebase.

mzabaluev:
RFC 105 has been taken for something else, and RFC 106 is pending in #1533.

melekes (approved):
I think this proposal is sound. Thanks @thanethomson 👍

Issue #731, opened 2023-04-19
proto: Investigate migrating to Google's Go code generator


thanethomson:
At present we make use of https://github.com/cosmos/gogoproto to generate our Go code from our Protobuf definitions. Even though the project from which it was forked (https://github.com/gogo/protobuf) has been deprecated some time ago, and the same for the project from which gogoproto was forked (https://github.com/golang/protobuf), the Cosmos gogoproto fork is maintained by the SDK team.

It would, however, be valuable to investigate the impact of migrating to Google's official Go code generator (https://pkg.go.dev/google.golang.org/protobuf). See https://go.dev/blog/protobuf-apiv2 for context.

See also https://github.com/cosmos/cosmos-sdk/issues/7488 for the SDK's deliberations on this front - it provides valuable context.

## Definition of Done

When we have a feature branch that demonstrates the pragmatic impact of migrating to Google's Go code generator. We especially need to quantify the impact on:
1. Our Go APIs
2. Our RPC and JSON serialization
3. Which user(s) this affects

Once we have quantified this impact, and if the impact is manageable, follow-up work should involve coordinating the release of this change - especially with the Cosmos SDK team.

thanethomson:
I've spent a little time investigating what this entails here, and it appears as though gogoproto is, unfortunately, _very_ deeply embedded throughout the codebase:

- Many parts of the codebase rely on functions generated specifically by gogoproto (comparators, equality operations, etc.). We would need to manually re-implement this functionality across many different types. For example: https://github.com/cometbft/cometbft/blob/feac56b6297eb0dd800d5e35b9c04a848758c612/proto/tendermint/types/params.proto#L9
- In cases where we have message fields whose types are other messages, the gogoproto annotations control whether the generated code refers to a _value_ or a _pointer_. Google's Go code generator actually produces structs that do not allow themselves to be copied - on purpose. This, imho, is the correct way of making use of these structs. But blindly changing our usage here by value to pointer may introduce new race conditions, and re-evaluating all the usages of these structs may take quite some time. But maybe the race detector will help us here?
- There are a handful of places where we explicitly make use of gogo types in marshalling/unmarshalling data. We would need to re-implement these types manually. For example: https://github.com/cometbft/cometbft/blob/feac56b6297eb0dd800d5e35b9c04a848758c612/evidence/pool.go#L340

thanethomson:
Marking the spike/investigation as done here. If we want to actually do the migration, we should open a separate issue.

PR #1168, opened 2023-07-26
Add BlockResults gRPC service

### LINKED ISSUE #1095: ADR-101: Implement gRPC `BlockResultsService`
We need an implementation of the `BlockResultsService` from ADR-101 on the `feature/adr101-pull-companion` branch.

glnro:
Closes #1095 

This PR implements the gRPC service for BlockResults as specified in the ADR-101 Data Companion API
- `GetBlockResults` gRPC endpoint implemented
- Defined protos for BlockResults service
- Client implemented to interface with BlockResults `GetBlockResults` endpoint
- e2e Test
- Changelog entry
---

#### PR checklist

- [x] Tests written/updated


proto/tendermint/services/block_results/v1/block_results.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+package tendermint.services.block_results.v1;
+
+import "tendermint/abci/types.proto";
+import "tendermint/types/params.proto";
+import "gogoproto/gogo.proto";
```
andynog, line 6: We should not use gogoproto, got this feedback from Thane

rpc/grpc/client/block_results_service.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+type blockResultServiceClient struct {
+	client v1.BlockResultsServiceClient
+}
+
+func (b blockResultServiceClient) GetBlockResults(ctx context.Context, req v1.GetBlockResultsRequest) (*ResultBlockResults, error) {
+	res, err := b.client.GetBlockResults(ctx, &v1.GetBlockResultsRequest{Height: req.Height})
+	if err != nil {
+		return &ResultBlockResults{}, err
```
andynog, line 35: It would be nice to have some logging and return a more friendly message to the client instead of bubbling up the error from the server. Please check this logic in the Block service client

rpc/grpc/server/server.go
```diff
@@ -5,21 +5,29 @@ import (
 	"net"
 	"strings"
 
+	sm "github.com/cometbft/cometbft/state"
+	"github.com/cometbft/cometbft/store"
+
+	v1 "github.com/cometbft/cometbft/proto/tendermint/services/block_results/v1"
```
andynog, line 11: Better call this something else instead of 'v1' because there will be other services, maybe 'pbblkrsltsvc' or something like that

rpc/grpc/server/services/blockresultservice/service.go
```diff
@@ -0,0 +1,48 @@
...
+	latestHeight := s.blockStore.Height()
+	if req.Height > latestHeight || req.Height < 0 {
+		return &v1.GetBlockResultsResponse{}, status.Error(codes.InvalidArgument, "Height is invalid.")
+	} else if req.Height == 0 {
+		height = latestHeight
+	}
+
+	res, err := s.stateStore.LoadFinalizeBlockResponse(height)
+	if err != nil {
+		return &v1.GetBlockResultsResponse{}, err
```
andynog, line 38: As commented in the client logic this internal error might bubble up to the client, don't need to change here but just pointing it out based on comment in the client logic

test/e2e/tests/grpc_test.go
```diff
@@ -30,3 +33,64 @@ func TestGRPC_Version(t *testing.T) {
...
+				res, err := gRPCClient.GetBlockResults(ctx, tc.request)
+				// First block tests
+				require.NoError(t, err)
+				require.NotNil(t, res)
+				require.Equal(t, res.Height, tc.expectedHeight)
+			}
+			for _, tc := range errorCases {
+				_, err = gRPCClient.GetBlockResults(ctx, tc.request)
+				require.Error(t, err)
+			}
```
andynog, line 93: Maybe add a test for response to be nil ?

andynog (changes requested):

test/e2e/tests/grpc_test.go
```diff
@@ -30,3 +33,64 @@ func TestGRPC_Version(t *testing.T) {
...
+		gRPCClient, err := node.GRPCClient(ctx)
+		require.NoError(t, err)
+
+		for _, block := range blocks {
+			successCases := []struct {
+				expectedHeight int64
+				request        v1.GetBlockResultsRequest
+			}{
+				{first, v1.GetBlockResultsRequest{Height: first}},
+				{last, v1.GetBlockResultsRequest{}},
```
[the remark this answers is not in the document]
glnro: In the endpoint definition it states that if no height is provided then the endpoint will fetch the latest block. The e2e test is indeed failing for the CI network setup (passes on simple) because of a race condition with the test setup. The test fails because the endpoint is fetching the latest endpoint which is greater than the value its being compared to (cached last block height).

proto/tendermint/services/block_results/v1/block_results.proto
thanethomson: Yup. We should avoid it at all costs to better facilitate a future transition away from it, as per #731.

rpc/grpc/server/services/blockresultservice/service.go
```diff
@@ -0,0 +1,61 @@
...
+
+type blockResultsService struct {
+	stateStore sm.Store
+	blockStore *store.BlockStore
+	logger     log.Logger
+}
+
+// New creates a new CometBFT block results service server.
+func New(bs *store.BlockStore, ss sm.Store, logger log.Logger) brs.BlockResultsServiceServer {
+	return &blockResultsService{stateStore: ss, blockStore: bs, logger: logger}
```
thanethomson, line 24:
```suggestion
	return &blockResultsService{
		stateStore: ss,
		blockStore: bs,
		logger: logger.With("service", "BlockResultsService"),
	}
```

Helps with logging context.

rpc/grpc/server/services/blockresultservice/service.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// GetBlockResults returns the block results of the requested height.
+// If no height is given, the block results for the latest height are returned.
+func (s *blockResultsService) GetBlockResults(_ context.Context, req *brs.GetBlockResultsRequest) (*brs.GetBlockResultsResponse, error) {
+	logger := s.logger.With("endpoint", "GetBlockResults")
+	height := req.Height
+	latestHeight := s.blockStore.Height()
+	if req.Height > latestHeight || req.Height < 0 {
+		logger.Error("Failed to validate height from BlockResults request", "height", height)
```
thanethomson, line 34:
```suggestion
```

No need for error logs in "bad request"-type responses. This might flood operators.

rpc/grpc/server/services/blockresultservice/service.go
```diff
@@ -0,0 +1,61 @@
...
+		return nil, status.Error(codes.InvalidArgument, "Invalid request: height must be between 0 and the last effective height.")
```
thanethomson, line 35:
```suggestion
		return nil, status.Errorf(codes.InvalidArgument, "Height must be between 0 and the last effective height (%d)", latestHeight)
```

Might be useful to instruct the caller as to what the latest height is at the time of the request. Also, the `Invalid request` part of the message isn't really necessary given the error code we're sending back.

rpc/grpc/server/services/blockresultservice/service.go
```diff
@@ -0,0 +1,61 @@
...
+	} else if req.Height == 0 {
+		height = latestHeight
+	}
+
+	res, err := s.stateStore.LoadFinalizeBlockResponse(height)
+	if err != nil {
+		logger.Error("Error fetching BlockResults", "height", height, "err", err)
+		return nil, status.Errorf(codes.Internal, "Internal server error: cannot fetch BlockResults for height %d", height)
```
thanethomson, line 43:
```suggestion
		return nil, status.Error(codes.Internal, "Internal server error")
```

Since the error and the height are being logged, operators can dig into what's going on here. As a stretch goal, I'd recommend looking into using this request tracing approach I've implemented here: https://github.com/cometbft/cometbft/pull/1142/files#diff-2cdfb60e6f68e99a22b64a5d79cb26e00adb76e670b0f6d144b7912bf1ae2354R84

That way, if a user asks the operator about why they're getting internal server errors, the operator will have logs to link their specific requests to.

thanethomson (commented):
Seems like the E2E test's failing at the moment?

rpc/grpc/server/services/blockresultservice/service.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+}
+
+// GetBlockResults returns the block results of the requested height.
+// If no height is given, the block results for the latest height are returned.
+func (s *blockResultsService) GetBlockResults(_ context.Context, req *brs.GetBlockResultsRequest) (*brs.GetBlockResultsResponse, error) {
+	logger := s.logger.With("endpoint", "GetBlockResults")
+	height := req.Height
+	latestHeight := s.blockStore.Height()
+	if req.Height > latestHeight || req.Height < 0 {
```
jmalicevic, line 37: As we now support separate pruning of the blockstore and the abci results , it can happen that we don't have blocks but we do have the abci results. What would make more sense maybe is try and fetch the results and if there is no error return that.

rpc/grpc/client/block_results_service.go
```diff
@@ -0,0 +1,64 @@
...
+	TxsResults            []*abci.ExecTxResult      `json:"txs_results"`
+	FinalizeBlockEvents   []*abci.Event             `json:"finalize_block_events"`
+	ValidatorUpdates      []*abci.ValidatorUpdate   `json:"validator_updates"`
+	ConsensusParamUpdates *cmtproto.ConsensusParams `json:"consensus_param_updates"`
+	AppHash               []byte                    `json:"app_hash"`
+}
+
+// BlockResultsServiceClient provides the block results of a given height (or latest if none provided).
+type BlockResultsServiceClient interface {
+	GetBlockResults(ctx context.Context, req brs.GetBlockResultsRequest) (*ResultBlockResults, error)
```
thanethomson, line 26: This should just be:
```suggestion
	GetBlockResults(ctx context.Context, height int64) (*ResultBlockResults, error)
```

rpc/grpc/client/block_results_service.go
```diff
@@ -0,0 +1,64 @@
...
+
+	abci "github.com/cometbft/cometbft/abci/types"
+	cmtproto "github.com/cometbft/cometbft/proto/tendermint/types"
+
+	"github.com/cosmos/gogoproto/grpc"
+
+	brs "github.com/cometbft/cometbft/proto/tendermint/services/block_results/v1"
+)
+
+type ResultBlockResults struct {
```
thanethomson, line 15: Does it not make sense to call this something like `BlockResults`? Seems simpler.

rpc/grpc/server/server.go
```diff
@@ -5,21 +5,29 @@ import (
...
 type Option func(*serverBuilder)
 
 type serverBuilder struct {
-	listener       net.Listener
-	versionService pbversionsvc.VersionServiceServer
-	logger         log.Logger
-	grpcOpts       []grpc.ServerOption
+	listener            net.Listener
+	versionService      pbversionsvc.VersionServiceServer
+	blockresultsService brs.BlockResultsServiceServer
```
thanethomson, line 28: `blockResultsService`?

rpc/grpc/client/block_results_service.go
```diff
@@ -0,0 +1,68 @@
...
+	TxsResults            []*abci.ExecTxResult      `json:"txs_results"`
+	FinalizeBlockEvents   []*abci.Event             `json:"finalize_block_events"`
+	ValidatorUpdates      []*abci.ValidatorUpdate   `json:"validator_updates"`
+	ConsensusParamUpdates *cmtproto.ConsensusParams `json:"consensus_param_updates"`
+	AppHash               []byte                    `json:"app_hash"`
+}
+
+// BlockResultsServiceClient provides the block results of a given height (or latest if none provided).
+type BlockResultsServiceClient interface {
+	GetBlockResults(ctx context.Context, height *int64) (*BlockResults, error)
```
thanethomson, line 26: Would it not be better to have `height` be of type `int64`, similar to the block service's client interface?

Using `int64` seems more ergonomic than having to supply pointers to this function.

glnro: See [the latest commit](https://github.com/cometbft/cometbft/pull/1168/commits/14372cb1c560f77d96f4da3cd73cdacd86c359d3) with an update to provide an additional endpoint specifically to retrieve `BlockResults` of the latest committed height.

rpc/grpc/server/services/blockresultservice/service.go
```diff
@@ -0,0 +1,89 @@
...
+// GetBlockResults returns the block results of the requested height.
+func (s *blockResultsService) GetBlockResults(_ context.Context, req *brs.GetBlockResultsRequest) (*brs.GetBlockResultsResponse, error) {
+	logger := s.logger.With("endpoint", "GetBlockResults")
+	ss, err := s.stateStore.Load()
+	if err != nil {
+		logger.Error("Error loading store", "err", err)
+		return nil, status.Error(codes.Internal, "Internal server error")
+	}
+	if req.Height > ss.LastBlockHeight || req.Height < 0 {
+		logger.Error("Error validating GetBlockResults request height")
```
thanethomson, line 40:
```suggestion
```

Could pollute operators' logs. And the full error's returned to the caller anyways.

test/e2e/tests/grpc_test.go
```diff
@@ -132,3 +135,60 @@ func TestGRPC_Block_GetLatestHeight(t *testing.T) {
...
+		require.GreaterOrEqual(t, last, latestBlockResults.Height)
+		require.NoError(t, err, "Unexpected error for GetLatestBlockResults")
+		require.NotNil(t, latestBlockResults)
+
+		successCases := []struct {
+			expectedHeight int64
+			requestHeight  int64
+		}{
+			{first, first},
+			{latestBlockResults.Height, latestBlockResults.Height},
```
thanethomson, line 173: The expected and requested should always be the same, right? If so I suspect we only need `requestHeight` in the test cases.

test/e2e/tests/grpc_test.go
```diff
@@ -132,3 +135,60 @@ func TestGRPC_Block_GetLatestHeight(t *testing.T) {
...
+		}
+		errorCases := []struct {
+			requestHeight int64
+		}{
+			{-1},
+			{100000},
```
thanethomson, line 179: This may end up being non-deterministic. I'd recommend doing something like:

```suggestion
			{first-1},
			{last+10},
```

thanethomson (approved):
Looking good overall! Just a few minor recommendations, but then I'd say this is good to merge.

test/e2e/tests/grpc_test.go
glnro: Yes fair, this is leftover from before but it's not needed now

test/e2e/tests/grpc_test.go
glnro: See [updated test](https://github.com/cometbft/cometbft/pull/1168/commits/8d3a2a6f7f209a6cf749d4c4a37a31154cb0fb8f)

config/toml.go
```diff
@@ -280,6 +280,11 @@ enabled = {{ .GRPC.VersionService.Enabled }}
 [grpc.block_service]
 enabled = {{ .GRPC.BlockService.Enabled }}
 
+# The gRPC block results service returns block results for a given height. If no height
+# is given, it will return the block results from the latest height.
```
jmalicevic, line 284: This text might be before your separation of the services into `GetLatestHeight` and `GetHeight`. Atm in the code if the req.Height is 0 we do not return the latest height right?

PR #1169, opened 2023-07-26
mempool: test FIFO ordering when disseminating txs

hvanz:
This code was originally in #1043, and extracted from there.

The main addition here is the test `TestReactorPeerLagging`. Also checking that transactions are in order in `TestReactorTxSendersMultiNode`.

---

#### PR checklist

- [ ] Tests written/updated


hvanz:
The tests in this PR fail intermittently because they test for FIFO ordering, that is, they require that transactions arrive to peers in the order in which they were sent. FIFOness is really difficult to enforce, so we should consider whether it must be a requirement on the mempool and an assumption by the application.

lasarojc:
FIFO is not really provided and we should check the documentation and code where this assumption might be made to remove/correct it.

lasarojc:
The test added in #1628 are more specific in demonstrating that FIFO is not provided. 
I am closing this PR in favor or #1628. Please reopen if you think this is a mistake.

PR #1170, opened 2023-07-26
cmd: Remove `replay` and `replay-console` subcommands

thanethomson:
While working towards #1055, I came across these subcommands that don't seem to be actively used anywhere. It's also not clear why these commands would be exposed at all, and some of the underlying code seems half-implemented.

They're also not exported by the Cosmos SDK.

I'll remove them on `main` and mark them as deprecated in v0.38. If anyone has any issues with this, please feel free to reach out and discuss your use case with us.

---

#### PR checklist

- [ ] Tests written/updated


adizere (approved):
Nice cleanup!

PR #1173, opened 2023-07-27
replace internet route-able ipv4s in docker-compose

gregnuj:
replace internet routeable ipv4s in docker-compose
ref: https://datatracker.ietf.org/doc/html/rfc1918
https://ip-netblocks.whoisxmlapi.com/lookup-report/8L2ojwyPRE

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,7 +13,7 @@ services:
       - ./build:/cometbft:Z
     networks:
       localnet:
-        ipv4_address: 192.167.10.2
+        ipv4_address: 192.168.10.2
 
   node1:
     container_name: node1
@@ -27,7 +27,7 @@ services:
       - ./build:/cometbft:Z
     networks:
       localnet:
-        ipv4_address: 192.167.10.3
+        ipv4_address: 192.168.10.3
 
   node2:
     container_name: node2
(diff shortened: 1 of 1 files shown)


thanethomson (changes requested):
Thanks for bringing this up @gregnuj - could you please explain the problem you're encountering to which this is the solution?

If this is a problem, there are two other changes that still need to be made for us to accept this:
1. The `localnet` target in the `Makefile` needs to be updated: https://github.com/cometbft/cometbft/blob/c3acdb866f885e94dcd0e1ef8ecce53ff689216c/Makefile#L287-L289
2. The [docs](https://github.com/cometbft/cometbft/blob/main/docs/networks/docker-compose.md) need to be updated

sergio-mena:
Further on @thanethomson's answer, what is the problem we are trying to solve?
RFC1918 defines a set of network prefixes (private networks) that MUST NOT be routable outside their subnet (10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16), but it does not mandate anything about the rest of prefixes (you still still use any other prefix for your private network, and it's up to you if you want it to be reachable).
In short, (as mentioned at the beginning) I don't see the problem this PR is trying to solve.

Issue #1171, opened 2023-07-26
Conform consensus state to Tendermint technical specification


BrendanChou:
### Setup

**CometBFT version**:
v0.37.2 and before

**Have you tried the latest version**:
Yes

**ABCI app**:
N/A

**Environment**:
N/A

**node command runtime flags**:
N/A

### Config
N/A

### What happened?

Discrepancies between the Tendermint implementation code and whitepaper were found
https://github.com/tendermint/tendermint/issues/6849
https://github.com/tendermint/tendermint/issues/6850

### What did you expect to happen?

These were fixed in Tendermint core in 2021, these changes did not seem to make it to CometBFT somehow, perhaps due to a code rollback of some sort

### How to reproduce it
N/A

### Anything else we need to know
These fixes are considered to block larger, previously-accepted consensus changes such as Proposer-Based Timestamps https://github.com/tendermint/tendermint/issues/6942

sergio-mena:
Taking a close look at tendermint/tendermint#6849 and tendermint/tendermint#6850 before forward-porting them to `main` (as a first step... then back to the relevant branches: possibly `v0.38.x` and `v0.37.x`)

BrendanChou:
The main difference for https://github.com/tendermint/tendermint/issues/6850 is that it was written before `ProcessProposal` existed. Therefore, the fall-through-case (the case when reaching the end of the function) is now different from when it was written, requiring a rewrite.

https://github.com/tendermint/tendermint/issues/6849 can be ported largely unchanged, but also includes some minor cleanup which I consider unnecessary. I have ported it here https://github.com/cometbft/cometbft/pull/1175

sergio-mena:
Closed by #1175, #1207, and #1203

PR #1175, opened 2023-07-27
consensus: remove logic to unlock block on +2/3 prevote for nil

BrendanChou:
- Fixes part of https://github.com/cometbft/cometbft/issues/1171 by implementing [this logic](https://github.com/cometbft/cometbft/blob/127d2fb78b271847de4dc0aaca7825bca98a2cb9/docs/introduction/README.md?plain=1#L305-L306)
- Largely copies logic from https://github.com/tendermint/tendermint/pull/6954
  -  (-) Doesn't include the renaming change of `blockID.IsZero()` to `blockID.IsNil()`
  - (-) Doesn't rename tests to include underscores.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/.changelog/unreleased/bug-fixes/1175-consensus-remove-unlock-behavior.md
+++ b/.changelog/unreleased/bug-fixes/1175-consensus-remove-unlock-behavior.md
@@ -1,0 +1,1 @@
+- `[consensus]` \#1175 remove logic to unlock block on +2/3 prevote for nil ([cometbft/cometbft\#1175](https://github.com/cometbft/cometbft/pull/1175))
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -1405,7 +1405,6 @@ func (cs *State) enterPrevoteWait(height int64, round int32) {
 // Enter: `timeoutPrecommit` after any +2/3 precommits.
 // Enter: +2/3 precomits for block or nil.
 // Lock & precommit the ProposalBlock if we have enough prevotes for it (a POL in this round)
-// else, unlock an existing lock and precommit nil if +2/3 of prevotes were nil,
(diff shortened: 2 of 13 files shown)


sergio-mena (approved):
Approving this PR, given that the differences between this and a mechanical forward port only amounts to spacing, some comments, and unimportant function renaming.
The differences mentioned in the previous sentence can be seen with the following command:
`git fetch --all; git diff 9cb38b9f6f..e5602e7dba`

.changelog/unreleased/bug-fixes/1175-consensus-remove-unlock-behavior.md
```diff
@@ -0,0 +1 @@
+- `[consensus]` \#1175 remove logic to unlock block on +2/3 prevote for nil ([cometbft/cometbft\#1175](https://github.com/cometbft/cometbft/pull/1175))
```
sergio-mena, line 1:
```suggestion
- `[consensus]` \#1175 remove logic to unlock block on +2/3 prevote for nil ([cometbft/cometbft\#1175](https://github.com/cometbft/cometbft/pull/1175): @BrendanChou )
```

consensus/state.go
```diff
@@ -1486,7 +1473,9 @@ func (cs *State) enterPrecommit(height int64, round int32) {
 		return
 	}
 
-	// If +2/3 prevoted for proposal block, stage and precommit it
+	// If greater than 2/3 of the voting power on the network prevoted for
```
lasarojc, line 1476: The previous if case is just a sub-case of the following if, and should be coded as such. It will be handled in a follow up PR.

lasarojc (approved):
LGTM

PR #1176, opened 2023-07-28
Anton/tx index pruning

### LINKED ISSUE #169: storage+indexer: Indexer is not pruned
## Summary

Storage usage of CometBFT nodes is becoming a problem for many operators. Reducing this usage is one of the major goals for 2023 (#44 ). To reduce storage usage, CometBFT allows operators to prune blocks below a certain height.  However, this pruning does not include data indexed by the indexers. 

## Problem definition

 As part of this effort, people have already gathered some information on the storage usage of CometBFT instances (informalsystems/interchain#1). Based on the reports mentioned in the issue, the storage usage of the kvindexer is almost 50%  of the total usage. The fact that pruning is enabled does not impact the indexer as indexer pruning is not implemented at the moment (Jan 2023). 

Note that the way the kvindexer is implemented at the moment, it stores all transaction results and block events (begin block and end block - which will be finalize block in the future). Thus, it has a summary of the entire blockchain which is never pruned. 

## Proposal

- We need to understand whether pruning the indexer impacts its users. My assumption is that, if users prune the blockstore (blockchain), they would be ok with pruning the indexer as well. But this is hard to know, there can be scenarios where for example relayers do depend on having all transaction events even though the actual transactions do not exist in the blockstore. 

- In any case, providing the users an option to prune the indexer would be of significant importance to some users. This can however be troublesome: 
    -  We have to find a way on how to identify an index entry - retrieve its key; in order to delete it. For the transaction indexer this is possible because the transaction results are kept in other stores. The key is derived from that (for each transaction it consists of the event attributes and their values- which are stored). Block events are not kept anywhere and it thus makes key retrieval impossible (the event attributes are not retrievable). We need to investigate whether we can use timestamps or heights to prune the store. 
    - Pruning the transaction indexer (based on a prototype implementation) can be slow. Thus this can be done as a maintenance task once a day. We would need to test the impact on performance and runtime to establish the impact indexer pruning has on performance. 

Even when we take the above in consideration as well as the current considerations on whether we should completely redesign the indexing capabilities of CometBFT(#73 and #82 ), given the huge storage usage of the indexer and the fact that v0.34.x is still widely used, I think it is worth investigating how to reduce it, even if we cannot entirely reduce it.

As we might not be able to be 100% sure that users are ok with pruning the indexer, we should make this a config flag that is false by default. At least for the initial release. 

## DoD

As a result of this work we will either decide to implement some kind of indexer pruning or have clear arguments as to why we will not do it and how will users problems with regards to its storage use be addressed (new design or alternative indexer implementation), 

- [ ] Understand whether users need all the event data to be present in the indexer
- [ ] Will ABCI 2.0 lead to no events being stored anywhere (thus making it impossible to generate event keys for the indexer)
   -  [ ] If yes - determine whether the indexer can be pruned based on timestamp or height
   -  [ ] If no - implement indexer pruning. 
- [ ] If pruning is implemented run experiments to evaluate impact on performance when indexer pruning is on - large testnet as well as e2e tests.

werty144:
Closes #169 

#### PR checklist

- [x] Tests written/updated


werty144:
# Pruning for Indexer
This PR introduces a pruning for the `tx_index.db` considered in #169.

Pruning is a process of removing old information from the database at the application demand. It was previously implemented for the `state.db` and `blockstrore.db` but not for the `tx_index.db`. 

The following chart shows the importance of being able to prune `tx_index.db`. As one can see, it is responsible for almost a half of node's disk space usage.
image

## Result
The following chart shows the disk usage behavior without pruning and after applying the changes of the current PR.
image

## Technical details
The pruning is triggered upon `state/execution.go ApplyBlock` call and propagates to the `IndexerService` (`state/txindex/indexer_service.go`) via the `eventBus`.

### Important files
- `state/indexer/block/kv/kv.go` - implementation of pruning for the block indexer
- `state/txindex/kv/kv.go` - implementation of pruning for the transaction indexer

## Future work

1. Currently the implementation does not support SQL backend.
2. The deletion of block events in `state/indexer/block/kv/kv.go deleteEvents` is done in linear time with respect to the number of keys in the storage. One might consider taking the measurements to see if that is an issue and needs to be addressed.

thanethomson (commented):
Great overall work here @werty144! Very good insights re: how much storage space could be saved by pruning the indexes.

I'd like understand how computationally intensive the approach is that you've taken - is there any way to get a good sense of that?

What I would, however, recommend is basing this work on top of #1150. If we go with this specific approach, it seems like it should actually be part of the `state.Pruner` behavior (probably part of the [pruningRoutine](https://github.com/cometbft/cometbft/pull/1150/files#diff-85847e2c654966518d4ab21f147cf243a6c7db53f897c5ed6e494ca18b939528R167), or its own goroutine if it runs less frequently than the other pruning operations), such that this pruning can happen automatically in the background on a regular basis.

state/indexer/block/kv/kv.go
```diff
@@ -86,6 +86,25 @@ func (idx *BlockerIndexer) Index(bh types.EventDataNewBlockEvents) error {
 	return batch.WriteSync()
 }
 
+func (idx *BlockerIndexer) Prune(retainHeight int64) {
+	heights, err := idx.Search(context.Background(),
```
jmalicevic, line 90: Do we need to search for the heights before trying out the code in 95? As in-will the deletion not fail anyways if we don't have this height.

state/txindex/kv/kv.go
```diff
@@ -40,6 +41,21 @@ type TxIndex struct {
 	log log.Logger
 }
 
+func (txi *TxIndex) Prune(retainHeight int64) {
+	ctx := context.Background()
+	results, err := txi.Search(ctx, query.MustCompile(fmt.Sprintf("tx.height < %d", retainHeight)))
```
jmalicevic, line 46: Same comment as for the block indexer. Can we not just scan the DB and try to delete a height. If it is not there, it's not there. This way we do a search to find the heights that are there and then we do a scan anyways and try to delete those keys.

state/indexer/block/kv/kv.go
werty144: Done

state/txindex/kv/kv.go
werty144: To me, it is not as easy as with blocks, since here `Search` returns the results, not just the heights. In particular, I can optimize the deletion of events (see lines 217-228):
```[go]
zeroKey := keyForEvent(compositeTag, attr.Value, result, 0)
endKey := keyForEvent(compositeTag, attr.Value, result, math.MaxInt64)
itr, err := txi.store.Iterator(zeroKey, endKey)
if err != nil {
	panic(err)
}
for ; itr.Valid(); itr.Next() {
	err := txi.store.Delete(itr.Key())
	if err != nil {
		panic(err)
	}
}
```

Here I use the `result` found by the search not to traverse all the keys, but only those that are for specific height and event type

types/events.go
```diff
@@ -75,9 +76,10 @@ type EventDataNewBlockHeader struct {
 }
 
 type EventDataNewBlockEvents struct {
-	Height int64        `json:"height"`
-	Events []abci.Event `json:"events"`
-	NumTxs int64        `json:"num_txs,string"` // Number of txs in a block
+	Height       int64        `json:"height"`
+	Events       []abci.Event `json:"events"`
+	NumTxs       int64        `json:"num_txs,string"` // Number of txs in a block
+	RetainHeight int64        `json:"retain_height"`  // TODO consider another place
```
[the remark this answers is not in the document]
jmalicevic: I know it complicates things even more, but ABCI responses and the indexed data are not subject to the same retention policy. ABCI responses are stored in the blockstore if `DiscardABCIResponses = true` . Events are indexed always upon firing `EventDataNewBlockEvents`. So I am not convinced we should have the same retain height for both.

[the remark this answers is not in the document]
thanethomson: Yes, actually, you're right.

The block retain height, however, will act as the minimum possible value for both the tx and block indexes would it not? Since when using `/tx_search` we need to load the relevant block(s) here: https://github.com/cometbft/cometbft/blob/24ebdd9df78d15828aa758e16a21d82fadcbb95b/rpc/core/tx.go#L116

And when using `/block_search` we need to load the relevant block(s) here: https://github.com/cometbft/cometbft/blob/24ebdd9df78d15828aa758e16a21d82fadcbb95b/rpc/core/blocks.go#L253

Does it then make sense really to prune indexes above the block retain height? I wouldn't think so - in that case it's better to just turn the indexers off.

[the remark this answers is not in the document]
jmalicevic: Hmm, so technically, even now, the blocks can be pruned and the `tx_search` would still have the indexing data for those heights. If you remember we tried to reason about this when we initially talked about pruning. And then I think some of our users said that even if they prune blocks they keep the indexer. Now I am not sure what scenario that could be for: Maybe use the indexer to see whether there is a match and then they have the blocks on special archive nodes where they fetch them from (not from the node which was queried via `tx_search`. This might be something we can ask Adi to find out?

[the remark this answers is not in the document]
thanethomson: So since we'll be offering folks the ability to index their data outside of Comet, does it really make sense to keep the tx index if the related blocks have been pruned?

cc @adizere

thanethomson:
@werty144 I've changed the base branch for this PR to `jasmina/1096-adr-101-pruning-mechanism` after your rebase to filter out all the base branch's commits from this PR. Feel free to change it back if that wasn't what you were aiming for.

state/indexer/block/kv/kv.go
```diff
@@ -86,6 +86,20 @@ func (idx *BlockerIndexer) Index(bh types.EventDataNewBlockEvents) error {
 	return batch.WriteSync()
 }
 
+func (idx *BlockerIndexer) Prune(retainHeight int64) {
+	for height := int64(0); height < retainHeight; height++ {
+		key, err := heightKey(height)
+		if err != nil {
+			panic(err)
+		}
+		err = idx.store.Delete(key)
```
mzabaluev, line 95: This looks massively inefficient. If this is performed regularly, the amount of work going through already pruned indices will grow as the retain height grows.

Can the store keys that are actually present be enumerated in an ordered way, from earliest height towards the latest?

mzabaluev (changes requested):

state/indexer/block/kv/kv.go
jmalicevic: What can be done is something similar to what we do in the pruning mechanism - remember the last retain height and prune between the last retain height and the retain height. See how it's retrieved and used [here](https://github.com/cometbft/cometbft/blob/0dd35c933a1ccea1a2e439a8a74b5fd5cc9103fd/state/store.go#L395C22-L395C40).

state/pruner.go
```diff
@@ -81,6 +92,7 @@ func (p *Pruner) SetObserver(obs PrunerObserver) {
 
 func (p *Pruner) OnStart() error {
 	go p.pruningRoutine()
+	go p.indexerPruningRoutine()
```
jmalicevic, line 95: great job! This will make sure the indexer pruning is not in the way of the other pruning processes.

state/pruner.go
```diff
@@ -59,17 +62,25 @@ func WithPrunerObserver(obs PrunerObserver) PrunerOption {
...
-		stateStore: stateStore,
-		logger:     logger,
-		interval:   cfg.interval,
-		observer:   cfg.observer,
+		bs:              bs,
+		stateStore:      stateStore,
+		indexerService:  indexerService,
+		logger:          logger,
+		interval:        cfg.interval,
+		indexerInterval: cfg.indexerInterval,
```
jmalicevic, line 82: For simplicity , I think it is OK to use the same interval for the indexer pruning as for the other pruning jobs.

state/indexer/block/kv/kv.go
```diff
@@ -86,6 +115,24 @@ func (idx *BlockerIndexer) Index(bh types.EventDataNewBlockEvents) error {
 	return batch.WriteSync()
 }
 
+func (idx *BlockerIndexer) Prune(retainHeight int64) {
+	if retainHeight < idx.lbRetainedHeight {
```
jmalicevic, line 119: This should be retrieved from the Database if it is not yet set.

state/txindex/kv/kv.go
jmalicevic, line 46: I think here you can benefit from the same optimization of caching the lastRetain height, so as to avoid calling `Search` for the same height over and over again.

state/indexer/block/kv/kv.go
```diff
@@ -648,28 +653,35 @@ func (idx *BlockerIndexer) indexEvents(batch dbm.Batch, events []abci.Event, hei
 				if err := batch.Set(key, heightBz); err != nil {
 					return err
 				}
+				keyArray = AppendToKeyArray(keyArray, key)
 			}
 		}
 	}
+	if err := batch.Set(getEventsForHeightKey(height), keyArray); err != nil {
```
jmalicevic: Instead of introducing a new key, you can actually reuse the already existing height key. In the block indexer you can [see](https://github.com/cometbft/cometbft/blob/1e0cf136d395db94728883355a136e4e34e693db/state/indexer/block/kv/kv.go#L77) that the height key has the heigth as well as the value. So you can easily replace it.  You can use the same key which will overwrite the previous value stored.

.gitignore
```diff
@@ -33,6 +33,8 @@ remote_dump
 rpc/test/.cometbft
 scripts/cutWALUntil/cutWALUntil
 scripts/wal2json/wal2json
+scripts/db_measurements/measurements.txt
```
jmalicevic, line 36: Should this be merged or  was this local?

go.mod
```diff
@@ -262,6 +265,8 @@ require (
 	github.com/t-yuki/gocover-cobertura v0.0.0-20180217150009-aaee18c8195c // indirect
 	github.com/tdakkota/asciicheck v0.2.0 // indirect
 	github.com/tecbot/gorocksdb v0.0.0-20191217155057-f0fad39f321c // indirect
+	github.com/tendermint/tendermint v0.35.9 // indirect
```
jmalicevic, line 268: This is weird, we should not be pulling this dependency in.

state/indexer/block/kv/kv.go
```diff
@@ -34,13 +36,44 @@ type BlockerIndexer struct {
...
 func New(store dbm.DB) *BlockerIndexer {
 	return &BlockerIndexer{
-		store: store,
+		store:            store,
+		lbRetainedHeight: getLBRetainHeight(store),
+	}
+}
+
+func getEventsForHeightKey(height int64) []byte {
+	return []byte(fmt.Sprintf("eventsForHeightKey%d", height))
```
jmalicevic, line 52: As mentioned below, you don't need a new key for this, you can reuse the existing height key , just instead of storing the height as the value (which is done right now), you store the new data.

node/node_test.go
```diff
@@ -96,6 +96,20 @@ func TestSplitAndTrimEmpty(t *testing.T) {
 	}
 }
 
+func TestCompanionInitialHeightSetup(t *testing.T) {
```
jmalicevic, line 99: Why is this part of this PR? MY guess is its because of a merge that did not fully go through?

state/pruner.go
werty144: Done

state/indexer/block/kv/kv.go
werty144: I retrieve it from the DB in the `func New(store dbm.DB) *BlockerIndexer`, inside `getLBRetainHeight`. 

My logic was to retrieve it a single time at the beginning and then operate with the in-memory value.

state/pruner.go
```diff
@@ -0,0 +1,375 @@
...
+				},
+			})
+			lastRetainHeight = newRetainHeight
+			time.Sleep(p.interval)
+		}
+	}
+}
+
+func (p *Pruner) indexerPruningRoutine() {
+	p.logger.Info("Index pruner started", "interval", p.indexerInterval.String())
```
jmalicevic, line 264:
```suggestion
	p.logger.Info("Index pruner started", "interval", p.indexerInterval.String())
	lastRetainHeight := int64(0)
```

state/pruner.go
```diff
@@ -0,0 +1,375 @@
...
+	for {
+		select {
+		case <-p.Quit():
+			return
+		default:
+			p.pruneIndexesToRetainHeight()
```
jmalicevic, line 270:
```suggestion
			lastRetainHeight = p.pruneIndexesToRetainHeight(lastRetainHeight)
```

go.mod
thanethomson: Yeah something's wrong here. We should _definitely_ not be seeing Tendermint or tm-db being pulled in here.

state/indexer/block/kv/util.go
```diff
@@ -209,3 +210,61 @@ func checkHeightConditions(heightInfo HeightInfo, keyHeight int64) (bool, error)
 	}
 	return true, nil
 }
+
+func GetEventsForTesting(height int64) types.EventDataNewBlockEvents {
```
jmalicevic, line 214: Can this go into a _test function so we avoid exposing this to the users?

state/indexer/block/kv/kv.go
```diff
@@ -86,6 +90,20 @@ func (idx *BlockerIndexer) Index(bh types.EventDataNewBlockEvents) error {
 	return batch.WriteSync()
 }
 
+func (idx *BlockerIndexer) Prune(lastRetainHeight int64, retainHeight int64) {
+	for height := lastRetainHeight; height < retainHeight; height++ {
+		key, err := heightKey(height)
+		if err != nil {
```
jmalicevic, line 96: The error here can be `ErrNotFound` which means the key does not exist, so I am not 100% sure you should panic here. I imagine we might have a block without events for a height.  The deletion errors are fine because they will not happen if the key does not exist.

state/txindex/indexer_service.go
```diff
@@ -119,9 +137,82 @@ func (is *IndexerService) OnStart() error {
...
+	lastRetainHeight, err := is.getLastIndexerRetainHeight()
+	if err != nil {
+		panic(err)
+	}
+	if retainHeight <= lastRetainHeight {
+		return
+	}
+	is.txIdxr.Prune(lastRetainHeight, retainHeight)
+	is.blockIdxr.Prune(lastRetainHeight, retainHeight)
+	is.setLastIndexerRetainHeight(retainHeight)
```
jmalicevic, line 150: Can it ever happen that the block and tx indexer do not prune until retain height?

state/txindex/indexer_service.go
```diff
@@ -119,9 +137,82 @@ func (is *IndexerService) OnStart() error {
...
+
+	if height < 0 {
+		return 0, ErrInvalidHeightValue
+	}
+
+	return height, nil
+}
+
+func (is *IndexerService) getLastIndexerRetainHeight() (int64, error) {
+	bz, err := is.getValue(LastIndexerRetainHeightKey)
```
jmalicevic, line 172: I think that, instead of making this a generic 'GetValue` , you should have a `getLastIndexerRetainHeight` where you explicitly query only for that key. I know these functions are all private so users cannot misuse them, but as this is the pattern followed elsewhere it might make more sense.

state/indexer/block/kv/kv.go
werty144: To my understanding, this is not the search for the key, but the creation of one as a function of height. This does not call the BD. I might have misunderstood you though

jmalicevic: ah right ,I understood it was trying to retrieve the key before deleting. All good, disregard the comment.

state/indexer/block/kv/kv.go
```diff
@@ -601,9 +620,32 @@ func (idx *BlockerIndexer) indexEvents(batch dbm.Batch, events []abci.Event, hei
...
 	}
+	err := batch.Set(getEventsForHeightKey(height), keyArray)
+	return err
+}
 
-	return nil
+func (idx *BlockerIndexer) deleteEvents(height int64) {
+	eventsKey := getEventsForHeightKey(height)
+	keyArray, err := idx.store.Get(eventsKey)
+	if err != nil {
```
jmalicevic, line 634: Maybe here it's worth checking whether the error is `ErrNotFound` and not panicking.

state/indexer/block/kv/kv.go
```diff
@@ -90,18 +90,23 @@ func (idx *BlockerIndexer) Index(bh types.EventDataNewBlockEvents) error {
...
-			panic(err)
+			return height, err
 		}
 		err = idx.store.Delete(key)
 		if err != nil {
-			panic(err)
+			return height, err
+		}
+		if err = idx.deleteEvents(height); err != nil {
+			return height, err
```
jmalicevic, line 106:
```suggestion
              if !errors.Is(err, ErrKeyNotFound) {
     			return height, err
     	       } 
     	        continue
```

state/txindex/indexer_service.go
```diff
@@ -180,22 +203,42 @@ func (is *IndexerService) getLastIndexerRetainHeight() (int64, error) {
...
 }
 
-func (is *IndexerService) setLastIndexerRetainHeight(height int64) {
-	if err := is.indexerStore.SetSync(LastIndexerRetainHeightKey, int64ToBytes(height)); err != nil {
+func (is *IndexerService) setLastTxIndexerRetainHeight(height int64) {
+	currentHeight, err := is.getLastTxIndexerRetainHeight()
+	if err != nil && !errors.Is(err, ErrKeyNotFound) {
+		panic(err)
+	}
+	if height < currentHeight {
```
jmalicevic, line 211:
```suggestion
	if height <= currentHeight {
```

state/txindex/kv/kv_test.go
```diff
@@ -782,3 +784,56 @@ func BenchmarkTxIndex500(b *testing.B)   { benchmarkTxIndex(500, b) }
 func BenchmarkTxIndex1000(b *testing.B)  { benchmarkTxIndex(1000, b) }
 func BenchmarkTxIndex2000(b *testing.B)  { benchmarkTxIndex(2000, b) }
 func BenchmarkTxIndex10000(b *testing.B) { benchmarkTxIndex(10000, b) }
+
+func equal(x []byte, y []byte) bool {
```
jmalicevic, line 788: Replace with `bytes.Equal()`. See: https://pkg.go.dev/bytes#Equal

jmalicevic: Actually I think many of these can be replaced with the default bytes functons.

state/txindex/kv/utils.go
```diff
@@ -106,3 +107,23 @@ func checkHeightConditions(heightInfo HeightInfo, keyHeight int64) (bool, error)
 	}
 	return true, nil
 }
+
+// GetKeys is used for testing purposes only
+func GetKeys(indexer *TxIndex) [][]byte {
```
jmalicevic, line 112: Can we make this not public, or be in `internal/indexer_utils`?

state/txindex/kv/kv_test.go
```diff
@@ -717,3 +786,41 @@ func BenchmarkTxIndex500(b *testing.B)   { benchmarkTxIndex(500, b) }
...
+			return bytes.Equal(i, elem)
+		}) {
+			return false
+		}
+	}
+	return true
+}
+
+func isEqualSets(x [][]byte, y [][]byte) bool {
+	return isSubset(x, y) && isSubset(y, x)
```
jmalicevic, line 802:
```suggestion
       if len(x) == len(y) {
           return isSubset(x, y) && isSubset(y, x)
       } else {
          return false
        }
```

state/indexer/block/kv/kv_test.go
```diff
@@ -437,3 +501,99 @@ func TestBigInt(t *testing.T) {
...
+			return bytes.Equal(i, elem)
+		}) {
+			return false
+		}
+	}
+	return true
+}
+
+func isEqualSets(x [][]byte, y [][]byte) bool {
+	return isSubset(x, y) && isSubset(y, x)
```
jmalicevic, line 575:
```suggestion
       if len(x) == len(y) {
	   return isSubset(x, y) && isSubset(y, x)
	} else {
	   return false
	}
```

state/indexer/block/indexer.go
```diff
@@ -20,28 +20,28 @@ import (
 // configuration.
 //
 //nolint:lll
-func IndexerFromConfig(cfg *config.Config, dbProvider config.DBProvider, chainID string) (txindex.TxIndexer, indexer.BlockIndexer, error) {
+func IndexerFromConfig(cfg *config.Config, dbProvider config.DBProvider, chainID string) (txindex.TxIndexer, indexer.BlockIndexer, dbm.DB, error) {
```
thanethomson, line 23: We'll need a specific changelog entry for this, given that it's a Go API-breaking change.

state/txindex/kv/utils.go
werty144: Not really, since it accesses `indexer.store`...

state/txindex/indexer_service.go
```diff
@@ -14,13 +17,23 @@ const (
...
+)
+
 // IndexerService connects event bus, transaction and block indexers together in
 // order to index transactions and blocks coming from the event bus.
 type IndexerService struct {
 	service.BaseService
 
 	txIdxr           TxIndexer
 	blockIdxr        indexer.BlockIndexer
+	indexerStore     dbm.DB
```
thanethomson, line 36: I'd like to understand a bit more about the reasoning here in keeping this `indexerStore` within this `IndexerService`. Does it not make more sense to implement this as part of the `TxIndexer` and `BlockIndexer` implementations?

werty144: Now it is implemented as part of `TxIndexer` and `BlockIndexer`. `IndexerService` is untouched

state/pruner_test.go
```diff
@@ -0,0 +1,181 @@
...
+			return bytes.Equal(i, elem)
+		}) {
+			return false
+		}
+	}
+	return true
+}
+
+func isEqualSets(x [][]byte, y [][]byte) bool {
+	return isSubset(x, y) && isSubset(y, x)
```
jmalicevic, line 168: You can check first whether the lengths are equal, to avoid scanning if they are not.

werty144: I don't really want to do that, cause formally speaking there might be duplicates, so two slices of different lengths might be equal as sets. 
Also, those slices in all cases are of size < 100, so that's a bit of an overkill optimization... 
And that's only used in testing:) 

But if you want to see that, I would do, not a big deal:)

jmalicevic: Ah so you can have duplicates and then you consider them as equal. Got it, that is fine then. Also as you said, if its just for testing the perf gain is not that big of a deal.

sergio-mena:
May I suggest to rename this PR to something clearer?

state/indexer/block/kv/kv.go
```diff
@@ -82,8 +91,101 @@ func (idx *BlockerIndexer) Index(bh types.EventDataNewBlockEvents) error {
 	if err := idx.indexEvents(batch, bh.Events, height); err != nil {
 		return fmt.Errorf("failed to index FinalizeBlock events: %w", err)
 	}
+	err = batch.WriteSync()
+	return err
+}
+
+func getKeys(indexer BlockerIndexer) [][]byte {
```
jmalicevic, line 98: is this used only in `kv_test.go` ? If so, I think it should be moved there.

state/indexer/block/kv/kv.go
```diff
@@ -601,9 +704,30 @@ func (idx *BlockerIndexer) indexEvents(batch dbm.Batch, events []abci.Event, hei
...
 		}
 	}
+	err := batch.Set(getEventsForHeightKey(height), keyArray)
+	return err
+}
 
-	return nil
+func (idx *BlockerIndexer) deleteEvents(height int64) error {
+	eventsKey := getEventsForHeightKey(height)
+	keyArray, err := idx.store.Get(eventsKey)
```
jmalicevic, line 717: I just realized that for indexed data, in old versions of comet, this will not exist. Therefore you still need to have a loop through the events if the error is that they key is not found. 

So below
```
  if err == ErrKeyNotFound{
      deleteEventsSequentially(height)
   } else {
          return err 
    }

```

state/txindex/kv/kv.go
```diff
@@ -40,6 +47,82 @@ type TxIndex struct {
...
+	results, err := txi.Search(ctx, query.MustCompile(
+		fmt.Sprintf("tx.height < %d AND tx.height >= %d", retainHeight, lastRetainHeight)))
+	if err != nil {
+		panic(err)
+	}
+	if len(results) == 0 {
+		return 0, lastRetainHeight, nil
+	}
+
+	sort.Sort(TxResultByHeight(results))
```
jmalicevic, line 70: I am not sure how expensive this is. I understand you need it because you need to determine the lastRetainHeight. But in theory, if the result was returned by the Search then it should also be deletable .

state/pruner.go
```diff
@@ -274,6 +350,75 @@ func (p *Pruner) pruneBlocksRoutine() {
...
+	if lastRetainHeight >= targetRetainHeight {
+		return lastRetainHeight
+	}
+
+	numPrunedBlockIndexer, newBlockIndexerRetainHeight, err := p.blockIndexer.Prune(targetRetainHeight)
+	if err != nil {
+		p.logger.Error("Failed to prune block indexer", "err", err, "targetRetainHeight", targetRetainHeight, "newBlockIndexerRetainHeight", newBlockIndexerRetainHeight)
+	} else if numPrunedBlockIndexer > 0 {
+		// TODO call metrics
+		p.logger.Info("Pruned block indexer", "count", numPrunedBlockIndexer, "newBlockIndexerRetainHeight", newBlockIndexerRetainHeight)
```
jmalicevic, line 417:
```suggestion
		p.logger.Debug("Pruned block indexer", "count", numPrunedBlockIndexer, "newBlockIndexerRetainHeight", newBlockIndexerRetainHeight)
```

We will introduce metrics to monitor this.

state/indexer/block/kv/kv.go
werty144: It's also exported for testing in `state/indexer/block/kv/export_testing.go`

state/indexer/block/kv/kv.go
werty144: Hmm... good point. I would just point out that it's `store.Get`, so if a key is not found, there is no error, but just the returned value is nil

state/txindex/kv/kv.go
werty144: Does the Search provide results sorted?

werty144: My logic was this is not expansive, since sorts work in _O(n*log(n))_ where n is the number of results. And we pay at least a cost linear in _n_ , so not that much of an overhead

jmalicevic: No they are not sorted so this is ok the way it is. We really do hope the indexer will go away eventually :)

go.mod
```diff
@@ -216,6 +216,7 @@ require (
 	github.com/nishanths/predeclared v0.2.2 // indirect
 	github.com/nunnatsa/ginkgolinter v0.13.3 // indirect
 	github.com/olekukonko/tablewriter v0.0.5 // indirect
+	github.com/onsi/ginkgo v1.16.4 // indirect
```
thanethomson, line 219: What're these additional dependencies for? Does running `go mod tidy` remove them?

state/indexer/block.go
```diff
@@ -24,4 +24,10 @@ type BlockIndexer interface {
 	Search(ctx context.Context, q *query.Query) ([]int64, error)
 
 	SetLogger(l log.Logger)
+
+	Prune(retainHeight int64) (int64, int64, error)
+
+	SetBlockIndexerRetainHeight(retainHeight int64) error
+
+	GetBlockIndexerRetainHeight() (int64, error)
```
thanethomson, line 32: Do we need such long function names? Could we call these `SetRetainHeight` and `GetRetainHeight` instead? They are already associated with the `BlockIndexer` interface, so unless we're composing this interface with another one that has similarly named functions, these names seem too verbose.

.changelog/unreleased/improvements/169-tx_index-pruning.md
```diff
@@ -0,0 +1,2 @@
+- `[state/indexer]` Add pruning for the tx_index.db.
```
thanethomson, line 1:
```suggestion
- `[state/indexer]` Add transaction and block index pruning
```

Users don't necessarily know what `tx_index.db` is, and this PR does more than just facilitate transaction index pruning. Changelog entries are not for us, they're for our users.

state/txindex/indexer.go
```diff
@@ -31,6 +31,12 @@ type TxIndexer interface {
 
 	//Set Logger
 	SetLogger(l log.Logger)
+
+	Prune(retainHeight int64) (int64, int64, error)
+
+	GetTxIndexerRetainHeight() (int64, error)
+
+	SetTxIndexerRetainHeight(retainHeight int64) error
```
thanethomson, line 39: Similar to my comments on the `BlockIndexer` interface - can we not just call these functions `GetRetainHeight` and `SetRetainHeight`?

thanethomson (commented):
Amazing work so far here @werty144!

state/pruner.go
```diff
@@ -116,8 +121,12 @@ func (p *Pruner) SetObserver(obs PrunerObserver) {
 
 func (p *Pruner) OnStart() error {
 	go p.pruneIndexesRoutine()
```
jmalicevic, line 123:
```suggestion
```

jmalicevic: I can't suggest this, but basically this should be put in line 128 below (under `p.pruneABCIResponses()` ) , because it should be done only if the data companion is enabled.

state/store_test.go
```diff
@@ -329,24 +333,16 @@ func TestSaveRetainHeight(t *testing.T) {
...
+	defer callbackF()
+	pruner := sm.NewPruner(stateStore, bs, blockIndexer, txIndexer, log.TestingLogger())
 	minHeight := pruner.FindMinRetainHeight()
 	require.Equal(t, int64(0), minHeight)
 
 	err := stateStore.SaveApplicationRetainHeight(10)
 	require.NoError(t, err)
 	minHeight = pruner.FindMinRetainHeight()
-	require.Equal(t, int64(0), minHeight)
+	require.Equal(t, int64(10), minHeight)
```
jmalicevic, line 345: This should not be changed? How is the test passing after this change? By default the data companion retain height is set o nstartup to 0, which means no pruning will occur and it should not return 10 as the minimum retain height.

state/store_test.go
```diff
@@ -329,24 +333,16 @@ func TestSaveRetainHeight(t *testing.T) {
...
-	stateStore := sm.NewStore(stateDB, sm.StoreOptions{
-		DiscardABCIResponses: false,
-	})
-	require.NoError(t, initStateStoreRetainHeights(stateStore, 0, 0, 0))
-	pruner := sm.NewPruner(
-		stateStore,
-		nil,
-		log.TestingLogger(),
-		sm.WithPrunerCompanionEnabled(),
-	)
```
jmalicevic, line 342: Why are you deleting these lines? They are not related to the changes in your PR and should probably be left as they are.

jmalicevic: The reason is that,by deleting this you have accidentally deleted the `initStatStoreRetainHeights` part which is probably why you had to replace the condition below to the height being 10 instead of 0

state/indexer/block.go
werty144: Done

.changelog/unreleased/improvements/169-tx_index-pruning.md
werty144: Done

state/txindex/indexer.go
werty144: Done

state/indexer/block/kv/kv.go
```diff
@@ -82,8 +87,138 @@ func (idx *BlockerIndexer) Index(bh types.EventDataNewBlockEvents) error {
 	if err := idx.indexEvents(batch, bh.Events, height); err != nil {
 		return fmt.Errorf("failed to index FinalizeBlock events: %w", err)
 	}
+	err = batch.WriteSync()
```
jmalicevic, line 90: You can revert this change because err will always have the value of `batch.WriteSync()` at this point (if it was not nil in line 87, it would have returned.

state/txindex/kv/kv.go
```diff
@@ -40,6 +47,101 @@ type TxIndex struct {
...
+			}
+			errWriteBatch := batch.WriteSync()
+			if errWriteBatch != nil {
+				return 0, lastRetainHeight, fmt.Errorf("error writing batch '%v' while handling result deletion error '%v'", errWriteBatch, errDeleteResult)
+			}
+			return result.Height - lastRetainHeight, result.Height, errDeleteResult
+		}
+	}
+	maxHeightPruned := results[len(results)-1].Height
+	newRetainedHeight := maxHeightPruned + 1
```
jmalicevic, line 96:
```suggestion
	newRetainedHeight := results[len(results)-1].Height + 1
```

state/txindex/kv/kv.go
```diff
@@ -40,6 +47,101 @@ type TxIndex struct {
...
+			errSetLastRetainHeight := txi.setIndexerRetainHeight(result.Height, batch)
+			if errSetLastRetainHeight != nil {
+				return 0, lastRetainHeight, fmt.Errorf("error setting last retain height '%v' while handling result deletion error '%v'", errSetLastRetainHeight, errDeleteResult)
+			}
+			errWriteBatch := batch.WriteSync()
+			if errWriteBatch != nil {
+				return 0, lastRetainHeight, fmt.Errorf("error writing batch '%v' while handling result deletion error '%v'", errWriteBatch, errDeleteResult)
+			}
+			return result.Height - lastRetainHeight, result.Height, errDeleteResult
+		}
```
jmalicevic, line 93: Maybe add a section here to flush the batch every X results. See https://github.com/cometbft/cometbft/blob/6a7dfb40d0302de5cdbca51cc4daede74b6d44c2/store/store.go#L312 which is called by https://github.com/cometbft/cometbft/blob/6a7dfb40d0302de5cdbca51cc4daede74b6d44c2/store/store.go#L379

state/txindex/mocks/indexer_service.go
```diff
@@ -0,0 +1,51 @@
+package mocks
```
jmalicevic, line 1: Did you add this mock file? If yes, why? Is it used anywhere?

state/indexer/block/kv/util.go
```diff
@@ -51,6 +51,30 @@ func heightKey(height int64) ([]byte, error) {
 	)
 }
 
+func (idx *BlockerIndexer) getEventKeys() (map[int64][][]byte, error) {
```
jmalicevic, line 54: My concern is this is effectively loading all the indexer keys into memory , which is potentiall 100s of GBs.

state/indexer/block/kv/kv_test.go
```diff
@@ -1,23 +1,96 @@
-package kv_test
+package kv
```
jmalicevic, line 1:
```suggestion
package kv_test
```

we cannot change package names.

state/indexer/block/kv/util.go
```diff
@@ -51,6 +51,51 @@ func heightKey(height int64) ([]byte, error) {
...
+		possibleHeight       int64
+	)
+	remaining, err := orderedcode.Parse(string(key), &blockHeightKeyPrefix, &possibleHeight)
+	if err != nil {
+		panic(err)
+	}
+	if len(remaining) == 0 && blockHeightKeyPrefix == types.BlockHeightKey {
+		return possibleHeight
+	}
+	panic(fmt.Errorf("key must be either heightKey or eventKey"))
```
jmalicevic, line 96: Are you not storing also the retain heights in the stores? This means that the iterator will hit them and it will cause a panic?

werty144: This function is meant to be called only with either heightKey or eventKey. I tried to emphasize it with [this](https://github.com/cometbft/cometbft/blob/7810d2654880cfcaf532fae234b82fc041b67e3c/state/indexer/block/kv/util.go#L79) annotation

This is ensured [here](https://github.com/cometbft/cometbft/blob/7810d2654880cfcaf532fae234b82fc041b67e3c/state/indexer/block/kv/kv.go#L146)

state/indexer/block/kv/kv_test.go
```diff
@@ -511,56 +512,6 @@ func TestBigInt(t *testing.T) {
 	}
 }
 
-func TestKeyArray(t *testing.T) {
```
jmalicevic: Do you have to delete this test with the renaming? Can you move all the unused functions from `util` back into this file?

state/indexer/block/kv/util.go
jmalicevic: I am not sure I see how this call enforces that it is called only for a key containing height? the iterator should get all keys unless you are somehow checking that the key has a height?

state/indexer/block.go
```diff
@@ -24,4 +24,10 @@ type BlockIndexer interface {
 	Search(ctx context.Context, q *query.Query) ([]int64, error)
 
 	SetLogger(l log.Logger)
+
+	Prune(retainHeight int64) (int64, int64, error)
```
sergio-mena, line 28: Since `BlockIndexer` is a public interface, I think we should add an extra changelog in the "breaking changes" directory to point at these changes

state/indexer/block/kv/export_testing.go
```diff
@@ -0,0 +1,5 @@
+package kv
+
+func GetKeys(indexer BlockerIndexer) [][]byte {
```
sergio-mena, line 3: Is this a function only used for unit tests? If that is the case, it is currently also being included when building the `cometbft` binary. If you want it to only be present when building UT binaries (i.e., when running `go test`) so should, either rename this file to have suffix "_test.go", or move this code to one of such files.

state/indexer/block/kv/util.go
jmalicevic: Clarified in person.

state/indexer/block/kv/kv.go
```diff
@@ -82,10 +87,138 @@ func (idx *BlockerIndexer) Index(bh types.EventDataNewBlockEvents) error {
...
+	}
+	if lastRetainHeight == 0 {
+		lastRetainHeight = 1
+	}
+
+	batch := idx.store.NewBatch()
+	defer func(batch dbm.Batch) {
+		err := batch.Close()
+		if err != nil {
+			idx.log.Error(fmt.Sprintf("Error when closing pruning batch: %v", err))
```
sergio-mena, line 124: If you replaced the batch (at least once) in line 160, won't you be trying to close the same batch twice? If that is the case, I don't think you should be producing an error here... it'll be misleading to the operator.

Same comment in the other deferred function below

jmalicevic (approved):
Thank you Anton, great job!

state/indexer/block/kv/kv.go
```diff
@@ -82,10 +87,138 @@ func (idx *BlockerIndexer) Index(bh types.EventDataNewBlockEvents) error {
...
+			affectedHeights[height] = struct{}{}
+			deleted++
+		}
+		if deleted%1000 == 0 && deleted != 0 {
+			err = flush(batch)
+			if err != nil {
+				return 0, lastRetainHeight, err
+			}
+			batch = idx.store.NewBatch()
+			defer func(batch dbm.Batch) {
```
sergio-mena, line 161: Minor: would it be worth defining another function, in the same way `flush` is defined, to contain code in lines 161-166, and 121-125? (this would remove code duplication)

state/indexer/block/kv/util.go
```diff
@@ -51,6 +51,51 @@ func heightKey(height int64) ([]byte, error) {
...
+	if err != nil {
+		return false
+	}
+	if len(remaining) == 0 &&
+		blockHeightKeyPrefix == types.BlockHeightKey &&
+		left <= possibleHeight &&
+		possibleHeight < right {
+		return true
+	}
+	return false
```
sergio-mena, line 75:
```suggestion
	return len(remaining) == 0 &&
		blockHeightKeyPrefix == types.BlockHeightKey &&
		left <= possibleHeight &&
		possibleHeight < right
```
Just a suggestion, up to you

state/indexer/block/kv/util.go
```diff
@@ -51,6 +51,51 @@ func heightKey(height int64) ([]byte, error) {
...
+		return eventHeight
+	}
+
+	var (
+		blockHeightKeyPrefix string
+		possibleHeight       int64
+	)
+	remaining, err := orderedcode.Parse(string(key), &blockHeightKeyPrefix, &possibleHeight)
+	if err != nil {
+		panic(err)
```
sergio-mena, line 91: Just for my own instruction, why do we panic here after calling `orderedcode.Parse`, but we just error out above (line 67)?

state/pruner.go
```diff
@@ -210,6 +221,58 @@ func (p *Pruner) SetABCIResRetainHeight(height int64) error {
...
+		}
+		return p.txIndexer.SetRetainHeight(height)
+	}
+	if currentRetainHeight > height {
+		return ErrPrunerCannotLowerRetainHeight
+	}
+	if err := p.txIndexer.SetRetainHeight(height); err != nil {
+		return err
+	}
+	// TODO call metrics
```
sergio-mena, line 246: Is this something we want to do in this PR?

state/txindex/indexer.go
```diff
@@ -31,6 +31,12 @@ type TxIndexer interface {
 
 	//Set Logger
 	SetLogger(l log.Logger)
+
+	Prune(retainHeight int64) (int64, int64, error)
```
sergio-mena, line 35: Same comment regarding the addition of a changelog in "breaking changes"

state/txindex/indexer_service.go
```diff
@@ -33,7 +33,11 @@ func NewIndexerService(
 	terminateOnError bool,
 ) *IndexerService {
 
-	is := &IndexerService{txIdxr: txIdxr, blockIdxr: blockIdxr, eventBus: eventBus, terminateOnError: terminateOnError}
+	is := &IndexerService{
+		txIdxr:           txIdxr,
+		blockIdxr:        blockIdxr,
+		eventBus:         eventBus,
+		terminateOnError: terminateOnError}
```
sergio-mena, line 40:
```suggestion
		terminateOnError: terminateOnError
	}
```

state/txindex/kv/export_testing.go
```diff
@@ -0,0 +1,5 @@
+package kv
+
+func GetKeys(indexer *TxIndex) [][]byte {
```
sergio-mena, line 3: Same comment regarding this code being part of the production binary. OK if it is intended, else we are introducing a public function to access a private one in prod code.

state/txindex/kv/kv.go
```diff
@@ -40,6 +47,138 @@ type TxIndex struct {
...
+
+	lastRetainHeight, err := txi.getIndexerRetainHeight()
+	if err != nil {
+		return 0, 0, fmt.Errorf("failed to look up last block indexer retain height: %w", err)
+	}
+	if lastRetainHeight == 0 {
+		lastRetainHeight = 1
+	}
+
+	ctx := context.Background()
```
sergio-mena, line 64: It would be good to add an issue to github to remember to thread this context up in the call chain.

state/txindex/kv/kv.go
```diff
@@ -40,6 +47,138 @@ type TxIndex struct {
...
+	}
+	if len(results) == 0 {
+		return 0, lastRetainHeight, nil
+	}
+
+	batch := txi.store.NewBatch()
+	defer func(batch dbm.Batch) {
+		err := batch.Close()
+		if err != nil {
+			txi.log.Error(fmt.Sprintf("Error when closing pruning batch: %v", err))
```
sergio-mena, line 78: Same comment on getting this error, even when everything went fine but we had more than 1000 pruned

state/txindex/kv/kv.go
```diff
@@ -40,6 +47,138 @@ type TxIndex struct {
...
+		return 0, 0, fmt.Errorf("failed to look up last block indexer retain height: %w", err)
+	}
+	if lastRetainHeight == 0 {
+		lastRetainHeight = 1
+	}
+
+	ctx := context.Background()
+	results, err := txi.Search(ctx, query.MustCompile(
+		fmt.Sprintf("tx.height < %d AND tx.height >= %d", retainHeight, lastRetainHeight)))
+	if err != nil {
```
sergio-mena, line 67: Question to @jmalicevic. Are we sure we're not gonna hit any error in line 66? (it's a "search" query so I imagine a lot of code is gonna be hit there)

state/txindex/kv/kv.go
```diff
@@ -40,6 +47,138 @@ type TxIndex struct {
...
+		// flush every 1000 blocks to avoid batches becoming too large
+		if pruned%1000 == 0 && pruned > 0 {
+			err := flush(batch)
+			if err != nil {
+				return numHeightsPersistentlyPruned, currentPersistentlyRetainedHeight, err
+			}
+			numHeightsPersistentlyPruned = numHeightsBatchPruned
+			currentPersistentlyRetainedHeight = currentBatchRetainedHeight
+			batch = txi.store.NewBatch()
+			defer func(batch dbm.Batch) {
```
sergio-mena, line 127: Same comment on unifying both (identical) defer functions into one

state/txindex/kv/kv.go
```diff
@@ -40,6 +47,138 @@ type TxIndex struct {
...
+		err := batch.Close()
+		if err != nil {
+			txi.log.Error(fmt.Sprintf("Error when closing pruning batch: %v", err))
+		}
+	}(batch)
+	pruned := uint64(0)
+	flush := func(batch dbm.Batch) error {
+		err := batch.WriteSync()
+		if err != nil {
+			return fmt.Errorf("failed to flush pruning batch %w", err)
```
sergio-mena, line 85: could differentiate the text of logs done by block indexer and those done by tx indexer?
E.g. `"failed to flush pruning batch %w"` --> `"failed to flush tx pruning batch %w"`

state/indexer/block/kv/kv.go
werty144: From what I understand, `batch.Close()` is idempotent, meaning one can call it multiple times. So closing a closed batch would not lead to an error, though I agree that this is far from being clear.

I didn't come up with this approach myself, but rather I took from [here](https://github.com/cometbft/cometbft/blob/6a7dfb40d0302de5cdbca51cc4daede74b6d44c2/store/store.go#L312)

state/indexer/block/kv/util.go
werty144: [keyBelongsToHeightRange](https://github.com/cometbft/cometbft/blob/18269cff546a7a3b67fe27179c8c6168f21582f6/state/indexer/block/kv/util.go#L54C6-L54C29) function requires no guarantees on the key passed to it. Its goal is to check whether the given key belongs to a range of heights. That might not be the case for two reasons: the key belongs to the height out of the specified range, or the key doesn't belong to any height, meaning it's neither `heightKey` nor `eventKey`. In the latter case, we might fail to parse the key (the bulk of random bytes) into `heightKey`. That's nothing critical, but simply means that it is not the key we are interested in. 

Whereas, [getHeightFromKey](https://github.com/cometbft/cometbft/blob/18269cff546a7a3b67fe27179c8c6168f21582f6/state/indexer/block/kv/util.go#L78C6-L78C22) [requires](https://github.com/cometbft/cometbft/blob/18269cff546a7a3b67fe27179c8c6168f21582f6/state/indexer/block/kv/util.go#L79) its argument key to be the key with height (either `heightKey` or `eventKey`).  So in case it's not, we panic. 

I wrote comments for these to functions to make it more clear.

state/indexer/block.go
werty144: Done: [.changelog/unreleased/breaking-changes/1176-blockindexer-pruning.md](https://github.com/cometbft/cometbft/blob/5b2522ff35ac86a28c703071c07bf72f9bd49124/.changelog/unreleased/breaking-changes/1176-blockindexer-pruning.md)

state/txindex/indexer.go
werty144: Done: [.changelog/unreleased/breaking-changes/1176-blockindexer-pruning.md](https://github.com/cometbft/cometbft/blob/5b2522ff35ac86a28c703071c07bf72f9bd49124/.changelog/unreleased/breaking-changes/1176-blockindexer-pruning.md)

state/indexer/block/kv/export_testing.go
werty144: Ahh, this is a tough one. It is used outside `kv` and `kv_test` packages, namely in `package state_test` in [state/pruner_test.go](https://github.com/cometbft/cometbft/blob/5b2522ff35ac86a28c703071c07bf72f9bd49124/state/pruner_test.go).

So renaming the file to "*_test.go" would make it invisible there, and it's not possible to move it outside `kv` package since it accesses private function `getKeys`.

state/indexer/block/kv/util.go
werty144: Done

state/txindex/kv/export_testing.go
werty144: Same answer to [this](https://github.com/cometbft/cometbft/pull/1176#discussion_r1318581213)

state/indexer/block/kv/kv.go
sergio-mena: I see, in that case, if you're sure that closing a closed batch returns `nil`, I'm OK with this approach.

state/txindex/kv/kv.go
werty144: I am not sure that I understood what you mean well enough to make it myself

state/txindex/kv/kv.go
werty144: Done

state/txindex/kv/kv.go
werty144: Done

.changelog/unreleased/breaking-changes/1176-blockindexer-pruning.md
```diff
@@ -0,0 +1,2 @@
+- `[state/indexer/block]` BlockIndexer now has additional method `Prune(retainHeight int64) (int64, int64, error)` ([\#1176](https://github.com/cometbft/cometbft/pull/1176))
+- `[state/txindex]` TxIndexer now has additional method `Prune(retainHeight int64) (int64, int64, error)` ([\#1176](https://github.com/cometbft/cometbft/pull/1176))
```
sergio-mena, line 2: Can you also mention the other to added methods? No need to mention their entire signature

state/indexer/block/kv/export_testing.go
sergio-mena: I see. Quite unfortunate, but I guess I can't think of any option than what you've done.

state/txindex/kv/kv.go
sergio-mena: Not a big deal... this will anyway have to be done as part of the "big" context threading task.

.changelog/unreleased/breaking-changes/1176-blockindexer-pruning.md
werty144: Done

state/txindex/kv/kv.go
jmalicevic: If all goes well there should not be an error, just an empty result returned. An error should happen only in exceptional cases... The only thing I am not 100% sure of is whether we should panic here as this is not critical consensus code , so maybe it is better to log an error?

state/pruner.go
jmalicevic: It will be done in a follow up PR.

state/txindex/kv/kv.go
sergio-mena: IMO, better return an error, if possible

jmalicevic: Yup, its fixed now.

PR #1182, opened 2023-07-31
ci: Remove Mergify automerge

thanethomson:
Mergify's automerge functionality clashes with merge queues and automatic Dependabot updates sometimes, which is particularly annoying. We don't need it on `main` any more.

---

#### PR checklist

- [ ] Tests written/updated


sergio-mena (approved):
MSTM (makes sense to me)

PR #1183, opened 2023-07-31
ADR 101: Add observer for pruner

thanethomson:
A suggestion for #1150 that is based on my recommendations in #1164 to potentially make our tests more deterministic.

Should be squash-merged into the base branch.

---

#### PR checklist

- [ ] Tests written/updated


state/pruner.go
```diff
@@ -231,19 +231,23 @@ func (p *Pruner) pruneABCIResToRetainHeight(lastABCIResPrunedHeight int64) int64
...
-		// lastPrunedHeight is the last height up until which we pruned successfully
-		// In case of an error it will be 0, but then it will also be ignored.
-		numPruned, lastPrunedHeight, err := p.stateStore.PruneABCIResponses(abciResRetainHeight)
-		if err != nil {
-			p.logger.Error("Failed to prune ABCI responses", "err", err, "abciResRetainHeight", abciResRetainHeight)
-		} else {
-			p.logger.Info("Pruned ABCI responses", "height", numPruned)
-			lastABCIResPrunedHeight = lastPrunedHeight
-		}
+		return lastABCIResPrunedHeight
```
jmalicevic, line 234: I added these changes to my branch so you can merge it again .. that way the observer branch contains only changes related to it.

jmalicevic:
We talked at some point about introducing metrics about how many blocks, abci results etc are pruned. This PR gives us the same information. Would we still need to implement those or we could reuse this code?

thanethomson:
> We talked at some point about introducing metrics about how many blocks, abci results etc are pruned. This PR gives us the same information. Would we still need to implement those or we could reuse this code?

Metrics are for operators, whereas this is an approach for consensus engine developers (such as us) to "hook into" the execution of certain processes. One could use these hooks to implement metrics, but that's not the main value this approach aims to provide. Our pruning metrics should be wired into the `Pruner` itself (perhaps in a follow-up PR to #1150).

See the usage from the test in this PR for an example of how this benefits testing in replacing potentially non-deterministic `time.Sleep` operations with hooks into specific events, and how it does so without having to expose private reactor/service methods exclusively for use in testing.

thanethomson:
I've now rebased this on top of the changes in #1201 such that we can clearly separate them.

jmalicevic (approved):

PR #1184, opened 2023-07-31
rpc: Improve abci_query parameter documentation

gibson042:
No issue or changelog entry because this PR is not non-trivial work (https://github.com/cometbft/cometbft/blob/main/CONTRIBUTING.md#github-issues ).


rpc/openapi/openapi.yaml
```diff
@@ -1057,29 +1057,39 @@ paths:
...
+            example: '"/store/foo"'
         - in: query
           name: data
-          description: Data
-          required: true
+          description: |
+            Request parameters for the application to interpret analogously to a
+            [URI query component](https://www.rfc-editor.org/rfc/rfc3986#section-3.4),
+            expressed as hexadecimal-serialized bytes.
+          required: false
```
jmalicevic, line 1078: Should we be changing the requirement here from `true` to `false`?

gibson042: This aligns the OpenAPI spec with implementation reality and with [other documentation](https://docs.cometbft.com/v0.37/spec/abci/abci++_methods#query) ("_`data`… Can be used with or in lieu of Path_" and "_`path`… Can be used with or in lieu of data_"). AFAICT, neither field has ever been required—there are explicit examples of kvstore app `path`-free queries like `curl -s 'localhost:26657/abci_query?data="abcd"'` [throughout docs.cometbft.com](https://github.com/search?q=repo%3Acometbft%2Fcometbft+%2Fabci_query%5B%3F%5Ddata%3D%2F&type=code), and you can run live `data`-free queries against the https://docs.cometbft.com/main/rpc/ default server:
```sh
$ curl -sS 'https://rpc.cosmos.directory/cosmoshub/abci_query?path="/app/version"' |
  jq -r '.result.response.value | @base64d'
v10.0.2
```

In practice, I suspect that most _real_ applications will require `path` but not `data` (cf. [cosmos-sdk `Query`](https://github.com/cosmos/cosmos-sdk/blob/e2037f7696fed4fdd4bc076f9e7053fe8178a881/baseapp/abci.go#L551-L554)), although committing to that in CometBFT would require changing the kvstore app.

gibson042:
I don't think this is stale; it just needs to be reviewed and merged.

gibson042:
> I don't think this is stale; it just needs to be reviewed and merged.

rpc/openapi/openapi.yaml
jmalicevic: Hi, sorry it took us a while to get back. You are right that this is in fact an application level concern (whether it is required or not), and from CometBFT's RPC these parameters are not required.

jmalicevic (approved):
Thank you for work and patience!

sergio-mena (approved):
🙏

Issue #1140, opened 2023-07-18
Export errors for easier programmatic use


thanethomson:
Comet currently doesn't export many of its errors, which hinders programmatic use (i.e. by consensus engine developers). The only alternative for them is to detect specific errors by their string representations, which is prone to breakage even in patch releases of our software.

We should, package by package, ensure that all possible errors returned by publicly exported methods are defined programmatically ([example](https://github.com/cometbft/cometbft/blob/90a627f21f2a6d41008acf57a9eb0b39b3e80c75/light/provider/errors.go#L8-L19)) and publicly exported too.

Per package (check it off if its errors are already all publicly exported or this is not applicable):

- [ ] `abci`
- [ ] `blocksync`
- [ ] `config`
- [ ] `consensus`
- [ ] `crypto`
- [ ] `evidence`
- [ ] `libs`
- [ ] `light`
- [ ] `mempool`
- [ ] `node`
- [ ] `p2p`
- [ ] `privval`
- [ ] `proxy`
- [ ] `rpc`
- [ ] `state`
- [ ] `statesync`
- [ ] `store`
- [ ] `types`
- [ ] `version`

thanethomson:
> hey @thanethomson , is it free to pick up?

It is! Are there particular packages you'd like to tackle, or would you like to handle the whole lot? If you'd like to tackle the issue in its entirety, I'd still recommend submitting separate PRs for each package to make review quicker.

pysel:
sounds awesome! not sure if I will be able to tackle the whole thing, but I will go in the sequential order packages are listed in the issue. I will "reserve" `abci`, `blocksync` and `config` for now, just in case anyone else wants to take over some packages of the list, to avoid working on the same thing

thanethomson:
Thanks @pysel! I'll assign you to this issue for now, given the above. If anyone else wants to take on other packages, please let us know on this thread.

4ever9:
hey @thanethomson, is it free to pick up 'mempool' package.

thanethomson:
Hey @allnil and @4ever9! Feel free to pick up any of the packages from the list in the issue description that don't have PRs associated with them (currently most of them).

Tikaryan:
Hey @thanethomson is `p2p` package is free to to pickup?.

Halimao:
I'm working on`light`, `node`, `privval` package, just to avoid someone gonna working on the same thing😀

Halimao:
@thanethomson Hi sir,

[...]

PR #1190, opened 2023-08-01
config: export errors

pysel:
---

Contributes to: #1140 
#### Changes

- Exports errors in `config`. More context: #1140 
- Adds a package to `types` specifically for errors


config/errors.go
```diff
@@ -0,0 +1,56 @@
+package config
+
+import (
+	"errors"
+	"fmt"
+)
+
+var (
+	// ErrNotEnoughRpcServers is returned if the number of rpc servers is less than two
+	ErrNotEnoughRPCServers = errors.New("at least two rpc_servers entries is required")
```
sergio-mena, line 10:
```suggestion
	ErrNotEnoughRPCServers = errors.New("at least two rpc_servers entries are required")
```

config/config.go
```diff
@@ -881,12 +887,12 @@ func TestBlockSyncConfig() *BlockSyncConfig {
 // ValidateBasic performs basic validation.
 func (cfg *BlockSyncConfig) ValidateBasic() error {
 	switch cfg.Version {
-	case "v0":
+	case v0:
 		return nil
-	case "v1", "v2":
-		return fmt.Errorf("blocksync version %s has been deprecated. Please use v0 instead", cfg.Version)
+	case v1, v2:
+		return ErrDeprecatedBlocksyncVersion{Version: cfg.Version, Allowed: AllowedBlockSyncConfigVersions}
```
sergio-mena, line 893: You are defining `AllowedBlockSyncConfigVersions` (for the error itself), but the _switch_ doesn't use it. So we will have to be careful in the future when modifying the blocksync versions to make sure that what shows in the _switch_ is consistent with the definition of `AllowedBlockSyncConfigVersions`.
My suggestion,

* either you get rid of `AllowedBlockSyncConfigVersions` and use `v0` directly hard-coded in line 893 (anyway it is hard-coded in the _case_ in line 890, isn't it?),
* or you modify the switch to use the information contained in `AllowedBlockSyncConfigVersions`

I'd personally lean toward the first (simpler) option.

config/errors.go
```diff
@@ -0,0 +1,56 @@
+package config
+
+import (
+	"errors"
+	"fmt"
+)
+
+var (
+	// ErrNotEnoughRpcServers is returned if the number of rpc servers is less than two
```
sergio-mena, line 9: this is not needed, the error message is self explanatoriy

config/errors.go
```diff
@@ -0,0 +1,56 @@
...
+	ErrNotEnoughRPCServers = errors.New("at least two rpc_servers entries is required")
+
+	// ErrEmptyRpcServerEntry is returned when an empty string corresponding to RPC entry is found during validate basic
+	ErrEmptyRPCServerEntry = errors.New("found empty rpc_servers entry")
+
+	// ErrInsufficientDiscoveryTime is returned when snapshot discovery time is less than 5 seconds
```
sergio-mena, line 15: Not needed

config/errors.go
```diff
@@ -0,0 +1,56 @@
...
+	ErrInsufficientDiscoveryTime = errors.New("discovery time must be at least five seconds")
```
sergio-mena, line 16:
```suggestion
	ErrInsufficientDiscoveryTime = errors.New("snapshot discovery time must be at least five seconds")
```

config/errors.go
```diff
@@ -0,0 +1,56 @@
...
+
+	// ErrInsufficientChunkRequestTimeout is returned when timeout for re-requesting a chunk is less than 5 seconds
```
sergio-mena, line 18: > timeout for re-requesting a chunk is less than 5 seconds

I would use that as the message for the error and get rid of this comment

config/errors.go
```diff
@@ -0,0 +1,56 @@
...
+	ErrInsufficientChunkRequestTimeout = errors.New("chunk_request_timeout must be at least 5 seconds")
+
+	ErrSubscriptionBufferSizeInvalid = fmt.Errorf("experimental_subscription_buffer_size must be >= %d", minSubscriptionBufferSize)
+
+	// ErrUnknownLogFormat is returned if provided log_format is invalid during validate basic
```
sergio-mena, line 23: Not needed

config/config.go
pysel: makes sense, thanks for advice! gonna lean towards the first one as well

config/errors.go
```diff
@@ -0,0 +1,47 @@
+package config
+
+import (
+	"errors"
+	"fmt"
+)
+
+var (
+	// ErrEmptyRpcServerEntry is returned when an empty string corresponding to RPC entry is found during validate basic
```
pysel, line 9:
```suggestion
```

config/errors.go
```diff
@@ -0,0 +1,46 @@
...
+	"fmt"
+)
+
+var (
+	ErrEmptyRPCServerEntry             = errors.New("found empty rpc_servers entry")
+	ErrNotEnoughRPCServers             = errors.New("at least two rpc_servers entries are required")
+	ErrInsufficientDiscoveryTime       = errors.New("snapshot discovery time must be at least five seconds")
+	ErrInsufficientChunkRequestTimeout = errors.New("timeout for re-requesting a chunk (chunk_request_timeout) is less than 5 seconds")
+	ErrSubscriptionBufferSizeInvalid   = fmt.Errorf("experimental_subscription_buffer_size must be >= %d", minSubscriptionBufferSize)
+	ErrUnknownLogFormat                = errors.New("unknown log_format (must be 'plain' or 'json')")
```
pysel, line 14:
```suggestion
	ErrUnknownLogFormat                = errors.New("unknown log_format (must be 'plain' or 'json')")
	ErrSubscriptionBufferSizeInvalid   = fmt.Errorf("experimental_subscription_buffer_size must be >= %d", minSubscriptionBufferSize)
```

pysel:
hey @sergio-mena! thanks for review, addressed, lmk if it is good to merge now please

sergio-mena (approved):
LGTM
Thanks @pysel for this PR! Much appreciated!

PR #1191, opened 2023-08-01
mempool: Store peer ids as p2p.ID instead of uint16 (remove mempoolIDs)

### LINKED ISSUE #1146: mempool: Remove list of peer ids in reactor
### Summary

The data structure `mempoolIDs` in the mempool reactor is used to assign a unique identifier of type `uint16` to each peer (identified by `p2p.ID`). These ids are only known locally to the node, and only used to keep track of the senders of each transaction.

PR https://github.com/tendermint/tendermint/pull/2778 introduced this data structure, where it says: "Instead of using the 20byte peer ID, it instead uses a local map from peerID to uint16 counter, so every peer adds 2 bytes. (Word aligned to probably make it 8 bytes)".

### Proposal

Remove `mempoolIDs` that is only used to save 12 bytes per connected peer.

See original discussion: https://github.com/cometbft/cometbft/pull/1043#discussion_r1251606701

hvanz:
Closes #1146 

---

#### PR checklist

- [x] Tests written/updated


.changelog/unreleased/breaking-changes/1146-mempool-remove-ids.md
```diff
@@ -0,0 +1,3 @@
+- `[mempool]` Remove `mempoolIDs` for internally storing peer ids as `p2p.ID`
+  instead of `uint16`.
+  ([\#1146](https://github.com/cometbft/cometbft/pull/1146))
```
thanethomson, line 3: I think it's mostly important just to mention which public Go APIs (`mempool.Reactor.InitPeer` and `mempool.Reactor.RemovePeer`) have been removed.

hvanz: They haven't been removed. They were redundant because now they are both empty, which is equal to the default behaviour provided by `BaseReactor`.

mempool/reactor.go
```diff
@@ -40,20 +37,13 @@ func NewReactor(config *cfg.MempoolConfig, mempool *CListMempool) *Reactor {
...
 	memR.BaseReactor = *p2p.NewBaseReactor("Mempool", memR)
 	memR.mempool.SetTxRemovedCallback(func(txKey types.TxKey) { memR.removeSenders(txKey) })
 	return memR
 }
 
-// InitPeer implements Reactor by creating a state for the peer.
-func (memR *Reactor) InitPeer(peer p2p.Peer) p2p.Peer {
-	memR.ids.ReserveForPeer(peer)
-	return peer
-}
```
thanethomson, line 55: This function is publicly exported, so its removal needs to be documented. Doesn't matter if it's redundant now.

mempool/reactor.go
```diff
@@ -97,12 +87,6 @@ func (memR *Reactor) AddPeer(peer p2p.Peer) {
 	}
 }
 
-// RemovePeer implements Reactor.
-func (memR *Reactor) RemovePeer(peer p2p.Peer, _ interface{}) {
-	memR.ids.Reclaim(peer)
-	// broadcast routine checks if peer is gone and returns
-}
```
thanethomson, line 104: Same here, this function is publicly exported, so its removal needs to be documented.

PR #1139, opened 2023-07-17
E2E: Add support for fast prototyping

otrack:
This PR addresses issue #1059.

It is built atop #1043.

**Summary:**
This PR adds a support for fast prototyping to the E2E test suite. Please read the REAME.md located [here](https://github.com/otrack/cometbft/blob/pierre/fast-prototyping-1059/test/e2e/fast-prototyping/README.md)  to start using this code.

**Progress:**
To date, the following goals are in full (or partially) done:
- [x] Speed-up docker image construction
- [x] Add new small docker image (new make target)
- [x] Include Prometheus server support to docker compose
- [ ] Provide manifests for medium-scale experiments (#1082)
- [x] (_partial)_ Allow customizing manifests to choose reactors (#1083)
- [x] (_partial)_ Provide scripts to collect and process metrics (#1084)
- [x] _(partial)_ Add full support for e2e tests in fast-prototyping mode (#1085))

---

#### PR checklist

- [ ] Tests written/updated
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments

Issue #1089, opened 2023-07-05
Evaluate solutions on QA infra


lasarojc:
- [ ] Run the prototypes on DO for the different experiment configurations. 
- [ ] We collect the generated metrics. 
- [ ] We generate a report that includes results for each run and a comparison of the results.

lasarojc:
Closing due to too broad a scope.
Will open issues for individual approaches.

Issue #1059, opened 2023-06-30
Add support for fast prototyping


lasarojc:
## Target audience

Consensus developers

## Problem definition

The e2e testbed demands large amounts of resources to run medium/large scale experiments (16-128 nodes).
This comes from the fact that the testbed runs CometBFT in full on the testbed nodes.
Furthermore, there is no support for fast prototyping.
For instance, building a Docker image is slow, and the image is heavy (several GBs).

## Upside

Run small-scale experiments on a laptop / server-class machine will speed up prototyping and testing fixes and/or improvements.

## Downside

Requires mocking reactors.

The system will be a slimmed down version of what actually runs on the nodes (less reactors).
Hence, there is a small chance that improvements in this simplified environment do not translate into improvements in real-world setups.
This is an inherent (and standard) risk to such an approach.
To mitigate it, the developper needs to understand beforehand the impact of mocking one (or more) reactors.

```[tasklist]
## Original Goals
- [x] Speed-up docker image construction
- [x] Add new small docker image (new make target)
- [x] Include Prometheus server support to docker compose
- [ ] https://github.com/cometbft/cometbft/issues/1082
- [ ] https://github.com/cometbft/cometbft/issues/1083
- [ ] https://github.com/cometbft/cometbft/issues/1084
- [ ] https://github.com/cometbft/cometbft/issues/1085
```

```[tasklist]
## Integration Goals (provide PRs for the following)
- [ ] Faster Docker builds (possibly leveraging go's cross-compilation simplicity)
- [ ] Plugging custom reactors (+ the "demo" custom reactor for the mempool)
- [ ] Interactive reporting/visualization tool
```

```[tasklist]
## Stretch goals
- [ ] Add Kubernetes support to e2e [2w]
- [ ] Integrate the scripting into the e2e test.
```

## Definition of done
- The e2e testbed includes a fast-prototyping mode. 
In this mode, the docker image is tailored to its bare minimum, and building it (along with the generator and runner) is immediate.
- The fast-prototyping mode allows to mock a reactor, such as consensus. 
- There is a new custom benchmark.
This benchmark starts a Prometheus node which targets all the validators.
At the end of a run, the benchmark return aggregated (customizable) key metrics from the data collected in Prometheus.
```

lasarojc:
The excellent work done in #1139 resulted in this [branch](https://github.com/cometbft/cometbft/tree/pierre/fast-prototyping-1059)

[...]

PR #1192, opened 2023-08-01
Porting fast-docker image from #1089

lasarojc:
Changes done in the context of #1059 resulted in changes in #1139 and branch `pierre/fast-prototyping-1059`
This PR isolates the changes related only building a lean image for fast prototyping.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/test/e2e/Makefile
+++ b/test/e2e/Makefile
@@ -16,12 +16,21 @@ docker-debug:
 		--tag cometbft/e2e-node:local-version \
 		-f docker/Dockerfile.debug ../..
 
+docker-fast:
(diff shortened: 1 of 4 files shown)


thanethomson (approved):
So is this effectively just building Comet with a smaller base image?

sergio-mena:
If this image makes existing e2e test pass, I'd try to keep this one and remove the existing one. Less maintenance burden.
The ~2GB images we have in e2e is something that has always been bothering me but never had spare cycles to improve the situation. This could be an opportunity to throw away the 2GB images we have.

lasarojc:
I've opened #1204 to track this idea.

PR #1201, opened 2023-08-02
ADR 101: Recommended updates to pruner service

thanethomson:
Recommendations to update #1150. Also pulls in the latest changes from the feature branch, resolving the conflict in the docs.

Since this already contains a bunch of commits that have been reviewed (i.e. the ones from the feature branch), please only review from https://github.com/cometbft/cometbft/pull/1201/commits/04a70393c8f0866557c53b8b92a90d91d465f8d2 onwards.

---

#### PR checklist

- [ ] Tests written/updated


state/pruner.go
```diff
@@ -23,11 +24,14 @@ var (
 type Pruner struct {
 	service.BaseService
 	logger log.Logger
+
+	mtx sync.Mutex
```
thanethomson, line 27: This was added after speaking to @mzabaluev today where we clearly want to serialize all requests that set any retain heights. For example, in the case where users may ignore our recommendations to only attach a single data companion and two calls are made simultaneously to set a particular retain height, allowing both calls to execute simultaneously may have undefined consequences.

state/pruner.go
```diff
@@ -178,6 +178,24 @@ func (p *Pruner) SetABCIResRetainHeight(height int64) error {
 	return err
 }
 
+// GetApplicationRetainHeight is a convenience method for accessing the
+// GetApplicationRetainHeight method of the underlying state store.
+func (p *Pruner) GetApplicationRetainHeight() (int64, error) {
```
thanethomson, line 179: These getters are added such that #1154 can access these methods without needing to access the underlying state store.

state/errors.go
```diff
@@ -51,6 +51,20 @@ type (
 	ErrNoABCIResponsesForHeight struct {
 		Height int64
 	}
+
+	ErrPrunerFailedToLoadState struct {
```
thanethomson, line 55: As per #1140.

state/pruner.go
jmalicevic: We discussed this initially and were hoping to rely on the databases' guarantees that these conflicts should be handled. But I agree this is a much cleaner approach, especially given the fact that we support multiple DB backends atm.

jmalicevic (approved):
LGTM, thanks Thane!

PR #1203, opened 2023-08-03
Forward-port: update state to prevote `nil` when proposal block does not match locked block

sergio-mena:
Contributes to #1171

Mechanical forward-port. For a thorough explanation of the contents (and context) of this PR, please check the original PR (tendermint/tendermint#6986) and original issue (tendermint/tendermint#6850).

N.B. The PR is currently open on a temporary base branch to only display the changes proper to the forward-port. When #1175 is merged, the base branch is to be changed to `main` (@BrendanChou)

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -1282,8 +1282,11 @@ func (cs *State) createProposalBlock(ctx context.Context) (*types.Block, error) 
 
 // Enter: `timeoutPropose` after entering Propose.
 // Enter: proposal block and POL is ready.
-// Prevote for LockedBlock if we're locked, or ProposalBlock if valid.
-// Otherwise vote nil.
+// If we received a valid proposal within this round and we are not locked on a block,
+// we will prevote for block.
+// Otherwise, if we receive a valid proposal that matches the block we are
+// locked on or matches a block that received a POL in a round later than our
+// locked round, prevote for the proposal, otherwise vote nil.
 func (cs *State) enterPrevote(height int64, round int32) {
 	logger := cs.Logger.With("height", height, "round", round)
 
@@ -1313,14 +1316,7 @@ func (cs *State) enterPrevote(height int64, round int32) {
 func (cs *State) defaultDoPrevote(height int64, round int32) {
 	logger := cs.Logger.With("height", height, "round", round)
 
-	// If a block is locked, prevote that.
-	if cs.LockedBlock != nil {
-		logger.Debug("prevote step; already locked on a block; prevoting locked block")
-		cs.signAddVote(cmtproto.PrevoteType, cs.LockedBlock.Hash(), cs.LockedBlockParts.Header())
-		return
-	}
-
-	// If ProposalBlock is nil, prevote nil.
+	// We did not receive a proposal within this round. (and thus executing this from a timeout)
 	if cs.ProposalBlock == nil {
 		logger.Debug("prevote step: ProposalBlock is nil")
 		cs.signAddVote(cmtproto.PrevoteType, nil, types.PartSetHeader{})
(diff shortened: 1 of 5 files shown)


lasarojc (approved):
LGTM

consensus/state.go
```diff
@@ -1363,11 +1359,67 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
...
+		AND this validator saw a 2/3 majority of the voting power prevote 'v' in round 'v_r', then we will
+		issue a prevote for 'v' in this round if 'v' is valid and either matches our locked value OR
+		'v_r' is a round greater than or equal to our current locked round.
+
+		'v_r' can be a round greater than to our current locked round if a 2/3 majority of
+		the network prevoted a value in round 'v_r' but we did not lock on it, possibly because we
+		missed the proposal in round 'v_r'.
+	*/
+	blockID, ok := cs.Votes.Prevotes(cs.Proposal.POLRound).TwoThirdsMajority()
+	if ok && cs.ProposalBlock.HashesTo(blockID.Hash) && cs.Proposal.POLRound >= 0 && cs.Proposal.POLRound < cs.Round {
```
hvanz, line 1406: Isn't `cs.Proposal.POLRound >= 0 && cs.Proposal.POLRound < cs.Round` redundant here? The function `defaultSetProposal` already validates the same for `POLRound`.

sergio-mena: Yes, but I'd keep the condition as defensive programming, since it takes a while to conclude that the checks done in `defaultSetProposal` also apply here

consensus/state.go
```diff
@@ -1363,11 +1359,67 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
...
+			logger.Debug("prevote step: ProposalBlock is valid and there is no locked block; prevoting the proposal")
+			cs.signAddVote(cmtproto.PrevoteType, cs.ProposalBlock.Hash(), cs.ProposalBlockParts.Header())
+			return
+		}
+		if cs.ProposalBlock.HashesTo(cs.LockedBlock.Hash()) {
+			logger.Debug("prevote step: ProposalBlock is valid and matches our locked block; prevoting the proposal")
+			cs.signAddVote(cmtproto.PrevoteType, cs.ProposalBlock.Hash(), cs.ProposalBlockParts.Header())
+			return
+		}
+	}
```
hvanz, line 1386: The "else" branch of the condition in line 23 of the pseudo code is missing. If `cs.Proposal.POLRound` is still -1 at this point, the line below to fetch `blockID` wouldn't make much sense.

consensus/state.go
hvanz, line 1406: Shouldn't we check that `!blockID.IsNil()`, as in most other places where we call `TwoThirdsMajority`?

consensus/state.go
```diff
@@ -1363,11 +1359,67 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
...
+
+	logger.Debug("prevote step: ProposalBlock is valid but was not our locked block or" +
```
hvanz, line 1420: What should happen at this point when `!ok` or `blockID.IsNil()`? According to the paper, the rule at line 28 should not apply, so here we should probably _not_ vote anything (and log something).

consensus/state.go
sergio-mena: Indeed. We also need to change `TestVoteSet_2_3Majority`. Bear with me

sergio-mena: ...although `HashesTo` will do the right thing if passed a `nil` block.

Nevertheless, adding your suggestion because it increases readability.

consensus/state.go
sergio-mena: My understanding of function `defaultDoPrevote` is that it will be called when we are ready to send our prevote:

* either because we detected a majority (outside `defaultDoPrevote`),
* or because we timed out

hence, we will prevote nil, rather than keep waiting.

Although I'm not 100% sure of this, I should probably examine all call sites.

consensus/state.go
```diff
@@ -1363,11 +1359,75 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
...
```
lasarojc, line 1428: given the change to address @hvanz previous comment, this case should now correspond to the timeout case only.

consensus/state.go
lasarojc: I have the same understanding.
If the timeout hit before having the full proposal, prevote nil at the beginning.
If the timeout hit before having a majority, prevote nil here.

Btw, in [state.go](https://github.com/cometbft/cometbft/blob/48de47782c2fb06d620b6825c5dc1e074a724aff/consensus/state.go#L1285-L1286), update the comment // Prevote for ~LockedBlock if we're locked, or ~ProposalBlock if valid.

sergio-mena: > Btw, in [state.go](https://github.com/cometbft/cometbft/blob/48de47782c2fb06d620b6825c5dc1e074a724aff/consensus/state.go#L1285-L1286), update the comment // Prevote for ~LockedBlock if we're locked, or ~ProposalBlock if valid.

But we're doing it already, aren't we?

hvanz (approved):
LGTM

consensus/state.go
```diff
@@ -1363,11 +1359,75 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
...
+		return
+	}
+
+	/*
+		28: upon <PROPOSAL, h_p, round_p, v, v_r> from proposer(h_p, round_p) AND 2f + 1 <PREVOTE, h_p, v_r, id(v)> while
+		step_p = propose && (v_r ≥ 0 && v_r < round_p) do
+		29: if valid(v) && (lockedRound_p ≤ v_r || lockedValue_p = v) then
+		30:   broadcast <PREVOTE, h_p, round_p, id(v)>
+		31: else
+		32:   broadcast <PREVOTE, h_p, round_p, id(v)>
```
hvanz, line 1399:
```suggestion
		32:   broadcast <PREVOTE, h_p, round_p, nil>
```

sergio-mena: Oops! Brainless copy-paste ☹️
Thanks for catching it!

consensus/state.go
lasarojc: This part of the comment is no longer true `Prevote for LockedBlock if we're locked`; this was the first if that you removed.

lasarojc: Pls go ahead with the merge

PR #1208, opened 2023-08-04
Add check for non-`nil` in `enterCommit`

### LINKED ISSUE #1207: Add additional invariant check to `enterCommit`
Was: tendermint/tendermint#7020

Part of: #1171

During tendermint/tendermint#6954 it was discovered that `enterCommit` does not check that the block that has received +2/3 precommits is non-nil. This check is performed outside of the call to [enterCommit](https://github.com/tendermint/tendermint/blob/ea539dcb98345f7af77aaca941004eee8c19cfd4/internal/consensus/state.go#L2164) but should likely also be added to `enterCommit` to safeguard against the check not being included outside the method for future uses.

This issue was cloned from Tendermint Core repo as a follow up of #1175 (tracked by #1171)

sergio-mena:
Closes #1207 

Small change, suggested [here](https://github.com/tendermint/tendermint/pull/6954/files/51db3f86c5ef40c9988255913a3de32211d2e5b6#r713821738), that makes the correctness of `enterCommit` independent from its (only ATM) callsite.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -1567,7 +1567,7 @@ func (cs *State) enterCommit(height int64, commitRound int32) {
 	}()
 
 	blockID, ok := cs.Votes.Precommits(commitRound).TwoThirdsMajority()
-	if !ok {
+	if !ok || blockID.IsNil() {
 		panic("RunActionCommit() expects +2/3 precommits")
 	}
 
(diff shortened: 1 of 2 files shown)


hvanz (commented):
LGTM

hvanz (approved):

lasarojc (approved):

PR #1209, opened 2023-08-04
ADR 101: Add GetLatest method to block service

thanethomson:
Just as in #1168, adds a `GetLatest` method to the gRPC block service to separate out the retrieval of blocks by height from the retrieval of the latest block.

Reviewing commit by commit, skipping the auto-generated commits, would be quickest here. I had to do some refactoring here to ensure the tests ran deterministically.

---

#### PR checklist

- [x] Tests written/updated


rpc/grpc/client/block_service.go
```diff
@@ -22,12 +40,32 @@ type LatestHeightResult struct {
...
-	// GetLatestHeight provides sends the latest committed block height to the given output
-	// channel as blocks are committed.
-	GetLatestHeight(ctx context.Context, resultCh chan<- LatestHeightResult)
+
+	// GetLatestBlock attempts to retrieve the latest committed block.
+	GetLatestBlock(ctx context.Context) (*Block, error)
+
+	// GetLatestHeight provides sends the latest committed block height to the
+	// resulting output channel as blocks are committed.
+	GetLatestHeight(ctx context.Context, opts ...GetLatestHeightOption) (<-chan LatestHeightResult, error)
```
thanethomson, line 68: @andynog I've gone back on my prior recommendation of supplying the channel to the function here because it turns out that the goroutine this function spawns could, non-deterministically, attempt to write to the channel after it's been closed, causing the client to panic.

This new approach causes the `GetLatestHeight` method to manage the whole lifecycle of the channel, but still provides the caller control over the channel size (and whether it's buffered or unbuffered).

jmalicevic (approved):
LGTM

PR #1211, opened 2023-08-06
consensus: export errors

pysel:
---
Contributes to: #1140 

Currently blocked by #1190 

#### Changes

Export errors in `consensus`. More context: #1140


types/errors/proto.go
```diff
@@ -0,0 +1,56 @@
...
+	"fmt"
+	"reflect"
+)
+
+type ErrMsgToProto struct {
+	MessageName string
+	Err         error
+}
+
+func NewErrMsgToProto[T any](msg T, err error) ErrMsgToProto {
```
pysel, line 13: To reviewers:

what is the opinion on this generic factory? This way there is no need to provide a string with struct name when returning the error, you can just pass the struct that caused the error.

pysel: if this design is desired, I could change [other](https://github.com/cometbft/cometbft/pull/1211/files#diff-f8a280b25c2ff3601ed557585a8ba277859b7782cdf5a277931dc62721462515R17) errors to use the same pattern

types/errors/sanity.go
```diff
@@ -12,6 +12,18 @@ type (
 	ErrRequiredField struct {
 		Field string
 	}
+
+	// ErrInvalidField is returned every time any field does not pass some validity check
```
pysel, line 16:
```suggestion
	// ErrInvalidField is returned every time a value does not pass a validity check
```

types/errors/sanity.go
```diff
@@ -12,6 +12,18 @@ type (
...
+	ErrInvalidField struct {
+		Field  string
+		Reason string
+	}
+
+	// ErrWrongField is returned every time a field does not pass validaty check, accompanied with error
```
pysel, line 22:
```suggestion
	// ErrWrongField is returned every time a value does not pass a validaty check, accompanied with error
```

consensus/errors.go
```diff
@@ -0,0 +1,44 @@
...
+	ErrProposalWithoutPreviousCommit = errors.New("propose step; cannot propose anything without commit for the previous block")
+)
+
+// Consensus sentinel errors
+var (
+	ErrInvalidProposalSignature   = errors.New("error invalid proposal signature")
+	ErrInvalidProposalPOLRound    = errors.New("error invalid proposal POL round")
+	ErrAddingVote                 = errors.New("error adding vote")
+	ErrSignatureFoundInPastBlocks = errors.New("found signature from the same key")
+	errPubKeyIsNotSet             = errors.New("pubkey is not set. Look for \"Can't get private validator pubkey\" errors")
```
thanethomson, line 23: Since this is returned by some functions, should it not be exported?

types/errors/proto.go
thanethomson: Given that all we're really using the `msg` field for is its name, I'd say the use of generics and reflection here is undesirable. Especially given that one has to instantiate the empty struct as well when calling the constructor in some cases.

My take is that supplying the type name as a string is simpler, and therefore better. Of course, if we had to rename any of the structs, we'd need to remember to rename the strings, but the chances of us needing to rename such structs often is _very_ low.

If we were embedding the source message itself in the error, then this approach could possibly make sense. But it's not clear whether that's desirable.

pysel: Thanks for a clear feedback, I will refactor it to the desirable design!

consensus/errors.go
pysel: it is only used in `consensus` package. however, same can be said about other errors under `Consensus sentinel errors`. these errors were introduced prior to the time I started working on exporting errors, so I only moved them to `errors.go` w/o changing anything else. personally, I also think it is worth to export it

pysel:
hey @thanethomson! thanks for review, addressed, lmk if it is ready to merge

thanethomson (approved):
LGTM, thanks @pysel! 🎉

PR #1231, opened 2023-08-09
consensus: Avoid process proposal when we know correct nodes validated it

### LINKED ISSUE #1230: `ProcessProposal`: avoid calls when we know one correct process has already accepted it
In the current implementation of `ProcessProposal` (and spec?), it is called just after doing Comet-level validations on the received proposal block.

However, there are some cases where the local node can conclude that the received proposal has already been processed and accepted by another correct node's `ProcessProposal`.

Since `ProcessProposal`'s main goal is to detect and drop bad proposals from byzantine validators, calling `ProcessProposal` on a proposal that we _know_ has passed `ProcessProposal` elsewhere is, at minimum, suboptimal.

These are the cases where we _know_ `ProcessProposal` has already accepted the proposal at some correct node:

* If the proposal's `POLRound` is -1 and the proposed block matches the one locally locked (part of line 23 of the arXiv algorithm)
* If the proposal's `POLRound` is not -1 and less than the current round, and
  * we have received 2f+1 (worth of voting power) valid prevotes for the proposed block in `POLRound`, and
    * either `POLRound` is greater than or `LockedRound?
    * or the proposed block matches the one locally locked
    * (line 29 of the arXiv algorithm)
* If the proposed block matches our `ValidBlock`

In the arXiv algorithm, this is equivalent to removing $valid(v)$ from line 29, and changing line 23 as follows:
`if` $(lockedRound_p = −1 ∧ valid(v)) ∨ lockedValue_p = v$ `then`.
Also, we assume that $valid(v)$ does not include `ProcessProposal` in lines 36 and 50 (do we really need to call $valid(v)$ at all there?).

This is a similar reasoning to the one followed to minimize calls to `timely` in the Proposer-Based Timestamp (PBTS) [specification](https://github.com/cometbft/cometbft/pull/1206#issuecomment-1666092213).

@hvanz @lasarojc @BrendanChou

hvanz:
Resolves #1230 

---

#### PR checklist

- [ ] Tests written/updated


consensus/state.go
```diff
@@ -1339,55 +1339,59 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
...
+		moment we locked it.
+
+		Otherwise, we have already locked on a value that is different from the
+		proposed value, so we prevote nil (line 26), because we are locked on a
+		different value.
 	*/
 	if cs.Proposal.POLRound == -1 {
-		if cs.LockedRound == -1 {
-			logger.Debug("prevote step: ProposalBlock is valid and there is no locked block; prevoting the proposal")
+		if cs.ProposalBlock.HashesTo(cs.LockedBlock.Hash()) {
```
sergio-mena, line 1365: Since we have moved `if cs.LockedRound == -1 {` below, we might be trying to compare hashes of a lockblock, but we haven't any (it may work, but it hinders readability).

How about leaving `if cs.LockedRound == -1 {` and embedding the `ProcessProposal` code inside that condition?

consensus/state.go
```diff
@@ -1339,55 +1339,59 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
...
+
+		// Before prevoting on the block received from the proposer for the current round and height,
+		// we request the Application, via `ProcessProposal` ABCI call, to confirm that the block is
+		// valid. If the Application does not accept the block, consensus prevotes nil.
+		//
+		// WARNING: misuse of block rejection by the Application can seriously compromise
+		// the liveness properties of consensus.
+		// Please see `PrepareProosal`-`ProcessProposal` coherence and determinism properties
+		// in the ABCI++ specification.
+		isAppValid, err := cs.blockExec.ProcessProposal(cs.ProposalBlock, cs.state)
```
sergio-mena, line 1379: Here we are calling `ProcessProposal`, but if `cs.LockedRound == -1` is false, we will be prevoting `nil`

consensus/state.go
hvanz: That will be better. Thanks for the suggestion.

consensus/state.go
hvanz: That's right. It's fixed with the above suggestion.

sergio-mena:
Thanks @hvanz for this. Before approving , do you agree with the third bullet point in the the description of the issue (#1230)?

If yes, what could we do to cover it in the code?

hvanz:
> do you agree with the third bullet point in the the description of the issue (https://github.com/cometbft/cometbft/issues/1230)?
> If yes, what could we do to cover it in the code?

That's a _valid_ point. I updated the code to cover this case according to what [we discussed](https://github.com/cometbft/cometbft/issues/1230#issuecomment-1672986776).

consensus/state.go
```diff
@@ -1339,58 +1339,75 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
...
+
+		Otherwise, we have already locked on a value that is different from the
+		proposed value, so we prevote nil (line 26), because we are locked on a
+		different value.
+
+		Note that there are two cases on which we know that the proposal is
+		application-valid, that is, it was validated by the application at least
+		by one correct node in a previous step:
+		- when the proposal matches our non-nil valid block AND we're not locked on a block, and
+		- when the proposal matches our locked block.
```
sergio-mena, line 1365:
```suggestion
		- when the proposal matches our non-nil locked block.
```
über-minor

sergio-mena (approved):
LGTM. Great job!

consensus/state.go
```diff
@@ -1339,58 +1339,75 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
...
-		we prevote nil since we are locked on a different value. Otherwise, if we're not locked on a block
-		or the proposal matches our locked block, we prevote the proposal.
+		If the application deems the proposal as valid AND we're not locked on a
+		block OR the proposal matches our locked block (line 23), we prevote the
+		proposal (line 24).
+
+		Otherwise, we have already locked on a value that is different from the
+		proposed value, so we prevote nil (line 26), because we are locked on a
+		different value.
+
```
lasarojc, line 1360:
```suggestion
		Otherwise, we have already locked on a value that is different from the
		proposed value, so we prevote nil (line 26).

```

lasarojc (approved):
LGTM

sergio-mena:
@hvanz please hold on merging as #1235 has this PR as base branch. Thanks!

lasarojc:
> @hvanz please hold on merging as #1235 has this PR as base branch. Thanks!

We can go ahead and merge this and I will change the target branch of #1235 to main.

PR #1235, opened 2023-08-10
Approximates code from algorithmic description

lasarojc:
lines 36 to 43 of the algorithm are broken into two separate conditions to detect if a value is being relocked.
Relocking is a subclass of locking and restructuring the case as such brings the code closer to the algorithm, while keeping the distinction in the logging.

---

#### PR checklist

- [ ] Tests written/updated


consensus/state.go
```diff
@@ -1548,33 +1548,33 @@ func (cs *State) enterPrecommit(height int64, round int32) {
...
-	}
-
+	// Lines 36-43
 	// If greater than 2/3 of the voting power on the network prevoted for
 	// the proposed block, update our locked block to this block and issue a
 	// precommit vote for it.
 	if cs.ProposalBlock.HashesTo(blockID.Hash) {
+		cs.LockedRound = round
+
+		// If we're already locked on that block, precommit it, and update the LockedRound
```
sergio-mena, line 1558:
```suggestion
		// If we're already locked on that block, precommit it, and update the LockedRound (above)
```

consensus/state.go
```diff
@@ -1548,33 +1548,33 @@ func (cs *State) enterPrecommit(height int64, round int32) {
...
+		if cs.LockedBlock.HashesTo(blockID.Hash) {
```
sergio-mena, line 1559:
```suggestion
		if cs.LockedBlock == nil {
```
Following our conversation (a few days ago), there are two king of relocking (only updating the round, and updating both the round and the block)

sergio-mena (approved):
Thanks for this @lasarojc, looking great!

cason (commented):
I think the changes are legit, but I have two main comments:

1. We are changing how the protocol works. In the case we locked on a block (so we have that block) in a previous round, and receive a Proposal from the current round for the locked block ID, plus the associated 2/3+ Prevotes, we are not issuing a Precommit for that (locked) block if we didn't receive the full proposed block again (the same block, but with updated round).
2. I wonder why the above mentioned situation is not covered by a test unit. The test unit would also serve as a way to document the protocol operation change we are introducing here. I still think that the ideal procedure should be to first create a test that fails, then update the code so that the test succeeds.

I am aware that creating and updating test units for consensus is a nightmare. But I think that, if we don't have that test, we should at least clearly describe in the PR the change in the scenario mentioned above (and others, if it is the case).

cason:
There is a test which is failining: `TestStateLockPOLUpdateLock`. And it is actually associated with relocking.

cason:
As a suggestion for test units to be considered/updated: `TestStateLockMissingProposalWhenPOLSeenDoesNotUpdateLock` and `TestStateLockMissingProposalWhenPOLSeenDoesNotUnlock`. They both refer to a lock change, so in both cases the node is expected to `Precommit` for `nil`. I wonder whether adapting one of them to cover this situation in which we don't change the lock, as the proposal matches the locked value, but nonetheless we should, with the new code, `Precommit` for `nil`.

cason:
The scenario that I mentioned in previous comments is tested in PR #1257.

consensus/state.go
```diff
@@ -1538,40 +1538,47 @@ func (cs *State) enterPrecommit(height int64, round int32) {
...
+			logger.Debug("precommit step; +2/3 prevoted locked block; relocking")
+
+			cs.LockedRound = round
+			if err := cs.eventBus.PublishEventRelock(cs.RoundStateEvent()); err != nil {
+				logger.Error("failed publishing event relock", "err", err)
+			}
+
+			cs.signAddVote(cmtproto.PrecommitType, blockID.Hash, blockID.PartSetHeader)
+			return
+
```
BrendanChou, line 1558: nit: extra blank line

lasarojc:
```suggestion
```

consensus/state.go
```diff
@@ -1538,40 +1538,46 @@ func (cs *State) enterPrecommit(height int64, round int32) {
...
 
 		// Validate the block.
 		if err := cs.blockExec.ValidateBlock(cs.state, cs.ProposalBlock); err != nil {
 			panic(fmt.Sprintf("precommit step; +2/3 prevoted for an invalid block: %v", err))
 		}
 
+		if cs.LockedRound == -1 {
+			if err := cs.eventBus.PublishEventLock(cs.RoundStateEvent()); err != nil {
+				logger.Error("failed publishing event lock", "err", err)
+			}
```
lasarojc, line 1570:
```suggestion
			if err := cs.eventBus.PublishEventUnlock(cs.RoundStateEvent()); err == nil {
			    if err := cs.eventBus.PublishEventLock(cs.RoundStateEvent()); err != nil {
				    logger.Error("failed publishing event lock after unlock", "err", err)
			    } 
			} else {
				logger.Error("failed publishing event unlock before lock", "err", err)
                        }
```

consensus/state.go
```diff
@@ -1538,40 +1538,47 @@ func (cs *State) enterPrecommit(height int64, round int32) {
...
+		} else {
+			if err := cs.eventBus.PublishEventRelock(cs.RoundStateEvent()); err != nil {
```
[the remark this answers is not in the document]
lasarojc: Indeed we are changing the locked value, which I considered to be a relocking.
We can add another event type to handle this case or send two events in sequence, unlock and lock, and in the suggestion in the previous comment.

lasarojc:
We are closing this PR without merging, as it would revert an optimization that is in place, even though the optimization is for the non-happy path and, as stated in the PR, causes the code to slightly diverge from the pseudo-code.

PR #1244, opened 2023-08-14
chore: Bump minimum Go version on `main` to v1.21

thanethomson:
Partially addresses #1245.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/.changelog/unreleased/minimum-go-version/1244-bump-go-version.md
+++ b/.changelog/unreleased/minimum-go-version/1244-bump-go-version.md
@@ -1,0 +1,2 @@
+- Bump minimum Go version to v1.21
+  ([\#1244](https://github.com/cometbft/cometbft/pull/1244))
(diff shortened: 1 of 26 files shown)


mzabaluev:
I'm on my distro-installed 1.20.6. I can update, but this could also be what outside developers still use.

thanethomson:
> I'm on my distro-installed 1.20.6. I can update, but this could also be what outside developers still use.

This is only for `main` (i.e. v0.39) and v0.38 (still RC). For the latest bleeding-edge versions of Comet, we should always strive to support the latest Go release. For other versions, we should always strive to support the supported versions of Go (e.g. now v1.19 is no longer supported, since the Go team only supports 2 "major" releases).

sergio-mena (approved):
LGTM.
Hopefully, they'll soon realise that supporting _only_ two versions is not really reasonable.

PR #1250, opened 2023-08-14
docs: Add logging guide to contributing guidelines

### LINKED ISSUE #1249: docs: Add logging guidelines
Given that there's such a mix of different perspectives/approaches to logging in the codebase, we should add some guidelines as to how to go about writing good log messages.

The best place I can think of for this right now is in the contributing guidelines, but if anyone has any other suggestions please let me know.

thanethomson:
Closes #1249.

Adds some lightweight/simple logging guidelines to the contributing guidelines.

:book: [Rendered](https://github.com/cometbft/cometbft/blob/thane/1249-logging-guidelines/CONTRIBUTING.md#logging)

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -285,7 +285,7 @@ logger.Error("Failed to execute important operation", "err", err)
 ```golang
 // Message starts with a lowercase letter, and will probably be called very
 // frequently, effectively spamming operators.
-logger.Debug("connected to peer", "peerID", peerID)
+logger.Info("connected to peer", "peerID", peerID)
 
 // Potentially prints huge quantities of data (unnecessary) in a single message,
 // and at info level, spamming operators.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -207,6 +207,96 @@ When updating dependencies, please only update the particular dependencies you
 need. Instead of running `go get -u=patch`, which will update anything, specify
 exactly the dependency you want to update.
 
+## Logging
+
+Operators, consensus engine and application developers all need information from
+the system while it is running. One avenue through which they get that
+information is via the logs. Whenever making contributions, please think
+carefully about what each of those groups of users would want to know about the
+operation of the system and try to adhere to the following guidelines as far as
+reasonably possible.
+
+### To log, or not to log
+
+Whether or not to log something at all should take into consideration how
+frequently the log message will appear. Users hate being spammed by large
+quantities of useless log messages. If you anticipate that a particular log
+message will occur frequently (e.g. a few times per minute), try to find ways to
+either eliminate that message or reduce its frequency (e.g. only logging every
+Nth message, or a summary message every minute or hour).
+
+### Log levels
+
+Different log levels should target different groups of users. At present, only
+**Debug**, **Info** and **Error** levels are supported.
+
+- **Debug**: Should primarily target consensus engine developers (i.e. core team
+  members and developers working on CometBFT forks).
+- **Info** and **Error**: Should primarily target operators and application
+  developers.
+
+### Log messages
+
+Log messages should always be tailored to the intended target audience. Unlike
+Go errors, **log messages must have the first letter of the message
+capitalized**. Only _errors_ in Go should start with a lowercase letter because
+they may end up being chained/embedded, but log messages are not chained in this
+same way.
+
+### Logging parameters
+
+In general, log messages should contain the **bare minimum** amount of
+information for those messages to be actionable by the target audience. So
+instead of dumping large quantities of raw data into the logs (e.g. RPC
+responses, transactions, block data), include lightweight references to the data
+that users can go look up via the RPC, CLI tools, etc. if they are genuinely
+interested in the details.
+
+When outputting variables, also keep in mind **concurrency concerns** of doing
+so. If outputting a pointer, understand that the value associated with that
+pointer at the time of calling the log function may differ to its value at the
+time it is finally serialized into the log message. Perhaps consider creating a
+temporary copy of the specific value you want to output and logging that
+temporary value. Also keep in mind potential data races when doing so.
+
+Finally, **use expensive operations like `fmt.Sprintf` sparingly**, as this can
+have a meaningful performance impact on a running production system. Consider
+and example where one may call `fmt.Sprintf` when logging something at **debug**
+level: even though an operator has configured their system to only log at
+**info** level and above, the expensive `fmt.Sprintf` calls will still take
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -207,6 +207,96 @@ When updating dependencies, please only update the particular dependencies you
...
+
+// Include information about the error.
+logger.Error("Failed to execute important operation", "err", err)
+```
+
+### Examples of bad log messages
+
+```golang
+// Message starts with a lowercase letter, and will probably be called very
+// frequently, effectively spamming operators.
```
andynog, line 287: Does this statement makes sense concerning operators ? 

On the `log levels` section it was mentioned that the `Debug` level should only be targeted to consensus engine developers, so should it still be a concern to operators even if the frequency is high ?

CONTRIBUTING.md
```diff
@@ -207,6 +207,96 @@ When updating dependencies, please only update the particular dependencies you
...
+logger.Debug("connected to peer", "peerID", peerID)
+
+// Potentially prints huge quantities of data (unnecessary) in a single message,
+// and at info level, spamming operators.
+logger.Info("Committed block", "block", fmt.Sprintf("%v", block))
+
+// Just as bad as the info-level message above because the (expensive)
+// fmt.Sprintf is always called, regardless of the operator's configured log
```
andynog, line 295: Same comment as above `Debug` for `operators`

CONTRIBUTING.md
```diff
@@ -207,6 +207,96 @@ When updating dependencies, please only update the particular dependencies you
...
+same way.
+
+### Logging parameters
+
+In general, log messages should contain the **bare minimum** amount of
+information for those messages to be actionable by the target audience. So
+instead of dumping large quantities of raw data into the logs (e.g. RPC
+responses, transactions, block data), include lightweight references to the data
+that users can go look up via the RPC, CLI tools, etc. if they are genuinely
+interested in the details.
```
andynog, line 253: I'd imagine that this statement is only valid for messages that can be retrieved later via a mechanism, like RPC or CLI. There are a lot of instances when this is not possible like Errors or Debug messages that are only available at runtime and need to be logged. Maybe re-word or add something saying that, like "in those instances where additional information can be retrieved...". Just a suggestion.

CONTRIBUTING.md
```diff
@@ -207,6 +207,96 @@ When updating dependencies, please only update the particular dependencies you
...
+temporary value. Also keep in mind potential data races when doing so.
+
+Finally, **use expensive operations like `fmt.Sprintf` sparingly**, as this can
+have a meaningful performance impact on a running production system. Consider
+and example where one may call `fmt.Sprintf` when logging something at **debug**
+level: even though an operator has configured their system to only log at
+**info** level and above, the expensive `fmt.Sprintf` calls will still take
+place, potentially slowing the system down. In such instances, consider printing
+values [lazily][log-lazy].
+
```
andynog, line 269: One construct that I find interesting is the `logger.With` that can embed messages, but I guess this is too much into the coding details. But if you think it's valuable to include some mentioning to it like 

```
logger:   logger.With("service", "BlockService"),
```

feel free to do so, just a suggestion if it makes sense, if not please disregard.

andynog (commented):
Looks good to me. Added a few comments

CONTRIBUTING.md
thanethomson: Yeah I'll change this to `Info`.

CONTRIBUTING.md
thanethomson: This still applies. As per the comment, it still causes a performance hit _regardless_ of the operator's configured log level.

CONTRIBUTING.md
thanethomson: Not sure whether rewording this really adds much value tbh, given that most substantial data that operators or developers would be interested in is stored somewhere in Comet in a way that can be retrieved via means other than logs.

Errors aren't the kind of data I'm thinking of here because they generally don't encode into "large quantities of raw data" the way that entire blocks or RPC responses do.

CONTRIBUTING.md
```diff
@@ -207,6 +207,96 @@ When updating dependencies, please only update the particular dependencies you
...
+that users can go look up via the RPC, CLI tools, etc. if they are genuinely
+interested in the details.
+
+When outputting variables, also keep in mind **concurrency concerns** of doing
+so. If outputting a pointer, understand that the value associated with that
+pointer at the time of calling the log function may differ to its value at the
+time it is finally serialized into the log message. Perhaps consider creating a
+temporary copy of the specific value you want to output and logging that
+temporary value. Also keep in mind potential data races when doing so.
+
```
lasarojc, line 261: Should it also include a warning not to log sensitive information? That will make it easier for validators to share info with the core team.

lasarojc (approved):

CONTRIBUTING.md
thanethomson: Makes sense! Will add that.

CONTRIBUTING.md
```diff
@@ -207,6 +207,96 @@ When updating dependencies, please only update the particular dependencies you
...
+capitalized**. Only _errors_ in Go should start with a lowercase letter because
+they may end up being chained/embedded, but log messages are not chained in this
+same way.
+
+### Logging parameters
+
+In general, log messages should contain the **bare minimum** amount of
+information for those messages to be actionable by the target audience. So
+instead of dumping large quantities of raw data into the logs (e.g. RPC
+responses, transactions, block data), include lightweight references to the data
```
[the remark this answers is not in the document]
thanethomson: Sounds valuable. Would this be similar to having a "trace" level? Or would "noise" messages be logged separately to the log messages? If they're logged separately, how does one link log messages to the logs to the noise messages?

Also, seems like a new type of feature we'd need to add, since our current logging framework doesn't support this.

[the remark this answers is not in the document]
sergio-mena: Yes, we don't support "noise" today (neither "warning", which seems like a miss to me TBH), so I take this convo as a discussion on possible improvements. My experience with "noise" in the past is that it was an super-verbose level (way more verbose than debug) were you dump all possible data to avoid having to attach a debugger (which is problematic with multithreaded or time sensitive processes).
You had to think twice before you enabled that level, just using it temporarily when debugging particular problems, and always restricted to one module only. Otherwise, boom!
Actually, a "noise" log level, combined with the idea I'm planning to work on as PI week (i described it in slack), would be super cool to have to say bye bye to dlv in most non-UT settings.

PR #1251, opened 2023-08-15
ADR 101: Add `Close` method to gRPC client

thanethomson:
As @glnro pointed out, this was an oversight on my part. Should fix it now.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/rpc/grpc/client/client.go
+++ b/rpc/grpc/client/client.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"net"
 
-	"github.com/cosmos/gogoproto/grpc"
 	ggrpc "google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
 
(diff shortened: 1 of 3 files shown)


rpc/grpc/client/client.go
```diff
@@ -50,13 +52,18 @@ func defaultDialerFunc(ctx context.Context, addr string) (net.Conn, error) {
 }
 
 type client struct {
-	conn grpc.ClientConn
+	conn *ggrpc.ClientConn
```
thanethomson, line 55: Not sure why this `grpc.ClientConn` from the gogoproto fork was pulled in - could've been done automatically by my IDE.

jmalicevic: Do you know what the diff is, this might be the cause of Lauren's problems maybe? :) anyways, PR looks good to me

jmalicevic (approved):

PR #1254, opened 2023-08-15
reduce the block size limit

catShaark:
There's a congestion problem on the stride network. The mempool is filled with txs ( prolly ibc update client txs ), this causes the proposer node to propose very big block (with 336 block parts). The peers in the network can't receive all the block parts of the proposed block causing them to vote no for the block. The next proposer will prolly have the same problem if their mempool is also full, and so on. This problem causes stride block times to be 15-30s and commited blocks has 0 txs for some period of time.

I think a very simple, non-breaking solution, which is just reducing the block size limit should solves this problem.

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/types/params.go
+++ b/types/params.go
@@ -96,7 +96,7 @@ func DefaultConsensusParams() *ConsensusParams {
 // DefaultBlockParams returns a default BlockParams.
 func DefaultBlockParams() BlockParams {
 	return BlockParams{
-		MaxBytes: 22020096, // 21MB
+		MaxBytes: 2000000, // 2MB
 		MaxGas:   -1,
 	}
 }


thanethomson:
Is there any way to reproduce this in a test to demonstrate the problem, and then demonstrate how reducing the block size limit fixes the problem?

In my mind, if we're going to change this, it firstly needs to be demonstrated as a definite solution to the problem, and secondly it needs to be opt-in (i.e. configurable via the config file).

The last thing we'd want to do here is make such a change and it doesn't actually fix the problem.

faddat:
I think that it is very likely that this is a correct solution to the problem.

I would also like to mention that I believe that this is one of several possible correct solutions to the problem, and that we may need to do additional solutions.

@thanethomson there isn't a test for this. The reason for that is relatively simple:

Peer to peer networks in the wild exhibit behaviors and tendencies that we cannot predict or model in for example a table-driven test or even a small test network that we spin up under circumstances that are in any way shape or form different from the real world.

sergio-mena:
This PR is proposing to change the _default_ value for the maximum block size, which is a `ConsensusParam`.
This value can, and should, be changed by the application via [ABCI](https://github.com/cometbft/cometbft/blob/v0.37.2/proto/tendermint/abci/types.proto#L257).
Since this problem seems to be affecting a particular chain: Stride. If the solution identified has been to change the block size limit, then it should be Stride application the one updating the `MaxBytes` consensus param via ABCI. This is the best solution, as it does not risk affecting other chains.

catShaark:
Thanks for your feedback. Changes to block size limit won't be state breaking right ?

cason:
> This PR is proposing to change the default value for the maximum block size, which is a ConsensusParam.
This value can, and should, be changed by the application via [ABCI](https://github.com/cometbft/cometbft/blob/v0.37.2/proto/tendermint/abci/types.proto#L257).

You can also set its initial value on the genesis file.

PR #1257, opened 2023-08-17
consensus: test for precommit/locking corner case

cason:
This PR adds a unit test that highlights and documents a corner case in the operation of the consensus protocol.

## Pseudo-code

Considering the abstract description of Tendermint consensus algorithm ([here](https://arxiv.org/pdf/1807.04938.pdf)), the conditions for a node to issue a `Precommit` for a value (block) and update its locked value to this value are:

```
36:  upon 〈PROPOSAL, h_p , round_p, v, ∗〉 from proposer(hp , round_p) AND 2f + 1 〈PREVOTE, hp, round_p , id(v)〉 while valid(v) ∧ step_p ≥ prevote for the first time do
```

This means that for issuing a `Precommit` and locking a value in a round, the node needs to receive a matching `PROPOSAL` message from the round's proposer. In terms of the implementation, this means that the node has to receive a `Proposal` message proposing the `BlockID` that is going to be voted, plus all the corresponding `BlockPart` messages, so that the node is able to build the full `Block` identified by `BlockID`.

## Implementation

The consensus protocol implementation, however, allows the above represented action to take place when the `BlockID` for which a 2/3+ majority of `Prevote` votes is received matches the node's `lockedValue`, which in the implementation is the `LockedBlock`. As mentioned in the [code](https://github.com/cometbft/cometbft/blob/147a47d892ec88c9d3857f335a1565dc17962403/consensus/state.go#L1543), since the value that received 2/3+ `Prevote` votes matches the current `LockedBlock`,  the node can issue a `Precommit` for this block again (the locked value is the last value for which this has happened) and has only to updated its `LockedRound` to the current round.

In practical terms, this approach allows re-using the `LockedBlock` proposed, constructed, and accepted in a previous round to replace the `ProposalBlock` field, which stores the block proposed and constructed in the current round. Observe that a similar approach is also taken when [committing a block](https://github.com/cometbft/cometbft/blob/147a47d892ec88c9d3857f335a1565dc17962403/consensus/state.go#L1657): if the decided `BlockID` matches the `LockedBlock`, we can use this block as it was the `ProposalBlock` for the round.

## Context

In historical terms, this approach of using the block stored in `LockedBlock` as a fallback when the `ProposalBlock` (from the current round) is missing can be tracked back to this [commit](https://github.com/cometbft/cometbft/blame/f6093caac9b5ba7291e681fb5f98210c77f00763/consensus/state.go#L360) from 2014. This behavior, however, is not covered by any of the implemented consensus test units.

The motivation for this PR was an attempt to, among other chances, remove this approach as part of PR #1235.

### DIFF
--- a/consensus/state_test.go
+++ b/consensus/state_test.go
@@ -51,6 +51,7 @@ x * TestStateLockPOLDoesNotUnlock 4 vals, one precommits, other 3 polka nil at
 next round, so we precommit nil but maintain lock
 x * TestStateLockMissingProposalWhenPOLSeenDoesNotUpdateLock - 4 vals, 1 misses proposal but sees POL.
 x * TestStateLockMissingProposalWhenPOLSeenDoesNotUnlock - 4 vals, 1 misses proposal but sees POL.
+x * TestStateLockMissingProposalWhenPOLForLockedBlock - 4 vals, 1 misses proposal but sees POL for locked block.
 x * TestStateLockPOLSafety1 - 4 vals. We shouldn't change lock based on polka at earlier round
 x * TestStateLockPOLSafety2 - 4 vals. After unlocking, we shouldn't relock based on polka at earlier round
 x * TestStatePrevotePOLFromPreviousRound 4 vals, prevote a proposal if a POL was seen for it in a previous round.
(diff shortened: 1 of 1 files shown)


sergio-mena (approved):
🙏

cason:
I will try backporting this PR to the release branches. I assume (aka hope) that not much work will be required.

PR #1261, opened 2023-08-17
ADR101:Fixing the retain height set logic

jmalicevic:
This PR addresses comments made internally on potential corner cases where the retain height is not accepted in case any height height is set , even though it is not the height used for pruning. 

Example
```
appRetainHeight:100
dataCompanionRetainHeight:20
```
Data companion wants to set the height to `50`. 
The old logic would refuse this because the application retain height is set to 100 and we assume blocks to 100 are pruned. However, we will prune only up until the lower of the two retain heights (20) in this case. So it makes sense to accept this value. 

If the application retain height  was the only one set, we would then refuse this as a way to signal to the data companion that the request height is invalid and those blocks might have indeed been pruned.

### DIFF
--- a/state/pruner.go
+++ b/state/pruner.go
@@ -8,7 +8,6 @@ import (
 	"github.com/cometbft/cometbft/config"
 	"github.com/cometbft/cometbft/libs/log"
 	"github.com/cometbft/cometbft/libs/service"
-	"github.com/oasisprotocol/curve25519-voi/curve"
 )
 
 var (
--- a/state/pruner.go
+++ b/state/pruner.go
@@ -8,6 +8,7 @@ import (
 	"github.com/cometbft/cometbft/config"
 	"github.com/cometbft/cometbft/libs/log"
 	"github.com/cometbft/cometbft/libs/service"
+	"github.com/oasisprotocol/curve25519-voi/curve"
 )
 
 var (
@@ -115,23 +116,15 @@ func (p *Pruner) SetApplicationRetainHeight(height int64) error {
 	if !p.checkHeightBound(height) {
 		return ErrInvalidHeightValue
 	}
-	currentAppRetainHeight, err := p.stateStore.GetApplicationRetainHeight()
-	if err != nil {
-		if !errors.Is(err, ErrKeyNotFound) {
-			return err
-		}
-		currentAppRetainHeight = height
-	}
-	currentCompanionRetainHeight, err := p.stateStore.GetCompanionBlockRetainHeight()
-	noCompanionRetainHeight := false
-	if err != nil {
-		if !errors.Is(err, ErrKeyNotFound) {
-			return err
-		}
-		noCompanionRetainHeight = true
-	}
-	if currentAppRetainHeight > height || (!noCompanionRetainHeight && currentCompanionRetainHeight > height) {
+	currentMinRetainHeight := p.findMinRetainHeight()
+	if height < currentMinRetainHeight {
 		return ErrPrunerCannotLowerRetainHeight
+	}
+	_, err := p.stateStore.GetApplicationRetainHeight()
+	if err != nil {
+		if !errors.Is(err, ErrKeyNotFound) {
+			return err
+		}
 	}
 	if err := p.stateStore.SaveApplicationRetainHeight(height); err != nil {
 		return err
@@ -152,9 +145,6 @@ func (p *Pruner) checkHeightBound(height int64) bool {
 //
 // If a higher retain height is already set, we cannot accept the requested
 // height because the blocks might have been pruned.
-//
-// If the application has already set a retain height to a higher value we also
-// cannot accept the requested height as the blocks might have been pruned.
 func (p *Pruner) SetCompanionRetainHeight(height int64) error {
 	// Ensure that all requests to set retain heights via the pruner are
 	// serialized.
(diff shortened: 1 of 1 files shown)


state/pruner.go
```diff
@@ -115,23 +115,15 @@ func (p *Pruner) SetApplicationRetainHeight(height int64) error {
...
 	}
-	currentAppRetainHeight, err := p.stateStore.GetApplicationRetainHeight()
-	if err != nil {
-		if !errors.Is(err, ErrKeyNotFound) {
-			return err
-		}
-		currentAppRetainHeight = height
+	currentMinRetainHeight := p.findMinRetainHeight()
+	if height < currentMinRetainHeight {
+		return ErrPrunerCannotLowerRetainHeight
```
andynog, line 120: Not sure if this logic returns the desirable outcome. For example, let's assume the following scenario:

| DCRetainHeight | AppRetainHeight | Height |
|----------------|-----------------|--------|
|       98       |       100       |   99   |

This logic would return a `currentMinRetainHeight = 98` and since `height (99) > currentMinRetainHeight (99)` this would set the `AppRetainHeight` to `99` but it was already at `100`, so after executing this logic you would end up with:

| DCRetainHeight | AppRetainHeight | Height |
|----------------|-----------------|--------|
|       98       |        99       |   99   |

I'm thinking that the logic in `p.findMinRetainHeight()` shouldn't be used in the logic to `Set` the retain heights, because it mixes the DC and the App retain height which might make the things more complicated. Maybe we just need to check here if the current AppRetainHeight is lower or not and set it accordingly. 

Then in the pruner logic to check between the AppRetainHeight and DCRetainHeight whichever is lower then we use the lower one (the one returned by `p.findMinRetainHeight`

state/pruner.go
```diff
@@ -164,23 +153,16 @@ func (p *Pruner) SetCompanionRetainHeight(height int64) error {
...
-	currentCompanionRetainHeight, err := p.stateStore.GetCompanionBlockRetainHeight()
-	if err != nil {
-		if !errors.Is(err, ErrKeyNotFound) {
-			return err
-		}
-		currentCompanionRetainHeight = height
+	currentMinHeight := p.findMinRetainHeight()
+
+	if height < currentMinHeight {
+		return ErrPrunerCannotLowerRetainHeight
```
andynog, line 159: same comment as above should apply here

andynog (changes requested):
added a few comments in regards the new logic

state/pruner.go
thanethomson: Can we perhaps introduce a comprehensive set of tests for the desired behaviour?

go.sum
```diff
@@ -325,6 +325,7 @@ github.com/go-logr/stdr v1.2.2/go.mod h1:mMo/vtBO5dYbehREoey6XUKy/eSumjCCveDpRre
 github.com/go-sql-driver/mysql v1.7.0 h1:ueSltNNllEqE3qcWBTD0iQd3IpL/6U+mJxLkazJ7YPc=
 github.com/go-sql-driver/mysql v1.7.0/go.mod h1:OXbVy3sEdcQ2Doequ6Z5BW6fXNQTmx+9S1MCJN5yJMI=
 github.com/go-stack/stack v1.8.0/go.mod h1:v0f6uXyyMGvRgIKkXu+yp6POWl0qKG85gN/melR3HDY=
+github.com/go-task/slim-sprig v0.0.0-20210107165309-348f09dbbbc0/go.mod h1:fyg7847qk6SyHyPtNmDHnmrv/HOrqktSC+C9fM+CJOE=
```
mzabaluev, line 328: Hmm, what are these new dependencies for...

state/pruner.go
```diff
@@ -164,23 +153,16 @@ func (p *Pruner) SetCompanionRetainHeight(height int64) error {
...
 	}
-	currentAppRetainHeight, err := p.stateStore.GetApplicationRetainHeight()
-	noAppRetainHeight := false
+	_, err := p.stateStore.GetCompanionBlockRetainHeight()
```
mzabaluev, line 161: What is this get operation for? Checking if the key already exists in the database?

thanethomson:
> I think I get a part of the problem @jmalicevic is on to: at startup, we lack explicit means to determine if the data companion will be there to prune. The previous behaviour was to prune to the height the application has commanded, but if we just do that, we might lose data the companion has not retrieved yet.

Why not just honour the [`storage.pruning.data_companion.enabled` flag](https://github.com/cometbft/cometbft/blob/thane/adr-084-data-companion-pull-api/docs/architecture/adr-101-data-companion-pull-api.md#configuration)?

mzabaluev:
> Why not just honour the [`storage.pruning.data_companion.enabled` flag](https://github.com/cometbft/cometbft/blob/thane/adr-084-data-companion-pull-api/docs/architecture/adr-101-data-companion-pull-api.md#configuration)?

Oh yes, I forgot we have that one too.

go.sum
thanethomson: No `go.mod` updates, so these should be removed by a `go mod tidy`.

state/pruner.go
andynog: no, it seems mostly to check if there's no error with the key in the database (e.g. < 0). Do we really need this ?

jmalicevic: You are totally right, this was left over from the version where I checked the new height of the dc companion is not lower than its own previous height, which I deemed not needed anymore because of the comparison against the minimum height ..allowing it to set a lower height in case it's own height was not the miminum.

jmalicevic:
Closing becuase we will adopt the logic in #1271

PR #1262, opened 2023-08-17
e2e: Scripts for random network topology generator

hvanz:
Scripts for generating random topologies. See README.md.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/test/e2e/scripts/README.md
+++ b/test/e2e/scripts/README.md
@@ -2,11 +2,13 @@
 
 ## Random network topology generator
 
-Generate a topology and a manifest:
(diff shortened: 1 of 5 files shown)


thanethomson (approved):
Nice!

So the topologies generated here are pretty much totally random, apart from the overall connectivity of the network, right?

hvanz:
> So the topologies generated here are pretty much totally random, apart from the overall connectivity of the network, right?

Right. Given the number of nodes, and the minimum and maximum number of peers per node, the connections between nodes are totally random.

PR #1263, opened 2023-08-18
update language

### LINKED ISSUE #1223: Replace usage of "we" in consensus timeouts explained
https://github.com/cometbft/cometbft/blob/cf230821c899f1f0cd3ef9cbbcc678ca01008d46/docs/core/configuration.md?plain=1#L516-L531

I have an uber nit comment regarding this section of the docs. It's not immediately clear to me what "we" refers to. I think "we" can be replaced with "an honest validator" which seems clearer to me.

alijnmerchant21:
Using "a validator should" instead of 'we'

Closes - https://github.com/cometbft/cometbft/issues/1223

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/docs/core/configuration.md
+++ b/docs/core/configuration.md
@@ -518,14 +518,14 @@ matter what is `timeout_commit`.
 
 Here's a brief summary of the timeouts:
 
-- `timeout_propose` = how long we wait for a proposal block before prevoting nil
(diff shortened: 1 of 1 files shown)


thanethomson (approved):
I assume we should be backporting to v0.38.x too?

PR #1271, opened 2023-08-18
ADR 101: Refactor height check-related logic and tests

thanethomson:
Alternative to #1261.

When we make the assumption that the retain heights will be available in the state store, and the pruner knows whether or not the data companion is enabled, we can simplify the code to a certain degree.

I've tried to structure this PR such that it can be reviewed commit by commit, but it may be easier to just review it in one go. I'm happy to also walk folks through it.

~The CI E2E test has been failing for me locally, so I'll mark this as ready to review once it passes.~

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -254,13 +254,14 @@ func NewNode(ctx context.Context,
 	if err != nil {
 		return nil, err
 	}
-	pruner := sm.NewPruner(
-		stateStore,
-		blockStore,
-		logger,
+	prunerOpts := []sm.PrunerOption{
 		sm.WithPrunerInterval(config.Storage.Pruning.Interval),
 		sm.WithPrunerMetrics(smMetrics),
-	)
+	}
+	if config.Storage.Pruning.DataCompanion.Enabled {
+		prunerOpts = append(prunerOpts, sm.WithPrunerCompanionEnabled())
+	}
+	pruner := sm.NewPruner(stateStore, blockStore, logger, prunerOpts...)
 
 	// make block executor for consensus and blocksync reactors to execute blocks
 	blockExec := sm.NewBlockExecutor(
--- a/node/node.go
+++ b/node/node.go
@@ -241,29 +241,16 @@ func NewNode(ctx context.Context,
 		return nil, err
 	}
 
-	err = initApplicationRetainHeight(stateStore)
+	pruner, err := createPruner(
+		config,
+		stateStore,
+		blockStore,
+		smMetrics,
+		logger.With("module", "state"),
+	)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to create pruner: %w", err)
 	}
-
-	if config.Storage.Pruning.DataCompanion.Enabled {
-		err = initCompanionRetainHeights(
-			stateStore,
-			config.Storage.Pruning.DataCompanion.InitialBlockRetainHeight,
-			config.Storage.Pruning.DataCompanion.InitialBlockResultsRetainHeight,
-		)
-		if err != nil {
-			return nil, err
-		}
-	}
-	prunerOpts := []sm.PrunerOption{
-		sm.WithPrunerInterval(config.Storage.Pruning.Interval),
-		sm.WithPrunerMetrics(smMetrics),
-	}
-	if config.Storage.Pruning.DataCompanion.Enabled {
-		prunerOpts = append(prunerOpts, sm.WithPrunerCompanionEnabled())
-	}
-	pruner := sm.NewPruner(stateStore, blockStore, logger, prunerOpts...)
 
 	// make block executor for consensus and blocksync reactors to execute blocks
 	blockExec := sm.NewBlockExecutor(
@@ -872,6 +859,37 @@ func makeNodeInfo(
 
 	err := nodeInfo.Validate()
 	return nodeInfo, err
+}
+
+func createPruner(
+	config *cfg.Config,
+	stateStore sm.Store,
+	blockStore *store.BlockStore,
+	metrics *sm.Metrics,
+	logger log.Logger,
+) (*sm.Pruner, error) {
+	if err := initApplicationRetainHeight(stateStore); err != nil {
+		return nil, err
+	}
+
+	if config.Storage.Pruning.DataCompanion.Enabled {
+		err := initCompanionRetainHeights(
+			stateStore,
+			config.Storage.Pruning.DataCompanion.InitialBlockRetainHeight,
+			config.Storage.Pruning.DataCompanion.InitialBlockResultsRetainHeight,
+		)
+		if err != nil {
+			return nil, err
+		}
+	}
+	prunerOpts := []sm.PrunerOption{
+		sm.WithPrunerInterval(config.Storage.Pruning.Interval),
+		sm.WithPrunerMetrics(metrics),
+	}
+	if config.Storage.Pruning.DataCompanion.Enabled {
+		prunerOpts = append(prunerOpts, sm.WithPrunerCompanionEnabled())
+	}
+	return sm.NewPruner(stateStore, blockStore, logger, prunerOpts...), nil
 }
 
 // Set the initial application retain height to 0 to avoid the data companion
--- a/node/node.go
+++ b/node/node.go
@@ -871,9 +871,9 @@ func makeNodeInfo(
 	return nodeInfo, err
 }
 
-// Set the initial application retain height to 0 to avoid the data companion pruning blocks
-// before the application indicates it is ok
-// We set this to 0 only if the retain height was not set before by the application
+// Set the initial application retain height to 0 to avoid the data companion
+// pruning blocks before the application indicates it is OK. We set this to 0
+// only if the retain height was not set before by the application.
 func initApplicationRetainHeight(stateStore sm.Store) error {
 	if _, err := stateStore.GetApplicationRetainHeight(); err != nil {
 		if errors.Is(err, sm.ErrKeyNotFound) {
--- a/node/node.go
+++ b/node/node.go
@@ -905,22 +905,25 @@ func initApplicationRetainHeight(stateStore sm.Store) error {
 	return nil
 }
 
-// Sets the data companion retain heights if one of two possible conditions is met:
+// Sets the data companion retain heights if one of two possible conditions is
+// met:
 // 1. One or more of the retain heights has not yet been set.
 // 2. One or more of the retain heights is currently 0.
 func initCompanionRetainHeights(stateStore sm.Store, initBlockRH, initBlockResultsRH int64) error {
-	if curBlockRH, err := stateStore.GetCompanionBlockRetainHeight(); err != nil || curBlockRH == 0 {
-		if !errors.Is(err, sm.ErrKeyNotFound) {
-			return fmt.Errorf("failed to obtain companion block retain height: %w", err)
-		}
+	curBlockRH, err := stateStore.GetCompanionBlockRetainHeight()
+	if err != nil && !errors.Is(err, sm.ErrKeyNotFound) {
+		return fmt.Errorf("failed to obtain companion block retain height: %w", err)
+	}
+	if curBlockRH == 0 {
 		if err := stateStore.SaveCompanionBlockRetainHeight(initBlockRH); err != nil {
 			return fmt.Errorf("failed to set initial data companion block retain height: %w", err)
 		}
 	}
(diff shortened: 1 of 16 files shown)


test/e2e/runner/setup.go
```diff
@@ -179,6 +179,10 @@ func MakeConfig(node *e2e.Node) (*config.Config, error) {
 	cfg.BlockSync.Version = node.BlockSyncVersion
 	cfg.Consensus.PeerGossipIntraloopSleepDuration = node.Testnet.PeerGossipIntraloopSleepDuration
 
+	if node.Mode == e2e.ModeValidator || node.Mode == e2e.ModeFull {
+		cfg.Storage.Pruning.DataCompanion.Enabled = true
+	}
```
thanethomson, line 184: I noticed that we didn't even enable the `storage.pruning.data_companion.enabled` flag for the E2E tests, so they were incorrectly passing.

state/errors.go
```diff
@@ -52,6 +52,11 @@ type (
 		Height int64
 	}
 
+	ErrPrunerFailedToGetRetainHeight struct {
+		Which string
```
mzabaluev, line 56: In Rust, you'd be told to create an enum with a `Display` impl for this field. In Go, it's probably OK :stuck_out_tongue:

state/pruner.go
```diff
@@ -92,45 +111,34 @@ func (p *Pruner) SetObserver(obs PrunerObserver) {
 
 func (p *Pruner) OnStart() error {
 	go p.pruneBlocksRoutine()
-	go p.pruneABCIResRoutine()
+	// We only care about pruning ABCI results if the data companion has been
+	// enabled.
+	if p.dcEnabled {
+		go p.pruneABCIResRoutine()
```
mzabaluev, line 117: Side note: I don't like the "routine" naming style.
It's, like, a programming language, every function is a routine of some sort.

mzabaluev (commented):
Looks good! The logic is easy to follow, there is no longer interference between retain heights commanded by the app and the data companion, and when the latter is disabled by config (as it is by default), the currently expected pruning behaviour is retained.

node/node.go
```diff
@@ -884,27 +905,24 @@ func initApplicationRetainHeight(stateStore sm.Store) error {
...
-				}
-			}
-		} else {
-			return fmt.Errorf("failed to obtain companion retain height: %w", err)
+// Sets the data companion retain heights if one of two possible conditions is met:
+// 1. One or more of the retain heights has not yet been set.
+// 2. One or more of the retain heights is currently 0.
+func initCompanionRetainHeights(stateStore sm.Store, initBlockRH, initBlockResultsRH int64) error {
+	if curBlockRH, err := stateStore.GetCompanionBlockRetainHeight(); err != nil || curBlockRH == 0 {
+		if !errors.Is(err, sm.ErrKeyNotFound) {
```
andynog, line 913: this logic is triggering an error in my tests, the `GetCompanionBlockRetainHeight` is returning `err = nil` and `curBlockRH = 0` (because the key is not set yet) and in the test to check if `err` is not `sm.ErrKeyNotFound` it is true (`nil` != `ErrKeyNotFound`) and triggered, and the node fails to start with a log message:

 `ERROR: failed to create node: failed to create pruner: failed to obtain companion block retain height: %!w(<nil>)`

node/node.go
```diff
@@ -884,27 +905,24 @@ func initApplicationRetainHeight(stateStore sm.Store) error {
...
+			return fmt.Errorf("failed to obtain companion block retain height: %w", err)
+		}
+		if err := stateStore.SaveCompanionBlockRetainHeight(initBlockRH); err != nil {
+			return fmt.Errorf("failed to set initial data companion block retain height: %w", err)
+		}
+	}
+	if curBlockResultsRH, err := stateStore.GetABCIResRetainHeight(); err != nil || curBlockResultsRH == 0 {
+		if !errors.Is(err, sm.ErrKeyNotFound) {
```
andynog, line 921: probably same as above

node/node.go
andynog: also, to rely on `curBlockRH == 0` to assume the value is not set might cause some undesired behaviors I believe, since in the `GetCompanionBlockRetainHeight` there's a test

```    buf, err := store.getValue(CompanionBlockRetainHeightKey)
	if err != nil {
		return 0, err
	}
```
I believe there's a possibility that if you get an error in `store.getValue()` (e.g. disk issue, etc) that might return 0, and this logic might reset the current retain height value stored to the default config value. 

For example, let's say current retain height was set to `100` by the Data Companion, then a node is stopped and re-started. There might be a possibility it will get reset to the default value in the config if there's an error in that function call. I don't think this might cause major issues because probably the default config value will be a lower value, but just mentioning here.

andynog (commented):
overall lgtm, but did some local testing and seems some logic might cause some issues, Added comments

node/node.go
thanethomson: Yeah this is what's causing the E2E failure. I've fixed it locally now, but that's causing other tests to fail. Once I'm done playing whack-a-mole with the test failures I'll update :slightly_smiling_face:

node/node.go
```diff
@@ -871,9 +861,40 @@ func makeNodeInfo(
...
+			return nil, err
+		}
+	}
+	prunerOpts := []sm.PrunerOption{
+		sm.WithPrunerInterval(config.Storage.Pruning.Interval),
+		sm.WithPrunerMetrics(metrics),
+	}
+	if config.Storage.Pruning.DataCompanion.Enabled {
+		prunerOpts = append(prunerOpts, sm.WithPrunerCompanionEnabled())
+	}
```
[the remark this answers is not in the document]
thanethomson: Yup, it would be more efficient. I've applied this now, thanks!

state/errors.go
thanethomson: Yeah in Go we'd just create a newtype (based on `string`) to represent the options and define a bunch of constants for each of the variants. It's a bit cumbersome but we could always do that in future if necessary.

node/node.go
```diff
@@ -884,27 +905,27 @@ func initApplicationRetainHeight(stateStore sm.Store) error {
...
+// Sets the data companion retain heights if one of two possible conditions is
+// met:
+// 1. One or more of the retain heights has not yet been set.
+// 2. One or more of the retain heights is currently 0.
+func initCompanionRetainHeights(stateStore sm.Store, initBlockRH, initBlockResultsRH int64) error {
+	curBlockRH, err := stateStore.GetCompanionBlockRetainHeight()
+	if err != nil && !errors.Is(err, sm.ErrKeyNotFound) {
+		return fmt.Errorf("failed to obtain companion block retain height: %w", err)
+	}
+	if curBlockRH == 0 {
```
jmalicevic, line 917: Do we know for sure this value will be 0 if the key did not exist before?

state/pruner.go
```diff
@@ -140,46 +148,32 @@ func (p *Pruner) SetApplicationRetainHeight(height int64) error {
...
 	p.mtx.Lock()
 	defer p.mtx.Unlock()
 
-	if !p.checkHeightBound(height) {
+	if !p.heightWithinBounds(height) {
 		return ErrInvalidHeightValue
 	}
-	currentCompanionRetainHeight, err := p.stateStore.GetCompanionBlockRetainHeight()
+	curRetainHeight, err := p.stateStore.GetCompanionBlockRetainHeight()
 	if err != nil {
```
jmalicevic, line 173: Again here you now assume that an initial block retain height **has** to be set in config otherwise this will always error. I don't have anything against this particular behaviour but we should make this clear in comments above.

state/pruner.go
```diff
@@ -323,39 +313,31 @@ func (p *Pruner) pruneABCIResToRetainHeight(lastRetainHeight int64) int64 {
...
+		return 0
+	}
+	// We only care about the companion retain height if pruning is configured
+	// to respect the companion's retain height.
+	if !p.dcEnabled {
+		return appRetainHeight
 	}
 	dcRetainHeight, err := p.stateStore.GetCompanionBlockRetainHeight()
 	if err != nil {
-		if !errors.Is(err, ErrKeyNotFound) {
```
jmalicevic, line 342: This again relies on the assumption the dc companion has to set at least the initial retain height in the initial config.

test/e2e/runner/setup.go
```diff
@@ -166,19 +166,33 @@ func MakeConfig(node *e2e.Node) (*config.Config, error) {
...
 	cfg.DBBackend = node.Database
 	cfg.StateSync.DiscoveryTime = 5 * time.Second
 	cfg.BlockSync.Version = node.BlockSyncVersion
 	cfg.Consensus.PeerGossipIntraloopSleepDuration = node.Testnet.PeerGossipIntraloopSleepDuration
 
+	// Assume that full nodes and validators will have a data companion
+	// attached, which will need access to the privileged gRPC endpoint.
+	if (node.Mode == e2e.ModeValidator || node.Mode == e2e.ModeFull) && node.EnableCompanionPruning {
+		cfg.Storage.Pruning.DataCompanion.Enabled = true
+		cfg.Storage.Pruning.DataCompanion.InitialBlockRetainHeight = 0
```
jmalicevic, line 190: This means that nothing will ever be pruned right?

node/node.go
thanethomson: Yes, as per https://github.com/cometbft/cometbft/blob/2aeee6ea81693a29d156bd267285de367c684d33/state/store.go#L606

node/node.go
```diff
@@ -884,27 +905,27 @@ func initApplicationRetainHeight(stateStore sm.Store) error {
...
+		if err := stateStore.SaveCompanionBlockRetainHeight(initBlockRH); err != nil {
```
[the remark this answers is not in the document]
thanethomson: It will be 0 by default, so the first value it will be set to will be 0. This dramatically simplifies all of the logic.

node/node.go
```diff
@@ -884,27 +905,27 @@ func initApplicationRetainHeight(stateStore sm.Store) error {
...
+			return fmt.Errorf("failed to set initial data companion block retain height: %w", err)
+		}
+	}
+	curBlockResultsRH, err := stateStore.GetABCIResRetainHeight()
+	if err != nil && !errors.Is(err, sm.ErrKeyNotFound) {
+		return fmt.Errorf("failed to obtain companion block results retain height: %w", err)
+	}
+	if curBlockResultsRH == 0 {
+		if err := stateStore.SaveABCIResRetainHeight(initBlockResultsRH); err != nil {
```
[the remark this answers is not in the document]
thanethomson: Same as above.

state/pruner.go
```diff
@@ -91,46 +110,35 @@ func (p *Pruner) SetObserver(obs PrunerObserver) {
...
-	if err != nil {
-		if !errors.Is(err, ErrKeyNotFound) {
-			return err
-		}
-		currentAppRetainHeight = height
-	}
-	currentCompanionRetainHeight, err := p.stateStore.GetCompanionBlockRetainHeight()
-	noCompanionRetainHeight := false
+	curRetainHeight, err := p.stateStore.GetApplicationRetainHeight()
 	if err != nil {
```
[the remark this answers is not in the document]
thanethomson: See https://github.com/cometbft/cometbft/pull/1271/files#diff-85847e2c654966518d4ab21f147cf243a6c7db53f897c5ed6e494ca18b939528R88

state/pruner.go
thanethomson: See https://github.com/cometbft/cometbft/pull/1271/files#diff-85847e2c654966518d4ab21f147cf243a6c7db53f897c5ed6e494ca18b939528R88

state/pruner.go
thanethomson: Yup, that's the reasoning behind this PR. If they don't explicitly set it, it'll default to 0 as per Go's conventions, which is intended in this PR.

test/e2e/runner/setup.go
thanethomson: Yes. Unless the data companion explicitly sets a different retain height. If I'd left these values out, they would have been initialized to 0 anyways as per Go's conventions - I was just making it explicit here.

state/pruner.go
jmalicevic: Ok then, as it is 0 it also means nothing will be pruned until there is a higher retain height set.

jmalicevic (approved):

PR #1281, opened 2023-08-24
fix: allow for default fuzztags from go-118-fuzz-build

arirubinstein:
---

Includes the new default tags that are provided during the fuzz build from go-118-fuzz-build

#### PR checklist

- [x] Tests written/updated
- [n/a] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)


thanethomson:
@arirubinstein apologies for not getting to this sooner. Should this be applied prior to https://github.com/google/oss-fuzz/pull/10883 being marked as ready for review?

Also, should this change be applied to our release branches?

PR #1284, opened 2023-08-26
evidence: export errors

pysel:
Contributes to: https://github.com/cometbft/cometbft/issues/1140

## Changes

Export errors in evidence. More context: https://github.com/cometbft/cometbft/issues/1140.

### DIFF
--- a/evidence/verify.go
+++ b/evidence/verify.go
@@ -30,19 +30,19 @@ func (evpool *Pool) verify(evidence types.Evidence) error {
 	}
 	evTime := blockMeta.Header.Time
 	if evidence.Time() != evTime {
-		return fmt.Errorf("evidence has a different time to the block it is associated with (%v != %v)",
-			evidence.Time(), evTime)
+		return ErrInvalidEvidence{fmt.Errorf("evidence has a different time to the block it is associated with (%v != %v)",
+			evidence.Time(), evTime)}
 	}
 
 	// checking if evidence is expired calculated using the block evidence time and height
 	if IsEvidenceExpired(height, state.LastBlockTime, evidence.Height(), evTime, evidenceParams) {
-		return fmt.Errorf(
+		return ErrInvalidEvidence{fmt.Errorf(
 			"evidence from height %d (created at: %v) is too old; min height is %d and evidence can not be older than %v",
 			evidence.Height(),
 			evTime,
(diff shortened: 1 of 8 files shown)


state/errors.go
```diff
@@ -51,6 +51,10 @@ type (
 	ErrNoABCIResponsesForHeight struct {
 		Height int64
 	}
+
+	ErrCannotLoadState struct {
```
pysel, line 55: note to reviewers: this PR adds this error to state, because evidence package returns it at some point: https://github.com/cometbft/cometbft/pull/1284/files#diff-4f5e7dc5896be906b72cb46fb61a2f935a428af838090279e7a4bded0679818dR58

p0mvn (approved):
🥪

evidence/verify.go
```diff
@@ -167,50 +166,40 @@ func VerifyLightClientAttack(
...
 	pubKey := val.PubKey
 
 	// H/R/S must be the same
 	if e.VoteA.Height != e.VoteB.Height ||
 		e.VoteA.Round != e.VoteB.Round ||
 		e.VoteA.Type != e.VoteB.Type {
-		return fmt.Errorf("h/r/s does not match: %d/%d/%v vs %d/%d/%v",
-			e.VoteA.Height, e.VoteA.Round, e.VoteA.Type,
-			e.VoteB.Height, e.VoteB.Round, e.VoteB.Type)
+		return ErrDuplicateEvidenceHRTMismatch{*e.VoteA, *e.VoteB}
```
sergio-mena, line 177: Minor: any reason why you prefer to pass the `e.VoteA`/`e.VoteB` by value? (you're copying the struct)

state/errors.go
sergio-mena: Makes sense to me

sergio-mena (approved):
Thanks for this, @pysel 
Only one minor comment, otherwise LGTM

evidence/verify.go
thanethomson: It's probably best given that the usage of the error could take place in another goroutine. Technically the original error was copying the `Height`, `Round` and `Type` values.

While `e.VoteA` and `e.VoteB` shouldn't change, there's no guarantee of that unfortunately.

thanethomson (approved):

PR #1288, opened 2023-08-30
refactor: remove redundant error return from mempool

p0mvn:
Addresses TODO about redundant error return in mempool.Update

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,9 +20,11 @@ security issues.
 
 ### IMPROVEMENTS
 
-- `[rpc]` Remove response data from response failure logs in order
-  to prevent large quantities of log data from being produced
+- `[mempool]` Remove redundant error return in `(*Mempool).Update`
   ([\#654](https://github.com/cometbft/cometbft/issues/654))
+- `[rpc]` Remove response data from response failure logs in order
+to prevent large quantities of log data from being produced
+([\#654](https://github.com/cometbft/cometbft/issues/654))
 
 ### SECURITY FIXES
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ security issues.
 ### IMPROVEMENTS
 
 - `[mempool]` Remove redundant error return in `(*Mempool).Update`
-  ([\#654](https://github.com/cometbft/cometbft/issues/654))
+  ([\#1288](https://github.com/cometbft/cometbft/pull/1288)
 - `[rpc]` Remove response data from response failure logs in order
 to prevent large quantities of log data from being produced
 ([\#654](https://github.com/cometbft/cometbft/issues/654))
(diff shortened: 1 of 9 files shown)


state/execution.go
```diff
@@ -388,7 +388,7 @@ func (blockExec *BlockExecutor) Commit(
 	)
 
 	// Update mempool.
-	err = blockExec.mempool.Update(
+	blockExec.mempool.Update(
 		block.Height,
 		block.Txs,
 		abciResponse.TxResults,
```
sergio-mena, line 394: Minor: can you please use `nil`, rather than `err` in line 399 below to improve readability?

CHANGELOG.md
```diff
@@ -20,9 +20,11 @@ security issues.
 
 ### IMPROVEMENTS
 
+- `[mempool]` Remove redundant error return in `(*Mempool).Update`
+  ([\#1288](https://github.com/cometbft/cometbft/pull/1288)
 - `[rpc]` Remove response data from response failure logs in order
-  to prevent large quantities of log data from being produced
-  ([\#654](https://github.com/cometbft/cometbft/issues/654))
+to prevent large quantities of log data from being produced
+([\#654](https://github.com/cometbft/cometbft/issues/654))
```
sergio-mena, line 27: We don't change the changelog manually. Please read [this section](https://github.com/cometbft/cometbft/blob/main/CONTRIBUTING.md#changelog) for instructions on how to proceed.

CHANGELOG.md
```diff
@@ -20,9 +20,11 @@ security issues.
 
 ### IMPROVEMENTS
 
+- `[mempool]` Remove redundant error return in `(*Mempool).Update`
```
sergio-mena, line 23: Please add the same entry (using unclog, see my comment on that) again in `breaking-changes` directory, as this is changing a public interface.

sergio-mena (commented):
Thanks for this @p0mvn
Mostly LGTM, only the changelog part needs adjustment.

@hvanz what do you think?

cason:
As a general comment here, I think that we should consider the possibility that future implementations of this method would return some form of error. I mean, the mempool is probably the reactor that had most versions in recent year, there is some current work in feature branches refactoring this implementation, and I am sure that it will not be the last one.

So, we either define that there is no way this API method should error, or I will be for maintaining the API as it is, even though the _current_ implementation never returns an error.

sergio-mena:
@cason that's a valid point actually

hvanz:
> I think that we should consider the possibility that future implementations of this method would return some form of error

I agree with @cason and I would suggest to not change the interface for this method, at least for the moment.

The `Mempool` interface needs to be redesigned with the goal of clearly separating the implementation details (`CListMempool`) from the gossip protocol in the reactor, and to allow different implementations to coexist. This will require discussing around an ADR. The current interface was not carefully designed; it is more of a result of many patches over the years. Also, many things happen inside `Update`, so it's not improbable that the current implementation or a different one returns an error.

p0mvn:
Great points. Totally makes sense. Let's close this then

PR #1293, opened 2023-09-01
node: verify genesis doc hash against the file contents rather than remarshalled JSON

### LINKED ISSUE #1287: Change the hash on the doc, to a hash on the genesis file itself.
This is a follow-up issue from the PR #1017 submitted and merged

During the PR review, there were a few points highlighted to address some concerns the CometBFT team had regarding this particular PR. Since this PR was merged from a fork, a few of the suggestions by the CometBFT team to address the concerns were not possible to be accepted directly on the PR. So we merged the PR #1017 and will address these concerns  in this issue.

_Concerns_ to be addressed:

1.  The logic in PR 1017 uses the [marshalled JSON bytes](https://github.com/cometbft/cometbft/blob/226628fe3c3aab1244adcfef7ef072b5bcdff337/node/setup.go#L547C2-L547C2) as the input for the hashing function. As discussed in the PR, this approach might cause some non-deterministic behaviours or it might be very error prone. JSON is not a canonical data representation, so any semantically insignificant variation, in the genesis file might result in a hash mismatch.
2. The `AppState` part of the genesis is application dependent and CometBFT has no "understanding" on its internal data structures. In the [`GenesisDoc` type, the `AppState` property is just a `json.RawMessage` ](https://github.com/cometbft/cometbft/blob/8e6e6ca0bcc732747e10020be2c5e9f3779e6b0c/types/genesis.go#L45)so any variations in its content might cause side effects that are beyond CometBFT's control such as hash mismatches.
3. In the current logic on PR 1017, if the node crashes before saving the genesis hash into storage some undesired side effects can happen. As per [comment](https://github.com/cometbft/cometbft/pull/1017#discussion_r1308796297), the logic on this [line](https://github.com/cometbft/cometbft/blob/226628fe3c3aab1244adcfef7ef072b5bcdff337/node/setup.go#L577) should be changed by applying the suggestion in the comment. 

 _Context_ on genesis file chain initialization and upgrades: 

5. Historically the way chain operators (node operators and validators) validate the `genesis` file is to compare a checksum obtained from a reference repository or a group of operators during a coordinated upgrade or new chain initialization. For example, during a chain upgrade, validators might export the current state via the CLI (e.g. `gaiad export`) and then they can generate a checksum of the exported genesis (based on a agreed hashing function (e.g. `sha256sum) and then they coordinate the upgrade by validating their own checksum against the "agreed" checksum of the majority of operators to ensure they all match in order to start the chain and achieve consensus. Therefore, implementing a logic that leverages the same hashing functionality on the same input used by operators is critical to ensure a reliable and trusted solution.
6. Currently, as far as we know, there's no central repository or on-chain stored information about these genesis checksum . Historically, most chains maintain a `mainnet` repository somewhere in Github where they list the checksum on a page for each genesis upgrade, or the checksum might be part of the release information for a chain mainnet or upgrade in Github.

Suggested _changes_ to be made to improve the solution:

7. The main concern to be addressed is around the hashing functionality as outlined in the sections `#1` and `#2` above. In our opinion, relying on a hash of the marshalled JSON input (bytes) is not the best solution (as mentioned in the sections above). We believe that a better approach is to rely on the checksum of the `genesis.json` file content. By applying this logic, we use the same approach chain operators use to manually validate a the `genesis` file (perform a checksum of the file contents, e.g. `$ sha256sum genesis.json`). We believe operators will have more confidence in the solution since they can also have an external way to validate the hash information. 
8. [OPTIONAL] Add a way for operators to specify the hash (checksum) value when starting (or initializing) a node. It could be a CLI command parameter or allow them to add the value in the configuration file (e.g. `genesis_hash = "b2fca9af31aae6953124439df5bc300c1ef11fb8be938aeb570767c5b6bfdaaf"` that can be used to validate the hash computed by CometBFT when reading the genesis file. If the hashes don't match (computed = specified) then the node fails to start, if they match then store the hash in the database. This will ensure that they are confident on the hash that will be stored in the database. As part of the future planning, we can also return the hash information via RPC and this will make it possible for operators and other users to validate the hash information against peers or public RPC nodes and that would address point `#6` above

```[tasklist]
### Tasks
- [ ] Create genesis hash using `sha`. Make sure it is properly saved into the DB.
```

mzabaluev:
Closes #1287

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/node/setup.go
+++ b/node/setup.go
@@ -60,6 +60,7 @@ func DefaultGenesisDocProviderFunc(config *cfg.Config) GenesisDocProvider {
 	return func() (ChecksummedGenesisDoc, error) {
 		// FIXME: find a way to stream the file incrementally,
 		// for the JSON	parser and the checksum computation.
+		// https://github.com/cometbft/cometbft/issues/1302
 		jsonBlob, err := os.ReadFile(config.GenesisFile())
 		if err != nil {
 			return ChecksummedGenesisDoc{}, fmt.Errorf("couldn't read GenesisDoc file: %w", err)
--- a/node/setup.go
+++ b/node/setup.go
@@ -52,22 +52,19 @@ type ChecksummedGenesisDoc struct {
 // GenesisDocProvider returns a GenesisDoc together with its SHA256 checksum.
 // It allows the GenesisDoc to be pulled from sources other than the
 // filesystem, for instance from a distributed key-value store cluster.
-// An optional checksum to verify the JSON document content against can be
-// passed as the parameter.
-type GenesisDocProvider func(expectedChecksum []byte) (ChecksummedGenesisDoc, error)
+type GenesisDocProvider func() (ChecksummedGenesisDoc, error)
 
 // DefaultGenesisDocProviderFunc returns a GenesisDocProvider that loads
 // the GenesisDoc from the config.GenesisFile() on the filesystem.
 func DefaultGenesisDocProviderFunc(config *cfg.Config) GenesisDocProvider {
-	return func(expectedChecksum []byte) (ChecksummedGenesisDoc, error) {
+	return func() (ChecksummedGenesisDoc, error) {
+		// FIXME: find a way to stream the file incrementally,
+		// for the JSON	parser and the checksum computation.
 		jsonBlob, err := os.ReadFile(config.GenesisFile())
 		if err != nil {
 			return ChecksummedGenesisDoc{}, fmt.Errorf("couldn't read GenesisDoc file: %w", err)
 		}
 		incomingChecksum := tmhash.Sum(jsonBlob)
-		if len(expectedChecksum) != 0 && !bytes.Equal(expectedChecksum, incomingChecksum) {
-			return ChecksummedGenesisDoc{}, fmt.Errorf("expected genesis doc hash does not match loaded genesis doc")
-		}
 		genDoc, err := types.GenesisDocFromJSON(jsonBlob)
 		if err != nil {
 			return ChecksummedGenesisDoc{}, err
@@ -565,7 +562,7 @@ func LoadStateFromDBOrGenesisDocProvider(
 	if err != nil {
 		return sm.State{}, nil, fmt.Errorf("error retrieving genesis doc hash: %w", err)
 	}
-	csGenDoc, err := genesisDocProvider(genDocHash)
+	csGenDoc, err := genesisDocProvider()
 	if err != nil {
 		return sm.State{}, nil, err
 	}
--- a/node/setup.go
+++ b/node/setup.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"os"
 	"strings"
 	"time"
 
@@ -21,7 +22,6 @@ import (
 	"github.com/cometbft/cometbft/evidence"
 	"github.com/cometbft/cometbft/statesync"
 
-	cmtjson "github.com/cometbft/cometbft/libs/json"
 	"github.com/cometbft/cometbft/libs/log"
 	"github.com/cometbft/cometbft/light"
 	mempl "github.com/cometbft/cometbft/mempool"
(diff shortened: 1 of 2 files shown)


node/node.go
```diff
@@ -155,21 +155,36 @@ func NewNode(ctx context.Context,
 		DiscardABCIResponses: config.Storage.DiscardABCIResponses,
 	})
 
-	genDoc, err := genesisDocProvider()
+	// Get the genesisDoc hash from db if exists
+	genDocHash, err := stateDB.Get(genesisDocHashKey)
```
jmalicevic, line 159: Note that, if the key did not exist there will be a `errKeyEmpty` or something like that return. Worth double checking whether that breaks the logic.

node/node.go
```diff
@@ -155,21 +155,36 @@ func NewNode(ctx context.Context,
...
 	}
 
-	state, err := loadStateFromDBOrGenesisDoc(stateStore, stateDB, genDoc)
+	state, err := loadStateFromDBOrGenesisDoc(stateStore, stateDB, csGenDoc)
 	if err != nil {
 		return nil, err
 	}
 
+	// Save the genDoc hash in the store if it doesn't already exist for future verification
+	if len(genDocHash) == 0 {
```
jmalicevic, line 180: if it is not 0 should we not compare it against `cs.Sha256Checksum` to make sure they are the same?

jmalicevic: Actually I think this is where Sergio suggested we always save it? If so then we should first check whether the existing one matches the new one we generate, and then always save it.

mzabaluev: > if it is not 0 should we not compare it against `cs.Sha256Checksum` to make sure they are the same?

We do, as per the new contract on `GenesisDocProvider`.

node/node.go
mzabaluev: It's handled the same as it was prior to this PR: https://github.com/cometbft/cometbft/pull/1293/files#diff-bbaa90ad1c80edb1f0084525de9b5e22da605e22d5e0b374d04c1e6a363d6f9eL543
If this is wrong, let's fix it here.
I actually assumed db `Get` method gives us an empty array and no error if the key is missing. Need to learn our internal APIs.

node/node.go
mzabaluev: > Actually I think this is where Sergio suggested we always save it?

No need to overwrite it if it's already there and matches the file, I think.

node/node.go
jmalicevic: You are right, I checked , the db wraps the error in this case and does not return it but rather just retuns a `nil` value so this is all good.

mzabaluev:
I don't think this needs a changelog entry, as this is a follow-up to #1017 and we haven't released the functional changes yet.

sergio-mena:
@mzabaluev PR #1295 is likely to affect this PR. Sorry for the inconveniences...

mzabaluev:
> @mzabaluev PR #1295 is likely to affect this PR. Sorry for the inconveniences...

No worries, I will wait (and review) and rebase once that one is merged.

jmalicevic:
Merging the PR now so you are free to rebase! Thanks Mikhail !

node/setup.go
```diff
@@ -42,16 +42,34 @@ import (
...
+// GenesisDocProvider returns a GenesisDoc together with its SHA256 checksum.
 // It allows the GenesisDoc to be pulled from sources other than the
 // filesystem, for instance from a distributed key-value store cluster.
-type GenesisDocProvider func() (*types.GenesisDoc, error)
+type GenesisDocProvider func() (ChecksummedGenesisDoc, error)
 
 // DefaultGenesisDocProviderFunc returns a GenesisDocProvider that loads
 // the GenesisDoc from the config.GenesisFile() on the filesystem.
 func DefaultGenesisDocProviderFunc(config *cfg.Config) GenesisDocProvider {
-	return func() (*types.GenesisDoc, error) {
```
jmalicevic, line 53: I wonder whether instead of changing the return value of this function (which is used in many other places), we can not move the checksum logic into the `LoadStateFromDBOrGenesisProvider` to be executed only once if there is no hash in the DB or if we have a flag provided via cli? Basically move it to line 576  and put the logic of reading the file and computing the sum there?

mzabaluev: > I wonder whether instead of changing the return value of this function (which is used in many other places), we can not move the checksum logic into the `LoadStateFromDBOrGenesisProvider` to be executed only once if there is no hash in the DB or if we have a flag provided via cli? Basically move it to line 576 and put the logic of reading the file and computing the sum there?

I have not come across any other places where the `GenesisDocProvider` signature was used. My concern is about eventually reworking this for streamability: the provider should not need to store the whole file content in RAM to parse the JSON or calculate the checksum. This signature allows implementing it with an incremental parser (e.g. `json.Decoder` from the standard library) and an incremental SHA-256 hasher. I have left a FIXME comment about this in the default provider function. Maybe I should turn it into a backlog issue.

mzabaluev: Filed away as #1302

jmalicevic: `DefaultGenesisProvider` is being used a lot although mostly to create a new node so technically it should be fine. But I am not sure whether it is used by users as well .

So instead of changing a public interface for a future optimization, could we not do one of two things:
- Leave this until we do the actual optimization and then we can even create a `DefaultStreamingDocProvider` or
- Keep the `genesisProvider` returning just the doc and have a special `createChecksum` function that could afterwards be altered to not read the file but rather stream through it based on a given path. This function is then called to get the hash (the fact that the genesisProvider has it in memory does not change anything as we anyways still keep it in memory).

mzabaluev: Also, we cannot really do this without changing the signature and contract for `GenesisDocProvider` in some way: in main, it only returns the parsed doc, and both retrieval or JSON content from the source and parsing is supposed to happen inside the provider function.

mzabaluev: There is one backward-compatible alternative: add a checksum field to the `GenesisDoc` struct and mandate every genesis doc provider set it with the computed checksum. But: 1) it feels conceptually wrong; 2) this special field has to be fenced away from normal handling of the `GenesisDoc` fields by the JSON marshal and unmarshal implementations.

jmalicevic: Yes, you are right it might be wrong.. What abut the solution of changing the signature of `LoadStateFromDBOrGenesisDocProvider` by adding it a path to the genesis file (as we will anyways change this function signature to pass the hash from command line)? Both are breaking, just not sure which is more (as in breaking the way people are used to use these things).

sergio-mena: The SDK uses `GenesisDocProvider` in three places (if I didn't miss any): [here](https://github.com/cosmos/cosmos-sdk/blob/release/v0.50.x/server/start.go#L398), [here](https://github.com/cosmos/cosmos-sdk/blob/release/v0.50.x/server/start.go#L362), and [here](https://github.com/cosmos/cosmos-sdk/blob/release/v0.50.x/server/cmt_cmds.go#L404).

If we include a changelog explaining the signature change, I doesn't seem highly disruptive. Thoughts?

mzabaluev: > If we include a changelog explaining the signature change, I doesn't seem highly disruptive. Thoughts?

I support this. But I do need to add a changelog, after all.

jmalicevic (approved):

sergio-mena (approved):
🙏

Issue #1037, opened 2023-06-27
storage: Alternative representation of the genesis file.


jmalicevic:
The genesis file is stored as a value in CometBFTs kv store. As the file can be large it can hit the internal file size limitations of databases. 

We need to figure out an alternative representation of this file. A proposal was to store the hash of the genesis file in the state store but the actual file is stored outside the store. 

Existing PRs targetting this :
- #1017 
Historically, this issue was mentioned in the original Tendermint repository under:
   - tendermint/tendermint#8416
   - tendermint/tendermint#6563

A potentially related problem is that at the moment the Genesis file is retrievable via RPC. Even though it is retrieved in chunks, it is still loaded entirely into memory before it is chunked up. We could improve on this if we take it into consideration. 

Things to understand and complete as part of this issue: 
* [x] #1286
* [x] #1078

After discussing, this issue will be addressed in two phases:

### Phase I 

* [x] Remove the genesis file from the kvstore and store it in an external file.
  * [x] Adopt PR #1017.
  * [x] #1287
  * [x] #1259

### Phase II
- [x] Reduce genesis's memory footprint: #1290

### Phase III
- [x] Optimize chunking logic (from storage and RPC perspective): #1289
  - [x] Reduce genesis chunks size: #4298 

### DoD

* [x] Genesis file is not stored as one entry in the state store
* [x] Tests written to confirm no errors are generated for large genesis files
* [x] Document how the genesis file is stored and represented
* [x] (stretch) Improve retrieval of Genesis file when requested via RPC

andynog:
Not sure if this would be also part of this issue, but I believe we need to better understand the use cases for people retrieving the genesis file via RPC, maybe we can provide another way to query parts of it only or something that returns like a hash that people can compare to another hash from a file retrieved from a different source (e.g. Github)

mzabaluev:
It's sad that the `/genesis_chunked` endpoint has to exist in the form it is. Because it's JSON-RPC, the protocol cannot do the usual HTTP thing: use the `Range` header on a `genesis` GET endpoint. That can also enable off-the-shelf caching/CDN solutions that abundantly exist for HTTP web services.

Or maybe, we can serve the genesis endpoint over plain HTTP and deprecate the JSON-RPC one? It could even be a start of converting the entire RPC to plain REST, which you can route and load-balance with off-the-shelf cloud components when things like [ADR 102](https://github.com/cometbft/cometbft/issues/707) are realized.
[...]

PR #1295, opened 2023-09-01
Re-implement removal of genesis persistence in state db

sergio-mena:
Contributes to #1037

This is a more self-contained implementation of #1017.
No API changes, it keeps the old logic when retrieving the genDoc at star-up time.
This is passing the e2e nightly tests that were failing in #1017.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/node/setup.go
+++ b/node/setup.go
@@ -529,7 +529,7 @@ func startStateSync(
 
 //------------------------------------------------------------------------------
 
-//var genesisDocKey = []byte("genesisDoc")
+// var genesisDocKey = []byte("genesisDoc")
 var genesisDocHashKey = []byte("genesisDocHash")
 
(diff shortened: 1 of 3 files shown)


jmalicevic (approved):
Looks good to me, thanks Sergio!

node/setup.go
```diff
@@ -530,6 +530,7 @@ func startStateSync(
 //------------------------------------------------------------------------------
 
 var genesisDocKey = []byte("genesisDoc")
```
sergio-mena, line 532:
```suggestion
//var genesisDocKey = []byte("genesisDoc")
```
Commenting it out to appease the linter, but we'll need it when implementing the upgrade path

mzabaluev (approved):
LGTM

PR #1297, opened 2023-09-03
state: remove genesis file from database 

### LINKED ISSUE #1259: Ensure smooth upgrades after removing genesis from kvstore.
The work done in PR#1017 will remove the genesis file from the database. We need to ensure that a node, upgrading to a version of CometBFT with this change can smoothly do so and retrieve the file from the new location. This should be automatic.

Additionally, we should delete the file from the database if it was persisted in a previous version. This operation needs to be automatically performed on boot. If we detect the genesis file in the database, we delete it and output a warning message to the operator that this information is no longer there. 

**Note** This can be merged only once we merge the PR removing the genesis file from the DB. 

DoD:

- [ ] Node boot up code is altered to check for the existence of the genesis file and deleting it
- [ ] Unit tests exist to test this behaviour. 
- [ ] Changelog entry

jmalicevic:
Closes #1259 

As the genesis file will not be retrieved from the database anymore, there is no need to keep it in the database and consume storage. 

When a node upgrades to a version of comet containing #1295 , we will delete the old contents of the database. 

This PR should be merged after #1293 to ensure it is compatible with the implementation. 

---

#### PR checklist

- [x] Tests written/updated


node/node_test.go
```diff
@@ -468,6 +468,62 @@ func TestNodeNewNodeCustomReactors(t *testing.T) {
...
+	// Start and stop to close the db for later reading
+	err = n.Start()
+	require.NoError(t, err)
+
+	err = n.Stop()
+	require.NoError(t, err)
+
+	stateDB, err = cfg.DefaultDBProvider(&cfg.DBContext{ID: "state", Config: config})
+	require.NoError(t, err)
+	genDocHash, err := stateDB.Get(genesisDocHashKey)
```
andynog, line 519: Not sure if it's useful but a test like `ok, err := stateDB.Has(genesisDocHashKey)` could be useful?

andynog (dismissed):
lgtm, added a few comments

node/node.go
```diff
@@ -160,6 +160,16 @@ func NewNode(ctx context.Context,
 		return nil, err
 	}
 
+	// The key will be deleted if it existed.
+	// Not checking whether the key is there in case the genesis file was larger than
+	// the max size of a value (in rocksDB for example), which would cause the check
+	// to fail and prevent the node from booting.
+	err = stateDB.Delete(genesisDocKey)
```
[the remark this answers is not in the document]
andynog: OK, makes sense. Is there an issue related to this problem of RocksDB with big files ?

node/node_test.go
andynog: This might not make sense because of the issue with RocksDB

node/node.go
```diff
@@ -160,6 +160,16 @@ func NewNode(ctx context.Context,
...
+	if err != nil {
```
mzabaluev, line 168: Just to confirm, `Delete` will not report an error if the key is not present, will it?

jmalicevic: It seems that for some DBs cometbft-db returns nil if delete returns the `ErrKeyNotFound` error and for some it just forwards the error. Some DBs on the other hand, themsleves do not return  an error. 

I think it is safe to leave this as is here (with the log message), as it will not disrupt the node, but open an issue in the cometbft-db repo to unify this behaviour across all the supported DBs.

node/node.go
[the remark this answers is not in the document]
jmalicevic: This was the original issue solved by PRs #1017 and #1295 .

.changelog/unreleased/improvements/1297-remove-genesis-from-db-upgrade.md
```diff
@@ -0,0 +1,2 @@
+- `[node]` On upgrade, after PR 1296, delete the genesis file existing in the DB.
```
sergio-mena, line 1:
```suggestion
- `[node]` On upgrade, after [\#1296](https://github.com/cometbft/cometbft/pull/1296), delete the genesis file existing in the DB.
```

node/node.go
```diff
@@ -281,6 +281,17 @@ func NewNode(ctx context.Context,
 		return nil, err
 	}
 
+	// The key will be deleted if it existed.
+	// Not checking whether the key is there in case the genesis file was larger than
+	// the max size of a value (in rocksDB for example), which would cause the check
+	// to fail and prevent the node from booting.
+	logger.Info("WARNING: deleting genesis file from database, the database stores a hash of the original genesis file now")
```
sergio-mena, line 288:
```suggestion
	logger.Info("WARNING: deleting genesis file from database if present, the database stores a hash of the original genesis file now")
```

node/node.go
sergio-mena: > to unify this behaviour across all the supported DBs.

... or wait until we chose one and remove the others

sergio-mena (approved):
🙏

node/node.go
andynog: I think for the scope of this PR it's OK just to leave as a log message if it's not causing any side effects. The scope to adapt or change the DB behaviour is a much bigger scope and agree that we deal with that in the future.

Issue #1296, opened 2023-09-01
docs: How to run a data companion and use the pruning API


andynog:
Add documentation for the ADR-101 feature branch. In the documentation, add information such as:

```[tasklist]
### Tasks
- [x] Introduction
- [x] gRPC services
- [x] Pruning services
```

andynog:
I believe I've added all the content needed for the docs

andynog:
Merged on PR #1307

PR #1307, opened 2023-09-07
docs: gRPC services and data companion pruning services

andynog:
close: #1296 

Added documentation related to the new Data Companion features such as the new gRPC services and the pruning mechanism.

#### PR checklist

- [ ] Tests written/updated


docs/data-companion/pruning.md
```diff
@@ -0,0 +1,161 @@
...
+
+# Using a Data Companion to influence data pruning on a CometBFT node
+
+CometBFT employs a sophisticated pruning logic to eliminate unnecessary data and reduce storage requirements.
+
+This document covers use cases where the data companion can influence the pruning process and explains how the data companion
+can influence the pruning logic on the node.
+
+CometBFT provides a privileged gRPC endpoint for the pruning service. This privileged endpoint is distinct from the
+regular gRPC endpoint and require separate configuration and activation. These "privileged" services
```
jmalicevic, line 16:
```suggestion
regular gRPC endpoint and requires separate configuration and activation. These "privileged" services
```

docs/data-companion/intro.md
```diff
@@ -0,0 +1,22 @@
...
+order: 1
+parent:
+    title: Introduction
+    order: 1
+---
+
+# Introduction
+
+In [ADR-101](https://github.com/cometbft/cometbft/blob/thane/adr-084-data-companion-pull-api/docs/architecture/adr-101-data-companion-pull-api.md),
+it was proposed that a Data Companion be introduced as a solution to enable external applications to have an impact on the
```
jmalicevic, line 11:
```suggestion
it was proposed that a Data Companion API be introduced as a solution to enable external applications to have an impact on the
```

docs/data-companion/intro.md
```diff
@@ -0,0 +1,22 @@
...
+pruning of data on a node. The Data Companion works by leveraging the newly introduced gRPC APIs to control the [pruning](./pruning.md)
```
jmalicevic, line 12: The ADR talks about the API, the data companion is then an arbitrary implementation using this API for pruning so I would rephrase this,

docs/data-companion/intro.md
```diff
@@ -0,0 +1,22 @@
...
+process on the node.
+
+By using the Data Companion Pull API, users can ensure that only the necessary data is kept on the node, while unnecessary
+data is efficiently pruned. This allows for more efficient storage management, as well as improved overall performance
+of the node. With the Data Companion, users can have greater control over their data and can optimize their storage usage
+to meet their specific needs.
+
```
jmalicevic, line 19: I would add somewhere here what can be pruned:
- Blocks and state via block retain height
- ABCI results (if enabled)
- The block indexer data
- The transaciton indexer data

docs/data-companion/pruning.md
```diff
@@ -0,0 +1,161 @@
...
+The node will preserve blocks up to the lowest value between the data companion retain height and the application block retain height.
+
+The "block results retain height" pruning parameter is the second crucial aspect of the pruning API.
+This parameter determines the height up to which the node will keep block results. By retaining block results to a
+certain height, the node can efficiently manage its storage and optimize its performance.
+
+With these two parameters, the pruning service ensures that the node efficiently manages its data storage.
+
+### Block Retain Height
+
```
jmalicevic, line 74: I would note here that this retain height can also be influenced by the application. AS in- it is the only retain height of which the application has control. The application controls this via appRetainHeight into which the operator can get insight via a gRPC call. Note that , at any point in time, only the lowest retain height is accepted. By default the application block retain height and the data companion retain height are set to 0. This is done in order to prevent any one of them pruning the data while the other has not indicated it is ok to do so.

docs/data-companion/pruning.md
```diff
@@ -0,0 +1,161 @@
...
+if err != nil {
+    // Do something with the error
+} else {
+    // Do something with the `retainHeight` value
+}
+```
+
+### Block Results Retain Height
+
+In order to set the block results retain height on the node, you have to enable the privileged services endpoint and the block results service in the
```
jmalicevic, line 119: In the current verison, if the config parameter `DiscardABCIResults=true` , no results are stored and this retain height is ignored.

docs/data-companion/quick-start.md
```diff
@@ -0,0 +1,269 @@
...
+order: 1
+parent:
+    title: Quick Start
+    order: 2
+---
+
+
+# Creating a Data Companion for CometBFT
+
+## Fetching data
```
jmalicevic, line 11: This title is a bit redundant with the Fetching data title below. I think it can even be removed here and this serves as an intro.

docs/data-companion/quick-start.md
```diff
@@ -0,0 +1,269 @@
...
+} else {
+    // Do something with the `blockResults`
+}
+
+```
+
+### Latest height stream
+
+There is a new way to subscribe to a stream of new blocks with the Block service. Previously, you could connect and
+subscribe to new block events using websockets through the RPC service.
```
jmalicevic, line 181: A benefit of this is that now you can subscribe only for the latest height so your gRPC endpoint is not transferring entire blocks to keep you up to date.. You can then fetch the blocks at your own pace via the `getBlockByHeight` endpoint .

docs/data-companion/quick-start.md
```diff
@@ -0,0 +1,269 @@
...
+pruning on the node.
+
+One important concept that can affect the pruning of nodes is the `retain_height`. The retain_height determines the specific
+height from which the data can be safely deleted from the node's storage. By considering the retain_height,
+nodes can effectively manage their storage usage and ensure that they are only retaining the data that is necessary for
+their operations. This is important because storage space is a finite resource and nodes with limited storage space may
+struggle to keep up with the growth of the blockchain.
+
+For a comprehensive understanding of the pruning service for the data companion, please see the document
+[Using a Data Companion to influence data pruning on a CometBFT node](./pruning.md).
```
jmalicevic, line 269: I think the diagram you once had on the workflow (which might be in the ADR even could be useful. It can be even simplified, just to visually show the steps described above.

docs/data-companion/intro.md
andynog: rewrite this part

docs/data-companion/pruning.md
andynog: added that information

docs/data-companion/pruning.md
andynog: I had added some information related to this but I'll augment that with info from this comment, thanks.

docs/data-companion/quick-start.md
andynog: fixed

docs/data-companion/quick-start.md
andynog: added

docs/data-companion/quick-start.md
andynog: The diagram is from ADR102, I can link this in the future for a better reference

docs/data-companion/grpc.md
```diff
@@ -0,0 +1,216 @@
...
+parent:
+    title: gRPC services
+    order: 2
+---
+
+
+# Fetching data from the node
+
+One of the most important steps to create a Data Companion service is to extract the necessary data from the node.
+Fortunately, CometBFT provides gRPC endpoints that allows you to fetch the data, such as `version`, `block` and
```
jmalicevic, line 12:
```suggestion
Fortunately, CometBFT provides gRPC endpoints that allow you to fetch the data, such as `version`, `block` and
```

docs/data-companion/grpc.md
```diff
@@ -0,0 +1,216 @@
...
+## Latest height streaming
+
+There is a new way to subscribe to a stream of new blocks with the Block service. Previously, you could connect and
+subscribe to new block events using websockets through the RPC service.
+
+One of the advantages of the new streaming service is that it allows you to opt for the latest height subscription.
+This way, the gRPC endpoint will not have to transfer entire blocks to keep you updated. Instead, you can fetch the
+blocks at the desired pace through the `GetBlockByHeight` method.
+
+To receive  the latest height from the stream, you need to call the method that returns the receive-only channel and then
```
jmalicevic, line 136:
```suggestion
To receive the latest height from the stream, you need to call the method that returns the receive-only channel and then
```

docs/data-companion/pruning.md
```diff
@@ -0,0 +1,338 @@
...
+
+# Pruning data via the pruning service
+
+CometBFT employs a sophisticated pruning logic to eliminate unnecessary data and reduce storage requirements.
+
+This document covers use cases where the pruning process on a CometBFT node can be influenced via the Data Companion
+pruning service API.
+
+CometBFT provides a privileged gRPC endpoint for the pruning service. This privileged endpoint is distinct from the
+non-privileged (regular) gRPC endpoint and require separate configuration and activation. These "privileged" services
```
jmalicevic, line 16:
```suggestion
non-privileged (regular) gRPC endpoint and requires separate configuration and activation. These "privileged" services
```

docs/data-companion/pruning.md
```diff
@@ -0,0 +1,338 @@
...
+
+CometBFT provides "privileged" services which are not intended to be exposed to the public-facing Internet.
+
+The privileged services offered by CometBFT can modify the data stored in the node, and hence, it's essential to keep
+them off by default to avoid any unintended modifications.
+
+However, when required, these services can be activated to set and retrieve a retained height, which can influence
+the pruning mechanism on the node.
+
+To be able to use the privileged gRPC services, it should be enabled through CometBFT's configuration.
```
jmalicevic, line 31:
```suggestion
To be able to use the privileged gRPC services, they should be enabled through CometBFT's configuration.
```

docs/data-companion/pruning.md
```diff
@@ -0,0 +1,338 @@
...
+height from which the data can be safely deleted from the node's storage. By considering the retain height,
+nodes can effectively manage their storage usage and ensure that they are only retaining the data that is necessary for
+their operations. This is important because storage space is a finite resource and nodes with limited storage space may
+struggle to keep up with the growth of the blockchain.
+
+## Pruning Blocks
+
+The pruning service uses the "block retain height" parameter to specify the height to which the node will
+preserve blocks. It is important to note that this parameter differs from the application block retain height, which
+the application sets in response to ABCI commit messages.
```
jmalicevic, line 112: I would add another sentence here of the form :
When deciding which blocks to prune, CometBFT choses the lower of the two retain heights. By default the application retain height is 0, which means that no blocks will be pruned until the application changes this. The default retain height for an enabled data companion is 0 as well. This ensures that the application will not trigger pruning until the data companion has done the same.

docs/data-companion/pruning.md
```diff
@@ -0,0 +1,338 @@
...
+    // `retainHeight.App`
+    // `retainHeight.PruningService`
+}
+```
+
+Retaining data is crucial to data management, and the application has complete control over the application retain height.
+The operator can monitor the application retain height with `GetBlockRetainHeight`, which returns a `RetainHeights`
+structure with both the block retain height and the application retain height (as shown in the code above).
+
+It's worth noting that at any given point in time, the node will only accept the lowest retained height.
```
jmalicevic, line 159:
```suggestion
It's worth noting that at any given point in time, the node will only accept the lowest retain height.
```

docs/data-companion/pruning.md
```diff
@@ -0,0 +1,338 @@
...
+If you try to set the `Block Retain Height` to a value that is lower to what is currently stored in the node, an error will
+be returned informing that.
+
+By default, both the application retain height and the data companion retain height are set to zero. This is done to prevent
+either one of them from prematurely pruning the data while the other has not indicated that it's okay to do so.
+
+In essence, the node will preserve blocks up to the lowest value between data companion block retain height and the application
```
jmalicevic, line 166: Ah I just saw this so disregard my comment above. But I would move this section up front, before the example. It is flows more natural in my opinion (but it is not a strong preference).

jmalicevic (approved):
Minor comments, great write -up Andy!

docs/data-companion/pruning.md
andynog: thanks

docs/data-companion/grpc.md
andynog: thanks

docs/data-companion/grpc.md
andynog: thanks

docs/data-companion/pruning.md
andynog: thanks

docs/data-companion/pruning.md
andynog: thanks

docs/data-companion/pruning.md
andynog: OK, I've moved the section up before the example, kept the original text (not the suggested one above since they are the same)

PR #1324, opened 2023-09-11
Add genesis_hash flag check on node startup

### LINKED ISSUE #1303: cli/genesis: allow operators to provide Comet with a hash of the genesis file to be verified on load
### Summary

On load CometBFT stores the hash of the genesis file in the state database but verifies it against a recomputed hash of the genesis file itself. Operators often get a precomputed hash along with a genesis file. We can provide them with an optional `cli` parameter via which they pass this hash for Comet to double check that the genesis file was not corrupted.

### Proposal

- Introducing a new flag - `genesis_hash` to be passed via command line to CometBFT
- This flag is optional as it can increase the safety guarantees provided by comet but it might be cumbersome for operators to update their existing node boot-up procedures with an extra required flags. 
- If the flag is not provided the we log a warning to the operator.
- This hash has to be checked against the sha256 hash of the genesis file document iteself (after merging PR #1293 ). If there is a mismatch between the two hashes, the node will not boot. 
- Only when these two match, we store the hash of the genesis file in stateDB for checks further in the future. There should not be a scenario where there has been a hash stored already and it mismatches (a correct) hash provided by the operator. The only scenarios where this is possible - a corrupt initial genesis file or a operator giving the wrong hash. In any case, we have to log the error and not boot.

glnro:
Closes #1303

Given there is already functionality to pass the genesis hash on node startup, this PR moves the verification step from within the scope of the cli execution to node creation. After the genesis hash is either retrieved or set, it is compared against the optional genesis hash provided by the `--genenesis_hash` flag, and will return an error if there is a mismatch. 

A new field in the `Storage` config was created to accommodate passing this flag value to scope of genesis hash verification.

---

#### PR checklist

- [x] Tests written/updated


cmd/cometbft/commands/run_node.go
```diff
@@ -94,8 +89,8 @@ func NewRunNodeCmd(nodeProvider nm.Provider) *cobra.Command {
 		Aliases: []string{"node", "run"},
 		Short:   "Run the CometBFT node",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if err := checkGenesisHash(config); err != nil {
-				return err
+			if len(genesisHash) != 0 {
+				config.Storage.GenesisHash = genesisHash
```
andynog, line 93: Would it be helpful to do some extra validation ? like if `len(genesisHash) == 64`.

config/config.go
```diff
@@ -1070,13 +1070,16 @@ type StorageConfig struct {
 	// required for `/block_results` RPC queries, and to reindex events in the
 	// command-line tool.
 	DiscardABCIResponses bool `mapstructure:"discard_abci_responses"`
+
+	GenesisHash []byte `mapstructure:"genesis_hash"`
```
andynog, line 1074: Since the hash is a fixed number of bytes should it be constrained here ? `[size]byte`

node/setup.go
```diff
@@ -560,6 +560,7 @@ var genesisDocHashKey = []byte("genesisDocHash")
 func LoadStateFromDBOrGenesisDocProvider(
 	stateDB dbm.DB,
 	genesisDocProvider GenesisDocProvider,
+	storeConfig *cfg.StorageConfig,
```
andynog, line 563: Instead of passing the whole `storageConfig` just pass the `[]byte` for the hash since that's all you use ? Also in line 595 you check against the bytes anyways

andynog (changes requested):
Good job. Looks good to me, made some suggestions to improve it if makes sense. Thanks !

cmd/cometbft/commands/run_node.go
jmalicevic: Maybe add the check that was done below - || config.Genesis == ""

node/node_test.go
```diff
@@ -535,6 +538,54 @@ func TestNodeNewNodeGenesisHashMismatch(t *testing.T) {
 	require.Equal(t, "genesis doc hash in db does not match loaded genesis doc", err.Error())
 }
 
+func TestNodeeGenesisHashFlagMismatch(t *testing.T) {
```
jmalicevic, line 541:
```suggestion
func TestNodeGenesisHashFlagMismatch(t *testing.T) {
```

node/setup.go
```diff
@@ -571,20 +572,30 @@ func LoadStateFromDBOrGenesisDocProvider(
...
 			return sm.State{}, nil, fmt.Errorf("failed to save genesis doc hash to db: %w", err)
 		}
 	} else {
 		if !bytes.Equal(genDocHash, csGenDoc.Sha256Checksum) {
 			return sm.State{}, nil, fmt.Errorf("genesis doc hash in db does not match loaded genesis doc")
 		}
 	}
+
+	// Validate that existing or recently saved genesis file hash matches optional --genesis_hash passed by operator
+	recentGenDoc, err := stateDB.Get(genesisDocHashKey)
```
jmalicevic, line 591: I think it is safe to use the same hash that was retrieved or `csGenDoc.Sha256Checksum` no need to read it again.

node/setup.go
```diff
@@ -571,20 +572,30 @@ func LoadStateFromDBOrGenesisDocProvider(
...
+	if err != nil {
+		return sm.State{}, nil, fmt.Errorf("error retrieving genesis doc hash: %w", err)
+	}
+	if len(storeConfig.GenesisHash) != 0 && !bytes.Equal(recentGenDoc, storeConfig.GenesisHash) {
```
jmalicevic, line 595: I wonder whether, in case they are not the same ,we should delete the hash from the DB to make sure there is not an incorrect hash in the database left in case of a mismatch.

node/setup.go
```diff
@@ -571,20 +572,30 @@ func LoadStateFromDBOrGenesisDocProvider(
...
+		return sm.State{}, nil, fmt.Errorf("genesis doc hash in db does not match passed --genesis_hash value")
```
jmalicevic, line 596:
```suggestion
                err = stateDB.Delete(genesisDocHashKey)
		return sm.State{}, nil, fmt.Errorf("genesis doc hash in db does not match passed --genesis_hash value")
```

test/e2e/node/config.go
```diff
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/BurntSushi/toml"
+
```
jmalicevic, line 8:
```suggestion
```

node/setup.go
andynog: Yes, that would be good. Yesterday testing something else on this branch like other genesis I had to reset the DB because the hash mismatch but not sure it is a possible scenario

sergio-mena:
Is there another way to pass the genesis hash from the command line to the node starting logic, other than the config?
If it is added to the config, it will also be visible to users there... unless that's what we want to do

glnro:
> Is there another way to pass the genesis hash from the command line to the node starting logic, other than the config? If it is added to the config, it will also be visible to users there... unless that's what we want to do

We would have to change the signature of [`DefaultNewNode`](https://github.com/cometbft/cometbft/blob/main/node/setup.go#L83C9-L83C9) and pass a pointer to the genesis hash which is an optional flag. Adding a field to the storage config felt less invasive but I'm happy to reconsider the alternative if you think it's preferable.

sergio-mena:
After some discussions with @glnro, she convinced me all alternatives for PR, other than using a new config knob are hackier

config/config.go
glnro: Forcing this causes test failures, I think the check in `run_node` that the length is 64 is sufficient

node/node_test.go
```diff
@@ -473,7 +476,6 @@ func TestNodeNewNodeCustomReactors(t *testing.T) {
 func TestNodeNewNodeDeleteGenesisFileFromDB(t *testing.T) {
 	config := test.ResetTestRoot("node_new_node_delete_genesis_from_db")
 	defer os.RemoveAll(config.RootDir)
-	fmt.Println(config.RootDir)
```
jmalicevic, line 476: This was accidentally left over from my last PR so I am piggy backing it here instead of opening a new PR.

config/config.go
andynog: makes sense

node/setup.go
andynog: resolved

node/setup.go
andynog: resolved

node/node_test.go
andynog: resolved

node/setup.go
andynog: I believe this was suggested here https://github.com/cometbft/cometbft/pull/1324/files/d2021d1aeac819d7079a0820775a37f3617e15e4#r1322667963 but that has been marked as resolved without commiting the suggested change so I'm assuming that you decided that this is not necessary

andynog (approved):
lgtm, I think most of the concerns have been addressed. There's only one last comment about deleting the hash from the DB in case of a mismatch but I'll let you decide on that.

config/config.go
```diff
@@ -1071,15 +1071,18 @@ type StorageConfig struct {
 	// command-line tool.
 	DiscardABCIResponses bool `mapstructure:"discard_abci_responses"`
 
-	GenesisHash []byte `mapstructure:"genesis_hash"`
+	// Hex representation of the hash of the genesis file.
+	// This is set when an operator provides a hash via the command line.
+	// It is used to verify the hash of the actual genesis file.
```
sergio-mena, line 1076: Can you say it's optional?

config/toml.go
```diff
@@ -467,6 +467,12 @@ peer_query_maj23_sleep_duration = "{{ .Consensus.PeerQueryMaj23SleepDuration }}"
 # reindex events in the command-line tool.
 discard_abci_responses = {{ .Storage.DiscardABCIResponses}}
 
+
+# Hash of the Genesis file (as hex string), passed to CometBFT via the command line. 
+# If this hash mismatches the hash that COmetBFT computes on the genesis file,
```
sergio-mena, line 472:
```suggestion
# If this hash mismatches the hash that CometBFT computes on the genesis file,
```

sergio-mena (approved):
LGTM

PR #1334, opened 2023-09-12
TxIndexer and BlockIndexer pruning metrics

### LINKED ISSUE #1282: ADR-101: Add metrics to monitor indexer pruning
### Summary

Once the indexer pruning mechanism and the corresponding gRPC service are merged, we need a way to indicate to the operator the retain heights requested by the data companion.

Suggested metrics are 

- `block_indexer_retain_height` : Requested block indexer retain height
- `tx_indexer_retain_height` : Requested tx indexer retain height
- `block_indexer_base_height`: The minimum height at which we have block events (should demonstrate the effects of pruning the block indexer) 
- `tx_indexer_base_height` : The minimum height at which we have transaction events (should demonstrate the effects of pruning the tx indexer)

 Why do we need this feature?

This will give operators of the data companion the opporutinity to monitor whether their requested retain height was accepted an whether the events were indeed pruned.

werty144:
Closes #1282
---

#### PR checklist

- [ ] Tests written/updated


jmalicevic (approved):
LGTM

Issue #656, opened 2023-04-05
Investigate new code coverage for integration tests (Go > 1.20)


andynog:
Investigate the new Go (from version 1.20) feature that supports code coverage for integration tests (at the package level). 

This is a [good article](https://go.dev/blog/integration-test-coverage) on this new feature that we can potentially use in the e2e tests in the future for example.

PR #1349, opened 2023-09-14
Add test/e2e/data to gitignore

hvanz:
~~`.gitignore` includes `test/p2p/data` but this directory is never created.~~ What is needed is an entry for `test/e2e/data` which  occurs when running e2e networks.

The line for `test/p2p/data` [was added](https://github.com/tendermint/tendermint/pull/5918) with the fuzz tests. I run the fuzz tests and they don't create this directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,7 @@ test/e2e/build
 test/e2e/data/
 test/e2e/networks/*/
 test/logs
+test/p2p/data/
 vendor
 test/fuzz/**/corpus
 test/fuzz/**/crashers
--- a/.gitignore
+++ b/.gitignore
@@ -40,9 +40,9 @@ terraform.tfstate.d
 test/app/grpc_client
 test/loadtime/build
 test/e2e/build
+test/e2e/data/
 test/e2e/networks/*/
 test/logs
-test/p2p/data/
 vendor
 test/fuzz/**/corpus
 test/fuzz/**/crashers


.gitignore
```diff
@@ -40,9 +40,9 @@ terraform.tfstate.d
 test/app/grpc_client
 test/loadtime/build
 test/e2e/build
+test/e2e/data/
 test/e2e/networks/*/
 test/logs
-test/p2p/data/
```
andynog, line 45: I've noticed this path `test/p2p/data` referenced in a couple other places:

- in `test/docker/Dockerfile` 
```
RUN cometbft testnet \
  --config $REPO/test/docker/config-template.toml \
  --node-dir-prefix="mach" \
  --v=4 \
  --populate-persistent-peers=false \
  --o=$REPO/test/p2p/data
```

- in `.dockerignore` there's an entry for it there too

andynog (commented):
Looks OK, but just need to double check if this folder is not created when running the Dockerfile below

.gitignore
thanethomson: I suspect the `test/docker` folder is no longer maintained and actually needs to be removed in its entirety.

hvanz: I failed to see those entries in `test/docker`. Thanks for pointing this out, @andynog! 

This `Dockerfile` is built on `golang:1.15` so it's rather old. It is used for the integration tests (`make test_integrations`) which I think they are unmaintained. 

So for the moment I suggest to leave `test/p2p/data/` in gitignore and just a new one for `test/e2e/data/`, which is actually needed. Removing `test/p2p/data/` should be part of another issue focusing on integration tests.

hvanz: Here's an existing issue on integration tests: #656

thanethomson (approved):

PR #1351, opened 2023-09-14
Bump Go version used in `v0.34.x` to v1.20

thanethomson:
Follows from the govulncheck failure in #1342, since Go v1.19 has reached EOL.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/p2p/conn/connection.go
+++ b/p2p/conn/connection.go
@@ -707,12 +707,13 @@ func (c *MConnection) Status() ConnectionStatus {
 	status.RecvMonitor = c.recvMonitor.Status()
 	status.Channels = make([]ChannelStatus, len(c.channels))
 	for i, channel := range c.channels {
+		channel := channel
 		status.Channels[i] = ChannelStatus{
 			ID:                channel.desc.ID,
 			SendQueueCapacity: cap(channel.sendQueue),
-			SendQueueSize:     int(atomic.LoadInt32(&channel.sendQueueSize)), //nolint:gosec
+			SendQueueSize:     int(atomic.LoadInt32(&channel.sendQueueSize)),
 			Priority:          channel.desc.Priority,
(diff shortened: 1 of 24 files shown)


thanethomson:
In order to get the linter to actually work on v0.34.x, I had to update the v0.34.x CI and code to use the latest golangci-lint, update the linter config, and add some ignores for specific lints in certain places.

If we're going to maintain the v0.34.x branch for much longer, we should possibly consider re-enabling some of the linters I've disabled and fix the lints (a trivial, but time-consuming and low-impact job).

p2p/conn/connection.go
```diff
@@ -708,9 +710,9 @@ func (c *MConnection) Status() ConnectionStatus {
 		status.Channels[i] = ChannelStatus{
 			ID:                channel.desc.ID,
 			SendQueueCapacity: cap(channel.sendQueue),
-			SendQueueSize:     int(atomic.LoadInt32(&channel.sendQueueSize)),
+			SendQueueSize:     int(atomic.LoadInt32(&channel.sendQueueSize)), //nolint:gosec
 			Priority:          channel.desc.Priority,
-			RecentlySent:      atomic.LoadInt64(&channel.recentlySent),
+			RecentlySent:      atomic.LoadInt64(&channel.recentlySent), //nolint:gosec
```
thanethomson, line 715: I had to manually skip the `gosec` linter checks here, otherwise it complains:

```
G601: Implicit memory aliasing in for loop.
```

My understanding of how this code works is that we're immediately loading the int values from the pointers during the current iteration, so it's irrelevant. Please correct me if I'm wrong though.

lasarojc (approved):

p2p/conn/connection.go
sergio-mena: I think it is complaining about the loop variable (potentially) being rewritten while using it as a reference. IICR @jmalicevic fixed this

sergio-mena (approved):

PR #1377, opened 2023-09-20
ADR101: Moved changelog entries

jmalicevic:
---

#### PR checklist

- [ ] Tests written/updated


sergio-mena (approved):
We took a look at it together and there seems to be a incompatibility with git's merge algorithm. When fixing conflicts "deleted by us" or "deleted by them", we have to be extra careful that those files should actually be there (sometimes git's suggestion is not the right one)

Issue #1376, opened 2023-09-20
docs: Update docs to reflect max bytes check


alijnmerchant21:
### Description

The documentation currently shows an example of simply returning the list of transactions sent in the `RequestPrepareProposal.` However, according to the CometBFT spec, the Application is responsible for ensuring the total bytes of transactions returned does not exceed `RequestPrepareProposal.max_tx_bytes.`

The current behavior in the codebase adheres to the spec by only returning transactions that fit within the max bytes limit. However, the documentation example could cause confusion by not demonstrating the max bytes check.

### Fix

The documentation example should be updated to demonstrate checking the total transaction bytes and only returning transactions up to the max bytes limit.

A note should be added to the documentation clarifying that the Application is responsible for enforcing the max bytes limit, even if more transactions are sent in the `RequestPrepareProposal.`

This will ensure the documentation matches both the spec and the actual code behavior.

PR #1384, opened 2023-09-21
Max byte check

alijnmerchant21:
This PR closes the issue #1376
---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/docs/guides/go-built-in.md
+++ b/docs/guides/go-built-in.md
@@ -40,7 +40,6 @@ guarantees as two processes would be communicating via established binary protoc
 CometBFT will not have access to application's state.
 If that is the way you wish to proceed, use the [Creating an application in Go](./go.md) guide instead of this one.
 
-
 ## 1.1 Installing Go
 
 Verify that you have the latest version of Go installed (refer to the [official guide for installing Go](https://golang.org/doc/install)):
@@ -380,7 +379,6 @@ The application state may have changed between the initial execution of `CheckTx
 
 Other methods, such as `Query`, rely on a consistent view of the application's state, the application should only update its state by committing the Badger transactions when the full block has been delivered and the Commit method is invoked.
 
-
 The `Commit` method tells the application to make permanent the effects of
 the application transactions.
 Let's update the method to terminate the pending Badger transaction and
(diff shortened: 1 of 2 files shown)


docs/guides/go-built-in.md
```diff
@@ -462,16 +460,29 @@ execution of the block once accepted.
...
+       if totalBytes > int64(proposal.MaxTxBytes) {
+         break
+       }
+     }
+
+     return &abcitypes.ResponsePrepareProposal{Txs: proposal.Txs}, nil
+ }
+ ```
+
+ Once a proposed block is received by a node, the proposal is passed to the application to determine its validity before voting to accept the proposal.
```
lasarojc, line 480: This paragraph refers the to snipped below, that is, to processProposal, not to prepareProposal.
However, the following 2 paragraphs, which you added/edited, refer to prepareProposal, not processProposal.
You need to reorder these paragraphs.

Also, you removed the comment about processProposal being used to validate the proposed block; I think you should bring it back.

lasarojc (approved):
Just a small fix required.

docs/guides/go-built-in.md
```diff
@@ -462,16 +460,33 @@ execution of the block once accepted.
...
+   totalBytes := int64(0)
+   txs := make([]byte, 0)
+
+   for _, tx := range proposal.Txs {
+     totalBytes += int64(len(tx))
+     txs = append(txs, tx...)
 
-Once a proposed block is received by a node, the proposal is passed to the application to give
-its blessing before voting to accept the proposal.
+       if totalBytes > int64(proposal.MaxTxBytes) {
```
odeke-em, line 471: The use of totalBytes is redundant and could instead just use len(txs) but the even better check would be to firstly check if the final added value is greater than proposal.MaxTxBytes

```go
txs := make([]byte, 0, len(proposal.Txs)*100)
max := int64(proposal.MaxTxBytes)

for _, tx := range proposal.Txs {
   if n := len(txs) + len(tx); n > max || n <= 0 {
       break
   }
```

lasarojc: Indeed, @odeke-em Thanks for pointing it out.
Given that this PR is now merged, would you mind opening a new PR with your suggestions?

sergio-mena: Isn't the problem here that the `txs` array is not of the right type? (is `[]byte`, should be `[][]byte`)
Then we are "coalescing" all transactions into a byte array... but that's not what we should be doing, is it?

lasarojc: You are right, @sergio-mena, but @odeke-em has a point in that the overflow check should be done before the last transaction has been added, otherwise it is too late to break the loop.

PR #1408, opened 2023-09-27
feat(consensus): additional sanity checks for the size of proposed blocks

cason:
Hardens tests regarding the size of proposed blocks, namely:

- The byte size of a proposal block `Part` should be constant (`== types.BlockPartSizeBytes`), except for the last part of a `PartSet`  (`<= types.BlockPartSizeBytes`)
-  A valid `Proposal` should not enclose a `PartSet` enabling the building of a `ProposalBlock` with size larger than the configured `ConsensusParams.Block.MaxBytes`. Notice that building a `ProposalBlock` larger than the allowed would fail in any case, but the proposed changes also invalidate the associated `Proposal`.

---

#### PR checklist

- [x] Tests written/updated


types/part_set.go
```diff
@@ -31,6 +32,11 @@ func (part *Part) ValidateBasic() error {
 	if len(part.Bytes) > int(BlockPartSizeBytes) {
 		return fmt.Errorf("too big: %d bytes, max: %d", len(part.Bytes), BlockPartSizeBytes)
```
sergio-mena, line 33: I know this is not strictly necessary, but we can turn this `fmt.Error` into a proper error as we're doing below?

types/part_set.go
```diff
@@ -31,6 +32,11 @@ func (part *Part) ValidateBasic() error {
...
 	}
+	// All parts except the last one should have the same constant size.
+	if part.Proof.Index < part.Proof.Total-1 &&
```
sergio-mena, line 36: Shouldn't this condition be using the actual Index of the part, rather than the index of the proof?

types/part_set.go
```diff
@@ -29,7 +31,11 @@ type Part struct {
 // ValidateBasic performs basic validation.
 func (part *Part) ValidateBasic() error {
 	if len(part.Bytes) > int(BlockPartSizeBytes) {
-		return fmt.Errorf("too big: %d bytes, max: %d", len(part.Bytes), BlockPartSizeBytes)
+		return ErrPartTooBig
+	}
+	// All parts except the last one should have the same constant size.
+	if int64(part.Index) < part.Proof.Total-1 && len(part.Bytes) != int(BlockPartSizeBytes) {
```
sergio-mena, line 37: We're still using `part.Proof.Total` here, are 100% sure this total will always be in sync with the total number of parts?

cason: Yes, this is a good point. But we don't have the `Total` field in each `Part`, we only have it on the `PartSet`.

cason: The point here is that wrong values for `Proof.Total` and `Proof.Index` should make the Merkle-tree verification to fail. But I am not sure whether we test this at this point. Notice that the `Part` tests have nothing on the arrays used to fill the `Part` and the `Part.Proof` bytes.

consensus/state.go
```diff
@@ -1975,6 +1975,15 @@ func (cs *State) defaultSetProposal(proposal *types.Proposal) error {
...
 	}
 
+	// Validate the proposed block size, derived from its PartSetHeader
+	maxBytes := cs.state.ConsensusParams.Block.MaxBytes
+	if maxBytes > -1 {
+		maxBytes = int64(types.MaxBlockSizeBytes)
+	}
+	if int64(proposal.BlockID.PartSetHeader.Total) > (maxBytes-1)/int64(types.BlockPartSizeBytes)+1 {
+		return ErrProposalTooManyParts
+	}
```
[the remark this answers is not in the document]
cason: Wow, this really a bug. The code should be the same as in the following method, namely the test is whether `maxBytes == -1`, meaning that it is _disabled_ .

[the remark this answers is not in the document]
sergio-mena: Thanks @lasarojc for spotting this!

types/part_set.go
sergio-mena: Is there a possibility that `Part` is used without populating `part.Proof`? (I guess no, but double-checking).

consensus/state_test.go
```diff
@@ -310,14 +310,21 @@ func TestStateOversizedBlock(t *testing.T) {
 				totalBytes += len(part.Bytes)
 			}
 
+			maxBlockParts := maxBytes / int64(types.BlockPartSizeBytes)
+			if maxBytes > maxBlockParts*int64(types.BlockPartSizeBytes) {
+				maxBlockParts += 1
+			}
+			numBlockParts := int64(propBlockParts.Total())
```
sergio-mena, line 317:
```suggestion
			maxBlockParts := (maxBytes-1) / int64(types.BlockPartSizeBytes) + 1
			numBlockParts := int64(propBlockParts.Total())
```

How about this instead?

cason: I prefer using a different method, even if it is less efficient/smart, when testing.

types/part_set.go
cason: It should not be possible, as the next check (line 40) would fail. This check, however, is also a valid check: integer fields are >= 0, hashes have the right length.

consensus/state_test.go
sergio-mena: Good point

types/part_set.go
cason: So, status regarding this:
- we can trust `Proof.Index`, as change it causes the proof verification to fail
- we can't trust `Proof.Total`, as it is irrelevant for the proof verification provided that `Proof.Index < Proof.Total`.

cason: Ok, now `Proof.Total` should be correct, otherwise the `Part` is not added. Lets check if something breaks due to this additional check.

sergio-mena: Excellent!

sergio-mena (approved):
I was convinced I had already approved this (sorry for delay)

cason:
> Is this going to be backported? If so, to which version(s)?

I would say all releases. Why?

thanethomson:
> I would say all releases. Why?

There are currently no backport labels on the PR.

Let's get this over the finish line please.

cason:
> There are currently no backport labels on the PR.

Sorry, my bad. @sergio-mena, can you confirm we are backporting this to all releases?

sergio-mena:
As this comes from our work in a security advisory, I'd say we should backport it to all releases

thanethomson:
What's blocking the merge here @cason?

andynog:
Moving this to `In Progress` since looks like more input is needed here. When it's `Ready for review` again please change the status.

consensus/state_test.go
```diff
@@ -310,14 +310,21 @@ func TestStateOversizedBlock(t *testing.T) {
 				totalBytes += len(part.Bytes)
 			}
 
+			maxBlockParts := maxBytes / int64(types.BlockPartSizeBytes)
+			if maxBytes > maxBlockParts*int64(types.BlockPartSizeBytes) {
```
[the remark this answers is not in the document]
sergio-mena: IIRC, @cason wanted to have it calculated differently in the test code and in the prod code, and I agreed with him

cason:
So, the main question that worries me regarding this PR is the following: is it a good idea to enforce block parts to have exactly `types.BlockPartSizeBytes`?

I can think on reasons for a proposer to opt for smaller parts, for performance or network dissemination reasons. And this should not be considered a bad behavior. In any case we already reject, at the reactor level, parts that are bigger than that.

The main reason we had for enforcing a fixed size for inner block parts is to infer the size of a proposed `Block` from the `BlockID` included in the associated `Proposal`, as its size would be at least its `(BlockID.PartSetHeader.Total- 1) x types.BlockPartSizeBytes`. The point is that, while less precise, we still get a decent upper bound for the size of a proposed block if we don't enforce a fixed size for inner block parts. We can still use the same logic to detect proposals for blocks that cannot be accepted in any way, despite the size of the block parts.

Notice that in any case, as reported in some comments, this logic to enforce all but the last block parts to have a fixed size is kind of hacky, because we use indexes and total number of parts from the Merkle proof etc. We could make this less hacky if we moved this verification from `Part.ValidateBasic()` to the consensus logic.

The last point, also for the same reasoning, is that types `Part` and `PartSet` are public, exported, and we don't really know the impact on enforcing this requirement for them. As the changes in our test units indicate, there is the possibility and maybe it is even common to build part set using different part sizes.

I would like some feedback on this before rendering this PR ready to merge.

melekes:
> The main reason we had for enforcing a fixed size for inner block parts is to infer the size of a proposed Block from the BlockID included in the associated Proposal,

Is this a reason for a fixed size or capping maximum size? I.e. why do we need a more precise value for a proposal block size?

cason:
> Is this a reason for a fixed size or capping maximum size? I.e. why do we need a more precise value for a proposal block size?

For fixed, not a maximum. I had the same confusion while writing. The introduced logic rejects the `Proposal` for a block when

`[num of parts - 1] x [part size] > [maximum configured block size]`

If someone uses smaller part sizes, we might reject proposals for blocks whose size does not exceed the configured maximum.

melekes:
> > Is this a reason for a fixed size or capping maximum size? I.e. why do we need a more precise value for a proposal block size?
> 
> For fixed, not a maximum. I had the same confusion while writing. The introduced logic rejects the `Proposal` for a block when
> 
> `[num of parts - 1] x [part size] > [maximum configured block size]`
> 
> If someone uses smaller part sizes, we might reject proposals for blocks whose size does not exceed the configured maximum.

In that case enforcing size makes sense to me 👍

sergio-mena:
@cason Did we see a valid use case being used on the filed where the proposer is using block parts of sizes different from the default? (except the last block part, obviously)

faddat:
I think this PR is a great idea.

faddat (approved):
I really like this approach as a way to further defend against storms.

melekes (approved):

cason:
Ok, lets land this on `main` like it is. I will propose another way to implement the same without forcing generic `PartSet`s to adhere to the sizes required by consensus.

cason:
A potential use-case for smaller block parts: #1776

Issue #1058, opened 2023-06-30
Analyze, prototype and experiment easily applicable solutions to reduce bandwidth usage


lasarojc:
## Problem

The bandwidth consumption in CometBFT is high. In particular, this comes from the fact that the reactor of the mempool is a basic flooding gossip protocol (à la Bitcoin). Because the network topology is dense, it is very likely that a node receives a transaction multiple times from different neighbors. Notice that such a redundancy is useful to prevent byzantine or selfish nodes to stop the dissemination of a transaction in the network.

Some evidence of this redundancy is seen in #613

## Target audience

Consensus developers

## Upside

Small improvements could cut bandwidth usage in CometBFT by (at least) 25%, based on other results.

## Downside

There are many possible improvements to test. To mitigate the risk of spending too much time on discussing improvements, we will divide the improvements into two batches, one with quick wins and one with ideas that require further though and refinement.

- [x] #1086

For each of the approaches that are chosen on the brainstorm we will 

- prototype the approach
- experiment with improvements using e2e and fast prototyping
- experiment with the approaches that turned out promising in the QA infra
- document the solutions as ADRs

The remaining ideas will become part of #11 and if there is room to try these and other ideas on that issue, we will iterate on brainstorming and prototyping.

Consider if #1048 is pre-requisite for this task of if backlog.

## Definition of done

- Provide in the form of an ADR a new gossip protocol, or an improvement to the current one, that reduces the bandwidth of mempool transactions, as demonstrated by experiments on a testnet that resembles as close as possible to a real-world mainnet.
- Implement, test, and merge into main the solution proposed in the ADR.
- Protocols or ideas discussed but not implemented or prototyped should also be documented.

hvanz:
Tentative list of improvements to the propagation protocol: 

1. For each transaction, keep track of the nodes that are known to have the transaction (currently it is only the senders)
1. Reply with TxAck instead of Tx itself if Tx is large.
1. Random sleep during gossip to allow receiving Tx before having to send it.
1. Send HaveTxId instead of Tx itself, depending on Tx size, to warn that Tx is not required.
1. Send HaveTxIdSet (BF) instead of Tx themselves, depending on the Tx set size, to let neighbors know which Txs are required.
1. Propose TxId instead of Tx and accept a block only when the Tx with TxId is known.
[...]

PR #1425, opened 2023-10-03
Allow skipping transactions when forwarding transactions

### LINKED ISSUE #1149: Investigate the effects of randomly skipping transactions in the tx gossiping.
As an alternative to throttling as proposed in #1148 , this issue investigates the effect of randomly skipping some transactions during the gossiping.

On the positive side, it leads to less bw usage. On the downside it could lead to less nodes being reached by transactions and on the FIFO ordering of channels being invalidated.

lasarojc:
Contributes to #1058
Closes #1149 

Adds a parameter to define a rate in which to skip transactions when forwarding; the default value is 0, the same as the current behavior. 

Also adds the receiver node to the senders list as soon as a transactions is successfully send or skipped.

This is a port of some changes made in the context of #1149, more specifically in this [branch](https://github.com/cometbft/cometbft/tree/hvanz/cat-mempool)

---

#### PR checklist

- [x] Tests written/updated


lasarojc:
We may revisit this idea in the future, once application are able to handle gaps in sequence numbers.

PR #1427, opened 2023-10-03
mempool: export errors

4ever9:
Partially addresses: https://github.com/cometbft/cometbft/issues/1140

Exports errors in mempool.


mempool/mempool.go
```diff
@@ -191,3 +191,29 @@ func (e ErrPreCheck) Error() string {
...
+	return e.Reason.Error()
+}
+
+type ErrFlushAppConn struct {
+	Reason error
+}
+
+func (e ErrFlushAppConn) Error() string {
+	return e.Reason.Error()
+}
```
pysel, line 219: I think it is best to define these errors in `errors.go` file to follow the standard

mempool/mempool.go
```diff
@@ -191,3 +191,29 @@ func (e ErrPreCheck) Error() string {
 func IsPreCheckError(err error) bool {
 	return errors.As(err, &ErrPreCheck{})
 }
+
+var ErrTxNotFound = errors.New("transaction not found in mempool")
+
+type ErrCheckTxAsync struct {
+	Reason error
```
pysel, line 198: Personally, I would call this field just `Err`

pysel: also, I think exported errors that wrap another `error` structure should have an `Unwrap() error` method that returns in underlying error. It is useful in many contexts, to give an example - it is useful when we want to check if an error has internally a specific error wrapped when writing unit tests

pysel (changes requested):
hey @4ever9! thank you very much for your contribution! left a few comments:

mempool/mempool.go
thanethomson: Agreed, I'd recommend renaming `Reason` to `Err`.

mempool/mempool.go
thanethomson: Agreed!

thanethomson (commented):
Thanks for this @4ever9! Looking forward to getting this merged!

4ever9:
hey @pysel @thanethomson  , thanks for review and all comments addressed.

mempool/errors.go
```diff
@@ -0,0 +1,95 @@
...
+func (e ErrPreCheck) Unwrap() error {
+	return e.Err
+}
+
+// IsPreCheckError returns true if err is due to pre check failure.
+func IsPreCheckError(err error) bool {
+	return errors.As(err, &ErrPreCheck{})
+}
+
+var ErrTxNotFound = errors.New("transaction not found in mempool")
```
pysel, line 59: nit: move this on top of the file so that all variables are in the same place

mempool/errors.go
```diff
@@ -0,0 +1,95 @@
...
+
+type ErrCheckTxAsync struct {
+	Err error
+}
+
+func (e ErrCheckTxAsync) Error() string {
+	return e.Err.Error()
```
pysel, line 66: nit: I think it would be better to wrap underlying error in some additional text for context. Because without that, there is no much reason to wrap an error. It is not critical, but rather my personal opinion. 

example: https://github.com/cometbft/cometbft/blob/main/blocksync/errors.go#L47-L49

pysel (approved):
lgtm! just left a few nits

mempool/errors.go
4ever9: @pysel thanks for your review, addressed.

mempool/errors.go
4ever9: Addressed.

thanethomson (approved):

hvanz:
@Mergifyio backport v0.38.x

PR #1436, opened 2023-10-05
indexer-respect-height-params-on-query

### LINKED ISSUE #1435: KV search by events ignores second height parameter if there is a event parameter
### Setup

**CometBFT version**: v0.37.2

**Have you tried the latest version**: no, but based on the code, the bug should be reproducible in the newest version as well.

**Environment**:
- **OS**: Ubuntu 22.04

### What happened?
I'm trying to query transactions using txSearch. I'm using three parameter: tx.height >= X AND tx.height <= Y and event.key = value. In this case the second tx.height is ignored. If event.key=value is missing then both tx.height params are respected.

The query should respect all of the parameters.

### How to reproduce it
Add two transactions to the indexer. One should be at height X and the other - at height Y. Let's both transactions have an event with key "account" and value "number" equals "1". Making a query tx.height >= X AND tx.height <= Y-1 and account.number = 1 MUST return only the transaction at height X but now it returns both transactions.

kstoykov:
Closes #1435

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/.changelog/unreleased/bug-fixes/indexer-respect-height-params-on-query.md
+++ b/.changelog/unreleased/bug-fixes/indexer-respect-height-params-on-query.md
@@ -1,0 +1,3 @@
+### BUG FIXES
+
+- `[state/indexer]` Respect both height params while quering for events
--- a/state/indexer/query_range.go
+++ b/state/indexer/query_range.go
@@ -87,14 +87,13 @@ func (qr QueryRange) UpperBoundValue() interface{} {
 func LookForRangesWithHeight(conditions []query.Condition) (queryRange QueryRanges, indexes []int, heightRange QueryRange) {
 	queryRange = make(QueryRanges)
 	for i, c := range conditions {
-		heightKey := false
+		heightKey := c.CompositeKey == types.BlockHeightKey || c.CompositeKey == types.TxHeightKey
 		if IsRangeOperation(c.Op) {
 			r, ok := queryRange[c.CompositeKey]
 			if !ok {
 				r = QueryRange{Key: c.CompositeKey}
 				if c.CompositeKey == types.BlockHeightKey || c.CompositeKey == types.TxHeightKey {
 					heightRange = QueryRange{Key: c.CompositeKey}
-					heightKey = true
 				}
 			}
 
--- a/state/txindex/kv/kv_test.go
+++ b/state/txindex/kv/kv_test.go
@@ -322,6 +322,89 @@ func TestTxSearchEventMatch(t *testing.T) {
 		})
 	}
 }
+
+func TestTxSearchEventMatchByHeight(t *testing.T) {
+
+	indexer := NewTxIndex(db.NewMemDB())
+
+	txResult := txResultWithEvents([]abci.Event{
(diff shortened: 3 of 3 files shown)


.changelog/unreleased/bug-fixes/indexer-respect-height-params-on-query.md
```diff
@@ -0,0 +1,3 @@
+### BUG FIXES
+
```
thanethomson, line 2:
```suggestion
```

.changelog/unreleased/bug-fixes/indexer-respect-height-params-on-query.md
```diff
@@ -0,0 +1,3 @@
...
+- `[state/indexer]` Respect both height params while quering for events
```
thanethomson, line 3: Please ensure there's a link to this PR at the end of the changelog entry line, e.g. https://github.com/cometbft/cometbft/blob/58cfaf973bcd88808e130ea454fe7fc5015e3078/.changelog/unreleased/enhancements/1096-state-background-pruning.md

thanethomson (commented):
First pass minor review, the team will review the logic change ASAP.

thanethomson:
@kstoykov we generally do not submit PRs to specific release branches. We usually submit them to `main` and then backport to the relevant release branches accordingly.

kstoykov:
@thanethomson should I close this PR and make a new one to main branch?

state/indexer/query_range.go
```diff
@@ -87,14 +87,13 @@ func (qr QueryRange) UpperBoundValue() interface{} {
 func LookForRangesWithHeight(conditions []query.Condition) (queryRange QueryRanges, indexes []int, heightRange QueryRange) {
 	queryRange = make(QueryRanges)
 	for i, c := range conditions {
-		heightKey := false
+		heightKey := c.CompositeKey == types.BlockHeightKey || c.CompositeKey == types.TxHeightKey
```
jmalicevic, line 90:
```suggestion
		
```

state/indexer/query_range.go
```diff
@@ -87,14 +87,13 @@ func (qr QueryRange) UpperBoundValue() interface{} {
 func LookForRangesWithHeight(conditions []query.Condition) (queryRange QueryRanges, indexes []int, heightRange QueryRange) {
 	queryRange = make(QueryRanges)
 	for i, c := range conditions {
```
jmalicevic, line 89:
```suggestion
	queryRange = make(QueryRanges)
	heightKey := false
	for i, c := range conditions {
```

state/indexer/query_range.go
jmalicevic: The reason why I removed it here before the `RangeOperation` check is that it might still be a `height` key but not a range operation. That is then yet another special case we need to handle. Otherwise the fix is legit ,thanks a lot for catching and fixing this!

jmalicevic:
> @thanethomson should I close this PR and make a new one to main branch?

Hi, yes, we need this bugfix in main, so please submit it there, we'll backport it then. Thank you for this!

kstoykov:
I had to create a new branch (therefore and a new PR) in order to fix the bug on main branch. The new PR is: https://github.com/cometbft/cometbft/pull/1529

Issue #1421, opened 2023-10-02
ADR-101 v0.38.x : BaseService and VersionService backport


jmalicevic:
Backporting PR #818  to `v0.38.x-experimental` branch

Commit to backport: 65a88b9f8 #1437

Branch `v0.38.x` contains a **legacy gRPC** implementation. 

Branch `main` does not have the legacy gRPC, which has been removed, and #818 adds a new gRPC service.

When backporting #818 we have two options:

1. Remove the legacy gRPC and replace it by the new gRPC service implemented in `main`
    - For preserving the existing functionalities in `v0.38`, this approach requires implementing the existing two legacy gRPC endpoints (`BroadcastTx` and `Ping`) using the new gRPC service
2. Preserve the legacy gRPC implementation and add the new gRPC implemented in `main` as an additional service
    - In this case, 3 RPC endpoints could be available on a node: the standard JSON-RPC, the legacy gRPC, and the newly introduced gRPC service

Additional changes:
- [x] Update tests to ensure that the legacy gRPC endpoints works as before
- [x] Update e2e test infra to: in case of approach 1, add test for the legacy endpoint to ensure it works with the API; in case of approach 2, extend the e2e node to support 3 endpoints
- [x] Changelog entries explaining what is breaking compared to v0.38.x, if any. (there were none)

cason:
We are more inclined towards option (2) because:
 - The legacy gRPC is optional, and probably not employed by most users
 - The legacy gRPC service is deprecated, being removed from `main`
 - Implementing the legacy gRPC endpoints on the new gRPC server is an additional issue by itself

cason:
Backporting 3479642e2 at this point is not ideal, as there are other PRs associated to #1419 that should be considered first:

```
% git log --oneline rpc/grpc/ 
4f0d5b4f0 (origin/feature/adr101-pull-companion) gRPC for Tx and Block indexer pruning (#1327)
6fc2c37c8 ADR 101: Refactor height check-related logic and tests (#1271)
3479642e2 ADR 101: Add `Close` method to gRPC client (#1251)
69843bed0 (origin/jasmina/benchmark_pruning) ADR-101: implement gRPC `PruningService` (#1154)
8006a7456 ADR 101: Add GetLatest method to block service (#1209)
f9c17b45d Add BlockResults gRPC service (#1168)
84e2e01e6 Add gRPC block service (#1142)
65a88b9f8 grpc: Add base gRPC server with version service (#818)
```

[...]

PR #1437, opened 2023-10-05
grpc: Add base gRPC server with version service to v0.38.x-experimental

cason:
Addresses #1421. 

Manual backport of #818, which introduces a new base gRPC service. The new gRPC service is implemented under `rpc/grpc/client` and  `rpc/grpc/server`. The destination branch contains a minimal, legacy gRPC implementation that has been deprecated. It is implemented under `rpc/grpc` and is **preserved** by this PR.

So, this PR introduces a new base gRPC service that is supposed to co-exist with the existing/legacy gRPC service. The unit tests for the legacy gRPC service were configured to have also the new gRPC service enabled. The e2e test suite was updated to enable the legacy gRPC service, in addition to the new base gRPC service. These changes are intended to attest that both gRPC service implementation can properly co-exist.

#### PR checklist

- [x] Tests written/updated


jmalicevic (approved):
LGTM

test/e2e/pkg/infra/docker/docker.go
```diff
@@ -78,7 +78,9 @@ services:
     init: true
     ports:
     - 26656
-    - {{ if .ProxyPort }}{{ .ProxyPort }}:{{ end }}26657
+    - {{ if .RPCProxyPort }}{{ .RPCProxyPort }}:{{ end }}26657
```
hvanz, line 81: This change should also be applied below (in line 107).

test/e2e/runner/start.go
```diff
@@ -56,21 +56,9 @@ func Start(ctx context.Context, testnet *e2e.Testnet, p infra.Provider) error {
...
-			logger.Info("start", "msg",
-				log.NewLazySprintf("Node %v up on http://%s:%v; with Prometheus on http://%s:%v/metrics",
-					node.Name,
-					node.ExternalIP,
-					node.ProxyPort,
-					node.ExternalIP,
-					node.PrometheusProxyPort,
-				),
-			)
+			logger.Info("start", "msg", log.NewLazySprintf("Node %v up on http://127.0.0.1:%v; with Prometheus on http://127.0.0.1:%v/metrics", node.Name, node.RPCProxyPort, node.PrometheusProxyPort))
```
hvanz, line 59: The IP address is fixed here, instead of using the variable `node.ExternalIP` as in the original code. The same happens in the `logger.Info` line below.

test/e2e/pkg/infra/docker/docker.go
cason: I assume the block below applies to running a different version (`UpgradeVersion`), while this block applies to the current version (`Version`). This has to be confirmed. In thesis, they should enable the same interactions, using the same ports.

However, as this is a backport, we are not supposed to introduce new behaviors or features.

cason: The version in `main` does not include these ports on the second block as well.

test/e2e/pkg/infra/docker/docker.go
```diff
@@ -78,7 +78,9 @@ services:
...
+    - {{ if .GRPCProxyPort }}{{ .GRPCProxyPort }}:{{ end }}26670
+    - {{ if .GRPCLegacyPort }}{{ .GRPCLegacyPort }}:{{ end }}26671
```
cason, line 83: When compared to main, it appears that this port is in use for the gRPC privileged service:

```
- {{ if .GRPCPrivilegedProxyPort }}{{ .GRPCPrivilegedProxyPort }}:{{ end }}26671
```

So we either change this port number here on this PR, or we have to perform this change on future backports.

test/e2e/runner/start.go
cason: We should revert this change, in my opinion.

Actually, the diff of this file should match the output of the following command:

```
git diff main v0.38.x test/e2e/runner/start.go
```

cason: Namely, `node.ProxyPort` renamed to `node.RPCProxyPort` should be the only change here.

test/e2e/pkg/infra/docker/docker.go
jmalicevic: Yes this is a bug we need to fix in main

Issue #1330, opened 2023-09-11
proto: protobuf renaming and versioning


thanethomson:
We should publish our Protobuf definitions to the Buf registry. These should ideally be pushed via CI when we make changes to the protos from our release branches.

[We are not tracking here the issues taking care of publishing `tendermint` protobufs to the buf repository (`v0.34.x` to `v0.38.x`) because, (1) they are not trivial and (2) they are not blocking`v1.0`. The issue #1402, is in the backlog]

- [x] #94 
- [x] #95
- [x] #736 
- [x] #1530
- [x] #1504
- [x] #1555
- [x] #1561
- [x] #1834
- [x] #1403
- [x] #1649
- [x] #1650
- [x] #1672 (merge feature branch into `main`)
- [x] #1733
  - [x] #1726 
  - [ ] Find a solution for dependencies (gogoproto, etc.)
  - [ ] (if keeping a copy in our repo) Goal: the files should be the same (published vs. released from Github)
  - [ ] [stretch] ideally, automatically via CI when tagging releases or pre-releases (see the pre-release and release workflows in the `.github/workflows` folder).

mzabaluev:
All tasks are done.

PR #1443, opened 2023-10-06
Merge main into feature/proto-upgrade

mzabaluev:
Contributes to #1330

Update the feature branch to the recent changes in main.

Of particular interest are the protobuf changes finalized in the
0.38.0 release, and the addition of the ADR-101 gRPC services.


mzabaluev:
Commits starting from 30ddb516845c5ed0943963d20bec6a595818d60c are reviewable.

api/cometbft/consensus/v1/message.go
```diff
@@ -38,12 +38,6 @@ func (m *HasVote) Wrap() proto.Message {
 	return cm
 }
 
-func (m *HasProposalBlockPart) Wrap() proto.Message {
```
sergio-mena: Not sure I understand this hunk

api/cometbft/consensus/v1/message.go
```diff
@@ -105,9 +99,6 @@ func (m *Message) Unwrap() (proto.Message, error) {
 	case *Message_HasVote:
 		return m.GetHasVote(), nil
 
-	case *Message_HasProposalBlockPart:
```
sergio-mena: Same here

api/cometbft/consensus/v1/message.go
sergio-mena: Got it. It was moved to `v2`

consensus/msgs.go
```diff
@@ -7,81 +7,82 @@ import (
...
 	"github.com/cometbft/cometbft/types"
 )
 
-// Takes a consensus message type and returns the proto defined consensus message,
-// wrapped in the discriminating Message container.
-func MsgToWrappedProto(msg Message) (cmtcons.Message, error) {
-	pb := cmtcons.Message{}
+// MsgToProto takes a consensus message type and returns the proto defined consensus message.
+//
+// TODO: This needs to be removed, but WALToProto depends on this.
```
sergio-mena, line 20: Can we open an issue on this?

sergio-mena (approved):
Reviewed the merge commit as a 3-way diff.
Reviewed the rest of commits on this PR

It all looks good to me, thanks for this! 🙏

As this will land as a "merge", please collapse 30ddb51 and following in one commit, then (optionally) "amend" the merge commit with the resulting collapsed commit.

consensus/msgs.go
mzabaluev: Sure, however this comment is present in main, I just merged it here.

mzabaluev:
Some dynamically-typed code ended up with a wrong message type again, will fix.

consensus/msgs.go
```diff
@@ -115,10 +115,10 @@ func MsgToProto(msg Message) (proto.Message, error) {
 			vsb.Votes = *bits
 		}
 
-		pb = vsb
+		pb.Sum = &cmtcons.Message_VoteSetBits{VoteSetBits: vsb}
 
 	default:
-		return nil, ErrConsensusMessageNotRecognized{msg}
+		return pb, ErrConsensusMessageNotRecognized{msg}
```
sergio-mena, line 121: Why this change? (from `nil` to `pb`)
If `pb` is `nil` in this code path, then `nil` is more readable

mzabaluev: The helper function got changed to return a struct value rather than a type-erased `proto.Message` interface pointer, eliminating a redundant roundtrip of a `Wrap` call and a dynamic downcast.

mzabaluev:
The remaining failure looks like some jitter in pruning service tests.

sergio-mena (approved):
Re-approving

mempool/clist_mempool.go
```diff
@@ -310,6 +315,9 @@ func (mem *CListMempool) globalCb(req *abci.Request, res *abci.Response) {
 			}
 			mem.metrics.RecheckTimes.Add(1)
 			mem.resCbRecheck(req, res)
+
+		default:
+			panic("unexpected value of RequestCheckTx.type")
```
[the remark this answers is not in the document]
mzabaluev: This should never happen because the `Type` field is set elsewhere in the same module before the struct value is passed to the callback. Added the formatting just to be more informative.

PR #1463, opened 2023-10-09
crypto: export errors

tshakalekholoane:
Exports errors from the `crypto` module per #1140 for the following submodules:

- [x] `armor`
- [x] `ed25519`
- [x] `encoding`
- [x] `merkle`
- [x] `sr25519`
- [x] `xchacha20poly1305`
- [x] `xsalsa20symmetric`

On an unrelated note: I noticed that the following `break` statement is [ineffectual if the intention is to break out of the outer loop](https://go.dev/ref/spec#Break_statements). I am not sure if this posses a problem.

https://github.com/cometbft/cometbft/blob/77fa8a00c524474a055d95bd5d522491c3bdbea3/crypto/merkle/proof.go#L207-L217

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/crypto/armor/armor.go
+++ b/crypto/armor/armor.go
@@ -8,28 +8,32 @@ import (
 	"golang.org/x/crypto/openpgp/armor" //nolint: staticcheck
 )
 
-// EncodeError represents an error from calling [EncodeArmor].
-type EncodeError struct{ Err error }
+// ErrEncode represents an error from calling [EncodeArmor].
+type ErrEncode struct {
+	Err error
+}
 
-func (e *EncodeError) Error() string {
+func (e *ErrEncode) Error() string {
 	return fmt.Sprintf("armor: could not encode ASCII armor: %v", e.Err)
 }
 
-func (e *EncodeError) Unwrap() error { return e.Err }
+func (e *ErrEncode) Unwrap() error {
+	return e.Err
+}
 
 func EncodeArmor(blockType string, headers map[string]string, data []byte) (string, error) {
 	buf := new(bytes.Buffer)
 	w, err := armor.Encode(buf, blockType, headers)
 	if err != nil {
-		return "", &EncodeError{Err: err}
+		return "", &ErrEncode{Err: err}
 	}
 	_, err = w.Write(data)
 	if err != nil {
-		return "", &EncodeError{Err: err}
+		return "", &ErrEncode{Err: err}
 	}
 	err = w.Close()
 	if err != nil {
-		return "", &EncodeError{Err: err}
+		return "", &ErrEncode{Err: err}
 	}
 	return buf.String(), nil
 }
--- a/crypto/armor/armor.go
+++ b/crypto/armor/armor.go
@@ -13,11 +13,11 @@ type ErrEncode struct {
 	Err error
 }
 
-func (e *ErrEncode) Error() string {
+func (e ErrEncode) Error() string {
 	return fmt.Sprintf("armor: could not encode ASCII armor: %v", e.Err)
 }
 
-func (e *ErrEncode) Unwrap() error {
+func (e ErrEncode) Unwrap() error {
 	return e.Err
 }
 
@@ -25,15 +25,15 @@ func EncodeArmor(blockType string, headers map[string]string, data []byte) (stri
 	buf := new(bytes.Buffer)
 	w, err := armor.Encode(buf, blockType, headers)
 	if err != nil {
-		return "", &ErrEncode{Err: err}
+		return "", ErrEncode{Err: err}
 	}
 	_, err = w.Write(data)
 	if err != nil {
-		return "", &ErrEncode{Err: err}
+		return "", ErrEncode{Err: err}
 	}
 	err = w.Close()
 	if err != nil {
-		return "", &ErrEncode{Err: err}
+		return "", ErrEncode{Err: err}
 	}
 	return buf.String(), nil
 }
--- a/crypto/armor/armor.go
+++ b/crypto/armor/armor.go
@@ -8,21 +8,30 @@ import (
 	"golang.org/x/crypto/openpgp/armor" //nolint: staticcheck
 )
 
-func EncodeArmor(blockType string, headers map[string]string, data []byte) string {
+// EncodeError represents an error from calling [EncodeArmor].
+type EncodeError struct{ Err error }
+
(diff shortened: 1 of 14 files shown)


crypto/ed25519/ed25519.go
```diff
@@ -15,7 +15,18 @@ import (
...
 
-//-------------------------------------
+var (
+	ErrNotEd25519Key    = errors.New("ed25519: pubkey is not Ed25519")
+	ErrInvalidSignature = errors.New("ed25519: invalid signature")
+)
+
+// InvalidKeyLenError describes an error resulting from an passing in a
+// key with an invalid key in the call to [BatchVerifier.Add].
+type InvalidKeyLenError struct{ got, want int }
```
thanethomson, line 25: `ErrInvalidKeyLen` and please make these fields publicly accessible.

crypto/armor/armor.go
```diff
@@ -8,21 +8,30 @@ import (
 	"golang.org/x/crypto/openpgp/armor" //nolint: staticcheck
 )
 
-func EncodeArmor(blockType string, headers map[string]string, data []byte) string {
+// EncodeError represents an error from calling [EncodeArmor].
+type EncodeError struct{ Err error }
```
thanethomson, line 12: Please stick to the current code formatting convention for struct definitions:

```suggestion
type ErrEncode struct { 
	Err error
}
```

Same comment applies to all other error struct names.

crypto/merkle/proof.go
```diff
@@ -16,6 +16,24 @@ const (
 	MaxAunts = 100
 )
 
+var ErrMaxAuntsLenExceeded = fmt.Errorf("merkle: maximum aunts length, %d, exceeded", MaxAunts)
+
+type InvalidHashError struct{ Err error }
```
thanethomson, line 21: `ErrInvalidHash`

crypto/merkle/proof.go
```diff
@@ -16,6 +16,24 @@ const (
...
+
+func (e *InvalidHashError) Error() string {
+	return fmt.Sprintf("merkle: invalid hash: %s", e.Err)
+}
+
+func (e *InvalidHashError) Unwrap() error { return e.Err }
+
+type InvalidProofError struct{ Err error }
```
thanethomson, line 29: `ErrInvalidProof`

crypto/merkle/proof_key_path.go
```diff
@@ -82,11 +82,19 @@ func (pth KeyPath) String() string {
 	return res
 }
 
+type InvalidKeyError struct{ Err error }
```
thanethomson, line 85: `ErrInvalidKey`

crypto/sr25519/batch.go
```diff
@@ -10,6 +11,28 @@ import (
...
+
+func (e *InvalidKeyError) Error() string {
+	return fmt.Sprintf("sr25519: invalid public key: %v", e.Err)
+}
+
+func (e *InvalidKeyError) Unwrap() error { return e.Err }
+
+// InvalidSignatureError wraps an error that could occur as a result of
+// generating an invalid signature.
+type InvalidSignatureError struct{ Err error }
```
thanethomson, line 28: `ErrInvalidSignature`

thanethomson:
Please also run the linter (`make lint` from the root of the repo) prior to committing and pushing changes.

tshakalekholoane:
You're welcome, @thanethomson!

Sorry about that. I was under the impression that a [different naming convention is used for structured errors](https://github.com/uber-go/guide/blob/master/style.md#error-naming) (which also what the standard library appears to do). But I've pushed the requested changes.

> Please also run the linter (`make lint` from the root of the repo) prior to committing and pushing changes.

The linter is flagging the use of naked returns which is unrelated to the changes I made in this PR. Do you want me to fix it anyway?

crypto/armor/armor.go
```diff
@@ -8,21 +8,34 @@ import (
 	"golang.org/x/crypto/openpgp/armor" //nolint: staticcheck
 )
 
-func EncodeArmor(blockType string, headers map[string]string, data []byte) string {
+// ErrEncode represents an error from calling [EncodeArmor].
+type ErrEncode struct {
+	Err error
+}
+
+func (e *ErrEncode) Error() string {
```
thanethomson, line 16: Ah, just noticed this now after a second pass. This should be:

```suggestion
func (e ErrEncode) Error() string {
```

We don't need to mutate any fields once an error's been constructed. Doing this also makes error construction slightly less cumbersome and more ergonomic (e.g. `ErrEncode{Err: err}` as opposed to `&ErrEncode{Err: err}`).

Applies to all the error types here.

thanethomson:
> The linter is flagging the use of naked returns which is unrelated to the changes I made in this PR. Do you want me to fix it anyway?

That's odd - I don't get any linter errors when running `make lint` on `main`. Any idea as to why that'd be?

crypto/armor/armor.go
tshakalekholoane: But that incurs a copy right? But I suppose the cost is negligible. I will change that as well.

tshakalekholoane:
> That's odd - I don't get any linter errors when running `make lint` on `main`. Any idea as to why that'd be?

I don't know. I am guessing the lint was added after the last change to the file? I think the checks are only done on modified files instead of the entire repository which would explain why it could only pick it up after a change was made albeit in an unrelated part.

> The linter can be a little finicky. If it fails, please do fix the lints.

I pushed the fix so it should pass now.

crypto/encoding/codec.go
```diff
@@ -33,7 +54,7 @@ func PubKeyToProto(k crypto.PubKey) (pc.PublicKey, error) {
 			},
 		}
 	default:
-		return kp, fmt.Errorf("toproto: key type %v is not supported", k)
+		return kp, &ErrUnsupportedKey{Key: k}
```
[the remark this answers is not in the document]
tshakalekholoane: Oh. My bad. Fixed.

crypto/armor/armor.go
thanethomson: Yeah it does, but it's negligible relative to the simplicity/readability/ergonomic benefits. If we had evidence that it was really hurting performance we could optimize specific cases, but there isn't any such evidence as of yet.

thanethomson (approved):

Issue #1468, opened 2023-10-11
replay missing begin block


forcodedancing:
### Setup

**CometBFT version** (use `cometbft version` or `git rev-parse --verify HEAD` if installed from source):
v0.37.2

**Have you tried the latest version**: no

**ABCI app** (name for built-in, URL for self-written if it's publicly available):

**Environment**:
- **OS** (e.g. from /etc/os-release): 
- **Install tools**:
- **Others**:

**node command runtime flags**:

### What happened?

When replay with mock app, the original implementation misses the logic for `begin block`. 

### Anything else we need to know

As we can see https://github.com/cometbft/cometbft/blob/v0.37.2/consensus/replay_stubs.go here, the begin block is not handled.

We should add something like this:

```
func (mock *mockProxyApp) BeginBlock(req abci.RequestBeginBlock) abci.ResponseBeginBlock {
	mock.txCount = 0
	return *mock.abciResponses.BeginBlock
}
```

cason:
Hey, thank you for your report.

Can you elaborate on why the lack of this method constitutes a bug? 

Have you identified a scenario where the lack of the proposed method produces an error or unexpected behavior?

andynog:
Hi @forcodedancing, just following up on this issue. Did you get a chance to review what @cason posted above ? Just want to check if you had a chance to review and if you're OK with his explanation and if we should close this issue. 

Or if not clear or if you still have concerns, could you please provide additional information so we can still keep assessing it?

PR #1469, opened 2023-10-11
Fix replay issue with mock app

forcodedancing:
Fix issue: https://github.com/cometbft/cometbft/issues/1468

When replaying with mock app, the original does not handle begin block - it will lead the begin block logic and block results are different from the real application's.

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/consensus/replay_stubs.go
+++ b/consensus/replay_stubs.go
@@ -86,6 +86,11 @@ func (mock *mockProxyApp) DeliverTx(req abci.RequestDeliverTx) abci.ResponseDeli
 	return *r
 }
 
+func (mock *mockProxyApp) BeginBlock(req abci.RequestBeginBlock) abci.ResponseBeginBlock {
+	mock.txCount = 0
+	return *mock.abciResponses.BeginBlock
+}
+
 func (mock *mockProxyApp) EndBlock(req abci.RequestEndBlock) abci.ResponseEndBlock {
 	mock.txCount = 0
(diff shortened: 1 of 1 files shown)


thanethomson:
Thanks for this @forcodedancing. We typically work by submitting a PR targeting `main`, which we then backport to the relevant release branches, as opposed to targeting a specific release branch.

Please re-target your PR to `main` and the team will review.

forcodedancing:
@thanethomson Thanks a lot.

The issue is - there are significant changes from v0.37 to the main branch, for the part. It could be invalid for the main branch now.

Could you help to evaluate the issue and fix firstly?

Thanks~

cason:
> it will lead the begin block logic and block results are different from the real application's.

Can you elaborate on this?

forcodedancing:
> > it will lead the begin block logic and block results are different from the real application's.
> 
> Can you elaborate on this?

image
For example, in normal there will be some begin block events in `block_results` rpc.

If we replay will mock app without begin block handler, the begin block events will be empty.

cason:
Ok, I see your point.

But for having a PR fixing a bug, we would need a test unit that fails due to the bug, then a solution for the bug.

Regarding the discussion of whether this is or not a bug, it should be carried out on the original issue #1468.

melekes:
Superseded by https://github.com/cometbft/cometbft/pull/1782

PR #1472, opened 2023-10-11
mempool: Add push-pull gossip protocol (CAT)

hvanz:
Closes: #2027.

Relates to #1058.

The CAT (for Content-Addressable Transaction) pool is a gossip protocol for the mempool originally implemented by Celestia. CAT is a push-pull protocol in contrast to CometBFT's default push protocol.

The code in this PR was ported from Celestia's [feature/cat](https://github.com/celestiaorg/celestia-core/tree/feature/cat) branch. The original protocol is built on top of the priority mempool implementation (aka `v1`), which existed in CometBFT until v0.37. The current code was ported on top of CometBFT's default mempool implementation (`CListMempool`), so we had to make to some changes to adapt it to the different underlying implementation.

---

#### PR checklist

- [ ] Tests written/updated


hvanz:
These are preliminary results on some experiments run on a laptop using the e2e framework. For more thorough results we would still need to run these experiments with around 200 nodes in the cloud, as we do for the [QA tests](https://github.com/cometbft/cometbft/blob/869833ee12c8c399fb0c487befcdac427236be2b/docs/qa/CometBFT-QA-38.md), probably also with a different network topology. 

Here, each experiment instance has:
- 8 nodes, in a complete graph topology, that is, each node is connected to all the others.
- The first node receives all the transaction load.
- Each run lasts 4 minutes.

Instances are defined by the permutation of transaction load rate (`r`) in tx/s and transaction size (`s`) in bytes, with the following values:
- `r = [100, 200, 400, 800]`
- `s = [256, 512, 1024, 2048]`

For each `(r,s)` value, we run two consecutive instances with the CAT mempool and then the default mempool (aka `v0`) as a baseline for comparison, one after the other.

Screenshot 2023-10-13 at 23 31 07

Screenshot 2023-10-16 at 09 26 04

We can see in the above graphs that: 
- Bandwidth consumption (bytes sent and received) of CAT is always lower than v0, with v0 being 3 to 6 times larger. 
- The chain height, chain size, and block size are, in each instance, almost the same value for CAT and v0, meaning that CAT does not miss transactions when compared to v0, or that all transactions sent to the nodes are eventually included in the chain. For big transactions, the chain size of v0 is lower than CAT. This is expected as CAT is supposed to work better with large transactions. 
- Remember that the saturation point defined in [the last QA experiments](https://github.com/cometbft/cometbft/blob/869833ee12c8c399fb0c487befcdac427236be2b/docs/qa/CometBFT-QA-38.md) is r=400, s=1024, meaning that the performance of a node is degraded for that value or bigger. This can be seen in the mempool size metrics, where for higher values the nodes become unstable.
- The metric 'Already received txs' counts the number of times a received transaction is already in the mempool cache. In all CAT instances, this metric is zero: the node receiving transactions from the client and pushing to the other nodes does not receive the transactions back from them. And the other nodes receive less duplicated transactions than with v0.

hvanz:
These are the same experiments as above but with the following topology, where node 1 receives all the load:
Screenshot 2023-10-16 at 17 28 29

Screenshot 2023-10-16 at 17 39 39

Screenshot 2023-10-16 at 17 39 50

These graphs look similar to the graphs of the above experiment. The main difference is that now the bandwidth of Flood is 2 times higher than that of Flood, instead of 3 to 6 times. This was expected, as now the push-pull protocol needs to transmit the transaction in various steps from node 1 to node 8.

config/config.go
```diff
@@ -840,6 +840,11 @@ type MempoolConfig struct {
 	// the $CMTHOME env variable or --home cmd flag rather than overriding this
 	// struct field.
 	RootDir string `mapstructure:"home"`
+	// Mempool reactor, valid options:
+	// - "cat": reactor with push-pull gossip protocol
+	// - "" (default): reactor with flooding protocol
+	Reactor string `mapstructure:"reactor"`
```
thanethomson, line 846: I would not call this parameter `reactor` - that's a concept that's internal to the CometBFT architecture that shouldn't be leaked to operators (such leakage is bad UX).

Operators would probably understand this better as `type` (as in the user story: "as an operator, I can select which type of mempool to run").

config/toml.go
```diff
@@ -380,6 +380,11 @@ dial_timeout = "{{ .P2P.DialTimeout }}"
 #######################################################
 [mempool]
 
+# Mempool reactor. Valid options:
+# - "cat" for push-pull gossip protocol
+# - "": default reactor with flood protocol
+reactor = "{{ .Mempool.Reactor }}"
```
thanethomson, line 386:
```suggestion
# Mempool type. Valid options:
# - "cat" for a mempool that uses the push-pull gossip protocol
# - "": default mempool with flooding gossip protocol
type = "{{ .Mempool.Type }}"
```

config/config.go
hvanz: I agree. I renamed it though to `gossip_protocol`, which I think it's more precise.

mempool/sync_reactor.go
```diff
@@ -0,0 +1,50 @@
+package mempool
+
+import (
+	"sync/atomic"
+
+	cfg "github.com/cometbft/cometbft/config"
+	"github.com/cometbft/cometbft/p2p"
+)
+
+type SyncReactor interface {
```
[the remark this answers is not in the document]
faddat: ok

mempool/sync_reactor.go
```diff
@@ -0,0 +1,50 @@
...
+	p2p.Reactor
+
+	WaitSync() bool
+
+	EnableInOutTxs()
+}
+
+type BaseSyncReactor struct {
```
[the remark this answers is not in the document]
faddat: ok

faddat:
but I must kindly ask that we backport this.

It is **badly** needed.

adizere:
> but I must kindly ask that we backport this.
> 
> It is **badly** needed.

Agree. Thanks Jacob! 

Is there a specific app-chain team that is waiting on it? If so, my bad, I was not aware!

Even more importantly, are Celestia mainnet nodes employing the CAT mempool? Our last chat with their team (Oct/Nov) we agreed we'd push CAT over the finish line if they give us the green light they will use it in their mainnet, and the Comet+Celestia teams will do complementary testing of CAT. But I have not re-checked since then so my info is very stale. We should get an update. cc @cmwaters

adizere:
@hvanz should we deprecate this PR in favor of #1971 ?

adizere:
Then deprecate 1971 in favor of present PR? Not sure 1971 still has anything worth keeping, let me know if so. cc @faddat 

BTW thanks for bringing the present PR up-to-date with `main` Hernán !

faddat:
Big thanks from me too. 

😍

I'll gladly close the other.

adizere:
For the moment deprioritized in favor of https://github.com/cometbft/cometbft/issues/3297

Issue #1470, opened 2023-10-11
e2e: Fix flakiness due to pruning


jmalicevic:
### Setup

unreleased - main

**Have you tried the latest version**: yes/no
yes
**ABCI app** (name for built-in, URL for self-written if it's publicly available):

### What happened?

Due to the pruning happening in the background, the e2e nightly tests in main fail occasionally because a pruned height is requested. 

The test does not always fail and usually passes on consecutive runs.

PR #1480, opened 2023-10-11
v0.38.x-experimental: Backport e2e flakiness fix

jmalicevic:
Manual backport of #1467

---

#### PR checklist

- [ ] Tests written/updated


cason:
So we are just skipping the tests when data companion is running?

jmalicevic:
Yes for the moment, we might figure out a better way to test this meaningfully, but right now there is no way to make the tests deterministic.

sergio-mena:
@jmalicevic Is there an issue so we don't forget to do this?
e2e tests are the best thing we have to catch problems ATM, so I think it is important that they test a newly shipped feature such as the new pruning mechanism

jmalicevic:
The pruning mechanism is tested but it should not be tested here. The functionality of this test is to verify that a node has the same blocks as an archive node. Which is not really deterministically possible atm when pruning happens ad-hoc. 

I am however working on reducing the flakiness of the gRPC tests which suffer from the same problem and the solution for that might be appicable here as well. #1470

sergio-mena (approved):

thanethomson (approved):

PR #1490, opened 2023-10-12
state/pruner: Notify observer only when retain height changes

### LINKED ISSUE #1489: store: flaky test due to pruner notifying observer when no pruning happens
### Setup

main , v0.38.x-experimental

- 
**Environment**:
- **Install tools**:
- **Others**:

### What happened?

We noticed this test  failure frequently in CI : https://github.com/cometbft/cometbft/actions/runs/6495267396/job/17639883516?pr=1474#step:5:108 .
The particular test (`TestPruningService` within `store/store_test.go` ) tests whether pruning is triggered when both the data companion and the application set different retain heights. 

Before reaching line 668, the application and the data companion have set their retain heights so that all blocks up to height 1200 were pruned. 

In line [668](https://github.com/cometbft/cometbft/blob/7e85f389c912f3908004fa99932c5b176a05d3ed/store/store_test.go#L668) , the retain height set by the application (1300) should be ignored because the data companion's retain height is still set to 1200. 

Due to this, no pruning should happen until line 668 and we would expect the observer to be notified only when pruning actually happens. That is why the condition on line [676](https://github.com/cometbft/cometbft/blob/7e85f389c912f3908004fa99932c5b176a05d3ed/store/store_test.go#L676C2-L676C2)  assumes that the base height of the store is now the one set by the application in L668. 

When the data companion increases its retain height to 1350 in line [671](https://github.com/cometbft/cometbft/blob/7e85f389c912f3908004fa99932c5b176a05d3ed/store/store_test.go#L671), 1300 is the lowest of the two and the pruning service can now prune blocks up to height 1300. 

However, the pruning routine of the pruner is restarted before the code in line 671 is executed, no pruning is performed but the observer is notified, the pruning service sleeps for the duration of `pruningInterval` and the test reaches the check in line 676. 

At that point the base value of the store is still 1200 as the pruning service did not realize the change in retain heights. 

The observer should not have been notified of the new retain height in case no pruning happens. Therefore the test should wait on a new value on the observer channel until the pruning service wakes up and realizes that blocks up to 1300 can be pruned.

### How to reproduce it

The failure can be reproduced locally by setting the pruning interval to a very low value (1 millisecond for example). This will trigger the pruning service to start before the new retain height is set.

jmalicevic:
Closes #1489 

This should be backported to `v0.38.x-experimental`

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/state/pruner.go
+++ b/state/pruner.go
@@ -312,10 +312,12 @@ func (p *Pruner) pruneABCIResponses() {
 			return
 		default:
 			newRetainHeight := p.pruneABCIResToRetainHeight(lastRetainHeight)
-			p.observer.PrunerPrunedABCIRes(&ABCIResponsesPrunedInfo{
-				FromHeight: lastRetainHeight,
-				ToHeight:   newRetainHeight - 1,
-			})
+			if newRetainHeight != lastRetainHeight {
+				p.observer.PrunerPrunedABCIRes(&ABCIResponsesPrunedInfo{
+					FromHeight: lastRetainHeight,
+					ToHeight:   newRetainHeight - 1,
+				})
+			}
 			lastRetainHeight = newRetainHeight
 			time.Sleep(p.interval)
 		}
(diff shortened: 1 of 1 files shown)


thanethomson (approved):
Thanks for this @jmalicevic!

PR #1492, opened 2023-10-13
e2e: Fix flakiness in grpc tests due to pruning

### LINKED ISSUE #1470: e2e: Fix flakiness due to pruning

jmalicevic:
Closes #1470 

---

#### PR checklist

- [ ] Tests written/updated


andynog:
I've removed the first block test to prevent flakiness on the test results due to race conditions with the pruning mechanism

lasarojc (approved):
LGTM in that it reduces the race window considerably.
Might add a couple of retries to make it even less likely.
I wonder how ready clients are to deal with this race.

PR #1501, opened 2023-10-16
Make BuildExtendedCommitInfo and BuildLastCommitInfo public and change their signature slightly

p-offtermatt:
Related: https://github.com/cometbft/cometbft/discussions/1406#discussioncomment-7254221

Makes the BuildExtendedCommitInfo and BuildLastCommitInfo functions public,
and makes them take a validator set instead of the store as an input.

These functions  are useful to also call from outside of CometBFT if one has a need to obtain the CommitInfo structs,
in particular I am using them in a mock for cometbft.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/state/execution.go
+++ b/state/execution.go
@@ -166,12 +166,16 @@ func (blockExec *BlockExecutor) ProcessProposal(
 	block *types.Block,
 	state State,
 ) (bool, error) {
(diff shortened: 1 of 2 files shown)


p-offtermatt:
Closed and cherry-picked the commits against main

PR #1502, opened 2023-10-16
Make BuildLastCommitInfo and BuildExtendedCommitInfo public

p-offtermatt:
Makes the BuildExtendedCommitInfo and BuildLastCommitInfo functions public,
and makes them take a validator set instead of the store as an input.

These functions are useful to also call from outside of CometBFT if one has a need to obtain the CommitInfo structs,
in particular I am using them in a mock for cometbft.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/state/execution.go
+++ b/state/execution.go
@@ -174,9 +174,9 @@ func (blockExec *BlockExecutor) ProcessProposal(
 	block *types.Block,
 	state State,
 ) (bool, error) {
-	lastValSet, err := blockExec.store.LoadValidators(block.LastCommit.Height)
+	lastValSet, err := blockExec.store.LoadValidators(block.Height - 1)
 	if err != nil {
-		return false, fmt.Errorf("failed to load validator set at height %d, initial height %d: %w", block.LastCommit.Height, state.InitialHeight, err)
+		return false, fmt.Errorf("failed to load validator set at height %d, initial height %d: %w", block.Height-1, state.InitialHeight, err)
 	}
 	resp, err := blockExec.proxyApp.ProcessProposal(context.TODO(), &abci.RequestProcessProposal{
 		Hash:               block.Header.Hash(),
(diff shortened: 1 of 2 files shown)


state/execution.go
```diff
@@ -423,18 +441,13 @@ func (blockExec *BlockExecutor) Commit(
 //---------------------------------------------------------
 // Helper functions for executing blocks and updating state
 
-func buildLastCommitInfo(block *types.Block, store Store, initialHeight int64) abci.CommitInfo {
+func BuildLastCommitInfo(block *types.Block, lastValSet *types.ValidatorSet, initialHeight int64) abci.CommitInfo {
```
sergio-mena, line 444: Can we structure this function at two levels, the outer being private, the inner being public, so that:

* The outer does the loading of the validators
* The inner does what is does in your patch

This way we avoid repeating the extracted code at all callsites.

state/execution.go
```diff
@@ -473,17 +486,12 @@ func buildLastCommitInfo(block *types.Block, store Store, initialHeight int64) a
 // data, it returns an empty record.
 //
 // Assumes that the commit signatures are sorted according to validator index.
-func buildExtendedCommitInfo(ec *types.ExtendedCommit, store Store, initialHeight int64, ap types.ABCIParams) abci.ExtendedCommitInfo {
+func BuildExtendedCommitInfo(ec *types.ExtendedCommit, valSet *types.ValidatorSet, initialHeight int64, ap types.ABCIParams) abci.ExtendedCommitInfo {
```
sergio-mena, line 489: Same comment here about creating an inner and an outer function, to reduce the amount of code repeated at callsites

state/execution.go
p-offtermatt: Yes, great idea!

state/execution.go
```diff
@@ -435,6 +434,17 @@ func buildLastCommitInfo(block *types.Block, store Store, initialHeight int64) a
 		panic(fmt.Errorf("failed to load validator set at height %d: %w", block.Height-1, err))
 	}
 
+	return BuildLastCommitInfo(block, lastValSet, initialHeight)
+}
+
+// Like
```
sergio-mena, line 440: Please write a one-liner with doc

sergio-mena (approved):
🙏

state/execution.go
p-offtermatt: My bad, I missed finishing this one!

PR #1506, opened 2023-10-17
tiny: reorder expected and actual values

insumity:
Tiny change: reorder expected and actual values to be consistent with other error messages.


sergio-mena (approved):
Thanks @insumity for catching this!

PR #1511, opened 2023-10-18
qa: Add parameter to loadtime/report for displaying results in one line

hvanz:
Currently the result of `test/loadtime/cmd/report` is too verbose to be consumed easily by other applications or scripts. This new flag parameter `-oneline` will display to the standard output only the values in one line separated by commas, without the extra text.

---

#### PR checklist

- [ ] Tests written/updated


thanethomson:
Is the aim here to output the data to CSV? If so, using the `encoding/csv` package is probably best.

lasarojc (approved):

PR #1517, opened 2023-10-19
mempool: Handle mempool errors in response callback

### LINKED ISSUE #1516: mempool: Edge case mempool error isn't propagated
## Bug Report
`CheckTx` doesn't propagate some mempool-related errors. For example, after a successful `CheckTx`, just before a transaction is added to the mempool, there's a check to see if it isn't full. If it's full, the transaction is discarded, a message is logged, but the response doesn't contain any errors. **So, when using the Sync mode of broadcasting, it may return success even though the transaction isn't in the mempool.**

We are seeing this edge case in our load tests.

### Setup
**CometBFT version**: 0.37.2
**Have you tried the latest version**:  No, but the problem is the same.
**ABCI app**: https://github.com/warp-contracts/sequencer

janekolszak:
Closes #1516

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -386,7 +386,7 @@ func (mem *CListMempool) resCbFirstTime(
 				r.CheckTx = &abci.ResponseCheckTx{
 					Code:      CodeTypePostCheckFailed,
 					Codespace: Codespace,
-					Log:       err.Error(),
+					Log:       postCheckErr.Error(),
 				}
 			}
 		}
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 	"sync/atomic"
 
+	"github.com/Workiva/go-datastructures/threadsafe/err"
 	abcicli "github.com/cometbft/cometbft/abci/client"
 	abci "github.com/cometbft/cometbft/abci/types"
 	"github.com/cometbft/cometbft/config"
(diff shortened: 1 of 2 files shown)


cason:
The errors detected by the mempool in this solution are returned as they were produced by the ABCI application. I don't think this is the right way to go.

janekolszak:
Not exactly, `Codespace` is different so anyone debugging could easily see that this is a response from mempool. 

An alternative would be to:
- return an error from the callback (add it err to the interface) - but it wouldn't handle async use case 
- add error as callback argument next to `res`
- another async mechanism that does essentially what res does 

Let me know if you see a better solution.

janekolszak:
@cason Do you see any other solution?

cason:
The list of solutions you provide are probably the possible ones.

The point here is that the RPC endpoint should return an operation result/response and possibly an error. I am not convinced that the situations covered in this PR are not an error. If they are actually errors, the RPC endpoint should return errors, not a result/response with an encoded error inside.

With that in mind, a possible approach is to:

- Produce an error at mempool level, with a special `Namespace`, like you suggested
- Make the RPC endpoints to interpret these special errors differently, returning an error instead of a response instance

melekes:
@janekolszak are you still working on this?

janekolszak:
Nope, got discouraged and left :)

hvanz:
Closing PR as it's not clear that this is the correct approach to solve the issue.

PR #1518, opened 2023-10-20
Reduce the default MaxBytes to 4mb and increase MaxGas to 10 million

faddat:
21MB blocks are very rare, and problematic, they tend to look like this on live production networks:

image

This PR reduces max_bytes default value to 2mb, addressing part of this: https://forum.cosmos.network/t/amulet-security-advisory-for-cometbft-asa-2023-002 inexplicably published security notice. 

**most networks have 21mb max....**

The suggestion of 10,000,000 as default max gas comes from Simon from Confio, and the wasmd readme file.

https://x.com/simon_warta/status/1715303771020914763

---

#### PR checklist

- [ ] Tests written/updated


adizere:
Hey can you pls open an issue to document the need for this, briefly? Thanks!

We can include this in v1 release. We'll need to:
- document the rationale behind the 5MB choice;
- ensure user teams get a chance to weigh in and we choose a sensible value that makes sense to them; 
- add a changelog entry.

lasarojc (approved):

adizere:
Let's lock on a smaller value, eg. 2MB.

adizere:
@faddat can you add complete context here -- an issue and a problem description therein? We need this to take decisions regarding the PR and maintain the code in the future. Let's avoid drive-by PRs pls.

Our team will handle the changelog.

faddat:
@adizere very happy to.  Also, I'm totally open and okay with reducing the default max to 2mb.

faddat:
@adizere tbh I'd rather not expound the whole issue, I think that Amulet has done that sufficiently well.

webmaster128:
> Let's lock on a smaller value, eg. 2MB.

There are transactions that are significantly larger than bank sends. E.g. uploading a CosmWasm smart contract gives you ~400-800KB payloads in a single tx. Some chains increase a contract size limit to 3MB of Wasm. This might be gzipped down to 50%, leading to a single tx of ~1.5MB leaving verly little safety margin. Atomically storing two of those would already increase the 2MB. The furture of Cosmos is hopefully more feature rich and includes gaming, NFTs stored on chain and do on and so forth. Now you can argue chains should find a good value individually. But I think a good default should be both safe and not affect 90% of users. Given that I think 4 MB is a very good value as long as there is not good reason to go lower.

adizere:
Thanks for sharing Simon, that's super valuable perspective I wasn't aware of.

> But I think a good default should be both safe and not affect 90% of users. Given that I think 4 MB is a very good value as long as there is not good reason to go lower.

Solid advice. We'll stick to 4mb then.

andynog:
will add a changelog and update the docs before this can be merged

webmaster128 (approved):

PR #1529, opened 2023-10-23
indexer-respect-height-params-on-query

### LINKED ISSUE #1435: KV search by events ignores second height parameter if there is a event parameter

kstoykov:
Closes #1435

---

#### PR checklist

- [x] Tests written/updated
- [x] Changelog entry added in `.changelog`

### DIFF
--- a/.changelog/unreleased/bug-fixes/1529-indexer-respect-height-params-on-query.md
+++ b/.changelog/unreleased/bug-fixes/1529-indexer-respect-height-params-on-query.md
@@ -1,0 +1,1 @@
+- `[state/indexer]` Respect both height params while quering for events
--- a/state/indexer/query_range.go
+++ b/state/indexer/query_range.go
@@ -96,14 +96,13 @@ func (qr QueryRange) UpperBoundValue() interface{} {
 func LookForRangesWithHeight(conditions []syntax.Condition) (queryRange QueryRanges, indexes []int, heightRange QueryRange) {
 	queryRange = make(QueryRanges)
 	for i, c := range conditions {
-		heightKey := false
 		if IsRangeOperation(c.Op) {
+			heightKey := c.Tag == types.BlockHeightKey || c.Tag == types.TxHeightKey
 			r, ok := queryRange[c.Tag]
 			if !ok {
 				r = QueryRange{Key: c.Tag}
 				if c.Tag == types.BlockHeightKey || c.Tag == types.TxHeightKey {
 					heightRange = QueryRange{Key: c.Tag}
-					heightKey = true
 				}
 			}
 
--- a/state/txindex/kv/kv_test.go
+++ b/state/txindex/kv/kv_test.go
@@ -325,6 +325,88 @@ func TestTxSearchEventMatch(t *testing.T) {
 	}
 }
 
+func TestTxSearchEventMatchByHeight(t *testing.T) {
+
+	indexer := NewTxIndex(db.NewMemDB())
+
+	txResult := txResultWithEvents([]abci.Event{
+		{Type: "account", Attributes: []abci.EventAttribute{{Key: "number", Value: "1", Index: true}, {Key: "owner", Value: "Ana", Index: true}}},
+	})
+
+	err := indexer.Index(txResult)
+	require.NoError(t, err)
+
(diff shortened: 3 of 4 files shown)


state/indexer/query_range.go
```diff
@@ -96,14 +96,13 @@ func (qr QueryRange) UpperBoundValue() interface{} {
 func LookForRangesWithHeight(conditions []syntax.Condition) (queryRange QueryRanges, indexes []int, heightRange QueryRange) {
 	queryRange = make(QueryRanges)
 	for i, c := range conditions {
-		heightKey := false
 		if IsRangeOperation(c.Op) {
+			heightKey := c.Tag == types.BlockHeightKey || c.Tag == types.TxHeightKey
```
jmalicevic, line 100: Sorry, I know this was my proposed change but maybe we can delete it from this line and just leave it as is (it will be set in line 106). I am worried I am missing a scenario where we set it here where for some reason the rest of the return values are not set and then there is some unexpected behaviour afterwards. 

The initial bug was due to always setting it to false in L99 (which we are not doing anymore). Do you mind verifying this makes sense?

kstoykov: To be honest, I didn't completely understand your solution initially. I just got that the heightKey check should be after the "if IsRangeOperaction(c.Op)". That's why I put it inside the check.

Based on my understanding of the code the "heightKey" must be TRUE only if the current processing operation is "height" or "block", otherwise it must be FALSE. That's why "heightKey" must be initialized on the beginning of each loop, because it is possible to process an operation that is not "heightKey". 

For example, if there are three operations - "height", "account.number", "height" then on first iteration the "heightKey" must be TRUE, then FALSE and finally TRUE again. Does it make sense or I'm missing something?

jmalicevic: Actually you are right. If you were to not set the `heightKey` every time, it would not set the `heightRange` values correctly. If it were not for that, and we needed to set only `queryRange` it would be fine. Your solution is good the way it is now, thank you for the clarifications. 

This indexer  code is quite tricky and has a bunch of corner cases so it is very easy to get lost in all the possibilities.

.changelog/unreleased/bug-fixes/1529-indexer-respect-height-params-on-query.md
```diff
@@ -0,0 +1 @@
+- `[state/indexer]` Respect both height params while quering for events
```
jmalicevic, line 1:
```suggestion
- `[state/indexer]` Respect both height params while querying for events
   ([\#1529](https://github.com/cometbft/cometbft/pull/1529))
```

jmalicevic (approved):
LGTM, thanks a lot for the effort! 

@thanethomson can you double check the changelog layout before we mark it as ok to merge.

jmalicevic:
@kstoykov , can you please address the linting errors? Running `make lint` locally should give you a better idea where the errors are exactly.

sergio-mena:
The linting errors are likely to be the ones we just fixed, so updating the PR branch to latest main should do

jmalicevic:
> The linting errors are likely to be the ones we just fixed, so updating the PR branch to latest main should do

@kstoykov  as its your branch, I don't have permissions to merge into it... Its probably worth trying that first then before looking at the exact errors.

kstoykov:
@jmalicevic I've merge main branch to my branch and there are no more lint errors so everything should be fine now.

jmalicevic:
Thank you @kstoykov , tiggered a merge!

PR #1510, opened 2023-10-18
Implement ADR-107: rename proto package version suffixes to v1beta* pattern

mzabaluev:
Closes: #1504

---

#### PR checklist

- [ ] Tests written/updated
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments

PR #1531, opened 2023-10-23
Fix linting on `main`

sergio-mena:
As per title

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/libs/tempfile/tempfile_test.go
+++ b/libs/tempfile/tempfile_test.go
@@ -13,6 +13,8 @@ import (
 	cmtrand "github.com/cometbft/cometbft/libs/rand"
 )
 
+const tmpStr = "/tmp/"
+
 func TestWriteFileAtomic(t *testing.T) {
 	var (
 		data             = []byte(cmtrand.Str(cmtrand.Intn(2048)))
@@ -67,7 +69,7 @@ func TestWriteFileAtomicDuplicateFile(t *testing.T) {
 	atomicWriteFileRand = defaultSeed
 	firstFileRand := randWriteFileSuffix()
 	atomicWriteFileRand = defaultSeed
-	fname := "/tmp/" + atomicWriteFilePrefix + firstFileRand
+	fname := tmpStr + atomicWriteFilePrefix + firstFileRand
 	f, err := os.OpenFile(fname, atomicWriteFileFlag, 0777)
 	defer os.Remove(fname)
 	// Defer here, in case there is a panic in WriteFileAtomic.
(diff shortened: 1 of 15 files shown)


test/e2e/app/app.go
```diff
@@ -462,7 +462,7 @@ func (app *Application) ExtendVote(_ context.Context, req *abci.RequestExtendVot
 		ext = make([]byte, 8)
 		if num, err := rand.Int(rand.Reader, big.NewInt(voteExtensionMaxVal)); err != nil {
 			panic(fmt.Errorf("could not extend vote. Len:%d", len(ext)))
-		} else {
+		} else { //nolint: revive
```
lasarojc, line 465: Can we just move this one out of the else?

lasarojc (approved):

test/e2e/app/app.go
sergio-mena: The problem is that `num` scope is the if-else, so if we remove the else, then we have to define num outside the `if`. I think the way it was done was more idiomatic (leveraging go scopes)

jmalicevic (approved):
LGTM

state/indexer/sink/psql/psql.go
```diff
@@ -1,6 +1,8 @@
-// Package psql implements an event sink backed by a PostgreSQL database.
+//nolint:goconst
```
mzabaluev, line 1: I actually went and created the string consts in #1510, those concatenations do useless work repeatedly.
Even better for performance would be to create prepared SQL statements, because these queries are used in a loop.

libs/tempfile/tempfile_test.go
```diff
@@ -67,7 +69,7 @@ func TestWriteFileAtomicDuplicateFile(t *testing.T) {
 	atomicWriteFileRand = defaultSeed
 	firstFileRand := randWriteFileSuffix()
 	atomicWriteFileRand = defaultSeed
-	fname := "/tmp/" + atomicWriteFilePrefix + firstFileRand
+	fname := tmpStr + atomicWriteFilePrefix + firstFileRand
```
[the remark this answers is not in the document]
sergio-mena: I agree about "/tmp" being hardcoded... but that's out of the scope of this PR. Feel free to submit another one to fix that (there are well-known ways in go to generate a temporary file/dir)

> Could be `tmpDir` for clarity?

Following your suggestion [above](https://github.com/cometbft/cometbft/pull/1531#discussion_r1369902621), this change will be reverted

state/indexer/sink/psql/psql.go
sergio-mena: FYI, this code will likely be removed soon, as ADR101 makes it unnecessary to be part of CometBFT

test/e2e/app/app.go
lasarojc: I actually prefer the `if-else` but am trying to get more go idiomatic :)
But I am ok with either way here.

state/indexer/sink/psql/psql.go
sergio-mena: I cherry-picked 90fae42c0d into this PR

mzabaluev: Thanks; sorry the commit description was misleading, I edited it before making more changes.

mzabaluev (approved):

state/indexer/sink/psql/psql.go
sergio-mena: NP, will get squashed :-)

PR #1533, opened 2023-10-24
Rename proto messages and services to satisfy default buf lints

### LINKED ISSUE #1530: proto: Rename messages and services to satisfy buf lints
Change the names in the latest package versions of proto files to satisfy these [buf lint rules](https://buf.build/docs/lint/rules):

    - RPC_REQUEST_RESPONSE_UNIQUE
    - RPC_REQUEST_STANDARD_NAME
    - RPC_RESPONSE_STANDARD_NAME
    - SERVICE_SUFFIX

This will make the proto files satisfy the `DEFAULT` lint category.

mzabaluev:
Resolves #1530

---

#### PR checklist

- [ ] Tests written/updated


proto/cometbft/services/block_results/v1beta1/block_results_service.proto
```diff
@@ -1,17 +1,17 @@
...
- */
+   BlockResultService provides the block results of a given or latestheight.
+*/
 service BlockResultsService {
   // GetBlockResults returns the BlockResults of the requested height.
   rpc GetBlockResults(GetBlockResultsRequest) returns (GetBlockResultsResponse);
 
   // GetLatestBlockResults returns the BlockResults of the latest committed height.
-  rpc GetLatestBlockResults(GetLatestBlockResultsRequest) returns (GetBlockResultsResponse);
+  rpc GetLatestBlockResults(GetLatestBlockResultsRequest) returns (GetLatestBlockResultsResponse);
```
mzabaluev, line 16: Here, we have to get into redundancy because in principle, types of different RPC methods may evolve independently, even though we are sure that in this case the return messages will always be the same.

But I sense a deeper design issue: `GetLatestBlockResults` is non-idempotent, unlike `GetBlockResults`, so perhaps it was a good idea to define it separately and not as a special case behaviour for the other method.

A somewhat more elegant way to work around this is to define a shared message type and reuse it in both responses as the single field.

proto/cometbft/services/block_results/v1beta1/block_results.proto
```diff
@@ -12,14 +12,22 @@ message GetBlockResultsRequest {
...
-  bytes                                    app_hash                = 6;
+  int64 height = 1;
+  repeated cometbft.abci.v1beta3.ExecTxResult txs_results = 2;
+  repeated cometbft.abci.v1beta2.Event finalize_block_events = 3;
+  repeated cometbft.abci.v1beta1.ValidatorUpdate validator_updates = 4;
+  cometbft.types.v1beta3.ConsensusParams consensus_param_updates = 5;
+  bytes app_hash = 6;
+}
+
+message GetLatestBlockResultsResponse {
```
mzabaluev, line 26: So this is now an exact copy of `GetBlockResultsResponse` above.
Should we rather define a common `BlockResults` message and reuse it in both response types?

abci/client/grpc_client.go
```diff
@@ -23,7 +23,7 @@ type grpcClient struct {
 	service.BaseService
 	mustConnect bool
 
-	client   types.ABCIClient
+	client   types.ABCIServiceClient
```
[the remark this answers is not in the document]
sergio-mena: Yeah, since this will cause a lot renaming on dependent teams anyway, we'd better get it consistent at least

proto/cometbft/services/block_results/v1beta1/block_results.proto
sergio-mena: Why not coalescing both methods (`GetBlockResults` and `GetLatestBlockResults`) into one, `GetBlockResults`, and letting `height = 0` denote latest height? This is used in many of our JSON-RPC methods, so users are used to it.

proto/cometbft/state/v1beta4/types.proto
```diff
@@ -0,0 +1,14 @@
...
+import "cometbft/abci/v1beta4/types.proto";
+import "cometbft/state/v1beta3/types.proto";
+
+option go_package = "github.com/cometbft/cometbft/api/cometbft/state/v1beta4";
+
+message ABCIResponsesInfo {
+  // Retains the responses of the legacy ABCI calls during block processing.
+  v1beta3.LegacyABCIResponses legacy_abci_responses = 1;
+  int64 height = 2;
+  cometbft.abci.v1beta4.FinalizeBlockResponse response_finalize_block = 3;
```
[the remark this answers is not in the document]
sergio-mena: > it's been marshalled to JSON for storage this way

Do you mean for an external project? I've looked up all references in CometBFT and we only marhsal/unmarshall protobufs to save to/load from the state store.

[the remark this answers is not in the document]
sergio-mena: Check out 1844afc33. It's a small test to show (as expected) that the field name change does not affect protobuf marshalling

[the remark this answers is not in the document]
mzabaluev: Ah, pardon my confusion, it's protobuf marshal. I will rename it to `finalize_block`, which it should have been from the beginning.

proto/cometbft/services/block_results/v1beta1/block_results.proto
mzabaluev: This split also exists in the block service, between `GetByHeight` and `GetLatest`.

There are important semantic differences between the two as hinted above:  `GetBlockResults` always returns the same response (in success case) for the same input, while the response of `GetLatestBlockResults` varies with time.
Protobuf has a way to mark down these semantics for binding generators, which we might decide to use at some later point. I can also imagine a proxying service which would cache data for `GetBlockResults`, but query latest node state for `GetLatestBlockResults`.
It would be surprising to have a non-idempotent special case for the otherwise idempotent getter method, and the difference in behavior, without distinction in name, increases potential for inadvertent misuse. I'd argue that it's JSON-RPC that muddles these cases too much, but it's too late to fix.

mzabaluev: > It would be surprising to have a non-idempotent special case for the otherwise idempotent getter method

Except that's what ADR-101 actually specified and it's even documented in the protos, but then @thanethomson thought it necessary to introduce the "latest" methods in #1209 and as one of the review changes in #1168. I can't find the exact rationale, but I think it's somewhere close to the reasons I pointed out above.

sergio-mena: I think no one objected to your proposal in Slack, so I'd advice to do the refactoring needed in this PR to follow your proposal (including any change necessary to ADR101).
Your propoal, IIRC, is to remove the "special behaviour of height 0", and keep two methods: one idempotent (and so, cache-able) and the other getting the latest info.

mzabaluev: I have filed an RFC to discuss the matter  in detail: #1545

mzabaluev: While that issue waiting its due discussion, the simplest non-intrusive change here is to duplicate the response struct with the same list of fields, so that it is binary compatible with the `tendermint` definition in the main branch.

sergio-mena: I'd suggest we go ahead and merge this, and tackle the outcome of this discussion in a further PR

sergio-mena (approved):
This is ready for merging IMO, as the open discussion on the interface we want can be solved (if necessary, see #1545) in a further PR

thanethomson (approved):

PR #1534, opened 2023-10-24
fix spelling of bandwidth

kevinheavey:
It's currently written "bandwith" in 3 places

#### PR checklist

Guessing this doesn't need a changelog entry?


sergio-mena (approved):
Thanks for spotting this

PR #1535, opened 2023-10-24
Use semaphore to limit concurrency

lasarojc:
use semaphores to simplify the code
---

#### PR checklist

- [ ] Tests written/updated


hvanz:
Good idea, though:
- I would use `sync.WaitGroup` instead, which is more idiomatic,
- in `broadcastTxRoutine`, move `if memR.config.MaxPeers > 0 { ...` after `if memR.WaitSync() {`
- rename `sem` to something more meaningful.

PR #1538, opened 2023-10-25
Correct go_package paths in service proto files

mzabaluev:
Contributes to #1330

Fix up after #1443.
The new location of the generated files is in the /api subdirectory.
This did not affect generated code besides some metadata, but needs
to be fixed to avoid confusion.


mzabaluev:
The test flakiness should be fixed by #1539.

sergio-mena (approved):
👍

PR #1545, opened 2023-10-27
RFC 106: non-idempotent methods in data companion

mzabaluev:
On added drag of some "convenience" methods in the data companion API.

[Rendered](https://github.com/cometbft/cometbft/blob/mikhail/rfc-separate-stateful-methods/docs/rfc/rfc-106-separate-stateful-methods.md)

### DIFF
--- a/docs/rfc/rfc-106-separate-stateful-methods.md
+++ b/docs/rfc/rfc-106-separate-stateful-methods.md
@@ -55,3 +55,13 @@ non-idempotent "get latest block data" API for a data companion
 progress of the chain) has never been discussed or demonstrated.
 If this was only a "nice to throw in" idea, perhaps it's better to remove
 these methods until the need for them becomes apparent.
+
+## Proposed actions
+
+* Remove the following gRPC methods:
+  - `GetLatest` from `BlockService`;
+  - `GetLatestBlockResults` from `BlockResultsService`.
+* Change the documentation for the `GetByHeight` and `GetBlockResults` methods
+  to not treat the height parameter of 0 as a special case.
+* Revise the specifications in ADR-101 to remove the special treatment of
+  the height value 0.
--- a/docs/rfc/rfc-106-separate-stateful-methods.md
+++ b/docs/rfc/rfc-106-separate-stateful-methods.md
@@ -2,7 +2,7 @@
 
 ## Changelog
 
-- 2023-10-27: Initial revision
+- 2023-10-27: Initial revision (@mzabaluev)
 
 ## Abstract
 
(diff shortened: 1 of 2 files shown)


docs/rfc/rfc-106-separate-stateful-methods.md
```diff
@@ -0,0 +1,57 @@
+# RFC 106: Separation of non-idempotent methods in data companion API
+
+## Changelog
+
+- 2023-10-27: Initial revision
```
sergio-mena, line 5:
```suggestion
- 2023-10-27: Initial revision (@mzabaluev)
```

docs/rfc/rfc-106-separate-stateful-methods.md
```diff
@@ -0,0 +1,57 @@
...
+  proto cleanup work for [#1530](https://github.com/cometbft/cometbft/issues/1530).
+
+## Discussion
+
+In this case, the changes driven by practicalities of implementation actually
+highlight the different use cases and semantics for the respective methods.
+`GetByHeight` (with a valid height) returns the same data for the same input
+and so the responses can be cached at the client side, while the response
+of `GetLatest` varies with time. It's also easy to erroneously use 0 as the
+actual requested height in workflows using `GetHeight` and process the result
```
sergio-mena, line 47:
```suggestion
actual requested height in workflows using `GetByHeight` and process the result
```

docs/rfc/rfc-106-separate-stateful-methods.md
```diff
@@ -0,0 +1,57 @@
...
+to some ill effects down the road, rather than be stopped by a timely failure.
+
+For these reasons, it seems better to cleanly separate the two use cases and
+remove the documentation language about the height parameter of 0 as the special
+case to retrieve the latest value. Furthermore, the usefulness of the
+non-idempotent "get latest block data" API for a data companion
+(other than the `GetLatestHeight` stream subscription used to follow the
```
sergio-mena, line 54: If there's no clear use for `GetLatest` for this interface (IIUC because you can always use `GetLatestHeight` prior to getting the data by providing the height) I would support removing it to keep it as lean as possible.

sergio-mena (approved):
I agree with the changes proposed here.
I would suggest to structure the last paragraph as bullet points with the proposed changes to render it easier to come back to as a reference.

sergio-mena:
If we decide to go ahead with these changes, do we need to modify the interface specified in ADR101?

mzabaluev:
> If we decide to go ahead with these changes, do we need to modify the interface specified in ADR101?

Yes, the specification needs to be changed to remove special treatment of value 0 in `GetByHeight` and `GetBlockResults`.

sergio-mena (approved):
I'm OK with the latest changes

PR #1557, opened 2023-10-31
Make `LoadBlock` also return block metadata

### LINKED ISSUE #1556: `BlockStore` method `LoadBlock` could also return metadata
The DB-backed implementation of the `BlockStore` interface provided by `store` performs `LoadBlockMeta` as part of retrieving block data. The retrieved metadata structure is discarded after its data is used. At a number of call sites, the caller also needs to retrieve the block metadata, which is performed by calling `LoadBlockMeta` directly, resulting in a second lookup in the metadata "table".

To eliminate redundancy, `LoadBlock` could return the metadata in a tuple. It's reasonable to expect that any other storage implementations either have a similar data model (as they also need to support `LoadBlockPart`), or can create the metadata structure cheaply from block data.

mzabaluev:
Fixes #1556

---

#### PR checklist

- [x] Tests written/updated


mzabaluev:
I don't understand why the tests using mocks are failing in CI. I cannot reproduce the failures locally.

consensus/replay_test.go
```diff
@@ -1183,13 +1184,17 @@ func newMockBlockStore(t *testing.T, config *cfg.Config, params types.ConsensusP
...
+func (bs *mockBlockStore) Base() int64                    { return bs.base }
+func (bs *mockBlockStore) Size() int64                    { return bs.Height() - bs.Base() + 1 }
+func (bs *mockBlockStore) LoadBaseMeta() *types.BlockMeta { return bs.LoadBlockMeta(bs.base) }
+func (bs *mockBlockStore) LoadBlock(height int64) (*types.Block, *types.BlockMeta) {
+	return bs.chain[height-1], bs.LoadBlockMeta(height)
+}
+
+func (bs *mockBlockStore) LoadBlockByHash([]byte) (*types.Block, *types.BlockMeta) {
+	height := int64(len(bs.chain)) - 1
+	return bs.chain[height], bs.LoadBlockMeta(height)
```
melekes, line 1197:
```suggestion
	return bs.chain[height-1], bs.LoadBlockMeta(height)
```

Otherwise, there's a mismatch in logic between `LoadBlockByHash` and `LoadBlock`, where we load `block(height-1)` and `header(height)`.

consensus/replay_test.go
```diff
@@ -1183,13 +1184,17 @@ func newMockBlockStore(t *testing.T, config *cfg.Config, params types.ConsensusP
...
+func (bs *mockBlockStore) Height() int64                  { return int64(len(bs.chain)) }
+func (bs *mockBlockStore) Base() int64                    { return bs.base }
+func (bs *mockBlockStore) Size() int64                    { return bs.Height() - bs.Base() + 1 }
+func (bs *mockBlockStore) LoadBaseMeta() *types.BlockMeta { return bs.LoadBlockMeta(bs.base) }
+func (bs *mockBlockStore) LoadBlock(height int64) (*types.Block, *types.BlockMeta) {
+	return bs.chain[height-1], bs.LoadBlockMeta(height)
+}
+
+func (bs *mockBlockStore) LoadBlockByHash([]byte) (*types.Block, *types.BlockMeta) {
+	height := int64(len(bs.chain)) - 1
```
melekes, line 1196:
```suggestion
	height := int64(len(bs.chain))
```

melekes (changes requested):
nice 👍 needs a small fix though in the mock

melekes (approved):

Issue #30, opened 2022-12-23
Bandwidth optimization


lasarojc:
Inspired by tendermint/tendermint#9880  but expanded here.

High-level tracking issue for general bandwidth optimization efforts.

CometBFT-based networks consume large amounts of bandwidth. 
This results in high operational costs for the network operators (typically, the validators). 
We need to reduce the bandwidth consumption of the solution while maintaining its core properties, in terms of correctness, progress and fault-tolerance.

Several strategies to reduce the bandwidth usage in CometBFT have been proposed by CometBFT developers and community members. 
While some of these ideas are simple to implement, others require much larger efforts, implying impactful modifications on the protocol. 
This results in a dilemma when looking for a solution. We also note that there is currently no general strategy to validate a solution, whether it would be through local tests, or by collecting information from the operators. 
A third concern comes from the fact that while there are clear quick wins, it seems also necessary to prepare a long term effort. 
The subject of improving bandwidth usage in peer-to-peer systems is vast, with decades of efforts in research and practical solutions. 

This issue tracks our efforts in addressing a large spectrum of approaches to reduce bandwidth usage in CometBFT. 
We prioritize the work based on their impact on use cases, on the effort needed to implement it, and on how it helps pave the way for future improvements. 
The main problem is subdivided as listed next. Note that some of the problems/solutions might overlap.

```[tasklist]
### Preliminary work
- [ ] #1048
- [ ] #1076
- [ ] #1081
```

```[tasklist]
### Short term goals
- [ ] #1058
- [ ] #1060
- [ ] #1059
- [ ] #1061
```

```[tasklist]
### Long term goals
- [ ] #11
- [ ] #2027
- [x] Feasibility study to replace the current gossip protocol
- [ ] #3297
```

## Previous efforts

- Investigate what kind of data is consuming the most bandwidth
   - https://github.com/tendermint/tendermint/pull/9706 ([Rendered RFC](https://github.com/cometbft/cometbft/blob/main/docs/rfc/tendermint-core/rfc-027-p2p-message-bandwidth-report.md))
   - #26
   - #734 
   -Changes from tendermint/tendermint#9760, released in  v0.34.25, testing for backwards compatibility in tendermint/tendermint#9928

## Related, but not covered by this issue.

- [ ] Investigate protocol change options for more substantial bandwidth optimization
    - [ ] #13
    - [ ] #14 

# References
[...]

PR #1558, opened 2023-10-31
Experimental - Reduce # of connections effectively used to gossip transactions out

adizere:
Ref: #30 
Contributes to: #1058 

---

#### PR checklist

- [x] Tests written/updated


config/config.go
```diff
@@ -876,6 +876,10 @@ type MempoolConfig struct {
 	// Including space needed by encoding (one varint per transaction).
 	// XXX: Unused due to https://github.com/tendermint/tendermint/issues/5796
 	MaxBatchBytes int `mapstructure:"max_batch_bytes"`
+	// Only broadcast txs to up to this many peers.
+	// If we are connected to more than this number of peers, only send txs to
+	// the first MaxOutboundPeers of them. If one of those peers goes offline, activate another peer.
+	MaxOutboundPeers int `mapstructure:"max_outbound_peers"`
```
thanethomson, line 882:
```suggestion
	ExperimentalMaxOutboundPeers int `mapstructure:"experimental_max_outbound_peers"`
```

config/config.go
```diff
@@ -886,10 +890,11 @@ func DefaultMempoolConfig() *MempoolConfig {
...
 		// ABCI Recheck
-		Size:        5000,
-		MaxTxsBytes: 1024 * 1024 * 1024, // 1GB
-		CacheSize:   10000,
-		MaxTxBytes:  1024 * 1024, // 1MB
+		Size:             5000,
+		MaxTxsBytes:      1024 * 1024 * 1024, // 1GB
+		CacheSize:        10000,
+		MaxTxBytes:       1024 * 1024, // 1MB
+		MaxOutboundPeers: 10,
```
thanethomson, line 897:
```suggestion
		ExperimentalMaxOutboundPeers: 10,
```

config/config.go
```diff
@@ -925,6 +930,9 @@ func (cfg *MempoolConfig) ValidateBasic() error {
 	if cfg.MaxTxBytes < 0 {
 		return cmterrors.ErrNegativeField{Field: "max_tx_bytes"}
 	}
+	if cfg.MaxOutboundPeers < 0 {
+		return cmterrors.ErrNegativeField{Field: "max_outbound_peers"}
```
thanethomson, line 934:
```suggestion
	if cfg.ExperimentalMaxOutboundPeers < 0 {
		return cmterrors.ErrNegativeField{Field: "experimental_max_outbound_peers"}
```

config/toml.go
```diff
@@ -425,6 +425,11 @@ max_tx_bytes = {{ .Mempool.MaxTxBytes }}
 # XXX: Unused due to https://github.com/tendermint/tendermint/issues/5796
 max_batch_bytes = {{ .Mempool.MaxBatchBytes }}
 
+# Only broadcast txs to up to this many peers.
+# If we are connected to more than this number of peers, only send txs to
+# the first MaxOutboundPeers of them. If one of those peers goes offline, activate another peer.
+max_outbound_peers = {{ .Mempool.MaxOutboundPeers }}
```
thanethomson, line 431:
```suggestion
experimental_max_outbound_peers = {{ .Mempool.ExperimentalMaxOutboundPeers }}
```

mempool/reactor.go
```diff
@@ -49,6 +53,8 @@ func NewReactor(config *cfg.MempoolConfig, mempool *CListMempool, waitSync bool)
 		memR.waitSyncCh = make(chan struct{})
 	}
 	memR.mempool.SetTxRemovedCallback(func(txKey types.TxKey) { memR.removeSenders(txKey) })
+	memR.sem = semaphore.NewWeighted(int64(memR.config.MaxOutboundPeers))
```
thanethomson, line 56:
```suggestion
	memR.sem = semaphore.NewWeighted(int64(memR.config.ExperimentalMaxOutboundPeers))
```

mempool/reactor.go
```diff
@@ -179,11 +185,20 @@ func (memR *Reactor) broadcastTxRoutine(peer p2p.Peer) {
 		}
 	}
 
+	if memR.config.MaxOutboundPeers > 0 {
```
thanethomson, line 188:
```suggestion
	if memR.config.ExperimentalMaxOutboundPeers > 0 {
```

thanethomson (commented):
This seems good so far. Let's get this merged ASAP and look into backporting to v0.38 and v0.37.

@hvanz please include your large-scale testnet results here or on the related issue(s).

And please ensure that the config parameter's renamed to `experimental_max_outbound_peers` before merging. This is especially critical for backports. But let's not remove the `experimental_` prefix until we've validated this change in production networks.

config/config.go
lasarojc: Changed the name. Does it sound good?
Also changed the default value to 0.

lasarojc:
Experimental results.
Tests were run with a load of 400tps, injected on a single node, on a network with 175 validators.
The number of peers effectively used for gossiping transactions outbound (E) were 20, 15, 10, 5 and 0 (0 equals no limitation).

In the graphs in the P2P section, yellow shows data related to mempool messages and green to block related messages.
Reducing the E consistently reduces inbound and output bandwidth of mempool gossiping.
Reducing E to 5 had an adverse impact on the block sizes, as shown in the last graph of the Consensus section. Green in this graph is the average block size of all nodes at a given instant and yellow is the 1m window average. 

image

The second set of graphs also show improvements in metrics when E is decreased, except for the recheck rate, which increases with E = 5.

image

Based on these results and expectation of size of networks, we recommend using E = 10. With this setup, for example, a transaction is spread to 10k nodes in just 4 hops, while considerably saving bandwidth as shown by the experimental results.

.changelog/unreleased/improvements/1558-experimental-gossip-limiting.md
```diff
@@ -0,0 +1,3 @@
+- `[mempool]` Only gossip transactions to a subset of the connections,
```
sergio-mena, line 1:
```suggestion
- `[mempool]` Only gossip transactions to a subset of the connected peers,
```

config/config.go
```diff
@@ -876,6 +876,10 @@ type MempoolConfig struct {
 	// Including space needed by encoding (one varint per transaction).
 	// XXX: Unused due to https://github.com/tendermint/tendermint/issues/5796
 	MaxBatchBytes int `mapstructure:"max_batch_bytes"`
+	// Only broadcast txs to up to this many peers.
+	// If we are connected to more than this number of peers, only send txs to
+	// the first ExperimentalMaxUsedOutboundPeers of them. If one of those peers goes offline, activate another peer.
+	ExperimentalMaxUsedOutboundPeers int `mapstructure:"experimental_max_used_outbound_peers"`
```
sergio-mena, line 882: Can you also explain here what setting it to 0 means?

config/toml.go
```diff
@@ -425,6 +425,14 @@ max_tx_bytes = {{ .Mempool.MaxTxBytes }}
 # XXX: Unused due to https://github.com/tendermint/tendermint/issues/5796
 max_batch_bytes = {{ .Mempool.MaxBatchBytes }}
 
+# Experimental parameter to limit broadcast of txs to up to this many peers
+# If we are connected to more than this number of peers, only send txs to
+# the first ExperimentalMaxOutboundPeers of them.; if one of those peers goes
+# offline, activate another peer.
+# Value disables the feature by not limiting the number of active connections.
```
sergio-mena, line 432:
```suggestion
# Value 0 disables the feature by not limiting the number of active connections.
```

config/toml.go
```diff
@@ -425,6 +425,14 @@ max_tx_bytes = {{ .Mempool.MaxTxBytes }}
...
+# If you enable this feature, a value of 10 is recommended.
```
sergio-mena, line 433:
```suggestion
# If you enable this feature, a value of 10 is recommended based on experimental performance results.
```
Not sure that's the reason tho

mempool/metrics.go
```diff
@@ -40,4 +40,8 @@ type Metrics struct {
 	// Number of times transactions were received more than once.
 	//metrics:Number of duplicate transaction reception.
 	AlreadyReceivedTxs metrics.Counter
+
+	// Experimental Number of connections being actively used for gossiping
+	// transactions outbound.
+	ExperimentalOutboundPeers metrics.Counter
```
sergio-mena, line 46: Two questions here:
* Does a metric really need to be experimental? (config knobs are a different story)
* Checking the [metric types](https://prometheus.io/docs/concepts/metric_types/), I'm not sure you want this to be a _counter_ (check the sentence saying "Do not use a counter to expose a value that can decrease"). Shouldn't it rather be a Gauge?

mempool/reactor.go
```diff
@@ -49,6 +53,8 @@ func NewReactor(config *cfg.MempoolConfig, mempool *CListMempool, waitSync bool)
 		memR.waitSyncCh = make(chan struct{})
 	}
 	memR.mempool.SetTxRemovedCallback(func(txKey types.TxKey) { memR.removeSenders(txKey) })
+	memR.activeConnectionsSemaphore = semaphore.NewWeighted(int64(memR.config.ExperimentalMaxUsedOutboundPeers))
```
sergio-mena, line 56: Should we gate this with "if memR.config.ExperimentalMaxUsedOutboundPeers > 0", similarly to what you're doing below?

mempool/reactor.go
```diff
@@ -93,7 +99,18 @@ func (memR *Reactor) GetChannels() []*p2p.ChannelDescriptor {
 // It starts a broadcast routine ensuring all txs are forwarded to the given peer.
 func (memR *Reactor) AddPeer(peer p2p.Peer) {
 	if memR.config.Broadcast {
-		go memR.broadcastTxRoutine(peer)
+		go func() {
+			if memR.config.ExperimentalMaxUsedOutboundPeers > 0 {
+				if err := memR.activeConnectionsSemaphore.Acquire(context.TODO(), 1); err != nil {
+					memR.Logger.Error("Failed to acquire semaphore: %v", err)
```
sergio-mena, line 105: Is this expected to happen in normal conditions? If not, consider a panic

sergio-mena (approved):
Some comments, otherwise LGTM

mempool/reactor.go
lasarojc: Should only happen if the context expires, which should not happen here.

mempool/reactor.go
lasarojc: We could do that, but the impact would be negligible here. This way the code seems cleaner.

config/config.go
```diff
@@ -876,6 +876,10 @@ type MempoolConfig struct {
 	// Including space needed by encoding (one varint per transaction).
 	// XXX: Unused due to https://github.com/tendermint/tendermint/issues/5796
 	MaxBatchBytes int `mapstructure:"max_batch_bytes"`
+	// Only broadcast txs to up to this many peers.
+	// If we are connected to more than this number of peers, only send txs to
+	// the first ExperimentalMaxUsedOutboundPeers of them. If one of those peers goes offline, activate another peer.
```
lasarojc, line 881:
```suggestion
 // Experimental parameter to limit broadcast of txs to up to this many peers
 // If we are connected to more than this number of peers, only send txs to
 // the first ExperimentalMaxOutboundPeers of them. If one of those peers goes
 // offline, activate another peer.
 // Value 0 disables the feature by not limiting the number of active connections.
 // If you enable this feature, a value of 10 is recommended based on experimental performance results.	
 ```

File: mempool/metrics.go (hunk 11 above)
Comment (hvanz) [reply mempool/metrics.go]: I agree. I renamed it and changed the type to gauge.

File: config/config.go (hunk 8 above)
Comment (hvanz) [reply config/config.go:887]: Done.

File: mempool/reactor.go [hunk 15]
```diff
@@ -93,7 +99,20 @@ func (memR *Reactor) GetChannels() []*p2p.ChannelDescriptor {
 // It starts a broadcast routine ensuring all txs are forwarded to the given peer.
 func (memR *Reactor) AddPeer(peer p2p.Peer) {
 	if memR.config.Broadcast {
-		go memR.broadcastTxRoutine(peer)
+		go func() {
+			if memR.config.ExperimentalMaxUsedOutboundPeers > 0 {
+				if err := memR.activeConnectionsSemaphore.Acquire(context.TODO(), 1); err != nil {
```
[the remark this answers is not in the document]
lasarojc:
```suggestion
				// Around (MaxOutboundPeers-ExperimentalMaxUsedOutboundPeers) goroutines will be
				// blocked here waiting for more peers disconnect and free some slots for running.
				if err := memR.activeConnectionsSemaphore.Acquire(context.TODO(), 1); err != nil {
```

lasarojc:
Merging this PR and opening another one to exclude non-persistent peers from the limit.

mempool/reactor_test.go
```diff
@@ -333,6 +333,45 @@ func TestReactorTxSendersMultiNode(t *testing.T) {
...
+	require.Zero(t, reactors[2].mempool.Size())
+	require.Zero(t, reactors[3].mempool.Size())
+
+	// In the first reactor, disconnect the second reactor; the third reactor should become active.
+	firstPeer := reactors[0].Switch.Peers().List()[0]
+	reactors[0].Switch.StopPeerGracefully(firstPeer)
+
+	// Now the third reactor should receive the transactions; the fourth reactor's mempool should
+	// still be empty.
+	checkTxsInMempool(t, txs, reactors[2], 0)
```
cason, line 371: How are you sure that is the second reactor (and not the third) that will get the semaphore's permit?

mempool/reactor.go
cason: Should we really consider this behaviour that shouldn't happen? I think it just causes confusion on readers, as the error might be interpreted as a failure to acquire (immediately). Moreover, context.TODO() in production code is not great.

mempool/reactor_test.go
hvanz: Yes, this test is quite brittle, it depends on the current implementation of the p2p layer. We know it works like that because of the order in which peers are created and connected, but this can change and brake the test.

Issue #1559, opened 2023-10-31
e2e: Emulate WAN latencies between nodes in testnets


cason:
Latencies between nodes in testnets used for end-to-end (e2e) tests and for quality assurance (QA) experiments are not realistic. In the case of e2e tests, nodes are running in docker images and the network latency between nodes are negligible (localhost). In the case of QA experiments, they are currently performed in Digital Ocean (DO) with all nodes located in the same availability zone/region; latency between nodes are the ones expected for a local-area (LAN) setup. In production networks, however, run in geographically distributed environments, with substantial (WAN) latencies between participant nodes. The infrastructure used for e2e testing and for QA experiments thus fail to reproduce the latencies observed in production networks.

The Linux kernel includes Network Emulation features that can be controlled using the [`tc` command](https://man7.org/linux/man-pages/man8/tc-netem.8.html). The goal of this issue is then to produce infrastructure tools to emulate realistic latencies between nodes running in a testnet, targeting both the Docker provider used for e2e tests and the Digital Ocean provider used in QA experiments.

## Definition of Done

- [ ] Document the methodology to emulate latencies between testnet nodes
- [ ] #1560 
- [ ] #1587

cason:
The tool available on https://github.com/paulo-coelho/latency-setter allows emulating WAN latencies between Linux machines. The tool assumes the definition of an arbitrary number `N` of zones, mimicking the availability regions/zones offered by cloud providers. The input of the tools is:

1. A `NxN` matrix representing the emulated (one-way) latency in ms between each pair of zones
2. A file mapping the IP of each node to one of the defined zones (multiple nodes can be hosted in the same zone)

The main component of the tool is a Python script that receives the above two files and used `tc` to configure the kernel to emulate the configured latencies between the node in which it runs and every other node mapped to a different zone. Nodes mapped to the same zone have no artificial latency emulated between them.

We opted for this tool because it was used in academical environments to emulate, in a LAN cluster, latencies between nodes observed when running experiments in AWS.

cason:
Limitations when adopting an adapted version of the latency setter script from https://github.com/paulo-coelho/latency-setter:

- No support for IPv6 addresses
- We are not sure if all the `tc` commands adopted are really needed for our use case. Understanding and documenting the role of each of them is a ToDo
- Some warnings are generated by the comment, which are muted by appending `2> /dev/null` to some commands
- We might consider re implementing it, for instance, in Go

PR #1560, opened 2023-10-31
e2e: Support emulated latencies between docker images

cason:
Part of #1559.

---

#### PR checklist

- [ ] Tests written/updated


test/e2e/pkg/infra/digitalocean/digitalocean.go
```diff
@@ -46,6 +46,12 @@ func (p Provider) StartNodes(ctx context.Context, nodes ...*e2e.Node) error {
 
 	return execAnsible(ctx, p.Testnet.Dir, playbookFile, nodeIPs)
 }
+
+// Currently unsuported.
+func (p Provider) SetLatency(_ context.Context, _ *e2e.Node) error {
+	return fmt.Errorf("SetLatency() currently unsuported for Digital Ocean")
```
sergio-mena, line 52:
```suggestion
// Currently unsupported.
func (p Provider) SetLatency(_ context.Context, _ *e2e.Node) error {
	return fmt.Errorf("SetLatency() currently unsupported for Digital Ocean")
```

test/e2e/latency/latency-setter.py
```diff
@@ -0,0 +1,98 @@
...
+with open(sys.argv[3], 'r') as f:
+    reader = csv.reader(f)
+    lats = list(list(rec) for rec in csv.reader(f, delimiter=',')) #reads csv into a list of lists
+
+myzone = ''
+mynode = ''
+for item in ipData:
+    if item['IP'] == myip:
+        myzone = item['Zone']
+        mynode = item['Node']
```
sergio-mena, line 49:
```suggestion
        mynode = item['Node']
        break
```

test/e2e/latency/latency-setter.py
```diff
@@ -0,0 +1,98 @@
...
+
+if myzone == '':
+    print(f'# No zone configured for node {myip}')
+    exit()
+
+azs = lats[0]
+tlats = {}
+for l in lats[1:]:
```
sergio-mena, line 57: Please add a comment why we're not using the first element of `lats`

test/e2e/pkg/infra/docker/docker.go
```diff
@@ -158,6 +188,24 @@ services:
 	return buf.Bytes(), nil
 }
 
+func zonesTableBytes(nodes []*e2e.Node) ([]byte, error) {
+	tmpl, err := template.New("zones").Parse(`Node,IP,Zone
+{{- range . }}
+{{- if .Zone }}
+{{ .Name }},{{ .InternalIP }},{{ .Zone }}
+{{- end }}
+{{- end }}`)
```
sergio-mena, line 197: Excellent!

test/e2e/pkg/testnet.go
```diff
@@ -663,3 +709,50 @@ func (g *ipGenerator) Next() net.IP {
 	}
 	return ip
 }
+
+func loadZoneLatenciesMatrix(filePath string) (map[ZoneID][]uint32, error) {
+	records, err := readCsvFile(filePath)
```
sergio-mena, line 714: Since we need to parse the csv file here anyway? Wouldn't it be easier to set the latencies from here? (rather than from a script inside the container) It would be easier to debug.
This is just a suggestion, not sure it's worth porting the script from python to go.

test/e2e/pkg/manifest.go
```diff
@@ -178,6 +178,9 @@ type ManifestNode struct {
 	// It defaults to false so unless the configured, the node will
 	// receive load.
 	SendNoLoad bool `toml:"send_no_load"`
+
+	// Geographical zone ID for simulating latencies.
+	Zone string `toml:"zone"`
```
sergio-mena, line 183: What happens with nodes that don't have a zone set? (my understanding is that the node doesn't get any latency set)
Does is have a valid physical interpretation?
Should we just set the no zone nodes to a default zone? (so latencies are always on, whether the user wants them or not)
Or should there be a general boolean to enable/disable latencies, and if enabled, nodes with no zone would go to a default zone?

test/e2e/pkg/testnet.go
```diff
@@ -359,6 +368,38 @@ func (t Testnet) Validate() error {
...
+	for zone := range zoneMatrix {
+		zones = append(zones, zone)
+	}
+
+	// Check that the zone ids of all nodes are valid when matrix file exists.
+	for _, node := range nodes {
+		if !node.ZoneIsSet() {
+			continue
+		}
+		if fileNotFoundErr {
```
melekes, line 390: Shall we move this condition out of the `for` loop?  https://github.com/cometbft/cometbft/pull/1560/files#diff-3a52d9f2dd1e93d9770b6c41bee2faae337bbdb1b73a41db4029b5f17e8fc91aR378 looks like the right place to me. wdyt?

test/e2e/pkg/testnet.go
```diff
@@ -663,3 +709,50 @@ func (g *ipGenerator) Next() net.IP {
...
+			if err != nil {
+				return nil, ErrParsingStringToInt{l, err}
+			}
+			matrix[zoneID][i] = uint32(lat)
+		}
+	}
+	return matrix, nil
+}
+
+type ErrParsingStringToInt struct {
```
melekes, line 734:
```suggestion
type ErrInvalidZoneId struct {
```

test/e2e/pkg/testnet.go
```diff
@@ -663,3 +709,50 @@ func (g *ipGenerator) Next() net.IP {
...
+	ZoneID string
+	Err    error
+}
+
+func (e ErrParsingStringToInt) Error() string {
```
melekes, line 739:
```suggestion
func (e ErrInvalidZoneId) Error() string {
```

test/e2e/pkg/testnet.go
hvanz: It's allowed to have no latencies file if all the nodes have no zone. But to check this condition, we need to inspect every node, and then return error if a node has a zone and there is no file.

test/e2e/pkg/manifest.go
hvanz: A node with no zone will have 0 added latency when sending or receiving messages to other nodes. It's like it belongs to a zone that is close to all zones, if that makes any sense. I don't know if this should be allowed or not, but it doesn't do any harm, except maybe when you forget to assign a zone to a node.

melekes (approved):
👍

lasarojc (approved):

test/e2e/pkg/manifest.go
sergio-mena: (sorry for delay getting back to this) I think it makes sense to define a default behavior that does make sense (a zone close to all others is a bit of a stretch IMO)

thanethomson (approved):
LGTM :+1: Just some minor spelling-related nits.

test/e2e/pkg/manifest.go
hvanz: I added a `default_zone` parameter to the testnet. Now either all nodes have a zone or none have.

PR #1584, opened 2023-11-08
mempool: Limit gossip connections to persistent and non-persistent peers (experimental)

### LINKED ISSUE #1058: Analyze, prototype and experiment easily applicable solutions to reduce bandwidth usage

lasarojc:
Closes #1058

Complements the work done in #1558: adds two independent configuration values for limiting outbound gossip connections, for persistent peers and non-persistent peers. Unconditional peers are not affected by this feature.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -876,16 +876,19 @@ type MempoolConfig struct {
 	// Including space needed by encoding (one varint per transaction).
 	// XXX: Unused due to https://github.com/tendermint/tendermint/issues/5796
 	MaxBatchBytes int `mapstructure:"max_batch_bytes"`
-	// Experimental parameter to limit broadcast of txs to up to this many peers.
-	// If we are connected to more than this number of peers, only send txs to the first
-	// ExperimentalMaxUsedOutboundPeers of them. If one of those peers disconnects, activate another
-	// peer.
-	// Persistent peers do not count towards this limit.
-	// If set to 0, this feature is disabled, that is, the number of active connections is not
-	// bounded.
-	// If enabled, a value of 10 is recommended based on experimental performance results using the
-	// default P2P configuration.
-	ExperimentalMaxUsedOutboundPeers int `mapstructure:"experimental_max_used_outbound_peers"`
+	// Experimental parameters to limit gossiping txs to up to the specified number of peers.
+	// We use two independent upper values for persistent peers and for non-persistent peers.
+	// Unconditional peers are not affected by this feature.
+	// If we are connected to more than the specified number of persistent peers, only send txs to
+	// the first ExperimentalMaxGossipConnectionsToPersistentPeers of them. If one of those
+	// persistent peers disconnects, activate another persistent peer. Similarly for non-persistent
+	// peers, with an upper limit of ExperimentalMaxGossipConnectionsToNonPersistentPeers.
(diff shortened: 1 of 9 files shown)


mempool/metrics.go
```diff
@@ -42,6 +42,7 @@ type Metrics struct {
 	AlreadyReceivedTxs metrics.Counter
 
 	// Number of connections being actively used for gossiping transactions
+	// excluding peristent peers.
```
lasarojc, line 45:
```suggestion
	// excluding persistent peers.
```

melekes (approved):
👍

andynog:
Moved to `In Progress` again as per conversation with @hvanz, still need some work

lasarojc (commented):
I cannot approve, since I opened the PR, but LGTM

config/config.go
```diff
@@ -876,15 +876,19 @@ type MempoolConfig struct {
...
-	// If set to 0, this feature is disabled, that is, the number of active connections is not
-	// bounded.
-	// If enabled, a value of 10 is recommended based on experimental performance results using the
-	// default P2P configuration.
-	ExperimentalMaxUsedOutboundPeers int `mapstructure:"experimental_max_used_outbound_peers"`
+	// Experimental parameters to limit gossiping txs to up to the specified number of peers.
+	// We use two independent upper values for persistent peers and for non-persistent peers.
+	// Unconditional peers are not affected by this feature.
+	// If we are connected to more than the specified number of persistent peers, only send txs to
+	// the first ExperimentalMaxGossipConnectionsToPersistentPeers of them. If one of those
```
ValarDragon, line 883: This seems like it should not be the first 10, but instead a random 10 peers, no?

PR #1565, opened 2023-11-03
ADR 110: Remote mempool

thanethomson:
Per recent discussions surrounding the default mempool in Comet and its flooding gossip mechanism, this ADR is proposed to allow application developers to build their own mempools (or mempool clusters) that reside in entirely separate processes with a simple, standardized interface.

:book: [Rendered](https://github.com/cometbft/cometbft/blob/thane/adr/mempool-client/docs/architecture/adr-110-remote-mempool.md)

Currently in early draft for review. TODO: Create a tracking issue for the ADR if the team decides it's a good idea to go ahead with this.

cc @zmanian @ValarDragon @bpiv400 

---

#### PR checklist

- [ ] Tests written/updated
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments

PR #1585, opened 2023-11-08
ADR 111: `nop` Mempool

sergio-mena:
With the same reasons and discussions that motivated #1565, this ADR presents a similar approach, with the same goal: pushing the mempool functionality out of CometBFT (optionally, for the time being).

The advantage of this approach is that, AFAICT, it is light weight (less code changes, no newly introduced interface) while also pushing the mempool functionality out of CometBFT (specifically, to the application). Just as in #1565, the application designers are free to implement whatever ad-hoc mechanism for transaction dissemination they deem appropriate for their appchain.

[Rendered](https://github.com/cometbft/cometbft/blob/sergio/nil-mempool/docs/architecture/adr-111-nil-mempool.md)

---

#### PR checklist

- [ ] Tests written/updated


docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
+# ADR 111: `nil` Mempool
```
thanethomson, line 1: Would it not be better to call this "Application-side mempool", given that that's the ultimate implication of this change?

Or is there some other way of emphasizing that that's the ultimate impact of this ADR?

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+
+- 2023-11-07: First version (@sergio-mena)
+
+## Status
+
+Accepted | Rejected | Deprecated | Superseded by
+
+## Context
+
+### In the Beginning...
```
thanethomson, line 13: I'd recommend putting all this content under the Detailed Design section, and have a very brief overview/summary here under Context. That could be another good opportunity to emphasize how the impact here is to fully enable app-side mempools.

thanethomson (commented):
Looking great so far! Only had capacity to give it a brief review, will review more thoroughly once I'm back up and running.

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+This has greatly improved the ability for appchains to influence the contents of proposed block.
+Further, it has enabled many new use cases for appchains. However some issues remain with
+the current model:
+
+* We are using the same p2p network for disseminating transactions and consensus-related messages.
+* Many mempool parameters are configured on a per-node basis by node operators,
+  allowing the possibility of inconsistent mempool configuration across the network
+  with potentially serious scalability effects
+  (even causing unacceptable performance degradation in some extreme cases).
+* The mempool is using a basic (robust but suboptimal) flood algorithm
```
andynog, line 52:
```suggestion
* The current mempool implementation uses a basic (robust but sub-optimal) flood algorithm
```

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+  * the CometBFT team are working on improving it as one of our current priorities,
```
andynog, line 53:
```suggestion
  * the CometBFT team is working on improving it as one of our current priorities,
```

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+    but any improvement we come up with must address the needs of a vast spectrum of applications,
+    as well as be heavily scaled-tested in various scenarios
+    (in an attempt to cover the applications' wide spectrum)
+  * a mempool designed specifically for one particular application
+    would reduce the search space as its designers can devise it with just their application's
+    needs in mind.
+* The interaction with the application is still somewhat convoluted:
+  * the app has to decide what logic to implement in `CheckTx`,
```
andynog, line 61:
```suggestion
  * the application has to decide what logic to implement in `CheckTx`,
```

just making it consistent to ensure we are talking about the same thing :wink:

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+    what to do with the transaction list coming in `RequestPrepareProposal`,
+    whether it wants to maintain an app-side mempool (more on this below), and whether or not
+    to combine the transactions in the app-side mempool with those coming in `RequestPrepareProposal`
+  * all those combinations are hard to fully understand, as the semantics and guarantees are
+    often not clear
```
andynog, line 66:
```suggestion
  * All those combinations are difficult to fully comprehend due to unclear semantics and guarantees.
```

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+    as in-transit transactions need to be kept in both mempools.
+
+The approach presented in this ADR builds on the app-mempool design released in `v0.47.x`
+of the Cosmos SDK,
+and briefly discussed in the last bullet point above (see [SDK app-mempool][sdk-app-mempool] for further details of this model).
+
+In the app-mempool design in Cosmos SDK `v0.47.x`
+an unconfirmed transaction must be both in CometBFT's mempool for dissemination and
+in the app's mempool so the application can decide how to manage the mempool.
+The many advantages of this approach are beyond question. However it has some implications:
```
andynog, line 80:
```suggestion
There is no doubt that this approach has numerous advantages. However, it also has some implications that need to be considered:
```

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+* Having every transaction both in CometBFT and in the application is suboptimal in terms of memory.
+  Additonally, the app developer has to be careful
+  that the contents of both mempools does not diverge over time
+  (hence the crucial role `re-CheckTx` plays post-ABCI++).
+* The main reason for a transaction needing to be in CometBFT's mempool is
+  because the design in Cosmos SDK `v0.47.x` does not consider an application
+  that has its own means of disseminating transactions.
+  It reuses the peer to peer network underneath CometBFT reactors.
+* So, if an app has an ad-hoc design of how to disseminate transactions,
+  there is no point in having transactions in CometBFT's mempool.
```
andynog, line 91:
```suggestion
* There is no point in having transactions in CometBFT's mempool if an application implements an ad-hoc design of disseminating transactions.
```

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+
+This proposal targets this kind of applications:
+those that have an ad-hoc (and likely more efficient) mechanism for transaction dissemination.
```
andynog, line 94:
```suggestion
This proposal targets this kind of applications that have an ad-hoc (and likely more efficient) mechanism for transaction dissemination.
```

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+We propose to introduce a `nil` mempool which will effectively act as a stubbed object
+internally:
+
+* it will reject any transaction being locally submitted or gossipped by a peer
+* when it is time to _reap_ (as it is currently called) the mempool the answer will always be empty
```
andynog, line 99:
```suggestion
* when a _reap_ (as it is currently called) is executed in the mempool, an empty answer will always be returned
```

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+* the application running on the proposer validator will add transactions it received
+  using the appchains's own mechanism via `PrepareProposal`.
```
andynog, line 101: should we re-phrase `appchains's own mechanism` adding `add-mempool` since that's the term it was described before ?

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+
+## Alternative Approaches
+
+These are the alternatives known to date:
+
+1. Keep the current model. Useful for basic apps, but clearly suboptimal for applications
+   with their own mechanism to disseminate transactions and particular performance requirements.
+2. Provide more efficient general-purpose mempool implementations.
```
andynog, line 109: Should we mention the CAT mempool here ? It's a good example of a new design that has been some R&D, testing, validation to ensure a different approach is a better alternative

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+albeit with a complex, poor UX.
+This core of this proposal is to make some internal changes so it is clear an simple for app developers,
+thus improving the UX.
+
+#### `nil` Mempool
+
+We propose a new mempool implementation, called `nil` Mempool, that effectively disables all mempool functionality
+within CometBFT.
+The `nil` Mempool implements the `Mempool` interface in a very simple manner:
+
```
andynog, line 131: Did we confirm that all these returns on the `nil` mempool will not cause any side effects or undesirable behaviours ? From what I read the idea of going with this model is that would not required any extra changes except on the interface implementation level.

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+We propose the following changes to the `config.toml` file:
+
+```toml
+[mempool]
+# The type of mempool for this CometBFT node to use.
+#
+# Valid types of mempools supported by CometBFT:
+# - "flood" : clist mempool with flooding gossip protocol (default)
+# - "nil"   : nil-mempool (app needs an alternative tx dissemination mechanism)
```
andynog, line 159:
```suggestion
# - "nil"   : nil-mempool (app has implemented an alternative tx dissemination mechanism)
```

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+
+The config validation logic will be modified to add a new rule that rejects a configuration file if:
+
+* the mempool is set to `nil`
+* `create_empty_blocks`, in `consensus` section, is set to `false`.
+
+### Impacted Workflows
+
+* *Submitting a transaction*. Users are not to submit transactions via CometBFT's RPC.
+  `BroadcastTx*` RPC methods will fail.
```
andynog, line 171: How is the impact of this ? From a client's perspective (eg. wallets) how do they determine if they can submit a tx via `BroadcastTx` or not, is that based on the mempool configuration ?

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+  - Ad-hoc ways to add, remove, merge, reorder, modify, prioritize transactions according
+    to application needs.
+  - A way to disseminate transactions (gossip-based or other) to get the submitted transactions
+    to proposers. The application developers can devise simpler, efficient mechanisms tailored
+    to their application.
+  - Back-pressure mechanisms to prevent malicious users from abusing the transaction
+    dissemination mechanism.
+- In this approach, CometBFT's peer-to-peer layer does not need to deal with transaction gossipping,
+  and its resources can be used by other reactors such as consensus, evidence,
+  block-sync, or state-sync.
```
andynog, line 218:
```suggestion
- In this approach, CometBFT's peer-to-peer layer is relieved from managing transaction gossip, freeing up its resources for other reactors such as consensus, evidence, block-sync, or state-sync.
```

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+- There is no risk for the operators of a network to provide inconsistent configurations
```
andynog, line 219: One scenario not clear to me is that if the application implements its ad-hoc mechanism, and some operators configured the `nil` mempool but others might not configure it (e.g. configure the CometBFT regular one). Will that cause any issues ? I think it should be OK because the app should handle the transaction dissemination, but just wondering if there are possible scenarios where a mis-configuration would cause side-effects or issues.

andynog (commented):
Looks good to me. Provided some suggestions, mostly typos or re-writing in a simpler/clear language, some questions too. Feel free to ignore if doesn't make sense. Good job!

docs/architecture/adr-111-nil-mempool.md
melekes: Good point 👍 We must ensure the node doesn't panic when receiving a slice of transactions via the p2p network.

`RequestPrepareProposal` will contain a list of transactions in such case, but as long as the app ignores it and uses a custom mempool, it should be okay.

melekes (approved):

docs/architecture/adr-111-nil-mempool.md
melekes: `abci` mempool ?

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+### Configuration
+
+We propose the following changes to the `config.toml` file:
+
+```toml
+[mempool]
+# The type of mempool for this CometBFT node to use.
+#
+# Valid types of mempools supported by CometBFT:
+# - "flood" : clist mempool with flooding gossip protocol (default)
```
melekes, line 158: local (or a similar word) would be my preference because

a) I doubt that CometBFT will have more than one mempool implementation in the future,
b) devs/operators usually don't know/care about the concrete mechanics (either it's point-to-point, gossip, or something else). I.e., I see some operators confused about flood option if they didn't read the docs

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+# - "nil"   : nil-mempool (app needs an alternative tx dissemination mechanism)
+type = "nil"
+```
+
+The config validation logic will be modified to add a new rule that rejects a configuration file if:
```
romac, line 163: It's quite obvious in the context of the ADR, but perhaps worth clarifying a little bit that both conditions need to be met for the configuration to be rejected:

```suggestion
The config validation logic will be modified to add a new rule that rejects a configuration file if all of these conditions are met:
```

romac: Can we maybe also add a one-line explanation for why `create_empty_block` must be set to `true` with the nil mempool?

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+
+* the mempool is set to `nil`
+* `create_empty_blocks`, in `consensus` section, is set to `false`.
```
jmalicevic, line 166: Are we sure that there are not applications who will not require empty blocks to be enabled? If not, do we have a way of providing support for this behaviour without erroring? How would such a scenario look?

docs/architecture/adr-111-nil-mempool.md
jmalicevic: We should offer a meaningful return method that this does not work. However, this endpoint is still open to attacks , needlessly perhaps so it can be disabled?

docs/architecture/adr-111-nil-mempool.md
jmalicevic: Ah I just realized from context that this should in fact be `true` ?

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+
+The changes needed for this approach, are fairly simple, and the logic is clear.
+This might allow us to even deliver it as part of CometBFT `v1` (our next release)
+even without a noticeable impact on `v1`'s delivery schedule.
+
+The CometBFT team (learning from past dramatic events) usually takes a conservative approach
+for backporting changes to release branches that have already undergone a full QA cycle
+(and thus are in code-freeze mode).
+For this reason, although the limited impact of these changes would limit the risks
+of backporting to `v0.38.x` and `v0.37.x`, a careful risk/benefit evaluation will
```
jmalicevic, line 195: backport to `v0.38.x-experimental` ? Or as an opt-in?

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+*	`Lock()`: does nothing
+* `Unlock()`: does nothing
+*	`Update(...) error`: returns `nil`
+* `FlushAppConn() error`: returns `nil`
+*	`Flush()`: does nothing
+* `TxsAvailable() <-chan struct{}`: returns a closed channel
+* `EnableTxsAvailable()`: does nothing
+* `SetTxRemovedCallback(cb func(types.TxKey))`: does nothing
+* `Size() int` returns 0
+* `SizeBytes() int64` returns 0
```
jmalicevic, line 145: I have a similar comment on the other ADR regarding the mempool, should we still allow  for some of these endpoints to return info to a user by checking whether the application provides implementation of those?

Also these functions are also accessed via RPC and we are not discussing in the ADR what happens to those RPC endpoints (if anything).

docs/architecture/adr-111-nil-mempool.md
andynog: I believe It might be possible to add a condition that remove this route from the RPC, not as easy as what we did with the new gRPC but I think it is doable and I would prefer that we enforce things more than have expectations that people will not use it (or abuse) in the scenario that it should not be available

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+#### `nil` Mempool
+
+We propose a new mempool implementation, called `nil` Mempool, that effectively disables all mempool functionality
+within CometBFT.
+The `nil` Mempool implements the `Mempool` interface in a very simple manner:
+
+*	`CheckTx(tx types.Tx) (*abcicli.ReqRes, error)`: returns `nil, ErrNotAllowed`
+*	`RemoveTxByKey(txKey types.TxKey) error`: returns `ErrNotFound`
+* `ReapMaxBytesMaxGas(maxBytes, maxGas int64) types.Txs`: returns `nil`
+* `ReapMaxTxs(max int) types.Txs`: returns `nil`
```
lasarojc, line 135: Instead of returning `nil`, could return an empty slice, matching the behavior in `flood`

lasarojc (approved):
Some suggestions, but overall LGTM

mzabaluev:
With the nil mempool, does ABCI++ provide enough information to implement optimized application-side tx propagation by, e.g. learning that the node will be a proposer for the next height, so its address can be gossiped for directed tx submissions? Is `PrepareProposal` generally too late for this?

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+
+These are the alternatives known to date:
+
+1. Keep the current model. Useful for basic apps, but clearly suboptimal for applications
+   with their own mechanism to disseminate transactions and particular performance requirements.
+2. Provide more efficient general-purpose mempool implementations.
+   This is an ongoing effort, but will take some time (and R&D effort) to come up with
+   advanced mechanisms -- likely highly configurable -- which then will have to be thoroughly tested.
+3. A similar approach to this one ([ADR110][adr-110]) whereby the application-specific
+   mechanism directly interacts with CometBFT via a newly defined gRPC interface.
```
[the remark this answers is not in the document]
zmanian: Publicly exposed broadcast tx on the validator mode is a Dos vulnerability. Level 1 is not secure.

docs/architecture/adr-111-nil-mempool.md
tac0turtle, line 166: I believe the reasoning here is because there is not a way to signify to comet to produce another block otherwise. Could we add a short sentence or two explaining this to avoid confusion for future readers.

docs/architecture/adr-111-nil-mempool.md
tac0turtle: instead of denying the rpc, id propose a short term solution of fire and forget. users can still submit via 26657 but checktx will fire and potentially not wait for the response, or discard the response. The reason is that that cosmjs puts a hard requirement on using tendermints rpc instead of the apps, this could cause lots of breakage for wallets and other clients

tac0turtle (commented):
i like this approach a lot more and it sets comet on a direction of absolving responsibility for things not crucial to consensus. 

Thank you for the well written ADR, left two comments.

docs/architecture/adr-111-nil-mempool.md
sergio-mena: The problem I have with "Application-side mempool", is that, for instance, what the SDK implemented in `v0.47.x` is also an "Application-side mempool", but (ATM) it does require a non-`nil` mempool on CometBFT's side.
I like like `nil`-mempool (or `null`-mempool, or stubbed mempool) as it points to what we are doing in CometBFT (not what the App can/must do when using our `nil`-mempool).

docs/architecture/adr-111-nil-mempool.md
sergio-mena: I'll add a short section on top of this one with a brief overview, as you suggest. But I'll leave this here, as it is context (not design)

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+(sometimes using _Reliable Broadcast_ [\[HT94\]][HT94] but not necessarily),
+is needed in a distributed system implementing State-Machine Replication (SMR).
+This is also the case in blockchains.
+Early designs such as Bitcoin or Ethereum include an _internal_ component,
+responsible for dissemination, called mempool.
+Tendermint Core chose to follow the same design given the success
+of those early blockchains.
+
+However, the design of ABCI clearly dividing the application logic (i.e., the appchain)
+and the consensus logic that provides SMR semantics to the app is a unique innovation in Cosmos
```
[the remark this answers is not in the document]
sergio-mena: I'd rather not introduce the p2p part here, as it doesn't add anything necessary to the reasoning going on here (all mentioned blockchains have some form of gossiping at the network level, so CometBFT is not _set apart_ in that respect)

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+it allows CometBFT to be used underneath (currently) tens of different appchains in production
+in the Cosmos ecosystem and elsewhere.
+But there are other implications for having an internal mempool
+in CometBFT: the interaction between the mempool, the application, and the network
+becomes more indirect, and thus more complex and hard to understand and operate.
+
+### ABCI++ Improvements and Remaining Shortcomings
+
+Before the release of ABCI++, `CheckTx` was the main mechanism the app had at its disposal to influence
+what made it to the mempool, and very indirectly what transactions got ultimately proposed in a block.
```
sergio-mena, line 39:
```suggestion
what transactions made it to the mempool, and very indirectly what transactions got ultimately proposed in a block.
```

docs/architecture/adr-111-nil-mempool.md
sergio-mena: Do you mean `app-mempool`?
If you do, I'd rather keep it general and not hint at one possible appchain mechanism (which is what an app mempool is)

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+
+## Alternative Approaches
+
+These are the alternatives known to date:
+
+1. Keep the current model. Useful for basic apps, but clearly suboptimal for applications
+   with their own mechanism to disseminate transactions and particular performance requirements.
+2. Provide more efficient general-purpose mempool implementations.
+   This is an ongoing effort, but will take some time (and R&D effort) to come up with
+   advanced mechanisms -- likely highly configurable -- which then will have to be thoroughly tested.
```
sergio-mena, line 111:
```suggestion
   advanced mechanisms -- likely highly configurable and thus complex -- which then will have to be thoroughly tested.
```

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+  using the appchains's own mechanism via `PrepareProposal`.
+
+## Alternative Approaches
+
+These are the alternatives known to date:
+
+1. Keep the current model. Useful for basic apps, but clearly suboptimal for applications
+   with their own mechanism to disseminate transactions and particular performance requirements.
+2. Provide more efficient general-purpose mempool implementations.
+   This is an ongoing effort, but will take some time (and R&D effort) to come up with
```
sergio-mena, line 110:
```suggestion
   This is an ongoing effort (e.g., [CAT mempool](https://github.com/cometbft/cometbft/pull/1472)), but will take some time (and R&D effort) to come up with
```

docs/architecture/adr-111-nil-mempool.md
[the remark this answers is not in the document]
sergio-mena: Level 1. If CometBFT's transactions are not gossiped by any full node, what is the point in having them on CometBFT's mempool? I mean, it's still gonna have all the interaction/config problems we have today (e.g., CometBFT not providing _all_ TX in `PrepProp` in some circumstances; what limits should the operator set on the mempool's capacity, etc). Also, the app will only learn about new TXs at `PrepProp` time, so it's only at that moment its own dissemination mechanism can kick in.
Level 2. Yes, this is the mechanism that allows us to say somewhere in this ADR that we're not offering anything that can't already be done with latest CometBFT version, but we're improving UX. I'll try to include your text here on one such reference to "nothing new, just better UX".

docs/architecture/adr-111-nil-mempool.md
sergio-mena: To the extent that I crawled back all callsites by hand and convinced myself there'll be no extra changes in other parts of the code.
But, when implementing, we might find something unexpected that might require this ADR to be updated (as we usually do).

docs/architecture/adr-111-nil-mempool.md
sergio-mena: I just quickly re-checked the call-site, and it doesn't make any difference.

docs/architecture/adr-111-nil-mempool.md
sergio-mena: Good point on not mentioning the RPC, will add a section.
To answer your question, my current understanding is that no code, except the (new) implementation of the `Mempool` interface will be changed/added (well, also the logic needed to be able to configure which mempool we will be using). So, the behavior of RPCs will be consistent with that model (size-related RPC methods will call the `Mempool` interface and thus return 0.

docs/architecture/adr-111-nil-mempool.md
sergio-mena: Not fully convinced (we might at CAT mempool land in the next weeks/months) but will change it to "local" as you folks seem to have a strong opinion on this one.

sergio-mena, line 158:
```suggestion
# - "local" : clist mempool with flooding gossip protocol (default)
```

docs/architecture/adr-111-nil-mempool.md
sergio-mena: Will do

docs/architecture/adr-111-nil-mempool.md
sergio-mena: Yes. Remember that, in previous team discussions:
* we struggled to see why a successful blockchain (and more generally an application under SMR) would need to temporarily stop consensus instances
* we realized the negative impact that temporarily stopping consensus instances had on other parts of CometBFT (e.g., having to wait for a block's apphash, which coming in the next block in order to be able to check proofs)

docs/architecture/adr-111-nil-mempool.md
sergio-mena: > is that based on the mempool configuration ?

Yes. Probably the error returned should be explicit about that _errored_ (e.g. "nil-mempool configured, which does not accept transactions")

> it can be disabled?
That is another option. Will mention it in the text

> The reason is that that cosmjs puts a hard requirement on using tendermints rpc instead of the apps, this could cause lots of breakage for wallets and other clients

Then they should not configure the nil-mempool until they've remove that requirement. If we return a silent response, the wallets, etc. will stay broken, as the behavior they're expecting won't materialize.
Don't forget that the proposal on this ADR is additive and _not_ the default config (the default is still the current mempool).

docs/architecture/adr-111-nil-mempool.md
sergio-mena: That's part of the discussion we'll need to have, and which I'd like to keep out of the scope of this ADR.

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -0,0 +1,239 @@
...
+for backporting changes to release branches that have already undergone a full QA cycle
+(and thus are in code-freeze mode).
+For this reason, although the limited impact of these changes would limit the risks
+of backporting to `v0.38.x` and `v0.37.x`, a careful risk/benefit evaluation will
+have to be carried out.
+
+Backporting to `v0.34.x` does not make sense as this version predates the release of `ABCI 1.0`,
+so using the `nil` mempool renders CometBFT's operation useless.
+
+>TODO: Need to add a small section here explaining the dilemma between a Consensus Param and a
```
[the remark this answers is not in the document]
sergio-mena: That's the point of my TODO. Bear with me while I work on the text

docs/architecture/adr-111-nil-mempool.md
mzabaluev: "local" sounds unclear. Does it refer to the mempool implementation being built-in, as opposed to out-of-process?

sergio-mena:
@mzabaluev Include some text explaining our approach allow for predicting next proposer(s)

docs/architecture/adr-111-nil-mempool.md
sergio-mena: The way I see it, in this particular case, there will be a risk of wasting resources by those nodes that have a misconfigured mempool (bandwidth, CPU, memory, etc) if they are submitted TXs via CometBFT's RPC.
It could get nasty, if there are TXs submitted (incorrectly) via CometBFT's RPC, but those TXs are never submitted (correctly) to the App. As those TXs risk being there for a long time.

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -126,7 +134,7 @@ TBD
 ## Detailed Design
 
 What this ADR proposes can already be achieved with an unmodified CometBFT since `v0.37.x`,
-albeit with a complex, poor UX.
+albeit with a complex, poor UX (see the last alternative in section [Alternative Approaches](#alternative-approaches)).
 This core of this proposal is to make some internal changes so it is clear an simple for app developers,
```
lasarojc, line 138:
```suggestion
The core of this proposal is to make some internal changes so it is clear an simple for app developers,
```

docs/architecture/adr-111-nil-mempool.md
melekes: I'm also against the silent responses and think we should return https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501

docs/architecture/adr-111-nil-mempool.md
melekes: > The type of mempool is not the same as the gossip protocol used by the mempool (if any)

💯 

> Not fully convinced (we might have CAT mempool landing on `main` in the next weeks/months) but will change it to "local" as you folks seem to have a strong opinion on this one.

Hmm, in that case, let's stick with `flood` or `clist` (`type CListMempool struct`) because of ^

docs/architecture/adr-111-nil-mempool.md
thanethomson: > Then they should not configure the nil-mempool until they've removed that requirement.

+1

> I'm also against the silent responses and think we should return https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501

Agreed. Technically the mechanism by which transactions _should_ be submitted when the nil mempool is enabled is via an API exposed by the application, executing `CheckTx` prior to insertion in the mempool.

The Cosmos SDK has an opportunity here to set the standard for such an API.

docs/architecture/adr-111-nil-mempool.md
```diff
@@ -286,10 +287,16 @@ In that sense, it is out of the scope of this ADR.
...
   to submit transactions and deliver those transactions to validators.
   This is a considerable endeavor, and more basic appchains may consider it is not worth the hassle.
-
+- There is a risk of wasting resources by those nodes that have a misconfigured
+  mempool (bandwidth, CPU, memory, etc). If there are TXs submitted (incorrectly)
+  via CometBFT's RPC, but those TXs are never submitted (correctly via an
+  app-specific interface) to the App. As those TXs risk being there until the node
+  is stopped. Moreover, those TXs will be replied & proposed every single block.
+  App developers will need to keep this in mind and panic on `CheckTx` or
+  `PrepareProposal` with non-empty list of transactions.
```
sergio-mena, line 299: Can you add a further disadvantage (it slipped through my mind) saying that optimizing block proposals by only including transaction IDs (e.g. TX hashes) is now more difficult as the mempool is outside CometBFT? (For context, this is one of the optimizations we have in mind for bandwidth).
TBH, the app could do it by submitting TX hashes (rather than TXs themselves) in prepare proposal, and then having a mechanism for pulling TXs from the network upon `FinalizeBlock`.

thanethomson (approved):
:tada:

docs/architecture/adr-111-nop-mempool.md
```diff
@@ -0,0 +1,308 @@
...
+  mempool (bandwidth, CPU, memory, etc). If there are TXs submitted (incorrectly)
+  via CometBFT's RPC, but those TXs are never submitted (correctly via an
+  app-specific interface) to the App. As those TXs risk being there until the node
+  is stopped. Moreover, those TXs will be replied & proposed every single block.
+  App developers will need to keep this in mind and panic on `CheckTx` or
+  `PrepareProposal` with non-empty list of transactions.
+- Optimizing block proposals by only including transaction IDs (e.g. TX hashes) is more difficult.
+  The ABCI app could do it by submitting TX hashes (rather than TXs themselves)
+  in `PrepareProposal`, and then having a mechanism for pulling TXs from the
+  network upon `FinalizeBlock`.
```
sergio-mena, line 303: Excellent! Thanks for this.

sergio-mena (commented):
LGTM 🙏 
Approved (Can't approve as I am the initial submitter)

PR #1588, opened 2023-11-10
fix: docs: default db provider moved from node to config

Plopmenz:
Small correction of the guide to setup a built-in cometbft application. Current example code gave me an error (DefaultDBProvider does not exist in nm/node), I believe this should be cfg/config instead. The code bellow is working for me.

I did not add anything to the changelog as that seems kinda unnecessary for such a small change to just the docs.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/docs/guides/go-built-in.md
+++ b/docs/guides/go-built-in.md
@@ -586,7 +586,7 @@ func main() {
         nodeKey,
         proxy.NewLocalClientCreator(app),
         nm.DefaultGenesisDocProviderFunc(config),
-        nm.DefaultDBProvider,
+        cfg.DefaultDBProvider,
         nm.DefaultMetricsProvider(config.Instrumentation),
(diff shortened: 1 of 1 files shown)


melekes (approved):
Thanks @Plopmenz 🙏

melekes:
@Plopmenz could you please also make the exact change on line 679?

Plopmenz:
Good catch! Done :)

melekes:
@mergify backport v0.38.x

sergio-mena:
@melekes Thanks for backporting this to `v0.38.x`. Does it make sense to also backport to `1.x`?

melekes:
> @melekes Thanks for backporting this to `v0.38.x`. Does it make sense to also backport to `1.x`?

v1 has these changes already.

PR #1593, opened 2023-11-11
chore: aligns function descriptions with godoc standards in the types package

staheri14:
This PR resolves inconsistencies in the `types` package, particularly in functions where their descriptions either do not follow the Godoc convention (where a comment describing a function should start with the function's name) or exhibit discrepancies with the actual function name.

#### PR checklist

- [ ] Tests written/updated


types/tx_test.go
```diff
@@ -123,7 +123,7 @@ func testTxProofUnchangable(t *testing.T) {
 	}
 }
 
-// This makes sure that the proof doesn't deserialize into something valid.
+// assertBadProof makes sure that the proof doesn't deserialize into something valid.
```
thanethomson, line 126: We don't follow this convention for private methods.

thanethomson (commented):
Thanks for this @staheri14!

melekes (approved):
👍

types/tx_test.go
adizere: Does this change need to be reverted, or can we accept the PR despite this change?

Seems good to me to have the consistency in comments, regardless of this being private method.

adizere (approved):

lasarojc (approved):

thanethomson (approved):

types/tx_test.go
thanethomson: It's fine, it's just inconsistent with the rest of the codebase, which isn't so great.

staheri14: Thank you for providing clarification, @thanethomson! Regarding consistency within the codebase, I've noticed that there are instances where private methods adhere to this convention [here](https://github.com/cometbft/cometbft/blob/38a4caeac0551c188af8a3e209e48380cd514e2d/statesync/chunks.go#L43). However, if it has been decided that this convention should not be followed strictly in the cometbft codebase, I will take note of this for future PRs.

Issue #1583, opened 2023-11-08
Error in the introductory guides


lasarojc:
## Bug Report

The introductory `go.md` and `go-builtin.md` guides have a wrong example of `prepareProposal`, as noted [here] (https://github.com/cometbft/cometbft/pull/1384#discussion_r1342041056)

```
func (app *KVStoreApplication) PrepareProposal(_ context.Context, proposal *abcitypes.RequestPrepareProposal) (*abcitypes.ResponsePrepareProposal, error) {
   totalBytes := int64(0)
   txs := make([]byte, 0)

   for _, tx := range proposal.Txs {
     totalBytes += int64(len(tx))
     txs = append(txs, tx...)

       if totalBytes > int64(proposal.MaxTxBytes) {
         break
       }
     }

     return &abcitypes.ResponsePrepareProposal{Txs: proposal.Txs}, nil
 }
```

First, `txs` should be a slice of slices.
Second, the overflow check should be done before appending the transaction to `txs`.

The fix need to be back ported to 0.38 and 0.37.

andynog:
Thanks @lasarojc for pointing this out. I'll take on this and fix the docs and backport it

melekes:
I agree with @lasarojc [here](https://github.com/cometbft/cometbft/pull/1603#discussion_r1391319868) that it's an intro document and we shouldn't overwhelm devs with details. Simply relaying existing transactions without modifying them is enough, imho.

PR #1603, opened 2023-11-13
docs: various improvements

melekes:
These are various minor improvements/fixes to the docs. It should be reviewed commit by commit.

Closes https://github.com/cometbft/cometbft/issues/1583

#### PR checklist

- [ ] Tests written/updated


docs/guides/go-built-in.md
```diff
@@ -101,10 +101,18 @@ github.com/cometbft/cometbft v0.38.0
 )
 ```
 
+XXX: CometBFT `v0.38.0` uses a slightly outdated `gogoproto` library, which
+may fail to compile with newer Go versions. To avoid any compilation errors,
+upgrade `gogoproto` manually:
+
+```bash
+go get github.com/cosmos/gogoproto@v1.4.11
```
thanethomson, line 109: Should we upgrade the gogoproto version we use on the `v0.38.x` branch?

thanethomson (approved):

docs/guides/go-built-in.md
```diff
@@ -457,30 +465,15 @@ The application is free to modify the group before returning from the call, as l
...
-       if totalBytes > int64(proposal.MaxTxBytes) {
-         break
-       }
-     }
-
-     return &abcitypes.ResponsePrepareProposal{Txs: proposal.Txs}, nil
- }
+func (app *KVStoreApplication) PrepareProposal(_ context.Context, proposal *abcitypes.RequestPrepareProposal) (*abcitypes.ResponsePrepareProposal, error) {
+    return &abcitypes.ResponsePrepareProposal{Txs: proposal.Txs}, nil
+}
```
lasarojc, line 474: It is possibly now that the list of transactions returned is bigger that what fits in the block. So the app is supposed to reduce the list of transactions to fit in the block.
Given that this is an intro document, I am fine with the simpler document, but wonder if we should have a warning here.

lasarojc (approved):

lasarojc:
Please add the backkport tags before merging.

docs/guides/go-built-in.md
melekes: > It is likely now that the list of transactions returned is bigger than what fits in the block.

Could you please explain how? CometBFT checks the size when reaping transactions from the mempool, right? So naturally, if the application does not modify the proposed transactions list, then it can safely skip the check, no?

melekes:
v0.37.x & v0.34.x documentations are correct concerning the changes in this PR — **no need to backport**.

docs/guides/go-built-in.md
melekes: https://github.com/cometbft/cometbft/pull/1609

docs/guides/go-built-in.md
lasarojc: This behavior was changed to allow reaping all the txs. See the following excerpt of https://docs.cometbft.com/v0.38/spec/abci/abci++_methods#prepareproposal

"The Application MAY configure CometBFT to include a list of transactions in RequestPrepareProposal.txs whose total size in bytes exceeds RequestPrepareProposal.max_tx_bytes. If the Application sets ConsensusParams.Block.MaxBytes to -1, CometBFT will include all transactions currently in the mempool in RequestPrepareProposal.txs, which may not fit in RequestPrepareProposal.max_tx_bytes. Therefore, if the size of RequestPrepareProposal.txs is greater than RequestPrepareProposal.max_tx_bytes, the Application MUST remove transactions to ensure that the RequestPrepareProposal.max_tx_bytes limit is respected by those transactions returned in ResponsePrepareProposal.txs. "

melekes: I didn't know that! I believe we all agree that the check shouldn't be in an introduction guide because the **default for `MaxBytes` is 4MB, not `-1`**. I will open a separate issue with documentation improvements and include this item. Thanks again 🙏

Issue #1484, opened 2023-10-11
ADR 109: Reduce Go API surface area


thanethomson:
At present, the surface area of the codebase is so large that it makes it difficult for the team to make significant changes quickly. It would be better if we had some way of minimizing this surface area prior to the next major release such that we can, in future, make changes quicker.

We should ideally implement these changes prior to cutting v1.

This is a tracking issue for the implementation of #1483.

- [x] #1485 
- [x] #1488 
- [x] #1605

PR #1610, opened 2023-11-14
Add public api `BlockStore.Rollback`

arno01:
Based on tendermint/tendermint#8574/81344ac9464db02421fa115041f162ab9ace9372

----

I've successfully tested it with [akash `0.26.2`](https://github.com/akash-network/node/tree/v0.26.2) (cosmos-sdk `v0.45.16`, tendermint `v0.34.27`) following [these](https://gist.github.com/andy108369/da1279257c018be6370310c990c25738) steps.

And [akash `0.16.4`](https://github.com/akash-network/node/tree/v0.16.4) (cosmos-sdk `v0.45.4`, tendermint `v0.34.19`) back in [`Jul 30, 2022`](https://github.com/cosmos/cosmos-sdk/pull/11361#issuecomment-1200469670) following [these](https://gist.github.com/andy108369/44f5a676935286e0115431015ef66e1c) steps.

- [ ] (`cosmos-sdk`) I am wondering whether [`deletePendingBlockIfExists`](https://github.com/cosmos/cosmos-sdk/pull/11982/commits/614906d69177db9555ea16ebb3e895b2f4617ff3#diff-e6374d17b47e21bb322f908dc81b6a544d9c57ca2569b1e9b28572af68551963R43) is actually also need to be backported to `cosmos-sdk` (the `rollback --delete-pending-block` is what I've tested in Akash);  It appears that the `deletePendingBlockIfExists` itself did not go to `cosmos-sdk` upstream cosmos/cosmos-sdk@51d2de582d036ece4af10267e889f6fdd97f8acf

cc @yihuang 

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -349,28 +349,8 @@ func (bs *BlockStore) PruneBlocks(height int64, state sm.State) (uint64, int64, 
 			evidencePoint = h
 		}
 
-		// if height is beyond the evidence point we dont delete the header
-		if h < evidencePoint {
-			if err := batch.Delete(calcBlockMetaKey(h)); err != nil {
-				return 0, -1, err
-			}
-		}
-		if err := batch.Delete(calcBlockHashKey(meta.BlockID.Hash)); err != nil {
+		if err := bs.deleteBlock(batch, meta, evidencePoint); err != nil {
 			return 0, -1, err
-		}
-		// if height is beyond the evidence point we dont delete the commit data
-		if h < evidencePoint {
-			if err := batch.Delete(calcBlockCommitKey(h)); err != nil {
-				return 0, -1, err
-			}
-		}
-		if err := batch.Delete(calcSeenCommitKey(h)); err != nil {
-			return 0, -1, err
-		}
-		for p := 0; p < int(meta.BlockID.PartSetHeader.Total); p++ {
-			if err := batch.Delete(calcBlockPartKey(h, p)); err != nil {
-				return 0, -1, err
-			}
 		}
 		pruned++
 
@@ -390,6 +370,34 @@ func (bs *BlockStore) PruneBlocks(height int64, state sm.State) (uint64, int64, 
 		return 0, -1, err
 	}
 	return pruned, evidencePoint, nil
+}
+
+func (bs *BlockStore) deleteBlock(batch dbm.Batch, meta *types.BlockMeta, evidencePoint int64) error {
+	h := meta.Header.Height
+	// if height is beyond the evidence point we dont delete the header
+	if h < evidencePoint {
+		if err := batch.Delete(calcBlockMetaKey(h)); err != nil {
+			return err
+		}
+	}
+	if err := batch.Delete(calcBlockHashKey(meta.BlockID.Hash)); err != nil {
+		return err
+	}
+	// if height is beyond the evidence point we dont delete the commit data
+	if h < evidencePoint {
+		if err := batch.Delete(calcBlockCommitKey(h)); err != nil {
+			return err
+		}
+	}
+	if err := batch.Delete(calcSeenCommitKey(h)); err != nil {
+		return err
+	}
+	for p := 0; p < int(meta.BlockID.PartSetHeader.Total); p++ {
+		if err := batch.Delete(calcBlockPartKey(h, p)); err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 // SaveBlock persists the given block, blockParts, and seenCommit to the underlying db.
@@ -538,6 +546,35 @@ func (bs *BlockStore) SaveSeenCommit(height int64, seenCommit *types.Commit) err
 
 func (bs *BlockStore) Close() error {
 	return bs.db.Close()
+}
+
+// Rollback rollbacks the latest block from BlockStore.
+func (bs *BlockStore) Rollback() error {
+	if bs.height <= 0 {
+		return fmt.Errorf("can't rollback height %v", bs.height)
+	}
+
+	meta := bs.LoadBlockMeta(bs.height)
+	if meta == nil {
+		return fmt.Errorf("block not found: %v", bs.height)
+	}
+
+	batch := bs.db.NewBatch()
+	defer batch.Close()
+	// Passing bs.height as the evidencePoint to satisfy the function signature.
+	// This assumes that rollback does not need to consider the evidencePoint logic
+	if err := bs.deleteBlock(batch, meta, bs.height); err != nil {
+		return fmt.Errorf("failed to delete block %v: %w", bs.height, err)
+	}
+
+	if err := batch.WriteSync(); err != nil {
+		return fmt.Errorf("failed to delete block %v: %w", bs.height, err)
+	}
+
+	bs.height--
+	bs.saveState()
+
+	return nil
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


store/store.go
```diff
@@ -539,6 +561,33 @@ func (bs *BlockStore) Close() error {
 	return bs.db.Close()
 }
 
+// Rollback rollbacks the latest block from BlockStore.
+func (bs *BlockStore) Rollback() error {
```
melekes, line 565: It would be great to have a test for this

melekes (commented):
Thanks for copying this over from tendermint 👍

store/store.go
arno01: added the tests

```
$ go test ./store -run TestRollback -v
=== RUN   TestRollback
=== RUN   TestRollback/Successful_Rollback
=== RUN   TestRollback/Rollback_at_Height_Zero
=== RUN   TestRollback/Rollback_on_Empty_Store
=== RUN   TestRollback/Rollback_State_Consistency
--- PASS: TestRollback (0.01s)
    --- PASS: TestRollback/Successful_Rollback (0.00s)
    --- PASS: TestRollback/Rollback_at_Height_Zero (0.00s)
    --- PASS: TestRollback/Rollback_on_Empty_Store (0.00s)
    --- PASS: TestRollback/Rollback_State_Consistency (0.00s)
PASS
ok  	github.com/cometbft/cometbft/store	0.016s
```

store/store.go
```diff
@@ -539,6 +561,33 @@ func (bs *BlockStore) Close() error {
...
+	defer batch.Close()
+	if err := bs.deleteBlock(batch, meta); err != nil {
+		return fmt.Errorf("failed to delete block %v: %w", bs.height, err)
+	}
+
+	bs.height--
+	bs.saveState()
+
+	if err := batch.WriteSync(); err != nil {
+		return fmt.Errorf("failed to delete block %v: %w", bs.height, err)
```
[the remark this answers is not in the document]
arno01: yeah, that was a good idea I think.
I've pushed the code changes and the tests for the failing WriteSync() :-)

image

after the fix:

image

image

arno01:
@melekes I believe I've addressed all concerns. And have added the tests including the test for failing WriteSync() scenario.

yihuang:
cosmos-sdk already provides a `rollback` command, does that help?

arno01:
> cosmos-sdk already provides a `rollback` command, does that help?
> 
> [cosmos/cosmos-sdk#11179](https://github.com/cosmos/cosmos-sdk/pull/11179)

`cosmos-sdk`'s is not enough, at least `0.45.16` can't be built without the `blockStore.Rollback` in `cometbft`, where `blockStore.Rollback` (in cometbft; this PR) is part of `deletePendingBlockIfExists`  [①](https://github.com/cosmos/cosmos-sdk/pull/11982/commits/614906d69177db9555ea16ebb3e895b2f4617ff3#diff-e6374d17b47e21bb322f908dc81b6a544d9c57ca2569b1e9b28572af68551963R114)

```
cosmos-sdk$ make build-linux 
...
server/rollback.go:116:24: blockStore.Rollback undefined (type *"github.com/tendermint/tendermint/store".BlockStore has no field or method Rollback)
```

I am not certain if the `rollback`, in the higher versions of `cosmos-sdk` (>=`0.46.0`), takes into account the **pending blocks** as higher `cosmos-sdk` versions do not have the `rollback --delete-pending-block` CLI arg. (`deletePendingBlockIfExists`).

The `deletePendingBlockIfExists` bit itself never went to `cosmos-sdk` upstream cosmos/cosmos-sdk@51d2de582d036ece4af10267e889f6fdd97f8acf


After this PR to `cometbft` we can then issue a RR to `cosmos-sdk` that brings back `deletePendingBlockIfExists` that enables `rollback --delete-pending-block` CLI command if it is necessary in `cosmos-sdk` of `>=0.46.0` version.

However, I am not 100% certain whether `deletePendingBlockIfExist` is required in `cosmos-sdk` >=`0.46.0` as per @yihuang 's comment back then https://github.com/cosmos/cosmos-sdk/pull/11982/commits/614906d69177db9555ea16ebb3e895b2f4617ff3#r875526169

yihuang:
You don't need to delete pending block to fix hashes mismatch I think?

yihuang:
Yeah, delete pending block is not landed in upstream, it's indeed useful in some edge cases.

arno01:
> Yeah, delete pending block is not landed in upstream, it's indeed useful in some edge cases.

Yeah, we've managed to restore three huge (3.6TiB) archival Akash nodes quickly thanks to that feature :-)
And generally, whenever there is a non-gov network upgrade happens (via binary swap), there is a very high chance of a "bad" pending block be proposed while the 2/3+ VP of the older binary is still running (for brief time during the switch).

This often leads the validators resorting to use the same snapshot so they start the new network binary off of the same height.

internal/store/store.go
```diff
@@ -540,6 +548,35 @@ func (bs *BlockStore) Close() error {
 	return bs.db.Close()
 }
 
+// Rollback rollbacks the latest block from BlockStore.
+func (bs *BlockStore) Rollback() error {
```
thanethomson, line 552: How do you intend on using this if the store's in `internal`?

arno01: it wasn't `internal` >2 days ago :thinking:  https://github.com/cometbft/cometbft/commit/10fa8d1016d98f8395be47cd2a9c40f2ae03e7bd#diff-b26d414d0eb20deebca0d2b785bd3e1090aa1a7a182e7dc30640d68c53cb40da

Ideas?

thanethomson: How were you intending on using it in the first place?

See ADR 109 (#1484) for context on the internalization of a number of our public APIs.

thanethomson: Technically, if you access the block store via the node's `BlockStore` method you should still be able to access the `Rollback` function.

We cannot, however, guarantee the stability of Go APIs that are defined within `internal`.

arno01: > How were you intending on using it in the first place?

I've described that in a step-by-step guide here https://gist.github.com/andy108369/da1279257c018be6370310c990c25738

thanethomson: What'd help is if we could understand where you want to call that function from. Would it be called from within Comet's own rollback command, within this SDK call here? https://github.com/cosmos/cosmos-sdk/blob/7799bba7bc889288607576aa11655b5fc31a2da9/server/rollback.go#L35

arno01: I believe within the SDK call based on the step-by-step guide shared prior.
Or whatever makes it easier to just run `akash rollback --delete-pending-block` and make sure it deletes it. :-)

thanethomson: Is there any reason not to modify the publicly exported `RollbackState` function in Comet to take a `deletePendingBlock` parameter? https://github.com/cometbft/cometbft/blob/ee99bf534a35f56f0e5ad463d93bbe1d796bf9f0/cmd/cometbft/commands/rollback.go#L54

thanethomson: If you want to do this in a non-breaking way (in terms of the public Go API), simply create a new publicly exported function there called `RollbackStateAndPendingBlock` that calls `RollbackState` and deletes the pending block.

arno01: Thank you @thanethomson ! I'll try it.

thanethomson: Hey @arno01, just checking in here to see whether you're going to implement this?

PR #1616, opened 2023-11-15
pruning:do not attempt to prune state if no blocks are pruned

### LINKED ISSUE #1615: pruning: Should not attempt to prune state store if no blocks were pruned
### Setup

**CometBFT version** 

main; v0.38.x-experimental 

yes 

### Config

Pruning is enabled. 

### What happened?

In the extreme case where pruning is  triggered while the chain is at height 1 and the retain height is set to 1, no blocks can be pruned as they should not be. However, we still attempt to prune the state store and get an error that pruning failed. 

If not blocks are pruned, there should be no attempt to prune the states. 

### How to reproduce it

The bug appeared in subsequent runs of the `TestProvider` test in `http_test`. The test added with the fix would fail if not for the changes fixing it.

jmalicevic:
Closes #1615 

Needs a backport to 0.38.x-experimental

---

#### PR checklist

- [x] Tests written/updated
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments - **not needed**


state/export_test.go
```diff
@@ -64,6 +64,10 @@ func (p *Pruner) PruneBlockIndexerToRetainHeight(lastRetainHeight int64) int64 {
 	return p.pruneBlockIndexerToRetainHeight(lastRetainHeight)
 }
 
+func (p *Pruner) PruneBlocksToHeight(height int64) (uint64, int64, error) {
```
mzabaluev, line 67: Just rename the private method?

mzabaluev (approved):

melekes (approved):
👍

lasarojc:
Blocked by #1619

state/export_test.go
jmalicevic: @thanethomson , @mzabaluev , we used to favor this approach as to avoid exposing functions that we use internally if they are exposed just for testing. I don't mind altering the code but want to make sure we are all on the same page.

thanethomson: Ah I see. Yes @mzabaluev if it's just exposed for testing purposes, we basically create a public wrapper method in a `_test.go` file. These `_test.go` files are not built when building the binary, but are built when running tests.

thanethomson (approved):

PR #1628, opened 2023-11-15
Adds tests that check for FIFO ordering being broken by gossip

### LINKED ISSUE #1620: FIFO ordering is violated by the clist_mempool
Tests have failed demonstrating that FIFO ordering is not ensured by the clist_mempool.
On instance [1](1), it was on a branch modifying the mempool but on instance [2](2) it was on a test that just limited the concurrency of the `broadcastTX` routine.
Resetting the clist `next` pointer when a block is decided could cause a transaction to be sent multiple times and filling the mempool could cause transactions to not be added to the pool, but these are not the cause here as the consensus was not playing a part in the test and the number of transactions was small.

[1]: https://github.com/cometbft/cometbft/pull/1169
[2]: https://github.com/cometbft/cometbft/actions/runs/6867258327/job/18675201973?pr=1611

lasarojc:
Closes #1620

This PR adds test `TestMempoolFIFOWithParallelCheckTx`, which fails frequently, demonstrating that FIFO is not preserved even in the absence of failures, disconnections and retransmissions. This test is kept disabled during CI, but may be verified to fail by running `go test -race -v -count=100 -failfast  -run TestMempool     ./mempool/...`

The test makes the same list of transactions be added in parallel to the same mempool, simulating receiving the transactions from different sources. At the end of the execution, the list in the mempool is different from the original list.

The differences arise in [`clist_mempool::CheckTx`](https://github.com/cometbft/cometbft/blob/1c907856fa2e22ceadb806453f017138092f46d9/mempool/clist_mempool.go#L251) because of context changes between adding the tx to the cache (`if added := mem.addToCache(tx); !added {`) and adding it to the mempool (`mem.proxyAppConn.CheckTxAsync()`)
or simply because the acquiring of mutex inside `mem.proxyAppConn.CheckTxAsync` does not follow the order of request.

Other tests have been added to show that if care is taken to avoid delivering the same same sequence of transactions in parallel (a likely situation), FIFO is ensured.
These may be run with  `CI=true go test -race -v -v -v -count=3 -failfast  -run TestMempool     ./mempool/...`

---

#### PR checklist

- [x] Tests written/updated


mempool/reactor.go
```diff
@@ -270,6 +270,7 @@ func (memR *Reactor) broadcastTxRoutine(peer p2p.Peer) {
 		// https://github.com/tendermint/tendermint/issues/5796
 
 		if !memR.isSender(memTx.tx.Key(), peer.ID()) {
+			memR.Logger.Info("TX BCAST", "string", memTx.tx.String())
```
lasarojc, line 273:
```suggestion
```

lasarojc:
The `TestMempoolParallelCheckTx` test fails if run multiple times as in `go test -v -v -v -count=100 -failfast  -run TestMempoolReactorMaxActiveOutboundConnections    ./mempool/...`

melekes (approved):
👍

mempool/reactor_test.go
```diff
@@ -333,6 +334,43 @@ func TestReactorTxSendersMultiNode(t *testing.T) {
...
+	}
+
+	// Deliver the same sequence of transactions from multiple sources, in parallel.
+	txs := newUniqueTxs(20)
+	mp := reactors[0].mempool
+	for i := 0; i < 3; i++ {
+		go func() {
+			for _, tx := range txs {
+				_, err := mp.CheckTx(tx)
+				assert.Nil(t, err)
```
hvanz, line 365: I'm getting an error "tx already exists in cache", which is expected but it's not the goal of this test.

mempool/reactor_test.go
```diff
@@ -333,6 +334,43 @@ func TestReactorTxSendersMultiNode(t *testing.T) {
...
+		}
+	}()
+	for _, r := range reactors {
+		for _, peer := range r.Switch.Peers().List() {
+			peer.Set(types.PeerStateKey, peerState{1})
+		}
+	}
+
+	// Deliver the same sequence of transactions from multiple sources, in parallel.
+	txs := newUniqueTxs(20)
```
hvanz, line 359:
```suggestion
	txs := newUniqueTxs(100)
```

hvanz: I need to increase the number of txs to see the error on my machine. With 20, I never see it. Starting with 100 I see it consistently.

lasarojc: Even when running a sequence of 100 tests? As in 
`go test -race -v -count=100 -failfast  -run TestMempool     ./mempool/...`

mempool/reactor_test.go
lasarojc: Thanks @hvanz
Fixe the code to not break on this error, just to output it and confirmed that it fails because of the inversion only.

PR #1629, opened 2023-11-15
Fixes back port of test in 0.37

lasarojc:
Contributes to #1620

This commit makes the test in 0.37 be the same as in main, that is, to ignore the order of transactions in the receiving reactor.
#1620 is not fixed by this PR. Once #1620 is fixed, it needs to be back ported as well.

#1627 does the same work as this PR, but for 0.38

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

PR #1654, opened 2023-11-17
Do not block indefinitely on the semaphore

lasarojc:
From time to time stop waiting on the semaphore limiting concurrency to check if the peer is still connected.

---

#### PR checklist

- [ ] Tests written/updated


mempool/reactor.go
```diff
@@ -103,36 +103,46 @@ func (memR *Reactor) GetChannels() []*p2p.ChannelDescriptor {
...
 
-				if !peer.IsPersistent() && memR.config.ExperimentalMaxGossipConnectionsToNonPersistentPeers > 0 {
-					// Block sending transactions to peer until one of the connections become
-					// available in the semaphore.
-					if err := memR.activeNonPersistentPeersSemaphore.Acquire(context.TODO(), 1); err != nil {
-						memR.Logger.Error("Failed to acquire semaphore: %v", err)
-						return
+				if peerSemaphore != nil {
+					for peer.IsRunning() {
+						ctxTimeout, cancel := context.WithTimeout(ctxParent, 30*time.Second)
```
melekes, line 122: Have you checked if it's possible to mount a DoS attack with a 30s timeout? How did you choose this value?

mempool/reactor.go
```diff
@@ -103,36 +103,46 @@ func (memR *Reactor) GetChannels() []*p2p.ChannelDescriptor {
...
+						// Block sending transactions to peer until one of the connections become
+						// available in the semaphore.
+						err := peerSemaphore.Acquire(ctxTimeout, 1)
+						cancel()
+
+						if err != nil {
+							select {
+							case <-ctxParent.Done():
```
melekes, line 130: `TODO()` context is never done, no? If so, can we remove it?

mempool/reactor.go
```diff
@@ -103,36 +103,46 @@ func (memR *Reactor) GetChannels() []*p2p.ChannelDescriptor {
...
-					// available in the semaphore.
-					if err := memR.activeNonPersistentPeersSemaphore.Acquire(context.TODO(), 1); err != nil {
-						memR.Logger.Error("Failed to acquire semaphore: %v", err)
-						return
+				if peerSemaphore != nil {
+					for peer.IsRunning() {
+						ctxTimeout, cancel := context.WithTimeout(ctxParent, 30*time.Second)
+						// Block sending transactions to peer until one of the connections become
+						// available in the semaphore.
+						err := peerSemaphore.Acquire(ctxTimeout, 1)
```
melekes, line 125: Does it make sense to move `defer peerSemaphore.Release(1)` here? Otherwise, it's never executed when you `return` on line `131`.

mempool/reactor.go
lasarojc: I don't have hard evidence but expect that 30 seconds wouldn't be enough to mount an attack because the go-routine being accumulated here is very small. Then again, we can lower the value. What do you think would be a more sensible value?

mempool/reactor.go
lasarojc: I don't think so. If there was an error, the semaphore wasn't acquired and releasing it would be wrong.

mempool/reactor.go
lasarojc: We can remove it, but I added it here to set up the the right structure for when contexts are properly used.
We should at least leave it as a comment; do you agree?

config/config.go
```diff
@@ -877,12 +877,13 @@ type MempoolConfig struct {
...
+	// We use two independent upper values for persistent and non-persistent peers.
 	// Unconditional peers are not affected by this feature.
 	// If we are connected to more than the specified number of persistent peers, only send txs to
-	// the first ExperimentalMaxGossipConnectionsToPersistentPeers of them. If one of those
-	// persistent peers disconnects, activate another persistent peer. Similarly for non-persistent
-	// peers, with an upper limit of ExperimentalMaxGossipConnectionsToNonPersistentPeers.
+	// `ExperimentalMaxGossipConnectionsToPersistentPeers` of them. If one of those
+	// persistent peers disconnects, activate another persistent peer.
+	// Similarly for non-persistent peers, with an upper limit of
+	// `ExperimentalMaxGossipConnectionsToNonPersistentPeers`.
```
thanethomson, line 886: Should this be updated in `config/toml.go` as well?

thanethomson (commented):
I'm still skeptical of the overall semaphore-based approach because it's hard to understand (and therefore it'd be hard to debug if we encounter corner cases we didn't anticipate). Just looking at this code in isolation, you have to think about how `AddPeer` can now conditionally block the calling goroutine (should possibly be documented in the `AddPeer` docstring).

Then we not only have one semaphore, but two that could block the execution of this function.

Then, it's tricky to reason about the acquisition and releasing of the semaphores.

Is there not a conceptually simpler way to handle this?

config/config.go
lasarojc: good point

lasarojc:
> I'm still skeptical of the overall semaphore-based approach because it's hard to understand (and therefore it'd be hard to debug if we encounter corner cases we didn't anticipate).  ... 

On an early version the semaphore were used inside `broadcastTx` itself, but that seemed worse for me; more confusing.
There are other approaches, using channels and mutexes, but they all seem to be reimplementing the semaphore.
Let's wait what other reviewers think; they may have simpler solutions.

With regards to `addPeer` blocking, that is not the case because the function body is encapsulated in its own go routine.

mempool/reactor.go
melekes: I'd simulate the attack by connecting and disconnecting from the node, which might be an overstretch for this PR. Feel free to leave the current value.

mempool/reactor.go
melekes: > We should at least leave it as a comment; do you agree?

sure

hvanz:
I agree with @lasarojc that semaphores is the way to go here. It's exactly made for what we need: to bound concurrent access to a resource. I guess we should better comment this part of the code because it's not easy to reason about concurrency, with or without the semaphore.

mempool/reactor.go
```diff
@@ -103,36 +103,50 @@ func (memR *Reactor) GetChannels() []*p2p.ChannelDescriptor {
 // AddPeer implements Reactor.
 // It starts a broadcast routine ensuring all txs are forwarded to the given peer.
 func (memR *Reactor) AddPeer(peer p2p.Peer) {
+	ctxParent := context.TODO() // This context should come from the caller.
```
melekes, line 106: Comment out?

mempool/reactor.go
lasarojc: Ok. Will leave this value for now.

mempool/reactor.go
lasarojc: Done. Cleanup up the other comment as well.

melekes (approved):

PR #1488, opened 2023-10-12
ADR 109: Modularize test infra

thanethomson:
Partially addresses #1484.

---

#### PR checklist

- [ ] Tests written/updated
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments

PR #1672, opened 2023-11-21
feature: Protobuf upgrade

thanethomson:
The primary contribution to #1330.

Must be merged by way of a merge commit (i.e. please do not squash).

Individual PRs have already been reviewed.

:heavy_check_mark: Manual run of the E2E nightlies on the `feature/proto-upgrade` branch: https://github.com/cometbft/cometbft/actions/runs/7025265160

---

#### PR checklist

- [ ] Tests written/updated
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments

PR #1676, opened 2023-11-22
Revert modularization of the test infra (#1488)

thanethomson:
This reverts commit e4a82ef64128786e1e2ec3a5d60a3be3a48918d9.

Modularizing the test infra, while useful for clearly separating E2E test dependencies from production ones, results in an incredibly painful development workflow that doesn't seem to be worth the cost. If we do want to clearly separate this in future, I suspect putting it into its own repository (for which there doesn't seem to be much benefit now either).

---

#### PR checklist

- [ ] Tests written/updated


.github/workflows/lint.yml
```diff
@@ -36,12 +28,11 @@ jobs:
         with:
           PATTERNS: |
             **/**.go
-            **/go.mod
-            **/go.sum
+            go.mod
+            go.sum
```
mzabaluev, line 32: Should there be a match for `api/go.{mod,sum}` as well, or should this rule be modified again before merging #1672?

thanethomson: That change should land on the `feature/proto-upgrade` branch.

melekes (approved):

lasarojc (approved):

PR #1677, opened 2023-11-22
Promote latest versions of Protobuf definitions to v1 packages

### LINKED ISSUE #1649: proto: Promote the latest message definitions into final `v1` packages
At present, the latest versions of the Protobuf definitions are all `v1beta*`. The versions of the types we will be using in CometBFT v1 must be `v1` instead.

For example, [`proto/cometbft/abci/v1beta4`](https://github.com/cometbft/cometbft/tree/feature/proto-upgrade/proto/cometbft/abci/v1beta4) should actually be `proto/cometbft/abci/v1` (or the `v1` folder should contain the latest, most up-to-date definitions of the types currently used by CometBFT v1).

mzabaluev:
Closes #1649

---

#### PR checklist

- [ ] Tests written/updated


mzabaluev:
Ignore the `./api/*.pb.go` changes as these are generated files.

proto/cometbft/abci/v1beta4/types.proto
```diff
@@ -4,7 +4,7 @@ package cometbft.abci.v1beta4;
 import "cometbft/abci/v1beta1/types.proto";
 import "cometbft/abci/v1beta2/types.proto";
 import "cometbft/abci/v1beta3/types.proto";
-import "cometbft/crypto/v1beta1/proof.proto";
```
mzabaluev, line 7: Retroactive changes in `v1beta*` packages update references to other packages which were simply renamed from `v1beta1` to `v1`, because there has only ever been one version in `crypto`, `libs/bits`, etc. and it does not make sense to produce duplicate definitions for those domains.

The rest of the changes to `**/v1beta*/*` is whitespace formatting: I had a different formatter setup on file save prior to this, but this style is true to clang-format.

mzabaluev: I'm also not sure if we should preserve the latest beta packages like `cometbft.abci.v1beta4` here: they are functionally equivalent to the new `v1` definitions, the only difference is they back-reference definitions in earlier package versions whereas the `v1` set is self-contained.

sergio-mena:
I went through the changes and they LGTM.
However, I see that files such as `abci/types/types.go`, `crypto/encoding/codec.go`, `internal/consensus/proto/aliases.go`, etc. are still using the `v1betaN` packages. Is this planned in a future PR?

mzabaluev:
> I see that files such as `abci/types/types.go`, `crypto/encoding/codec.go`, `internal/consensus/proto/aliases.go`, etc. are still using the `v1betaN` packages. Is this planned in a future PR?

No, I did not overhaul the manually written code at that time yet. Now it should be ready.

melekes (approved):

mzabaluev:
> Are these files actually needed?

Yes, these correspond to the protocol as it was released in 0.34, 0.37, and 0.38 branches.
I will add a table listing which version each beta-versioned package was introduced with.

> Also the I think the following files shouldn't be under v1betaN but they are referenced from other parts of the code:

Oops. Could you point exactly where (apart from generated code for other beta packages, of course)? I can't find anything with search.

> Finally, do we still need to generate the go code for the `v1betaN` proto files?

I think it can be useful for customers who want to build their own CometBFT code adhering to the specific pre-1.0 releases.
It does not cost us anything to have these files generated.

sergio-mena:
> Oops. Could you point exactly where (apart from generated code for other beta packages, of course)? I can't find anything with search.

So, I've repeated the operation on my IDE I did yesterday and didn't get anything this time. So I'd say, ignore my comment. Sorry for the noise.

sergio-mena (approved):
🙏

mzabaluev:
> I will add a table listing which version each beta-versioned package was introduced with.

This is done in #1694.

Issue #1675, opened 2023-11-22
Block Proposal Failure（proposal block parts exceeds maximum block bytes）


lx-xiang:
## Environment Information

- Cometbft Version: v0.37.2
- Cosmos Sdk Version: v0.47.3
- Cosmos Mempool Used: SenderNonceMempool
- Number of Validator: 3

## Err log
`ERR failed to process message err="total size of proposal block parts exceeds maximum block bytes (2097374 > 2097152)" height=2771 module=consensus msg_type=*consensus.BlockPartMessage peer=b9018b4260f19b5e0e8b0429c603d4156e6c2702 round=259`
`ERR failed to process message err="total size of proposal block parts exceeds maximum block bytes (2097354 > 2097152)" height=2771 module=consensus msg_type=*consensus.BlockPartMessage peer=1af9376873ae9bfa5cbb33a954f344afe96d55c6 round=260`

https://github.com/cometbft/cometbft/blob/v0.37.2/consensus/state.go#L1931

## Consensus Params
for all validators are configured as follows
`{"jsonrpc":"2.0","id":-1,"result":{"block_height":"2770","consensus_params":{"block":{"max_bytes":"2097152","max_gas":"-1"},"evidence":{"max_age_num_blocks":"100000","max_age_duration":"172800000000000","max_bytes":"1048576"},"validator":{"pub_key_types":["ed25519"]},"version":{"app":"0"}}}}`
`{"jsonrpc":"2.0","id":-1,"result":{"block_height":"2771","consensus_params":{"block":{"max_bytes":"2097152","max_gas":"-1"},"evidence":{"max_age_num_blocks":"100000","max_age_duration":"172800000000000","max_bytes":"1048576"},"validator":{"pub_key_types":["ed25519"]},"version":{"app":"0"}}}}`

andynog:
Hi @lx-xiang, thanks for reporting this. I believe the message in the log is just letting you know that the `max_bytes` configured might not be enough to accomodate the total size of proposal block parts, so doesn't seem to be a bug. But I'm not sure how you are testing/running this so I can't tell if it's just a matter of increasing that parameter value. 

Also, the `SenderNonceMempool` that you've mentioned in this issue, it is something implemented at the `cosmos-sdk` level. If you still think this is a bug, have you tried to open an issue in their repo?
https://github.com/cosmos/cosmos-sdk/blob/main/types/mempool/sender_nonce.go

lx-xiang:
[...]

PR #1687, opened 2023-11-24
fix: Txs Validate

lx-xiang:
The calculation method of tx size returned by calling proxyapp should be consistent with that of mempool

https://github.com/cometbft/cometbft/issues/1675
---

#### PR checklist

- [ ] Tests written/updated


melekes:
Hey @lx-xiang! Thanks for looking into this :+1:. Did the fix solve your problem? From what I can see, with this fix we'll return an error in https://github.com/cometbft/cometbft/blob/ecb35fbb04efb0ecfa6491242e2af2d31192f802/state/execution.go#L155 during proposal block creation. Is that a desired outcome?

lx-xiang:
@melekes The current fix does not solve my problem, but it will expose the issue in the Create ProposalBlock to avoid broadcasting invalid blocks，for cometbft, only this modification is needed

melekes:
@lx-xiang could you please add a changelog entry in `.changelog`?

lx-xiang:
I added

lasarojc:
May have hit this bug in a test without malicious nodes.
The receiving side got too big a block:  `msg_type=*consensus.BlockPartMessage err="total size of proposal block parts exceeds maximum block bytes (4195546 > 4194304)"`

Confirming that the problem is not reproduced with the fix (applied to an internal branch).

lasarojc (approved):
The test missing in this PR is added in #1712

CHANGELOG.md
```diff
@@ -47,6 +47,8 @@ gossip.
 
 - `[state/indexer]` Respect both height params while querying for events
    ([\#1529](https://github.com/cometbft/cometbft/pull/1529))
+- `[mempool]` The calculation method of tx size returned by calling proxyapp should be consistent with that of mempool
+  ([\#1687](https://github.com/cometbft/cometbft/pull/1687))
```
lasarojc, line 51:
```suggestion
```

lasarojc:
unclog changelog also added in #1712

lasarojc:
Thank you @lx-xiang . The fix is now merged into main and we'll back port it.

lasarojc:
@Mergifyio refresh

PR #1694, opened 2023-11-25
proto: document the cometbft.* versioned protos

mzabaluev:
Explain the new structure of the protobuf files and document the
mapping between version suffixes and CometBFT releases.

### DIFF
--- a/proto/README.md
+++ b/proto/README.md
@@ -1,7 +1,7 @@
 <!-- NB: Ensure that all hyperlinks in this doc are absolute URLs, not relative
 ones, as this doc gets published to the Buf registry and relative URLs will fail
 to resolve. -->
-# CometBFT v0.38.x Protocol Buffers Definitions
+# CometBFT Protocol Buffers Definitions
 
 This is the set of [Protobuf][protobuf] definitions of types used by various
 parts of [CometBFT]:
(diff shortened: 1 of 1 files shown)


proto/README.md
```diff
@@ -20,7 +20,36 @@ The canonical Protobuf definitions live in the `proto` folder of the relevant
...
+| `crypto`        | `v1`      | same      | same      |
+| `libs/bits`     | `v1`      | same      | same      |
+| `mempool`       | `v1`      | same      | same      |
+| `p2p`           | `v1`      | same      | same      |
+| `privval`       | `v1beta1` | same      | `v1`      |
+| `rpc/grpc`[^1]  | `v1beta1` | `v1beta2` | `v1beta3` |
+| `state`         | `v1beta1` | `v1beta2` | `v1beta3` |
+| `statesync`     | `v1`      | same      | same      |
+| `types`         | `v1beta1` | `v1beta2` | `v1`      |
+| `version`       | `v1`      | same      | same      |
```
thanethomson, line 48: This table's great, thanks for putting it together! What do the `same` entries mean though?

Also, could you please add a "1.0" column? My assumption is that every field value will be `v1`? Might be good to just have that there to be crystal clear to users as to what to expect.

proto/README.md
```diff
@@ -20,7 +20,36 @@ The canonical Protobuf definitions live in the `proto` folder of the relevant
...
+
+[^1]: Retired in 1.0
+
+## Why does CometBFT provide `tendermint` Protobuf definitions?
```
thanethomson, line 52: Please remove this section - I think it's already explained earlier in the readme.

proto/README.md
mzabaluev: "same" means the definitions have not changed in this release, so the same package as recorded to the left can be used.
Yes, in 1.0 everything is `v1` as the text explains, but I'll add the column as well.

proto/README.md
```diff
@@ -29,28 +58,9 @@ transitioning from Tendermint Core to CometBFT.
 As of CometBFT v1, however, the project will transition to using and providing a
 `cometbft` package of Protobuf definitions (see [\#1330]).
 
-## How are `tendermint` Protobuf definitions versioned?
```
andynog, line 32: Should we still keep links to the Buf registry ? I see there's a general link in the introduction at the top but maybe associating each version with the published version would be good.

proto/README.md
```diff
@@ -20,7 +20,36 @@ The canonical Protobuf definitions live in the `proto` folder of the relevant
...
+| `privval`       | `v1beta1` | same      | `v1`      | same |
+| `rpc/grpc`[^1]  | `v1beta1` | `v1beta2` | `v1beta3` |      |
+| `state`         | `v1beta1` | `v1beta2` | `v1beta3` | `v1` |
+| `statesync`     | `v1`      | same      | same      | same |
+| `types`         | `v1beta1` | `v1beta2` | `v1`      | same |
+| `version`       | `v1`      | same      | same      | same |
+
+[^1]: Retired in 1.0
+
+## Why does CometBFT provide `tendermint` Protobuf definitions?
```
andynog, line 52: Is the assumption that the packages were just renamed ? Like if someone was using the `tendermint.*` packages before, and want to upgrade, they would just need to rename to `cometbft.*` ? like they are 1:1 compatible (in structure)? If so, maybe a small blurb of text explaining (and assuring that) would be beneficial I'd think.

andynog (approved):
lgtm, added a few comments

proto/README.md
thanethomson: They were renamed _and_ versioned.

proto/README.md
mzabaluev: Restored the relevant text and the links in https://github.com/cometbft/cometbft/pull/1694/commits/c9fb60f915c9d486f979b5c734df622d38bf9123

melekes (approved):
I think this is super useful. Thanks for writing 👍

PR #1703, opened 2023-11-28
chore(proto): add missing docs

melekes:
and enable the same linting rules used by Cosmos SDK.

---

#### PR checklist


adizere (approved):
Seems like a good idea, any thoughts @mzabaluev ?

proto/buf.yaml
```diff
@@ -7,6 +7,16 @@ breaking:
...
   use:
     - DEFAULT
+    - COMMENTS
+    - FILE_LOWER_SNAKE_CASE
+  except:
+    - UNARY_RPC
+    - COMMENT_FIELD
+    - SERVICE_SUFFIX
+    - PACKAGE_VERSION_SUFFIX
+    - RPC_REQUEST_STANDARD_NAME
```
mzabaluev, line 17: Do these lints need to be overridden across the board, or only for legacy packages?
In the `ignore_only` object below, there is a pattern on how to only waive lints for the beta packages which did not follow best practices.

melekes: See https://github.com/cometbft/cometbft/pull/1703/commits/fe990e73983223f3c78db79a10666ba916668724

We only need to exclude `COMMENT_FIELD` **across the board**.

proto/buf.yaml
```diff
@@ -15,40 +15,33 @@ lint:
     - tendermint
   ignore_only:
     ENUM_VALUE_PREFIX:
-      - tendermint
```
mzabaluev, line 12: Why remove this? As long as the `buf.yaml` manifest does not exclude the `tendermint.*` packages, they will be linted by `buf lint`. You'd have to select the file set with `--path` every time to avoid the lints on `tendermint` protos.

proto/cometbft/abci/v1/types.proto
```diff
@@ -164,7 +179,7 @@ message ExtendVoteRequest {
   bytes proposer_address = 8;
 }
 
-// Request to verify a vote extension
+// Request to verify a vote extension.
```
thanethomson, line 182:
```suggestion
// VerifyVoteExtensionRequest is a request for the application to verify a vote extension
// produced by a different validator.
```

proto/cometbft/abci/v1/types.proto
```diff
@@ -275,87 +301,121 @@ message CheckTxResponse {
...
   PROCESS_PROPOSAL_STATUS_UNKNOWN = 0;
-  PROCESS_PROPOSAL_STATUS_ACCEPT  = 1;
-  PROCESS_PROPOSAL_STATUS_REJECT  = 2;
+  // Accepted
+  PROCESS_PROPOSAL_STATUS_ACCEPT = 1;
+  // Rejected
+  PROCESS_PROPOSAL_STATUS_REJECT = 2;
 }
 
+// ExtendVoteResponse contains the resulting bytes after extending the vote.
```
thanethomson, line 392:
```suggestion
// ExtendVoteResponse contains the vote extension that the application would like to
// attach to its next precommit vote.
```

proto/cometbft/abci/v1/types.proto
```diff
@@ -221,12 +239,15 @@ message ExceptionResponse {
   string error = 1;
 }
 
+// EchoResponse indicates that the connection is still alive.
 message EchoResponse {
   string message = 1;
 }
 
+// FlushResponse indicates that the ABCI application state was flushed?
```
thanethomson, line 247:
```suggestion
// FlushResponse indicates that the write buffer was flushed.
```

proto/buf.yaml
melekes: https://github.com/cometbft/cometbft/pull/1703/commits/7dc757b89e51ce1720662936890fb8a5b9a9b0ee

mzabaluev (approved):

PR #1709, opened 2023-11-29
feature/proto-upgrade: Last-minute fixes

thanethomson:
In reviewing #1672 I picked up on a few minor things:

1. If we don't include the `Makefile` change, Go will simply use whichever version of the Buf CLI is currently installed locally, which may not match the version used in CI (which is always the latest version).
2. CI should always check that the code generation has happened correctly.
3. I ran `make proto-gen` to update the generated code.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/.github/workflows/check-generated.yml
+++ b/.github/workflows/check-generated.yml
@@ -6,8 +6,6 @@
 name: Check generated code
 on:
   pull_request:
-    branches:
-      - main
   merge_group:
 
 permissions:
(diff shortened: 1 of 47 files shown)


melekes (approved):
👍

mzabaluev:
The command line does not follow `go.sum` even if it's run in the repository's checked out directory, right?

mzabaluev (approved):

PR #1726, opened 2023-12-01
buf improvements prior to publishing on BSR

mzabaluev:
Update buf configuration files to:

- Update the gogo-proto dependency to the latest revision;
- Exclude the `tendermint` files from any consideration by buf.


thanethomson (approved):

melekes (approved):
👍

Issue #1729, opened 2023-12-03
ABCI method InitChain fails if genesis file is larger than 100MB and ABCI app is started in socket mode


troykessler:
### Setup

**CometBFT version**: tendermint v0.34.9 - cometbft v0.38.2

**Have you tried the latest version**: yes

**ABCI app**: Gaia, Evmos, Archway

### What happened?

During the development of [KSYNC](https://github.com/KYVENetwork/ksync), a tool to sync blocks stored on Arweave to any tendermint/cometbft based app chain, an error occured on the app every time the ABCI method [InitChain](https://github.com/cometbft/cometbft/blob/main/spec/abci/abci++_methods.md#initchain) was called over the [socket protocol](https://github.com/cometbft/cometbft/blob/main/spec/abci/abci++_client_server.md#socket) (error log below).

### What did you expect to happen?

That the ABCI method InitChain can be called successfully called over the socket protocol by Tendermint/CometBFT without the app crashing.

### How to reproduce it

It can be reproduced by trying to start the Gaia chain (Cosmos Hub) from genesis not in the default embedded in-process mode, but with tendermint running externally over the socket protocol.

1. Install gaia with latest genesis version `v4.2.1`: https://github.com/cosmos/gaia/releases/tag/v4.2.1
2. Init with `./gaiad init test --chain-id cosmoshub-4`
3. Install gaia [genesis file](https://raw.githubusercontent.com/cosmos/mainnet/master/genesis/genesis.cosmoshub-4.json.gz)
4. Start gaia without embedded tendermint process: `./gaiad start --with-tendermint=false --transport=socket`

(open new terminal)

5. Install tendermint `v0.34.9`, which was used for gaia `v4.2.1`: https://github.com/tendermint/tendermint/releases/tag/v0.34.9
6. Start tendermint: `./tendermint node --home ~/.gaia`

### Logs

Gaia:

```
4:51PM INF Waiting for new connection... module=abci-server
4:51PM INF Accepted a new connection module=abci-server
4:51PM INF Waiting for new connection... module=abci-server
4:51PM ERR Connection error err="error reading message: short buffer" module=abci-server
4:51PM ERR Connection was closed by client module=abci-server
4:51PM ERR Connection was closed by client module=abci-server
4:51PM ERR Connection was closed by client module=abci-server
```

Tendermint:

```
I[2023-12-03|16:51:52.461] ABCI Handshake App Info                      module=consensus height=0 hash= software-version= protocol-version=0
[...]

PR #1730, opened 2023-12-03
fix: increase abci socket message size limit to 2GB

troykessler:
This PR addresses #1729 by increasing the maximum message size for the ABCI socket protocol to 2GB. The current maximum message size of 100MB caused InitChain to fail if the genesis file is bigger than 100MB because the app state is  a parameter in this ABCI method (see [here](https://github.com/cometbft/cometbft/blob/main/spec/abci/abci++_methods.md#initchain)). The new message limit should cover all existing genesis files while still being a protection against large spam messages.

---

#### PR checklist

- [ ] Tests written/updated


.changelog/unreleased/improvements/1730-increase-abci-socket-message-size-limit
```diff
@@ -0,0 +1 @@
+- `[fix]` Increase ABCI socket message size limit to 2GB ([\#1730](https://github.com/cometbft/cometbft/pull/1730))
```
melekes, line 1:
```suggestion
- `[abci]` Increase ABCI socket message size limit to 2GB ([\#1730](https://github.com/cometbft/cometbft/pull/1730))
```

troykessler:
Changed `maxMsgSize` from `2147483648` to `2147483647` (MaxInt32) to allow successful builds on 32-bit systems where a message size of 2GB (MaxInt32 + 1) resulted in the following error: `cannot use maxMsgSize (untyped int constant 2147483648) as int value in argument to protoio.NewDelimitedReader (overflows)`

.changelog/unreleased/improvements/1730-increase-abci-socket-message-size-limit
```diff
@@ -0,0 +1 @@
+- `[abci]` Increase ABCI socket message size limit to 2GB ([\#1730](https://github.com/cometbft/cometbft/pull/1730))
```
thanethomson, line 1:
```suggestion
- `[abci]` Increase ABCI socket message size limit to 2GB ([\#1730](https://github.com/cometbft/cometbft/pull/1730): @troykessler)
```

thanethomson (approved):
Thanks @troykessler!

PR #1750, opened 2023-12-05
`VerifyCommitLight` and `VerifyCommitLightTrusting` _never_ check all signatures

### LINKED ISSUE #1749: Bogus signatures in LightClientAttackEvidence
### Bug Report (low severity)
At a high level, a malicious validator with more than 1/3 of the voting power could generate `LightClientAttackEvidence` that contains _correct_ signatures but also contains _bogus_ (e.g., unverifiable) signatures. Such `LightClientAttackEvidence` can be added to the _evidence pool_ without an issue if the correct signatures stem from validators that have more than 1/3 of the voting power. As a result when we go through the signatures to slash validators we could go through bogus signatures as well and potentially slash innocent validators.
 
Specifically, the issue has to do in the way that [`LightClientAttackEvidence`](https://github.com/cometbft/cometbft/blob/v0.38.2/proto/tendermint/types/evidence.proto#L28) is verified before being added to the evidence pool.  [`VerifyLightClientAttack`](https://github.com/cometbft/cometbft/blob/v0.38.2/evidence/verify.go#L110) in case of a lunatic attack calls [`VerifyCommitLightTrusting`](https://github.com/cometbft/cometbft/blob/v0.38.2/types/validation.go#L94) with `e.ConflictingBlock.Commit`. Subsequently, `VerifyCommitLightTrusting` calls [`verifyCommitBatch`](https://github.com/cometbft/cometbft/blob/v0.38.2/types/validation.go#L152) or [`verifyCommitSingle`](https://github.com/cometbft/cometbft/blob/v0.38.2/types/validation.go#L265) with the `countAllSignatures` argument  set to `false`. This means that both `verifyCommitBatch` and `verifyCommitSingle` do not go through _all_ the signatures contained in `e.ConflictingBlock.Commit.Signatures` if the signatures that have been verified so far exceed 1/3 (i.e., the [default trust level](https://github.com/cometbft/cometbft/blob/v0.38.2/light/verifier.go#L16)) of the voting power we stop checking when the following condition evaluates to true:
```
if !countAllSignatures && talliedVotingPower > votingPowerNeeded {
```
As a result, a malicious validator (or set of validators) that has more than 1/3 of the voting power could generate `LightClientAttackEvidence` that has all the verifiable signatures at the beginning of the `conflictingBlock.Commit.Signatures` slice and afterwards it has bogus signatures that are not verifiable.
Nevertheless, in such a scenario, the `LightClientAttackEvidence` with the bogus signatures would still be [added](https://github.com/cometbft/cometbft/blob/v0.38.2/evidence/pool.go#L158) in the evidence pool.

When we slash based on this `LightClientAttackEvidence` we [extract the byzantine validators](https://github.com/cometbft/cometbft/blob/v0.38.2/types/evidence.go#L257) that would go through all the signatures:
```
for _, commitSig := range l.ConflictingBlock.Commit.Signatures {
    ...
    _, val := commonVals.GetByAddress(commitSig.ValidatorAddress)
```
And hence we could slash validators that have their `ValidatorAddress` in some of the signatures but those validators have not performed a light client attack. This way, a malicious validator would get slashed, but can also succeed in slashing innocent validators.

**CometBFT version:** Note that the above provided links stem from CometBFT v0.38.2 but I would assume the issue exists since the introduction of `LightClientAttackEvidence`.

Thanks to @jmalicevic and @sergio-mena for looking into this.

sergio-mena:
Closes #1749

Please see the description of #1749 for a thorough description of the problem.

This PR consists of two commits (initially):

* The first commit are changes in our `e2e` tests that allow for a controlled end-to-end reproduction of the problem, and extra logic in the `e2e` logic to properly catch it as test failure. Check the little ❌ next to the `e2e` CI run on this commit to see the error.
* The second commit is the fix on production code. Although the initially proposed fix was to set parameter `countAllSignatures` to true in _all_ calls to `verifyCommitBatch` and `verifyCommitSingle`, it would actually be overkill
  * the light client algorithm does not need to verify all signatures, as it doesn't produce or check any evidence
  * so the proposed fix is rather setting that parameter to true whenever a light client attack evidence is being checked, but to false when it is being used by the light client.
  * I also adapt some of the UTs testing this functionality to also check that VerifyCommitLight` and `VerifyCommitLightTrusting` are also working as expected when called with `countAllSignatures` set to to true
  * Check the little ✅ next to the `e2e` CI run on this commit to see that the error is now gone (i.e., malformed evidences are rejected, and correctly constructed evidences are accepted)

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/test/e2e/runner/evidence.go
+++ b/test/e2e/runner/evidence.go
@@ -24,7 +24,7 @@ import (
 // 1 in 4 evidence is light client evidence, the rest is duplicate vote evidence
 const lightClientEvidenceRatio = 4
 
-// InjectEvidence takes a running testnet and generates an amount of valid
+// InjectEvidence takes a running testnet and generates an amount of valid/invalid
 // evidence and broadcasts it to a random node through the rpc endpoint `/broadcast_evidence`.
 // Evidence is random and can be a mixture of LightClientAttackEvidence and
 // DuplicateVoteEvidence.
@@ -87,10 +87,12 @@ func InjectEvidence(ctx context.Context, r *rand.Rand, testnet *e2e.Testnet, amo
 	}
 
 	var ev types.Evidence
-	for i := 1; i <= amount; i++ {
+	for i := 0; i < amount; i++ {
+		validEv := true
 		if i%lightClientEvidenceRatio == 0 {
+			validEv = i%(lightClientEvidenceRatio*2) != 0 // Alternate valid and invalid evidence
 			ev, err = generateLightClientAttackEvidence(
-				ctx, privVals, evidenceHeight, valSet, testnet.Name, blockRes.Block.Time,
+				ctx, privVals, evidenceHeight, valSet, testnet.Name, blockRes.Block.Time, validEv,
 			)
 		} else {
 			var dve *types.DuplicateVoteEvidence
@@ -110,7 +112,15 @@ func InjectEvidence(ctx context.Context, r *rand.Rand, testnet *e2e.Testnet, amo
 		}
 
 		_, err := client.BroadcastEvidence(ctx, ev)
-		if err != nil {
+		if !validEv {
+			// The tests will count committed evidences later on,
+			// and only valid evidences will make it
+			amount++
+		}
+		if validEv != (err == nil) {
+			if err == nil {
+				return errors.New("submitting invalid evidence didn't return an error")
+			}
 			return err
 		}
 	}
@@ -155,6 +165,7 @@ func generateLightClientAttackEvidence(
 	vals *types.ValidatorSet,
 	chainID string,
 	evTime time.Time,
+	validEvidence bool,
 ) (*types.LightClientAttackEvidence, error) {
 	// forge a random header
 	forgedHeight := height + 2
@@ -164,7 +175,7 @@ func generateLightClientAttackEvidence(
 
 	// add a new bogus validator and remove an existing one to
 	// vary the validator set slightly
-	pv, conflictingVals, err := mutateValidatorSet(ctx, privVals, vals)
+	pv, conflictingVals, err := mutateValidatorSet(ctx, privVals, vals, !validEvidence)
 	if err != nil {
 		return nil, err
 	}
@@ -177,6 +188,10 @@ func generateLightClientAttackEvidence(
 	commit, err := test.MakeCommitFromVoteSet(blockID, voteSet, pv, forgedTime)
 	if err != nil {
 		return nil, err
+	}
+
+	if !validEvidence {
+		commit.Signatures[len(commit.Signatures)-1].Signature[0]++
 	}
 
 	ev := &types.LightClientAttackEvidence{
@@ -292,7 +307,11 @@ func makeBlockID(hash []byte, partSetSize uint32, partSetHash []byte) types.Bloc
 	}
 }
 
-func mutateValidatorSet(ctx context.Context, privVals []types.MockPV, vals *types.ValidatorSet,
+func mutateValidatorSet(
+	ctx context.Context,
+	privVals []types.MockPV,
+	vals *types.ValidatorSet,
+	nop bool,
 ) ([]types.PrivValidator, *types.ValidatorSet, error) {
 	newVal, newPrivVal, err := test.Validator(ctx, 10)
 	if err != nil {
@@ -300,10 +319,14 @@ func mutateValidatorSet(ctx context.Context, privVals []types.MockPV, vals *type
 	}
 
 	var newVals *types.ValidatorSet
-	if vals.Size() > 2 {
-		newVals = types.NewValidatorSet(append(vals.Copy().Validators[:vals.Size()-1], newVal))
+	if nop {
+		newVals = types.NewValidatorSet(vals.Copy().Validators)
 	} else {
-		newVals = types.NewValidatorSet(append(vals.Copy().Validators, newVal))
+		if vals.Size() > 2 {
+			newVals = types.NewValidatorSet(append(vals.Copy().Validators[:vals.Size()-1], newVal))
+		} else {
+			newVals = types.NewValidatorSet(append(vals.Copy().Validators, newVal))
+		}
 	}
 
 	// we need to sort the priv validators with the same index as the validator set
--- a/internal/evidence/pool.go
+++ b/internal/evidence/pool.go
@@ -133,11 +133,11 @@ func (evpool *Pool) Update(state sm.State, ev types.EvidenceList) {
 
 // AddEvidence checks the evidence is valid and adds it to the pool.
 func (evpool *Pool) AddEvidence(ev types.Evidence) error {
-	evpool.logger.Debug("Attempting to add evidence", "ev", ev)
+	evpool.logger.Info("Attempting to add evidence", "ev", ev)
 
 	// We have already verified this piece of evidence - no need to do it again
 	if evpool.isPending(ev) {
-		evpool.logger.Debug("Evidence already pending, ignoring this one", "ev", ev)
+		evpool.logger.Info("Evidence already pending, ignoring this one", "ev", ev)
 		return nil
 	}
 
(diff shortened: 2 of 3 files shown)


types/validation.go
```diff
@@ -57,8 +57,14 @@ func VerifyCommit(chainID string, vals *ValidatorSet, blockID BlockID,
...
 // signatures.
-func VerifyCommitLight(chainID string, vals *ValidatorSet, blockID BlockID,
-	height int64, commit *Commit) error {
+func VerifyCommitLight(
+	chainID string,
+	vals *ValidatorSet,
+	blockID BlockID,
+	height int64,
+	commit *Commit,
+	countAllSignatures bool,
```
thanethomson, line 66: Why add this as an optional parameter to the method? Why not either introduce a new method that checks all signatures, or just make checking all signatures the default behaviour?

Boolean fields like this are, to me, generally indicative of a need for a new function as opposed to simply adding more configuration parameters and complexity to an existing one.

types/validation.go
```diff
@@ -92,7 +98,13 @@ func VerifyCommitLight(chainID string, vals *ValidatorSet, blockID BlockID,
 //
 // This method is primarily used by the light client and does not check all the
 // signatures.
-func VerifyCommitLightTrusting(chainID string, vals *ValidatorSet, commit *Commit, trustLevel cmtmath.Fraction) error {
+func VerifyCommitLightTrusting(
+	chainID string,
+	vals *ValidatorSet,
+	commit *Commit,
+	trustLevel cmtmath.Fraction,
+	countAllSignatures bool,
```
thanethomson, line 106: Same here.

types/validation.go
```diff
@@ -57,8 +57,14 @@ func VerifyCommit(chainID string, vals *ValidatorSet, blockID BlockID,
 //
 // This method is primarily used by the light client and does not check all the
 // signatures.
```
insumity, line 59: Remove "does not check all the signatures." because this now depends on `countAllSignatures`. Similarly for `VerifyCommitLightTrusting` below.

types/validation_test.go
```diff
@@ -184,8 +195,10 @@ func TestValidatorSet_VerifyCommitLight_ReturnsAsSoonAsMajorityOfVotingPowerSign
...
 	commit.Signatures[3] = vote.CommitSig()
 
-	err = valSet.VerifyCommitLight(chainID, blockID, h, commit)
+	err = valSet.VerifyCommitLight(chainID, blockID, h, commit, false)
 	assert.NoError(t, err)
+	err = valSet.VerifyCommitLight(chainID, blockID, h, commit, true)
+	assert.Error(t, err) // counting all signatures detects the malleated signature
 }
 
 func TestValidatorSet_VerifyCommitLightTrusting_ReturnsAsSoonAsTrustLevelOfVotingPowerSigned(t *testing.T) {
```
insumity, line 204: Maybe rename the test here because now it contains a test case where we do not `ReturnAsSoonAsTrustLevelOfVotingPowerSigned`.

test/e2e/runner/evidence.go
```diff
@@ -164,7 +175,7 @@ func generateLightClientAttackEvidence(
 
 	// add a new bogus validator and remove an existing one to
 	// vary the validator set slightly
-	pv, conflictingVals, err := mutateValidatorSet(ctx, privVals, vals)
+	pv, conflictingVals, err := mutateValidatorSet(ctx, privVals, vals, !validEvidence)
```
insumity, line 178: If the `nop` parameter in `mutateValidatorSet` is a no-op, then maybe it would make sense to keep `mutateValidatorSet` as is and do (?)
```
if validEvidence {
    pv, conflictingVals, err := mutateValidatorSet(ctx, privVals, vals)
} else {
   // reuse privVals and vals
}
```

test/e2e/runner/evidence.go
```diff
@@ -179,6 +190,10 @@ func generateLightClientAttackEvidence(
 		return nil, err
 	}
 
+	if !validEvidence {
```
insumity, line 193: nit: could potentially add a comment in the style of "malleate the last signature of the commit by adding one to its first byte"

insumity:
Thanks Sergio!
I left some minor comments but looks really good to me.

Another idea: I wonder if, for simplicity, it would make sense to leave `VerifyCommitLight` and `VerifyCommitLightTrusting` as is and just change [`GetByzantineValidators`](https://github.com/cometbft/cometbft/blob/v0.38.2/types/evidence.go#L252). `GetByzantineValidators` could be the one that verifies the signatures and return an `err`or if one of the signatures cannot be verified. Then [`ValidateABCIEvidence`](https://github.com/cometbft/cometbft/blob/v0.38.2/evidence/verify.go#L231) called from inside [`VerifyLightClientAttack`](https://github.com/cometbft/cometbft/blob/v0.38.2/evidence/verify.go#L110) would fail and hence the evidence won't be added in the evidence pool. This would be similar to what Simon did in [interchain-security](https://github.com/cosmos/interchain-security/blob/ac0597fced07ff51f7f272aa92f440d57d4d4548/x/ccv/provider/keeper/consumer_equivocation.go#L198) where we verify the signatures when trying to extract the byzantine validators.

test/e2e/runner/evidence.go
sergio-mena: 👍

sergio-mena:
@insumity 
> Another idea

Looks like a good alternative, but I'll stick with the present solution if you don't mind, as we spent some time testing it

types/validation_test.go
```diff
@@ -186,9 +205,11 @@ func TestValidatorSet_VerifyCommitLight_ReturnsAsSoonAsMajorityOfVotingPowerSign
 
 	err = valSet.VerifyCommitLight(chainID, blockID, h, commit)
 	assert.NoError(t, err)
+	err = valSet.VerifyCommitLightAllSignatures(chainID, blockID, h, commit)
```
insumity, line 208: nit: Maybe rename `[...]_ReturnsAsSoonAsMajorityOfVotingPowerSigned` as well because it checks something else as well.

insumity (approved):
👍 LGTM

thanethomson (approved):
Great work folks! :tada:

types/validation_test.go
sergio-mena: Good catch

thanethomson:
Should we not also backport this to v0.37.x and v0.34.x?

melekes (approved):
👍

internal/blocksync/reactor.go
```diff
@@ -460,7 +460,7 @@ FOR_LOOP:
 			// first.Hash() doesn't verify the tx contents, so MakePartSet() is
 			// currently necessary.
 			// TODO(sergio): Should we also validate against the extended commit?
-			err = state.Validators.VerifyCommitLight(
+			err = state.Validators.VerifyCommitLightAllSignatures(
```
ValarDragon, line 463: (Being pedantic) I actually don't think this is right. This line is using the received second block's commit, whose voteset isn't finalized.

Agreed that every other usage is correct (and important to fix)!

So it should suffice for correctness here that 2/3rds of signatures are valid.

However for reasons noted in #1830 , I think the new change is better, and we refactor the logic to have the next two calls not verify these signatures.

sergio-mena: @ValarDragon Not pedantic at all. Spot on, I'd say.

The reason why we are using `VerifyCommitLightAllSignatures` here is that we decided to leave the previous behavior (i.e., skipping signatures) only in the light client logic, for the rest we took a conservative approach (bear in mind that the raison d'être of this PR was that skipping signatures was problematic when dealing with evidence verification).

However, after careful consideration of your proposal, we realize that the usage of `VerifyCommitLight*` in blocksync can be seen as a light client-ish verification as well, in that it does not produce or verify any evidence of misbehavior. So we'll open a PR to change this shortly. Bear with us.

sergio-mena: #1858

PR #1755, opened 2023-12-06
perf(store): Batch save `Block`

melekes:
Refs https://github.com/cometbft/cometbft/issues/1040

LSMs don't handle big batches very well ([1](https://github.com/cosmos/iavl/issues/619)). The ideal for `goleveldb` is ~ 100KB ([2](https://github.com/ethereum/go-ethereum/pull/15115)), which is more or less equal to our block part size (64kB).

Here are the results from running the Anton's benchmark on my local PC:

<details>

<summary>Benchmark</summary>

```
func BenchmarkBlockStore_SaveBlockWithExtendedCommit(b *testing.B) {
	dbTypes := []dbm.BackendType{
		dbm.GoLevelDBBackend,
	}
	NTXs := int64(1024)

	for _, dbType := range dbTypes {
		state, bs, cleanup := makeStateAndBlockStoreSpecificBackend(dbType)
		b.Run(fmt.Sprintf("DBType:%s;NTxs:%d", dbType, NTXs),
			func(b *testing.B) {
				for i := 0; i < b.N; i++ {
					height := bs.Height() + 1
					block := state.MakeBlock(height, test.MakeNTxs(height, NTXs), new(types.Commit), nil, state.Validators.GetProposer().Address)
					validPartSet, err := block.MakePartSet(types.BlockPartSizeBytes)
					if err != nil {
						b.Error(err)
					}
					seenCommit := makeTestExtCommit(block.Header.Height, cmttime.Now())

					bs.SaveBlockWithExtendedCommit(block, validPartSet, seenCommit)
				}
			})

		cleanup()
	}

}

type cleanupFunc func()

func makeStateAndBlockStoreSpecificBackend(backendType dbm.BackendType) (sm.State, *BlockStore, cleanupFunc) {
	config := test.ResetTestRoot("blockchain_reactor_test")
	blockDB, err := dbm.NewDB("block", backendType, config.DBDir())
	if err != nil {
		panic(fmt.Errorf("error creating %s: %w", backendType, err))
	}
	stateDB, err := dbm.NewDB("state", backendType, config.DBDir())
	if err != nil {
		panic(fmt.Errorf("error creating %s: %w", backendType, err))
	}
		DiscardABCIResponses: false,
	})
	state, err := stateStore.LoadFromDBOrGenesisFile(config.GenesisFile())
	if err != nil {
		panic(fmt.Errorf("error constructing state from genesis file: %w", err))
	}
	return state, NewBlockStore(blockDB), func() { os.RemoveAll(config.RootDir) }
}
```

</details>

```
## batching / 1 MB block / goleveldb
155921071 ns/op
## NO batching / 1 MB block / goleveldb
134169606 ns/op
## batching / 10 MB block / goleveldb
1135896143 ns/op
## NO batching / 10 MB block / goleveldb
1096632817 ns/op
```

If we want to support blocks bigger than 100kB on average, we should not batch single blocks. This becomes more important for big blocks (>1MB). The current max size is 4MB. This implies that CometBFT must handle such a scenario efficiently, which won't be the case w/ a single batch.

Now, the current block size of Cosmos Hub is 64kB, with peaks ranging from 200kB to 700kB. For Osmosis, it's 74kB. If we want to target the current customers, then creating a single batch makes sense.

The solution presented here is to batch save the whole block only if it's within 640kB (10 block parts * 64kB each). Otherwise, save each block part individually. All the commits are still saved as part of the batch.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -324,7 +324,7 @@ func (bs *BlockStore) PruneBlocks(height int64, state sm.State) (uint64, int64, 
 		bs.mtx.Lock()
 		bs.base = base
 		bs.mtx.Unlock()
-		bs.saveState()
+		bs.saveState(batch)
 
 		err := batch.WriteSync()
 		if err != nil {
@@ -403,12 +403,26 @@ func (bs *BlockStore) SaveBlock(block *types.Block, blockParts *types.PartSet, s
 	if block == nil {
 		panic("BlockStore can only save a non-nil block")
 	}
-	if err := bs.saveBlockToBatch(block, blockParts, seenCommit); err != nil {
+
+	batch := bs.db.NewBatch()
+	defer func(batch dbm.Batch) {
+		err := batch.Close()
+		if err != nil {
+			panic(err)
+		}
+	}(batch)
+
+	if err := bs.saveBlockToBatch(block, blockParts, seenCommit, batch); err != nil {
 		panic(err)
 	}
 
 	// Save new BlockStoreState descriptor. This also flushes the database.
-	bs.saveState()
+	bs.saveState(batch)
+
+	err := batch.WriteSync()
+	if err != nil {
+		panic(err)
+	}
 }
 
 // SaveBlockWithExtendedCommit persists the given block, blockParts, and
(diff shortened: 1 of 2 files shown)


.changelog/unreleased/improvements/1755-batch-save-block.md
```diff
@@ -0,0 +1,2 @@
+- `[store]` Save block using a single DB batch ([\#1755](https://github.com/cometbft/cometbft/pull/1755))
+  Only if block is less than 640kB, otherwise each block part is saved individually.
```
thanethomson, line 2:
```suggestion
- `[store]` Save block using a single DB batch if block is less than 640kB, otherwise each block part is saved individually
  ([\#1755](https://github.com/cometbft/cometbft/pull/1755))
```

thanethomson (approved):
LGTM!

internal/store/store.go
```diff
@@ -324,7 +330,7 @@ func (bs *BlockStore) PruneBlocks(height int64, state sm.State) (uint64, int64,
 		bs.mtx.Lock()
 		bs.base = base
 		bs.mtx.Unlock()
-		bs.saveState()
+		bs.saveState(batch)
```
lasarojc, line 333: Could we rename this variable so it does not shadow the outer one?

lasarojc: Not related to this PR, but this function closes the batch, but this function is called on a batch for which `defer Close()` has also been called. The second close should be a nop, but it would be better to clearly specify who closes it.

lasarojc: If the close inside the function remains, then it should be a deferred call.

lasarojc (approved):

lasarojc:
Some coments not really related to this PR and a non-required suggestion to change the code. Otherwise, LGTM

Issue #1041, opened 2023-06-27
Reconsider representation of keys for state and block store


jmalicevic:
To save state and block data, CometBFT generates keys with string + height prefixes. They are at the moment internally sorted lexicographically. 

This potentiall impacts both data access time (lookup) as well as the performance of compaction. 

Some work in Q2 2023 indicates that changing the way keys are represented and forcing them to be ordered by height improved compaction and the disk footprint of the blockstore (28MB shrank to 18MB). This was however a POC and the generated data size was probably not big enough to be certain of gains. 

Also, it is not 100% clear whether sorting all types of data by height is desired. Talking to users and analyzing the access patterns should help with that. 

We need to: 
* [x] Understand which data is access together to make sure the keys are in a layout that is beneficial for common access patterns. For example, we often fetch the last commits of a previous height and those should be located nearby. 
* [ ] https://github.com/cometbft/cometbft/pull/2327 [ ] #1041 (Adding support for a different key ordering. Previous work done on this topic for Tendermint 0.35 and 0.36:
https://github.com/tendermint/tendermint/pull/5771 . It is ported into this work. )

This would potentially resolve or improve https://github.com/informalsystems/interchain/issues/1 

DoD

* [x] Understand the way data is layout by the database
* [x] Decide on optimal key representation to suit our use cases
* [ ] Benchmark the code with a different layout against the baseline in #67 
* [ ] Evaluate the impact on pruning and compaction
* [ ] Update documentation
* [ ] Write tests to confirm behaviour is correct
* [ ] Provide users with a tool to migrate to the new data layout in case one is chosen. 
* [ ] Remove `mtx, base, height` fields from the `BlockStore` struct https://github.com/cometbft/cometbft/pull/1856/files#r1430831931

sergio-mena:
The work done regarding this in `v0.35.x`/`v0.36.x` should also be considered

melekes:
<details>

<summary>vanilla 4 node setup:</summary>

```
#key: blockStore
#key: genesisDocHash
#key: stateKey
#key: stateKey
#key: stateKey
#key: AppRetainHeightKey
#key: offlineStateSyncHeightKey
#key: AppRetainHeightKey
#key: validatorsKey:1
#key: validatorsKey:1
#key: H:1
#key: validatorsKey:2
#key: validatorsKey:1
#key: validatorsKey:2
#key: validatorsKey:1
#key: validatorsKey:2
#key: validatorsKey:1
#key: H:2
[...]

PR #1764, opened 2023-12-07
feat(store)!: order-preserving varint key encoding

melekes:
uses the [orderedcode](https://github.com/google/orderedcode) library to lexicographically encode the key's in the state store, block store, evidence pool and light store.

Port of https://github.com/tendermint/tendermint/pull/5771

Closes https://github.com/cometbft/cometbft/issues/1041

---

#### PR checklist

- [ ] Tests written/updated


melekes:
The migration script needs to be copied and adapted from https://github.com/cometbft/cometbft/pull/1814.

mzabaluev:
Would this include a fix for https://github.com/cometbft/cometbft/issues/1857?

jmalicevic:
> Would this include a fix for #1857?

That is a separate concern I believe.

light/store/db/db.go
```diff
@@ -169,13 +171,8 @@ func (s *dbs) FirstLightBlockHeight() (int64, error) {
...
 	defer itr.Close()
 
-	for itr.Valid() {
-		key := itr.Key()
-		_, height, ok := parseLbKey(key)
-		if ok {
-			return height, nil
-		}
-		itr.Next()
+	if itr.Valid() {
```
jmalicevic, line 174: Implementing this for the other PR. This should return the height if there is no error. Otherwise, there might be a reason for the error where we should keep iterating. In the new layout I think it is these two conditions

```	if lightBlockPrefix != prefixLightBlock {
		err = fmt.Errorf("expected light block prefix but got: %d", lightBlockPrefix)
	}
	if dbPrefix != storePrefix {
		err = fmt.Errorf("parsed key has a different prefix. Expected: %s, got: %s", storePrefix, dbPrefix)
	}
	
	
``` 

And in the old one it is the first condition when there is no submatch.

PR #1783, opened 2023-12-11
fix(consensus): return last saved BeginBlock, not a empty one

### LINKED ISSUE #1468: replay missing begin block

melekes:
Otherwise, the events from app's BeginBlock won't be fired.

Closes #1468

---

#### PR checklist

- [ ] Tests written/updated


thanethomson (approved):

sergio-mena:
This is awesome!

PR #1798, opened 2023-12-11
chore(docs): explain the effect of `timeout_propose`

### LINKED ISSUE #1586: About `some block time` in config.toml will affect the selection of block producer?
Hello, friends from the community. I conducted an experiment on a chain launched with Tendemint and Cosmos SDK and had some questions about the selection of block producers. Here's my experiment(using cosmos-sdk v0.47.3, comebft v0.37.2):

- I launched a chain using Cosmos SDK.
- started two validator nodes with equal stake amounts. 

I observed that both nodes took turns to produce blocks, which aligns with the principles of Tendermint's round-robin consensus.
Now, let's make some configuration changes on top of the default settings mentioned above:
- The first node uses the default consensus configuration with:
image

while the second node modifies the consensus field in the config.yaml with this: 
image

At this point, I noticed that the round-robin consensus rule wasn't implemented, as `the second node continued to produce blocks while the first node didn't get the chance`. 
image

**Why is this happening? Can block production time be modified by the validator to gain an advantage in block production? Does this deviate from the definition of round-robin consensus?**

Since the first validator didn't produce blocks, I attempted to increase the delegation amount for the first validator, making Validator1's voting power 60% and Validator2's voting power 40%. Surprisingly, Validator1 also started producing blocks.

- **Why is this?** Isn't block production time determined by voting power? Why is it influenced by the node's block production time configuration? 
- If it's the case, how can we address situations where someone arbitrarily adjusts their block production time to gain an advantage, especially since it seems like these configurations are entirely under the control of the nodes themselves?

melekes:
on the proposer selection process.

Closes #1586

---

#### PR checklist


docs/core/configuration.md
```diff
@@ -604,3 +604,54 @@ Here's a brief summary of the timeouts:
...
+...
+```
+
+What happens in reality is, however, a little bit different:
+
+```
+#1 block - Bob
+#2 block - Bob
+#3 block - Bob
+#1 block - Alice
```
hvanz, line 634:
```suggestion
#4 block - Alice
```

docs/core/configuration.md
```diff
@@ -604,3 +604,54 @@ Here's a brief summary of the timeouts:
...
+#3 block - Bob
+#4 block - Alice
+```
+
+That's because Bob is too fast at proposing blocks. This leaves Alice very
+little chances to propose a block and not always be catching up. Note every
+block Bob creates needs a vote from Alice to constitute 2/3+.
+
+Imagine now there are ten geographically distributed validators. One of them
+(Bob) sets `timeout_propose` to 1s. Now, Bob won't be able to move with the
```
andynog, line 642: do we need to add that the other validators set the timeout_propose to a value of something like 3s or higher just to make it more clear?

docs/core/configuration.md
```diff
@@ -604,3 +604,54 @@ Here's a brief summary of the timeouts:
 - `timeout_commit` = how long a validator should wait after committing a block, before starting
   on the new height (this gives us a chance to receive some more precommits,
   even though we already have +2/3)
+
+### The effect of `timeout_propose` on the proposer selection process
```
andynog, line 608: this is good, I think eventually we might also have to explain other timeouts (prevote, precommit, commit) and how they related to each other.

andynog (approved):
lgtm, small suggestion

docs/core/configuration.md
sergio-mena, line 608: Consider changing this to something like:
"The adverse effect of using inconsistent `timeout_commit` valued in a network"

docs/core/configuration.md
```diff
@@ -604,3 +604,54 @@ Here's a brief summary of the timeouts:
...
+
+Here's an interesting question. What if the particular validator sets a very
+small `timeout_propose`?
```
sergio-mena, line 611: Consider rephrasing the question as:

"What happens if a particular validator sets a very small `timeout_commit` (as compared to the rest of the network?"

sergio-mena:
After convo with @melekes , we decided to slightly change the text. 🙏 @melekes

Issue #1622, opened 2023-11-15
Spec/Implementation Discrepancy: Which fields of ABCI must be deterministic?


p-offtermatt:
In the specification, for ABCI, it seems that events need to be deterministic:

See here:
https://github.com/cometbft/cometbft/blob/main/spec/abci/abci%2B%2B_basic_concepts.md#deterministic-state-machine-replication

Quite

>     Note that some methods (Query, FinalizeBlock) return non-deterministic data in the form of Info and Log fields. The Log is intended for the literal output from the Application's logger, while the Info is any additional info that should be returned. These are the only fields that are not included in block header computations, so we don't need agreement on them. All other fields in the Response* must be strictly deterministic.

Emphasis on this:

>     All other fields in the Response* must be strictly deterministic.

Additional context:
Events are returned by FinalizeBlock (or EndBlock under ABCI-without-++)
https://github.com/cometbft/cometbft/blob/main/spec/abci/abci%2B%2B_basic_concepts.md#events

>     Method FinalizeBlock includes an events field at the top level in its Response*, and one events field per transaction included in the block.
>     ...
>     The value of the index flag is non-deterministic and may vary across different nodes in the network.

This makes it seem like CometBFT requires all fields except for the index to be deterministic in events.
However, in the code it does not look like it. In particular, the quite-strong assertion from the spec that all other fields must be strictly deterministic seems wrong, see the code here:
https://github.com/cometbft/cometbft/blob/main/abci/types/types.go#L143

I think in the ABCI methods doc, for each response field, it should be made obvious whether
that field needs to be deterministic, or whether it can be nondeterministic.
Obviously, it seems like a recommendation that all fields except for info and
log are deterministic, but it should be clear for which fields CometBFT enforces it.

melekes:
Thanks for highlighting this 👍 We need to update the docs to indicate that events are nondeterministic. I also like the proposal to add a separate column indicating whether the response field must be deterministic.

sergio-mena:
If you want to understand the reason for the current status, check out [ADR 058](https://github.com/cometbft/cometbft/blob/main/docs/architecture/tendermint-core/adr-058-event-hashing.md).

PR #1804, opened 2023-12-12
chore(spec): specify which fields must be deterministic

melekes:
Closes https://github.com/cometbft/cometbft/issues/1622

[**Rendered** 📄 ](https://github.com/cometbft/cometbft/blob/f54fc315153b6a15ba42ae939b46f6cf8d0b942b/spec/abci/abci%2B%2B_methods.md)

---

#### PR checklist


spec/abci/abci++_basic_concepts.md
```diff
@@ -311,9 +314,11 @@ message Event {
...
 ```
 
-The attributes of an `Event` consist of a `key`, a `value`, and an `index` flag. The
-index flag notifies the CometBFT indexer to index the attribute. The value of
-the `index` flag is non-deterministic and may vary across different nodes in the network.
+The attributes of an `Event` consist of a `key`, a `value`, and an `index`
+flag. The index flag notifies the CometBFT indexer to index the attribute.
+
+The `type` and `attributes` fields are non-deterministic and may vary across
+different nodes in the network.
```
mzabaluev, line 321: What about the `index` flag, it should keep being listed as non-deterministic too?

melekes: `attributes` includes `index`

spec/abci/abci++_methods.md
```diff
@@ -93,26 +93,26 @@ title: Methods
...
-    | codespace | string                | Namespace for the `code`.                                                                                                                                                                                          | 10           |
+    | Name      | Type                  | Description                                                                                                                                                                                                        | Field Number | Deterministic |
+    |-----------|-----------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|--------------|---------------|
+    | code      | uint32                | Response code.                                                                                                                                                                                                     | 1            | No            |
```
sergio-mena, line 107: For `Query`, rather than "No", I'd say it's an "N/A", and that ABCI method is not involved in SMR, it's rather a local query

spec/abci/abci++_methods.md
```diff
@@ -747,42 +748,42 @@ Most of the data structures used in ABCI are shared [common data structures](../
...
 * **Fields**:
 
-    | Name | Type   | Description                                    | Field Number |
-    |------|--------|------------------------------------------------|--------------|
-    | type | string | Type of Merkle proof and how it's encoded.     | 1            |
-    | key  | bytes  | Key in the Merkle tree that this proof is for. | 2            |
-    | data | bytes  | Encoded Merkle proof for the key.              | 3            |
+    | Name | Type   | Description                                    | Field Number | Deterministic |
+    |------|--------|------------------------------------------------|--------------|---------------|
+    | type | string | Type of Merkle proof and how it's encoded.     | 1            | No            |
```
sergio-mena, line 772: Similar comment, since this is used in queries to the local app, I'd say this is rather an "N/A"

sergio-mena (approved):
Awesome! Thanks for this!

PR #1814, opened 2023-12-13
feat!(store): "ideal" key layout

melekes:
https://github.com/cometbft/cometbft/issues/1041#issuecomment-1851452642

Most keys are now grouped by height except for block hash's key (-1)
and pending evidence (-2):

```
header: 11/0
block part: 11/1/0
block part 2: 11/1/1
...
commit: 11/2
seen commit: 11/3
ext commit: 11/4

validator set: 11/5
consensus params: 11/6
abci responses: 11/7

committed evidence: 11/8
```

```
block hash: -1/{hash}
pending evidence: -2/11/{ev.Hash}
```

```
light block: {chainID}/9/11
size: {chainID}/10
```

Refs https://github.com/cometbft/cometbft/issues/1041
Fixes https://github.com/cometbft/cometbft/issues/1857

---

#### PR checklist

- [ ] Tests written/updated


internal/evidence/pool.go
```diff
@@ -556,19 +553,36 @@ func evMapKey(ev types.Evidence) string {
...
+	subkeyCommitted = int64(8)
+
+	// prefixes must be unique within a single DB
+	prefixPending = int64(-2)
+)
 
-func keyCommitted(evidence types.Evidence) []byte {
-	return append([]byte{baseKeyCommitted}, keySuffix(evidence)...)
+func prefixPendingBytes() []byte {
+	prefix, err := orderedcode.Append(nil, prefixPending)
```
melekes, line 567: this can be calculated once in `init()`

internal/state/store.go
```diff
@@ -932,3 +911,49 @@ func int64ToBytes(i int64) []byte {
...
+func encodeKey(height, prefix int64) []byte {
+	res, err := orderedcode.Append(nil, height, prefix)
+	if err != nil {
+		panic(err)
+	}
+	return res
+}
+
+func validatorsKey(height int64) []byte {
+	// Since CometBFT requests block H and validators H+1 on every height, we
```
melekes, line 941: This is the most controversial change. But this is undoubtedly an improvement because the access pattern is clear (block H, parts H, validators H+1). Any thoughts?

jmalicevic: I think it should be fine although there are places where we fetch the validators info explicitly (like RPC `status` and pruning). But those are less frequent than block validation. Another frequent access pattern is loading the lastCommit (validators from height -1 ) which is called on every `ApplyBlock`

.changelog/unreleased/breaking-changes/1041-reconsider-representation-of-keys.md
```diff
@@ -0,0 +1,9 @@
+- Change `store`, `state`, `evidence` and `light` databases key format. The
+  keys are now grouped by height and lexicographically sorted using the
+  [orderedcode](https://github.com/google/orderedcode) library.
+
+  To upgrade, run `cometbft migrate-db`. **Before doing so, it's highly
```
tac0turtle, line 5: saw this pr on my feed, the title intrigued me. This step would prevent the sdk from adopting this change. This will cause downtime when users upgrade, there should be a way to make it lazy so users dont experience this downtime

ValarDragon: Would it be fine to live migrate the latest {N} blocks which hopefully happened quickly, and then migrate the rest in the background?

tac0turtle: im not sure there needs to be a migration at the point of upgrade, the node could be able to tell what is the old format vs new, then migrate in the background. Potentially causing 0 donwtime

melekes:
Closing in favor of https://github.com/cometbft/cometbft/pull/1764

@jmalicevic: "Based on the experiments, this change performs worse than #1764. The only improvement compared to #1764  is in the IO stats, but for compaction and general impact on the blockchain performance when it is on, it underperforms."

PR #1829, opened 2023-12-14
Fixing the ABCI grammar and updating the code for e2e tests to account for the new grammar. 

### LINKED ISSUE #1581: e2e: The new e2e tests found 2 scenarios where CometBFT is not respecting ABCI 1.0 grammar.
## Problem explanation

The new e2e tests (#902, #930) that verify whether CometBFT respects the ABCI grammar disclosed two scenarios in which this is not the case. Specifically, the scenarios were observed while running `ci.toml` testnet and they are the following: 

1. One of the validator nodes is calling `init_chain` even though it runs in a `recovery` mode and not in a `clean-start` mode.
2. One of the full nodes is not calling `init_chain` in the beginning.

The same can be reproduced by running e2e tests with `ci.toml` testnet. Please remember to enable ABCI tests by setting 
`abci_tests_enabled = true` in `ci.toml` file.

nenadmilosevic95:
---

This PR closes #1581. 
We discovered that the scenarios displayed in #1581 are not caused by a bug in the CometBFT but in the ABCI grammar. We fixed the grammar and updated the code for e2e tests accordingly. After doing this, the e2e tests that check whether CometBFT respects ABCI grammar are passing. 

#### PR checklist

- [ ] Tests written/updated


spec/abci/abci++_comet_expected_behavior.md
```diff
@@ -104,12 +105,19 @@ Let us now examine the grammar line by line, providing further details.
 >start               = clean-start / recovery
 >```
 
-* If the process is starting from scratch, CometBFT first calls `InitChain`, then it may optionally
-  start a _state-sync_ mechanism to catch up with other processes. Finally, it enters normal
-  consensus execution.
+* If the process is starting from scratch, depending on whether the _state-sync_ is enabled, it involves in the handshake
```
sergio-mena, line 108:
```suggestion
* If the process is starting from scratch, depending on whether the _state-sync_ is enabled, it engages in the handshake 
```
Just a suggestion

test/e2e/pkg/grammar/abci_grammar.md
```diff
@@ -1,15 +1,17 @@
 ```
-package "github.com/cometbft/cometbft/test/e2e/pkg/grammar/clean-start/grammar-auto"
+package "github.com/cometbft/cometbft/test/e2e/pkg/grammar/grammar-auto"
 
-Start : CleanStart ;
+Start : CleanStart | Recovery;
 
-CleanStart : InitChain StateSync ConsensusExec | InitChain ConsensusExec ;
+CleanStart : InitChain ConsensusExec | StateSync ConsensusExec ;
```
sergio-mena, line 6: My understanding of the grammar changes is that, now `InitChain` is always preceded by `Info`, is that the case here?

Oh wait... is it because we're not checking `Info` in the tests? So no point in adding it?

test/e2e/pkg/grammar/abci_grammar.md
```diff
@@ -29,32 +31,39 @@ ProcessProposal : "process_proposal" ;
...
 apply-chunk         = %s"<ApplySnapshotChunk>"
 info                = %s"<Info>"
 prepare-proposal    = %s"<PrepareProposal>"
 process-proposal    = %s"<ProcessProposal>"
+extend-vote         = %s"<ExtendVote>"
+got-vote            = %s"<VerifyVoteExtension>"
+finalize-block      = %s"<FinalizeBlock>"
+commit              = %s"<Commit>"
 
+*Note* `Info` and parts related to vote extensions are missing.
```
sergio-mena, line 66: Could you add a sentence briefly explaning that `Info` is missing because we're ignoring it in the tests, due to the fact that it is also triggered by the e2e tests at unpredictable places?

test/e2e/pkg/grammar/checker.go
```diff
@@ -161,33 +154,33 @@ func (g *Checker) verifyCleanStart(execution string) []*Error {
...
-			height:      err.Line - 1,
+	if len(errors) != 0 {
+		return errors
+	}
+	eType := symbols.NT_Recovery
+	if isCleanStart {
+		eType = symbols.NT_CleanStart
+	}
+	roots := bsrForest.GetRoots()
+	for _, r := range roots {
```
sergio-mena, line 170: Question: is this returning _all_ possible roots and all possible way the grammar could be applied to the trace?

This is important because of the following. I'll explain it with an example:

* Imagine that the last recovery trace (so an _actual_ recovery) contains the following trace: `InitChain`, `FinalizeBlock, `Commit`.
* This trace, obtained from a recovering process, can be misinterpreted as a clean-start trace

Hence it is important when we are checking for the root(s) [I understand: the starting non-terminal(s)]

test/e2e/tests/e2e_test.go
```diff
@@ -197,6 +197,11 @@ func fetchBlockChain(t *testing.T) []*types.Block {
 	return blocks
 }
 
+type ABCIExecution struct {
+	Type string
+	Reqs []*abci.Request
+}
```
sergio-mena, line 203: Where is this struct used?

sergio-mena:
This is awesome @nenadmilosevic95!
I think you can mark it as ready for review

test/e2e/pkg/grammar/abci_grammar.md
nenadmilosevic95: Yes, two things are new:
1. the previous grammar allowed both InitChain and StateSync. Now we know this is not correct, so we have InitChain or StateSync
2. Info is missing in the grammar. We do not check it because it can happen at any point.

test/e2e/pkg/grammar/abci_grammar.md
nenadmilosevic95: I will do this!

test/e2e/tests/e2e_test.go
nenadmilosevic95: It is not used. It was used in one of the previous versions. I will remove it. Thanks!

test/e2e/pkg/grammar/checker.go
nenadmilosevic95: My understanding is that the answer to your question is `Yes`. This is what is written in the `GetRoots` description: `// GetRoots returns all the roots of parse trees of the start symbol of the grammar`. Moreover, in my unit tests I am actually checking whether the execution you just described is passing the check as both the clean-start and recovery execution. See tests `empty-block-1` and `empty-block-2` in the `TestVerify` function in `pkg/grammar/checker_test.go` file.

sergio-mena (approved):
This is excellent!

sergio-mena:
Due to the merging of `feature/proto-upgrade` after this PR was started, I think the best way to backport this is:

* merge this into `v1.x`
* take the resulting commit and cherry-pick it on top of `main`, resolve the conflicts, and then open a PR against `main`

Let me know if you need assistance with this.

PR #1833, opened 2023-12-15
ci: fix failing golangci-lint action

melekes:
was failing for me in https://github.com/cometbft/cometbft/pull/1782

---

#### PR checklist

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,8 +1,9 @@
 linters:
+  disable-all: true
   enable:
     - asciicheck
     - bodyclose
-    - depguard
+    # - depguard
     - dogsled
     - dupl
     - errcheck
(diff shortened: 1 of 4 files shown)


.golangci.yml
```diff
@@ -1,22 +1,23 @@
 linters:
+  disable-all: true
   enable:
     - asciicheck
     - bodyclose
-    - depguard
+    # - depguard
```
sergio-mena, line 6: Rather than disabling these here, so far, we've been adding directives to disable the check on the code, on a case-by-case basis

thanethomson: This specific depguard linter requires a "whitelist" of allowed dependencies for use in the binary and testing, so there's no way to disable individual lints in the code.

melekes: fixed by https://github.com/cometbft/cometbft/pull/1833/commits/fd7dcb442afe35eb7ed1fecaa962c40ac33c67eb

thanethomson (approved):

PR #1836, opened 2023-12-15
chore(README.md): updating discord

Frierened:
Updated discord's channel link to cometbft's channel. Current one is a channel verified don't have right to see (it shows up empty).

Also I believe it would be nice to add an invite link (https://discord.com/invite/cosmosnetwork) so people that aren't in already don't get an error message.

I thought of something like :

Join the [Discord](https://discord.com/invite/cosmosnetwork) and discuss in the [chat](https://discord.com/channels/669268347736686612/1069933855307472906)

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ in order of preference:
   GitHub](https://github.com/cometbft/cometbft/discussions)
 - Reach out to us via [Telegram](https://t.me/CometBFT)
 - [Join the chat on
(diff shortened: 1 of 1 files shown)


thanethomson:
> Also I believe it would be nice to add an invite link (https://discord.com/invite/cosmosnetwork) so people that aren't in already don't get an error message.

Good idea @Frierened! I've incorporated that suggestion into #1874. Thanks for bringing this to our attention!

PR #1858, opened 2023-12-19
Allow blocksync to not verify all signatures

sergio-mena:
Contributes to #1749

Following [this comment](https://github.com/cometbft/cometbft/pull/1750#discussion_r1431035852) in #1750, we are allowing blocksync to follow its pre-#1749 behavior: it won't verify all signatures unconditionally, but will exit the loop when enough voting power is verified.

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

Issue #1878, opened 2023-12-22
blocksync: stopping a node while block sync (catch-up) is running often causes a panic


jchappelow:
### Setup

**CometBFT version**: `v0.38.x` branch at 721ac3cc, or `v0.38.2` tag.

**Have you tried the latest version**: yes

**ABCI app** (name for built-in, URL for self-written if it's publicly available): Unavailable to link our app yet, sorry.

**Environment**:
- **OS** (e.g. from /etc/os-release): Arch linux or Ubuntu or Darwin
- **Install tools**: `go build` with Go 1.21.5

### What happened?

With a node still in block sync (catch-up) while rapidly syncing blocks, if you stop the `Node` cleanly, the `blocksync.(*Reactor).poolRoutine` goroutine continues after `blocksync.(*Reactor).OnStop` returns, and even after `node.(*Node).OnStop` returns.  This typically results in one of two panics as it continues to operate using databases that have been closed.

This does not happen after block sync has switched to the consensus reactor.

### What did you expect to happen?

Wait for `poolRoutine` to return, and shutdown cleanly with no panics.

### How to reproduce it

Get one node running and synced on a chain with at least few hundred blocks to make it easier to catch.  It can be the only node using the only validator on this test network.

Start a second node as a non-validator, connect to validator.  It begins blocksync (do not use snapshot sync).  After a few seconds, the block rate picks up steadily.  When it is processing blocks quickly, interrupt/cancel to trigger `(*Node).Stop` -> `OnStop` where it begins stopping services and reactors.

Usually observe a panic with `blocksync.(*Reactor).poolRoutine` in the call stack.  A couple attempts may be needed, particularly if blocks are still going relatively slowly.  There seems to be a short warm-up period before it becomes steady and fast.

### Logs

```
panic: Failed to process committed block (327:D2EB81F27986A5CCD2E9C1C9EEED4D11B75BD99B2318E0662D8AE42BA9553D61): failed to create new app hash: DB Closed

goroutine 224 [running]:
github.com/cometbft/cometbft/blocksync.(*Reactor).poolRoutine(0xc00b6cc1e0, 0x0)
	/home/jon/github/cometbft/cometbft/blocksync/reactor.go:511 +0x18c8
created by github.com/cometbft/cometbft/blocksync.(*Reactor).OnStart in goroutine 220
	/home/jon/github/cometbft/cometbft/blocksync/reactor.go:124 +0x6e
```

(`DB Closed`)

OR

```
[...]

PR #1879, opened 2023-12-22
blocksync: wait for poolRoutine to stop in (*Reactor).OnStop

jchappelow:
Resolves https://github.com/cometbft/cometbft/issues/1878

```
blocksync: wait for poolRoutine to stop in (*Reactor).OnStop

blocksync.(*Reactor).poolRoutine goroutine lifetime was not supervised,
so OnStop returning did not provide a guarantee that it was done.  As a
result, the other services could be stopped and a subsequent call to
bcR.blockExec.ApplyBlock would cause a panic, typically a leveldb
closed error.

This change uses a sync.WaitGroup to ensure that OnStop returns after
poolRoutine has returned.

This also triggers the poolRoutine method to return on a signal from
bcR.pool.Quit() instead of just bcR.Quit(), which seems to be important
as (*Reactor).OnStop itself can only stop the BlockPool (bcR.pool), while
the BaseReactor.BaseService's quit channel will only be closed _after_
(*Reactor).OnStop has returned.
```

A consequence of the issue can be unrecoverable node data, so it can be kinda nasty if the node is interrupted while in catch up.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/internal/blocksync/reactor.go
+++ b/internal/blocksync/reactor.go
@@ -57,11 +57,11 @@ type Reactor struct {
 	// immutable
 	initialState sm.State
 
-	blockExec *sm.BlockExecutor
-	store     sm.BlockStore
-	pool      *BlockPool
-	blockSync bool
-	wg        sync.WaitGroup
+	blockExec     *sm.BlockExecutor
+	store         sm.BlockStore
+	pool          *BlockPool
+	blockSync     bool
+	poolRoutineWg sync.WaitGroup
 
 	requestsCh <-chan BlockRequest
 	errorsCh   <-chan peerError
@@ -127,7 +127,7 @@ func (bcR *Reactor) OnStart() error {
 		if err != nil {
 			return err
 		}
-		bcR.wg.Add(1)
+		bcR.poolRoutineWg.Add(1)
 		go bcR.poolRoutine(false)
 	}
 	return nil
@@ -143,7 +143,7 @@ func (bcR *Reactor) SwitchToBlockSync(state sm.State) error {
 	if err != nil {
 		return err
 	}
-	bcR.wg.Add(1)
+	bcR.poolRoutineWg.Add(1)
 	go bcR.poolRoutine(true)
 	return nil
 }
@@ -154,7 +154,7 @@ func (bcR *Reactor) OnStop() {
 		if err := bcR.pool.Stop(); err != nil {
 			bcR.Logger.Error("Error stopping pool", "err", err)
 		}
-		bcR.wg.Wait()
+		bcR.poolRoutineWg.Wait()
 	}
 }
 
@@ -289,7 +289,7 @@ func (bcR *Reactor) Receive(e p2p.Envelope) {
 // Handle messages from the poolReactor telling the reactor what to do.
 // NOTE: Don't sleep in the FOR_LOOP or otherwise slow it down!
 func (bcR *Reactor) poolRoutine(stateSynced bool) {
-	defer bcR.wg.Done()
+	defer bcR.poolRoutineWg.Done()
 
 	bcR.metrics.Syncing.Set(1)
 	defer bcR.metrics.Syncing.Set(0)
@@ -449,6 +449,9 @@ FOR_LOOP:
 				panic(fmt.Errorf("peeked first block without extended commit at height %d - possible node store corruption", first.Height))
 			}
 
+			// Before priming didProcessCh for another check on the next
+			// iteration, break the loop if the BlockPool has quit. This avoids
+			// case ambiguity of the outer select when two channels are ready.
 			select {
 			case <-bcR.pool.Quit():
 				break FOR_LOOP
--- a/internal/blocksync/reactor.go
+++ b/internal/blocksync/reactor.go
@@ -3,6 +3,7 @@ package blocksync
 import (
 	"fmt"
 	"reflect"
+	"sync"
 	"time"
 
 	bcproto "github.com/cometbft/cometbft/api/cometbft/blocksync/v1"
@@ -60,6 +61,7 @@ type Reactor struct {
 	store     sm.BlockStore
 	pool      *BlockPool
 	blockSync bool
+	wg        sync.WaitGroup
 
 	requestsCh <-chan BlockRequest
 	errorsCh   <-chan peerError
@@ -125,6 +127,7 @@ func (bcR *Reactor) OnStart() error {
 		if err != nil {
 			return err
 		}
+		bcR.wg.Add(1)
 		go bcR.poolRoutine(false)
 	}
 	return nil
@@ -140,6 +143,7 @@ func (bcR *Reactor) SwitchToBlockSync(state sm.State) error {
 	if err != nil {
 		return err
 	}
+	bcR.wg.Add(1)
 	go bcR.poolRoutine(true)
 	return nil
 }
@@ -150,6 +154,7 @@ func (bcR *Reactor) OnStop() {
 		if err := bcR.pool.Stop(); err != nil {
 			bcR.Logger.Error("Error stopping pool", "err", err)
 		}
+		bcR.wg.Wait()
 	}
 }
 
@@ -284,6 +289,8 @@ func (bcR *Reactor) Receive(e p2p.Envelope) {
 // Handle messages from the poolReactor telling the reactor what to do.
 // NOTE: Don't sleep in the FOR_LOOP or otherwise slow it down!
 func (bcR *Reactor) poolRoutine(stateSynced bool) {
+	defer bcR.wg.Done()
+
 	bcR.metrics.Syncing.Set(1)
 	defer bcR.metrics.Syncing.Set(0)
 
@@ -442,6 +449,11 @@ FOR_LOOP:
 				panic(fmt.Errorf("peeked first block without extended commit at height %d - possible node store corruption", first.Height))
 			}
 
+			select {
+			case <-bcR.pool.Quit():
+				break FOR_LOOP
+			default:
+			}
 			// Try again quickly next loop.
 			didProcessCh <- struct{}{}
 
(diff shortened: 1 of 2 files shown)


internal/blocksync/reactor.go
```diff
@@ -442,6 +449,11 @@ FOR_LOOP:
 				panic(fmt.Errorf("peeked first block without extended commit at height %d - possible node store corruption", first.Height))
 			}
 
+			select {
+			case <-bcR.pool.Quit():
```
melekes, line 453: q: why did you add `select` to this line specifically?

internal/blocksync/reactor.go
```diff
@@ -60,6 +61,7 @@ type Reactor struct {
 	store     sm.BlockStore
 	pool      *BlockPool
 	blockSync bool
+	wg        sync.WaitGroup
```
melekes, line 64: can we rename this to `poolRoutineWg`?

internal/blocksync/reactor.go
jchappelow: Sorry for the delay in responding.  This `select` is here to prevent a contest between the `Quit` and `didProcessCh` receives and on the outer select in the next iteration.  Just below this select it primes `didProcessCh`, which we want to avoid if we know we're shutting down.  I've just pushed a commit with a comment.

The quirk of `select` `case` with channels is essentially this the "uniform pseudo-random selection" that Go uses when there are multiple cases ready to go: https://play.golang.com/p/GNdVVAdSKIz

jchappelow:
Rebased with requested changes made.
Please let me know if you'd like me to squash this too.

internal/blocksync/reactor.go
melekes: Should we check for `case <-bcR.Quit():` as well then?

jchappelow: That is a good question.  Based on my review of this code (first time in cometbft code however) I don't think any of the `case <-bcR.Quit():` in the other `select`s in this file are even necessary since the `BlockPool`'s quit channel seems to be closed before the `Reactor`'s quit channel.

I think for consistency I should add that `case` however.

melekes (approved):
👍 

@jchappelow could you please add the changelog entry as well https://github.com/cometbft/cometbft/blob/main/CONTRIBUTING.md#changelog ?

melekes:
@jchappelow please avoid force-pushing in the future, as it makes incremental reviews impossible.

jchappelow:
That was just a rebase, no commits squashed, so all incremental changes are visible.  What's the preferred method to update as `main` changes?  I just used the Update button on github, and it appears to have retained all the commits with the requested revisions.

melekes:
> That was just a rebase, no commits squashed, so all incremental changes are visible. What's the preferred method to update as `main` changes? I just used the Update button on github, which did a rebase, and it has retained all the commits with the requested revisions.

I use `git merge main --no-ff`

jchappelow:
OK, I'll use merge commits in the future then.  I do agree that it's problematic if someone squashes and then rebases.  I've made sure to keep the individual revisions as separate commits for review, but will just merge main next time.

internal/blocksync/reactor.go
```diff
@@ -442,6 +449,17 @@ FOR_LOOP:
...
+			// iteration, break the loop if the BlockPool or the Reactor itself
+			// has quit. This avoids case ambiguity of the outer select when two
+			// channels are ready.
+			select {
+			case <-bcR.pool.Quit():
+				break FOR_LOOP
+			case <-bcR.Quit():
+				break FOR_LOOP
+			default:
+			}
```
cason, line 462: The same effect is achieved with an if testing if the two services are running (`IsRunning()`). With the advantage that the "atomic bool" used by this call is updated before closing the quit channel.

internal/blocksync/reactor.go
```diff
@@ -284,6 +289,8 @@ func (bcR *Reactor) Receive(e p2p.Envelope) {
 // Handle messages from the poolReactor telling the reactor what to do.
 // NOTE: Don't sleep in the FOR_LOOP or otherwise slow it down!
 func (bcR *Reactor) poolRoutine(stateSynced bool) {
+	defer bcR.poolRoutineWg.Done()
```
cason, line 292: Honestly, the `bcR.poolRoutineWg.Add(1)` should come here, instead of putting whenever this routine is started. We add a permit to the semaphore when starting this routine, then we defer returning the permit.

internal/blocksync/reactor.go
cason: Moreover, this is not really relevant for the solution.

The panic happens because this reactor is stopped, but this routine is not. If you solve this issue, you shouldn't have the panics coming from the block executor and the associated databases.

The reason for which is that when stopping the node (`Node.Stop()`) we first stop all reactors (the `Switch.OnStop()` does that), to then stop the other components. So while this reactor has not being stopped, which only happens when this routine leaves, we should not observe the reported problem.

cason (commented):
Thank for detecting this issue and proposing a solution. I have left some specif comments, to which I would add some more general comments.

First, I really doubt that this problem may jeopardize the node state. In particular because the main/persistent source of state are the DBs, which when are closed cannot be changed. Moreover, the DBs are atomic, meaning that either a write is fully complete, or nothing is written.

Second, I think that despite the provided solution, the best way to prevent panic errors is by using a `recover()` method stopping this reactor and the associated pool, while producing an error log message.

In any case, since we don't really have a test case that reproduces the reported behavior, it is hard to make sure that this solution indeed solves the problem.

jchappelow:
> Thank for detecting this issue and proposing a solution. I have left some specif comments, to which I would add some more general comments.

Thanks for reviewing.

> First, I really doubt that this problem may jeopardize the node state. In particular because the main/persistent source of state are the DBs, which when are closed cannot be changed. Moreover, the DBs are atomic, meaning that either a write is fully complete, or nothing is written.

Not directly, no.  What I've observed is probably a secondary effect of the untimely (and unrecovered) panic whereby other systems of the application rather than cometbft are killed.  Ideally everything would be robust to this, however, I would agree.  At the same time, I think it's best for cometbft to have a clean shutdown if possible.

> Second, I think that despite the provided solution, the best way to prevent panic errors is by using a `recover()` method stopping this reactor and the associated pool, while producing an error log message.

Here I think would argue it is proper to supervise the goroutine so that when the `Reactor` is stopped we are sure the `poolRoutine` isn't running still.   That said, I'm happy for any solution that prevents panics.

> In any case, since we don't really have a test case that reproduces the reported behavior, it is hard to make sure that this solution indeed solves the problem.

Yeah, apologize that I don't have a minimal reproduction.  I can try to craft one, but the description of it in https://github.com/cometbft/cometbft/issues/1878 does reliably create the panics in the blocksync code.  I'll consider a way to make that happen with a toy app catching-up on blocks if needed.

internal/blocksync/reactor.go
jchappelow: I might be misunderstanding the comment here.  The pattern with a waitgroup is to `Add` before (in the caller goroutine) launching it with `go` and `defer` the `Done` in the goroutine.  e.g.

```
wg.Add(1)
go func() {
	defer wg.Done()
	// .. do stuff e.g. bcR.poolRoutine(true)
	time.Sleep(time.Second)
}()

wg.Wait()
```

Such as in the cometbft example: https://github.com/cometbft/cometbft/blob/8acc13c31c86fc3dff9da071a1a151b1c5ca46e5/cmd/cometbft/commands/compact.go#L46-L48

internal/blocksync/reactor.go
jchappelow: > The same effect is achieved with an if testing if the two services are running (`IsRunning()`). With the advantage that the "atomic bool" used by this call is updated before closing the quit channel.

Ah, yes, I see the CAS in `(*BaseService).Stop` in advance of closing the quit channel.  Good call.  No need for a select in this spot unlike the outer `select`.

> The panic happens because this reactor is stopped, but this routine is not. If you solve this issue, you shouldn't have the panics coming from the block executor and the associated databases.
> 
> The reason for which is that when stopping the node (Node.Stop()) we first stop all reactors (the Switch.OnStop() does that), to then stop the other components. So while this reactor has not being stopped, which only happens when this routine leaves, we should not observe the reported problem.

Yes, absolutely correct that this is not necessary to synchronize the `poolRoutine` return with the service shutdown.  I just noticed the outer select was in a contest due to the subsequent `didProcessCh` being primed for the "Trying again quickly".

Waiting on `poolRoutine` did indeed solve the issue, but there were often many iterations of the loop after stop due to this quick retry path not being escaped by shutdown.

Thanks for the insightful comments here!

cason: I think you still have to check the `bcR.pool` also here for having the purposed effect.

The logic, as it appears is: stop the reactor, `OnStop` stops the `pool`, then waits for this routine to leave.

cason: >  I just noticed the outer select was in a contest due to the subsequent didProcessCh being primed for the "Trying again quickly". That is likely unintended.

This is true and, to be honest, a very debatable approach.

internal/blocksync/reactor.go
cason: You are probably right here, we should first add the permit to then start the thread, otherwise the thread may not be scheduled soon enough to add the permit.

internal/blocksync/reactor.go
```diff
@@ -149,7 +149,7 @@ func (bcR *Reactor) SwitchToBlockSync(state sm.State) error {
 	bcR.poolRoutineWg.Add(1)
 	go func() {
 		defer bcR.poolRoutineWg.Done()
-		bcR.poolRoutine(false)
+		bcR.poolRoutine(true)
```
jchappelow, line 152: copy pasta error in the last iteration.  The bool arg matches `main` again.

cason (approved):

melekes (approved):

PR #1892, opened 2023-12-25
chore(docs): alter the section about the effects of `timeout_commit`

melekes:
Refs https://github.com/cometbft/cometbft/pull/1798#pullrequestreview-1794258798

[📜 RENDERED](https://github.com/cometbft/cometbft/blob/6e6e28725acac4d637fbc91ce6216e63e9ed7470/docs/explanation/core/configuration.md#the-adverse-effect-of-using-inconsistent-timeout_propose-in-a-network)


docs/core/configuration.md
```diff
@@ -605,13 +605,13 @@ Here's a brief summary of the timeouts:
...
+### The adverse effect of using inconsistent `timeout_commit` in a network
 
-Here's an interesting question. What if the particular validator sets a very
-small `timeout_propose`?
+Here's an interesting question. What happens if a particular validator sets a
+very small `timeout_commit`, as compared to the rest of the network?
 
 Imagine there are only two validators in your network: Alice and Bob. Bob sets
-`timeout_propose` to 1s. Alice uses the default value of 3s. Bob will create
+`timeout_commit` to 1s. Alice uses the default value of 3s. Bob will create
```
cason, line 614: The default `timeout_commit` is 1s: https://github.com/cometbft/cometbft/blob/main/config/config.go#L1148

docs/core/configuration.md
```diff
@@ -605,13 +605,13 @@ Here's a brief summary of the timeouts:
...
 blocks ~ every second, Alice - every 3 seconds (given `create_empty_blocks` is
 `true`). Let's say they both have an equal voting power. Given the proposer
```
cason, line 616: The default for `create_empty_blocks` is `true`, maybe it is the case to mention this, something like "given the default value of..." or "given that that ... was not changed to `false`.

docs/core/configuration.md
```diff
@@ -634,12 +634,12 @@ What happens in reality is, however, a little bit different:
 #4 block - Alice
 ```
 
-That's because Bob is too fast at proposing blocks. This leaves Alice very
+That's because Bob is too fast at committing blocks. This leaves Alice very
```
cason, line 637: Line 638/639 below is inconsistent. If Bob and Alice have the same voting power, and they are the only validators, both have to vote to commit a block.

docs/core/configuration.md
```diff
@@ -634,12 +634,12 @@ What happens in reality is, however, a little bit different:
...
 little chances to propose a block and not always be catching up. Note every
 block Bob creates needs a vote from Alice to constitute 2/3+.
 
 Imagine now there are ten geographically distributed validators. One of them
-(Bob) sets `timeout_propose` to 1s. Others have it set to 3s. Now, Bob won't be
+(Bob) sets `timeout_commit` to 1s. Others have it set to 3s. Now, Bob won't be
```
cason, line 642: This is not precise as well. If Bob is the proposer, it will receive its own proposal relatively fast and move to the `prevote` round step. At this point it will wait for `2/3+` Prevotes. If they match, Bob precommits the matching value (a block ID or nil). Otherwise, it schedules `timeout_prevote`, which has default duration of 1s.

This is to say that Bob will block until it gets `2/3+` Prevotes, and doesn't matter that it entered the `propose` step before other nodes (due to smaller `timeout_commit`).

cason (commented):
I think the description of the scenarios is not precise (see my comments).

The problem I see on having different `timeout_commit` values is that non-proposes are more likely to Prevote `nil`, because `timeout_propose` expires, in the case the proposer has longer `timeout_commit`. In other words, the non-proposer start waiting for a Proposal and a full  Block, this only applies to round 0, after `timeout_commit` expires. The proposer, in its turn, only proposes a value in round 0 after `timeout_commits` is expired. If different values of `timeout_commit` are adopted, in particular the proposer has a longer one, its proposal is likely to arrive at non-proposes when they are already at `propose` step.

docs/explanation/core/configuration.md
```diff
@@ -632,18 +631,21 @@ What happens in reality is, however, a little bit different:
...
 #3 block - Bob
-#4 block - Alice
+#4 block - Bob
 ```
 
-That's because Bob is too fast at proposing blocks. This leaves Alice very
-little chances to propose a block and not always be catching up. Note that every
-block Bob creates needs a vote from Alice to constitute 2/3+.
+That's because Bob doesn't wait for a proposal from Alice (prevotes `nil`) and
+always moves onto the next round (NOT height). This leaves Alice no chances to
```
cason, line 638: I would remove the second part of this sentence: "and always moves onto the next round (NOT height)".

Both move to the next round when prevotes are received but do not form a quorum (as Bob prevoted nil), so that both precommit nil, thus eventually moving to the next round. This is a slow path because, in addition to receiving all the votes, they have to wait for two timeouts (prevote and precommit).

docs/explanation/core/configuration.md
```diff
@@ -656,3 +658,12 @@ being available and must move on if such is not responding.
...
 decentralization.
+
+### The adverse effect of using inconsistent `timeout_commit` in a network
+
+Let's look at the same scenario as before. There are ten geographically
+distributed validators. One of them (Bob) sets `timeout_commit` to 0s. Others
+have it set to 1s (the default value). Now, Bob will be the fastest producer
+because he doesn't wait for additional precommits after creating a block. If
+waiting for precommits (`timeout_commit`) is not incentivized, Bob will accrue
+more rewards compared to the other 9 validators.
```
cason, line 669: Not sure about the last sentence. What can happen does not wait for `timeout_commit` (i.e., it is zero):
1. Not all precommits are collected, so some nodes are not rewarded for them
1. Bob starts `timeout_propose` before all the others, therefore will stop waiting for other nodes' blocks early, this possibly prevoting `nil` for blocks other will accept an prevote for.

To achieve more rewards, Bob has to consider its vote in case 1. and ignore others. But it has in any case to include 2/3+ of them.

In case 2., if Bob has the same voting power than the others, or, more precisely, if it has `<1/3` of the voting power, the fact it prevotes for `nil` should not have impact, as other nodes will accept blocks proposed by other nodes.

Also regarding 1., Bob has the advantage of broadcasting its proposal early. This makes more likely that every other node will receive its proposal. That is, if the proposal takes 3.5s to be propagated, it will still be accepted, since the other processes are actually waiting 4s before prevoting nil. This is not the case of proposals produced by other nodes.

cason (commented):
Asking for some more clarification.

I do like the way you summarize pretty complex scenarios, I just would not like to produce wrong scenarios when simplifying their description.

docs/explanation/core/configuration.md
melekes: > If blocks are accepted faster, the other proposers turns will also arrive earlier, correct, and therefore everybody will benefit?

True, they will benefit too. But Bob will still be the fastest producer

docs/explanation/core/configuration.md
```diff
@@ -632,18 +631,21 @@ What happens in reality is, however, a little bit different:
...
+Alice has `timeout_propose` set to 3s. Alice never gets one because Bob has it
+set to 0s.
 
 Imagine now there are ten geographically distributed validators. One of them
-(Bob) sets `timeout_propose` to 1s. Others have it set to 3s. Now, Bob won't be
-able to move with the speed of 1s blocks because it won't gather 2/3+ of votes
-for its block proposal in time (1s). I.e., the network moves with the speed of
-time to accumulate 2/3+ of votes, not with the speed of the fastest proposer.
+(Bob) sets `timeout_propose` to 0s. Others have it set to 3s. Now, Bob won't be
+able to move with his own speed because it still needs 2/3 votes of the other
```
cason, line 645: I think the point here is that Bob will now pass his block faster because it has a lower timeout commit, right?

docs/explanation/core/configuration.md
cason: I think that the problem here is having a node progressing faster than the others. The `timeout_commit` is the minimum duration of a height, do you agree?

If few nodes ignore this parameter, as described above, I don't really see a problem here. Except for the fact that they will not reward all the validators that helped deciding the previous block.

cason: An information here that I am pretty sure, is that a node that is waiting for `timeout_commit` will not prevote until it expires. So, yes, the proposal arrives faster but is not that everyone will move faster because the proposer doesn't wait before proposing at round `0` .

cason: I think someone mentioned here that there is no clear incentive to wait for `timeout_commit` and this is simply bad.

melekes: > So, yes, the proposal arrives faster but is not that everyone will move faster because the proposer doesn't wait before proposing at round 0 .

won't other nodes decide on value faster in such case? i.e. when they receive proposal, they `X = (prevote, then precommit, then wait for timeout_commit)`. the `X` time doesn't change, but the rate of blocks changes

ancazamfir: > I think someone mentioned here that there is no clear incentive to wait for `timeout_commit` and this is simply bad.

fwiw, `The more precommits the proposer includes, the larger P is`, where `P` determines the proposer reward, see:
https://docs.cosmos.network/v0.45/modules/distribution/03_begin_block.html#reward-to-the-validators

ancazamfir: So if Alice was the proposer at `H` and Bob will be the proposer at `H+1`, Bob has an incentive to wait for more commits for `H`, commits that he will include in the proposal for `H+1`.

PR #1900, opened 2023-12-28
feat: DefaultHttpClient support setting proxy from env

levisyin:
This pr added a new default behavior to `DefaultHttpClient`: set proxy from environment, so that we can control it conveniently from environment


melekes (approved):
Thanks @levisyin ❤️ 

Could you please add a changelog entry https://github.com/cometbft/cometbft/blob/main/CONTRIBUTING.md#changelog ?

levisyin:
> Thanks @levisyin ❤️
> 
> Could you please add a changelog entry https://github.com/cometbft/cometbft/blob/main/CONTRIBUTING.md#changelog ?

@melekes Hi sir, changelog is added😀

PR #1907, opened 2023-12-30
feat: enable revive linter on the pbts branch

faddat:
This PR uses revive in largely the same manner as ibc-go does, and is targeted at the pbst branch.

It also became a great illustration of test flakiness:

* #1915 
* #1917 
* #1916 
* #1914 

In the course of writing this PR, all of the above tests were identified to be flaky.  We don't know if that is the test, or the code it is testing, please keep in mind. 

## important

please don't merge or attempt to review this before merging #1906.

---

#### PR checklist

- [ ] Tests written/updated


internal/state/indexer/block/kv/kv.go
```diff
@@ -364,7 +364,6 @@ func (idx *BlockerIndexer) Search(ctx context.Context, q *query.Query) ([]int64,
 
 		select {
 		case <-ctx.Done():
-			break
```
melekes, line 367: we need to revisit these breaks as it might be a developer error ⚠️

p2p/conn/connection.go
```diff
@@ -663,10 +663,6 @@ FOR_LOOP:
 
 	// Cleanup
 	close(c.pong)
-
-	for range c.pong {
```
melekes, line 667: interesting.. the intention was probably to drain before closing

rpc/jsonrpc/types/types_test.go
```diff
@@ -67,30 +67,38 @@ func TestUnmarshallResponses(t *testing.T) {
 		assert.Equal(*response, a)
 	}
 	response := &RPCResponse{}
-	err := json.Unmarshal([]byte(`{"jsonrpc":"2.0","id":true,"result":{"Value":"hello"}}`), response)
```
melekes, line 70: revert?

test/e2e/pkg/infra/digitalocean/digitalocean.go
```diff
@@ -1,200 +0,0 @@
-package digitalocean
```
melekes, line 1: revert?

light/detector.go
```diff
@@ -120,7 +120,6 @@ func (c *Client) compareNewHeaderWithWitness(ctx context.Context, errc chan erro
 	switch err {
 	// no error means we move on to checking the hash of the two headers
 	case nil:
-		break
```
melekes, line 123: same comment as above

melekes (approved):
👍

faddat:
cool man, a bit of self - review -- both of the items you flagged, made me raise my eyebrows a little bit too.

internal/state/indexer/block/kv/kv.go
faddat: on my part, or on the original author?

I had same "revisit" vibe.

faddat: that is to say that I am not 100% sure I made a mistake with them.  

but could have.

melekes: >  on the original author?

original author

melekes:
FYI, we're waiting for PBTS https://github.com/cometbft/cometbft/tree/feature/pbts to land on `main`.

faddat:
Note to self: 
2606eb3 is the last commit before merging the pbst branch into here, can PR to main from 2606eb3

sergio-mena (changes requested):
PBTS is a feature branch. The way we use these branches is that only changes relevant to the feature being worked on should land directly from a PR.
The rest should be reviewed and committed to `main`. And, via the periodic merges from `main` into the feature branch that we do, we will eventually get those changes. But, by planning the moment of those merges,  we _choose_ when those changes are to land on the feature branch.

faddat:
OK, figure I should close this @sergio-mena ?

I was basically trying to reduce future diffs.  

could just keep merging main to the other one.

faddat:
@sergio-mena the reason for this PR is that @melekes let me know that we were waiting for pbts to hit main.  I will close it for now, it can be reopened later if need be.  Could be a good reference later.

PR #1921, opened 2024-01-02
perf(crypto/merkle, crypto/tmhash): simplify+optimize SHA256 hashing of multiple byteslices

### LINKED ISSUE #1881: crypto/merkle: innerHash is unnecessarily inefficient by creating an entire fresh byteslice, copying just to invoke tmhash.Sum once yet it cloud invoke sha256.Hash.Write multiple times
## Bug Report
Noticed while in a security audit, if we look at this code https://github.com/cometbft/cometbft/blob/f68ba599a4c1968ecde34477d28930eb00dfc82c/crypto/merkle/hash.go#L34-L40

* creates a fresh byteslice with make(...)
* copies 3 byte slices in
* invokes tmhash.Sum(data)

meanwhile that code could simply be self contained and much simpler and more efficient just by using crypto/sha256.New then .Write on each byte slice like this

## Suggested change
```go
// returns tmhash(0x01 || left || right).
func innerHash(left []byte, right []byte) []byte {
        h := sha256.New()
        h.Write(innerPrefix)
        h.Write(left)
        h.Write(right)
        return h.Sum(nil)
}
```

or as a diff

```diff
diff --git a/crypto/merkle/hash.go b/crypto/merkle/hash.go
index 2acbb1ef3..523064842 100644
--- a/crypto/merkle/hash.go
+++ b/crypto/merkle/hash.go
@@ -1,6 +1,7 @@
 package merkle
 
 import (
+	"crypto/sha256"
 	"hash"
 
 	"github.com/cometbft/cometbft/crypto/tmhash"
@@ -32,11 +33,11 @@ func leafHashOpt(s hash.Hash, leaf []byte) []byte {
 
 // returns tmhash(0x01 || left || right).
 func innerHash(left []byte, right []byte) []byte {
-	data := make([]byte, len(innerPrefix)+len(left)+len(right))
-	n := copy(data, innerPrefix)
-	n += copy(data[n:], left)
-	copy(data[n:], right)
-	return tmhash.Sum(data)
+	h := sha256.New()
+	h.Write(innerPrefix)
+	h.Write(left)
+	h.Write(right)
+	return h.Sum(nil)
 }
 
 func innerHashOpt(s hash.Hash, left []byte, right []byte) []byte {
```

### Benchmarking code
```go
package merkle

import (
	"crypto/sha256"
	"strings"
	"testing"
)

var sink any = nil

type innerHashTest struct {
	left, right string
}

var innerHashTests = []*innerHashTest{
	{"aaaaaaaaaaaaaaa", "                    "},
	{"", ""},
	{"                        ", "a    ff     b    f1    a"},
	{"ffff122fff", "ffff122fff"},
	{"😎💡✅alalalalalalalalalallalallaallalaallalalalalalalalaallalalalalalala", "😎💡✅alalalalalalalalalallalallaallalaallalalalalalalalaallalalalalalalaffff122fff"},
	{strings.Repeat("ff", 1<<10), strings.Repeat("00af", 4<<10)},
	{strings.Repeat("f", sha256.Size), strings.Repeat("00af", 10<<10)},
	{"aaaaaaaaaaaaaaaaaaaaaaaaaaaffff122fffaaaaaaaaa", "aaaaaaaaaffff1aaaaaaaaaaaaaaaaaa22fffaaaaaaaaa"},
}

func BenchmarkInnerHash(b *testing.B) {
	b.ReportAllocs()

	for i := 0; i < b.N; i++ {
		for _, tt := range innerHashTests {
			got := innerHash([]byte(tt.left), []byte(tt.right))
			if g, w := len(got), sha256.Size; g != w {
				b.Fatalf("size discrepancy: got %d, want %d", g, w)
			}
			sink = got
		}
	}

	if sink == nil {
	}
}
```

### Benchmark results

```shell
name         old time/op    new time/op    delta
InnerHash-8     357µs ± 3%     353µs ± 2%     ~     (p=0.129 n=9+10)

name         old alloc/op   new alloc/op   delta
InnerHash-8    69.1kB ± 0%    60.1kB ± 0%  -12.98%  (p=0.000 n=9+9)

name         old allocs/op  new allocs/op  delta
InnerHash-8      24.0 ± 0%      23.0 ± 0%   -4.17%  (p=0.000 n=10+10)
```

/cc @elias-orijtech @thanethomson @lasarojc

odeke-em:
This change adds a more efficient API in tmhash with "SumMany" whose job is to produce the SHA256 sum of multiple byteslices. It is used inside crypto/merkle.innerHash which used a naive and inefficient way of hashing multiple byteslices. Benchmark results reflect these improvements:

```shell
name         old time/op    new time/op    delta
InnerHash-8     161µs ± 1%     160µs ± 5%     ~     (p=0.143 n=10+10)

name         old alloc/op   new alloc/op   delta
InnerHash-8    69.1kB ± 0%    60.1kB ± 0%  -12.98%  (p=0.000 n=10+9)

name         old allocs/op  new allocs/op  delta
InnerHash-8      24.0 ± 0%      23.0 ± 0%   -4.17%  (p=0.000 n=10+10)
```

Fixes #1881

### DIFF
--- a/.changelog/unreleased/improvements/1921-crypto-merkle-innerHash.md
+++ b/.changelog/unreleased/improvements/1921-crypto-merkle-innerHash.md
@@ -1,0 +1,1 @@
+- `[crypto/merkle]` faster calculation of hashes ([#1921](https://github.com/cometbft/cometbft/pull/1921))
--- a/crypto/tmhash/hash.go
+++ b/crypto/tmhash/hash.go
@@ -21,6 +21,18 @@ func Sum(bz []byte) []byte {
 	return h[:]
 }
 
+// SumMany takes at least 1 byteslice along with a variadic
+// number of other byteslices and produces the SHA256 sum from
+// hashing them as if they were 1 joined slice.
+func SumMany(data []byte, rest ...[]byte) []byte {
+	h := sha256.New()
+	h.Write(data)
+	for _, data := range rest {
+		h.Write(data)
+	}
+	return h.Sum(nil)
+}
+
 //-------------------------------------------------------------
 
 const (
--- a/crypto/merkle/bench_test.go
+++ b/crypto/merkle/bench_test.go
@@ -1,0 +1,42 @@
+package merkle
+
+import (
+	"crypto/sha256"
+	"strings"
+	"testing"
+)
+
+var sink any = nil
+
+type innerHashTest struct {
+	left, right string
+}
+
+var innerHashTests = []*innerHashTest{
+	{"aaaaaaaaaaaaaaa", "                    "},
+	{"", ""},
+	{"                        ", "a    ff     b    f1    a"},
+	{"ffff122fff", "ffff122fff"},
+	{"😎💡✅alalalalalalalalalallalallaallalaallalalalalalalalaallalalalalalala", "😎💡✅alalalalalalalalalallalallaallalaallalalalalalalalaallalalalalalalaffff122fff"},
+	{strings.Repeat("ff", 1<<10), strings.Repeat("00af", 4<<10)},
+	{strings.Repeat("f", sha256.Size), strings.Repeat("00af", 10<<10)},
+	{"aaaaaaaaaaaaaaaaaaaaaaaaaaaffff122fffaaaaaaaaa", "aaaaaaaaaffff1aaaaaaaaaaaaaaaaaa22fffaaaaaaaaa"},
+}
+
+func BenchmarkInnerHash(b *testing.B) {
+	b.ReportAllocs()
+
+	for i := 0; i < b.N; i++ {
+		for _, tt := range innerHashTests {
+			got := innerHash([]byte(tt.left), []byte(tt.right))
+			if g, w := len(got), sha256.Size; g != w {
+				b.Fatalf("size discrepancy: got %d, want %d", g, w)
+			}
+			sink = got
+		}
+	}
+
+	if sink == nil {
+		b.Fatal("Benchmark did not run!")
+	}
+}
--- a/crypto/merkle/hash.go
+++ b/crypto/merkle/hash.go
@@ -32,11 +32,7 @@ func leafHashOpt(s hash.Hash, leaf []byte) []byte {
 
 // returns tmhash(0x01 || left || right).
 func innerHash(left []byte, right []byte) []byte {
-	data := make([]byte, len(innerPrefix)+len(left)+len(right))
-	n := copy(data, innerPrefix)
-	n += copy(data[n:], left)
-	copy(data[n:], right)
-	return tmhash.Sum(data)
+	return tmhash.SumMany(innerPrefix, left, right)
 }
 
 func innerHashOpt(s hash.Hash, left []byte, right []byte) []byte {
--- a/crypto/tmhash/bench_test.go
+++ b/crypto/tmhash/bench_test.go
@@ -1,0 +1,52 @@
+package tmhash
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"strings"
+	"testing"
+)
+
+var sink any = nil
+
+var manySlices = []struct {
+        name string
+	in   [][]byte
+	want [32]byte
+}{
+	{
+                name: "all empty",
+		in:   [][]byte{[]byte(""), []byte("")},
+		want: sha256.Sum256(nil),
+	},
+	{
+                name: "ax6",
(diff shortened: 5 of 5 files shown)


ValarDragon (approved):
LGTM, thanks for reducing allocation overhed! Though the changelog comment of "much more efficient" seems a bit odd. Its a non-existent time speedup, and a 10% allocation reduction!

CHANGELOG.md
```diff
@@ -268,6 +268,7 @@ the stability guarantees we provide for pre-releases.
 
 ### IMPROVEMENTS
 
+- crypto/merkle.innerHash is much more efficient ([#1921](https://github.com/cometbft/cometbft/pull/1921))
```
melekes, line 271:
```suggestion
```

Please see https://github.com/cometbft/cometbft/blob/main/CONTRIBUTING.md#changelog

crypto/tmhash/hash.go
```diff
@@ -21,6 +21,20 @@ func Sum(bz []byte) []byte {
 	return h[:]
 }
 
+// SumMany takes at least 1 byteslice along with a variadic
+// number of other byteslices and produces the SHA256 sum from
+// hashing them as if they were 1 joined slice.
+// Its purpose is to avoid programmers from thinking of inefficient patterns
```
melekes, line 27:
```suggestion
```

crypto/tmhash/hash.go
```diff
@@ -21,6 +21,20 @@ func Sum(bz []byte) []byte {
...
+// like manually joining byteslices before hashing them altogether.
```
melekes, line 28:
```suggestion
```

It's sufficient to add this to the issue's / PR description.

melekes (commented):
Thanks @odeke-em ❤️ minor comments, otherwise LGTM

CHANGELOG.md
odeke-em: Done, than you @melekes!

crypto/tmhash/hash.go
odeke-em: Gotcha, I've removed it. Thank you @melekes!

odeke-em:
>LGTM, thanks for reducing allocation overhed! Though the changelog comment of "much more efficient" seems a bit odd. Its a non-existent time speedup, and a 10% allocation reduction!

Thanks for the reivew @ValarDragon! I don't follow what you mean by "non-existent time speedup and 10% allocation reduction" precluding efficiency -- innerHash is called a whole lot inside crypto/merkle which is a critical package, in fact even if it were a 5% allocation reduction I'd call it that because the API change means no wastage of CPU cycles but also the runtime doing less work. Also we've had quite the rhetoric in the Cosmos community of "optimizations must be huge updates or re-engineer the whole thing" but sadly that leads to a ton of inefficient code patterns that linger for 7+ years, moreover very few people (except yourself and Osmosis) actually profile code but even then folks aren't looking at profiles intently hence massive bloats. 

Thanks for the code reviews @ValarDragon and @melekes!

CHANGELOG.md
```diff
@@ -268,6 +268,7 @@ the stability guarantees we provide for pre-releases.
 
 ### IMPROVEMENTS
 
+- `[crypto/merkle]` innerHash is much more efficient ([#1921](https://github.com/cometbft/cometbft/pull/1921))
```
melekes, line 271: this must be in a separate file as per https://github.com/cometbft/cometbft/blob/main/CONTRIBUTING.md#changelog

odeke-em: Oh, damn I totally missed that, I've made the new file and moved it. Thank you @melekes!

melekes (approved):
👍

melekes:
@odeke-em could you please fix the build error:

```
# github.com/cometbft/cometbft/crypto/merkle [github.com/cometbft/cometbft/crypto/merkle.test]
Error: crypto/merkle/bench_test.go:4:2: "bytes" imported and not used
```

.changelog/unreleased/improvements/1921-crypto-merkle-innerHash.md
```diff
@@ -0,0 +1 @@
+- `[crypto/merkle]` innerHash is much more efficient ([#1921](https://github.com/cometbft/cometbft/pull/1921))
```
melekes, line 1:
```suggestion
- `[crypto/merkle]` faster calculation of hashes ([#1921](https://github.com/cometbft/cometbft/pull/1921))
```

odeke-em: Done and thanks!

PR #1922, opened 2024-01-02
feat: add support for state sync at specific height

Pythonberg1997:
### Description

This pr is to introduce an enhancement to support state sync at a specific block height.

### Rationale

In the current code, state sync automatically selects the best (usually the most recent) snapshot for syncing, which may not be very convenient. Sometimes, we might prefer to sync to a specific height of our choosing.

### Changes

Notable changes:
* add support for state sync at a specific block height

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -984,6 +984,7 @@ type StateSyncConfig struct {
 	DiscoveryTime       time.Duration `mapstructure:"discovery_time"`
 	ChunkRequestTimeout time.Duration `mapstructure:"chunk_request_timeout"`
 	ChunkFetchers       int32         `mapstructure:"chunk_fetchers"`
+	TargetHeight 	  	int64         `mapstructure:"target_height"`
 }
 
 func (cfg *StateSyncConfig) TrustHashBytes() []byte {
--- a/internal/statesync/snapshots.go
+++ b/internal/statesync/snapshots.go
@@ -2,6 +2,7 @@ package statesync
 
 import (
 	"crypto/sha256"
+	"errors"
 	"fmt"
 	"math/rand"
 	"sort"
@@ -124,6 +125,19 @@ func (p *snapshotPool) Best() *snapshot {
 		return nil
 	}
 	return ranked[0]
+}
+
+// At returns the known snapshot at the height.
+// If there are multiple snapshots or no snapshot at the height, it will return nil.
+func (p *snapshotPool) At(height uint64) (*snapshot, error) {
+	keys := p.heightIndex[height]
+	if len(keys) > 1 {
+		return nil, errors.New("multiple snapshots at height")
+	}
+	for key := range keys {
+		return p.snapshots[key], nil
+	}
+	return nil, errors.New("no snapshot at height")
 }
 
 // GetPeer returns a random peer for a snapshot, if any.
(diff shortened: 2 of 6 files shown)


internal/statesync/snapshots.go
```diff
@@ -126,6 +127,19 @@ func (p *snapshotPool) Best() *snapshot {
 	return ranked[0]
 }
 
+// At returns the known snapshot at the height.
+// If there are multiple snapshots or no snapshot at the height, it will return nil.
+func (p *snapshotPool) At(height uint64) (*snapshot, error) {
+	keys := p.heightIndex[height]
+	if len(keys) > 1 {
+		return nil, errors.New("multiple snapshots at height")
```
melekes, line 135: from the above comment: "If there are multiple snapshots or no snapshot at the height, it will return nil."

please amend the comment or the code

config/config.go
```diff
@@ -984,6 +984,7 @@ type StateSyncConfig struct {
 	DiscoveryTime       time.Duration `mapstructure:"discovery_time"`
 	ChunkRequestTimeout time.Duration `mapstructure:"chunk_request_timeout"`
 	ChunkFetchers       int32         `mapstructure:"chunk_fetchers"`
+	TargetHeight 	  	int64         `mapstructure:"target_height"`
```
melekes, line 987: Please format w/ `make lint-format` or default Go formatter

internal/statesync/snapshots.go
```diff
@@ -126,6 +127,19 @@ func (p *snapshotPool) Best() *snapshot {
...
+	}
+	for key := range keys {
+		return p.snapshots[key], nil
+	}
+	return nil, errors.New("no snapshot at height")
```
melekes, line 140: same

config/config.go
melekes, line 987: Could you add a validation rule in `ValidateBasic` to check that `TargetHeight` is `>= 0`?

melekes (changes requested):
seems like a reasonable feature-request to me 👍  Thanks for contributing @Pythonberg1997

internal/statesync/syncer.go
```diff
@@ -161,7 +161,14 @@ func (s *syncer) SyncAny(discoveryTime time.Duration, retryHook func()) (sm.Stat
 	for {
 		// If not nil, we're going to retry restoration of the same snapshot.
 		if snapshot == nil {
-			snapshot = s.snapshots.Best()
+			if targetHeight > 0 {
+				snapshot, err = s.snapshots.At(uint64(targetHeight))
+				if err != nil {
```
cason, line 166: Not finding a snapshot is not a reason for leaving this method. As you can observe after line 174, we try again retrieving the snapshot (we send a further `SnapshotsRequest` to all peers) if, for instance, we didn't receive it yet.

internal/statesync/syncer_test.go
cason, line 1: We would really need some unit tests covering this new use case.

cason (commented):
Our policy requires opening an issue before submitting a PR. In that case we can discuss the specifics of the implementation.

Regarding the PR, we would need some test cases to verify the operation of the proposed solution. Moreover, we need documentation and to add the new configuration flag to the configuration file (`config/toml.go`).

sergio-mena:
Following up on @cason 's comment. The way we tackle user requests is to
* first understand the problem we are trying to solve. For that, the user needs to create an issue or a discussion
* once the problem is understood, we then explore the solution space. The user may submit a companion PR containing _one_ possible solution
* a decision is made on why solution/approach will be followed to fix the problem

PR #1930, opened 2024-01-02
run tests 40x

faddat:
This PR should (please) never be merged.

While I was working on the linting I hit a lot of test failures that I simply couldn't explain, which could be solved by stuff like changing godoc.  So This PR will run the tests 40 times so that we can know which are flaky.

I will continue to run it until unique failures stop happening.

I suspect it is the only way.

😬 - I just realized that if we want to really be sure, we should add a job like this as default, until the issues disappear.  

The tests need to run a million times the same way.

Right now we cannot achieve 40.

NOTE: this is not a personal attack it is a bug report.

NOTE2: every green check mark on this repo is just luck

---

#### PR checklist

- [ ] Tests written/updated


faddat:
But just to make sure it is understood this isn't some kind of personal attack let's try v0.37.x which is used widely in production.

Shall it pass?

faddat:
Marko gave reallly sage feedback on this approach -- I should have been doing this against master, so I'll give that a go now.

Anyone know how to capture stats?

I think we can solve these issues.

faddat:
Note: since this PR had lots of changes (which I'd love to see merged, except for 40xing the tests) -- here's a PR that is just against master which shows the same failures. 

https://github.com/cometbft/cometbft/actions/runs/7389641484/job/20102926059?pr=1950

melekes:
Closing due to inactivity. Please re-open if needed.

PR #1936, opened 2024-01-02
40x for the 37 series

faddat:
Please do not merge this.  It is for finding flaky tests and the like.

This is a self-documenting test that will leave permanent evidence in CI logs of the state of the 37 series.


melekes:
We've got what we wanted, so closing.

Issue #1915, opened 2024-01-01
TestListenerTimeoutReadWrite is flaky


faddat:
## Bug Report

TestListenerTimeoutReadWrite is flaky

### Setup

I just used CI

### Config

CI

### What happened?

With no code changes other than a godoc, TestListenerTimeoutReadWrite decided to fail

### What did you expect to happen?

I expected no tests to fail other than those that failed in the previous commit

### How to reproduce it

Just work on comet

### logs
<img width="527" alt="Screenshot 2024-01-02 at 3 12 09 AM" src="https://github.com/cometbft/cometbft/assets/7142025/fb2803af-985d-428e-9a4c-6e522ebce95c">

* https://github.com/cometbft/cometbft/pull/1907/commits/9f5f93cb5c3cd9badf695dd520ff2bad7997a048

Another one:

* https://github.com/cometbft/cometbft/actions/runs/7379907607/job/20076706226

But then I added a godoc, and another test failed!

TestListenerTimeoutReadWrite

There were no code changes other than the godoc.  Godoc should not affect test success or failure.  The situation with flaky tests is robbing us of a source of truth for comet.

### Anything else we need to know

Making CI reliable provides a source of truth as to the state of the code and eliminates "works on my machine" problems.  If we wish to improve the reliability of comet, we need to determine why these tests are flaky, and fix it.

faddat:
note to self: TestByzantinePrevoteEquivocation could be too

faddat:
fixed in #1954

PR #1947, opened 2024-01-02
fix TestListenerTimeoutReadWrite

faddat:
This is a fix to a frequently seen test failure, TestListenerTimeoutReadWrite.

Please evaluate it carefully, as I do wish to preserve the rigor of the original test and in this case I have not gone deeply through the underlying code.

closes: #1915

---

#### PR checklist

- [x] Tests written/updated


privval/socket_listeners_test.go
```diff
@@ -97,19 +97,18 @@ func TestListenerTimeoutAccept(t *testing.T) {
 
 func TestListenerTimeoutReadWrite(t *testing.T) {
 	const (
-		// This needs to be long enough s.t. the Accept will definitely succeed:
```
melekes, line 100: I find these comments helpful. Would you mind putting them back?

melekes (commented):
Thanks @faddat ❤️ one minor comment

privval/socket_listeners_test.go
faddat: I don't mind at all :)

faddat:
its good to go

privval/socket_listeners_test.go
melekes: Maybe you forgot to push the changes?

faddat: I was hesitant to touch the repo at all:

https://github.com/cometbft/cometbft/discussions/1424

privval/socket_listeners_test.go
```diff
@@ -104,33 +104,39 @@ func TestListenerTimeoutReadWrite(t *testing.T) {
 		// Note: this controls how long this test actually runs.
 		timeoutReadWrite = 10 * time.Millisecond
 	)
-	for _, tc := range listenerTestCases(t, timeoutAccept, timeoutReadWrite) {
-		go func(dialer SocketDialer) {
-			_, err := dialer()
+
+	for i := 0; i < 100; i++ {
```
melekes, line 108: Do we need to repeat 100 times?

faddat: image


(and I checked and it costs a few seconds)

melekes: But the same could be said about all tests. Does that mean we have to repeat them 100 times too?

faddat: I suppose your POV is also accurate.

It passes 100 rounds well.

melekes (approved):
👍

Issue #1917, opened 2024-01-01
consensus: TestByzantinePrevoteEquivocation is flaky


faddat:
## Bug Report

TestByzantinePrevoteEquivocation is flaky

### Setup

CI

### Config

CI

### What happened?

Sometimes TestByzantinePrevoteEquivocation works.  Sometimes TestByzantinePrevoteEquivocation does not work. 

In flaky test cases we don't know if it is the code or the test, by the way.  This means that ByzantinePrevoteEquivocation could be malfunctioning.

### What did you expect to happen?

I expect tests to work consistently.

### How to reproduce it

Work on Comet

### Logs

Fail Log
* https://github.com/cometbft/cometbft/actions/runs/7379656614/job/20076181734?pr=1907

Looking at the surrounding commits in:

#1907 it is quite clear no relevant changes were made.

Magic Success 1 commit later with no relevant code changes:
* https://github.com/cometbft/cometbft/actions/runs/7379829117/job/20076535412?pr=1907

### Anything else we need to know

While we call these flaky tests, the name doesn't do reality justice.  They're ambiguity as to weather or not one function or another actually works as designed, and they greatly slow development on comet, while making it more expensive.

cason:
After investigating the logs of tests that have failed, I am starting to suspect that in some runs validators _do not receive_ the conflicting votes from height 2. I am sure, from these logs, that validators don't add conflicting votes and don't produce the associated evidence of misbehavior.

I can think on some reasons for the above mentioned behavior:
 - A node receives the conflicting Prevote of height 2 when it already moved to the next height
 - A node does not send one of the conflicting Prevote messages to a peer (we need that, since the Byzantine node splits the network with two conflicting Prevotes) because it realized that that peer does not need it anymore (e.g., it moved to the Precommit round step)
 - A node does not send one of the conflicting Prevote messages to a peer essentially because by randomly picking Prevote messages to send to that peer, it happens it never peak the vote in question before the two cases above are observed.
 
In the failed test case I have found, only one of the 3 correct validators receive both Prevotes. Unfortunately some log messages are not indexed by the node who produces them, but following the second Receive (the one that would be the equivocation), there are messages referring to committing a block. Since the Receive log entry is produced by the reactor, its is always possible that the vote was received (network level), but ignored at consensus level.

faddat:
@cason this sounds to me like the scary place where p2p meets consensus.

cason:
[...]

PR #1948, opened 2024-01-02
fix TestByzantinePrevoteEquivocation

faddat:
This PR adds additional logging to TestByzantinePrevoteEquivocation and ensures that when it fails it gives a clearer picture as to why.  Please check it carefully, but in this case, I am confident in these changes as both safe and effective.

closes: #1917

---

#### PR checklist

- [x] Tests written/updated


internal/consensus/byzantine_test.go
```diff
@@ -286,7 +288,8 @@ func TestByzantinePrevoteEquivocation(t *testing.T) {
 				assert.Equal(t, prevoteHeight, ev.Height())
 			}
 		}
-	case <-time.After(20 * time.Second):
+	case <-time.After(timeout):
+		t.Logf("Evidence from each validator: %v", evidenceFromEachValidator) // Log evidence
```
melekes, line 292: Have you seen `evidenceFromEachValidator` containing anything other than `[<nil> <nil> <nil> <nil>]` in case of failure?

internal/consensus/byzantine_test.go
```diff
@@ -276,6 +276,8 @@ func TestByzantinePrevoteEquivocation(t *testing.T) {
 	pubkey, err := bcs.privValidator.GetPubKey()
 	require.NoError(t, err)
 
+	const timeout = 30 * time.Second // Increase timeout to 30 seconds
```
melekes, line 279: Does this reduce the frequency of test failures?

faddat: yes, I think there was a performance aspect to it

internal/consensus/byzantine_test.go
faddat: no, this is all I've seen.

faddat:
my understanding is that evidence submission is really slow and could…
… even be a halt vector. So, I propose like:

* for now we bump the timeout to 35s, but we shouldn't go higher
* we open an issue to see if we can make evidence submission faster

cason:
I think the relevant information to log here is how many heights were executed (blocks committed) without the evidence being committed. In this case, the wait clause should refer to the number of heights executed, not to an arbitrary amount of time.

internal/consensus/byzantine_test.go
cason: This indicates that, after that time, no evidence was committed to a block. It is pretty unlikely (although possible) that some nodes have seen that block, while others don't. In this case the output will contain non-nil values.

faddat:
Love this.  I am going to give it a poke.

faddat:
@cason @melekes -- I have decided to try two things:

1) raise the time to 180s and run 40x (just to see)
2) add logging for # of blocks past.  

Thanks to you both for your help.

faddat:
.....opinion: this test is enormous. 

That is why it is failing.

faddat:
closing in favor of #1954

Issue #1932, opened 2024-01-02
TestMempoolProgressInHigherRound is flaky


faddat:
## Bug Report

TestMempoolProgressInHigherRound is flaky

### Setup

CI

### Config

CI

### What happened?

https://github.com/cometbft/cometbft/actions/runs/7387307772/job/20095751485?pr=1930

### What did you expect to happen?

More failures than actually occured

### How to reproduce it

run the tests 40 times.  On every in-prod release version. 

### Logs

https://github.com/cometbft/cometbft/actions/runs/7387307772/job/20095751485?pr=1930

### `dump_consensus_state` output

N/A

### Anything else we need to know

If we do this rigorously, we can eliminate the test failures or the underlying bugs

Issue #1923, opened 2024-01-02
Multiple flaky tests


faddat:
## Bug Report

https://github.com/cometbft/cometbft/actions/runs/7384251905/job/20086817549?pr=1907

### Setup

CI

### Config

CI

### What happened?

made changes that should not cause these failures

### What did you expect to happen?

consistency

### How to reproduce it

work on comet

faddat:
I've begun to submit fixes to the flaky tests.  I would like to urge the maintainers to check the fixes extremely harshly, such that none of the fixes endangers the purpose of any of the tests.

faddat:
I think I figured out `TestReader`

```log
--- FAIL: TestReader (0.44s)
    io_test.go:94: r.Status(1)
        expected: {2024-01-19 18:27:55.48 +0000 UTC 10 1 100 100 100 100 0 100ms 0s 0s 0.000% true}
        got     : {2024-01-19 18:27:55.48 +0000 UTC 10 1 100 100 100 100 0 100ms 20ms 0s 0.000% true}
FAIL
```

So basically, we're expecting that to do.... something in 0s but it takes 20ms.  Also, I kinda think we can tolerate the 20ms.

https://github.com/cometbft/cometbft/blob/d766d20c0609e3018e26f30aadf91bd322f8cad9/internal/flowrate/io_test.go#L82-L83

PR #1953, opened 2024-01-02
fix: make TestMempoolProgressInHigherRound less flaky

faddat:
This PR should make TestMempoolProgressInHigherRound less flaky and a teensy big more rigorous. 

The critical change is the order of the steps in the test itself, which does seem to make the test more reliable.

If the order should not have changed then we should probably just close this PR and try to find another way to make the test more reliable.

closes: #1932 

found in: #1954 

---

#### PR checklist

- [x] Tests written/updated


internal/consensus/mempool_test.go
```diff
@@ -70,41 +70,58 @@ func TestMempoolProgressAfterCreateEmptyBlocksInterval(t *testing.T) {
...
+
+	// Subscribe to events
 	newBlockCh := subscribe(cs.eventBus, types.EventQueryNewBlock)
 	newRoundCh := subscribe(cs.eventBus, types.EventQueryNewRound)
 	timeoutCh := subscribe(cs.eventBus, types.EventQueryTimeoutPropose)
+
+	// Set proposal function
 	cs.setProposal = func(proposal *types.Proposal) error {
 		if cs.Height == 2 && cs.Round == 0 {
-			// dont set the proposal in round 0 so we timeout and
```
melekes, line 86: why this was removed?

melekes (commented):
Thanks @faddat ❤️ Unsure how changing/adding comments fixes the test's flackiness though. Could you please explain?

internal/consensus/mempool_test.go
faddat: no good reason and hapy to restore it

faddat: it's back

faddat:
wrangling.... will mark this as draft.

faddat:
That's right draft is a good way for this one.

privval/socket_listeners_test.go
cason, line 1: Are these changes somehow related with the mempool test?

internal/consensus/mempool_test.go
```diff
@@ -90,21 +100,30 @@ func TestMempoolProgressInHigherRound(t *testing.T) {
 		}
 		return cs.defaultSetProposal(proposal)
 	}
+
+	// Start test round
+	height, round := cs.Height, cs.Round
```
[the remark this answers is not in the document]
cason: Don't see why. Height is always 1, round is always 0.

internal/consensus/mempool_test.go
```diff
@@ -90,21 +100,30 @@ func TestMempoolProgressInHigherRound(t *testing.T) {
...
 	startTestRound(cs, height, round)
 
-	ensureNewRound(newRoundCh, height, round) // first round at first height
-	ensureNewEventOnChannel(newBlockCh)       // first block gets committed
+	// Ensure new round at first height and block commit
+	ensureNewRound(newRoundCh, height, round)
```
cason, line 109: The problem here is that the new round event takes more than 200ms?

faddat: These are both good questions that I had missed. I will review this tomorrow and thank you for your review

faddat:
@cason I think we should defer further review here until https://github.com/cometbft/cometbft/pull/2021

is merged.

faddat:
@cason please see #1923 

I just added some commentary about another time-related test issue and I'm not sure how to handle these safely in all cases.

PR #1925, opened 2024-01-02
configure the gci linter

faddat:
- remove unused linters that produced a warning and ensure that unparam is enabled
- apply the linters to tests
- additional autofixes
- test helpers
- all helper functions start from t.Helper()
- mainly assert -> require and other hygene changes
- further updates and attempts at fixes
- further cleanups
- unused-recievers
- mainly unused recievers
- Fix import paths and receiver names in various files
- Fix various return statements in the code
- "save game"
- fix double declaration
- b.helper()
- use stdlib http methods
- Fix test assertions in various test files
- Fix validation test assertions
- Fix test assertions and clean up code
- Fix field element comparison in TestGenPrivKeySecp256k1
- Fix various test failures and errors in the code
- Fix assertion in TestClientReplacesPrimaryWithWitnessIfPrimaryIsUnavailable
- Fix error handling and add arguments to execAnsible function
- as a piece of Free software, cometbft does NOT need to support digitalocean.
- start over with internal
- comment-spacings
- redoing internal
- Revert "b.helper()"
- error handling
- reactor_test.go removed in master
- fix a bug from removing an error that is always nil
- use correct require calls
- begin reversing expected and equal values
- complete reversing expected and actual
- favor require.NoError over require.Nil
- fix issue with len in kvstore test
- use the correct numeral type in TestBlockStoreSaveLoadBlock
- Fix test assertions in rpc/jsonrpc/client/args_test.go, mempool/clist_mempool_test.go, crypto/merkle/proof_test.go, and internal/store/store_test.go
- Fix assertions in test cases
- Fix test assertions in rpc/client/helpers_test.go and types/block_test.go
- revert a store test fix that was causing issues
- Fix test assertions in state_test.go, cache_test.go, clist_mempool_test.go, helpers_test.go, and status_test.go
- assert.Gerater
- Fix assert statements in test files
- Fix assertion order in TestBlockFetchAtHeight
- fix regressions
- Fix test assertions in various test files
- Fix test assertions in detector_test.go, netaddress_test.go, addrbook_test.go, and switch_test.go
- Add gofmt linter and update assertions in tests
- Fix test helper functions
- Refactor code to use strconv.Itoa instead of fmt.Sprintf
- Fix assertion errors in unit tests
- Fix test assertions in types_test.go, secp256k1_internal_test.go, group_test.go, pool_test.go, verify_test.go, and state_test.go
- Fix test cases and update provider configuration
- Fix ValidateBasic assertion in http_test.go
- attempt to fix a regression
[...]

PR #1959, opened 2024-01-03
Disable undesired linting code

### LINKED ISSUE #1958: Disable undesired linting
Disable `lint-format` and `format` targets in the `Makefile`.

The below comment by Adi (and a bunch of PRs opened about linting) shows that this could be a useful note in the `Makefile`.
_Originally posted by @adizere in https://github.com/cometbft/cometbft/issues/1925#issuecomment-1875127862_

It looks like these targets are not used in our automation and their existence is misleading for newcomer developers too.

My proposal is to comment out the relevant targets with a note to Adi's comment and revisit linting after v1 was released.

greg-szabo:
Closes #1958 

This PR addresses two small bumps I have experienced while setting up my dev environment.

1. Linting
As it turns out this was extensively discussed in #1925 and other PRs. Some linting is configured in the `Makefile` but we do not make use of it because of the reasons [Adi mentioned](https://github.com/cometbft/cometbft/pull/1925#issuecomment-1875127862).

To make life easier for incoming developers, Adi's comment is added to the Makefile with a deadline and the (ir)relevant linting is commented out. I'm interested in the team's opinion on this kind of code commenting. I feel it would have saved me a ton of time during my initial setup searching for issues around it. (I didn't expect a bunch of files to be linted.)

2. Dependencies
The `Makefile` now makes a quick check if you have `codespell` installed for `lint-typo` as it's not part of the Go interpreter bundle. If it's not installed it will try to install it. This may still fail on some non-standard systems and it's up to the developer to find out what failed but it gives you a good indication of where to start.

I think this second aspect should have gone into a new PR but it was already made by the time I went through the linting problem. If the team wants, I can separate it.

---

#### PR checklist

- [X] Tests written/updated (I ran the `Makefile` locally and checked the workflows if formatting is ran.)


melekes (approved):

Makefile
```diff
@@ -222,26 +222,31 @@ clean_certs:
...
 	@go run github.com/golangci/golangci-lint/cmd/golangci-lint@latest run
 .PHONY: lint
 
-lint-format:
-	@go run github.com/golangci/golangci-lint/cmd/golangci-lint@latest run --fix
-	@go run mvdan.cc/gofumpt -l -w ./..
-.PHONY: lint-format
+# https://github.com/cometbft/cometbft/pull/1925#issuecomment-1875127862
+# Revisit using lint-format after CometBFT v1 release and/or after 2024-06-01.
+#lint-format:
```
melekes, line 239: I don't think we should comment these commands. The current `main` branch passes them.

faddat: I'm in agreement here, because I find that linting helps enormously with keeping code legible and frequently catches a surprising number of issues that are easy for humans to miss.

faddat:
@greg-szabo would this also mean that the "canonical / correct" way to run linting in cometbft is now `make lint`?

Also, codespell is a spell checker?

golangci-lint includes one of those, and afaik it is currently enabled.

greg-szabo:
Codespell is a spell checker, mainly for source code: https://github.com/codespell-project/codespell

I can't comment on the "canonical" way of linting in CometBFT as I'm just picking up the pieces. My current assumption is that whatever is in CI is canonical. I hope the CI uses the `Makefile`, but if not, then we may want to make a note of that. As a first-time developer, one would expect the `Makefile` to be at least in line with whatever is in CI.

I'll check the CI when I can. In this PR, I was addressing whatever I found so far.

I'd also like to propose linting in pre-commit hooks but the community might or might not be open to that. But that's for another discussion.

Issue #1964, opened 2024-01-05
perf: Blocksync is double calling types.Block.ValidateBasic at costs to sync speed


ValarDragon:
## Feature Request

Block sync is double-calling types.Block.ValidateBasic, which re-hashes every transaction in a block. In a 100 block CPU profile sync on Osmosis, over a period with especially large blocks, we found that this was the cause of some notable overhead. (18 seconds out of 185 seconds over all processes, at least 70s of which were in non-blocking operations) Under normal block sizes, this time is lower but still high/worth speeding up. (4.1 seconds for 100 blocks)

![image](https://github.com/cometbft/cometbft/assets/6440154/05295faf-c2cf-4dd6-a22c-b24320dde813)

We should not be double verifying this.

### Summary

We are double verifying block hashes in block sync by calling this in two different spots: 
[ValidateMsg here](https://github.com/cometbft/cometbft/blob/main/internal/blocksync/reactor.go#L233-L237), which calls types.BlockFromProto which calls validate basic, and [types.BlockFromProto right here](https://github.com/cometbft/cometbft/blob/main/internal/blocksync/reactor.go#L245)

### Problem Definition

Speed up block sync

### Proposal

Do some mild API or order of operations changes to reuse the `types.BlockFromProto` result and only compute it once.

faddat:
gonna check this out.

faddat:
Got it :).

I think that this will make your flame graph much happier.

melekes:
Thanks @ValarDragon ❤️ Definitely something we should fix

PR #1954, opened 2024-01-02
fix: test reliability patches

faddat:
## This PR fixes 

* flakiness in testwriter
* flakiness in testbyzantineprevoteequivocation

@melekes - I'd like to merge this branch, and then open another torture branch and run the tests a few thousand times.  But I think we are in good shape. 

----

## NOTES FOR REVIEWERS

Some of the fixes here involved small changes to the tolerances of tests in situations where the code may behave in a slightly variable way without malfunctioning.  I bet that is the key item for review. 

This needs to be broken into pieces.

~~I consider this branch to be mergable now~~, and I suppose what I'll do is make another one where the other tests I identified as having issues are run 100 times.  That was totally key to fixing `TestTransportMultiplexAcceptNonBlocking` because it got me the data I needed to understand the problem.

---

#### PR checklist

- [x] Tests written/updated
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments

PR #1970, opened 2024-01-05
perf: improve blocksync performance

faddat:
Closes: #1964

This PR:

* ensures we only call validatebasic once
* continues to process blocks even if a peer sends us something that's incorrect
* makes info level logging print every 100 blocks
* ensures that debug level logging prints every block

---

#### PR checklist

- [ ] Tests written/updated


faddat:
Once we get #1954 merged, ci won't be a lotto.

melekes (changes requested):
I don't understand how changes in this PR improve `blocksync` performance. @faddat could you please explain?

faddat:
@melekes -- I think I made a code management mistake here, and that..... I did write such changes but they are not here.

faddat:
(or worse, I may have actually reverted them)

faddat:
@melekes - pretty sure that I somehow conflated this and #1977 

I think I can get it sorted tonight.

Thanks for the review!

faddat:
closing in favor of #2016

Issue #49, opened 2022-12-23
Add in-process compaction support to LevelDB


lasarojc:
Was tendermint/tendermint#9743

## Summary
Experiment with adding in-proces compaction, so that nodes don't need to be stopped to perform compaction.

### Background
One of the most common problem that operators signal is that storage growth is unbounded and compaction doesn't work. Some operators stop their node, trigger `experimental-compact-goleveldb` (#8564) which deletes old data, and then restart their node.

### Why do we need this feature?
The use of command `experimental-compact-goleveldb` has the disadvantage that while this is running the node is stopped and is missing blocks. It typically take on the order of tens of minutes to finish compaction of a node on a production network, so the number of missed blocks can be significant.

## Proposal
We'll go about this incrementally
* [ ]  Tendermint team does initial de-risking and sanity checks to see that in-process compaction can be implemented safely
  * Add a new database type that does compaction
* [ ]  We ask an operator to deploy an early experiment replacing one of their full nodes with the patched tendermint version that has in-process compaction   
  * relayer team tests relaying against that node, monitor general health
* [ ]  We collect advanced metrics on latency in particular, as well as storage growth evolution  
  * [consensus: additional timing metrics  #9733](https://github.com/tendermint/tendermint/pull/9733)

jmalicevic:
This should apply the results done by Anton Paramonov in https://github.com/cometbft/cometbft/blob/anton/db_measurements/tools/goleveldb_perf/GoLevelDBCompaction.ipynb

melekes:
https://github.com/ethereum/go-ethereum/blob/4d086430bd8c0b9549fde5259f74f6df6c96df65/ethdb/leveldb/leveldb.go#L236-L245

This is the function for in-process compaction.

jmalicevic:
Closed with https://github.com/cometbft/cometbft/pull/1972 and https://github.com/cometbft/cometbft-db/pull/111

PR #1972, opened 2024-01-05
pruning: trigger explicitly compaction upon pruning

jmalicevic:
Blocked on https://github.com/cometbft/cometbft-db/pull/111  and benchmarking

Addresses #49 

Upon pruning we explicitly call the compaction function of the DB backend. This has shown to immediately shrink the storage footprint. 

We need to evaluate the duration of this compaction depending on the storage size to be able to reason about the impact on Comet's regular operations. 

ToDo: 
-Extend the `storage` config section with following parameters:
    - [ ] `in-process-compaction = false #Enable or disable in-process compaction. False by default`
    - [ ] `in-process-compaction-interval = 10 #Interval in number of blocks to trigger explicit compaction; 10 by default`

---

#### PR checklist

- [ ] Tests written/updated


melekes:
1. What are the downsides of having in-process compaction enabled by default?
2. What is the average compaction duration? Can we compact every block? That way we won't even need the `in-process-compaction-interval` config param.

jmalicevic:
> 1. What are the downsides of having in-process compaction enabled by default?

Something that I hope to establish during our benchmarking. This is a background process, and it should have minimal impact, especially if it happens frequently enough. But compaction also requires allocating in-memory buffers and it might have an impact on the performance. 

> 2. What is the average compaction duration? Can we compact every block? That way we won't even need the `in-process-compaction-interval` config param.

I think adding this parameter in any case makes sense because people might have different use cases. The duration of compaction is also something we will measure before merging this PR. But it does depend on the amount of data compacted, files moved around etc. We did once a benchmark to measure the duration of `PruneBlocks` and usually it is low but if we prune a lot of blocks at once it would go up to 200 or even 900ms. Note though that this does not necessarily mean the blocks were actually compacted, rather just marked as deleted.

config/config.go
```diff
@@ -1272,7 +1272,11 @@ type StorageConfig struct {
 	DiscardABCIResponses bool `mapstructure:"discard_abci_responses"`
 	// Configuration related to storage pruning.
 	Pruning *PruningConfig `mapstructure:"pruning"`
-
+	// Compaction config
+	CompactOnPruning bool `mapstructure:"compact_on_pruning"`
+	// Compaction interval - number of blocks to try explicit compaciton on
```
andynog, line 1277:
```suggestion
	// Compaction interval - number of blocks to try explicit compaction on
```

config/config.go
```diff
@@ -1272,7 +1272,11 @@ type StorageConfig struct {
 	DiscardABCIResponses bool `mapstructure:"discard_abci_responses"`
 	// Configuration related to storage pruning.
 	Pruning *PruningConfig `mapstructure:"pruning"`
-
+	// Compaction config
```
andynog, line 1275:
```suggestion
	// Compaction on pruning - enable or disable in-process compaction
	// false by default (compaction on pruning is disabled)
```

config/toml.go
```diff
@@ -583,6 +583,12 @@ initial_block_retain_height = {{ .Storage.Pruning.DataCompanion.InitialBlockReta
 # already set a block results retain height, this is ignored.
 initial_block_results_retain_height = {{ .Storage.Pruning.DataCompanion.InitialBlockResultsRetainHeight }}
 
+# If set to true, CometBFT will force compaction to happen for databases that support this feature
+compact_on_pruning = true
```
andynog, line 587: From the PR description it says it's `false` by default. Just want to confirm if `true` should be here. If true, then my comment above needs to be updated to say default is `false`

config/toml.go
```diff
@@ -583,6 +583,12 @@ initial_block_retain_height = {{ .Storage.Pruning.DataCompanion.InitialBlockReta
...
+
+# To avoid forcing compaction every time, this parameter instructs Comet to wait
```
andynog, line 589:
```suggestion
# To avoid forcing compaction every time, this parameter instructs CometBFT to wait 
```

go.mod
```diff
@@ -39,7 +39,7 @@ require (
 	github.com/Masterminds/semver/v3 v3.2.1
 	github.com/btcsuite/btcd/btcec/v2 v2.3.2
 	github.com/btcsuite/btcd/btcutil v1.1.5
-	github.com/cometbft/cometbft-db v0.10.0
+	github.com/cometbft/cometbft-db v0.11.0
```
andynog, line 42: Do we need any not in the `CHANGELOG` that we are upgrading this ? I think this is just the compaction changes for the DB and not breaking as far as I understand

internal/state/store.go
```diff
@@ -420,6 +427,14 @@ func (store dbStore) PruneStates(from int64, to int64, evidenceThresholdHeight i
 		return err
 	}
 
+	// We do not want to panic or interrupt consensus on compaction failure
+	if store.StoreOptions.Compact && (to%store.StoreOptions.CompactionInterval == 0 || pruned >= uint64(store.StoreOptions.CompactionInterval)) {
+		err = store.db.Compact(nil, nil)
+		if err != nil && store.StoreOptions.Logger != nil {
+			store.StoreOptions.Logger.Error("error compacting state store ", "err", err)
```
andynog, line 434: I think it should be OK to return an error here, because this will bubble up to `pruneBlocksToRetainHeight` in the `pruner.go` and there an error is logged too afaiu.

internal/state/store.go
```diff
@@ -420,6 +427,14 @@ func (store dbStore) PruneStates(from int64, to int64, evidenceThresholdHeight i
 		return err
 	}
 
+	// We do not want to panic or interrupt consensus on compaction failure
+	if store.StoreOptions.Compact && (to%store.StoreOptions.CompactionInterval == 0 || pruned >= uint64(store.StoreOptions.CompactionInterval)) {
```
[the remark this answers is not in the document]
andynog: I'm trying to understand this second part of the `if`. In `to%store.StoreOptions.CompactionInterval == 0` is this a good condition ? I'm thinking that let's say `to = 100` but the `CompactionInterval = 10` then a compaction would happen, the question is, does the `to` increases sequently or it could skip (e.g. 100, 105, etc) ? If the latter then the compaction might not happen in the desired interval.

internal/store/store.go
```diff
@@ -55,17 +56,46 @@ type BlockStore struct {
...
+
+// WithCompaction sets the compaciton parameters.
+func WithCompaction(compact bool, compactionInterval int64) BlockStoreOption {
+	return func(bs *BlockStore) {
+		bs.compact = compact
+		bs.compactionInterval = compactionInterval
+	}
+}
+
+// WithLogger sets the logger for the blockstore.
```
andynog, line 77:
```suggestion
// WithLogger sets the logger for the BlockStore.
```
just to make it consistent with other comments

internal/store/store.go
```diff
@@ -390,6 +420,17 @@ func (bs *BlockStore) PruneBlocks(height int64, state sm.State) (uint64, int64,
...
 		return 0, -1, err
 	}
+	bs.blocksDeleted += int64(pruned)
+
+	if bs.compact && bs.blocksDeleted >= bs.compactionInterval {
+		// Error on compaction should not be reason to halt the chain
+		// TODO add logger to keep track of error
+		err = bs.db.Compact(nil, nil)
+		if err != nil && bs.logger != nil {
+			bs.logger.Error("Error compacting block store", "err", err)
```
andynog, line 430: same comment as above, returning an error here I don't think will cause any issues because it's just logged

internal/store/store.go
```diff
@@ -390,6 +420,17 @@ func (bs *BlockStore) PruneBlocks(height int64, state sm.State) (uint64, int64,
...
+		}
+		bs.blocksDeleted = 0
```
andynog, line 432: what is the reason to set `bs.blocksDeleted` to `0` here after compaction ? Is this a statement saying that compaction worked (didn't return an error) ? 

So if you store this value in the block store for future reference on another prune round, does this line `bs.blocksDeleted += int64(pruned)` makes sense ? If compaction is working and you reset `blocksDeleted` to 0, in your condition `if bs.compact && bs.blocksDeleted >= bs.compactionInterval` this will only trigger when there are more blocks pruned than the interval. I am thinking you want to keep track of pruned blocks in more than one pruning iteration no ?

andynog (approved):
lgtm, I've added a few comments related to the logic and also suggestions for renaming some text in comments

config/config.go
```diff
@@ -1272,7 +1272,11 @@ type StorageConfig struct {
...
+	CompactOnPruning bool `mapstructure:"compact_on_pruning"`
```
melekes, line 1276:
```suggestion
	Compact bool `mapstructure:"compact"`
```

Do we have to tie compaction to pruning?

go.mod
melekes: > I think this is just the compaction changes for the DB and not breaking as far as I understand

correct

internal/state/store.go
```diff
@@ -420,6 +427,14 @@ func (store dbStore) PruneStates(from int64, to int64, evidenceThresholdHeight i
...
+		err = store.db.Compact(nil, nil)
+		if err != nil && store.StoreOptions.Logger != nil {
```
melekes, line 433: I think you can set Logger to `NopLogger` and remove the 2nd condition

internal/store/store.go
```diff
@@ -390,6 +420,17 @@ func (bs *BlockStore) PruneBlocks(height int64, state sm.State) (uint64, int64,
 	if err != nil {
 		return 0, -1, err
 	}
+	bs.blocksDeleted += int64(pruned)
+
+	if bs.compact && bs.blocksDeleted >= bs.compactionInterval {
+		// Error on compaction should not be reason to halt the chain
+		// TODO add logger to keep track of error
+		err = bs.db.Compact(nil, nil)
+		if err != nil && bs.logger != nil {
```
melekes, line 429: same comment about logger

internal/state/store.go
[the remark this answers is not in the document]
melekes: But `pruned` is never reset to 0

melekes (commented):
👍

internal/state/store.go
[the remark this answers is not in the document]
jmalicevic: @andynog  Yes you are right, if `to` is not a multiple of `CompactionInterval` then we have a problem. I will see how to change it. 

> But pruned is never reset to 0

It is set at the beginning of the function every time we call it.

internal/state/store.go
jmalicevic: Ah yes you are right. This was done based on the 0.37 backport where an error here actually caused a panic eventually. This can be indeed returned then.

internal/store/store.go
jmalicevic: What Andy pointed out is that these errors won't cause a panic in the end. I was tricked by the way this is done in 0.37 where it would panic and I don't think a compaction failure should be cause for a panic. I think we should remove the loggers alltogether given this?

internal/store/store.go
jmalicevic: It will be reset to 0 only if we have pruned enough blocks that we can trigger compaction and then we can start counting again. Otherwise the value is preserved accross calls. `bs` is a pointer so the value remains, unlike with the state store where it would always be reset to 0 so I removed it.

config/config.go
jmalicevic: I did experiments and if we do not prune, compaction makes no sense. It makes sense to do it only if we actually delete stuff.

config/toml.go
jmalicevic: Should be false, will edit.

config/config.go
melekes: What if we take into account https://github.com/cometbft/cometbft/pull/2017?

internal/store/store.go
melekes: > I think we should remove the loggers alltogether given this?

Agree

config/config.go
jmalicevic: I see.. We could then rename compact to be just what you are suggesting but in that context, should CompactionInterval be applied to the ABCI results the same way as it is for blocks or would you call the `db.Compact` function from #2017 every time you delete? Note that in that case, if the pruner is pruning and you prune here as well we might trigger compaction more frequently on the state store.

melekes: What's the overhead of calling pruning every block?

jmalicevic: Depends on the size of the DB. If pruning is enabled and state is pruned frequently and the DB is not large it is not that big. But if you prune rarely, or not at all but you just want to compact the DB because you want to delete these ABCI results then it will traverse all the levels of the DB on every call.

melekes: > should CompactionInterval be applied to the ABCI results the same way as it is for blocks or would you call the db.Compact function from https://github.com/cometbft/cometbft/pull/2017 every time you delete?

Then I would say the former (i.e. the same way it is for blocks).

internal/store/store.go
andynog: OK, in this case if you decide to refactor the code above and return an error instead of just logging, then you will also need to modify this logic a bit to account for the early return.

jmalicevic: I will reset this only if there is no error in compaction. Otherwise it stays the same.

internal/state/store.go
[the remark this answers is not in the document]
jmalicevic: Changed it but had to adapt the interface of the PruneStates function .

config/config.go
jmalicevic: Do you think it would then make sense, once this is merged to add a call to compact ABCI results from #2017? I am wondering whether we should then not call `Compact` within the `PruneStates` and `PruneBlocks` function but rather extend the interfaces of the state and block store to include a `Compact` function. Then in the pruner service , if the retain heights changed call Compact. If they have not changed but 

``` 
if config.Compact && discardABCIResults = false  && height % CompactionInterval == 0  {

   stateStore.Compact(nil,nil)
   
}
```

Before making this change we just need to make sure this is not overkill for achieving the effect in 2017? If it makes more sense from a design perspective then I am happy to do this change.

.changelog/unreleased/breaking-changes/1972-store-pruning-api-breaking.md
```diff
@@ -0,0 +1 @@
+- `[state/store]` go-API breaking change in `PruneStates`: added parameter to pass number of pruned states and return pruned entries in current pruning iteration.. ([\#1972](https://github.com/cometbft/cometbft/pull/1972))
```
melekes, line 1:
```suggestion
- `[state/store]` go-API breaking change in `PruneStates`: added parameter to pass the number of pruned states and return pruned entries in current pruning iteration. ([\#1972](https://github.com/cometbft/cometbft/pull/1972))
```

config/config.go
melekes: > Then in the pruner service , if the retain heights changed call Compact. If they have not changed but

This makes sense, but let's delay this decision until #2017.

I'm in favor of renaming `CompactOnPruning` to just `Compact`, though.

jmalicevic: Ok yes I am totally fine with this. Then lets try and merge this logic as is but make sure we finalize the changes before v1 as they are potentially breaking afterwards.  I'll edit the name of the config variable. I realized there is a potential corner case where the data companion is enabled but the state and block stores are never pruned because the app does not allow it. The data companion can still pruned the ABCI results and indexer and we should explicitly call compact in those cases (not relying that state and block stores will be pruned). Will push that change in a min

config/toml.go
```diff
@@ -583,6 +583,12 @@ initial_block_retain_height = {{ .Storage.Pruning.DataCompanion.InitialBlockReta
 # already set a block results retain height, this is ignored.
 initial_block_results_retain_height = {{ .Storage.Pruning.DataCompanion.InitialBlockResultsRetainHeight }}
 
+# If set to true, CometBFT will force compaction to happen for databases that support this feature
+compact_on_pruning = false
```
melekes, line 587:
```suggestion
compact = false
```

melekes (approved):
👍

PR #1973, opened 2024-01-05
PBTS: migrating to `main` the new version of the specification

cason:
This PR cherry-picks a number of commits present in the `master` branch of this repository.

Some of the commits are from the old `tendermint/spec` repository, namely:

- tendermint/spec#369
- tendermint/spec#373
- tendermint/spec#375
- tendermint/spec#398
- tendermint/spec#399

This content has then been migrated to the `tendermint/tendermint` repository, from which follow PRs were considered:

- tendermint/tendermint#7804 (**not cherry-picked**)
- tendermint/tendermint#8018
- tendermint/tendermint#8096

Most of the conflicts and the fixes needed refer to links to content, that have been updated or referred to old repositories.

The MD files that are currently under `spec/consensus/proposer-based-timestamp` were moved to the `v1` subdir. Files with similar names, but the `-002` suffix were added with the new solution, replacing the old ones.

The differences from the original content, on `master` and the content of this PR are minimal, as one can check by running

` git diff master spec/consensus/proposer-based-timestamp/`

Finally, the entry point of the added content can be found [here](https://github.com/cometbft/cometbft/blob/cason/pbts-spec/spec/consensus/proposer-based-timestamp/README.md).


melekes (approved):

sergio-mena:
If we are to release PBTS as part of `v1`, we need to backport this to the `v1` branch. But let's not do it until we're sure PBTS will indeed land in `v1`. I'll add a task in the tracking issue

sergio-mena:
Regarding the comment above, now is the time 😄

PR #1974, opened 2024-01-05
feat(storage/metrics): Metrics to measure storage

### LINKED ISSUE #46: Establish and implement the relevant metrics to understand storage workloads

jmalicevic:
This PR superceeds #79 with some adjustments to the segments of code timed as well as bucket sizes. The majority of the code was done by William in #79.  I tried to fine tune the measurements to exclude proto marshalling/unmarshalling where I think it made sense. 

Closes #46 

Blocked on benchmarking to confirm it is measuring what we want.
  (follow up) The metrics gave us nice and realistic measurements in our benchmarks.

---

#### PR checklist

- [ ] Tests written/updated


cmd/cometbft/commands/inspect.go
```diff
@@ -58,7 +58,7 @@ func runInspect(cmd *cobra.Command, _ []string) error {
 	if err != nil {
 		return err
 	}
-	blockStore := store.NewBlockStore(blockStoreDB)
+	blockStore := store.NewBlockStore(blockStoreDB, store.BlockStoreOptions{Metrics: store.NopMetrics()})
```
melekes, line 61:
```suggestion
	blockStore := store.NewBlockStore(blockStoreDB, store.BlockStoreOptions{})
```

We can remove `NopMetrics` because it's a default

internal/state/store.go
```diff
@@ -130,12 +139,21 @@ func IsEmpty(store dbStore) (bool, error) {
 
 // NewStore creates the dbStore of the state pkg.
 func NewStore(db dbm.DB, options StoreOptions) Store {
```
melekes, line 141: What do you think about having two methods: `NewStore` and `NewStoreWithOptions`?

internal/store/store.go
```diff
@@ -56,14 +60,23 @@ type BlockStore struct {
...
 }
 
+type BlockStoreOptions struct {
+	Metrics *Metrics
+}
+
 // NewBlockStore returns a new BlockStore with the given DB,
 // initialized to the last height that was committed to the DB.
-func NewBlockStore(db dbm.DB) *BlockStore {
+func NewBlockStore(db dbm.DB, o BlockStoreOptions) *BlockStore {
```
melekes, line 69: Same. I'd propose adding a second constructor: `NewBlockStoreWithOptions`.

melekes (commented):
nice 👍

internal/state/store.go
jmalicevic: This interface was there from before so I would not necessarily change it in this PR although I think it is nicer to do it differently. I did it more elegantly for the blockstore now btw.

internal/store/store.go
jmalicevic: Done as an optional parameter to the constructor of the blockstore.

jmalicevic:
Do we maybe want documentation on the metrics, the methods they cover? At the moment there are two metrics :
- block_store_access_time
-state_store_access_time

But the time is measured individually per method accessing the storage. And some methods include others (`save_blocks` includes the time tow rite and marshall all block parts). So maybe documenting this for users might be benefitial?

internal/state/store.go
```diff
@@ -234,12 +258,17 @@ func (store dbStore) save(state State, key []byte) error {
 		state.LastHeightConsensusParamsChanged, state.ConsensusParams, batch); err != nil {
 		return err
 	}
-	if err := batch.Set(key, state.Bytes()); err != nil {
+	stateMarshallTime := time.Now()
+	stateBytes := state.Bytes()
```
greg-szabo, line 262: Why is the byte-conversion taken out of the measurement? Can we add (maybe above line 261) a comment with the reason?

internal/state/store.go
```diff
@@ -296,17 +326,19 @@ func (store dbStore) Bootstrap(state State) error {
...
+	defer addTimeSample(store.metrics.StoreAccessDurationSeconds.With("method", "prune_states"), time.Now())()
 	if from <= 0 || to <= 0 {
 		return fmt.Errorf("from height %v and to height %v must be greater than 0", from, to)
 	}
 	if from >= to {
 		return fmt.Errorf("from height %v must be lower than to height %v", from, to)
 	}
 
-	valInfo, err := loadValidatorsInfo(store.db, min(to, evidenceThresholdHeight))
+	valInfo, elapsedTime, err := loadValidatorsInfo(store.db, min(to, evidenceThresholdHeight))
```
greg-szabo, line 337: I'm not sure about the patterns used in our codebase, but it looks to me that `loadValidatorsInfo` doesn't benefit from `elapsedTime` at all and maybe the measurement should happen in this (`PruneStates`) function, not in the `loadValidatorsInfo`.

I can see that the unmarshaling of the data from the database is not measured, which gives us a more precise measurement for database activity. However, this is not indicated in `loadValidatorsInfo`. Maybe a solution is to document that the result is the elapsed time for _database access_ or something. But maybe we should move this measurement to `PruneStates` and accept that the unmarshaling is also part of the measurement.

internal/state/store.go
```diff
@@ -333,7 +365,8 @@ func (store dbStore) PruneStates(from int64, to int64, evidenceThresholdHeight i
 		// params, otherwise they will panic if they're retrieved directly (instead of
 		// indirectly via a LastHeightChanged pointer).
 		if keepVals[h] {
-			v, err := loadValidatorsInfo(store.db, h)
+			v, tmpTime, err := loadValidatorsInfo(store.db, h)
```
greg-szabo, line 368: Again, I lack some context, but do we allow variable shadowing? `err` is previously defined in the function.

internal/store/store.go
```diff
@@ -57,15 +60,33 @@ type BlockStore struct {
 	height int64
 }
 
+type BlockStoreOption func(*BlockStore)
+
+// WithMetrics sets the metrics.
+func WithMetrics(metrics *Metrics) BlockStoreOption {
```
greg-szabo, line 66: I like this way better than the `NewStore` function in StateStore. Maybe I'm too Rust-oriented.

greg-szabo (commented):
Great work here! I made a few comments in the StateStore. I kind of lost track in the middle of BlockStore so I'll have another go at it. Until then, I think some of the comments are worth a bit of discussion.

internal/state/store.go
greg-szabo: I agree that the `WithMetrics` solution is nicer. I also have to agree that we can't fix all the code in one go. But maybe it's worth opening a new issue on Anton's comment and we'll get back to it later.

andynog (commented):
lgtm, I've added a few comments but more like suggestions or questions to ensure nothing is breaking.

internal/store/store.go
```diff
@@ -57,15 +60,33 @@ type BlockStore struct {
...
+	return func(bs *BlockStore) { bs.metrics = metrics }
+}
+
 // NewBlockStore returns a new BlockStore with the given DB,
 // initialized to the last height that was committed to the DB.
-func NewBlockStore(db dbm.DB) *BlockStore {
+func NewBlockStore(db dbm.DB, options ...BlockStoreOption) *BlockStore {
```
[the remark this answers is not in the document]
jmalicevic: The signature change is indeed API breaking but the function can still be called without any option. Do you think we should still consider it breaking in this case?

internal/state/store.go
jmalicevic: Will do, is that ok with you @melekes ?

internal/state/store.go
jmalicevic: The point was to avoid measuring marshalling/unmarshalling to make sure the measurements include only DB access time. You are however right that the function `loadValidatorsInfo` itself will not report this. The function has no object that has metrics associated with it to enter the value. That is why I report the time spent in this function only within functions that call it.

internal/state/store.go
```diff
@@ -931,3 +982,11 @@ func int64ToBytes(i int64) []byte {
 	n := binary.PutVarint(buf, i)
 	return buf[:n]
 }
+
+// addTimeSample returns a function that, when called, adds an observation to m.
+// The observation added to m is the number of seconds elapsed since addTimeSample
+// was initially called. addTimeSample is meant to be called in a defer to calculate
+// the amount of time a function takes to complete.
+func addTimeSample(m metrics.Histogram, start time.Time) func() {
+	return func() { m.Observe(time.Since(start).Seconds()) }
```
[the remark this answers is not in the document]
andynog: Yes, it address. Please disregard for now. Changing this would require many other changes and since `Seconds()` return a `flota64` it's all good. Thanks !

internal/store/store.go
```diff
@@ -698,3 +786,9 @@ func (bs *BlockStore) DeleteLatestBlock() error {
 	bs.height = targetHeight - 1
 	return bs.saveStateAndWriteDB(batch, "failed to delete the latest block")
 }
+
+func addTimeSample(h metrics.Histogram, start time.Time) func() {
```
melekes, line 790:
```suggestion
// addTimeSample returns a function that, when called, adds an observation to m.
// The observation added to m is the number of seconds elapsed since addTimeSample
// was initially called. addTimeSample is meant to be called in a defer to calculate
// the amount of time a function takes to complete.
func addTimeSample(h metrics.Histogram, start time.Time) func() {
```

internal/state/store.go
```diff
@@ -131,12 +139,21 @@ func IsEmpty(store dbStore) (bool, error) {
 
 // NewStore creates the dbStore of the state pkg.
 func NewStore(db dbm.DB, options StoreOptions) Store {
-	return dbStore{db, options}
+	m := NopMetrics()
+	if options.Metrics != nil {
```
melekes, line 143: why do we need the 2nd pointer if we can have a private method `func (store dbStore) metrics { return store.StoreOptions.Metrics }` ?

jmalicevic: Removed!

internal/state/store.go
jmalicevic: So this was here from before, I don't think we have explicit rules but given that on any error before or after we return immediately it can pass here.. It would be messier to declare the vars above just to avoid overwriting this one.

.changelog/unreleased/features/1974-statestore-metrics.md
```diff
@@ -0,0 +1 @@
+- `[state/store]` Added metrics to monitor state store access. ([\#1974](https://github.com/cometbft/cometbft/pull/1974))
```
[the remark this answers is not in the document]
jmalicevic: They are in different packages but if you think its better to merge this into one changelog entry I don't particularly mind.

melekes (approved):

internal/state/store.go
andynog: Agree with @jmalicevic, if the goal is to measure access time, then any side effects or processing that is not related to the IO should not be considered (or eliminated as much as possible)

internal/store/store.go
andynog: Not Rust oriented, we have this Builder pattern in several places in the codebase including the RPC APIs, I think it's a good pattern too and it helps with unit testing.

PR #1975, opened 2024-01-05
Update localnet-start Makefile

greg-szabo:
This is yet another small infrastructure-related fix that does not impact current developers but makes life easier for newcomers. It is not part of the mainstream source code, so I didn't add it to unclog.

Situation:
- I'm building on an Apple M1 computer (darwin/arm64).
- I want to set up a localnet using the provided scripts (`make localnet-start`)
  - I cross-compile the binary to linux/amd64 as the docker image instructs.
  - The `localnode` docker container still fails.

Problem:
The `localnode` docker container is built using the host's architecture but expects x86_64 libraries within. This works perfectly fine on an older Intel Mac or a Linux host. But in my case Apple M1 is an arm64 architecture and localnode is built with that in mind. Then it expects to run x86_64 binaries.

Resolution:
Even though in 99% of the cases this is redundant, we should explicitly tell the docker build process that we want an x86_64 container. This can be done with the `--platform` parameter when using the `buildx` build system.

Fallout:
- No impact for current users.
- Apple M1 users can run `localnet-start`
- Old docker users will have to upgrade their engine to v19 or above. (v19 is three years old so I expect everyone has buildx already.)

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/networks/local/Makefile
+++ b/networks/local/Makefile
@@ -1,7 +1,7 @@
 # Makefile for the "localnode" docker image.
 
 all:
-	docker build --tag cometbft/localnode localnode
+	docker buildx build --platform linux/amd64 --tag cometbft/localnode localnode
 
 .PHONY: all
 


faddat:
The team Jacob unclog on newcomers so you should add it to unc

melekes (approved):

Issue #1976, opened 2024-01-06
perf: De-duplicate writes in SaveABCIResponses / SaveFinalizeBlockResponse


ValarDragon:
## Feature Request

SaveABCIResponses is called on every block during consensus execution + block sync, and is [blocking consensus](https://github.com/cometbft/cometbft/blob/main/internal/state/execution.go#L253-L256) from proceeding. 

We see that for nodes which want to serve the `block_results` query, we pay a 2x write amplification over the data we want to serve.

The 2x write amplification in SaveABCIResponses can get extremely expensive for nodes, when there are many events. This occurs during a current problematic mainnet scenario detailed below, that currently affects all cosmos chains. Would be great if we can get a reduction at the comet layer soon for its part of the slowdown in this process!

### Summary

Over a 100 block sync, where only a fraction of blocks have spam txs as detailed below, we see SaveABCIResponses take a large amount of time. (10 seconds! The state machine exec + commit time for this is 55 seconds. For a more "normal" 100 block range, we see this take .23 seconds)
![image](https://github.com/cometbft/cometbft/assets/6440154/16caa412-59ef-4105-828c-b09d36bd2861)

### Problem Definition / Context

When people send IBC packets with large amounts of data (400kb), nodes begin falling behind. (This is the "bananaking" attacks)
The biggest culprit of problems for nodes falling behind comes from SaveABCIResponse where we suffer a 2x write overhead: https://github.com/cometbft/cometbft/blob/main/internal/state/store.go#L561-L601 . We see nodes having to write an extra 100mb/s when these txs come in. We also see 10+ of these getting into a block. (Which matches the expected 100MB/s overhead being measured)

During this time, we observe all RPC nodes have a very hard time keeping up, and 

There are other issues that go into having problems here, but this particular comet location definitely seems like a strong mitigation here going forward. The fact this msg, if `x` bytes, causes a `13x` increase in event size relative to bandwidth used leading to a `26x` write overhead in Comet here. CPU overheads in the SDK + IBC + comet call stack are present for this pattern. The fact that gas per byte is underpriced. CometBFT block gossip also gets quite expensive on big blocks.

Without knowing this can become a system bottleneck, the reason for the 2x write overhead is quite reasonable imo!

### Proposal

Notice that in https://github.com/cometbft/cometbft/blob/main/internal/state/store.go#L561-L601 we are setting the same data twice.

[...]

PR #1977, opened 2024-01-06
perf: enhance performance in state

faddat:
Working on #1976 and also examining other ways to improve performance. 

closes: #1976 

---

#### PR checklist

- [x] Tests written/updated


faddat:
Needs #1954

faddat:
@ValarDragon - I don't think that this is quite done correctly yet, but could you take a peek?

faddat:
oh I def found a bug.... it is strange that tests pass

faddat:
I was afraid of that...

faddat:
@ValarDragon @melekes - what would be a good way to benchmark here?

And what should be "the standard"?

faddat:
This isn't the same thing reported by @ValarDragon but I think it's related:

https://github.com/cometbft/cometbft-db/pull/112#issuecomment-1883788848

internal/state/store.go
```diff
@@ -131,12 +139,26 @@ func IsEmpty(store dbStore) (bool, error) {
 
 // NewStore creates the dbStore of the state pkg.
 func NewStore(db dbm.DB, options StoreOptions) Store {
-	return dbStore{db, options}
+	return &dbStore{
+		db:              db,
+		valInfoCache:    make(map[int64]*cmtstate.ValidatorsInfo),
```
melekes, line 144: Why do we need these? I don't see them used anywhere in this PR

melekes (commented):
While there are some nice improvements here, this PR fails to address the main issue described in https://github.com/cometbft/cometbft/issues/1976

internal/state/store.go
faddat: I believe that I made a branch management mistake here.  Thanks for review.

internal/state/store.go
```diff
@@ -403,20 +426,19 @@ func (store dbStore) PruneStates(from int64, to int64, evidenceThresholdHeight i
 		}
 		pruned++
 
-		// avoid batches growing too large by flushing to database regularly
-		if pruned%1000 == 0 && pruned > 0 {
-			err := batch.Write()
-			if err != nil {
+		// Commit the batch after a certain number of operations to avoid large batches
+		if pruned%batchSize == 0 {
+			if err := batch.WriteSync(); err != nil {
```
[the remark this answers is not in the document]
faddat: let me see if batch.Write() works just as well.

faddat:
@melekes I will do more here, once 

* https://github.com/cometbft/cometbft/pull/2021

has been merged.

internal/state/store.go
faddat: Hey, I'm wrong.  Basically one finding was that it was faster to use pointers and lay things out first.  That's why that's there.

faddat:
I'm opening this for review again -- close if anyone thinks is not needed, it seems like I went outside of the original intended scope.

faddat:
Hey I'm going to close this with the note to self that we should have more benchmarks across the code base

PR #1978, opened 2024-01-06
feat: app hash error channel 

czarcas7ic:
This PR implements an error channel which allows cometBFT to stream app hash errors to the sdk. With this information, the sdk can provide a more useful error message, specifically the hash of every module at the problematic block height.

Screenshot 2024-01-02 at 9 58 28 PM

The respective cosmos sdk PR: https://github.com/osmosis-labs/cosmos-sdk/pull/500

Osmosis will be using this fork in our next release. Would be great to have this upstreamed here!

### DIFF
--- a/internal/blocksync/reactor.go
+++ b/internal/blocksync/reactor.go
@@ -3,6 +3,7 @@ package blocksync
 import (
 	"fmt"
 	"reflect"
+	"strings"
 	"time"
 
 	bcproto "github.com/cometbft/cometbft/api/cometbft/blocksync/v1"
@@ -61,8 +62,9 @@ type Reactor struct {
 	pool      *BlockPool
 	blockSync bool
 
-	requestsCh <-chan BlockRequest
-	errorsCh   <-chan peerError
+	requestsCh      <-chan BlockRequest
+	errorsCh        <-chan peerError
+	appHashErrorsCh chan p2p.AppHashError
 
 	switchToConsensusMs int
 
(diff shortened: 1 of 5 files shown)


faddat (approved):

cason:
Not sure if adding a new method for _every_ reactor is the solution for retrieving some messages from the blockchain reactor. Moreover, a blocking channel in a for loop is not really a great approach.

sergio-mena:
All interactions between the application and CometBFT MUST happen via ABCI. Moreover, any solution we come up with, should work with all possible ABCI clients: local, TCP, and gRPC. The latter two are thought for applications that are not written in golang.

ValarDragon:
The context for why we forked to include this right now, was because there have been node stability problems for months with folks getting random app hashes due to some issue in a query somewhere, and we've been massively struggling to get proper info to debug them. (It affects several cosmos chains, not just osmosis)

So we wanted to add some basic logging app side (block results, module app hashes, we'd eventually do logging the gas steps in a tx, etc.) so we can start finding the source of this issue.

Definitely don't see us being on a long term fork for this, but we did need some more debug info on the default binary we give nodes for now!

sergio-mena:
Thanks @ValarDragon for the context. This is what I propose to move this forward:

* Please create an issue on this repo explaining the problem you are facing (basically an extended version of your comment [here](https://github.com/cometbft/cometbft/pull/1978#issuecomment-1883896416))
* You can reference this PR in that issue to provide context: "This is how Osmosis dealt with the problem temporarily"
* Then we can hook that issue into the ABCI 3.0 list in #9. We are planning to revive the ABCI bi-weeklies shortly, where we will discuss holistic solutions for all items in that list (which will include this one if we hook the corresponding issue)

adizere:
Feature development tracked in #2051

PR #1983, opened 2024-01-07
refactor: optimize the order of fields in each struct

kokeshiM0chi:
## Optimize the order of fields in each struct

Memory size is saved by optimizing the order of fields in a struct.

## Reproduction steps: How were structures discovered and optimized?

I applied field alignment tool.

```
go install golang.org/x/tools/go/analysis/passes/fieldalignment/cmd/fieldalignment@latest
```

```
find -E ./ -type f -iregex ".+\.go" | grep -v "pb.go" | xargs -n1 fieldalignment -fix
```

## For example 

Take abci/client as an example.  (16 + 24 + 24 + 8 + 8 = ) 80 bytes could be saved.

```
cd abci/client && fieldalignment .
```

```
cometbft/abci/client/client.go:61:13: struct with 48 pointer bytes could be 32
cometbft/abci/client/grpc_client.go:21:17: struct with 152 pointer bytes could be 128
cometbft/abci/client/socket_client.go:31:19: struct with 160 pointer bytes could be 136
cometbft/abci/client/unsync_local_client.go:11:24: struct with 96 pointer bytes could be 88
cometbft/abci/client/socket_client_test.go:201:29: struct of size 16 could be 8
```

---

#### PR checklist

- [ ] Tests written/updated (not needed)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments (not needed)


kokeshiM0chi:
The code comment has disappeared, so close it once. sorry.

PR #1986, opened 2024-01-07
No need to support ancient go's

faddat:
There is no need to support ancient versions of go.

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

melekes (approved):
👍

PR #1995, opened 2024-01-08
feat: conventional commits

melekes:
Check PR title follows the Conventional Commits spec.

---

#### PR checklist


CONTRIBUTING.md
```diff
@@ -501,22 +501,25 @@ Before merging a pull request:
...
 
 [potentially longer description in the body]
 
 Fixes #nnnn
 ```
 
-Each PR should have one commit once it lands on `main`; this can be accomplished
-by using the "squash and merge" button on GitHub. Be sure to edit your commit
-message, though!
-
```
sergio-mena, line 519: I think this still makes sense (except for PR that represent a merge from `main` into a feature branch)

mzabaluev:
As this PR suggests adding and enforcing another discipline to our workflow that is bound to add some overhead, let's not skip the usual requirements we have for PRs. I'd like to have these points discussed:

* What are the improvements and problems that this convention addresses?
* Does this overlap with our existing `.changelog` process of recording changes?
* If this is going to be some "source of truth" for analyzing change history, are there ways to make corrections that do not involve force-pushing to e.g. `main`?

sergio-mena:
@mzabaluev makes sense. Can we add it to next Monday's agenda?
@melekes can we hold on merging this PR until the discussion takes place?

CONTRIBUTING.md
mzabaluev: This concerns me as well: does this mean we change from squash-and-merge to preserving commit history in PRs, and therefore imposing more discipline on developers' branches? I'm not a fan of squash-and-merge myself, but I feel this needs some buy-in from the team.

This `!` thing may be particularly vexing: do we want to be forced to think of which changes might be breaking, when making every single commit (or, perhaps, when grooming commits pre-merge)?

mzabaluev (dismissed):
As discussed in earlier comments, let's go through the suggested process changes as a team first.

.github/workflows/conventional-pr-title.yml
```diff
@@ -0,0 +1,17 @@
...
+      - reopened
+      - edited
+      - synchronize
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    permissions:
+      statuses: write
+    steps:
+      - uses: aslafy-z/conventional-pr-title-action@v3
```
mzabaluev, line 15: This checks that the PR metadata conforms to the conventions, not the commits in the branch, correct?

I think this granularity of enforcement will be good, if we keep the squash-and-merge method of merging PRs.

faddat:
I think this is a really good idea.

I'm not really a fan of unclog.

.github/workflows/conventional-pr-title.yml
melekes: > not the commits in the branch, correct?

correct ✅

melekes:
> @melekes can we hold on merging this PR until the discussion takes place?

sure 👌

melekes:
> What are the improvements and problems that this convention addresses?

Lack of structure in commits that land on the `main` branch.

> Does this overlap with our existing .changelog process of recording changes?

Short answer: no. Long answer: we could replace `unclog` with something like https://github.com/conventional-changelog/conventional-changelog if we want in the future, which generates a changelog from the Git history. But this is not why I created this PR.

> If this is going to be some "source of truth" for analyzing change history, are there ways to make corrections that do not involve force-pushing to e.g. main?

Do you mean that you can always edit the `unclog` entry prior to release but can't edit the commit message in the `main` branch? This would be one reason not to use https://github.com/conventional-changelog/conventional-changelog and continue using `unclog` then. But again, the intention behind this PR is not to replace `unclog`.

CONTRIBUTING.md
melekes: > does this mean we change from squash-and-merge to preserving commit history in PRs,

no. I removed this paragraph because we enforce it via Github settings, so I felt that there's no real need to state that **if there's no other way to merge your PR other than squash merge**. I can revert of course.

> This ! thing may be particularly vexing: do we want to be forced to think of which changes might be breaking, when making every single commit (or, perhaps, when grooming commits pre-merge)?

no

adizere: It would help to continue keeping this remark about squash and merge in the CONTRIBUTING.md doc.

thanethomson:
For context, the whole reason why I created unclog was to deliberately separate the Git commit log (which targets _developers_) from the changelog (which targets _users_). And also to avoid the annoying and unnecessary conflicts creating by multiple PRs attempting to update the `CHANGELOG_PENDING.md` file (the old process).

We considered conventional commits as a standard previously, but this suffers from the problems that @mzabaluev has outlined here and assumes that we only have one target audience for the changelog (which is not true).

adizere:
Let's write down the categories, then LGTM

PR #1997, opened 2024-01-09
feat(state)!: do not save genesis file in DB (backport #1295, #1293, #1297)

melekes:
Backports #1295, #1293, #1297


sergio-mena:
What is the reason we are backporting this?

melekes:
> What is the reason we are backporting this?

https://github.com/mmulji asked us to backport this for `gaia`.

melekes:
cc @mmulji

PR #1998, opened 2024-01-09
Fix paths to scripts/metricsgen in go:generate commands

hvanz:
Some paths to the script `metricsgen` were broken when the files were moved to `internal/`.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/internal/blocksync/metrics.go
+++ b/internal/blocksync/metrics.go
@@ -12,7 +12,7 @@ const (
 	MetricsSubsystem = "blocksync"
 )
 
-//go:generate go run ../scripts/metricsgen -struct=Metrics
+//go:generate go run ../../scripts/metricsgen -struct=Metrics
 
 // Metrics contains metrics exposed by this package.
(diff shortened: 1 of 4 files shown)


jmalicevic:
There is one `../` missing  in `internal/store` too 🙏

hvanz:
I don't see any `metrics.go` file or `go:generate` command in `internal/store`.

jmalicevic:
> I don't see any `metrics.go` file or `go:generate` command in `internal/store`.

Ah sorry they are on my branch where i introduce new metrics for the first time. Ignore :) But the indexer should have them (on main and in v1 at least)?

hvanz:
I don't see any problem with the indexer either. I checked all `go:generate` commands in `internal` and they seem to be ok.

melekes (approved):
👍

jmalicevic (approved):
Sorry for the confusion, again something local :/ Thanks for this!

PR #2000, opened 2024-01-09
feat(state)!: do not save genesis file in DB (backport #1295, #1293, #1297)

melekes:
Backports #1295, #1293, #1297


sergio-mena:
`v0.37x` is a stable branch. If we have to backport this because some user asked for it, we should backport it to `v0.37.x-experimental`

melekes:
I don't think `gaia` (cosmos hub) will switch to an experimental branch. cc @mmulji

sergio-mena:
This is a new feature. If a user won't use the experimental branch, then they need to upgrade to the version that features it (unless there are strong security implications). It is unsustainable to backport major improvements (some of them breaking previous behavior) onto stable release branches: some users may be _surprised_ to bump into those behavioral changes on a "point upgrade" as opposed to a major upgrade where it is somehow expected.

This is a discussion we have already had within the team. There is no way to avoid this tradeoff.

Issue #646, opened 2023-04-05
`(*pex.addrBook).Wait()` never called, occasionally continues writing after `(*node.Node).Wait()` returns


mark-rushakoff:
### Setup

**CometBFT version** (use `cometbft version` or `git rev-parse --verify HEAD` if installed from source): v0.37.0

**Have you tried the latest version**: no

**ABCI app** (name for built-in, URL for self-written if it's publicly available): simapp in Cosmos SDK

**Environment**:
- **OS** (e.g. from /etc/os-release): `go test` inside macOS
- **Install tools**:
- **Others**:

**node command runtime flags**:

### Config

Default config.

### What happened?

In the Cosmos SDK, I am working on a set of tests that run a real CometBFT `*node.Node` in-process along with a simapp instance.

The tests defer `node.Stop(); node.Wait()`. Usually they pass.

Occasionally, there are failures like: `testing.go:1206: TempDir RemoveAll cleanup: unlinkat /var/folders/.../TestCometStarter_PortContentionattempt_0278742771/001/config: directory not empty`. If I switch away from using `t.TempDir` so that I can inspect the `config` directory which isn't empty, it always contains `addrbook.json`.

If I print `debug.Stack() from inside `(*pex.addrBook).saveToFile`, I see that the final call writing out `addrbook.json` is the final `a.saveToFile` call in `(*pex.addrBook).saveRoutine`: https://github.com/cometbft/cometbft/blob/365b0a7107259b6a18db55ed5dfa6fd6e92c9114/p2p/pex/addrbook.go#L494-L509

At the top of that function you will see a `defer a.wg.Done()` call. If we call `(*pex.addrBook).Wait()` (which simply calls `a.wg.Wait()`), then we would block until `a.SaveRoutine` returned.

However... nothing in the codebase calls `(*pex.addrBook).Wait()`. If you comment out the function, `go test -short ./...` passes without any compilation errors.

### How to reproduce it

```
$ git clone -b mr/testnet-node-stop-wait --depth=1 https://github.com/mark-rushakoff/cosmos-sdk
$ cd cosmos-sdk/simapp
$ while GORACE=halt_on_error=1 GODEBUG=tracebackancestors=3 go test ./internal/testnet/ -run=Contention -count=10 -failfast -race -v; do date; done
```

There are usually several successful runs before it fails.

### Logs

n/a

### `dump_consensus_state` output

n/a

### Anything else we need to know

The simplest solution looks like it would be:
[...]

Issue #652, opened 2023-04-05
PEX reactor does not have a `Wait()` method,  lacks way to wait for goroutine started in `OnStart()`


mark-rushakoff:
### Setup

**CometBFT version** v0.37.0

**Have you tried the latest version**: no

**ABCI app** (name for built-in, URL for self-written if it's publicly available): Cosmos SDK simapp

**Environment**:
- **OS** (e.g. from /etc/os-release): `go test` in macOS
- **Install tools**:
- **Others**:

**node command runtime flags**:

### Config

Default config.

### What happened?

Using the same setup as #646 -- a test that instantiates multiple `*node.Node` instances, starts them, checks something, then calls `Stop()` and `Wait()` on each Node -- and locally applying the patch mentioned in the issue to eliminate that test failure, the test almost always passes. On very rare occasions, there is a failure due to a log occurring after the test has ended. The stack looks like this:

```
panic: Log in goroutine after TestCometStarter_PortContention pex_reactor.go:458: <nil> INF Ensure peers module=pex numDialing=0 numInPeers=0 numOutPeers=0 numToDial=10 rootmodule=comet_node-0

goroutine 1969 [running]:
testing.(*common).logDepth(0xc0015d4ea0, {0xc0129c1a40, 0xd9}, 0x3)
        /opt/homebrew/Cellar/go/1.20.1/libexec/src/testing/testing.go:1003 +0x4b4
testing.(*common).log(...)
        /opt/homebrew/Cellar/go/1.20.1/libexec/src/testing/testing.go:985
testing.(*common).Logf(0xc0015d4ea0, {0x1030b4ff5, 0xb}, {0xc004c5c600, 0x4, 0x4})
        /opt/homebrew/Cellar/go/1.20.1/libexec/src/testing/testing.go:1036 +0x84
github.com/rs/zerolog.TestWriter.Write({{0x131279e18?, 0xc0015d4ea0?}, 0xc00b130350?}, {0xc003ae5040, 0xb7, 0x1a0})
        /Users/hh/go/pkg/mod/github.com/rs/zerolog@v1.29.0/writer.go:140 +0x358
bytes.(*Buffer).WriteTo(0xc00cf66c30, {0x10509a820, 0xc0063cc3f0})
        /opt/homebrew/Cellar/go/1.20.1/libexec/src/bytes/buffer.go:252 +0xe4
[...]

PR #2010, opened 2024-01-10
fix(p2p/pex): gracefully shutdown `Reactor`

melekes:
Closes https://github.com/cometbft/cometbft/issues/646, https://github.com/cometbft/cometbft/issues/652

---

#### PR checklist

- [ ] Tests written/updated


adizere:
cc @mark-rushakoff

p2p/pex/pex_reactor.go
```diff
@@ -166,11 +168,16 @@ func (r *Reactor) OnStart() error {
...
 }
 
-// OnStop implements BaseService.
-func (r *Reactor) OnStop() {
+// Stop implements BaseService.
+func (r *Reactor) Stop() error {
+	if err := r.BaseReactor.Stop(); err != nil {
+		return err
+	}
 	if err := r.book.Stop(); err != nil {
```
cason, line 176: We should stop the address book first.

cason: This in particular will block all other calls to `r.book`, since for saving the address book it acquires the lock.

cason: Actually, for clarity, closing the address book should be implemented in the `OnStop` method.

p2p/pex/pex_reactor.go
```diff
@@ -414,6 +421,8 @@ func (r *Reactor) SetEnsurePeersPeriod(d time.Duration) {
 
 // Ensures that sufficient peers are connected. (continuous).
 func (r *Reactor) ensurePeersRoutine() {
+	defer r.peersRoutineWg.Done()
+
```
cason, line 425: Consider also deferring closing the ticker.

p2p/pex/pex_reactor.go
```diff
@@ -650,6 +664,8 @@ func (r *Reactor) AttemptsToDial(addr *p2p.NetAddress) int {
 // Seed/Crawler Mode causes this node to quickly disconnect
 // from peers, except other seed nodes.
 func (r *Reactor) crawlPeersRoutine() {
+	defer r.peersRoutineWg.Done()
+
```
cason, line 668: Consider also deferring closing the ticker.

p2p/pex/pex_reactor.go
```diff
@@ -414,6 +421,8 @@ func (r *Reactor) SetEnsurePeersPeriod(d time.Duration) {
 
 // Ensures that sufficient peers are connected. (continuous).
 func (r *Reactor) ensurePeersRoutine() {
+	defer r.peersRoutineWg.Done()
+
 	var (
 		seed   = cmtrand.NewRand()
 		jitter = seed.Int63n(r.ensurePeersPeriod.Nanoseconds())
```
cason, line 428: At startup we sleep for this period. We should consider a select inside the if of line 434 so that we don't block `Stop()` while sleeping there.

p2p/pex/pex_reactor.go
```diff
@@ -475,6 +485,10 @@ func (r *Reactor) ensurePeers() {
 	maxAttempts := numToDial * 3
 
 	for i := 0; i < maxAttempts && len(toDial) < numToDial; i++ {
+		if !r.IsRunning() || !r.book.IsRunning() {
```
cason, line 488: The first condition here does make much sense, as we don't stop the reactor until this thread is terminates, right?

cason (approved):
This problem is super tricky and we probably need to find a more comprehensive solution for it (namely, ensuring that routines are concluded as part of closing a service).

I think this patch mostly addresses the reported issues. Notice, however, that we still are reading/updating the address book in several other parts of the code, while it is possibly stopping or already stopped. Those changes will be in-memory only, not persisted, but this is not that bad.

I having said so, I have the impression that try breaking the `ensurePeers` method or the `crawlPeers` method (not addressed here) while stopping the reactor is... incomplete, as we handle some few corner cases while not really solving the problem.

p2p/pex/pex_reactor.go
melekes: right

p2p/pex/pex_reactor.go
melekes: Why? It will make the code more complicated, but the chance of someone stopping the node right after start is super low, no?

p2p/pex/pex_reactor.go
cason: >  This in particular will block all other calls to r.book, since for saving the address book it acquires the lock.

Every call to the address bock acquires a lock. This lock is also acquired by `Save()`. So while saving, other calls to the address book will hang. From this point, assuming that the address book is stopped, if we change something on the address book, this changes won't be persisted.

cason:
In the light of #2041, probably the best solution here is:
1. Create a local channel to be closed when we are stopping the service
2. Quit the sub0routines when this channel is closed
3. Wait for the sub-routines to quit, using the wait group

Issue #493, opened 2023-03-10
Clarify the operation of a recovering node


cason:
CometBFT considers the crash-recovery failure model, meaning that nodes may crash and then recovery, rejoining the distributed computation in a _consistent_ state. For this to happen, nodes should persist relevant information and state changes during their regular operation, so that during recovery they are able to restore the state they had just before crashing.

Recovering the state of a node after a crash is a tricky operation. Several modules of CometBFT persist information that they are expected to recover after a crash. The consensus protocol keeps a Write-Ahead Log (WAL) to persist crucial information. The block store, the state store, the evidence reactor, the transaction indexer, and the address book persist data to their own DBs. And the application itself should adhere to the crash-recovery failure model, implementing a persistence strategy.

Among the mentioned modules, probably the best documented recovery procedure regards ABCI [applications](https://github.com/cometbft/cometbft/blob/main/spec/abci/abci++_app_requirements.md#crash-recovery). The consensus WAL is very superficially [covered](https://github.com/cometbft/cometbft/blob/main/spec/consensus/wal.md), while the other DBs are essentially not documented. In any case, the assumptions regarding the persisted state and its recovery are not documented.

It is worth noting that when the state persistence is delegated to a database, the recovery procedure tends to be straightforward, as it is provided by the database implementation. As far as I known, consensus is the only module that adopts transactional semantics for persisted data, based on a WAL. The recovery of the consensus WAL is particularly tricked and not well documented.

Definition of Done:
- [ ] List all databases adopted by CometBFT modules, summarize the persistence assumptions, and document, when it is the case, the relevant aspects of the recovery procedures
- [ ] Document the consensus Write-Ahead Log and the operation of the consensus protocol during recovery. This should include the interaction between consensus and the ABCI application, cover only from the side of the application in the existing documentation.

PR #2017, opened 2024-01-11
perf(internal/state): avoid double-saving FinalizeBlockResponse

melekes:
for performance reasons.
Closes https://github.com/cometbft/cometbft/issues/1976

---

#### PR checklist

- [ ] Tests written/updated


melekes:
cc @ValarDragon

ValarDragon (approved):
Amazing tysm! This looks great to me

internal/state/store.go
```diff
@@ -524,39 +524,40 @@ func (store dbStore) LoadFinalizeBlockResponse(height int64) (*abci.FinalizeBloc
...
+			return info.FinalizeBlock, nil
 		}
-		return responseFinalizeBlockFromLegacy(info.LegacyAbciResponses), nil
+		// END OF DEPRECATED lastABCIResponseKey
+		return nil, fmt.Errorf("expected last ABCI responses at height %d, but none are found", height)
 	}
-
-	return info.FinalizeBlock, nil
+	resp := new(abci.FinalizeBlockResponse)
+	err = resp.Unmarshal(buf)
```
cason, line 559: Here actually we should check for corruption, and maybe exit.

melekes:
> I wonder if the description should mention the deprecation of the key

And what will be the reason for this? The change is non-breaking. Plus the key is not exported.

> and the new way to store/retrieve the last ABCI responses.

The best place for this would probably be the spec, but that is outside of the scope for this PR (spec for state/store doesn't exist).

cason:
You are right. Regarding the documentation of the crash-recovery mechanisms of Comet we have this old, never tackled issue: https://github.com/cometbft/cometbft/issues/493. Because this is a worm hole.

internal/state/store.go
melekes: Good call 👍 Fixed in https://github.com/cometbft/cometbft/pull/2017/commits/f7dc7b4ffe0c12783a84b8f5620df22a961dfc3a

internal/state/store.go
```diff
@@ -575,30 +578,21 @@ func (store dbStore) SaveFinalizeBlockResponse(height int64, resp *abci.Finalize
...
 		return err
 	}
 
-	return store.db.SetSync(lastABCIResponseKey, bz)
+	// Save the ABCI response.
+	//
+	// We always save the last ABCI response for crash recovery.
+	// If `store.DiscardABCIResponses` is true, then we delete the previous ABCI response.
+	if store.DiscardABCIResponses && height > 1 {
+		if err := store.db.Delete(calcABCIResponsesKey(height - 1)); err != nil {
```
jmalicevic, line 591: While we do not double save a lot of data, we do have 2 DB operations here. Note that if compaction does not work as expected, this might still be blowing up `state.db` even though we are deleting data. 

If `DiscardABCIResponses` is true then we could potentially just save everything under `LastABCIResponsesKey`. 
So we have only one key in the DB and don't risk having potentially a big DB due to lack of compaction kicking in. 

The changes could look something like this: 

```
func  (store dbStore) LoadLastFinalizeBlockResponse(height int64) (*abci.FinalizeBlockResponse, error) {

.....
var bz
if store.DiscareABCIResponse == true {

    bz, err := store.db.Get(lastABCIResponseKey)
    
 } else {

    bz, err := store.db.Get(calcABCIResponseKey(height))
}
// Perform checks on height

....

```

Then in 

```
func (store dbStore) SaveFinalizeBlockResponse(height int64, resp*abci.FinalizeBlockResponse) error {

...

var resp 
if store.DiscardABCIResponses == true {
      store.db.SetSync(lastABCIResponseKey, bz)
} else {

  store.db.SetSync(calcLastABCIResponseKey(height), bz)
}
```

The only potential problem I see is if this setting (`DiscardABCIResponses`) can be changed between node boot-ups so there might be some additional checks to be added.

melekes: > While we do not double save a lot of data, we do have 2 DB operations here. Note that if compaction does not work as expected, this might still be blowing up state.db even though we are deleting data.

Good point 👍 under what conditions compaction does not work as expected?

> The only potential problem I see is if this setting (DiscardABCIResponses) can be changed between node boot-ups so there might be some additional checks to be added.

This is a valid concern.

jmalicevic: > > While we do not double save a lot of data, we do have 2 DB operations here. Note that if compaction does not work as expected, this might still be blowing up state.db even though we are deleting data.
> 
> Good point 👍 under what conditions compaction does not work as expected?
> 

At the moment (without the changes we plan to introduce with in-process compaction, there is no guarantee that golevleDB will actually compact the storage, the same way it has not been doing so previously for blocks. . This is worth testing out  in this case though. If the problem exists, then calling `CompactRange` after a couple of these delete operations would solve the storage footprint problem. 

> > The only potential problem I see is if this setting (DiscardABCIResponses) can be changed between node boot-ups so there might be some additional checks to be added.
> 
> This is a valid concern.

I am not sure this is a problem in your version of the code, only in what I propose.

jmalicevic: @melekes , what could be worth running is a small testnet on this branch and monitor the storage usage. Even on one node.

internal/state/store.go
```diff
@@ -524,39 +523,43 @@ func (store dbStore) LoadFinalizeBlockResponse(height int64) (*abci.FinalizeBloc
 // This method is used for recovering in the case that we called the Commit ABCI
 // method on the application but crashed before persisting the results.
 func (store dbStore) LoadLastFinalizeBlockResponse(height int64) (*abci.FinalizeBlockResponse, error) {
```
ancazamfir, line 525: Not sure I understand the crash scenario fully. And this is the first time I am looking at this code. With the new change if `height` is smaller than the height of the `lastABCIResponseKey` it will return stored results (or not found if  `DiscardABCIResponses` was true). Maybe this is not possible in the crash scenario. If that's the case, could you maybe document this?

ancazamfir:
Was looking at `TestLastFinalizeBlockResponses`. Are the comments wrong?
```
diff --git a/internal/state/store_test.go b/internal/state/store_test.go
index a80d0342d..9cf34a24a 100644
--- a/internal/state/store_test.go
+++ b/internal/state/store_test.go
@@ -505,7 +505,7 @@ func TestFinalizeBlockResponsePruning(t *testing.T) {

 func TestLastFinalizeBlockResponses(t *testing.T) {
        // create an empty state store.
-       t.Run("Not persisting responses", func(t *testing.T) {
+       t.Run("Persisting responses", func(t *testing.T) {
                stateDB := dbm.NewMemDB()
                        DiscardABCIResponses: false,
@@ -540,7 +540,7 @@ func TestLastFinalizeBlockResponses(t *testing.T) {
                require.Equal(t, response1, responses)
        })

-       t.Run("persisting responses", func(t *testing.T) {
+       t.Run("Not persisting responses", func(t *testing.T) {
                stateDB := dbm.NewMemDB()
                height := int64(10)
                // stub the second abciresponse.
```

ancazamfir:
Small nit:
```
diff --git a/internal/state/store.go b/internal/state/store.go
index e438e0b4e..c6b58c6dd 100644
--- a/internal/state/store.go
+++ b/internal/state/store.go
@@ -479,7 +479,7 @@ func TxResultsHash(txResults []*abci.ExecTxResult) []byte {
 }

 // LoadFinalizeBlockResponse loads the DiscardABCIResponses for the given height from the
-// database. If the node has D set to true, ErrABCIResponsesNotPersisted
+// database. If the node has D set to true, ErrFinalizeBlockResponsesNotPersisted
 // is persisted. If not found, ErrNoABCIResponsesForHeight is returned.
 func (store dbStore) LoadFinalizeBlockResponse(height int64) (*abci.FinalizeBlockResponse, error) {
        if store.DiscardABCIResponses {
```

internal/state/store.go
melekes: > With the new change if height is smaller than the height of the lastABCIResponseKey

This is not possible IMHO. I.e. if we know we crashed at height 5, then we'll try to restore last responses for this height.

ValarDragon:
Anything needed to progress this? Really excited about halving this I/O write time, to improve latency under heavy event load!

melekes:
Nothing is blocking this afaik. The follow-up to this PR should be calling compaction (similar to https://github.com/cometbft/cometbft/pull/1972) because it deletes the previous ABCI response with each new height. See https://github.com/cometbft/cometbft/pull/1972#discussion_r1482421058

melekes:
@mergify backport v1.x

melekes:
@mergify backport v1.x --force

Issue #642, opened 2023-04-03
CListMempool has data race(s)


peterbourgon:
[mempool.CListMempool.Update writes to the height field](https://github.com/cometbft/cometbft/blob/8d6d2bff3ec443dd8c79fccd13f111acba3d4ee2/mempool/clist_mempool.go#L580-L589) under the assumption that the updateMtx field on the CListMempool struct has been locked by its caller. But e.g. [mempool.CListMempool.resCbFirstTime reads from the height field](https://github.com/cometbft/cometbft/blob/8d6d2bff3ec443dd8c79fccd13f111acba3d4ee2/mempool/clist_mempool.go#L371-L394) without taking that lock. And that method is called by [reqResCb](https://github.com/cometbft/cometbft/blob/8d6d2bff3ec443dd8c79fccd13f111acba3d4ee2/mempool/clist_mempool.go#L293-L305) via [abci/client.ReqRes.SetCallback](https://github.com/cometbft/cometbft/blob/8d6d2bff3ec443dd8c79fccd13f111acba3d4ee2/abci/client/client.go#L89-L97), neither of which take the lock. This is a data race.

The height field (and others) are commented as "atomic" which suggests they are expected to be accessed with sync/atomic functions like [mempoolTx.Height](https://github.com/cometbft/cometbft/blob/8d6d2bff3ec443dd8c79fccd13f111acba3d4ee2/mempool/clist_mempool.go#L683-L686). But that's not the case at the moment. If that's the intent, then they should only be accessed with sync/atomic functions, and do not require the updateMtx to be locked. But this kind of mixing of individually-atomic fields and a mutex that governs (apparently arbitrary) fields in the type is basically impossible to get right :) and probably deserves a refactor.

adizere:
Closed by #2021

PR #2021, opened 2024-01-11
mempool: Fix data races in CListMempool's height and notifiedTxsAvailable

hvanz:
Addresses #642

---

#### PR checklist

- [X] Tests written/updated

### DIFF
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -24,12 +24,11 @@ import (
 // mempool uses a concurrent list structure for storing transactions that can
 // be efficiently accessed by multiple concurrent readers.
 type CListMempool struct {
-	// Atomic integers
-	height   int64 // the last block Update()'d to
-	txsBytes int64 // total size of mempool, in bytes
+	height   atomic.Int64 // the last block Update()'d to
+	txsBytes atomic.Int64 // total size of mempool, in bytes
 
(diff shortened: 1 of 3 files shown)


mempool/clist_mempool.go
```diff
@@ -24,12 +24,11 @@ import (
 // mempool uses a concurrent list structure for storing transactions that can
 // be efficiently accessed by multiple concurrent readers.
 type CListMempool struct {
-	// Atomic integers
-	height   int64 // the last block Update()'d to
-	txsBytes int64 // total size of mempool, in bytes
+	height   atomic.Int64 // the last block Update()'d to
+	txsBytes atomic.Int64 // total size of mempool, in bytes
```
melekes, line 28: The issue only mentions the data race for `height` field. Did you also discover that there are data races for `txsBytes` and `notifiedTxsAvailable` fields?

melekes (approved):

mempool/clist_mempool.go
hvanz: The issue uses height just as an example. But yes, in the tests for this PR, I also saw that there were data races in the other variables.

hvanz: `txBytes` was actually implemented correctly. It was declared as an `int64` and accessed with `atomic.*` methods. I changed its type to `atomic.Int64` to keep it consistent with the other variables.

Issue #1309, opened 2023-09-08
Log if a follower receives a proposal with POLRound == LockedRound and the locked value is different than the proposed one


angbrav:
## Problem Definition
The code assumes that when a follower receives a proposal with `POLRound == LockedRound`, the locked value is equal to the proposed one ([line](https://github.com/cometbft/cometbft/blob/main/consensus/state.go#L1435)). This follows from the 2/3 assumption (at least 2/3 of the voting power is under the control of correct validators), which guarantees that (*) a consensus round cannot generate two certificates (+2/3) for different values.

## Proposal
I propose to log when a follower receives a proposal with `POLRound == LockedRound` and its locked value is different from the proposed one. This may indicate that the 2/3 assumption does not hold anymore. It should be discussed whether more drastic measures should be taken, as violating Property (*) may break safety.

angbrav:
> I assume that you are referring to this [line](https://github.com/cometbft/cometbft/blame/49127d27ef23e82f2e7cbc1dea9dd3b90db726c3/consensus/state.go#L1459) of the code, when we compared the POLRound, assume to be set (> -1), to the locked round:

Yes

cason:
A point that has to be made is that to accept a value with `POLRound > -1` we have to receive, or have received, `2/3+` voting power equivalent `Prevote` messages from `POLRound` for the proposed value. In terms of implementation, a node that proposes, or forward a `Proposal` with such content, also makes it available the corresponding `Prevote` messages that enable the receiver to attest that in round `POLRound` the proposed value has been declared valid.

At the same time, if `LockedRound == X`, the node has necessarily received `2/3+` voting power equivalent `Prevote` messages for `LockedValue` at round `X`.

So, if it happens that a `Proposal` with `POLRound == LockedRound > -1` but with proposed value different `LockedValue` is received and accepted, the node will have in its vote set `2/3+` voting power equivalent votes for both its `LockedValue` and the value proposed in the current round. In order to have this two quorums from different values at the same round, we necessarily (*) need some validators to double vote. This situation is observed when adding the second (equivocating) vote from one of such validators, whic leads the node to produce an evidence of misbehavior.

In other words, this situation is already caught by the evidence module of CometBFT, as it requires validators to double voting.

(*) This is true provided that the voting power of Byzantine nodes is not a majority of the total voting power. The implementation assumes that this does not happen, in particular because a majority of Byzantine voting power would prevent the propagation of evidences of misbehaviour.

[...]

PR #2042, opened 2024-01-15
consensus: increase log level for corner case when prevoting a proposal

cason:
Addresses #1309.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/internal/consensus/state.go
+++ b/internal/consensus/state.go
@@ -1457,8 +1457,8 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
 	*/
 	blockID, ok := cs.Votes.Prevotes(cs.Proposal.POLRound).TwoThirdsMajority()
 	ok = ok && !blockID.IsNil()
-	if ok && cs.ProposalBlock.HashesTo(blockID.Hash) && cs.Proposal.POLRound >= 0 && cs.Proposal.POLRound < cs.Round {
-		if cs.LockedRound <= cs.Proposal.POLRound {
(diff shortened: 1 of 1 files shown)


internal/consensus/state.go
```diff
@@ -1457,8 +1457,8 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
 	*/
 	blockID, ok := cs.Votes.Prevotes(cs.Proposal.POLRound).TwoThirdsMajority()
 	ok = ok && !blockID.IsNil()
-	if ok && cs.ProposalBlock.HashesTo(blockID.Hash) && cs.Proposal.POLRound >= 0 && cs.Proposal.POLRound < cs.Round {
```
cason, line 1460: POLRound > -1 here, by construction. Moreover, the `Proposal.ValidateBasic()` rejects POLRound < -1.

angbrav:
LGTM, thanks for taking care of it @cason! The logging is definitely in place.

I wonder if in such a worrying scenario: there is evidence that two quorum certificates have been formed for conflicting proposals, one should rather panic as malicious parties may have been able to violate safety and fork the chain. But this is probably a separate discussion.

cason:
> one should rather panic as malicious parties may have been able to break safety and fork the chain.

We have other situations where this may happen. This occurs when the assumption 2/3+ majority of honest processes is violated. We do not panic in this case, but I do agree that we should log this to make debugging easier.

melekes (approved):
👍

cason:
I was considering whether to backport this change. But I realized that the code of consensus `cs.defaultDoPrevote` method is substantially different (was refactored) on the release branches. So I would not.

PR #2046, opened 2024-01-16
misc(tools)!: remove `tools` package

faddat:
- 1) protos are now built with the sdk's proto builder 
- 2) go tools methodology just bloats .mod and .sum making comet harder to work with
- 3) tidy: mod and sum don't have maintenance tools in them anymore
- 4) stop using an abandoned Dockerfile

---------

tl;dr: by removing things from go.mod we can make comet easier to maintain, even if that means less stylish golang.

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
I'm actually in favor of this PR because in our `Makefile`, we always use the `@latest` version, which defeats the purpose of having a `tools` package in the first place.

Thanks @faddat ❤️

tools/README.md
```diff
@@ -1,5 +0,0 @@
-# tools
-
-Tools for working with CometBFT and associated technologies.
-Documentation for these tools can be found online in the
-[CometBFT tools documentation](https://docs.cometbft.com/main/tools/).
```
adizere, line 5: Does removing anything here affect what is written in https://docs.cometbft.com/main/tools/? If so, we should adapt the docs. From a quick glance it looks like not.

adizere (commented):
Concept ACK.

@melekes anyone else from the team needs to review/approve this ?

melekes:
> @melekes anyone else from the team needs to review/approve this ?

I'd like to hear what @mzabaluev and/or @sergio-mena think about this.

cason:
Should we check whether this at least currently works? I mean the `Dockerfile`, in particular, which still uses a pretty old version of Go.

Dropping a lot of dependencies for the repo is plus of this proposal.

scripts/mockery_generate.sh
```diff
@@ -2,5 +2,8 @@
 #
 # Invoke Mockery v2 to update generated mocks for the given type.
 
-go run github.com/vektra/mockery/v2 --disable-version-string --case underscore --name "$*"
+
+
+go install github.com/vektra/mockery/v2@latest
+mockery --disable-version-string --case underscore --name "$*"
```
sergio-mena, line 8:
```suggestion
go run github.com/vektra/mockery/v2@latest --disable-version-string --case underscore --name "$*"
```
Consider this alternative

sergio-mena:
> I'd like to hear what @mzabaluev and/or @sergio-mena think about this.

I've applied this PR's patch and run "make build", "make proto-gen", "make mockery", "make lint". Everything seems to be working fine, so I'm in favor of adopting this.

BTW, thanks @faddat for submitting this 👍

mzabaluev (approved):
I never had to use this subdirectory so I don't have a very informed opinion, but the content does not seem like a healthy way of doing things and it does bring a lot of unneeded dependencies.

I think we can treat buf as a required tool installed externally, at least that's what I always did locally. Its results are fairly stable, and more dependent on compiler plugins which can be pinned to a revision in `buf.gen.yaml`.

melekes:
@mergify backport v0.34.x

melekes:
@mergify backport v0.37.x

PR #2047, opened 2024-01-16
fix(privval): retry accepting a connection on errors

### LINKED ISSUE #452: Unclear private validator accept connection logic
## Bug Report

version: 0.34.27

While diving into the `privval` package source code as part of an investigation into [horcrux](https://github.com/strangelove-ventures/horcrux) we discovered that the "accept connection logic" is unclear.

See [source](https://github.com/cometbft/cometbft/blob/main/privval/signer_listener_endpoint.go#L197-L200). Writing to the unbuffered channel `sl.connectRequestCh` after attempting to accept a connection in `SignerListenerEndpoint.serviceLoop` has no affect. This is because a single goroutine cannot read and write to the same unbuffered channel, it can only do that with buffered channels.

It would therefore be clearer to remove that select statement since it never actually writes to the channel.

If the reasoning behind the serviceLoop is to successfully accept one connection for each `connectRequest` then one could change `connectRequestCh` into a buffered channel with length 1 and only write to it when `acceptNewConnection` errored.

melekes:
Before the change, we'd error after just 1 attempt of accepting a connection.

Co-authored by: @corverroos

Closes #452

Port of https://github.com/corverroos/tendermint/pull/1

---

#### PR checklist

- [x] Tests written/updated


faddat:
Before you started to work on comet again or tenderment or whatever, we are calling the consensus mechanism these days, there was a pretty serious bug discovered by @jhernandez (Jorge) from stargaze

I think that the short version of the description of the bug is that it was possible to cause very serious CPU stress by initiating a P2P connection and then breaking it when the peer has begun to send blocks.

We should try to check this to make sure that we don't allow that back in, but please note that I'm not 100% certain that this will affect that

melekes:
> I think that the short version of the description of the bug is that it was possible to cause very serious CPU stress by initiating a P2P connection and then breaking it when the peer has begun to send blocks.

This change only affects the `privval` connection (between this node and signer, usually an HSM), which is private and never exposed to the outside the internet.

faddat:
ohhhh I missed this, thank you!

sergio-mena (approved):
Thanks for the synchronous explanations!

melekes:
@mergify backport v0.37.x

PR #2049, opened 2024-01-16
feat: add gofumpt

### LINKED ISSUE #604: Formatting with gofumpt

melekes:
Our codebase is already sufficiently formatted, so we can enable **`gofumpt` by default**.

This PR also merges `lint-typo`, `lint-fix-typo`, `format`, `lint-format`, and `lint` Makefile
targets into a single `lint` target to ease the software development process. `codespell` CI job is removed in favor of `lint` job.

Closes #604


scripts/gofumpt.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+# Format staged Go files with `gofumpt`
+
+# Redirect output to stderr.
+exec 1>&2
```
mzabaluev, line 5: Why not let the operator redirect streams the way they see fit? Especially as the script outputs to `>&2` as if it makes a difference after this.

mzabaluev (commented):
Good, let's establish more order.

faddat:
gofumpt is enforced by .golangci.yml currently

scripts/gofumpt.sh
melekes: Fixed in https://github.com/cometbft/cometbft/pull/2049/commits/0a73d731aaa9dfba29687dc28eaca1f64f95489b

melekes:
> gofumpt is enforced by .golangci.yml currently

I find it interesting that the `gofumpt` linter in `golangci` does not cover all the files. I.e. when I run `gofumpt` separately, it will find and format more files than the linter in `golangci`. Hence, it's added to `Makefile` targets in this PR.

melekes:
NOTE: `golangci-lint` CI job was renamed to `lint`

Makefile
```diff
@@ -240,41 +242,28 @@ clean_certs:
...
-#	find . -name '*.go' -type f -not -path "*.git*"  -not -name '*.pb.go' -not -name '*pb_test.go' | xargs goimports -w -local github.com/cometbft/cometbft
-#.PHONY: format
-
-#? lint: Run latest golangci-lint linter
+#? lint: Lint, format and fix typos
 lint:
-	@echo "--> Running linter"
-	@go run github.com/golangci/golangci-lint/cmd/golangci-lint@latest run
+	@echo "--> Linting"
+	@go run github.com/golangci/golangci-lint/cmd/golangci-lint@latest run --fix
```
cason, line 248: Are you sure you want to automatically fix here? I think I liked it fails when there are problems.

Makefile
```diff
@@ -240,41 +242,28 @@ clean_certs:
...
+	@echo "--> Formatting"
+	@go run mvdan.cc/gofumpt@latest -l -w ./..
```
cason, line 250: Not sure if I understood why `./..`.

cason: Moreover, for some reason it reports here files inside `.git` dir.

cason (approved):
👍

Makefile
melekes: > Are you sure you want to automatically fix here?

Why would you not prefer the linter to fix the problem for you? `golangci-lint` will fail if the linter doesn't provide `AutoFix` https://golangci-lint.run/usage/linters/, which is largely still the case for many linters.

Makefile
melekes: fixed https://github.com/cometbft/cometbft/pull/2049/commits/fe80aad63d0095c95257399ac22441d7dabd0fe6

Makefile
greg-szabo: I agree that fixing is the right thing to do. If it causes issues down the line, we'll revert to no-fixing.

greg-szabo (approved):
I have more ideas that I would like to add/improve in a subsequent PR. Until then, this reduces chaos, it's a good thing.

melekes:
FYI, I don't have a strong opinion about merging 3 workflows into 1 (golangci-lint, gofumpt, codespell). I can break them apart if needed.

PR #2060, opened 2024-01-18
Backport ASA-2024-001

sergio-mena:
Manual backport of fix and UTs of advisory ASA-2024-001

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/types/params.go
+++ b/types/params.go
@@ -205,27 +205,58 @@ func (params ConsensusParams) ValidateBasic() error {
 	return nil
 }
 
+// ValidateUpdate validates the updated VoteExtensionsEnableHeight.
+// | r | params...EnableHeight | updated...EnableHeight | result (nil == pass)
+// | 1 | *                     | (nil)                  | nil
+// | 2 | *                     | < 0                    | VoteExtensionsEnableHeight must be positive
+// | 3 | <=0                   | 0                      | nil
+// | 4 | > 0; <=height         | 0                      | vote extensions cannot be disabled once enabled
+// | 5 | > 0; > height         | 0                      | nil (disable a previous proposal)
+// | 6 | *                     | <=height               | vote extensions cannot be updated to a past height
+// | 7 | <=0                   | > height (*)           | nil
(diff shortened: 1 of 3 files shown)


types/params.go
```diff
@@ -205,27 +205,58 @@ func (params ConsensusParams) ValidateBasic() error {
...
+	if updated == nil || updated.Abci == nil {
 		return nil
 	}
-	if params.ABCI.VoteExtensionsEnableHeight == updated.Abci.VoteExtensionsEnableHeight {
+	// 2
+	if updated.Abci.VoteExtensionsEnableHeight < 0 {
+		return errors.New("VoteExtensionsEnableHeight must be positive")
+	}
+	// 3
+	if params.ABCI.VoteExtensionsEnableHeight <= 0 && updated.Abci.VoteExtensionsEnableHeight == 0 {
```
mzabaluev, line 229: Side note: why do we use all-caps in the human-written field name (that's not what you should do to acronyms per CamelCase), when protobuf has it differently?

mzabaluev (approved):
Looks good.

types/params.go
sergio-mena: Good point. But I think it's too late to change that now for backwards compatibility

sergio-mena: Also ABCI is an acronym, so not sure how CameCase usually deals with that

mzabaluev: > Also ABCI is an acronym, so not sure how CameCase usually deals with that

At least in Rust, you make it a "word" e.g. `TcpStream`.

mzabaluev: > it's too late to change that now for backwards compatibility

Or the last opportunity to clean up the style prior to the 1.0 release? :wink:

greg-szabo: Yeah, the `params.ABCI` vs `updated.Abci` got me in the first half. It's not something we can fix here but I can open an issue with your side note, if we want to fix it.

greg-szabo (approved):

PR #2062, opened 2024-01-18
fix(ci): lint ci fail

levisyin:
This pr fixed a small lint error and added `lint-fix` to Makefile, introducing a convenient way for developers to fix make lint errors

---

#### PR checklist

- [ ] Tests written/updated


melekes:
Thanks @levisyin ❤️ , but the correct solution I think is to uncomment `lint-format` Makefile target https://github.com/cometbft/cometbft/pull/2049, not to introduce yet another target.

PR #2066, opened 2024-01-18
`e2e`: test vote extension activation via `InitChain` and `FinalizeBlock`

### LINKED ISSUE #2065: `e2e` should be able to test on-the-fly activation of vote extensions
Current version of our `e2e` tests tests vote extension activation via setting consensus parameter `VoteExtensionEnableHeight` at genesis.

`e2e`should also be able to test the activation of `VoteExtensionEnableHeight` via `InitChain` and `FinalizeBlock`. Which will typically be the case for existing chains upgrading to `v0.38.x`.

Ideally, the generator should also be able to output manifests that test this new way of activating vote extensions.

sergio-mena:
Closes #2065

This PR introduces a new `e2e` manifest option `VoteExtensionsUpdateHeight`, which allows the `e2e` application to test vote extension activation mechanisms other than via genesis file.

The value of `VoteExtensionsUpdateHeight` is interpreted as follows:
* -1: the value of `VoteExtensionsEnableHeight` is to be set in the genesis file (this is equivalent to the mechanism currently present in `e2e`)
* 0: the value of  `VoteExtensionsEnableHeight` is to be set in `InitChainResponse`
* _H_ (where _H_ >= `InitialHeight`, and _H_ < `VoteExtensionsEnableHeight`): the value of  `VoteExtensionsEnableHeight` is set in the `FinalizeBlockResponse` corresponding to height _H_ (i.e., at the end of height _H_)

The PR also extends the manifest generator to produce different (valid) values of the new option `VoteExtensionsUpdateHeight` in combination with `VoteExtensionsEnableHeight`

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

Issue #1731, opened 2023-12-04
consensus: proposer-based timestamps (PBTS)


melekes:
### Summary

**Use proposer-based timestamps instead of calculating the median time of vote timestamps.**

<details>
<summary>Detailed description of previous efforts</summary>

### Context

This collapsed section describes the previous work on the adoption of proposer-based timestamps (PBTS) for the blocks produced in CometBFT networks and points out relevant aspects to be considered when porting this work to the current code base.

Part of the references included in this section are linked in the PBTS project board in the old Tendermint Core repository: https://github.com/orgs/tendermint/projects/10/views/4.

### Rationale

The adoption of proposer-based timestamps (PBTS) was originally proposed in 2018 (tendermint/tendermint#2840, tendermint/tendermint#1615) to replace the BFT Time algorithm (tendermint/tendermint#2013) that is currently used to compute the timestamp of new blocks. In short, the [BFT Time algorithm](https://github.com/cometbft/cometbft/blob/main/spec/consensus/bft-time.md) sets as timestamp of a block (stored in its `Time` field) the voting-power-weighted median of the `Precommit` vote messages (stored at the block's `LastCommit` field) that decided the previous block.

The main problems with BFT Time are the following:
 * it can't be used for (future adoption of) tendermint/tendermint#1319 (since each vote has a unique timestamp component)
 * it gives +1/3 of the validator set complete control over the timestamp (see tendermint/tendermint#2653)

The consensus algorithm adopted by CometBFT needs to be adapted for the use of proposer-based timestamps. Throughout 2021, two proposals of changes in the consensus algorithm were specified (tendermint/spec#261 and tendermint/spec#369).  In addition, the architectural changes required for the adoption of PBTS were discussed in [ADR 071](https://github.com/cometbft/cometbft/blob/main/docs/architecture/tendermint-core/adr-071-proposer-based-timestamps.md), originally built from the first proposal (tendermint/tendermint#6799), then adapted for the second proposal (tendermint/tendermint#7764).

## Spec

At the moment, the PBTS spec on [`main`](https://github.com/cometbft/cometbft/tree/main/spec/consensus/proposer-based-timestamp) refers to the _first proposal_ of changes in the consensus algorithm. The second and latest proposal of changes is currently available on [`master`](https://github.com/cometbft/cometbft/blob/master/spec/consensus/proposer-based-timestamp/) (https://github.com/tendermint/tendermint/pull/8096).

[...]

PR #2089, opened 2024-01-22
PBTS: forward port of tendermint/tendermint#7709

cason:
Part of #1731.

**Must not be merged before  #2018 is merged**. When this happens, please change the base branch to [feature/pbts](https://github.com/cometbft/cometbft/tree/feature/pbts).

The original PR precedes tendermint/tendermint#8488, a PR that adopted `metricsgen` for metric generation. This PR was forward ported to this repository as in tendermint/tendermint#9178. Some changes were added to the forward-ported PR to match the behaviour and description of the added metric with this follow-up change.

---

#### PR checklist

- [ ] Tests written/updated


internal/consensus/metrics.gen.go
```diff
@@ -192,39 +192,48 @@ func PrometheusMetrics(namespace string, labelsAndValues ...string) *Metrics {
...
+		ValidatorsPower:             discard.NewGauge(),
+		ValidatorPower:              discard.NewGauge(),
+		ValidatorMissedBlocks:       discard.NewGauge(),
+		MissingValidators:           discard.NewGauge(),
+		MissingValidatorsPower:      discard.NewGauge(),
+		ByzantineValidators:         discard.NewGauge(),
+		ByzantineValidatorsPower:    discard.NewGauge(),
+		BlockIntervalSeconds:        discard.NewHistogram(),
+		NumTxs:                      discard.NewGauge(),
+		BlockSizeBytes:              discard.NewGauge(),
```
glnro, line 222: Is this supposed to be a gauge or histogram https://github.com/tendermint/tendermint/pull/7709/files#diff-8ba8e127ec86116d1c551254adca1654f36b94251c593748e05d91dcceb1f4beR267

cason: Good point.

Originally it was a Gauge: https://github.com/tendermint/tendermint/commit/489d9b9184f06d1c7e38f5895d0b0e42a9839aa3

This is the version on main.

Then it was changed to an histogram: https://github.com/tendermint/tendermint/commit/2f6141645bdc92a7f16653ce544a5ba1f4c907d5

This version is on master, but not on main.

cason: In any case, in `metrics.go` it is a Gauge (line 58).

internal/consensus/metrics.go
```diff
@@ -122,6 +122,12 @@ type Metrics struct {
 	// correspond to earlier heights and rounds than this node is currently
 	// in.
 	LateVotes metrics.Counter `metrics_labels:"vote_type"`
```
sergio-mena, line 124: At this point (roughly) I see two hunks in the original commit (#7709) that have been discarded during the forward-port. Looking at those changes I see they are adding `ProposalTimestampDifference` to two constructors. IIRC these constructors are now autogenerated by "metrics.gen" aren't they? Hence the changes in the "make metrics" commit.

(just double-checking here)

internal/consensus/metrics.gen.go
sergio-mena: tendermint/tendermint#6549 (from 2f6141645 found by @cason above) is the commit that changed from Gauge to Histogram. This happened in the `v0.35.x`/`v0.36.x`/`master` timeline, but not on `v0.34.x`/`v0.37.x`/`v0.38.x`/`main`.

We can decide to also forward-port tendermint/tendermint#6549 to `main`, but that would be independent from the PBTS effort.

sergio-mena: Actually, I've just read tendermint/tendermint#5752, which is the issue tendermint/tendermint#6549 was addressing, and it may make sense to forward-port it to main, so I copied tendermint/tendermint#5752 into our repo: #2104

sergio-mena (approved):

melekes (approved):

internal/consensus/metrics.go
cason: Yes, most of the original changes in this file (`metrics.go`) are now at the automatically generated (`metrics.gen.go`). It is maybe useful to compare the current autogenerated file content to the manually generate metrics setup on the original commit (I've checked, they match, but please double check).

sergio-mena: I did check 👍

PR #2093, opened 2024-01-22
consensus: Add `chain_size_bytes` metric

hvanz:
This metric measures the total size of the chain in bytes.

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

PR #2108, opened 2024-01-24
test(localnet): Add monitoring tools for localnet

### LINKED ISSUE #2107: Grafana dashboards for local use
## Feature Request
Docker Compose to start Granafa and Prometheus servers that monitor locally running nodes.
### Summary
The Grafana dashboards in the qa-infra repository have some interesting graphs that we could use during local testing. We could also improve those dashboards easier if we have a locally running Grafana server.

### Problem Definition
I have a hard time understanding the health of a node in some cases: how long does consensus take, what's the memory usage, etc. The Comet log file has a lot of information but we should leverage the Prometheus metrics during development better.

With this tool we'd get easier visibility into Prometheus metrics. It's an external tool, no Go code needs to change.

### Proposal

* Docker Compose file that starts up a Grafana and a Prometheus container.
* They need to connect to each other.
* Prometheus needs to be able to connect to nodes running locally on the host machine.
* Pre-populated dashboards would be nice.
*

greg-szabo:
Closes #2107

This PR introduces a new folder `test/monitoring` which contains a Docker Compose file that creates a Grafana and a Prometheus server locally for the developer. It connects up with a localnet installation.

The localnet Docker compose file (`compose.yml`) was extended to allow Prometheus traffic from the container and documentation was updated to reflect the changes.

Short way to execute:
```
make build-linux localnet-start monitoring-start
```
---

The JSON files are pre-populated Grafana dashboards. Two of them are from the qa-infra repository, one is part of host monitoring with "node-explorer" and one is a new one created for storage experiments that are currently ongoing.

#### PR checklist

- [ ] Tests written/updated - no Go code change
- [X] Updated relevant documentation (`docs/` or `spec/`) and code comments - test docs were updated


greg-szabo:
Sigh, I'll check YAML linting tomorrow. I have a feeling that copy-pasting random Grafana dashboards is not going well with the linter.

Makefile
```diff
@@ -327,14 +327,22 @@ build_c-amazonlinux:
...
 localnet-stop:
 	docker-compose down
 .PHONY: localnet-stop
 
+#? monitoring-start: Start Prometheus and Grafana servers for localnet monitoring
+monitoring-start:
+	cd test/monitoring && docker-compose up -d
+
+#? monitoring-stop: Stop the Prometheus and Grafana servers
+monitoring-start:
```
melekes, line 343:
```suggestion
monitoring-stop:
```

networks/local/localnode/Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN apk update && \
 
 VOLUME /cometbft
 WORKDIR /cometbft
-EXPOSE 26656 26657
+EXPOSE 26660 26656 26657
```
melekes, line 9:
```suggestion
EXPOSE 26656 26657 26660
```

melekes (changes requested):
nice 👍

test/monitoring/config-grafana/provisioning/dashboards-data/storage_experiments.json
```diff
@@ -0,0 +1,998 @@
+{
```
melekes, line 1: Is the idea to eventually rename this to `storage` once it's past `experiments` stage?

melekes (approved):

test/monitoring/config-grafana/provisioning/dashboards-data/storage_experiments.json
greg-szabo: Yes! Currently, it has metrics that were not introduced on `main` yet and also some metrics that Jasmina uses for the current storage experiments but they may not be introduced in a release at all. (Honestly, I released this because I had to share the storage board with her.)

Actually, we can rename it right now, so we don't forget. But the content might change in the future.

PR #2112, opened 2024-01-24
fix: `ValidateUpdate`: allow no-change updates regardless of current height

### LINKED ISSUE #2109: v0.34.8: possible regression in `ConsensusParams#ValidateUpdate`
### Setup

**CometBFT version** (use `cometbft version` or `git rev-parse --verify HEAD` if installed from source): v0.38.4

**ABCI app** (name for built-in, URL for self-written if it's publicly available): berachain

**Environment**:
- **Install tools**:
- **Others**:

### What happened?

```
6:42PM ERR failed to unmarshal commit info err="proto: wrong wireType = 2 for field Round" module=server
6:42PM ERR failed to process proposal err="proto: wrong wireType = 2 for field Round" hash=2819B22226DB5BC1FC70D173E81F6CEB9B8AD2FA17BCA78608CA0F97CBBE30A0 height=237778 module=server
6:42PM ERR prevote step: state machine rejected a proposed block; this should not happen:the proposer may be misbehaving; prevoting nil err=null height=237778 module=consensus round=63

4:45PM ERR failure when running app err="error during handshake: error on replay: commit failed for application: updating consensus params: vote extensions cannot be updated to a past or current height, enable height: 2, current height 216184"
```

successful replay

https://github.com/cometbft/cometbft/commit/d766d20c0

`params.ABCI.VoteExtensionsEnableHeight == updated.Abci.VoteExtensionsEnableHeight`

I think this is the offending change
which is not present in v0.38.4
before v0.38.4, after upgrade && during replay ApplyBlock would succeed because `params.ABCI.VoteExtensionsEnableHeight == updated.Abci.VoteExtensionsEnableHeight` `2 == 2` passes :white_check_mark:

sergio-mena:
Closes #2109

In `v0.38.2`, when the App was setting `VoteExtensionsEnableHeight` to a value that was the same as the current one, even if is useless as a ConsensusParams update, we were accepting it because it doesn't do any harm.

When we released `v0.38.3` we refactored function `ValidateUpdate` and we started rejecting those (useless but harmless) cases.

Since `v0.38.3` is a point release, we shouldn't add restrictions on inputs to a function without a good reason for it. This PR is adding a rule to `ValidateUpdate` so re-allow all cases where the app is not really updating the value of `VoteExtensionsEnableHeight`, regardless of which height we are currently in.

### Testing done

Apart form CI tasks, we carried out the following extra (manual) activities:
* We reproduced the error reported (#2109) using our e2e test infra, by configuring the manifest to set a vote extension to a height in the past, which is the same height that was previously passed via the genesis file
* [THE FIX] In the production code, we introduced an early rule in function `ValidateUpdate` to return no error if the value to update is equal to the current value
* We also updated two unit tests that should not be returning an error now (and that were returning an error in `v0.38.3`)
* We re-ran the e2e tests as explained above with the new fix, and the problem reported in #2109 no longer shows up
* Then, to make sure no other extra restrictions were introduced in `v0.38.3`, we did the following:
  * We temporarily replaced the implementation of `ValidateUpdate` with the one existing before `v0.38.3` (these temporary changes are not included in the PR, it was just done for testing pruposes)
  * We ran the new exhaustive unit tests introduced in `v0.38.3` and modified in this PR against the replaced (`v0.38.3`) implementation
  * We went through all the UTs that are now failing in `TestConsensusParamsUpdate_VoteExtensionsEnableHeight` (getting an unexpected error, or not getting an expected error), to make sure they don't represent a set of inputs to `ValidateUpdate` that were accepted in v0.38.2 but no longer accepted with the current version (the one in this PR)
  * We found no problematic test case

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/types/params.go
+++ b/types/params.go
@@ -210,7 +210,7 @@ func (params ConsensusParams) ValidateBasic() error {
 // |  1 | *                    | (nil)                  | nil
 // |  2 | *                    | < 0                    | VoteExtensionsEnableHeight must be positive
 // |  3 | <=0                  | 0                      | nil
-// |  4 | X                    | X                      | nil
+// |  4 | X                    | X (>=0)                | nil
 // |  5 | > 0; <=height        | 0                      | vote extensions cannot be disabled once enabled
 // |  6 | > 0; > height        | 0                      | nil (disable a previous proposal)
 // |  7 | *                    | <=height               | vote extensions cannot be updated to a past height
--- a/types/params.go
+++ b/types/params.go
@@ -207,15 +207,16 @@ func (params ConsensusParams) ValidateBasic() error {
 
 // ValidateUpdate validates the updated VoteExtensionsEnableHeight.
 // | r | params...EnableHeight | updated...EnableHeight | result (nil == pass)
-// | 1 | *                     | (nil)                  | nil
-// | 2 | *                     | < 0                    | VoteExtensionsEnableHeight must be positive
-// | 3 | <=0                   | 0                      | nil
-// | 4 | > 0; <=height         | 0                      | vote extensions cannot be disabled once enabled
-// | 5 | > 0; > height         | 0                      | nil (disable a previous proposal)
-// | 6 | *                     | <=height               | vote extensions cannot be updated to a past height
-// | 7 | <=0                   | > height (*)           | nil
-// | 8 | (> 0) <=height        | > height (*)           | vote extensions cannot be modified once enabled
-// | 9 | (> 0) > height        | > height (*)           | nil
+// |  1 | *                    | (nil)                  | nil
+// |  2 | *                    | < 0                    | VoteExtensionsEnableHeight must be positive
+// |  3 | <=0                  | 0                      | nil
+// |  4 | X                    | X                      | nil
+// |  5 | > 0; <=height        | 0                      | vote extensions cannot be disabled once enabled
+// |  6 | > 0; > height        | 0                      | nil (disable a previous proposal)
+// |  7 | *                    | <=height               | vote extensions cannot be updated to a past height
+// |  8 | <=0                  | > height (*)           | nil
+// |  9 | (> 0) <=height       | > height (*)           | vote extensions cannot be modified once enabled
+// | 10 | (> 0) > height       | > height (*)           | nil
 // The table above reflects all cases covered.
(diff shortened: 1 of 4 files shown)


melekes (approved):
👍 

we should also update the docs if we're going to remove this condition. otherwise app devs can easily bump into this again.

sergio-mena:
@melekes Good point. I took a look at the spec, and this behavior wasn't allowed by the text, so I updated the spec's text. Please take a look

melekes:
> @melekes Good point. I took a look at the spec, and this behavior wasn't allowed by the text, so I updated the spec's text. Please take a look

thanks ❤️

types/params.go
```diff
@@ -207,15 +207,16 @@ func (params ConsensusParams) ValidateBasic() error {
...
-// | 4 | > 0; <=height         | 0                      | vote extensions cannot be disabled once enabled
-// | 5 | > 0; > height         | 0                      | nil (disable a previous proposal)
-// | 6 | *                     | <=height               | vote extensions cannot be updated to a past height
-// | 7 | <=0                   | > height (*)           | nil
-// | 8 | (> 0) <=height        | > height (*)           | vote extensions cannot be modified once enabled
-// | 9 | (> 0) > height        | > height (*)           | nil
+// |  1 | *                    | (nil)                  | nil
+// |  2 | *                    | < 0                    | VoteExtensionsEnableHeight must be positive
+// |  3 | <=0                  | 0                      | nil
+// |  4 | X                    | X                      | nil
```
greg-szabo, line 213:
```suggestion
// |  4 | X                    | X (>=0)                | nil
```

If you don't mind, I'd like to clarify that the rule has the implicit knowledge that `updated...EnableHeight` should be positive. (Derived from the previous rules.)

Since not all rules were processed yet where `updated...EnableHeight` can be zero, it's best to claim it's `>=0`.

greg-szabo (approved):
I made two small suggestions to clarify in the comments, but it looks good to me.

types/params.go
sergio-mena: I would like to keep it as is, otherwise we risk breaking the existing behavior in another way: ">=0" is a restriction that didn't exist in `v0.38.2`. If you're keen on adding it, we can leave it on `main` but not backport it to `v0.38.x`

ancazamfir (approved):

alexanderbez:
LGTM

PR #2005, opened 2024-01-09
fix: unflake the majority of flaky tests

faddat:
## Note

I've marked this as ready for review because I think it solves a whole bunch of legitimate test related issues. 

Weather it is merged in this form, or as a bunch of smaller PR's I leave up to the maintainers. 

Since I've marked it as ready for review, I'm also going to make a modification to the number of times that it runs tests -- 5x each, and indeed, until tests are rock solid, I do think that we should merge that to main.  We will get to that goal >5x faster if we run tests 5x per test.

An explicit goal of this work has been to NEVER weaken a test, either make changes that are neutral or make the test more rigorous.  When reviewing, anything that looks like it makes a test weaker, I request that it be treated like a bug.

## Now Contains:

* #2146 + refactor thereof
* #2295 since this PR is really about tests
  * It should be no problem to merge #2295 first, or merge this first, and not merge that. 
* refactor of many inidivdual tests
* reduces the test timeout time to 90 seconds, to conserve capital, mainly human labor costs in the form of waiting for the CI system, which incidentally reduces the cost of using GitHub Actions. 
* a more specific way of handling errors from test byzantine prevote equivocation: I think there may be an issue with "evidence" in general
* CI configuration to run each test once to control ci cost
* CI configuration to split work across 20 workers instead of 6, enabling a reduction in test time from 15 minutes to 90 seconds

## Request from reviewers

I think this is a great candidate for one of two merge approaches:

1) we merge it, and be done with it, I think there is a case for that here, after removing the 40x bits in the .github/workflows folder
2) We break it into chunks and merge those, I think there's a case for that as well.

I'm happy to do either, so please let me know which is preferable, though since I did get to the goal of 100% passage, I figure it makes the most sense to merge as-is.

---

#### PR checklist

- [x] Tests written/updated
- [x] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments

PR #2120, opened 2024-01-24
ci: run the tests 40 times (with pebble)

faddat:
# tl;dr

If it doesn't pass 40 times, it simply doesn't pass.  Running the tests 40x before any merge will clarify the situation and prevent us from putting trust in the code when that's not warranted. 

## Run tests 40x by default
This PR runs cometbft tests 40 times instead of once.  The logic behind this PR is that by running the tests 40 times, we can identify issues like those found in v0.38.x before code is shipped, instead of after.

In general, we will find that this will reveal tests that depend on timing in ways they likely should not, broken code under test, or tests that do not work correctly.

I recommend that we merge this into all of comet's release branches, so that:

* nightly ci runs don't provide a false sense of security
* we can all get a better understanding of which tests fail and why

I can also say that this approach does not add significant time to PR's.  Once we've found that the tests are running well across all release branches, we could reduce the number of test runs. 

I think that the key learning about tests that I've gotten -- and am hoping to transmit, from:

* #2005 
* #1930 
* #1954 

Is that the majority of test failures relate to timing, and that the environment that the tests should pass in by default is github's CI system.  

I am also suggesting this for inclusion to main, because I think that this makes the majority of PR's fail.  I think that is easily resolved, but will require some attention to the tests that are failing:

* #1923 

Please note that the goal of this PR is NOT to block merges.  Instead, it is to ensure that we provide downstream with the stability needed to thrive.  When a PR is merged into comet, we should **know** that it improves the state of the code, and that all tests pass in github CI, so that downstream can trust their own tests. 

It occurred to me today that I still don't know root cause on a lot of these test failures, and my "40x" pr series has provided me with a lot of unique knowledge.  One learning I got in the past few weeks is that we might need to refactor some tests, because we may have unrealistic expectations for test behavior.  

Even more interesting is the fact that this is the first time that CometBFT has passed all 40 tests.

~~Might indicate that some recent changes made performance improvements that allowed every test to pass.~~

**actually indicates that we should NOT run tests conditionally, watch the next commit**

### Why shouldn't we run tests conditionally?

We shouldn't use 

```yaml
        if: env.GIT_DIFF
```

for the following reasons:

* machine time is cheaper than human time and if it isn't that's a bug
* a change in one of the six areas for testing, might affect tests in the other five but if we continue to use ```if: env.GIT_DIFF``` this will never show up in CI
* reviewers base assessments on the output of the CI system

### Alternative

I am going to make another PR that **only** removes

```yaml
        if: env.GIT_DIFF
```

But I don't prefer this alternative because I don't think it goes as far to address root causes.  Surely, if we are familiar with the way that various components have affects on different pieces of the stack, then it can easily be seen that it is **_folly of the highest order_** to split tests so that some times, some of them do not run.

A change in the 1/6th can break things in the 5/6th and the conditional running of tests is "how we got here"

* #2121 

---

#### PR checklist

- [ ] Tests written/updated


melekes:
Closing as completed.

PR #2121, opened 2024-01-24
fix: do not run tests conditionally

faddat:
This PR ensures that reviewers are not fooled by green check marks that should not exist by ensuring that comet does not run tests conditionally based on a diff.  A change in a teeny piece of the codebase can break things in other pieces of the codebase, thus:

```yaml
        if: env.GIT_DIFF
```

is folly and dangerous.

Please see #2120 for a very detailed description of the issue and note that #2120 is my preferred remedy.

### This is a mind hack

The green check mark makes all of us feel comfortable.  We want to merge stuff that earns the green check mark. The eyes have the optic nerve -- the highest bandwidth path to our brains where we make merge/don't merge decisions.   The CI system is currently designed to deceive our minds by presenting a green check mark at the wrong times.  It must be meaningfully earned.

I couldn't really figure out why this issue was largely hidden from the maintainership team but running tests conditionally is the precise reason.

While I prefer to merge a PR that runs the tests 40 times, and ensures they're not conditional, this is the more important of the two changes and removes the blinder from our brains.

### This is a bug fix

the reason that this is a bug fix is that what we have here is a **bug** that **causes other bugs**

### My thoughts while working on the `run tests 40 times` PR series

Over and over thinking how can they not know about this, what is wrong, the tests pretty much don't work and can't be relied upon to function. How is it that the maintenance team doesn't know about this?

This is how.

a commit might change 1% of the codebase, causing only 1/6th of the tests to run. 

It could cause problems in the other 5/6th of the codebase, which relies on that 1% to function in some clearly defined manner.

---

#### PR checklist

- [ ] Tests written/updated


faddat:
Oh look the tests pass!

What changed between commits here?

**nothing**

faddat:
* https://github.com/cometbft/cometbft/actions/runs/7649974837/job/20845289857?pr=2123

faddat:
note: there's now been another flake out, above. 

If the code does not change, the result of running tests should not change.

faddat:
....and another.  Maybe it will pass this time.  Maybe not.

melekes:
Thank you for your contribution and the effort you've put into this pull request. I appreciate the intent of removing the conditional run.

However, I'm currently leaning towards not merging this PR. The reason is that removing the conditional run may lead to more frequent setup failures, especially in pull requests that only modify the documentation rather than the source code. There may be better solutions than running tests on every occasion to address non-deterministic failures.

**The practical approach would focus on resolving the underlying issue of non-determinism in the tests. Doing so allows us to maintain a more reliable testing environment without necessarily resorting to unconditional test runs.**

PR #2130, opened 2024-01-25
refactor: simplify large functions

faddat:
This PR will enable some new linters, and be based off of:

main + #1907

... so #1907 should be merged before this one, because that will greatly reduce the number of changed files from 203 to somewhere in the teens.  If at any time it's needed to break any of this into its own PR's, just let me know.

## New Linters

* nestif set to 15
* cyclomatic complexity via revive set to 71
* cognitive complexity via revive set to 71

I got (in my opinion) amazing results from this.  After the configuration, I was guided to what are more or less known trouble areas in comet, nearly every time.  

## Ongoing plan

Over time, we can lower the complexity rules and nestif to their defaults, and I think that this would likely have a great effect on the software overall.  

## Notes

I really liked some of the changes I started to get as I worked with these linters, which are all related to each other to some degree.  

This PR takes some of the largest functions in cometbft, and breaks them into smaller pieces.  

When I started it, I wasn't too sure, but by using this approach I was definitely able to make changes that seem to:

* make things more reliable
* make things easier to understand

## Changes unique to this PR

This PR takes anything that the nestif/cyclomaticcomplexity/cognitivecomplexity linters finds to be too complex, and refactors it into individual functions.  I wasn't too sure about this course of action when I started, but in a number of places, the results I started to get are (I think) quite convincing.  

In places where the actual code has been changed, I'm finding that the reworked style is **actually** easier to read and think about.

It's likely that I will write out the whole PR, kind of like 

* #1954 

And then submit the whole thing as a series of smaller patches. 

For example, this technique has led me to poolRoutine, which is indeed a large and complex function.  It runs consistently, and it is probably better for many reasons to break it into separate functions.  

---

#### PR checklist

- [ ] Tests written/updated


faddat:
note: I'm not sure that these linters will improve anything.  I am not very pleased with my first round, but will play with them some more.

faddat:
update: these seem to improve things now that I've gotten a hang of it.  

the `nestif` linter makes this PR rather broad, so I'm doing it commit by commmit so that we can cherry pick anything desired/undesired.

faddat:
This PR can also be used as a case study in why we should fix the tests so that they consistently pass.  Writing it is mainly struggling to tell the difference between existing problems, and ones I could have caused while refactoring.

faddat:
will break all poolRoutine / blocksync changes into their own PR.  I think they're ready, and its... really good.

faddat:
Note: this is likely ready to be merged, though if we did merge it we would want to turn off the linters enabled in this PR:

* nestif
* cyclomatic complexity
* cognitive complexity

and we would also want to merge #1954

melekes:
I can see that you've already started breaking this PR into smaller PRs. I think this is the right direction 👍

faddat:
If it's no different to you, I would like to reopen this as a draft pull request for tracking purposes.

This PR enables several of the linters that you mentioned in #2274 .

Would you like to go linter by linter or how do you want to do it?

Issue #2134, opened 2024-01-25
Deprecated Dockerfiles in the CometBFT repository


faddat:
## Bug Report

* https://github.com/cometbft/cometbft/pull/2132

Not 100% sure this is a bug but this issue is closed by:

* #2135

faddat:
Actually, please leave this issue here but I need to expand on it and I might fix it in a couple of PRS from now. Basically the end-to-end test docker file is actually based off of the database docker file, but I really can't say for sure why we have any other docker files at all.

PR #2135, opened 2024-01-25
chore: remove amazon linux Dockerfile and update test/docker/Dockerfile

faddat:
* amazon linux Dockerfile is configured for go v1.12.x -- it is clearly unused
* tests Dockerfile is configured for go v1.15.x -- it's clearly unused

The presence of these files is a bug/technical debt or such.  They make the code harder to read and understand and create the impression that they may serve a purpose.

closes: #2134 

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

andynog (changes requested):
Should the `Makefile` in the root of the repo also be updated ? I see this entry there 

```
build_c-amazonlinux:
	$(MAKE) -C ./DOCKER build_amazonlinux_buildimage
	docker run --rm -it -v `pwd`:/cometbft cometbft/cometbft:build_c-amazonlinux
.PHONY: build_c-amazonlinux
```

faddat:
Nice catch @andynog -- I'll address that now

faddat:
@andynog I just reviewed the Pr and I do believe that it updates the make file as well.

faddat:
ah :).  Also that one lol :)

melekes:
`test/docker/Dockerfile` is used by `test_integrations` cmd, which I'm not sure we use anymore. So we should remove it along with any sub-commands it calls.

faddat:
Makes sense to me. I'll have a look.

faddat:
@melekes -- the Dockerfile was broken anyhow, as it tried to build `make tools` which no longer exists.  

I think it might make sense to put the rest of this into an issue, unless you think it is alright to change the e2e image to include a folder, let's call it `/testnet` in which case I think that we could make this a part of the e2e image.  

So, now instead of removing this extra Dockerfile, it's fixed, so that when we next make a release, that process will go smoothly.

melekes (approved):
👍

PR #2136, opened 2024-01-26
fix: `Rollback`: wrong modification of `state.LastHeightValidatorsChanged` while rollback at a special height

cosinlink:
closes: #2137 #1074

## Description
Rollback at a specific height causes the node to fail to start. This is caused by the error code: 
https://github.com/cometbft/cometbft/blob/635d0b596c7300c98caa57e7aab26a7a4579ab97/internal/state/rollback.go#L71

According to the definition of the `State` struct as below
https://github.com/cometbft/cometbft/blob/635d0b596c7300c98caa57e7aab26a7a4579ab97/internal/state/state.go#L63
`LastHeightValidatorsChanged` satisfy the following condition under any circumstances
1 <= `state.LastHeightValidatorsChanged` <= `state.LastBlockHeight + 1 + 1`

this condition can also be confirmed by looking at the code in `func (store dbStore) save(state State, key []byte) error`
https://github.com/cometbft/cometbft/blob/635d0b596c7300c98caa57e7aab26a7a4579ab97/internal/state/store.go#L229

`state.LastHeightValidatorsChanged` is with `nextHeight + 1`(`state.LastBlockHeight + 1 + 1`) corresponds.

When we rollback from block height `state.LastHeightValidatorsChanged - 1` to height `state.LastHeightValidatorsChanged - 2`, `newState.LastHeightValidatorsChanged` would be wrong to modified into the `state.LastHeightValidatorsChanged - 1`
 

## Issue
Let me give an example to describe the issue:

 
1. Assume that the current height of the block is 243, current state is `state2` and `state2.LastHeightValidatorsChanged` == 244, validatorSet changed to 4 validators in the block 244 and state2

2. Rollback to block 242 and state1, the target `rollbackHeight` == 242

3. Since the error code, https://github.com/cometbft/cometbft/blob/635d0b596c7300c98caa57e7aab26a7a4579ab97/internal/state/rollback.go#L71
`state1.LastHeightValidatorsChanged` was changed to 243 incorrectly and it should be 244

4. After rollbacked to block 242, node could not start properly

## Solution
See code for details.

#### PR checklist

- [x] Tests written/updated


adizere:
Thanks for the contribution @cosinlink, we'll have a look and review this soon.

melekes (approved):
Thanks @cosinlink ❤️ 

Could you add a changelog entry `.changelog/unreleased/bug-fixes/2136-fix-rollback.md`?

internal/state/rollback.go
```diff
@@ -65,10 +65,11 @@ func Rollback(bs BlockStore, ss Store, removeBlock bool) (int64, []byte, error)
 		return -1, nil, err
 	}
 
+	nextHeight := rollbackHeight + 1
 	valChangeHeight := invalidState.LastHeightValidatorsChanged
 	// this can only happen if the validator set changed since the last block
```
cason, line 70: Do you mind updating the comment here. Mentioning, in short, how we define `valChangeHeight`? You may also add a link to the existing issue.

cason (commented):
I think the fix is correct, although:
1. I am curious how tests were passing before this fix
2. I would prefer having a test unit revealing the bug that is fixed here.

internal/state/rollback.go
```diff
@@ -65,10 +65,11 @@ func Rollback(bs BlockStore, ss Store, removeBlock bool) (int64, []byte, error)
 		return -1, nil, err
 	}
 
+	nextHeight := rollbackHeight + 1
```
sergio-mena, line 68: I know this is unrelated to this PR, but, why are we:

* loading the previous `LastValidatorSet` from `rollbackHeight`, and
* loading the previous `ConsensusParams`from `rollbackHeight + 1` (i.e., `invalidState.LastBlockHeight`)

@melekes what do you think? I can't find an explanation, and might be a similar bug lurking

internal/state/rollback.go
```diff
@@ -65,10 +65,11 @@ func Rollback(bs BlockStore, ss Store, removeBlock bool) (int64, []byte, error)
...
 	valChangeHeight := invalidState.LastHeightValidatorsChanged
 	// this can only happen if the validator set changed since the last block
-	if valChangeHeight > rollbackHeight {
-		valChangeHeight = rollbackHeight + 1
+	if valChangeHeight > nextHeight+1 {
+		valChangeHeight = nextHeight + 1
```
sergio-mena, line 72: If I run the unit tests changing this line with this:
```go
valChangeHeight = nextHeight + 2
```
which is the correct one?

internal/state/rollback.go
sergio-mena: I'm working on #2283 to be able to test this (and repro the problem) using `e2e`. I still need to introduce a "rollback" perturbation to `e2e` functionality. Will get to when I'm back.

cosinlink:
> Thanks @cosinlink ❤️
> 
> Could you add a changelog entry `.changelog/unreleased/bug-fixes/2136-fix-rollback.md`?

ok, I will add a changelog

internal/state/rollback.go
cosinlink: @sergio-mena,  different definition by origin author,  you could check these code to compare  `LastValidatorSet` with `ConsensusParams`,
`LastValidatorSet`: https://github.com/cometbft/cometbft/blob/635d0b596c7300c98caa57e7aab26a7a4579ab97/internal/state/store.go#L229

`ConsensusParams`: https://github.com/cometbft/cometbft/blob/635d0b596c7300c98caa57e7aab26a7a4579ab97/internal/state/store.go#L233

internal/state/rollback.go
cosinlink: `valChangeHeight = nextHeight + 1` is right, if it changed to `valChangeHeight = nextHeight + 2`, will cause error in some edge case

sergio-mena (approved):
Thanks for answering my questions. I still feel the need to test this via blackbox techniques, as the UTs are mocking a lot of things.
I'm approving this PR now, and will try to drive #2283 to completion so that we can test the rollback command in CI (`e2e`) with constant changes in validator sets and consensus params.

melekes:
@Mergify backport v0.37.x

PR #1950, opened 2024-01-02
run tests 40x against the unmodified master branch

faddat:
My original "run the linter against the code" PR contained many, many changes from a linting session.  So, as with the others, please do not merge this. 

It's being used to find and fix flaky tests.

Probably we don't want to merge this, though it might be interesting as a temporary measure till we have gotten all of the tests (or underlying code) tuned right)

---

#### PR checklist

- [ ] Tests written/updated
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments

PR #1940, opened 2024-01-02
Run comet tests on v0.38.x, which is in production, 40x

faddat:
This PR will run the comet tests on v0.38.x 40 times.  

---

#### PR checklist

- [ ] Tests written/updated
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments

PR #2146, opened 2024-01-26
feat: break out testbyzantineprevoteequivocation

faddat:
This PR breaks out testbyzantineprevotequivocation.

tbpe is the number 1 test failure in the comet repository as can be seen:

* #1954 
* #2120 
* #1936 
* #1950 
* #1940 

and other similar branches.  This PR improves the failure rate of tbpe without changing what it does.  

Failure rate is improved, but the issue is not yet solved.  Much easier to work on with the test in its own file, however.

I'll take this PR out of draft when the changes have been better isolated.

This is a checkout of internal/consensus from #1954 with changes unrelated to tbpa removed, except for a handy comment in mempool_test.go

---

#### PR checklist

- [ ] Tests written/updated


faddat:
Note:

* https://github.com/cometbft/cometbft/security/advisories/GHSA-hq58-p9mv-338c

Is not what Notional reported, and we disagree with all claims made in that security advisory.

faddat:
closed in favor of #2005 since that has updated versions of this

PR #2147, opened 2024-01-26
fix(flowrate): fix non-determinism in flowrate tests

faddat:
This PR addresses test reliability issues in testreader and testwriter. 

From: #1954 

---

#### PR checklist

- [ ] Tests written/updated


internal/flowrate/io_test.go
```diff
@@ -198,3 +233,22 @@ func ratesAreEqual(r1 int64, r2 int64) bool {
 	}
 	return false
 }
+
+func statusesAreEqualWithinTolerance(s1 *Status, s2 *Status, maxDeviationForDuration time.Duration, maxDeviationForRate int64) bool {
```
melekes, line 237: Can't we modify `statusesAreEqual`? Having two functions that do the same thing seems excessive.

faddat: Think so?  I'll check on my flight.

faddat: hey, it seems my response did not post -- the answer to your question is"

- "if you are oaky with all the things being statusesAeEqualWithinTolerance" then yep, we can do that.  Currently equal means exactly equal and within tolerance gives them some wiggle room.

mzabaluev:
It looks like the flakiness still [pops up](https://github.com/cometbft/cometbft/actions/runs/8169417865/job/22333459466?pr=2527) in `v1.x` despite this having been backported in #2248.

PR #1587, opened 2023-11-09
e2e: Implement latency emulation for DigitalOcean

hvanz:
Part of #1559

Latency emulation is currently supported only for the Docker infrastructure. This PR implements the methods for supporting latency emulation in DigitalOcean.

---

#### PR checklist

- [X] Manual testing (depends on https://github.com/cometbft/qa-infra/issues/27)
- [ ] Tests written/updated
- [X] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments

PR #2148, opened 2024-01-26
fix(e2e): path to latency script files in DO

hvanz:
Running testnets with latency emulation in DO is failing because of we are looking for the script files in the wrong location.

Related to #1587

---

#### PR checklist

- [ ] Tests written/updated


faddat:
Practice how you play: 

do not run these testnets with comet only, in DO.

Run these testnets with comet+sdk+wasmd, on the internet.

melekes (approved):

PR #2150, opened 2024-01-26
fix: simultaneous rpc test failures

faddat:
This PR should resolve the odd case where a group of rpc tests fail simultaneously.

from: #1954 

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
Thanks @faddat ❤️

PR #2156, opened 2024-01-28
refactor(privval): reverse conditional + more idiomatic Go code with early returns

### LINKED ISSUE #2154: style: privval/FilePVLastSignState.CheckHRS can use early returns and refactor by reverse conditional for more idiomatic Go code to reduce nesting
If we examine this code https://github.com/cometbft/cometbft/blob/f4d73cd5a091a997d1f040850710b6937e650125/privval/file.go#L100-L131 there is a bunch of unnecessary nesting which can be made much neater by

```go
func (lss *FilePVLastSignState) CheckHRS(height int64, round int32, step int8) (bool, error) {
	if lss.Height > height {
		return false, fmt.Errorf("height regression. Got %v, last height %v", height, lss.Height)
	}

	if lss.Height != height {
		return false, nil
	}

	if lss.Round > round {
		return false, fmt.Errorf("round regression at height %v. Got %v, last round %v", height, round, lss.Round)
	}

	if lss.Round != round {
		return false, nil
	}

	if lss.Step > step {
		return false, fmt.Errorf(
			"step regression at height %v round %v. Got %v, last step %v",
			height,
			round,
			step,
			lss.Step,
		)
	}

	if lss.Step != step {
		return false, nil
	}

	if lss.SignBytes == nil {
		return false, errors.New("no SignBytes found")
	}
	if lss.Signature == nil {
		panic("pv: Signature is nil but SignBytes is not!")
	}
	return true, nil
}
```

odeke-em:
This change applies the refactoring method of reverse conditional and using early returns for more idiomatic code which makes the code much more direct and less nested.

Fixes #2154

### DIFF
--- a/privval/file.go
+++ b/privval/file.go
@@ -102,32 +102,40 @@ func (lss *FilePVLastSignState) CheckHRS(height int64, round int32, step int8) (
 		return false, fmt.Errorf("height regression. Got %v, last height %v", height, lss.Height)
 	}
 
-	if lss.Height == height {
-		if lss.Round > round {
-			return false, fmt.Errorf("round regression at height %v. Got %v, last round %v", height, round, lss.Round)
-		}
-
-		if lss.Round == round {
-			if lss.Step > step {
-				return false, fmt.Errorf(
-					"step regression at height %v round %v. Got %v, last step %v",
-					height,
-					round,
-					step,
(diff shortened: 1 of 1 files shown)


melekes (approved):
Thanks @odeke-em ❤️

PR #2159, opened 2024-01-28
fix(p2p): make PeerSet.Remove more efficient

### LINKED ISSUE #2157: p2p: PeerSet.Remove naively & inefficiently creates new slices each time discarding all the work preallocating, but could simply re-slice
## Bug Report
Noticed during one of my audits and security vulnerability hunts that this code https://github.com/cometbft/cometbft/blob/f4d73cd5a091a997d1f040850710b6937e650125/p2p/peer_set.go#L34-L39 pre-allocates a slice with capacity of 256 and size of 0, for performance reasons, and then add continues with that trend https://github.com/cometbft/cometbft/blob/f4d73cd5a091a997d1f040850710b6937e650125/p2p/peer_set.go#L57

However, take a look at .Remove https://github.com/cometbft/cometbft/blob/f4d73cd5a091a997d1f040850710b6937e650125/p2p/peer_set.go#L123-L140 which to pop a peer out of the slice, naively creates a fresh slice, copies all the elements, but really it could simply use Go's slicing.

This code has been like this since 8 years ago, essentially the birth of Tendermint per first commit https://github.com/cometbft/cometbft/commit/abc3a2cc3c5ec777fecb5e0eac551ad17bdb1c6e

## Remedy
Given that that code hasn't been touched since Tendermint's first commit, we can deduce that perhaps it was that we didn't understand Go more, because re-slicing is a common operation in Go per https://go.dev/blog/slices-intro 

PeerSet.Remove should simply use re-slicing like this

```diff
diff --git a/p2p/peer_set.go b/p2p/peer_set.go
index ba43f6417..4f1501d31 100644
--- a/p2p/peer_set.go
+++ b/p2p/peer_set.go
@@ -119,25 +119,25 @@ func (ps *PeerSet) Remove(peer Peer) bool {
 		return false
 	}
 
+	if len(ps.list) == 0 {
+		return false
+	}
+
 	index := item.index
-	// Create a new copy of the list but with one less item.
-	// (we must copy because we'll be mutating the list).
-	newList := make([]Peer, len(ps.list)-1)
-	copy(newList, ps.list)
-	// If it's the last peer, that's an easy special case.
+	lastPeer := ps.list[len(ps.list)-1]
+	ps.list[len(ps.list)-1] = nil // nil the array's entry to ensure it isn't reachable in the array, hence can be GC'd.
 	if index == len(ps.list)-1 {
-		ps.list = newList
+		ps.list = ps.list[:len(ps.list)-1]
 		delete(ps.lookup, peer.ID())
 		return true
 	}
 
 	// Replace the popped item with the last item in the old list.
-	lastPeer := ps.list[len(ps.list)-1]
 	lastPeerKey := lastPeer.ID()
 	lastPeerItem := ps.lookup[lastPeerKey]
-	newList[index] = lastPeer
+	ps.list[index] = lastPeer
+	ps.list = ps.list[:len(ps.list)-1]
 	lastPeerItem.index = index
-	ps.list = newList
 	delete(ps.lookup, peer.ID())
 	return true
 }
```

which produces these performance improvements

we get back these improved performance

```shell
name                 old time/op    new time/op    delta
PeerSetRemoveOne-8      100µs ± 9%      96µs ±13%     ~     (p=0.218 n=10+10)
PeerSetRemoveMany-8    1.56ms ± 2%    1.50ms ± 1%   -3.76%  (p=0.000 n=9+8)

name                 old alloc/op   new alloc/op   delta
PeerSetRemoveOne-8     8.48kB ± 0%    7.92kB ± 0%   -6.60%  (p=0.000 n=9+10)
PeerSetRemoveMany-8     149kB ± 0%      65kB ± 0%  -56.44%  (p=0.000 n=10+10)

name                 old allocs/op  new allocs/op  delta
PeerSetRemoveOne-8       85.0 ± 0%      73.0 ± 0%  -14.12%  (p=0.000 n=10+10)
PeerSetRemoveMany-8     1.32k ± 0%     1.22k ± 0%   -7.51%  (p=0.000 n=10+10)
```

that are reproducible from down below:
## Exhibit
If we run such a benchmark
```go
package p2p
                
import (
        "testing"
)

var sink any = nil
        
func BenchmarkPeerSetRemoveOne(b *testing.B) {
        b.ReportAllocs()
        
        for i := 0; i < b.N; i++ {
                testPeerSetAddRemoveOne(b)
                sink = i
        }       
                
        if sink == nil {
        }       
                
        // Reset the sink.
        sink = nil
}       
        
func BenchmarkPeerSetRemoveMany(b *testing.B) {
        b.ReportAllocs()
                
        for i := 0; i < b.N; i++ {
                testPeerSetAddRemoveMany(b)
                sink = i
        }
        
        if sink == nil {
        }
        
        // Reset the sink. 
        sink = nil
}

func testPeerSetAddRemoveOne(tb testing.TB) {
        
        peerSet := NewPeerSet()
                
        var peerList []Peer
        for i := 0; i < 5; i++ {
                p := newMockPeer(net.IP{127, 0, 0, byte(i)})
                if err := peerSet.Add(p); err != nil {
                        tb.Error(err)
                }       
                peerList = append(peerList, p)
        }

        n := len(peerList)
        // 1. Test removing from the front
        for i, peerAtFront := range peerList {
                removed := peerSet.Remove(peerAtFront)
                assert.True(tb, removed)
                wantSize := n - i - 1
                for j := 0; j < 2; j++ { 
                        assert.Equal(tb, false, peerSet.Has(peerAtFront.ID()), "#%d Run #%d: failed to remove peer", i, j)
                        assert.Equal(tb, wantSize, peerSet.Size(), "#%d Run #%d: failed to remove peer and decrement size", i, j)
                        // Test the route of removing the now non-existent element
                        removed := peerSet.Remove(peerAtFront)
                        assert.False(tb, removed)
                }
        }

        // 2. Next we are testing removing the peer at the end
        // a) Replenish the peerSet
        for _, peer := range peerList {
                if err := peerSet.Add(peer); err != nil {
                        tb.Error(err)
                }
        }

        // b) In reverse, remove each element
        for i := n - 1; i >= 0; i-- {
                peerAtEnd := peerList[i]
                removed := peerSet.Remove(peerAtEnd)
                assert.True(tb, removed)
                assert.Equal(tb, false, peerSet.Has(peerAtEnd.ID()), "#%d: failed to remove item at end", i)
                assert.Equal(tb, i, peerSet.Size(), "#%d: differing sizes after peerSet.Remove(atEndPeer)", i)
        }
}


func testPeerSetAddRemoveMany(tb testing.TB) {
        peerSet := NewPeerSet()

        peers := []Peer{}
        N := 100
        for i := 0; i < N; i++ {
                peer := newMockPeer(net.IP{127, 0, 0, byte(i)})
                if err := peerSet.Add(peer); err != nil {
                        tb.Errorf("failed to add new peer")
                }
                if peerSet.Size() != i+1 {
                        tb.Errorf("failed to add new peer and increment size")
                }
                peers = append(peers, peer)
        }       
                
        for i, peer := range peers {
                removed := peerSet.Remove(peer)
                assert.True(tb, removed)
                if peerSet.Has(peer.ID()) {
                        tb.Errorf("failed to remove peer")
                }
                if peerSet.Size() != len(peers)-i-1 {
                        tb.Errorf("failed to remove peer and decrement size")
                }
        }
}
```

## FYI & security considerations
There is a possible security issue/vector due to this behavior but that requires a whole new discussion and a different mitigation that would be slightly mitigated by the fix I suggested above, but to completely solve it, it would require a design addition which I I privately communicated.

Kindly cc-ing @elias-orijtech @ValarDragon

### LINKED ISSUE #2158: p2p: Switch.OnStop uses stale and already mutated PeerSet.List() which could have data race consequences due to a leaky abstraction of using .List()
## Bug Report
The code in p2p/Switch.OnStop invokes sw.peers.List() per https://github.com/cometbft/cometbft/blob/f4d73cd5a091a997d1f040850710b6937e650125/p2p/switch.go#L251 
and if we look at the code for .List() https://github.com/cometbft/cometbft/blob/f4d73cd5a091a997d1f040850710b6937e650125/p2p/peer_set.go#L153-L156 which returns a copy of the slice while under a lock but after handing it over, it is no longer under a lock and is modified by .Remove() which causes .OnStop() to read stale data.

I found this problem while working on the correct fix for https://github.com/cometbft/cometbft/issues/2157 in which the correct way to shorten a slice is to ensure you insert a nil at the last element before reslicing to ensure that there are no references to it and that the backing array can be garbage collected i.e.

```go
endItem := slice[len(slice)-1]
slice[len(slice)-1] = nil // nil it so that it'll be garbage collected
```

I can reliably reproduce this issue when I apply the fix for the referenced issue

## Remedy
Check and ensure that the peer is non-nil.

odeke-em:
This change makes PeerSet.Remove much more efficient simply by using more idiomatic Go re-slicing to avoid the prior mechanisms of creating fresh peer set lists on just a single remove. While here also added a remedy for a found bug #2158 due to an abstraction that returns a stale slice to its caller in Switch.OnStop.

Benchmark results:

```shell
name                 old time/op    new time/op    delta
PeerSetRemoveOne-8     90.5µs ± 4%    95.9µs ±13%     ~     (p=0.218 n=10+10)
PeerSetRemoveMany-8    1.58ms ± 4%    1.50ms ± 1%   -4.98%  (p=0.000 n=10+8)

name                 old alloc/op   new alloc/op   delta
PeerSetRemoveOne-8     8.48kB ± 0%    7.92kB ± 0%   -6.60%  (p=0.000 n=10+10)
PeerSetRemoveMany-8     149kB ± 0%      65kB ± 0%  -56.44%  (p=0.000 n=10+10)

name                 old allocs/op  new allocs/op  delta
PeerSetRemoveOne-8       85.0 ± 0%      73.0 ± 0%  -14.12%  (p=0.000 n=10+10)
PeerSetRemoveMany-8     1.32k ± 0%     1.22k ± 0%   -7.51%  (p=0.000 n=10+10)
```

which savings become so much more when peers are removed much more frequently for a longer period of time and could mitigate DOS vectors.

Fixes #2157
Fixes #2158

### DIFF
--- a/p2p/peer_set.go
+++ b/p2p/peer_set.go
@@ -109,7 +109,7 @@ func (ps *PeerSet) Remove(peer Peer) bool {
 	defer ps.mtx.Unlock()
 
 	item := ps.lookup[peer.ID()]
-	if item == nil {
+	if item == nil || len(ps.list) == 0 {
 		// Removing the peer has failed so we set a flag to mark that a removal was attempted.
 		// This can happen when the peer add routine from the switch is running in
 		// parallel to the receive routine of MConn.
@@ -120,24 +120,22 @@ func (ps *PeerSet) Remove(peer Peer) bool {
 	}
 
 	index := item.index
-	// Create a new copy of the list but with one less item.
-	// (we must copy because we'll be mutating the list).
-	newList := make([]Peer, len(ps.list)-1)
-	copy(newList, ps.list)
+	lastPeer := ps.list[len(ps.list)-1]
+	ps.list[len(ps.list)-1] = nil // nil the last entry of the slice to shorten, so it isn't reachable & can be GC'd.
+
 	// If it's the last peer, that's an easy special case.
 	if index == len(ps.list)-1 {
-		ps.list = newList
+		ps.list = ps.list[:len(ps.list)-1]
 		delete(ps.lookup, peer.ID())
 		return true
 	}
 
 	// Replace the popped item with the last item in the old list.
-	lastPeer := ps.list[len(ps.list)-1]
 	lastPeerKey := lastPeer.ID()
 	lastPeerItem := ps.lookup[lastPeerKey]
-	newList[index] = lastPeer
+	ps.list[index] = lastPeer
+	ps.list = ps.list[:len(ps.list)-1]
 	lastPeerItem.index = index
-	ps.list = newList
 	delete(ps.lookup, peer.ID())
 	return true
 }
--- a/p2p/bench_test.go
+++ b/p2p/bench_test.go
@@ -1,0 +1,39 @@
+package p2p
+
+import (
+	"testing"
+)
+
+var sink any = nil
+
+func BenchmarkPeerSetRemoveOne(b *testing.B) {
+	b.ReportAllocs()
+
+	for i := 0; i < b.N; i++ {
+		testPeerSetAddRemoveOne(b)
+		sink = i
+	}
+
+	if sink == nil {
+		b.Fatal("Benchmark did not run!")
+	}
+
+	// Reset the sink.
+	sink = nil
+}
+
+func BenchmarkPeerSetRemoveMany(b *testing.B) {
+	b.ReportAllocs()
+
+	for i := 0; i < b.N; i++ {
+		testPeerSetAddRemoveMany(b)
+		sink = i
+	}
+
+	if sink == nil {
+		b.Fatal("Benchmark did not run!")
+	}
+
+	// Reset the sink.
+	sink = nil
+}
--- a/p2p/peer_set_test.go
+++ b/p2p/peer_set_test.go
@@ -48,13 +48,19 @@ func newMockPeer(ip net.IP) *mockPeer {
 }
 
 func TestPeerSetAddRemoveOne(t *testing.T) {
+	testPeerSetAddRemoveOne(t)
+}
+
+func testPeerSetAddRemoveOne(tb testing.TB) {
+	tb.Helper()
+
 	peerSet := NewPeerSet()
 
 	var peerList []Peer
 	for i := 0; i < 5; i++ {
 		p := newMockPeer(net.IP{127, 0, 0, byte(i)})
 		if err := peerSet.Add(p); err != nil {
-			t.Error(err)
+			tb.Error(err)
 		}
 		peerList = append(peerList, p)
 	}
@@ -63,14 +69,14 @@ func TestPeerSetAddRemoveOne(t *testing.T) {
 	// 1. Test removing from the front
 	for i, peerAtFront := range peerList {
 		removed := peerSet.Remove(peerAtFront)
-		assert.True(t, removed)
+		assert.True(tb, removed)
 		wantSize := n - i - 1
 		for j := 0; j < 2; j++ {
-			assert.False(t, false, peerSet.Has(peerAtFront.ID()), "#%d Run #%d: failed to remove peer", i, j)
-			assert.Equal(t, wantSize, peerSet.Size(), "#%d Run #%d: failed to remove peer and decrement size", i, j)
+			assert.False(tb, false, peerSet.Has(peerAtFront.ID()), "#%d Run #%d: failed to remove peer", i, j)
+			assert.Equal(tb, wantSize, peerSet.Size(), "#%d Run #%d: failed to remove peer and decrement size", i, j)
 			// Test the route of removing the now non-existent element
 			removed := peerSet.Remove(peerAtFront)
-			assert.False(t, removed)
+			assert.False(tb, removed)
 		}
 	}
 
@@ -78,7 +84,7 @@ func TestPeerSetAddRemoveOne(t *testing.T) {
 	// a) Replenish the peerSet
 	for _, peer := range peerList {
 		if err := peerSet.Add(peer); err != nil {
-			t.Error(err)
+			tb.Error(err)
 		}
 	}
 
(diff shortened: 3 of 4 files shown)


p2p/peer_set.go
```diff
@@ -119,25 +119,27 @@ func (ps *PeerSet) Remove(peer Peer) bool {
 		return false
 	}
 
+	if len(ps.list) == 0 {
+		return false
```
melekes, line 123: Shouldn't we panic here? I mean, if the `item` is not `nil`, then there must be at least 1 peer in `ps.list`, no?

melekes:
```suggestion
                peer.SetRemovalFailed()
		return false
```

at the very least we should call `peer.SetRemovalFailed()` (for same reasons as above)

melekes (commented):
Thanks @odeke-em ❤️

p2p/peer_set.go
odeke-em: Gotch and thank you @melekes! Updated

p2p/peer_set.go
```diff
@@ -119,25 +119,33 @@ func (ps *PeerSet) Remove(peer Peer) bool {
 		return false
 	}
 
+	if len(ps.list) == 0 {
```
melekes, line 122: let's combine with above condition

melekes:
```
if item == nil || len(ps.list) == 0 {
```

melekes (approved):

melekes:
@odeke-em can you update the PR or allow core developers to update your fork so we can merge latest main into your branch?

melekes:
Closing in favor of https://github.com/cometbft/cometbft/pull/2246

PR #2181, opened 2024-01-29
chore: fix typos

omahs:
chore: fix typos


cason:
Most of this PR applies to old ADRs from the previous repository (Tendermint Core). I don't see benefits on fixing typos on them.

PR #2190, opened 2024-01-30
fix(abci): fix ABCI grammar (backport #1829)

melekes:
This PR closes https://github.com/cometbft/cometbft/issues/1581.
We discovered that the scenarios displayed in https://github.com/cometbft/cometbft/issues/1581 are not caused by a bug in the CometBFT but in the ABCI grammar. We fixed the grammar and updated the code for e2e tests accordingly. After doing this, the e2e tests that check whether CometBFT respects ABCI grammar are passing.


sergio-mena (approved):
Thanks for this!

PR #2201, opened 2024-01-31
test(abci): update e2e tests to check against ABCI 2.0 grammar

### LINKED ISSUE #1404: Extend e2e tests that verify CometBFT's behaviour with respect to ABCI 1.0 to ABCI 2.0
## Problem definition

We want to extend the e2e tests that verify CometBFT's behaviour with respect to ABCI 1.0 (#353) to ABCI 2.0. Namely, we want to cover vote extensions which are missing in the current version. 

## DoD

- Update current ADR (https://github.com/cometbft/cometbft/blob/main/docs/architecture/adr-108-e2e-abci%2B%2B.md).
- Extend the implementation.

nenadmilosevic95:
This PR partially closes #1404. Namely, it updates the grammar and other parts of the code to be able to check whether CometBFT respects ABCI 2.0. The only thing missing is to update ADR when this implementation is accepted. 

---

#### PR checklist

- [ ] Tests written/updated


cason:
@nenadmilosevic95, can you run `make mockery`  and `make lint`. The latter is failing.

If you want to automatically fix (most) of the lint problems, you can run `golangci-lint run --fix`.

test/e2e/pkg/grammar/abci_grammar.md
```diff
@@ -18,8 +18,11 @@ ConsensusHeight : ConsensusRounds FinalizeBlock Commit | FinalizeBlock Commit ;
 ConsensusRounds : ConsensusRound | ConsensusRound ConsensusRounds ;
 ConsensusRound : Proposer | NonProposer ; 
 
-Proposer : PrepareProposal | PrepareProposal ProcessProposal ; 
-NonProposer: ProcessProposal ;
+Proposer : GotVotes | ProposerSimple | Extend | GotVotes ProposerSimple | GotVotes Extend | ProposerSimple Extend | GotVotes ProposerSimple Extend ;
```
sergio-mena, line 21: Is this line considering the case where:
* no votes are received
* no prepare proposal is received (as a proposer)
* no extend vote is received

test/e2e/pkg/grammar/checker_test.go
```diff
@@ -17,88 +17,159 @@ var (
...
+}
 
-func TestVerify(t *testing.T) {
-	tests := []struct {
-		name         string
-		abciCalls    []*abci.Request
-		isCleanStart bool
-		result       bool
-	}{
+func TestVerifyCleanStart(t *testing.T) {
```
sergio-mena, line 112: Why do we need to split it back into "clean start" and "recovery" ?

test/e2e/pkg/grammar/abci_grammar.md
nenadmilosevic95: Do you mean all three together?

test/e2e/pkg/grammar/checker_test.go
nenadmilosevic95: Because a valid "clean start" execution is not necessarily a valid "recovery" execution and vice versa.

test/e2e/pkg/grammar/abci_grammar.md
sergio-mena: Yes, the original grammar is:
```abnf
proposer = *got-vote [prepare-proposal [process-proposal]] [extend]
```
so all three parts are optional.

Also, I just realized the same applies to `NonProposer`

test/e2e/pkg/grammar/checker_test.go
sergio-mena: I see. I trust your judgement here.
My question was motivated because we used to have the distinction between clean start and recovery in an earlier version (we even had two grammars), and then we managed to unify it. Hence I was curious why we're splitting this function whereas before this PR we only had `TestVerify`.

test/e2e/pkg/grammar/abci_grammar.md
nenadmilosevic95: This is actually covered with this rule: 

`ConsensusHeight : ConsensusRounds FinalizeBlock Commit | FinalizeBlock Commit ;`
`ConsensusRounds : ConsensusRound | ConsensusRound ConsensusRounds ;`

So, in this case, `ConsensusHeight` will be `FinalizeBlock Commit`.

test/e2e/pkg/grammar/checker_test.go
nenadmilosevic95: The question is good. The thing is that since we now have one grammar, the parser will return that the execution is valid if it represents a valid clean start or recovery execution. In addition, it will return a parse tree. Consequently, to check that an execution is of a valid type also, I pass to the Verify method the execution type and then check if the parse tree matches the execution type. But this check is done inside the verify method; the parser alone cannot do this check since, for the grammar and the parser, the execution is valid if it satisfies any of the types.

sergio-mena:
@nenadmilosevic95 can you turn this into a (non-draft) PR, so I can approve?

sergio-mena (approved):
🙏

PR #2205, opened 2024-01-31
revert(pbts): revert removal of BFT Time related functions

### LINKED ISSUE #2203: Restore `state.MedianTime` and affected call sites (block creation and validation)

### LINKED ISSUE #2204: PBTS: Restore and adapt `voteTime()` consensus method accordingly

lasarojc:
Closes #2203
Closes #2204

PBTS will be enabled through a flag and therefore the code to use BFT Time should continue to exist in v1

Pls note that this is branched of `glnro/con-params` and not directly from `feature/pbts` for now. Will change the base once `glnro/con-params` is merged.

---

#### PR checklist

- [ ] Tests written/updated


lasarojc:
This PR fails some tests, but these failures seem to be inherited from parent branches, hence making it ready for review.

internal/consensus/state_test.go
```diff
@@ -1609,6 +1609,104 @@ func TestStateLock_POLSafety2(t *testing.T) {
 	*/
 	ensureNewProposal(proposalCh, height, round)
 
+	ensurePrevote(voteCh, height, round)
+	validatePrevote(t, cs1, round, vss[0], propBlockID1.Hash)
```
lasarojc, line 1613: Reverted the change here, for when PBTS is disabled.

internal/consensus/state_test.go
```diff
@@ -1609,6 +1609,104 @@ func TestStateLock_POLSafety2(t *testing.T) {
...
+	// Add the pol votes
+	addVotes(cs1, prevotes...)
+
+	ensureNewRound(newRoundCh, height, round)
+	t.Log("### ONTO Round 2")
+	/*Round2
+	// now we see the polka from round 1, but we shouldn't unlock
+	*/
+	ensureNewProposal(proposalCh, height, round)
+
```
lasarojc, line 1709: For when is enabled, the test compares with nil, two lines below.

types/params.go
```diff
@@ -164,7 +164,7 @@ func DefaultSynchronyParams() SynchronyParams {
 	// TODO(@wbanfield): Determine experimental values for these defaults
 	// https://github.com/tendermint/tendermint/issues/7202
 	return SynchronyParams{
-		Precision:    500 * time.Millisecond,
+		Precision:    time.Duration(10000 * 366 * 24 * 60 * 60 * 1000), // 500 * time.Millisecond, // Default is to use BFT Time. // TODO: use something else to determine if PBTS is enabled.
```
lasarojc, line 167:
```suggestion
		Precision:    500 * time.Millisecond, // Default is to use BFT Time.
```

internal/blocksync/reactor.go
```diff
@@ -484,7 +484,7 @@ FOR_LOOP:
 
 			if err == nil {
 				// validate the block before we persist it
-				err = bcR.blockExec.ValidateBlock(state, first)
```
sergio-mena, line 487: Why is this change popping up here?

internal/consensus/invalid_test.go
```diff
@@ -80,12 +79,14 @@ func invalidDoPrevoteFunc(t *testing.T, cs *State, sw *p2p.Switch, pv types.Priv
...
 		blockHash := bytes.HexBytes(cmtrand.Bytes(32))
+		timestamp := cs.voteTime(cs.Height)
+
 		precommit := &types.Vote{
 			ValidatorAddress: addr,
 			ValidatorIndex:   valIndex,
 			Height:           cs.Height,
 			Round:            cs.Round,
-			Timestamp:        cmttime.Now(),
+			Timestamp:        timestamp,
```
sergio-mena, line 89: Does this change really matter?

internal/consensus/pbts_test.go
```diff
@@ -521,6 +532,9 @@ func TestPBTSTooFarInTheFutureProposal(t *testing.T) {
 		height2ProposedBlockOffset:        100 * time.Millisecond,
 		height2ProposalTimeDeliveryOffset: 10 * time.Millisecond,
 		height4ProposedBlockOffset:        150 * time.Millisecond,
+		pbtsParams: types.PBTSParams{
+			PBTSEnableHeight: 1,
+		},
```
sergio-mena, line 537: Do we need another unit test here (or maybe adapt an existing one) to test activation? (i.e., PBTS enable height > 1) Or non-activation? (PBTS enable height stays at 0)

internal/consensus/state.go
```diff
@@ -2700,3 +2727,8 @@ func proposerWaitTime(lt cmttime.Source, bt time.Time) time.Duration {
 	}
 	return 0
 }
+
+// isPBTSEnabled returns true if PBFT is enabled at the current height.
```
sergio-mena, line 2731:
```suggestion
// isPBTSEnabled returns true if PBTS is enabled at the current height.
```

internal/consensus/state_test.go
sergio-mena: Sorry, I think I don't understand the change. Is the unite test testing with PBTS enabled or disabled?

internal/consensus/state_test.go
sergio-mena: Let me see if I understood this change: you duplicated `TestStateLock_POLSafety2` into two test cases: one with PBTS and one without.
If that's the case, could it be possible to refactor it into two test cases (with and without) but reusing most of the logic of the original test case? It would be much easier to maintain IMO

internal/state/execution.go
```diff
@@ -195,7 +195,19 @@ func (blockExec *BlockExecutor) ProcessProposal(
 // Validation does not mutate state, but does require historical information from the stateDB,
 // ie. to verify evidence from a validator at an old height.
 func (blockExec *BlockExecutor) ValidateBlock(state State, block *types.Block) error {
-	err := validateBlock(state, block)
+	err := validateBlock(state, block, false)
```
sergio-mena, line 198: Hmm. why is this popping here?

internal/state/validation.go
```diff
@@ -119,6 +119,15 @@ func validateBlock(state State, block *types.Block) error {
...
+			medianTime := MedianTime(block.LastCommit, state.LastValidators)
+			if !block.Time.Equal(medianTime) {
+				return fmt.Errorf("invalid block time. Expected %v, got %v",
+					medianTime,
+					block.Time,
+				)
+			}
+		}
 
 	case block.Height == state.InitialHeight:
```
sergio-mena, line 132: Not sure about this, but I think one of the PRs previously forward-ported was removing the requirement that that the timestamp of the first block has to match the genesis time. If that's that case, this condition should _only_ be checked if PBTS is not enabled

internal/state/validation.go
```diff
@@ -119,6 +119,15 @@ func validateBlock(state State, block *types.Block) error {
 				state.LastBlockTime,
 			)
 		}
+		if !state.ConsensusParams.PBTS.PBTSEnabled(block.Height) && !isBlockSync {
```
sergio-mena, line 122: Why the `!isBlockSync`?

cason (commented):
Left a bunch of comments.

Regarding the PR itself, please add to the description its dependency (as it does not target the feature branch).

I would actually prefer if it targeted the feature branch.

internal/consensus/invalid_test.go
sergio-mena: Hmm, I'm guessing @lasarojc brought it because it was a change that the forward-ported code had been doing. If that's the case, I'm OK leaving this change, as it'll disappear when merging back to `main`

internal/state/validation.go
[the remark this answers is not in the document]
sergio-mena: I see. I think I remember. Was the reason for skipping this in blocksync that, since the blocks are light-client-validated, we assume the data is correct? (as we make the same assumption with the `app_hash`?

If that's the case, @cason, I think the answer to your question is that, by skipping the unneeded check, we are being faster.

internal/state/validation.go
```diff
@@ -12,7 +12,7 @@ import (
 //-----------------------------------------------------
 // Validate block
 
-func validateBlock(state State, block *types.Block) error {
+func validateBlock(state State, block *types.Block, isBlockSync bool) error {
```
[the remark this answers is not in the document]
sergio-mena: It's a private method, what problem do you forsee @cason ?

internal/consensus/state.go
```diff
@@ -2510,6 +2511,32 @@ func (cs *State) signVote(
 	return vote, err
 }
 
+func (cs *State) voteTime(height int64) time.Time {
+	now := cmttime.Now()
+	if cs.isPBTSEnabled(height) {
+		return now
```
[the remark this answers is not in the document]
sergio-mena: if you return `now`, you save a system call

api/cometbft/types/v1/params.pb.go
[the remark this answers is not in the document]
lasarojc: Yes. It still is in the parent, so once fixed there (it will disappear here)

internal/consensus/state.go
```diff
@@ -2700,3 +2727,8 @@ func proposerWaitTime(lt cmttime.Source, bt time.Time) time.Duration {
 	}
 	return 0
 }
+
+// isPBTSEnabled returns true if PBFT is enabled at the current height.
+func (cs *State) isPBTSEnabled(height int64) bool {
```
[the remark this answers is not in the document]
lasarojc: We will merge the other one first and then change the base for this one back to `feature/pbts`

internal/state/validation.go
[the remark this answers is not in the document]
lasarojc: Exactly. The whole test had been removed as not needed when using PBTS, but since we need to support BFT time, it was brought back.
However during the discussion we concluded that even with BFT Time the test is not needed during block sync, since the timestamp has been verified when it was decided and we check the block hash (which includes the time) that this is the one decided.

lasarojc: see previous answer.

internal/consensus/common_test.go
```diff
@@ -496,6 +496,12 @@ func randStateWithApp(nValidators int, app abci.Application) (*State, []*validat
 	return randStateWithAppImpl(nValidators, app, c)
 }
 
+func randStateWithAppVePbts(nValidators int) (*State, []*validatorStub) {
```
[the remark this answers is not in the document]
lasarojc: Removed VE. On the first iteration I was using randStateWithAppHeight.

Renamed, but think the `Enabled` part would be redundant.

internal/consensus/invalid_test.go
lasarojc: I am reverting here to keep the feature branch simpler (As @sergio-mena mentioned, this was forward-ported and I am reverting with as little further changes as possible).

internal/state/execution.go
sergio-mena: Please ignore this comment: I realize it's a change added by this PR, as we're not checking BFT Time from blocksync

internal/consensus/pbts_test.go
```diff
@@ -382,13 +381,16 @@ func TestPBTSProposerWaitsForGenesisTime(t *testing.T) {
 		height2ProposalTimeDeliveryOffset: 10 * time.Millisecond,
 		height2ProposedBlockOffset:        10 * time.Millisecond,
 		height4ProposedBlockOffset:        30 * time.Millisecond,
+		pbtsParams: types.PBTSParams{
```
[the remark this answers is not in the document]
lasarojc: Inlined `pbtsParams to make it more pleasing.

internal/consensus/pbts_test.go
lasarojc: We need a test, but that should be done in the base branch, which adds the parameter.

internal/consensus/pbts_test.go
```diff
@@ -521,6 +524,7 @@ func TestPBTSTooFarInTheFutureProposal(t *testing.T) {
 		height2ProposedBlockOffset:        100 * time.Millisecond,
 		height2ProposalTimeDeliveryOffset: 10 * time.Millisecond,
 		height4ProposedBlockOffset:        150 * time.Millisecond,
+		pbtsParams:                        types.PBTSParams{PBTSEnableHeight: 1},
```
lasarojc, line 527: undo. Moved to base branch.

internal/state/validation.go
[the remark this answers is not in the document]
sergio-mena: OK, we changed our mind yesterday and decided to go for _not_ modifying the checks we do when in BFT Time w.r.t. the checks performed in `v0.38.x`

internal/consensus/pbts_test.go
```diff
@@ -482,6 +483,7 @@ func TestPBTSTimelyProposal(t *testing.T) {
 		genesisTime:                       initialTime,
 		height2ProposedBlockOffset:        15 * time.Millisecond,
 		height2ProposalTimeDeliveryOffset: 30 * time.Millisecond,
+		pbtsParams:                        types.PBTSParams{PBTSEnableHeight: 1},
```
lasarojc, line 486:
```suggestion
```

internal/consensus/pbts_test.go
```diff
@@ -503,6 +505,7 @@ func TestPBTSTooFarInThePastProposal(t *testing.T) {
 		timeoutPropose:                    50 * time.Millisecond,
 		height2ProposedBlockOffset:        15 * time.Millisecond,
 		height2ProposalTimeDeliveryOffset: 27 * time.Millisecond,
+		pbtsParams:                        types.PBTSParams{PBTSEnableHeight: 1},
```
lasarojc, line 508:
```suggestion
```

internal/consensus/pbts_test.go
```diff
@@ -526,6 +529,7 @@ func TestPBTSTooFarInTheFutureProposal(t *testing.T) {
 		height2ProposedBlockOffset:        100 * time.Millisecond,
 		height2ProposalTimeDeliveryOffset: 10 * time.Millisecond,
 		height4ProposedBlockOffset:        150 * time.Millisecond,
+		pbtsParams:                        types.PBTSParams{PBTSEnableHeight: 1},
```
lasarojc, line 532:
```suggestion
```

internal/state/state.go
```diff
@@ -240,13 +240,21 @@ func (state State) MakeBlock(
 	// Build base block with block data.
 	block := types.MakeBlock(height, txs, lastCommit, evidence)
 
-	// TODO(@glnro): Update logic once merged with #2205
-	// Check if PBTS Enabled: state.ConsensusParams.PBTS.PBTSEnabled(height)
+	// Set time.
+	var timestamp time.Time
+	switch {
+	case state.ConsensusParams.PBTS.PBTSEnabled(height):
```
lasarojc, line 246:
```suggestion
	case state.ConsensusParams.Feature.PbtsEnabled(height):
```

internal/state/validation.go
```diff
@@ -119,6 +119,15 @@ func validateBlock(state State, block *types.Block) error {
 				state.LastBlockTime,
 			)
 		}
+		if !state.ConsensusParams.PBTS.PBTSEnabled(block.Height) {
```
lasarojc, line 122:
```suggestion
		if !state.ConsensusParams.Feature.PbtsEnabled(block.Height) {
```

internal/consensus/state.go
```diff
@@ -2686,3 +2712,8 @@ func proposerWaitTime(lt cmttime.Source, bt time.Time) time.Duration {
 	}
 	return 0
 }
+
+// isPBTSEnabled returns true if PBTS is enabled at the current height.
+func (cs *State) isPBTSEnabled(height int64) bool {
+	return cs.state.ConsensusParams.PBTS.PBTSEnabled(height)
```
lasarojc, line 2718:
```suggestion
	return cs.state.ConsensusParams.Feature.PbtsEnabled(height)
```

internal/consensus/state_test.go
lasarojc: Refactored and merged.

internal/state/state.go
```diff
@@ -240,13 +240,21 @@ func (state State) MakeBlock(
 	// Build base block with block data.
 	block := types.MakeBlock(height, txs, lastCommit, evidence)
 
-	// TODO(@glnro): Update logic once merged with #2205
-	// Check if PBTS Enabled: state.ConsensusParams.PBTS.PBTSEnabled(height)
+	// Set time.
+	var timestamp time.Time
+	switch {
+	case state.ConsensusParams.Feature.PbtsEnabled(height):
+		timestamp = time.Now()
```
cason, line 247: Why `time` and not `cmttime` here?

internal/consensus/common_test.go
```diff
@@ -496,6 +496,13 @@ func randStateWithApp(nValidators int, app abci.Application) (*State, []*validat
 	return randStateWithAppImpl(nValidators, app, c)
 }
 
+func randStateWithAppPbts(nValidators int) (*State, []*validatorStub) {
```
cason, line 499: I think that this must be the default. See https://github.com/cometbft/cometbft/pull/2329.

Namely, if we want to run **without** PBTS we need to set enableHeight to 0 for that test.

internal/consensus/pbts_test.go
cason, line 1: Nothing has changed here, ehehe.

internal/consensus/state.go
```diff
@@ -1382,7 +1382,7 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
 		return
 	}
 
-	if cs.Proposal.POLRound == -1 && cs.LockedRound == -1 && !cs.proposalIsTimely() {
+	if cs.isPBTSEnabled(height) && cs.Proposal.POLRound == -1 && cs.LockedRound == -1 && !cs.proposalIsTimely() {
```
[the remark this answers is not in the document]
cason: Moreover, I would be more didactic with the code reader here.

If POLRound == 1 and PBTS is enabled. Then inside the if check timely. If it fails, log this and prevote nil. But this is only a suggestion.

internal/consensus/state.go
```diff
@@ -2510,6 +2511,32 @@ func (cs *State) signVote(
...
+	if cs.isPBTSEnabled(height) {
+		return now
+	}
+
+	minVoteTime := now
+
+	// Minimum time increment between blocks
+	const timeIota = time.Millisecond
+	// TODO: We should remove next line in case we don't vote for v in case cs.ProposalBlock == nil,
+	// even if cs.LockedBlock != nil. See https://github.com/cometbft/cometbft/tree/main/spec/.
```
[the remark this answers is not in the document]
cason: Still valid the previous comment.

internal/consensus/state.go
```diff
@@ -2686,3 +2712,8 @@ func proposerWaitTime(lt cmttime.Source, bt time.Time) time.Duration {
 	}
 	return 0
 }
+
+// isPBTSEnabled returns true if PBTS is enabled at the current height.
```
cason, line 2716: We have the lock when using this, haven't we? If so, probably we should write a "contract" comment.

internal/state/state.go
```diff
@@ -240,13 +240,21 @@ func (state State) MakeBlock(
...
+	case height == state.InitialHeight:
```
cason, line 248: Shouldn't this condition come before?

internal/state/state.go
```diff
@@ -255,6 +263,29 @@ func (state State) MakeBlock(
 	return block
 }
 
+// MedianTime computes a median time for a given Commit and the
```
cason, line 266:
```suggestion
// MedianTime computes the median time for a Commit based on the associated validator set.
// The median time is the weighted median of the Timestamp fields of the commit votes,
// with heights defined by the validator's voting powers.
// The BFT Time algorithm ensures that the computed median time is always picked among
// the timestamps produced by honest processes, i.e., faulty processes cannot arbitrarily
// increase or decrease the median time.
// See: https://github.com/cometbft/cometbft/blob/main/spec/consensus/bft-time.md
```

internal/state/validation.go
```diff
@@ -119,6 +119,15 @@ func validateBlock(state State, block *types.Block) error {
 				state.LastBlockTime,
 			)
 		}
+		if !state.ConsensusParams.Feature.PbtsEnabled(block.Height) {
+			medianTime := MedianTime(block.LastCommit, state.LastValidators)
+			if !block.Time.Equal(medianTime) {
+				return fmt.Errorf("invalid block time. Expected %v, got %v",
+					medianTime,
```
cason, line 126:
```suggestion
					medianTime.Format(time.RFC3339Nano),
```

internal/state/validation.go
```diff
@@ -119,6 +119,15 @@ func validateBlock(state State, block *types.Block) error {
...
+					block.Time,
```
cason, line 127:
```suggestion
					block.Time.Format(time.RFC3339Nano),
```

internal/consensus/state.go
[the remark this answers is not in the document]
cason: Although in `main` the comment is the same.

internal/consensus/state_test.go
```diff
@@ -1582,91 +1582,114 @@ func TestStateLock_POLSafety2(t *testing.T) {
...
+	csT2, vssT2 := randStateWithAppPbts(4)
+
+	tcs := []struct {
+		name  string
+		state *State
+		vss   []*validatorStub
+		exp   error
+	}{
+		{name: "Without PBTS", state: csT1, vss: vssT1, exp: nil},
+		{name: "With PBTS", state: csT2, vss: vssT2, exp: nil},
```
cason, line 1595: I don't understand whether the result `exp`, which is not even used , should change when running BFT Time or PBTS.

internal/consensus/state_test.go
```diff
@@ -1582,91 +1582,114 @@ func TestStateLock_POLSafety2(t *testing.T) {
...
+	csT1, vssT1 := randState(4)
+	csT2, vssT2 := randStateWithAppPbts(4)
+
+	tcs := []struct {
+		name  string
+		state *State
+		vss   []*validatorStub
+		exp   error
+	}{
+		{name: "Without PBTS", state: csT1, vss: vssT1, exp: nil},
```
cason, line 1594:
```suggestion
		{name: "With BFT Time", state: csT1, vss: vssT1, exp: nil},
```

internal/state/state.go
cason: Notice that they are different:

```
=== RUN   TestTime
    time_test.go:11: cmttime.Now() 2024-02-15 16:47:32.96113 +0000 UTC
    time_test.go:12: time.Now() 2024-02-15 17:47:32.961536 +0100 CET m=+0.012614460
```

cason: The question here is how that hell this doesn't break any test.

cason (commented):
Good, my only relevant question is the use of `time.Now()` in `state/state.go`, as it produces a value pretty different from `cmttime.Now()`. And how this don't break everything.

internal/consensus/state.go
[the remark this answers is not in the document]
lasarojc: Yes, remember that this is a reversion of removal of this code block.
We should fix this against main, if at all.

internal/consensus/state_test.go
lasarojc: exp is nil in both cases. I will remove it.

internal/consensus/common_test.go
lasarojc: Good point. Done.

internal/state/state.go
lasarojc: Internally they are the same; `cmttime` is just `time` at UTC, but the localtime is the same. In any case, changing to `cmttime`.

internal/state/state.go
lasarojc: If using PBTS from genesis, then the genesis time will not be timely, as normally the genesis will be created long before the chain is started the genesis timestamp will be well in the past.

internal/consensus/state.go
lasarojc: This is true for `IsPBTSEnabled` (except under testing).
Not true for `state.ConsensusParams.Feature.PbtsEnabled(height)`, which is also called in other places (under testing and when performing block sync). This is actually causing a race caught in #2335

So I propose to treat this in the context of #2335 as well

internal/consensus/state.go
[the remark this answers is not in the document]
lasarojc: Do you know the issue?

internal/state/state.go
cason: It is also stripped.

internal/consensus/state.go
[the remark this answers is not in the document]
lasarojc: > I know it was not introduced here, but why we have the third condition (`cs.LockedRound == -1`)?

If `cs.LockedRound == -1` then there is a chance that the block will be accepted in line 1471, since its timeliness has been tested and passed before, speeding up a decision.

internal/consensus/state.go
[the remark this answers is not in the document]
cason: I don't know the issue. I followed the original commit, and there is no comment on that. The spec refers to this, indeed.

I think we can remove this comment completely.

internal/consensus/state.go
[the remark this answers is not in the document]
cason: If `cs.Proposal.POLRound == -1` we don't even get there.

cason (approved):
After looong discussions, this is good to go.

Issue #2110, opened 2024-01-24
PBTS: add/update introduced parameters to data structures spec


cason:
_Originally posted by @cason in https://github.com/cometbft/cometbft/pull/2018#discussion_r1458852653_

- [x] Update the description of `Time` field of `Proposal` and block `Header`: https://github.com/cometbft/cometbft/blob/main/spec/core/data_structures.md#proposal
- [ ] Update `ConsensusParams` table: https://github.com/cometbft/cometbft/blob/main/spec/core/data_structures.md#consensusparams
- [x] Uncomment/update documentation on https://github.com/cometbft/cometbft/blob/main/spec/abci/abci%2B%2B_app_requirements.md?plain=1#L682

cason:
Closed after merging #2206

PR #2206, opened 2024-01-31
spec(consensus/pbts): update description for introduced parameters.

cason:
Addresses #2110.

Includes some general fixes in `spec/core/data_structures.md`.

---

#### PR checklist

- [ ] ~Tests written/updated~
- [ ] ~Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)~


spec/core/data_structures.md
```diff
@@ -129,7 +129,7 @@ the data in the current block, the previous block, and the results returned by t
...
+| Time              | [Time](#time)             | The timestamp can be computed using [PBTS][pbts] or [BFT Time][bfttime] algorithms. In case of PBTS, it is the time at which the proposer has produced the block. In case of BFT Time, it is equal to the weighted median of timestamps validators present in the last commit.                                                                                                  | Time must be larger than the Time of the previous block header. The timestamp of the first block should not be smaller than the genesis time. When BFT Time is adopted, it should match the genesis time (since there's no votes to compute the median). |
```
sergio-mena, line 132:
```suggestion
| Time              | [Time](#time)             | The timestamp can be computed using [PBTS][pbts] or [BFT Time][bfttime] algorithms. In case of PBTS, it is the time at which the proposer has produced the block. In case of BFT Time, it is equal to the weighted median of timestamps validators present in the last commit.                                                                                                  | Time must be larger than the Time of the previous block header. The timestamp of the first block should not be smaller than the genesis time. When BFT Time is adopted, it should match the genesis time (since there's no votes to compute the median with). |
```

spec/core/data_structures.md
```diff
@@ -199,8 +199,8 @@ Commit is a simple wrapper for a list of signatures, with one for each validator
 
 | Name       | Type                             | Description                                                          | Validation                                                                                               |
 |------------|----------------------------------|----------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------|
-| Height     | int64                            | Height at which this commit was created.                             | Must be > 0                                                                                              |
-| Round      | int32                            | Round that the commit corresponds to.                                | Must be > 0                                                                                              |
+| Height     | int64                            | Height at which this commit was created.                             | Must be >= 0                                                                                             |
```
sergio-mena, line 202: Maybe it's worth adding a note after the table explaining when this field is set to 0 (IIRC, it's on the first block, to signal the absence of signatures)

spec/core/data_structures.md
```diff
@@ -270,8 +270,8 @@ The vote includes information about the validator signing it. When stored in the
 | Name               | Type                            | Description                                                                              | Validation                               |
 |--------------------|---------------------------------|------------------------------------------------------------------------------------------|------------------------------------------|
 | Type               | [SignedMsgType](#signedmsgtype) | The type of message the vote refers to                                                   | Must be `PrevoteType` or `PrecommitType` |
-| Height             | int64                           | Height for which this vote was created for                                               | Must be > 0                              |
-| Round              | int32                           | Round that the commit corresponds to.                                                    | Must be > 0                              |
+| Height             | int64                           | Height for which this vote was created for                                               | Must be >= 0                             |
```
sergio-mena, line 273: When is this field set to 0 ?
Looks like it can't be zero, because of [this](https://github.com/cometbft/cometbft/blob/main/types/vote.go#L279)

spec/core/data_structures.md
```diff
@@ -289,8 +289,8 @@ and uses a different ordering of the fields.
 | Name      | Type                            | Description                             | Validation                               |
 |-----------|---------------------------------|-----------------------------------------|------------------------------------------|
 | Type      | [SignedMsgType](#signedmsgtype) | The type of message the vote refers to  | Must be `PrevoteType` or `PrecommitType` |
-| Height    | int64                           | Height in which the vote was provided.  | Must be > 0                              |
-| Round     | int64                           | Round in which the vote was provided.   | Must be > 0                              |
+| Height    | int64                           | Height in which the vote was provided.  | Must be >= 0                             |
```
sergio-mena, line 292: Please double check this change

spec/core/data_structures.md
cason, line 273:
```suggestion
| Height             | int64                           | Height for which this vote was created for                                               | Must be > 0                             |
```

spec/core/data_structures.md
cason, line 292:
```suggestion
| Height    | int64                           | Height in which the vote was provided.  | Must be > 0                             |
```

cason: There is not validation for this type. But it can only be produced from votes, for which `Height > 0` is required. Fixing.

spec/core/data_structures.md
```diff
@@ -323,25 +323,25 @@ This is the structure to marshall in order to obtain the bytes to sign or verify
...
-| Type      | [SignedMsgType](#signedmsgtype) | Represents a Proposal [SignedMsgType](#signedmsgtype)                                 | Must be `ProposalType`  [signedMsgType](#signedmsgtype) |
-| Height    | uint64                           | Height for which this vote was created for                                            | Must be > 0                                             |
-| Round     | int32                           | Round that the commit corresponds to.                                                 | Must be > 0                                             |
-| POLRound  | int64                           | Proof of lock                                                                         | Must be > 0                                             |
+| Type      | [SignedMsgType](#signedmsgtype) | Represents a Proposal [SignedMsgType](#signedmsgtype).                                | Must be `ProposalType`                                  |
+| Height    | uint64                          | Height for which this vote was created for                                            | Must be >= 0                                            |
```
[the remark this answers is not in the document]
cason: You are probably right, but the `ValidateBasic()` only rejects when it is negative. Plus, the test unit only considers this case for rejection...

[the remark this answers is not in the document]
cason: Namely, if I change this here, we would have to change also the implementation and tests.

spec/core/data_structures.md
cason: Right.

spec/core/data_structures.md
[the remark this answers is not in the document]
sergio-mena: Yeah, when I said "but it is a bug"? I was referring to the current status of `ValidateBasic`.
I suggest you leave ">=0" here for the moment, and file an issue to look into `ValidateBasic`, to make sure we don't have a bug there.

sergio-mena (approved):
🙏

spec/core/data_structures.md
cason: Mentioned this below, on the rules for the signatures.

spec/core/data_structures.md
```diff
@@ -129,7 +129,7 @@ the data in the current block, the previous block, and the results returned by t
...
+| Time              | [Time](#time)             | The timestamp can be computed using [PBTS][pbts] or [BFT Time][bfttime] algorithms. In case of PBTS, it is the time at which the proposer has produced the block. In case of BFT Time, it is equal to the weighted median of timestamps validators present in the last commit.                                                                                                  | Time must be larger than the Time of the previous block header. The timestamp of the first block should not be smaller than the genesis time. When BFT Time is used, it should match the genesis time (since there's no votes to compute the median with). |
```
lasarojc, line 132:
```suggestion
| Time              | [Time](#time)             | The timestamp can be computed using [PBTS][pbts] or [BFT Time][bfttime] algorithms. In case of PBTS, it is the time at which the proposer has produced the block (the value of its local clock). In case of BFT Time, it is equal to the weighted median of timestamps present in the previous commit.                                                                                                  | Time must be larger than the Time of the previous block header. The timestamp of the first block should not be smaller than the genesis time. When BFT Time is used, it should match the genesis time (since there's no votes to compute the median with). |
```

spec/core/data_structures.md
```diff
@@ -323,25 +325,25 @@ This is the structure to marshall in order to obtain the bytes to sign or verify
...
+| Height    | uint64                          | Height for which this vote was created for                                            | Must be >= 0                                            |
+| Round     | int32                           | Round that the commit corresponds to.                                                 | Must be >= 0                                            |
+| POLRound  | int64                           | Proof of lock round.                                                                  | Must be >= -1                                           |
 | BlockID   | [BlockID](#blockid)             | The blockID of the corresponding block.                                               | [BlockID](#blockid)                                     |
-| Timestamp | [Time](#time)                   | Timestamp represents the time at which a validator signed.                            | [Time](#time)                                           |
+| Timestamp | [Time](#time)                   | Timestamp represents the time at which the proposer produced the block.               | [Time](#time)                                           |
```
lasarojc, line 346:
```suggestion
| Timestamp | [Time](#time)                   | Timestamp represents the time at which the block was produced.               | [Time](#time)                                           |
```

lasarojc (approved):
Small tweaks proposed.

Issue #2143, opened 2024-01-26
blocksync: Double conversion from proto not needed


jmalicevic:
## Code Inefficiency 

I noted that in the blocksync reactor there double conversion from/to proto which is not needed. 
In [`respondToPeer`](https://github.com/cometbft/cometbft/blob/28ad4d2230134045e2f5ce9fdab6673e276f2579/internal/blocksync/reactor.go#L203)  we call `LoadBlock` which loads relevant block pieces (parts, meta etc.) and stores them in proto. The function then internally calls `BlockFromProto` to return the block built from block parts as a data structure.  `respondToPeer` then takes this block (as a data structure) and converts it into proto (calling `BlockToProto` ) before sending it to the peer. And we don’t even need the block (data structure) in any other format than proto in `respondToPeer`.

### Notes

There might be other places in the code where we do this. What might be worth doing is following the call-chain of all the DB functions that load blocks/meta etc. to confirm that the caller does not re-convert the things. Especially if we do not need the data structure (as in the description above). If we do need it, we can have a different Load function that returns both proto and the data structure so we don't do the work twice. 

This should probably not be backported a lot as it is not impacting safety(especially if we change function signatures).

felipetojal:
Hey, can I pick this up?

PR #2207, opened 2024-01-31
fix: Double conversion from proto not needed

faddat:
WIP PR that I'm making for visibility and so it's easy for me to access later.  Please do not review yet.

closes: #2143


andynog:
Thanks @faddat, since this is a draft PR we tend to not review until it's ready.

PR #2152, opened 2024-01-27
chore: simplify poolRoutine

faddat:
By various metrics, poolRoutine is the most complex function in CometBFT.

* cyclomatic complexity
* cognitive complexity

This PR is from #2130, which I'll break down into smaller PRs similar to the flaky test PR.  I was able to identify this function as excessively complex by configuring linters in the way described in #2130.

### review notes

I ended up removing some comments from the code, but it's possible that they should have been moved to one of the new helper functions.  Please let me know if the comments should be restored.

---

#### PR checklist

- [x] Tests written/updated
- [x] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments
- [x] Title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) spec

PR #2209, opened 2024-01-31
refactor: simplify state pruning

faddat:
Like #2152, this PR has been checked out of #2130.  

This PR breaks a pruning down into a function with helpers, instead of a single large function.

---

#### PR checklist

- [ ] Tests written/updated


internal/state/store.go
```diff
@@ -329,90 +315,11 @@ func (store dbStore) PruneStates(from int64, to int64, evidenceThresholdHeight i
...
-		// avoid batches growing too large by flushing to database regularly
-		if pruned%1000 == 0 && pruned > 0 {
-			err := batch.Write()
-			if err != nil {
-				return err
-			}
-			batch.Close()
-			batch = store.db.NewBatch()
-			defer batch.Close()
-		}
```
mzabaluev, line 415: This goes beyond code simplification... Why was batch size management removed here?

faddat: thanks for your review.  That was in error.

faddat:
@mzabaluev fyi, this has been updated to fix the missing batching.  Thanks!

faddat:
this now needs rework, and I'm hesitant to do the merge because of the risk of missing something.

PR #2215, opened 2024-01-31
fix: if-return and early-return

faddat:
This is cherry-picked from:

* #1907

There were a number of opportunities to simplify code in comet by using revive's if-return and early-return linters.  This PR does not enable the linters as default, but instead just applies the fixes suggested by them.

There are two cases where I've marked conditionals with //nolint -- they're both a little bit odd and it may make sense to look deeper into weather it is possible to benefit from early returns. 

As we've seen in the break statements in #1907 using a linter can help to enforce good practices across the entire repository and uncover issues.

## inspired by

* #2156 

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/.changelog/unreleased/fix/2215.md
+++ b/.changelog/unreleased/fix/2215.md
@@ -1,1 +1,0 @@
-fix: use more idiomatic early return pattern, and if-return linter.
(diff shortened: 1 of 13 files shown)


melekes (approved):
👍

PR #2218, opened 2024-02-01
fix(consensus): prevote nil upon timeout when Proposal is missing

cason:
This bug was possibly introduced by https://github.com/cometbft/cometbft/pull/1203. This PR refactored the consensus operation, propose -> prevote transition (`defaultDoPrevote` method), to match the pseudo-code.

The refactoring assumed that when `cs.ProposalBlock` is unset (nil), so should be `cs.Proposal` (meaning that the node didn't receive the round proposal). In this case, the node must prevote nil.

The added test unit describes a scenario where `cs.Proposal` is nil (not received), but the block ` cs.ProposalBlock` is received because the node sees a Polka for a block, then receives the full associated block. The node must prevote nil in this scenario, in line with #1203. But the lack of the test on `cs.Proposal` leads to a bug.

This bug does not affect previous releases except for `v1.0`.

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

sergio-mena (approved):
Excellent! Thanks for this!

Issue #2119, opened 2024-01-24
PBTS: do we need the `Proposal` when the proposed block received a Polka?


cason:
_Originally posted by @cason in https://github.com/cometbft/cometbft/pull/2018#discussion_r1458778006_:

      A note to remember: we are forcing a `Proposal` to be received to run the prevote logic. While this is necessary when POLRound == -1, it might not be otherwise.

The PBTS-enabled consensus code adds the requirement of receiving and accepting a `Proposal` to precommit a block:

https://github.com/cometbft/cometbft/blob/1733219c2fec4be370b672081f8daf84b768a0e9/internal/consensus/state.go#L1619-L1621

This requirement is not present in the original code, in the `main` branch,  where the absence of a valid proposal (`cs.Proposal`) does not cause the node to precommit nil. The reason for which is that a node may realize that a proposal block has become valid, because it received a Polka, in the case the `blockID` in the following code:

https://github.com/cometbft/cometbft/blob/1733219c2fec4be370b672081f8daf84b768a0e9/internal/consensus/state.go#L2374-L2376

What happens here is that the node starts "looking" for this block,  by updating its `cs.ProposalBlockParts` field. There are two possibilities to consider here:

1. The node received a `Proposal` for a block that doesn't match the `blockID` that became valid. This happens when a Byzantine proposer produces multiple proposals in a round. This is not the case considered int this issue, since the `cs.Proposal` should be set, probably for a different block, but the code doesn't care.
2. The node has not received any valid `Proposal`, so it is not building any block. This is the scenario addressed here.

The general point here is that if we received a Polka for a block, it does not matter the (first valid) `Proposal` we received, not even if we have not received any `Proposal`. This block has been deemed valid by `2/3+` validators, so there is no reason for not precommiting it. Note that we use the same rationale in PBTS when we receive a `Proposal` backed by a Polka in a `POLRound`: we don't check the block time again, as it was accepted by a quorum of validators.

With this in mind, this additional check added to the `precommit` round step does not make sense:

https://github.com/cometbft/cometbft/blob/1733219c2fec4be370b672081f8daf84b768a0e9/internal/consensus/state.go#L1625-L1628

While it is probably the only reason for requiring `cs.Proposal != nil` to get to this point. My proposal is then:

- [x] Remove the `cs.Proposal != nil` condition to precommit a block (`cs.enterPrecommit`)
[...]

PR #2221, opened 2024-02-01
fix(consensus): do not precommit nil if proposal block is received

cason:
Addresses #2119.

ADR 112, describing the PBTS implementation, [states](https://github.com/cometbft/cometbft/pull/2223/files#diff-81d55f6e52eabd6cce264a76fba015de46092b9ccda241de10cd265890ab8ad9R302):

    The precommit step will not require much modification.
    Its proposal validation rules will change in the same ways that validation will change in the prevote step with the exception of the `timely` check: precommit validation will never check that the timestamp is `timely`.

The changes proposed here are in line with this description.

Moreover, this PR in fact revert the changes to the precommit step (`enterPrecommit` method) added by https://github.com/tendermint/tendermint/pull/7480 and https://github.com/tendermint/tendermint/pull/7391.

---

#### PR checklist

- [x] Tests written/updated


internal/consensus/state.go
```diff
@@ -1615,20 +1615,14 @@ func (cs *State) enterPrecommit(height int64, round int32) {
 	}
 	// At this point, +2/3 prevoted for a particular block.
 
-	// If we never received a proposal for this block, we must precommit nil
-	if cs.Proposal == nil || cs.ProposalBlock == nil {
+	// If we did not receive a proposal block to vote for, we must precommit nil
+	// FIXME: although +2/3 prevoted for the block, therefore it is valid.
```
mzabaluev, line 1619: Who's the FIXME label for? As stated in #2119, this is correct accordingly to the spec, what's there for the code monkeys to fix?

mzabaluev (approved):
Looks good, with a nitpick on a comment.

internal/consensus/state.go
cason: Good point. I think the question here is that when adapting the code to match the spec we are losing some opportunities to, in a smart way, speed up decisions when we have Byzantine proposers, for example.

In our in production (current) releases, we can prevote and precommit to a block that matches our locked block. Because this is safe (at least, I don't see how it would not be safe). This code, for the precommit step is still here, btw, see lines 1633/1627.

When we require having `cs.ProposalBlock` to precommit a block, we are not using the possibility added to the above mentioned logic/lines.

In other words, what we would need here is a way to re-use the `LockedBlock`, setting `cs.Proposal` to it, in case the block we are looking for matches the locked one. This save us, in particular, the cost of downloading it again in a round `r > cs.LockedRound`. We may also consider the `cs.ValidRound`, for the same reasons. But this should be addressed by another issue/PR.

cason: If you navigate the history of commits, you realize that I removed this check altogether. The problem of which is that `TestStateLock_MissingProposalWhenPOLForLockedBlock` test fails. I then reintroduced this check to make this test to pass.

The fun part is that the description of the test does not match the test itself. Which is odd.

lasarojc:
LGTM

internal/consensus/state.go
mzabaluev: Hmm, not sure, maybe the FIXME explanation should be expanded or placed elsewhere, to reflect on the missed opportunities and/or the oddly specified test, and the note should state what it does now.

sergio-mena (approved):
👍

Issue #2063, opened 2024-01-18
PBTS: support both PBTS and legacy BFT Time


glnro:
In order to ensure backwards compatibility, PBTS should be enabled via consensus params.

In #2018 `MedianTime` was removed per the original work in tendermint/tendermint#7382. This does not consider that PBTS can be adopted in existing chains, with blocks produced and that need to be verified using the BFT Time algorithm. 

### Specification

- Addressed by #2415 (depends on 2376 below)
  - [x] Specify the transition from BFT Time to PBTS via consensus parameter
     - Add it to `spec/consensus/time.md`
  - [x] Update documentation of `FeatureParams`, remove references to `ABCIParams`

### Documentation

- Addressed by #2376
  - [x] Update ADR with the adopted solution: #2223
  - [x] Update the [PBTS documentation](https://github.com/cometbft/cometbft/blob/feature/pbts/docs/explanation/core/proposer-based-timestamps.md) accordingly
  - [x] Add to the [PBTS documentation](https://github.com/cometbft/cometbft/blob/feature/pbts/docs/explanation/core/proposer-based-timestamps.md) the workaround for transitioning to PBTS proposed in the ADR (see this [comment](https://github.com/cometbft/cometbft/pull/2223#discussion_r1481259421)).

### Implementation

- "Rescue" code removed by cherry-picks:
  - [x] #2203
    - #2205
  - [x] #2204
    - #2205
  - [x] #2397 
  - Adapt calls to the construction `types.NewProposal` to accept a timestamp (see [comment](https://github.com/cometbft/cometbft/pull/2018#discussion_r1458000432))
      - This is not needed, as in if we didn't update them yet, compilation errors and test failing would be observed
- Introduce and implement consensus param to enable PBTS
  - [x] #2197
    - #2231
    - #2325
  - [x] #2326
  - [x] #2438

### Testing

- [x] #2324
   - https://github.com/cometbft/cometbft/pull/2329
- [x] #2393
   - https://github.com/cometbft/cometbft/pull/2404 (depends on 2339 above)
- [x] #2227

cason:
Some relevant a priori questions:

1. Should we still support BFT Time?
2. Should we allow chains to stick with BFT Time, i.e., to not adopt PBTS from v1.0?

sergio-mena:
Summary of decisions made during the meeting today (participants @glnro, @lasarojc, @cason , @sergio-mena):
[...]

PR #2223, opened 2024-02-01
docs: ADR-112 Proposer-Based Timestamps

cason:
ADR 071, from Tendermint Core repository was renamed to ADR 112 and updated to match the current implementation of PBTS ( #1731).

Rendered [ADR 112](https://github.com/cometbft/cometbft/blob/cason/pbts-adr/docs/references/architecture/adr-112-proposer-based-timestamps.md).

---

Since the file was copied, renamed, then adapted, Git lost track of the actual changes, that can be manually checked using:

    git diff cason/pbts-adr:docs/architecture/adr-112-proposer-based-timestamps.md feature/pbts:docs/architecture/tendermint-core/adr-071-proposer-based-timestamps.md

**Important:** there are missing parts in this ADR, that should be addressed before merging it. It is open for discussion, as it is expected for an ADR.

---

#### PR checklist

- [ ] Tests written/updated


docs/architecture/adr-112-proposer-based-timestamps.md
```diff
@@ -0,0 +1,367 @@
...
+If we remove the `Timestamp` field from the `Precommit` messages, we are able to aggregate signatures.
+
+## Consequences
+
+### Positive
+
+* `<2/3` of validators can no longer influence block timestamps.
+* Block timestamp will have stronger correspondence to real time.
+* Improves the reliability of light client block verification.
+* Enables BLS signature aggregation.
```
lasarojc, line 341:
```suggestion
* Will enable BLS signature aggregation.
```

docs/architecture/adr-112-proposer-based-timestamps.md
```diff
@@ -0,0 +1,367 @@
...
+* Alters CometBFT’s liveness properties.
+Liveness now requires that all correct validators have synchronized clocks within a bound.
+Liveness will now also require that validators’ clocks move forward, which was not required under `BFT Time`.
+
+### Negative
+
+* May increase the length of the propose step if there is a large skew between the previous proposer and the current proposer’s local Unix time.
+This skew will be bound by the `PRECISION` value, so it is unlikely to be too large.
+
+* Current chains with block timestamps far in the future will either need to pause consensus until after the erroneous block timestamp or must maintain synchronized but very inaccurate clocks.
```
lasarojc, line 355: is it the case of any important chain?

lasarojc (approved):

docs/architecture/adr-112-proposer-based-timestamps.md
```diff
@@ -9,40 +9,40 @@
...
+Additionally, any amount of voting power `>1/3` may control the block timestamp.
 As a result, it is quite possible that the timestamp is not particularly meaningful.
 
-These drawbacks present issues in the CometBFT protocol.
+These drawbacks present issues in CometBFT.
 Timestamps are used by light clients to verify blocks.
-Light clients rely on correspondence between their own currently known Unix time and the block timestamp to verify blocks they see;
-However, their currently known Unix time may be greatly divergent from the block timestamp as a result of the limitations of `BFTTime`.
+Light clients rely on correspondence between their own currently known Unix time and the block timestamp to verify blocks they see.
+However, their currently known Unix time may be greatly divergent from the block timestamp as a result of the limitations of `BFT Time`.
```
sergio-mena, line 36: Should we mention that we want to get rid of timestamps in votes in the long term, so that they are more amenable to aggregation?

docs/architecture/adr-112-proposer-based-timestamps.md
```diff
@@ -9,40 +9,40 @@
...
 
-The proposer-based timestamps specification suggests an alternative approach for producing block timestamps that remedies these issues.
+The [Proposer-Based Timestamps specification (PBTS)][pbts-spec] suggests an alternative approach for producing block timestamps that remedies these issues.
 Proposer-based timestamps alter the current mechanism for producing block timestamps in two main ways:
 
-1. The block proposer is amended to offer up its currently known Unix time as the timestamp for the next block instead of the `BFTTime`.
+1. The block proposer is amended to offer up its currently known Unix time as the timestamp for the next block instead of the `BFT Time`.
 1. Correct validators only approve the proposed block timestamp if it is close enough to their own currently known Unix time.
 
 The result of these changes is a more meaningful timestamp that cannot be controlled by `<= 2/3` of the validator voting power.
```
sergio-mena, line 44: In when sense is it better than BFT Time in this aspect?
My understanding:
* In BFT Time: Byzantine validators (say, close to 1/3 of voting power) can affect the time by timestamping their votes with nonsensical values. They will affect the median value, but the median value will still be that from a correct validator
* In PBTS: Byzantine proposers also have some degree of "freedom" to propose values that deviate from the real value of their clock, as long as `timely` passes in 2/3 of validators
So, in both cases, byzantine validators have a limited influence on the final timestamp value, albeit, via different mechanisms.

docs/architecture/adr-112-proposer-based-timestamps.md
```diff
@@ -125,56 +127,61 @@ type Vote struct {
...
 }
 ```
 
 ```go
 type SynchronyParams struct {
-	MessageDelay time.Duration `json:"message_delay"`
-	Precision    time.Duration `json:"precision"`
+        Precision    time.Duration `json:"precision,string"`
+        MessageDelay time.Duration `json:"message_delay,string"`
 }
```
sergio-mena, line 163: I'm assuming we're adding the enable height in a further PR?

docs/architecture/adr-112-proposer-based-timestamps.md
```diff
@@ -125,56 +127,61 @@ type Vote struct {
...
-In proposer-based timestamps, the proposer will still set a timestamp into the `Header.Timestamp`.
-The timestamp the proposer sets into the `Header` will change depending on if the block has previously received a [polka](https://github.com/cometbft/cometbft/blob/053651160f496bb44b107a434e3e6482530bb287/docs/introduction/what-is-cometbft.md#consensus-overview) or not.
+In PBTS, the proposer will still set a timestamp into the `Header.Timestamp`.
+The timestamp the proposer sets into the `Header` will change depending on if the block has previously received a Polka, i.e., `2/3+` prevotes in a previous round.
 
-#### Proposal of a block that has not previously received a polka
+#### Proposal of a block that has not previously received a Polka
 
 If a proposer is proposing a new block then it will set the Unix time currently known to the proposer into the `Header.Timestamp` field.
 The proposer will also set this same timestamp into the `Timestamp` field of the `Proposal` message that it issues.
```
sergio-mena, line 180: Do we check at any point that the timestamp set to the proposal matches the proposed block's timestamp?

docs/architecture/adr-112-proposer-based-timestamps.md
```diff
@@ -125,56 +127,61 @@ type Vote struct {
...
 
-#### Re-proposal of a block that has previously received a polka
+#### Re-proposal of a block that has previously received a Polka
 
-If a proposer is re-proposing a block that has previously received a polka on the network, then the proposer does not update the `Header.Timestamp` of that block.
+If a proposer is re-proposing a block that has previously received a Polka on the network, then the proposer does not update the `Header.Timestamp` of that block.
 Instead, the proposer simply re-proposes the exact same block.
```
sergio-mena, line 185: What does it do with the `timestamp` field in the proposal message?

docs/architecture/adr-112-proposer-based-timestamps.md
```diff
@@ -183,58 +190,61 @@ The proposer will set the re-proposed block's `Header.Timestamp` as the `Proposa
...
+The rules for validating a proposed block will be modified to implement PBTS.
 We will change the validation logic to ensure that a proposal is `timely`.
 
-Per the proposer-based timestamps spec, `timely` only needs to be checked if a block has not received a +2/3 majority of `Prevotes` in a round.
+Per the PBTS spec, `timely` only needs to be checked if a block has not received a +2/3 majority of `Prevotes` in a round.
 If a block previously received a +2/3 majority of prevotes in a previous round, then +2/3 of the voting power considered the block's timestamp near enough to their own currently known Unix time in that round.
 
 The validation logic will be updated to check `timely` for blocks that did not previously receive +2/3 prevotes in a round.
-Receiving +2/3 prevotes in a round is frequently referred to as a 'polka' and we will use this term for simplicity.
+Receiving +2/3 prevotes in a round is frequently referred to as a 'Polka' and we will use this term for simplicity.
```
sergio-mena, line 212: Can we move this sentence to the place in the text where the term Polka is first used?

docs/architecture/adr-112-proposer-based-timestamps.md
```diff
@@ -183,58 +190,61 @@ The proposer will set the re-proposed block's `Header.Timestamp` as the `Proposa
...
 
 #### Current timestamp validation logic
 
 To provide a better understanding of the changes needed to timestamp validation, we will first detail how timestamp validation works currently in CometBFT.
```
sergio-mena, line 216:
```suggestion
To provide a better understanding of the changes needed for timestamp validation, we will first detail how timestamp validation works currently in CometBFT.
```

docs/architecture/adr-112-proposer-based-timestamps.md
```diff
@@ -183,58 +190,61 @@ The proposer will set the re-proposed block's `Header.Timestamp` as the `Proposa
...
+#### Timestamp validation when a block has not received a Polka
 
-The [POLRound](https://github.com/cometbft/cometbft/blob/68ca65f5d79905abd55ea999536b1a3685f9f19d/types/proposal.go#L29) in the `Proposal` message indicates which round the block received a polka.
+The [`POLRound`](https://github.com/cometbft/cometbft/blob/1f430f51f0e390cd7c789ba9b1e9b35846e34642/types/proposal.go#L29) in the `Proposal` message indicates which round the block received a Polka.
 A negative value in the `POLRound` field indicates that the block has not previously been proposed on the network.
-Therefore the validation logic will check for timely when `POLRound < 0`.
+Therefore the validation logic will check for timely when `POLRound == -1`.
 
-When a validator receives a `Proposal` message, the validator will check that the `Proposal.Timestamp` is at most `PRECISION` greater than the current Unix time known to the validator, and at maximum `PRECISION + MSGDELAY` less than the current Unix time known to the validator.
+When a validator receives a `Proposal` message, it records it `proposalReceiveTime` as the current Unix time known to the validator.
```
sergio-mena, line 246: General comment. You call "validator" when explaining that a proposal is received. But, if I'm not wrong, this will be run by all full nodes, won't it?

docs/architecture/adr-112-proposer-based-timestamps.md
cason: The point here, and this is some analysis from PBTS v1 (namely, I was not part of it): https://github.com/cometbft/cometbft/blob/main/spec/consensus/proposer-based-timestamp/v1/pbts_001_draft.md

- BF Time: If more than 1/2 of the voting power is held by faulty validators, then the time is under total control of the faulty validators. Here faulty combines Byzantine and "correct"  validators with non synchronized clocks.
- PBTS: the block time can be corrupted only in the extreme case when >2/3 of the validators are faulty. Corrupted here means outside the synchrony window.

docs/architecture/adr-112-proposer-based-timestamps.md
cason: Yes, at the prevote step.

docs/architecture/adr-112-proposer-based-timestamps.md
cason: Sure, didn't notice that inconsistency in the original text.

docs/architecture/adr-112-proposer-based-timestamps.md
cason: Yes. There is a commented part referring that vote timestamps become useless. This is commented because we don't change the vote message structure here. But this future/possible approach deserves at least a comment.

docs/architecture/adr-112-proposer-based-timestamps.md
```diff
@@ -183,58 +190,61 @@ The proposer will set the re-proposed block's `Header.Timestamp` as the `Proposa
...
+The validator will check that the `Proposal.Timestamp` is at most `PRECISION` greater than `proposalReceiveTime`, and at maximum `PRECISION + MSGDELAY` less than `proposalReceiveTime`.
 If the timestamp is not within these bounds, the proposed block will not be considered `timely`.
```
sergio-mena, line 248:
```suggestion
If the timestamp is not within these bounds, the proposed block will not be considered `timely` and the validator will thus prevote `nil`.
```

docs/architecture/adr-112-proposer-based-timestamps.md
```diff
@@ -267,15 +277,18 @@ If a CometBFT network selects a `MessageDelay` parameter that does not accuratel
 A very common way to update `ConsensusParams` is by executing a transaction included in a block that specifies new values for them.
 However, if the network is unable to produce blocks because of this liveness issue, no such transaction may be executed.
 To prevent this dangerous condition, we will add a relaxation mechanism to the `Timely` predicate.
+
+<!---
```
sergio-mena, line 281: Can we describe here the decision we make on Mon? Or is it coming in a further PR?

docs/architecture/adr-112-proposer-based-timestamps.md
```diff
@@ -292,17 +305,31 @@ The only change we will make to the prevote step is to what a validator consider
...
+
+### Backwards compatibility
+
+In order to ensure backwards compatibility, PBTS should be enabled using a consensus parameter.
+The proposed approach is similar to the one adopted to enable vote extensions:
+[`ABCIParams.VoteExtensionsEnableHeight`](https://github.com/cometbft/cometbft/blob/main/spec/abci/abci++_app_requirements.md#abciparamsvoteextensionsenableheight).
+
+In summary, the network will migrate from the `BFT Time` method for assigning
+and validating timestamps to the new method for assigning and validating
+timestamps adopted by `PBTS` from a given, configurable height.
```
sergio-mena: I would add here:
* Once PBTS is activated, there are no provisions for a chain to revert back to BFT Time (the latter being considered obsolete).
* We decided that blocksync will not run BFT Time validation in blocks created before PBTS was activated on the chain. The reasoning is that, if the light client algorithm accepts an already produced block, we assume the BFT Time algorithm was followed. This reasoning is analogous to why we don't check `timely` when precommiting a proposal with a polka.

docs/architecture/adr-112-proposer-based-timestamps.md
```diff
@@ -292,17 +305,31 @@ The only change we will make to the prevote step is to what a validator consider
...
+
+For more discussion of this, see [issue 2063][issue2063].
 
 ## Future Improvements
 
 * Implement BLS signature aggregation.
-By removing fields from the `Precommit` messages, we are able to aggregate signatures.
+If we remove the `Timestamp` field from the `Precommit` messages, we are able to aggregate signatures.
```
sergio-mena:
```suggestion
If we remove the `Timestamp` field from the `Precommit` messages, we are able to aggregate signatures.
We have however left this improvement out for the time being, as it would break the block format and validation
rules (signature verification) and thus may force a hard-fork on chains upgrading to the latest version of CometBFT.
We will remove the timestamps in votes when changing the block format is supported in CometBFT without
requiring a hard-fork (this feature is called Soft Upgrades).
```

docs/architecture/adr-112-proposer-based-timestamps.md
```diff
@@ -318,7 +345,7 @@ By removing fields from the `Precommit` messages, we are able to aggregate signa
 
 * Alters CometBFT’s liveness properties.
 Liveness now requires that all correct validators have synchronized clocks within a bound.
-Liveness will now also require that validators’ clocks move forward, which was not required under `BFTTime`.
+Liveness will now also require that validators’ clocks move forward, which was not required under `BFT Time`.
```
sergio-mena: Above, the first bullet of positive, I'm not sure I agree with the wording of that. Since byzantine validators (which will eventually, and infinitely often, become proposers) can influence the time within the limits of MSGDELAY and PRECISION

docs/architecture/adr-112-proposer-based-timestamps.md
```diff
@@ -329,6 +356,12 @@ This skew will be bound by the `PRECISION` value, so it is unlikely to be too la
```
sergio-mena: On line above, if such a situation currently exists on a chain, its community may decide to keep BFT Time (by not activating PBTS via its enable height) until the problem is fixed: BFT Time will let the chain make progress, by adding one millisecond to each new block; so if the average block time is > 1ms, the chain's time will eventually catch up.

docs/architecture/adr-112-proposer-based-timestamps.md
sergio-mena: I agree. I'm always confused because of the way our code is organized: we run the same code (validator vs full node), the only difference being that, at the moment of voting, we skip that if not a validator.

docs/architecture/adr-112-proposer-based-timestamps.md
sergio-mena: Yeah, but is you allow "correct" process to have unsynchronized clocks on one side and forbid is on the other side, aren't we then comparing apples and oranges?

docs/architecture/adr-112-proposer-based-timestamps.md
cason: Not accepted in these terms, because of the confusion between node and validator. Fixed in a different commit.

docs/architecture/adr-112-proposer-based-timestamps.md
cason: Yes, we are. But we have also changed the system model.

I updated point 2 there with the assumption of synchronized clocks.

docs/architecture/adr-112-proposer-based-timestamps.md
cason: Not sure. I would wait for the implementation. Moreover, we have to document the chosen solution, so that PR could also update this ADR.

docs/architecture/adr-112-proposer-based-timestamps.md
cason: The second point is something worth discussing.

docs/architecture/adr-112-proposer-based-timestamps.md
cason: I hate merging stuff which prevents accepting suggestions. Added this change in a new commit.

docs/architecture/adr-112-proposer-based-timestamps.md
cason: Added `arbitrarily` to the sentence.

docs/architecture/adr-112-proposer-based-timestamps.md
cason: This is on the future work section. More detailed now, following your suggestion in other comment.

docs/architecture/adr-112-proposer-based-timestamps.md
cason: Yes, this [comment](https://github.com/cometbft/cometbft/pull/2223#discussion_r1480350702), right?

docs/architecture/adr-112-proposer-based-timestamps.md
cason: We would need a section for that, if we are making this change (which I think it is not needed), see this [comment](https://github.com/cometbft/cometbft/pull/2205#discussion_r1480207485).

docs/architecture/adr-112-proposer-based-timestamps.md
cason: Added this "workaround" to the description, while improving the description of this "pause".

cason: Added an action item to #2063 to address this "workaround" in the documentation.

docs/architecture/adr-112-proposer-based-timestamps.md
cason: Added the first point to this description.

docs/architecture/adr-112-proposer-based-timestamps.md
sergio-mena: OK

docs/architecture/adr-112-proposer-based-timestamps.md
sergio-mena: Thanks, and the second, we have anyway changed the decision, to not change the checks when in BFT Time

docs/architecture/adr-112-proposer-based-timestamps.md
sergio-mena: 👍

sergio-mena (approved):
🚀

PR #2226, opened 2024-02-02
Allow customization of metrics

firelizzard18:
This is a proof of concept for allowing applications to manage the labels applied to metrics. The application I work on (written in Go) incorporates two related but distinct CometBFT networks. Thus each node runs two CometBFT nodes within the same process. Due to how CometBFT initializes metrics, I am forced to use a different namespace for metrics for each node, which leads to unpleasantly awkward queries in Grafana.

The change I propose:
1. Modifies the `PrometheusMetrics` methods so that they **do not apply labels** immediately.
2. Adds helper `With` methods so that change (1) is not a burden.
3. Exposes metrics struct types (some are currently internal).

As a result:
- Applications can add their own labels that are applied to all metrics.
- My application can register the metrics once with the default namespace, instead of being forced to register with different namespaces to avoid a panic. See the example in node/metrics_test.go.
- I can use more natural queries.

Here's a comparison of a query before and after:
```
// Returns multiple different metric names due to the wildcard
{__name__=~"cometbft_.*_consensus_num_txs"}

// All results have the same name, with different label sets (like a normal query)
cometbft_consensus_num_txs
```

### DIFF
--- a/internal/blocksync/metrics.gen.go
+++ b/internal/blocksync/metrics.gen.go
@@ -8,42 +8,38 @@ import (
 	stdprometheus "github.com/prometheus/client_golang/prometheus"
 )
 
-func PrometheusMetrics(namespace string, labelsAndValues ...string) *Metrics {
-	labels := []string{}
-	for i := 0; i < len(labelsAndValues); i += 2 {
-		labels = append(labels, labelsAndValues[i])
-	}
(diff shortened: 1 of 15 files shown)


firelizzard18:
@ebuchman :wave:

Issue #2225, opened 2024-02-01
mempool: panic when app's CheckTx returns an error while rechecking


hvanz:
Thanks @p4u for reporting this bug!

### Setup

**CometBFT version**: from v0.38 and up

### What happened?

Comet panics with the message "recheck cursor is not nil before resCbFirstTime", which is found in this line: https://github.com/cometbft/cometbft/blob/1f430f51f0e390cd7c789ba9b1e9b35846e34642/mempool/clist_mempool.go#L313

Briefly, when we are using a local ABCI client and the app's `CheckTx` method returns a non-nil error, the rechecking process is broken at [this line](https://github.com/cometbft/cometbft/blob/1f430f51f0e390cd7c789ba9b1e9b35846e34642/mempool/clist_mempool.go#L679), leaving `recheckCursor` and `recheckEnd` in an inconsistent state.

#### These are the steps to reach the panic line:
- Rechecking is started by `recheckTxs`, which:
  - sets `recheckCursor` to the first element of the transaction list `txs`
  - for every element in `txs`, it sends a CheckTx request to the app. We are using a local client, so the call to `proxyAppConn.CheckTxAsync` is synchronous, meaning that it will the `CheckTx` method on the app and immediately call the `globalCb` callback to process the response.
  - If the app's `CheckTx` returns an error, the function `recheckTxs` [returns](https://github.com/cometbft/cometbft/blob/1f430f51f0e390cd7c789ba9b1e9b35846e34642/mempool/clist_mempool.go#L679) without finishing the rechecking process and leaving `recheckCursor` with a non-nil value. `recheckCursor` is set to nil again only in `resCbRecheck`, when it's equal to `recheckEnd`.
- A `CheckTx` request of type `CHECK_TX_TYPE_CHECK` is sent to the app for a new tx; a response is handled by `globalCb`, which will panic because `recheckCursor` is not nil.

### Notes

On previous versions of CometBFT, `CheckTxAsync` didn't return an error, so this problem didn't exist. And when this new behaviour was introduced, [the documentation](https://github.com/cometbft/cometbft/blob/main/spec/abci/abci%2B%2B_basic_concepts.md?plain=1#L392-L424) was not updated.

p4u:
I've opened a hotfix PR (#2228) to tackle an immediate panic issue.

[...]

PR #2228, opened 2024-02-02
fix(mempool): panic when app's CheckTx returns an error while rechecking

p4u:
Hotfix for #2225

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -676,7 +676,6 @@ func (mem *CListMempool) recheckTxs() {
 		})
 		if err != nil {
 			mem.logger.Error("recheckTx", err, "err")
-			return
 		}
 	}
 


melekes (approved):

mempool/clist_mempool.go
```diff
@@ -676,7 +676,6 @@ func (mem *CListMempool) recheckTxs() {
 		})
 		if err != nil {
 			mem.logger.Error("recheckTx", err, "err")
-			return
 		}
```
mzabaluev, line 679: Wouldn't it be better to nullify `recheckCursor` (and maybe `recheckEnd`) and still make an early return? After a transaction check has failed, is it wasted work to check the remaining transactions?

We also may have a "too much object state" problem where this method modifies mempool object members that may be used by other methods outside of a single recheck run. This could be despaghettified by factoring the recheck state off to a separate object that is passed to the global callback instead of, or alongside with, the mempool object.

hvanz (changes requested):
Without at least a test we cannot have some certainty that this PR fixes the problem. And I think that removing the `return` line is not the solution.

Suppose that the request on the last tx on the list returns an error. Then `globalCb` won't be called and `recheckCursor` won't be set to `nil`.
This is because the [local client](https://github.com/cometbft/cometbft/blob/1f430f51f0e390cd7c789ba9b1e9b35846e34642/abci/client/local_client.go#L53-L60) will first call CheckTx on the app, and if there is an error, it will return the error without calling the callback.

mempool/clist_mempool.go
hvanz: This may work for the local client but not for the socket client. The first one makes sync calls to the app and calls the callback after each response. The second sends all the requests and then handles the responses asynchronously.

p4u:
What's the reason to add errors to the checkTx() application call? I don't understand the point for the current behavior: "if one of the transactions from the mempool returns an error, no more transactions are checked".

This behavior is dangerous, since it's not what usually a developer expects from a callback. To stop iterations, usually there is a boolean that indicates the caller should stop, not an error. 

@hvanz @mzabaluev

hvanz:
I agree that CheckTx should not return an error but we would need to change the interface for that. For the moment I'd recommend to return always nil, and encode any error in the CheckTx response.

hvanz:
Here's a test that fails with and without the `return` line: https://github.com/cometbft/cometbft/compare/hvanz/mempool-test-recheck-return-error

melekes:
It sounds like a more complex solution is needed. Let's continue the conversation in the issue.

PR #2236, opened 2024-02-05
fix(go.mod): do not use `replace` for api

melekes:
otherwise it's hard to use our qa-infra


mzabaluev (approved):
:bowing_man:

mzabaluev:
> otherwise it's hard to use our qa-infra

More than that, it's impossible to use `cometbft/cometbft` as a remote module. Which is a shame since it's supposed to be a library as well.

mzabaluev:
I have verified that qa-infra works now.

PR #2241, opened 2024-02-05
docs(ADR): #113 Modular transaction hashing

### LINKED ISSUE #342: ADR for modular transaction hashing
CometBFT should support pluggable cryptographic functionality in order to allow users to address a wider range of use cases than possible with the default cryptography we provide. At present the only way they can do this is to fork the project.

Specific use cases here are to be clarified so the team understands what kinds of interface(s) to implement. Another concern potentially addressed by this is one of mitigating security-related risk through isolation of cryptography-related dependencies (probably our highest-impact dependencies) in order to make them easier to replace in case of vulnerabilities.

The first thing we need here is a design in the form of an ADR, and it must be clear from the ADR as to how the aforementioned use cases could be implemented using the design.

melekes:
[RENDERED](https://github.com/cometbft/cometbft/blob/anton/342-txi/docs/references/architecture/adr-113-modular-transaction-hashing.md)

Closes #342


adizere:
@itsdevbear @fedekunze @JayT106 super simple ADR, please have a look.

adizere (approved):
Concept ACK, thanks Anton!

I pinged a couple of users to gather some feedback on this, let's wait on their input before merging.

cason:
Hey, before reviewing this, do you mind using 113 for this ADR? We are working in the PBTS feature branch on an ADR 112.

docs/references/architecture/adr-112-txi.md
```diff
@@ -0,0 +1,84 @@
...
+
+- 2024-02-05: First version (@melekes)
+
+## Status
+
+TBD
+
+## Context
+
+Transaction hashing in CometBFT is currently implemented using `tmhash`
```
cason, line 13:
```suggestion
Transaction hashing in CometBFT is currently implemented using `crypto/tmhash`
```

docs/references/architecture/adr-112-txi.md
```diff
@@ -0,0 +1,84 @@
...
+package, which itself relies on `sha256` to calculate a transaction's hash.
```
cason, line 14:
```suggestion
package, which itself relies on `[sha256](https://pkg.go.dev/crypto/sha256)` to calculate a transaction's hash.
```

docs/references/architecture/adr-112-txi.md
```diff
@@ -0,0 +1,84 @@
...
+
+`tmhash` is also used to calculate various hashes in CometBFT (e.g.,
+evidence, consensus params, commit, header, partset header).
+
+The problem some application developers are facing is a mismatch between the
+internal/app representation of transactions and the one employed by CometBFT. For
+example, [Evmos](https://evmos.org/) wants transactions to be hashed using
+the [RLP][rlp].
+
+In order to be flexible, CometBFT needs to allow changing the hashing algorithm
```
cason, line 28:
```suggestion
In order to be flexible, CometBFT needs to allow changing the transaction hashing algorithm
```

docs/references/architecture/adr-112-txi.md
```diff
@@ -0,0 +1,84 @@
...
+## Status
+
+TBD
+
+## Context
+
+Transaction hashing in CometBFT is currently implemented using `tmhash`
+package, which itself relies on `sha256` to calculate a transaction's hash.
+
+The hash is then used by the built-in indexer (to index this
```
cason, line 16: Can you elaborate more this list?

The mempool implementation uses transaction hashes. Both to produce log messages and as a key to maps of transactions (`types.Tx.Key()` is a hash as an array with fixed size).

docs/references/architecture/adr-112-txi.md
```diff
@@ -0,0 +1,84 @@
...
+transaction) and by the RPC `tx` and `tx_search` endpoints, which allow users
+to search for a transaction using a hash.
+
+`tmhash` is also used to calculate various hashes in CometBFT (e.g.,
+evidence, consensus params, commit, header, partset header).
```
cason, line 21: Not true for commit, header and partset header. 

They use the `merkle` package which, it is true, uses `sha256`.

docs/references/architecture/adr-112-txi.md
```diff
@@ -0,0 +1,84 @@
...
+
+The problem some application developers are facing is a mismatch between the
+internal/app representation of transactions and the one employed by CometBFT. For
+example, [Evmos](https://evmos.org/) wants transactions to be hashed using
+the [RLP][rlp].
```
cason, line 26: Do we have an issue or a reference for that?

docs/references/architecture/adr-112-txi.md
```diff
@@ -0,0 +1,84 @@
...
+In order to be flexible, CometBFT needs to allow changing the hashing algorithm
+if desired by the app developers.
+
+## Alternative Approaches
+
+None.
+
+## Decision
+
+Give app developers a way to provide their own hash function.
```
cason, line 37:
```suggestion
Give app developers a way to provide their own transaction hash function.
```

docs/references/architecture/adr-112-txi.md
```diff
@@ -0,0 +1,84 @@
...
+func (TMHash) Size(bz []byte) {
+    return tmhash.Size
+}
+
+func DefaultHashFn() HashFn {
+    return TMHash{}
+}
+```
+
+And then use it to calculate a transaction's hash.
```
cason, line 64: Can you be more specific regarding that?

In the case of `types/Tx`, calculating the hash is hard-coded. How can we parameterize this?

cason: In the case we want this change to apply to the RPC endpoints and, possibly, indexer is one thing. In this case we can use any customized hash function.

But if we want to adopt customized hash functions for our types, this is a completely different discussion.

docs/references/architecture/adr-112-txi.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Positive
+
+- Modular transaction hashing
+- Paving a way for the pluggable cryptography. hashFn can later be used to
+  calculate header's hash, evidence hash.
+
+### Negative
+
+- One more argument in `NewNode`
```
cason, line 76: Not really, it can be an `Option`: https://github.com/cometbft/cometbft/blob/main/node/node.go#L96

docs/references/architecture/adr-112-txi.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Neutral
```
cason, line 78: Not sure if performance should be considered here at some of the consequences? Probably neutral, since it really depends on the hash method adopted.

cason (commented):
Nice write up.

I agree with the approach, but I think we should really restrict the scope of the proposed changes. I know, for instance, that the RPC endpoints use some weird way to "print" (i.e., representing as strings) transactions and their IDs/hashes. But this is a problem of `String()` method. That I am afraid that is not solved using this approach.

cason:
Recall also to add an entry to the `README.md` file in the same directory.

docs/references/architecture/adr-112-txi.md
melekes: > Not true for commit, header and partset header.
> They use the merkle package which, it is true, uses sha256.

https://github.com/cometbft/cometbft/blob/d02b25e0fd278bc7e70142190ae5b3b8e722a8df/crypto/merkle/hash.go

are you sure?

andynog (commented):
The overall approach seems OK, but my concern is related to the scope and implications of this change. 

For example, in `types/tx.go` the code below might affect the Merkle root hash
```
func (txs Txs) Hash() []byte {
	hl := txs.hashList()
	return merkle.HashFromByteSlices(hl)
}
```

because in `txs.hashList()` the `Hash()` is invoked
```
func (txs Txs) hashList() [][]byte {
	hl := make([][]byte, len(txs))
	for i := 0; i < len(txs); i++ {
		hl[i] = txs[i].Hash()
	}
	return hl
}
```
which in `merkle.HashFromByteSlices(hl)` will make the Merkle root hash different
```
func HashFromByteSlices(items [][]byte) []byte {
	return hashFromByteSlices(sha256.New(), items)
}
```
maybe this will be OK, but will this have implications later on ? Like during an upgrade, could the hash algorithm be changed ? what are the implications on verifying and validating if the Merkle is not valid.

Also, is this OK with IBC and proofs ? Like what if one chain decides to switch the hash algorithm? what are the validity of proof guarantees ? 

Anyway, the scope and implications of this change seems to be a lot broader and it is a good idea to think this through. Maybe my concerns are not valid but just bringing this up 😉

docs/references/architecture/adr-112-txi.md
cason: True, in the verification side, sorry.

cason:
> The overall approach seems OK, but my concern is related to the scope and implications of this change.

Fully agree with that, much better explained by @andynog. Hashing transactions is something useful for the mempool reactor and for RPC endpoints.

Other uses of this library includes block, merkle tree, evidence production and verification. For this case, being compatible is essential. A suggestion here, for instance, is to adopt `ed25519` directly, so it is clear to everyone that this is the standard for blockchain hashes.

docs/references/architecture/adr-113-txi.md
```diff
@@ -0,0 +1,95 @@
...
+to search for a transaction using a hash.
+
+`tmhash` is also used to calculate various hashes in CometBFT (e.g.,
+evidence, consensus params, commit, header, partset header).
+
+The problem some application developers are facing is a mismatch between the
+internal/app representation of transactions and the one employed by CometBFT. For
+example, [Evmos](https://evmos.org/) wants transactions to be hashed using
+the [RLP][rlp].
+
```
lasarojc, line 27: Is there any need to also RLP encoding or really just the hasn?

robert-zaremba:
NOTE: We should be able to customize the Merkle Tree hashing as well. But this has to be done in the IAVL level.

docs/references/architecture/adr-113-txi.md
```diff
@@ -0,0 +1,95 @@
...
+- 2024-02-05: First version (@melekes)
+
+## Status
+
+TBD
+
+## Context
+
+Transaction hashing in CometBFT is currently implemented using `crypto/tmhash`
+package, which itself relies on `[sha256](https://pkg.go.dev/crypto/sha256)` to calculate a transaction's hash.
```
cason, line 14:
```suggestion
package, which itself relies on [`sha256`](https://pkg.go.dev/crypto/sha256) to calculate a transaction's hash.
```

docs/references/architecture/adr-113-txi.md
```diff
@@ -0,0 +1,95 @@
...
+
+None.
+
+## Decision
+
+Give app developers a way to provide their own transaction hash function.
+
+## Detailed Design
+
+Add `hashFn HashFn` option to `NewNode` in `node.go`.
```
cason, line 41: Should we be specific and add Transaction or Tx to the name and description of the methods?

docs/references/architecture/adr-113-txi.md
```diff
@@ -0,0 +1,95 @@
...
+
+```go
+// HashFn defines an interface for the hashing function.
+type HashFn interface {
```
cason, line 45: Doesn't this interface needs a `Sum()` method?

docs/references/architecture/adr-113-txi.md
```diff
@@ -0,0 +1,95 @@
...
+func DefaultHashFn() HashFn {
+    return TMHash{}
+}
+
+// TMHash uses tmhash.
+type TMHash struct {}
+func (TMHash) New() hash.Hash {
+    return tmhash.New()
+}
+func (TMHash) Size(bz []byte) {
```
cason, line 69: This is `Sum`, I think.

Missing `Size() int`.

docs/references/architecture/adr-113-txi.md
```diff
@@ -0,0 +1,95 @@
...
+    return tmhash.Size
+}
+```
+
+And then use it to calculate a transaction's hash.
```
cason, line 74: Can we get a list of where in the code we need to use this hash function? And which interfaces we would need to change?

docs/references/architecture/adr-113-txi.md
```diff
@@ -0,0 +1,95 @@
...
+## Consequences
+
+### Positive
+
+- Modular transaction hashing
+- Paving a way for the pluggable cryptography. hashFn can later be used to
+  calculate header's hash, evidence hash.
+
+### Negative
+
```
cason, line 85: Added complexity? Not really a problem, but a negative side.

adizere: agree

9ee941ab2

docs/references/architecture/adr-113-txi.md
adizere: Judging from the [Evmos fork and initial investigation in this comment](https://github.com/cometbft/cometbft/issues/342#issuecomment-1888821207), encoding was not a concern.

adizere:
It would be good if we can demonstrate that the approach in this ADR can achieve the same effects as in https://github.com/evmos/cometbft/pull/1.

docs/references/architecture/adr-113-txi.md
```diff
@@ -0,0 +1,97 @@
...
+## Detailed Design
+
+Add `hashFn HashFn` option to `NewNode` in `node.go`.
+
+```go
+// HashFn defines an interface for the transaction hashing function.
+type HashFn interface {
+    // New returns a new hash.Hash calculating the given hash function.
+    New() hash.Hash
```
fedekunze, line 47: what is a `hash.Hash`? It's not defined anywhere. 

Is it an interface? We need to use geth's `common.Hash` which has 32 bytes of size and formats the hash using `0x....` instead of using uppercase of the hex value without the `0x` prefix

docs/references/architecture/adr-113-txi.md
```diff
@@ -0,0 +1,97 @@
...
+    // Size returns the length, in bytes, of a digest resulting from the given hash function.
+    Size() int
+}
+
+// Use CustomHashFn to supply custom hashing function.
+func CustomHashFn(hashFn HashFn) Option {
```
fedekunze, line 53: what's `Option`? I assume it's for the node but needs to be defined

fedekunze: it's also not clear where you connect this custom hashing function to CometBFT

docs/references/architecture/adr-113-txi.md
```diff
@@ -0,0 +1,97 @@
...
+
+Give app developers a way to provide their own transaction hash function.
+
+## Detailed Design
+
+Add `hashFn HashFn` option to `NewNode` in `node.go`.
+
+```go
+// HashFn defines an interface for the transaction hashing function.
+type HashFn interface {
```
fedekunze, line 45: we should also have a custom `String` for formatting the hash result

docs/references/architecture/adr-113-txi.md
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// DefaultHashFn is a TMHash.
+func DefaultHashFn() HashFn {
+    return TMHash{}
+}
+
+// TMHash uses tmhash package to implement transaction hashing.
+type TMHash struct {}
+func (TMHash) New() hash.Hash {
```
fedekunze, line 66: this needs to be more explicit. Consider renaming the function to `Hash() hash.Hash`

docs/references/architecture/adr-113-txi.md
fedekunze: agree, this is missing

docs/references/architecture/adr-113-txi.md
```diff
@@ -0,0 +1,97 @@
...
+	}
+}
+
+// DefaultHashFn is a TMHash.
+func DefaultHashFn() HashFn {
+    return TMHash{}
+}
+
+// TMHash uses tmhash package to implement transaction hashing.
+type TMHash struct {}
```
fedekunze, line 65: let's add an example with the ethereum tx hash:

```go
// EVMTxRLPHash uses RLP from an Ethereum tx to implement transaction hashing.
type EVMTxRLPHash struct {
   tx *types.Transaction
}

func (h EVMTxRLPHash) Hash() hash.Hash {
   return h.tx.Hash()
}

func (h EVMTxRLPHash) String() string {
   return h.tx.Hash().String()
}

func (EVMTxRLPHash) Size() int {
  return common.HashLength
}
```

docs/references/architecture/adr-113-txi.md
cason: This is a go standard interface: https://pkg.go.dev/hash#Hash

fedekunze: I see. The `Sum(b []byte) []byte` function from `hash.Hash` is not relevant in the case of Ethereum txs because the hash is computed from the transaction fields and not the transaction bytes. Unless the bytes are unmarshalled to the `Transaction` struct and from there call `tx.Hash()`

docs/references/architecture/adr-113-txi.md
```diff
@@ -0,0 +1,97 @@
...
+
+- 2024-02-05: First version (@melekes)
+
+## Status
+
+TBD
+
+## Context
+
+Transaction hashing in CometBFT is currently implemented using `crypto/tmhash`
```
fedekunze, line 13: Is this ADR meant only for transactions or also for the `Block` hash?

docs/references/architecture/adr-113-txi.md
fedekunze, line 45: Not, should be renamed to `TxHashFn` for clarity

docs/references/architecture/adr-113-txi.md
cason: At the moment, only for transaction's hash. I would make it very precise in the description.

adizere:
blocked on #2610

andynog:
> blocked on #2610

#2610 got merged, so I'm assuming this is good to merge @adizere and @melekes ? I fixed a small conflict so should pass all the tests now

adizere:
> > blocked on #2610
> 
> #2610 got merged, so I'm assuming this is good to merge @adizere and @melekes ? I fixed a small conflict so should pass all the tests now

I think we have two options:
1. merge this, open a new PR with a new ADR 114 (?) which fixes [Fede's comments](https://github.com/cometbft/cometbft/pull/2241#pullrequestreview-1920405754), mark ADR 113 as 'not adopted' and the new ADR 114 as 'proposed' or something similar
2. fix the present ADR to address Fede's comments

probably option 2 is faster and better, so I would suggest doing that

adizere (dismissed):
Fix Fede's & Daniels comments

docs/references/architecture/adr-112-txi.md
melekes: > But if we want to adopt customized hash functions for our types, this is a completely different discussion.

I'd like for us to have both discussions.

docs/references/architecture/adr-113-txi.md
melekes: If we decide to restrict the scope to just txs, yes.

docs/references/architecture/adr-113-txi.md
melekes: Fixed by using a global variable.

docs/references/architecture/adr-113-modular-transaction-hashing.md
```diff
@@ -0,0 +1,154 @@
...
+Give app developers a way to provide their own transaction hash function.
+
+## Detailed Design
+
+Use `sha256` by default, but give developers a way to change the hashing function:
+
+```go
+import (
+	"crypto"
+    "hash"
```
sergio-mena, line 55:
```suggestion
	"hash"
```

docs/references/architecture/adr-113-txi.md
melekes: Fixed in new version.

> what's Option?

https://pkg.go.dev/github.com/cometbft/cometbft@v0.38.7/node#Option

docs/references/architecture/adr-113-txi.md
melekes: Done

docs/references/architecture/adr-113-txi.md
melekes: This is up for discussion. @fedekunze would you benefit from changing `Block` hash?

docs/references/architecture/adr-113-txi.md
melekes: Pending discussion.

docs/references/architecture/adr-113-modular-transaction-hashing.md
```diff
@@ -0,0 +1,154 @@
...
+	}
+)
+
+// New returns a new hash.Hash calculating the given hash function.
+func New() hash.Hash {
+	return Hash.New()
+}
+
+// ReplaceDefaultHashWith replaces the default SHA256 hashing function with a given one.
+func ReplaceDefaultHashWith(h crypto.Hash) {
```
sergio-mena, line 80: How about `SetHashFn` ?

docs/references/architecture/adr-113-modular-transaction-hashing.md
```diff
@@ -0,0 +1,154 @@
...
+`ReplaceStringFuncWith` allows developers to swap the default string function
+(`fmt.Sprintf("%X", bz)`) with their own implementation.
+
+The above solution changes hashes across the whole CometBFT, meaning header's
+hash, evidence's hash, data's hash, etc.
+
+If the application developer decides to change the default hashing scheme, they
+can only do so once before launching their app. If they attempt to upgrade
+after without a hard fork, the resulting hashes won't match. A hard fork would
+work.
```
sergio-mena, line 118: That's a very important point. And I would add some reference to the existing light clients which will all probably get broken if the app dev's change the hashing function after genesis.

docs/references/architecture/adr-113-modular-transaction-hashing.md
```diff
@@ -0,0 +1,154 @@
...
+
+The hashing function needs to be added to `Header`, so that light clients are aware of
+the function to use for verification.
+
+```go
+type Header struct {
+    // ...
+	Hash crypto.Hash `json:"hash"`
+}
+```
```
sergio-mena, line 128: If I am not missing something obvious, this would change the block format, forcing all chains upgrading to this version to hard-fork. Not an option at this stage.
We can't do this before we tackle Soft Upgrades (hopefully in Q3).

docs/references/architecture/adr-113-modular-transaction-hashing.md
```diff
@@ -0,0 +1,154 @@
...
+
+Light clients would then use ^ when calculating validators hash or header's hash.
```
sergio-mena, line 135: Light clients of a chain MUST know the hash function used in that chain. Maybe, we should include the hashing algorithm in a filed in the genesis file.

docs/references/architecture/adr-113-modular-transaction-hashing.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Positive
+
+- Modular transaction hashing
+
+### Neutral
+
+- App developers need to take performance into account when choosing custom
+  hash function.
+- IBC is not affected by this change since the proof contains `Hash` (hashing
```
sergio-mena, line 147: > since the proof contains `Hash`

Not sure I understand this part. Could you clarify?

docs/references/architecture/adr-113-modular-transaction-hashing.md
```diff
@@ -0,0 +1,154 @@
...
+)
+
+var (
+
+    // Hash used
+    Hash = crypto.SHA256
+
+	stringFunc = func(bz []byte) string {
+		return fmt.Sprintf("%X", bz)
+	}
```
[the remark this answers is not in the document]
melekes: Yes, I did. I found that having two separate things (object and a function is a cleaner approach)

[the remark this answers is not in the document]
sergio-mena: Fair enough

docs/references/architecture/adr-113-modular-transaction-hashing.md
melekes: You're right. We don't need to put it in the header.

docs/references/architecture/adr-113-modular-transaction-hashing.md
```diff
@@ -127,17 +127,34 @@ can only do so once before launching their app. If they attempt to upgrade
...
+) (*Client, error) {
 ```
 
-Light clients would then use ^ when calculating validators hash or header's hash.
+To help discover the hash function used by the other zone, we can add a new
+field to the genesis file:
+
+```json
+{
+    "hash": 5, // crypto.SHA256 (https://pkg.go.dev/crypto#Hash)
```
sergio-mena, line 155: What does "5" mean here?

sergio-mena (approved):
🚀

docs/references/architecture/adr-113-modular-transaction-hashing.md
```diff
@@ -0,0 +1,181 @@
...
+The transaction hash is used by:
+
+- the built-in transaction indexer;
+- the `/tx` and `/tx_search` RPC endpoints, which allow users
+to search for a transaction using its hash;
+
+The problem some application developers are facing is a mismatch between the
+internal/app representation of transactions and the one employed by CometBFT. For
+example, [Evmos](https://evmos.org/) wants transactions to be hashed using
+the [RLP][rlp].
```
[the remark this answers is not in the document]
melekes: It's a reference-style link (https://www.markdownguide.org/basic-syntax/#reference-style-links). The actual path is defined at the end of the document.

docs/references/architecture/adr-113-modular-transaction-hashing.md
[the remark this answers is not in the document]
melekes: I agree, but the problem is that the standard `hash.Hash` https://pkg.go.dev/hash#Hash doesn't have a `String` method, so this will require our own interface, meaning we won't be able to use `crypto.Hash` from the Go stdlib. Again, nothing is wrong, but ideally, we'd use standard interfaces.

docs/references/architecture/adr-113-modular-transaction-hashing.md
melekes: Go's stdlib `Hash` https://pkg.go.dev/crypto#Hash. We can also use strings like `SHA256` if that's more practical

docs/references/architecture/adr-113-modular-transaction-hashing.md
melekes: I was looking at the wrong thing, sorry. Fixed

melekes:
@fedekunze If we only allowed changing the hash function for transactions (how they are hashed in mempool and displayed, plus the Merkle root hash of transactions—`DataHash` in the block's header), would it suffice for your use case?

docs/references/architecture/adr-113-modular-transaction-hashing.md
```diff
@@ -0,0 +1,142 @@
...
+- 2024-05-28: Complete refactor (@melekes)
+- 2024-06-07: Limit the scope to transaction hashing (@melekes)
+
+## Status
+
+Proposed
+
+## Context
+
+Hashing in CometBFT is currently implemented using `crypto/tmhash`
```
cason, line 15: Although we directly import `crypto/sha256` in many places:

```
$ grep crypto/sha256 . -R 2> /dev/null
./statesync/snapshots.go:	"crypto/sha256"
./crypto/merkle/tree.go:	"crypto/sha256"
./crypto/merkle/bench_test.go:	"crypto/sha256"
./crypto/bls12381/key_bls12381.go:	"crypto/sha256"
./crypto/secp256k1/secp256k1.go:	"crypto/sha256"
./crypto/tmhash/hash_test.go:	"crypto/sha256"
./crypto/tmhash/hash.go:	"crypto/sha256"
./crypto/tmhash/bench_test.go:	"crypto/sha256"
./crypto/hash.go:	"crypto/sha256"
./types/tx.go:	"crypto/sha256"
./test/e2e/app/state.go:	"crypto/sha256"
./p2p/conn/secret_connection.go:	"crypto/sha256"
./mempool/cache_test.go:	"crypto/sha256"
./mempool/mempool.go:	"crypto/sha256"
```

docs/references/architecture/adr-113-modular-transaction-hashing.md
```diff
@@ -0,0 +1,142 @@
...
+Among the things which are hashed are the block's header, evidence, consensus
+params, commit, partset header, transactions and others.
+
+### Transaction hashing
+
+The transaction hash is used by:
+
+- the built-in transaction indexer;
+- the `/tx` and `/tx_search` RPC endpoints, which allow users
+to search for a transaction using its hash;
```
cason, line 27: It is also used in the mempool. Is it worth mentioning this?

docs/references/architecture/adr-113-modular-transaction-hashing.md
```diff
@@ -0,0 +1,142 @@
...
+all structs, not just transactions. But the implification of that change is
+quite significant. For example, if the chain is using a different hashing
+scheme, then it looses IBC-compatibility. The IBC modules assumes fixed hashing
+scheme. The destination chain needs to know the hashing function of the source
+chain in order to verify the validators hash.
+
+## Alternative Approaches
+
+1. Do nothing => not flexible.
+2. Add `HashFn` argument to `NewNode` and pass this function down the stack =>
```
cason, line 49: `HashFn` is a transaction hasher implementation?

docs/references/architecture/adr-113-modular-transaction-hashing.md
```diff
@@ -0,0 +1,142 @@
...
+   complicates the code.
+3. Allow changing the hashing function for all structs => breaks IBC
+   compatibility (see 'General hashing' above).
+
+## Decision
+
+Give app developers a way to provide their own hash function.
+
+## Detailed Design
```
cason, line 58: One solution I see in several places in the code base is:

1. Introduce an interface
2. Produce a standard implementation for the interface
3. Enable changing the implementation for the interface, which is essentially only used for testing

So I wonder why this wouldn't be a better solution.

docs/references/architecture/adr-113-modular-transaction-hashing.md
```diff
@@ -0,0 +1,142 @@
...
+    txHash = crypto.SHA256
+
+    // fmtHash is a function that converts a byte slice to a string.
+    fmtHash = func(bz []byte) string {
+        return fmt.Sprintf("%X", bz)
+    }
+)
+
+// SetFmtHash sets the function used to convert a checksum to a string.
+func SetFmtHash(f func([]byte) string) {
```
cason, line 80: This can only be called before anything is started, right? Otherwise we have a real synchronization problem here.

docs/references/architecture/adr-113-modular-transaction-hashing.md
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// Sum returns the checksum of the data as Bytes.
+func Sum(bz []byte) Bytes {
+	return Bytes(TxHash.Hash.Sum(bz))
+}
+```
+
+Let's break this down. By default, we use `sha256` standard crypto library.
+`SetTxHash` allows developers to swap the default hashing function
```
cason, line 107: Again, this is for transactions or anything?

docs/references/architecture/README.md
```diff
@@ -43,6 +43,7 @@ numbering our ADRs from 100 onwards.
 ### Proposed
 
 - [ADR-115: Predictable Block Times](adr-115-predictable-block-times.md)
+- [ADR-113: Modular crypto](adr-113-modular-transaction-hashing.md)
```
cason, line 46: The name here does not match the title of the next file.

cason (commented):
A further review round.

Should we consider a prototype implementation to test the idea?

docs/references/architecture/adr-113-modular-transaction-hashing.md
melekes: Clarified.

docs/references/architecture/adr-113-modular-transaction-hashing.md
melekes: Correct.

docs/references/architecture/adr-113-modular-transaction-hashing.md
melekes: It's an option that we pass to `NewNode` and forward to mempool, tx indexer, and tx RPC endpoints. However, the use case is rare, so I chose to opt for global variables instead.

docs/references/architecture/adr-113-modular-transaction-hashing.md
melekes: https://github.com/cometbft/cometbft/pull/2241#discussion_r1633278671

My main argument against passing a hasher everywhere is that it's rare when somebody wants a different hashing function, so changing the interface of mempool, tx indexer and RPC endpoints is unjustified. But maybe I'm wrong. Please let me know what you think.

cason:
I think we should also add a short paragraph mentioning that it is probably not possible/feasible to implement this via configuration parameter or command-line option, am I right? Because this could be seen as a trivial solution, but... unfeasible.

cason:
Did we mention https://github.com/tendermint/tendermint/issues/6539 and https://github.com/tendermint/tendermint/pull/6773 here? It would be nice, also to see the discussions back then.

melekes:
> I think we should also add a short paragraph mentioning that it is probably not possible/feasible to implement this via configuration parameter or command-line option, am I right? 

It's possible but I'm against this because

a) we're not sure this will be a final design/solution
b) we're optimizing for the default hashing function (by analogy, we may support different consensus algorithms, but we're optimizing for the default one)

PR #2246, opened 2024-02-06
refactor(p2p)!: Refactor PeerSet to eliminate data races & improve performance

### LINKED ISSUE #2157: p2p: PeerSet.Remove naively & inefficiently creates new slices each time discarding all the work preallocating, but could simply re-slice

### LINKED ISSUE #2158: p2p: Switch.OnStop uses stale and already mutated PeerSet.List() which could have data race consequences due to a leaky abstraction of using .List()

melekes:
Original PR: https://github.com/cometbft/cometbft/pull/2159

## Breaking changes

- `[p2p]` Rename `IPeerSet#List` to `Copy`, add `Random`, `ForEach` methods.
   Rename `PeerSet#List` to `Copy`, add `Random`, `ForEach` methods.

Fixes #2158

## Performance improvement

This change makes PeerSet.Remove much more efficient simply by using more idiomatic Go re-slicing to avoid the prior mechanisms of creating fresh peer set lists on just a single remove. While here also added a remedy for a found bug #2158 due to an abstraction that returns a stale slice to its caller in Switch.OnStop.

Benchmark results:

```shell
name                 old time/op    new time/op    delta
PeerSetRemoveOne-8     90.5µs ± 4%    95.9µs ±13%     ~     (p=0.218 n=10+10)
PeerSetRemoveMany-8    1.58ms ± 4%    1.50ms ± 1%   -4.98%  (p=0.000 n=10+8)

name                 old alloc/op   new alloc/op   delta
PeerSetRemoveOne-8     8.48kB ± 0%    7.92kB ± 0%   -6.60%  (p=0.000 n=10+10)
PeerSetRemoveMany-8     149kB ± 0%      65kB ± 0%  -56.44%  (p=0.000 n=10+10)

name                 old allocs/op  new allocs/op  delta
PeerSetRemoveOne-8       85.0 ± 0%      73.0 ± 0%  -14.12%  (p=0.000 n=10+10)
PeerSetRemoveMany-8     1.32k ± 0%     1.22k ± 0%   -7.51%  (p=0.000 n=10+10)
```

which savings become so much more when peers are removed much more frequently for a longer period of time and could mitigate DOS vectors.

Fixes #2157


p2p/peer_set.go
```diff
@@ -120,24 +120,22 @@ func (ps *PeerSet) Remove(peer Peer) bool {
...
 		return true
 	}
 
 	// Replace the popped item with the last item in the old list.
-	lastPeer := ps.list[len(ps.list)-1]
 	lastPeerKey := lastPeer.ID()
 	lastPeerItem := ps.lookup[lastPeerKey]
-	newList[index] = lastPeer
+	ps.list[index] = lastPeer
+	ps.list = ps.list[:len(ps.list)-1]
```
cason, line 137: Could you move this line to where it was before? The two changes that happen in both cases is to reduce `ps.list` by one element and removing the key from `ps.lookup`.

p2p/switch.go
```diff
@@ -364,6 +364,13 @@ func (sw *Switch) StopPeerGracefully(peer Peer) {
 }
 
 func (sw *Switch) stopAndRemovePeer(peer Peer, reason interface{}) {
+	if peer == nil {
+		// This condition can occur due to the fact that Switch.OnStop uses stale data and
+		// there is no easy fix for it per https://github.com/cometbft/cometbft/issues/2158
```
cason, line 369: Should this fix be mentioned in the changelog? Although we are not sure that it indeed solves the problem.

cason (commented):
The changes in the test unit have no effect at all. I would revert them.

cason:
By the way, the tests that are failing are failing because of race conditions. Isn't the case that by not copying the list of peers, instead changing it, is producing these errors?

cason:
> Fixes https://github.com/cometbft/cometbft/issues/2158

I don't think it solves this issue.

p2p/switch.go
melekes: No, it doesn't solve the fundamental problem. I agree with you. But it doesn't harm to have this condition imho.

melekes:
@cason I've cleaned up the code a bit + removed `Fixes #2158` from description.

melekes:
One thing that still bothers me here is that `*peer.Send` is not thread-safe, but it's called in multiple places.

rpc/core/net.go
```diff
@@ -13,20 +13,19 @@ import (
...
-	peersList := env.P2PPeers.Peers().List()
-	peers := make([]ctypes.Peer, 0, len(peersList))
-	for _, peer := range peersList {
-		nodeInfo, ok := peer.NodeInfo().(p2p.DefaultNodeInfo)
+	peers := make([]ctypes.Peer, 0)
+	env.P2PPeers.Peers().ForEach(func(p p2p.Peer) {
+		nodeInfo, ok := p.NodeInfo().(p2p.DefaultNodeInfo)
 		if !ok {
-			return nil, fmt.Errorf("peer.NodeInfo() is not DefaultNodeInfo")
+			panic(fmt.Sprintf("peer %v has the invalid node info type: %T ", p.ID(), p.NodeInfo()))
```
melekes, line 20: note that we refuse handshake if `p.NodeInfo()` is not `DefaultNodeInfo`, but I can revert the code to error instead.

cason (approved):
Great work, really.

I just wonder if there is a way to this not being a breaking change. For instance, preserving the `List()`  method while updating its implementation. Adding methods to an interface should not be breaking (I think).

.changelog/unreleased/breaking-changes/2246-p2p-peerset.md
```diff
@@ -0,0 +1,3 @@
+- `[p2p]` Rename `IPeerSet#List` to `Copy`, add `Random`, `ForEach` methods.
+   Rename `PeerSet#List` to `Copy`, add `Random`, `ForEach` methods.
```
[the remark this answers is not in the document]
melekes: No, there's a difference between IPeerSet and PeerSet

rpc/core/net.go
melekes: But let's play safe.

melekes:
> I just wonder if there is a way to this not being a breaking change. For instance, preserving the List() method while updating its implementation. Adding methods to an interface should not be breaking (I think).

We can do that, for sure. But why? If we're not going to backport it to v0.38. Copy better captures the semantics (`Peers().Copy()` instead of `Peers().List()`)

p2p/switch.go
```diff
@@ -273,7 +273,7 @@ func (sw *Switch) OnStop() {
 func (sw *Switch) Broadcast(e Envelope) chan bool {
 	sw.Logger.Debug("Broadcast", "channel", e.ChannelID)
 
-	peers := sw.peers.List()
+	peers := sw.peers.Copy()
```
cason, line 276: So, here I see a performance cost for this approach. `Broadcast` is used by several reactors (details [here](https://github.com/cometbft/cometbft/blob/main/spec/p2p/reactor-api/p2p-api.md#switch-api)) to broadcast (of course), periodic messages.

I wonder whether we can use the `ForEach` method here too. In particular because we start go routines to actually send messages, so we shouldn't keep the lock for long.

cason (commented):
Could you investigate whether this [information](https://github.com/cometbft/cometbft/blob/main/spec/p2p/reactor-api/p2p-api.md#peers-state) about the `Switch.Peers()` method is true? And if is not, do you mind updating the spec?

I think that most accesses to the peers lists are done through this method.

Except for tests, where the concurrency problems of `PeerSet.List()` are not really, to not say at all, a problem. And I've noticed that most of the changes in this PR are on test units.

cason:
> > I just wonder if there is a way to this not being a breaking change. For instance, preserving the List() method while updating its implementation. Adding methods to an interface should not be breaking (I think).
> 
> We can do that, for sure. But why? If we're not going to backport it to v0.38. Copy better captures the semantics (`Peers().Copy()` instead of `Peers().List()`)

I don't have strong opinions here. I would say to put someone else on the loop here.

melekes:
> Could you investigate whether this [information](https://github.com/cometbft/cometbft/blob/main/spec/p2p/reactor-api/p2p-api.md#peers-state) about the Switch.Peers() method is true? And if is not, do you mind updating the spec?

Done ☑️

p2p/switch.go
```diff
@@ -248,15 +248,15 @@ func (sw *Switch) OnStart() error {
...
+	for _, p := range sw.peers.Copy() {
 		sw.stopAndRemovePeer(p, nil)
 	}
 
 	// Stop reactors
 	sw.Logger.Debug("Switch: Stopping reactors")
 	for _, reactor := range sw.reactors {
 		if err := reactor.Stop(); err != nil {
-			sw.Logger.Error("error while stopped reactor", "reactor", reactor, "error", err)
+			sw.Logger.Error("Error while stopped reactor", "reactor", reactor, "err", err)
```
sergio-mena, line 259:
```suggestion
			sw.Logger.Error("error while stopped reactor", "reactor", reactor, "err", err)
```

sergio-mena: log lines should start with lowercase

p2p/switch.go
```diff
@@ -364,11 +362,14 @@ func (sw *Switch) StopPeerGracefully(peer Peer) {
 }
 
 func (sw *Switch) stopAndRemovePeer(peer Peer, reason interface{}) {
-	sw.transport.Cleanup(peer)
+	// Returning early if the peer is already stopped prevents data races because
+	// this function may be called from multiple places at once.
 	if err := peer.Stop(); err != nil {
-		sw.Logger.Error("error while stopping peer", "error", err) // TODO: should return error to be handled accordingly
+		sw.Logger.Error("Error stopping peer", "peer", peer.ID(), "err", err)
```
sergio-mena, line 368:
```suggestion
		sw.Logger.Error("error stopping peer", "peer", peer.ID(), "err", err)
```

p2p/switch.go
```diff
@@ -377,13 +378,14 @@ func (sw *Switch) stopAndRemovePeer(peer Peer, reason interface{}) {
...
 	// RemovePeer is finished.
 	// https://github.com/tendermint/tendermint/issues/3338
-	if sw.peers.Remove(peer) {
-		sw.metrics.Peers.Add(float64(-1))
-	} else {
+	if !sw.peers.Remove(peer) {
 		// Removal of the peer has failed. The function above sets a flag within the peer to mark this.
 		// We keep this message here as information to the developer.
-		sw.Logger.Debug("error on peer removal", ",", "peer", peer.ID())
+		sw.Logger.Debug("Error on peer removal", "peer", peer.ID())
```
sergio-mena, line 384:
```suggestion
		sw.Logger.Debug("error on peer removal", "peer", peer.ID())
```

melekes:
```
--- FAIL: TestReactorValidatorSetChanges (1.92s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x108284a]

goroutine 11988 [running]:
testing.tRunner.func1.2({0x130dfe0, 0x1d6e7d0})
	/opt/hostedtoolcache/go/1.21.6/x64/src/testing/testing.go:1545 +0x3f7
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.21.6/x64/src/testing/testing.go:1548 +0x716
panic({0x130dfe0?, 0x1d6e7d0?})
	/opt/hostedtoolcache/go/1.21.6/x64/src/runtime/panic.go:920 +0x270
github.com/cometbft/cometbft/p2p.(*PeerSet).Remove(0xc002aebd70, {0x16b4de8, 0xc001adb450})
	/home/runner/work/cometbft/cometbft/p2p/peer_set.go:137 +0x48a
github.com/cometbft/cometbft/p2p.(*Switch).stopAndRemovePeer(0xc001dc1c20, {0x16b4de8, 0xc001adb450}, {0x0, 0x0})
	/home/runner/work/cometbft/cometbft/p2p/switch.go:381 +0x37a
github.com/cometbft/cometbft/p2p.(*Switch).OnStop(0xc001dc1c20)
	/home/runner/work/cometbft/cometbft/p2p/switch.go:252 +0xa5
github.com/cometbft/cometbft/internal/service.(*BaseService).Stop(0xc001dc1c20)
	/home/runner/work/cometbft/cometbft/internal/service/service.go:181 +0x57b
github.com/cometbft/cometbft/internal/consensus.stopConsensusNet({0x16a9850, 0x1dbb900}, {0xc000472f00, 0x7, 0xc008265f60?}, {0xc000472f40, 0x7, 0xc007cd17a0?})
	/home/runner/work/cometbft/cometbft/internal/consensus/reactor_test.go:97 +0x28a
github.com/cometbft/cometbft/internal/consensus.TestReactorValidatorSetChanges(0x0?)
	/home/runner/work/cometbft/cometbft/internal/consensus/reactor_test.go:633 +0x1325
testing.tRunner(0xc0048f31e0, 0x154f7a8)
	/opt/hostedtoolcache/go/1.21.6/x64/src/testing/testing.go:1595 +0x262
created by testing.(*T).Run in goroutine 1
	/opt/hostedtoolcache/go/1.21.6/x64/src/testing/testing.go:1648 +0x846
FAIL	github.com/cometbft/cometbft/internal/consensus	13.454s
```

cason:
Notice that `attemptDisconnects()` is only used by nodes in seed mode. Not sure if anyone still uses this mode, as the node in this case will be super slow and problematic (due to the number of peers).

In any case, this method is completely best-effort, and it runs periodically. So a partial view of the peer set here is not a problem at al.

PR #2250, opened 2024-02-06
ci: run golangci-lint linters for all files (not only ones which were modified)

melekes:
on modified files.
See https://pre-commit.com/hooks.html

cc @greg-szabo


melekes:
> If we expect the full linting to pass, shouldn't we request that in pre-commit too?

Short answer: I don't know. As of now, pre-commit only runs **fast linters** (slow linters are skipped)

```
golangci-lint - Fast linters runner for Go. Note that only modified files are linted, so linters like 'unused' that need to scan all files won't work as expected.
golangci-lint-full - Fast linters runner for Go. Runs on all files in the repo. Use this hook if you use pre-commit in CI.
```

**I.e. it's not enough to run golangci-lint for all files, we also want to run both fast & slow linters**

> Do we expect the full linting to uncover issues that the user caused?

Yes

> Is the user required to fix linting errors uncovered by the CI, even if it wasn't detected locally?

Yes

> Shall we request a full linting in pre-commit? (The pre-commit command has an --all-files option and we can also set the full golangci-lint parameters in the .pre-commit-config.yml.)

I would say no because it would be slow (this is my hunch, but let's verify) and break the developer process. I.e. if he/she commits often, it will be annoying to wait for full lint to finish every time.

sergio-mena:
> I.e. if he/she commits often, it will be annoying to wait for full lint to finish every time.

+ 1

greg-szabo:
Running linting locally shouldn't take too much time, at least in my experience.

I'm also a bit perplexed with the type of linters and their configuration:

I need to find where slow/fast linters are configured. The [`golangci-lint` config](https://github.com/golangci/golangci-lint/blob/master/.pre-commit-hooks.yaml) only makes a slight difference between the two ways of running: using `--new-from-rev HEAD ` or not. The configuration should come from our repository, which is preconfigured for a specific set of linters.

I would really like it if the developer had a way to run these checks locally without parsing what's in the GitHub Actions, but maybe that's only my preference. The only case where this becomes problematic is security vulnerability fixes that are created on a private branch. Those branches can't run CI for obvious reasons and the only linting (if it ever happens) is on the developer machine.

greg-szabo (approved):

melekes:
> Running linting locally shouldn't take too much time, at least in my experience.

I just checked it and it doesn't add too much difference. Changed pre-commit job to `golangci-lint-full`.

mzabaluev (approved):

mempool/clist_mempool_test.go
```diff
@@ -289,7 +289,7 @@ func TestMempoolUpdateDoesNotPanicWhenApplicationMissedTx(t *testing.T) {
 		reqRes := newReqRes(tx, abci.CodeTypeOK, abci.CHECK_TX_TYPE_CHECK)
 		callback(reqRes.Request, reqRes.Response)
 	}
-	require.Equal(t, len(txs), mp.Size())
```
greg-szabo, line 292: Is that the result of some linting? Neat.

greg-szabo (approved):

mempool/clist_mempool_test.go
melekes: Yep

PR #2259, opened 2024-02-07
perf(mempool/cache): clear map in Reset

### LINKED ISSUE #2243: mempool: LRUTxCache.Reset could benefit from using the map clearing idiom to reduce CPU & RAM pressure
If we look at this code https://github.com/cometbft/cometbft/blob/f92bace913312bbca2ca7e14ca2dcf35eec227e3/mempool/cache.go#L56-L62

we can see that .Reset simply takes a lock then allocates a fresh map by discarding the previous one.

## Suggestion
We can reduce on CPU & RAM pressure by using the map clearing idiom aka
```go
func (c *LRUTxCache) Reset() { 
 	c.mtx.Lock() 
 	defer c.mtx.Unlock() 
  
        for key := range c.cacheMap {
             delete(c.cacheMap, key)
 	}
 	c.list.Init() 
 } 
```

which in performance wisdom is a known and accepted Go idiom for reducing CPU & RAM pressure by reusing the map.

melekes:
instead of allocating a new one
Closes #2243

It should not be backported onto <= v0.37 because their min Go version is 1.20, and `clear` was introduced in Go 1.21.

---

#### PR checklist

- [ ] Tests written/updated


hvanz (approved):
LGTM

odeke-em:
Thanks for this change @melekes!

PR #2263, opened 2024-02-07
fix(mempool/tests): Reduce tests duration

hvanz:
- Do not print the whole transaction in log messages. In particular, `TestMempool_CheckTxChecksTxSize` sometimes timeouts because it logs the whole transaction, which can be huge in this test.
- Decrease the timeouts of some tests that are unnecessarily high.
- Remove verbose or redundant debug log messages.

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

cason:
Jumping late here, I think that the representation (in logs, in RPC) of transactions (e.g., `String()` vs `Hash()`) is problematic. This topic should be discussed, at least I suggest to, in #2241.

Issue #1827, opened 2023-12-14
mempool: Data race when rechecking with socket connection


hvanz:
### What happened?

There is a data race to access the variable `recheckCursor` in `CListMempool` on a particular recheck scenario when connected to the app via sockets.

This is the scenario:
- At the end of a mempool `Update`, all txs in the mempool are sent to the app for rechecking their validity. So, for each tx one CheckTx request is sent to the app. Each request is sent in the order in which the txs appear in the mempool list.
- Because we are using a socket connection, CheckTx responses may arrive in a different order in which the requests were sent.
- If a tx that is being rechecked is received again for validation, it may occur a data race when accessing `recheckCursor`, which is used to traverse `clist` in order.

Related to https://github.com/tendermint/tendermint/issues/5519.

### How to reproduce it

Running the following test should produce a data race. This test was originally written for #1116, which includes a potential fix.
```
func TestMempoolRecheckRace(t *testing.T) {
	sockPath := fmt.Sprintf("unix:///tmp/echo_%v.sock", cmtrand.Str(6))
	app := kvstore.NewInMemoryApplication()
	_, server := newRemoteApp(t, sockPath, app)
	t.Cleanup(func() {
		if err := server.Stop(); err != nil {
			t.Error(err)
		}
	})
	cfg := test.ResetTestRoot("mempool_test")
	cfg.Mempool.Size = 1000
	cfg.Mempool.CacheSize = 1000
	mp, cleanup := newMempoolWithAppAndConfig(proxy.NewRemoteClientCreator(sockPath, "socket", true), cfg)
	defer cleanup()

	// Add a bunch of transactions to the mempool.
	var err error
	txs := newUniqueTxs(100)
	for _, tx := range txs {
		err = mp.CheckTx(tx, nil, TxInfo{})
		require.NoError(t, err)
	}
	err = mp.FlushAppConn()
	require.NoError(t, err)

	// Update one transaction to force rechecking the rest.
	mp.Lock()
	err = mp.Update(1, txs[:1], abciResponses(1, abci.CodeTypeOK), nil, nil)
	mp.Unlock()
	require.NoError(t, err)

	// Add again the same transaction that was updated. Checking this
	// transaction and rechecking the others simultaneously should not result
	// in a data race for the variable recheckCursor.
	err = mp.CheckTx(txs[:1][0], nil, TxInfo{})
	require.Equal(t, err, ErrTxInCache)
}
```

cason:
[...]

PR #2268, opened 2024-02-08
fix(mempool): Fix data race when rechecking with async ABCI client

### LINKED ISSUE #2225: mempool: panic when app's CheckTx returns an error while rechecking

hvanz:
Fixes ~~#2225 and~~ #1827 

(#2225 is now fixed in a separate PR, #2894)

The bug: during rechecking, when the `CheckTxAsync` request for the last transaction fails, then the `resCbRecheck` callback on the response is not called, and the recheck variables end up in a wrong state (`recheckCursor != nil`, meaning that recheck has not finished). This will cause a panic next time a new transaction arrives, and the `CheckTx` response finds that rechecking hasn't finished.

This problem only happens when using the non-local ABCI client, where `CheckTx` responses may arrive late or never, so the response won't be processed by the callback. We have two options to fix this.
1. When we call `CheckTxAsync`, block waiting for a response. If the response never arrives, it will block `Update` forever.
2. After sending all recheck requests, we flush the app connection and set a timer to wait for late recheck responses. After the timer expires, we finalise rechecking properly. If a CheckTx response arrives late, we consider that it is safe to ignore it.

This PR implements option 2, as we cannot allow the risk to block the node forever waiting for a response.

With the proposed changes, now when we reach the end of the rechecking process, all requests and responses will be processed or discared, and `recheckCursor` will always be `nil`.

This PR also:
- refactors all recheck logic to put it into a separate `recheck` struct. The fix to the bug described above is the only change in the recheck logic.
- adds 4 new tests.

---

#### PR checklist

- [x] Tests written/updated


faddat:
thank you

mempool/clist_mempool_test.go
```diff
@@ -271,14 +271,15 @@ func TestMempoolUpdateDoesNotPanicWhenApplicationMissedTx(t *testing.T) {
 	mockClient.On("SetLogger", mock.Anything)
 	mockClient.On("Error").Return(nil).Times(4)
 	mockClient.On("SetResponseCallback", mock.MatchedBy(func(cb abciclient.Callback) bool { callback = cb; return true }))
+	mockClient.On("CheckTxAsync", mock.Anything, mock.Anything).Return(nil, nil)
+	mockClient.On("Flush", mock.Anything).Return(nil)
```
cason, line 275: This test passes in `main`. Curious on why we have to mock `Flush` here.

mempool/clist_mempool_test.go
```diff
@@ -805,6 +785,211 @@ func TestMempoolNotifyTxsAvailable(t *testing.T) {
...
+	require.Nil(t, mp.recheck.cursor)
+	require.Nil(t, mp.recheck.end)
+
+	// mp.recheck.done() should be true only before and after calling recheckTxs.
+	mp.recheckTxs()
+	require.True(t, mp.recheck.done())
+	require.Nil(t, mp.recheck.cursor)
+	require.NotNil(t, mp.recheck.end)
+	require.Equal(t, mp.recheck.end, mp.txs.Back())
+	require.Equal(t, len(txs)-1, mp.Size()) // one invalid tx was removed
```
cason, line 847: This fails with the `main` implementation, the fix solve this, good.

mempool/clist_mempool_test.go
```diff
@@ -805,6 +785,211 @@ func TestMempoolNotifyTxsAvailable(t *testing.T) {
...
+		callback(reqRes2.Request, reqRes2.Response)
+	}).Return(nil)
+
+	// mp.recheck.done() should be true only before and after calling recheckTxs.
+	mp.recheckTxs()
+	require.True(t, mp.recheck.done())
+	require.Nil(t, mp.recheck.cursor)
+	require.NotNil(t, mp.recheck.end)
+	require.Equal(t, mp.recheck.end, mp.txs.Back())
+	require.Equal(t, len(txs)-1, mp.Size()) // one invalid tx was removed
```
cason, line 908: This fails with the `main` implementation, the fix solve this, good.

mempool/clist_mempool_test.go
```diff
@@ -805,6 +785,211 @@ func TestMempoolNotifyTxsAvailable(t *testing.T) {
...
+	require.Len(t, txs, mp.Size())
+
+	// Check that recheck has not started yet.
+	require.True(t, mp.recheck.done())
+	require.Nil(t, mp.recheck.cursor)
+	require.Nil(t, mp.recheck.end)
+
+	// The first tx is valid, the third is invalid, and the request for the second and fourth tx
+	// were not sent, so the callback is not called.
+	mockClient.On("Flush", mock.Anything).Run(func(_ mock.Arguments) {
```
cason, line 895: Why we need to mock the flush? I why it has to return content?

cason: This code is not really easy to understand.

mempool/clist_mempool_test.go
```diff
@@ -805,6 +785,211 @@ func TestMempoolNotifyTxsAvailable(t *testing.T) {
...
+	require.NotNil(t, mp.recheck.end)
+	require.Equal(t, mp.recheck.end, mp.txs.Back())
+	require.Equal(t, len(txs)-1, mp.Size()) // one invalid tx was removed
+	require.Equal(t, int32(2), mp.recheck.numPendingTxs.Load())
+
+	mockClient.AssertExpectations(t)
+}
+
+// This test used to cause a data race when rechecking (see https://github.com/cometbft/cometbft/issues/1827).
+func TestMempoolRecheckRace(t *testing.T) {
```
cason, line 915: This test passes with the original code (in `main`).

mempool/clist_mempool_test.go
```diff
@@ -805,6 +785,211 @@ func TestMempoolNotifyTxsAvailable(t *testing.T) {
...
+	// transaction and rechecking the others simultaneously should not result
+	// in a data race on the variable recheck.cursor.
+	_, err = mp.CheckTx(txs[:1][0])
+	require.Equal(t, err, ErrTxInCache)
+	require.Zero(t, mp.recheck.numPendingTxs.Load())
+}
+
+// Test adding txs while a concurrent routine reaps txs and updates the mempool, using an async ABCI
+// client.
+func TestMempoolConcurrentCheckTxAndUpdate(t *testing.T) {
```
cason, line 945: This test also passes with the original code (in `main`).

mempool/clist_mempool.go
```diff
@@ -45,11 +50,8 @@ type CListMempool struct {
 
 	proxyAppConn proxy.AppConnMempool
 
-	// Track whether we're rechecking txs.
-	// These are not protected by a mutex and are expected to be mutated in
-	// serial (ie. by abci responses which are called in serial).
-	recheckCursor *clist.CElement // next expected response
-	recheckEnd    *clist.CElement // re-checking stops here
+	// For keeping track of the rechecking process.
```
cason, line 53:
```suggestion
	// Keeps track of the rechecking process.
```

mempool/clist_mempool.go
```diff
@@ -306,14 +307,16 @@ func (mem *CListMempool) globalCb(req *abci.Request, res *abci.Response) {
...
-			if mem.recheckCursor != nil {
+			if !mem.recheck.done() {
 				// this should never happen
 				panic("recheck cursor is not nil before resCbFirstTime")
 			}
 			mem.resCbFirstTime(req.GetCheckTx().Tx, res.GetCheckTx())
 
 		case abci.CHECK_TX_TYPE_RECHECK:
-			if mem.recheckCursor == nil {
+			tx := types.Tx(req.GetCheckTx().Tx)
```
cason, line 317: This is only used in case of error. We can move it to inside the next if clause.

mempool/clist_mempool.go
```diff
@@ -649,29 +610,139 @@ func (mem *CListMempool) Update(
 	return nil
 }
 
+// recheckTxs sends all transactions in the mempool to the app for re-validation. When the function
+// returns, all recheck responses from the app have been processed.
```
cason, line 614: This is not about this PR, but blocking here is terrible. Why we need to conclude the recheck before move on? And, if this is indeed the case, why we use `CheckTxAsync`, expected to be... asynchronous.

mempool/clist_mempool.go
```diff
@@ -649,29 +610,139 @@ func (mem *CListMempool) Update(
...
-	// In <v0.37 we would call FlushAsync at the end of recheckTx forcing the buffer to flush
-	// all pending messages to the app. There doesn't seem to be any need here as the buffer
-	// will get flushed regularly or when filled.
+	// Flush any pending asynchronous recheck requests to process.
+	mem.proxyAppConn.Flush(context.TODO())
+
+	// Give some time to finish processing the responses; then finish rechecking, even if not all
+	// txs were rechecked.
+	select {
+	case <-time.After(timeoutRecheck):
```
cason, line 647: Why? We are now imposing the the application a maximum delay for performing operations?

mempool/clist_mempool.go
```diff
@@ -18,6 +19,10 @@ import (
 	"github.com/cometbft/cometbft/types"
 )
 
+const (
+	timeoutRecheck = 100 * time.Millisecond
```
cason, line 23: This appears to be pretty arbitrary.

cason (commented):
Left some comments, some are just me thinking aloud.

I am not sure that forcing a maximum delay for ReCheckTx to return is the right approach here. Is is possible to come up with a different (asynchronous) solution?

mempool/clist_mempool_test.go
cason: I think it should break there, and pass here.

mempool/clist_mempool.go
hvanz: Done.

mempool/clist_mempool.go
hvanz: For async ABCI clients, where CheckTx responses may arrive late or never, we have two options. 
- One is to block waiting for CheckTx responses. In particular during rechecking, which happens inside Update, where the updateMtx lock is held. If the response never arrives, Update will block forever.
- The second option is to put a timeout, which is what I did here. It's not critical if some response arrive late and is not processed. IMO this option is better than blocking.

mempool/clist_mempool.go
hvanz: It's not blocking, that's why there is a timeout, so that rechecking can finish even when not all the CheckTx responses have arrived.

Remember that recheck happens after the mempool has been updated. And once rechecking finishes then the `updateMtx` lock is released and adding new transactions via `CListMempool.CheckTx` is possible again.

mempool/clist_mempool_test.go
hvanz: On the socket client, CheckTxAsync does not send the request, it just adds it to a queue. It's the Flush method that actually sends it and invokes the callback.

sergio-mena:
@hvanz ready for review?

mempool/clist_mempool_test.go
hvanz: `Flush` is called when sending the requests to the app, so if it's not mocked the mock client will complain.

config/toml.go
```diff
@@ -397,6 +397,11 @@ type = "flood"
 # you can disable rechecking.
 recheck = {{ .Mempool.Recheck }}
 
+# RecheckTimeout is the time the application has to return CheckTx responses after the
```
melekes, line 400: Maybe we should write a sentence explaining how an APP developer is supposed to measure / tweak this parameter? Or the default 3s is good for the majority of the existing applications?

mempool/clist_mempool.go
```diff
@@ -303,14 +300,14 @@ func (mem *CListMempool) globalCb(req *abci.Request, res *abci.Response) {
...
-				panic("recheck cursor is not nil before resCbFirstTime")
+			if !mem.recheck.done() {
+				panic(log.NewLazySprintf("rechecking has not finished; cannot check new tx %v", types.Tx(req.GetCheckTx().Tx).Hash()))
 			}
 			mem.resCbFirstTime(req.GetCheckTx().Tx, res.GetCheckTx())
 
 		case abci.CHECK_TX_TYPE_RECHECK:
-			if mem.recheckCursor == nil {
+			if mem.recheck.done() {
+				mem.logger.Info("rechecking has finished; discard late recheck response", "tx", log.NewLazySprintf("%v", types.Tx(req.GetCheckTx().Tx).Hash()))
```
melekes, line 310: Error?

mempool/clist_mempool.go
```diff
@@ -646,28 +601,138 @@ func (mem *CListMempool) Update(
...
 
-	// In <v0.37 we would call FlushAsync at the end of recheckTx forcing the buffer to flush
-	// all pending messages to the app. There doesn't seem to be any need here as the buffer
-	// will get flushed regularly or when filled.
+	// Flush any pending asynchronous recheck requests to process.
+	mem.proxyAppConn.Flush(context.TODO())
+
+	// Give some time to finish processing the responses; then finish the rechecking process, even
+	// if not all txs were rechecked.
+	select {
```
melekes, line 636: how is it non-blocking (https://github.com/cometbft/cometbft/pull/2268/files#r1579726481) if we're blocking here waiting for timeout or `doneRechecking`? Should'nt we run this in a goroutine if we want this non-blocking?

mempool/clist_mempool.go
```diff
@@ -646,28 +601,138 @@ func (mem *CListMempool) Update(
...
+// tryFinish will check if the cursor is at the end of the list and notify the channel that
+// rechecking has finished. It returns true iff it's done rechecking.
+func (rc *recheck) tryFinish() bool {
+	if rc.cursor == rc.end {
+		// Reached end of the list without finding a matching tx.
+		rc.setDone()
+	}
+	if rc.done() {
+		// Notify that recheck has finished.
+		go func() {
```
melekes, line 698: why you need a goroutine?

mempool/clist_mempool.go
```diff
@@ -646,28 +601,138 @@ func (mem *CListMempool) Update(
...
+// end of the list, so it can finish rechecking.
+//
+// The goal is to guarantee that transactions are rechecked in the order in which they are in the
+// mempool. Transactions whose recheck response arrive late or don't arrive at all are skipped and
+// not rechecked.
+func (rc *recheck) findNextEntryMatching(tx *types.Tx) bool {
+	found := false
+	for ; !rc.done(); rc.setNextEntry() {
+		expectedTx := rc.cursor.Value.(*mempoolTx).tx
+		if bytes.Equal(*tx, expectedTx) {
```
melekes, line 720: can we use https://github.com/bits-and-blooms/bloom to speed up the search?

melekes (commented):
👍

mempool/clist_mempool.go
hvanz: Sorry, ignore above when I say it's non-blocking. Rechecking needs to block waiting for all reCheckTx responses because it needs to finish before new transactions are allowed to be checked again. This is to not break the sequential order when rechecking transactions.

mempool/clist_mempool.go
hvanz: Right, there's no need for a goroutine. The select is non-blocking. I removed it.

mempool/clist_mempool.go
hvanz: Does this library allow removing elements from the set?

hvanz: Probably it could work. The set of elements we would need to keep in the Bloom filter doesn't change; it's just those elements left in the CList that are being rechecked. So we would create a new Bloom filter at the beginning of  `recheckTxs` and remove it when the rechecking process has finished.

hvanz: By the way, this optimization is independent of this PR; it could also be applied to the current code.

melekes (approved):

mempool/clist_mempool.go
sergio-mena: Bear in mind bloom filters don't guarantee 100% accuracy when checking for inclusion

config/config.go
```diff
@@ -904,10 +908,11 @@ type MempoolConfig struct {
 // DefaultMempoolConfig returns a default configuration for the CometBFT mempool.
 func DefaultMempoolConfig() *MempoolConfig {
 	return &MempoolConfig{
-		Type:      MempoolTypeFlood,
-		Recheck:   true,
-		Broadcast: true,
-		WalPath:   "",
+		Type:           MempoolTypeFlood,
+		Recheck:        true,
+		RecheckTimeout: 300 * time.Millisecond,
```
sergio-mena, line 913: This might be too little for heavy app implementations for `CheckTx`

docs/references/config/config.toml.md
```diff
@@ -1205,6 +1205,20 @@ might become invalid. Setting `recheck = true` will go through the remaining tra
...
+started. Responses that arrive after the timeout expires are discarded.
+```toml
+recheck_timeout = "300ms"
+```
+
+| Value type          | string (duration) |
+|:--------------------|:------------------|
+| **Possible values** | &gt;= `"300ms"`   |
+
+This setting only applies to asynchronous ABCI clients and when `recheck` is enabled.
```
sergio-mena, line 1220:
```suggestion
This setting only applies to asynchronous ABCI methods and when `recheck` is enabled.
```

docs/references/config/config.toml.md
```diff
@@ -1205,6 +1205,20 @@ might become invalid. Setting `recheck = true` will go through the remaining tra
 If your application may remove transactions passed by CometBFT to your `PrepareProposal` handler,
 you probably want to set this configuration to `true` to avoid possible leaks in your mempool
 (transactions staying in the mempool until the node is next restarted).
+
+### mempool.recheck_timeout
+Time to wait for the application to return CheckTx responses after the rechecking process was
+started. Responses that arrive after the timeout expires are discarded.
+```toml
+recheck_timeout = "300ms"
```
sergio-mena, line 1213: As mentioned above, I find 300ms too low for advanced apps (which have to verify signatures, etc)

mempool/clist_mempool.go
sergio-mena: Bear in mind that, even if we wait here until all TXs have been re-Checked, it still makes sense that `CheckTx` is async: in non-local ABCI clients we will pipeline this:
* Request serialization, request transmission (over TCP or gRPC), Request deserialization at the app, etc
If we make `CheckTx` synchronous, all that will have to happen for one TX until we start with the next. This may make no difference with local client, but is likely to have a high performance impact over a socket or gRPC

sergio-mena: @cason In order not to have to wait here, we need to tackle one of the items in our laundry list: `Re-CheckTx`

But, as you said, that's not part of this PR

mempool/clist_mempool.go
```diff
@@ -646,28 +603,136 @@ func (mem *CListMempool) Update(
...
 
+// recheckTxs sends all transactions in the mempool to the app for re-validation. When the function
+// returns, all recheck responses from the app have been processed.
 func (mem *CListMempool) recheckTxs() {
-	if mem.Size() == 0 {
-		panic("recheckTxs is called, but the mempool is empty")
+	if mem.Size() <= 0 {
+		return
 	}
+	mem.logger.Debug("recheck txs", "height", mem.height.Load(), "num-txs", mem.Size())
```
sergio-mena, line 612: I would move this debug log to the beginning of the function

mempool/clist_mempool.go
```diff
@@ -646,28 +603,136 @@ func (mem *CListMempool) Update(
...
 
-	mem.recheckCursor = mem.txs.Front()
-	mem.recheckEnd = mem.txs.Back()
+	mem.recheck.init(mem.txs.Front(), mem.txs.Back())
```
sergio-mena, line 614: Sorry, I forgot why we need to keep track of `mem.txs.Back()`. I imagine it's to avoid re-checking newly received TXs while running `recheckTxs()`.
However, haven't we locked the mempool so that this doesn't happen?

mempool/clist_mempool.go
```diff
@@ -646,28 +603,136 @@ func (mem *CListMempool) Update(
...
+}
+
+func newRecheck() *recheck {
+	return &recheck{
+		doneCh: make(chan struct{}, 1),
+	}
+}
+
+func (rc *recheck) init(first, last *clist.CElement) {
+	rc.cursor = first
```
sergio-mena, line 671: Can you please make sure that `cursor` and `end` are `nil` when we call this function?
My understanding is that, if that is not the case, we are overlapping the Re-CheckTx of two different `Update`s

config/toml.go
hvanz: I added a paragraph to `config.toml.md`.

docs/references/config/config.toml.md
hvanz: Yes, I thought the same. I changed to 1 second. I think more than that will be too much, though it greatly depends on the application.

mempool/clist_mempool.go
hvanz: Done.

mempool/clist_mempool.go
hvanz: It's to check if `recheck.cursor` reached the end of the list, when `recheck.cursor == recheck.end`. This is part of the original recheck logic, though it's true that the `recheck.end` variable never changes and that the CList also doesn't change during rechecking because the mempool is locked.

mempool/clist_mempool.go
hvanz: Done. Now it will panic when trying to call init when a rechecking process is in place.

config/config.go
hvanz: That's right. I changed it to 1 second. I also added a paragraph to `config.toml.md` explaining how to roughly calculate this value.

mempool/clist_mempool_test.go
hvanz: This test have been very useful in debugging the code. It performs most mempool operations in the typical scenario where consensus updates the mempool while transactions are being added, concurrently and multiple times.

mempool/clist_mempool.go
hvanz: Changed to a config parameter. And the default value is now 1000ms.

mempool/clist_mempool_test.go
hvanz: I added more comments to improve clarity of the code.

docs/references/config/config.toml.md
hvanz: I changed it to "non-local ABCI clients".

PR #2284, opened 2024-02-10
feat(pbts): enable `e2e` manifest config to set `PbtsEnableHeight`

### LINKED ISSUE #2227: Add infra in `e2e` runner and generator to set the new consensus param in `e2e` (or DO) tests

sergio-mena:
Closes #2227 

Similarly to how `VoteExtensionsEnableHeight` is managed in `e2e`, this PR introduces `PbtsEnableHeight` and `PbtsUpdateHeight` to be able to configure when the consensus param is set in the `e2e` run.

It also adds support for the generator.

And I added a test case that checks that the block times according to the algorithm used at that height (PBTS/BFT Time)

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/test/e2e/tests/block_test.go
+++ b/test/e2e/tests/block_test.go
@@ -6,7 +6,6 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	"github.com/cometbft/cometbft/internal/state"
 	e2e "github.com/cometbft/cometbft/test/e2e/pkg"
 )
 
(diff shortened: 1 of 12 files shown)


test/e2e/tests/block_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/cometbft/cometbft/internal/state"
```
sergio-mena, line 9: This (e2e tests and runner) can be considered external to CometBFT, so I don't think it's best to access "internal" from here. I'd propose to move `MedianTime` elsewhere in CometBFT (out of "internal"). What do you think?

cason: Totally. It could be moved to `types/block.go` or `types/validation.go`. The latter contains "abstract" methods receiving a `Commit` and a `ValidatorSet` as well.

cason (approved):
Legit!

test/e2e/tests/block_test.go
sergio-mena: Thanks @cason for the input!
I moved it to `types/block.go`, and made it a method of `Commit`. Please take a look at the last commit

cason: In terms of readability, your solution is great :)

lasarojc (approved):

PR #2294, opened 2024-02-11
refactor(internal/statesync): replace ticker with `time.After(X)`

hoanguyenkh:
- Remove `defer` from `ticker.Stop()` inside the loop.
- Manually call `ticker.Stop()` at necessary points in the loop to ensure tickers are stopped correctly after each iteration.


faddat:
curious: how'd you identify this?

cason:
This code doesn't make sense in general. We create a `ticker`, that is something that produces ticks every `X` time. But we only use it once. The right solution here would to use `time.After(X)`, which has some performance/GC issues...

melekes:
> The right solution here would to use time.After(X)

@hoanguyenkh would you mind rewriting the code to use `time.After(s.retryTimeout)`?

hoanguyenkh:
> > The right solution here would to use time.After(X)
> 
> @hoanguyenkh would you mind rewriting the code to use `time.After(s.retryTimeout)`?

thanks @melekes, @cason i rewrite this code

melekes (approved):
Thanks @hoanguyenkh

PR #2296, opened 2024-02-11
chore: updated doc link

Pan-chao:
Current link is broken, thought of fixing it myself instead of opening an Issue, less time wasted.


hvanz (approved):
Thanks @Pan-chao!

PR #2316, opened 2024-02-12
spec(consensus): update Block Time documentation in spec

### LINKED ISSUE #2020: PBTS: update Block Time documentation in Spec
Originally https://github.com/tendermint/tendermint/issues/8591

With the introduction of PBTS, the references to timestamp generation in the remainder of the spec should be updated. At the moment they refer to the [BFT Time](https://github.com/cometbft/cometbft/blob/main/spec/consensus/bft-time.md) algorithm.

Differently from the original approach, for retrocompatibility, we are not fully dropping BFT Time. The documentation of timestamps generation for blocks should them reflect this scenario, referring to the two methods adopted, possibly tagging BFT Time as a legacy method.

Definition of Done:
- [x] Create a general spec for block times: `spec/consensus/time.md`
- [x] Add to it a general description of block times (e.g., monotonicity?)
- [x] Add to it pointers to both BFT Time and PBTS specifications
- [x] Update pointers to BFT Time spec to the new Time spec

cason:
Closes #2020.

- Create a general spec for block times: `spec/consensus/time.md` ([rendered](https://github.com/cometbft/cometbft/blob/cason/2020-time-spec/spec/consensus/time.md))
- Add to it a general description of block times properties
- Add to it pointers to both BFT Time and PBTS specifications
- Refactors the description of BFT Time

---

#### PR checklist

- [ ] Tests written/updated


andynog (approved):
Overall looks good to me. But the `BFT Time` spec I haven't truly validated because I'm not sure if there will be changes in PBTS that will affect the implementation of the old BFT Time. If there are not changes then it's all good.

spec/core/data_structures.md
```diff
@@ -182,8 +182,9 @@ Part defines a part of a block. In CometBFT blocks are broken into `parts` for g
 
 ## Time
 
-CometBFT uses the [Google.Protobuf.Timestamp](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Timestamp)
+CometBFT uses the [Google.Protobuf.Timestamp](https://protobuf.dev/reference/protobuf/google.protobuf/#timestamp)
```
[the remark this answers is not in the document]
mzabaluev: The gogoproto decoration is indeed an implementation detail of the GometBFT Go implementation and as such, should be taken out of the proto spec if possible. It has no bearing on the wire protocol, though.

In terms of valid values, both types should be interchangeable within the timespan of our civilization. `time.Time` in Go provides precision up to the nanosecond, so it should be interoperable with any other representations. It adds the time zone baggage, so there is an increased potential for programmer mistakes in processing the timestamps. But this is outside the scope of this discussion on specification.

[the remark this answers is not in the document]
cason: Wow, this is something to detail, no questions about.

Can you suggest a new wording summarizing this information? I can also work on that later.

spec/consensus/readme.md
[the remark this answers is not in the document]
cason: Yep, this is need to be included, thanks for catching it up :)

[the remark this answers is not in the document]
cason: Updated this section. But this spec in general is very outdated, we should consider reviewing it, or even removing it, since most of the information is already present in other pages (e.g., the data structures spec).

spec/core/data_structures.md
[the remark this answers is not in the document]
cason: Spec-wise, I think the relevant part here is whether this affect implementations in other languages.

I can mention that a specific sub-type of this proto is used. I think the more detail here, the better.

spec/consensus/readme.md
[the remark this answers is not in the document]
cason: https://github.com/cometbft/cometbft/issues/2348

spec/consensus/signing.md
```diff
@@ -30,14 +30,30 @@ All signed messages must correspond to one of these types.
...
+- [BFT Time](./bft-time.md): the `Timestamp` field of `Precommit` vote messages
+  are used to compute the `Time` for the next proposed block.
+  Correct validators are expected to report their local clock time, provided
+  that the time is higher than the current block's time.
+  Otherwise, the reported time is the current block's time plus 1ms.
+
+- [PBTS](./proposer-based-timestamp/README.md): the `Timestamp` field of a
+  `Proposal` message must match the proposed `Block.Time`.
+   Otherwise, the `Proposal` will be rejected by correct validators.
+   There are no requirements for `Vote.Timestamp` values.
```
andynog, line 55: I'm just wondering why there's not need to validate the vote timestamp, shouldn't there be some validation that the timestamp of the vote be within some bounds of the timestamp of the proposal ?

Or is that not relevant because if the validator clock is drifted too much it will consider the proposal invalid and won't sign it ? Does PBTS changes the chances of this happening, or not really because you in the timestamp validation it still relies on local clock?

Anyway, was just trying to understand if there was a possibility to have votes with drifted timestamps as part of the vote set.

Ideally, the timestamp of a vote should closely align with the timestamp of the proposal to ensure that the voting reflects a synchronized view of time across the network.

andynog (approved):
lgtm, added a few comments but some might not be relevant, feel free to ignore.

spec/consensus/bft-time.md
```diff
@@ -3,54 +3,123 @@ order: 2
...
+from the `Timestamp` field of the `Precommit` messages broadcast by
+validators in the commit round of the previous height `H-1`.
+
+In order to commit a block, a node needs to receive `Precommit` messages for
+the corresponding `BlockID` from validators whose cumulative voting power is
+more than `2/3` of the total voting power.
+The received `Precommit` messages should refer to the same round, the _commit round_.
+A set of `Precommit` messages with the properties above mentioned is a `Commit`.
+A `Commit` set of height `H` is included in blocks proposed in height `H+1`.
+
```
lasarojc, line 26:
```suggestion
In order to commit a block, a node needs to receive `Precommit` messages for
the corresponding `BlockID` from validators whose cumulative voting power is
more than `2/3` of the total voting power.
The received `Precommit` messages should refer to the same round, the _commit round_.
A set of `Precommit` messages with the properties above mentioned is a `Commit`.
A `Commit` set of height `H` is included in blocks proposed in height `H+1`.

BFT Time computes the `Time` of a block proposed in height `H` of consensus
from the `Timestamp` field of the messages in the previous `Commit`, that is is, of height `H-1`.

```

spec/consensus/bft-time.md
```diff
@@ -3,54 +3,123 @@ order: 2
...
+The `Time` field of a block proposed in height `H` is deterministically
+computed from the `LastCommit` field of the block, which is a `Commit` set from
+height `H-1`, using the `MedianTime` method defined as follows:
```
lasarojc, line 29:
```suggestion
In order to commit a block, a node needs to receive `Precommit` messages for
the corresponding `BlockID` from validators whose cumulative voting power is
more than `2/3` of the total voting power.
The received `Precommit` messages should refer to the same round, the _commit round_.
A set of `Precommit` messages with the properties above mentioned is a `Commit`.
A `Commit` set of height `H` is included in blocks proposed in height `H+1`.

BFT Time computes the `Time` field of a block proposed in height `H` deterministically
from the `LastCommit` field of the block, which is a `Commit` set from
height `H-1`, using the `MedianTime` method defined as follows:
```

spec/consensus/bft-time.md
```diff
@@ -3,54 +3,123 @@ order: 2
...
+
+- `MedianTime`: the median time of a `Commit` is equal to the median of the
+  `Timestamp` fields of the `Precommit` messages forming the `Commit`, where the value of
+   each `Timestamp` field is counted a number of times proportional to the voting power
+   of the validator that produced and signed that `Precommit` message.
+
+In other words, the `MedianTime` is the weighted median of `Timestamp` fields, with
+heights defined by validators' voting powers.
```
lasarojc, line 37:
```suggestion
- `MedianTime`: the weighted median of `Timestamp` fields, of the previous block with
 heights defined by validators' voting powers or, in other words, the median of the
  `Timestamp` fields of the `Precommit` messages forming a `Commit`, where the value of
   each `Timestamp` field is counted a number of times proportional to the voting power
   of the validator that produced and signed that `Precommit` message.
```

spec/consensus/bft-time.md
```diff
@@ -3,54 +3,123 @@ order: 2
...
+Notice that, no matter what set of `Precommit` messages with at least `2/3` of
+the total voting power we choose, the `MedianTime` value will always be a
+value among the `Timestamp` values produced by correct validators.
+
+## Operation
+
+In order to implement BFT Time, validators need to set the `Timestamp` field of
+`Precommit` messages they sign and broadcast, and block proposers need to
+compute the block `Time` from the `LastCommit` set.
+
```
lasarojc, line 69:
```suggestion
```

spec/consensus/bft-time.md
```diff
@@ -3,54 +3,123 @@ order: 2
...
+1. Let `now` be the clock time of the validator.
+2. If `LockedBlock` is defined, set `Timestamp = max(now, LockedBlock.Time + 1ms)`.
+3. Else if `ProposalBlock` is defined, set `Timestamp = max(now, ProposalBlock.Time + 1ms)`.
+4. Otherwise, set `Timestamp = now`.
+
+The rationale is that, for ensuring [Time Monotonicity](./time.md#properties),
+the `Time` of the next block should be higher than the time of the current block.
+The `LockedBlock`, if set, is the block for which the validator is issuing a `Precommit`.
+The `ProposalBlock` is the block proposed in that round; in favorable runs, it
+matches the `LockedBlock`.
```
lasarojc, line 83:
```suggestion
The `LockedBlock`, if set, is the block for which the validator is issuing a `Precommit`.
The `ProposalBlock` is the block proposed in that round; in favorable runs, it
matches the `LockedBlock`.
To ensure [Time Monotonicity](./time.md#properties), the `Time` of the next block should be 
higher than the time of the current block, hence the `+1ms` in the computation.

```

spec/consensus/bft-time.md
```diff
@@ -3,54 +3,123 @@ order: 2
...
+
+The validator in practice _proposes_ the `Time` for the next block when setting
+the `Timestamp` of its `Precommit`.
+The proposed `Time` is, by default, the validator's current clock time.
+The validator, however, cannot propose a `Time` which its smaller than the
+`Time` of the block to be committed in the current height.
+So in this case, the validator proposes the `Time` of the block to be committed
+plus a small delta, set to `1ms`.
```
lasarojc, line 91:
```suggestion
To ensure [Time Monotonicity](./time.md#properties), the `Time` of the next block should be 
higher than the `Time` of the block to be committed in the current height.
So if `now` is smaller than `Time`, the validator proposes the `Time` of the block to be committed
plus a small delta, set to `1ms`.
```

spec/consensus/signing.md
```diff
@@ -30,14 +30,30 @@ All signed messages must correspond to one of these types.
...
 Timestamps are expected to be strictly monotonic for a given validator, though
-this is not currently enforced.
-
+this is not enforced.
+
+Some timestamps, however, are used by the algorithms adopted for computing
+[block times](./time.md):
+
+- [BFT Time](./bft-time.md): the `Timestamp` field of `Precommit` vote messages
+  are used to compute the `Time` for the next proposed block.
```
lasarojc, line 47:
```suggestion
  is used to compute the `Time` for the next proposed block.
```

spec/consensus/signing.md
lasarojc: > Or is that not relevant because if the validator clock is drifted too much it will consider the proposal invalid and won't sign it ?

Good observation. Voting for a proposal implies that the validator's clocks were synchronized within accepted bounds and therefore the timestamp field of the vote becomes useless and could even be removed at some point.

spec/consensus/time.md
```diff
@@ -0,0 +1,52 @@
...
+
+- **Byzantine Fault Tolerance**: faulty or malicious processes should not be able
+  to arbitrarily influence (increase or decrease) the block Time.
+  In other words, the Time of blocks should be defined by correct processes.
+
+In addition, the Time produced by CometBFT is expected, by external observers, to provide:
+
+- **Relation to real time**: block times bear some resemblance to real time.
+  In other words, block times should represent, within some reasonable accuracy,
+  the actual time at which blocks were produced.
```
lasarojc, line 27: This definition is problematic but I have no better one to offer.

spec/consensus/time.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Implementations
+
+CometBFT implements two algorithms for computing block times:
+
+- [BFT Time][bft-time]: the algorithm adopted in versions up to `v0.38.x`;
+  available, in legacy mode, in version `v1.x`.
+
+- [Proposer-Based Timestamps][pbts-spec]: introduced in version `v1.x`,
+  intended to be a replacement for BFT Time.
```
lasarojc, line 37:
```suggestion
- [Proposer-Based Timestamps][pbts-spec]: introduced in version `v1.x`,
  as a replacement for BFT Time.
```

lasarojc (approved):
some reorganization suggestions, to reduce redundancy, but LGTM

spec/consensus/proposer-based-timestamp/README.md
andynog, line 1: Can we remove the section 
```
<!---
## BFT Time
...
-->
```

This might render in the documentation site

andynog (commented):
added one comment about a hidden section of BFT Time in the PBTS main doc

spec/consensus/time.md
cason: Indeed...

spec/consensus/bft-time.md
```diff
@@ -3,54 +3,120 @@ order: 2
...
-
-    - otherwise, `vote.Time = time.Now())`. In this case vote is for `nil` so it is not taken into account for
-    the timestamp of the next block.
+BFT Time is a Byzantine fault-tolerant algorithm for computing [block times](./time.md).
+
+> :warning:
+> CometBFT `v1.x` introduced [Proposer-Based Timestamps (PBTS)][pbts-spec],
+> intended to be a replacement for BFT Time.
+> Users are strongly encouraged to adopt PBTS in new chains, or when upgrading
+> existing chains.
```
sergio-mena, line 12: I'd add something like:

_BFT-Time MAY be deprecated in a future version of CometBFT_

... even if we know it's not likely to happen, as it would be a mess for blocksync

spec/consensus/bft-time.md
```diff
@@ -3,54 +3,120 @@ order: 2
...
+A `Commit` set of height `H` is included in blocks proposed in height `H+1`.
+
+BFT Time computes the `Time` field of a block proposed in height `H` deterministically
+from the `LastCommit` field of the block, which is a `Commit` set from
+height `H-1`, using the `MedianTime` method defined as follows:
+
+- `MedianTime`: the weighted median of `Timestamp` fields of `Precommit`
+  messages forming a `Commit` set, with weights defined by the validators' voting powers.
+  The weighted median is produced by considering the value of each `Timestamp`
+value a number of times proportional to the voting power of the corresponding validator.
```
sergio-mena, line 34:
```suggestion
  value a number of times proportional to the voting power of the corresponding validator.
```

spec/consensus/bft-time.md
```diff
@@ -3,54 +3,120 @@ order: 2
...
+
+The median of a set of values is one of the values of the set, so that the
```
sergio-mena, line 36:
```suggestion
The median of a set of values is one of the values of the set, so the
```

spec/consensus/bft-time.md
```diff
@@ -3,54 +3,120 @@ order: 2
...
+messages included in the `LastCommit` set of that block.
+
+### Example
+
+Consider the following example:
+
+- We have four validators p1, p2, p3 and p4, with voting power
+  distribution: (p1, 23), (p2, 27), (p3, 10), (p4, 10).
+  The total voting power is 70, so we assume that the faulty validators have at
+most 23 of voting power (since `N = 3F + 1`, where `N` is the total voting
```
sergio-mena, line 47:
```suggestion
most 23 of voting power (since `N < 3F`, where `N` is the total voting
```

spec/consensus/bft-time.md
```diff
@@ -3,54 +3,120 @@ order: 2
...
+power and `F` is the maximum voting power of faulty validators).
```
sergio-mena, line 48:
```suggestion
  power and `F` is the maximum voting power of faulty validators).
```

spec/consensus/bft-time.md
```diff
@@ -3,54 +3,120 @@ order: 2
...
+- We have the following `Precommit` messages in some `LastCommit` field (we
+ignore all fields except the `Timestamp` field): (p1, 100), (p2, 98), (p3, 1000), (p4, 500).
+We assume that p3 and p4 are faulty validators.
```
sergio-mena, line 51:
```suggestion
  ignore all fields except the `Timestamp` field): (p1, 100), (p2, 98), (p3, 1000), (p4, 500).
  We assume that p3 and p4 are faulty validators.
```

spec/consensus/bft-time.md
```diff
@@ -3,54 +3,120 @@ order: 2
...
+
+The proposer of a round of consensus produces a block to be proposed.
+The proposed block must include a `Commit` set from the commit round of the
+previous height, as the block's `LastCommit` field.
+
+The `Time` for the proposed block is then set as `Block.Time = MedianTime(block.LastCommit)`.
+
+Since the block `Time` is produced in a deterministic way, every node that
+receives the proposed block, can validate `Block.Time` using the same
+procedure.  Block with wrongly computed block times are rejected.
```
sergio-mena, line 100:
```suggestion
procedure. Blocks with wrongly computed block times are rejected.
```

spec/consensus/bft-time.md
```diff
@@ -3,54 +3,120 @@ order: 2
...
+  `Precommit` messages at correct validators ensures that the `Time` proposed
+   for the next block is higher than the `Time` of the current block.
+   Since the `Time` of a block is retrieved from a `Precommit`
+   produced by a correct validator, monotonicity is guaranteed.
+
+- **Byzantine Fault Tolerance**: given a `Commit` set that forms the
+  `LastCommit` field, a range of [valid values](#proposed-time) for the `Time` field of the
+block is defined only by `Precommit` messages produced by correct validators,
+i.e., faulty validators cannot arbitrarily influence (increase or decrease) the
+`Time` value.
```
sergio-mena, line 116:
```suggestion
  block is defined only by `Precommit` messages produced by correct validators,
  i.e., faulty validators cannot arbitrarily influence (increase or decrease) the
  `Time` value.  
```

spec/consensus/signing.md
sergio-mena: > and could even be removed at some point

It will, for BLS

spec/consensus/time.md
```diff
@@ -0,0 +1,52 @@
...
+
+The Time produced by CometBFT satisfies the following properties:
+
+- **Time Monotonicity**: time is monotonically increasing.  More precisely, given
+  two block headers `H1` of height `h1` and `H2` of height `h2`,
+  it is guaranteed that if `h2 > h1` then `H2.Time > H1.Time`.
+
+- **Byzantine Fault Tolerance**: faulty or malicious processes should not be able
+  to arbitrarily influence (increase or decrease) the block Time.
+  In other words, the Time of blocks should be defined by correct processes.
```
sergio-mena, line 21: I have a (small) problem with this sentence. If a byzantine process (which is byzantine for the whole run) decides to use a "correct" time value, this sentence is not correct, is it?
I know it can't happen in BFT-Time, but it _could_ in PBTS if the byzantine node in question is the proposer.

spec/consensus/time.md
sergio-mena: An idea would be to express it with a bound, unknown but existing. Something like:

_ block times bear some resemblance to real time. For every decided block `H` of height `h`, there exists an unknown bound `B` such that, `| H.Timestamp - real_time | < B`, where `real_time` is the actual time when `H` was first proposed.

Anyway, I'm not sure about this. You may well decide to keep the current text, I'm OK with that.

spec/consensus/time.md
```diff
@@ -0,0 +1,52 @@
...
+- **Relation to real time**: block times bear some resemblance to real time.
+  In other words, block times should represent, within some reasonable accuracy,
+  the actual time at which blocks were produced.
+
+## Implementations
+
+CometBFT implements two algorithms for computing block times:
+
+- [BFT Time][bft-time]: the algorithm adopted in versions up to `v0.38.x`;
+  available, in legacy mode, in version `v1.x`.
```
sergio-mena, line 34:
```suggestion
  available, in legacy mode, in version `v1.x` and later.
```

sergio-mena (approved):
Excellent!

spec/consensus/time.md
cason: We had an `ACCURACY` parameter in the original spec. But relating it to `PRECISION` is not really trivial, although one can assume that `ACCURACY <= 2*PRECISION`.

In this case, we have to add a `2` factor for all `PRECISION` entries in the timely properties. Which has rendered the spec more complex, and we are not sure if this reasoning is right.

cason: See: https://github.com/cometbft/cometbft/blob/master/spec/consensus/proposer-based-timestamp/pbts-sysmodel_002_draft.md#accuracy

spec/consensus/time.md
cason: In this case they are not arbitrarily influencing the block time, are them?

spec/consensus/time.md
cason: Are we making this promise?

spec/consensus/time.md
sergio-mena: My suggestion is really basic... definitely not proposing to change the sysmodel in the spec. It was just another possible wording for the "Relation to real time" definition.

spec/consensus/time.md
sergio-mena: Agreed, that's about the first sentence, which I'm fine with.
The problem is the second sentence, which says "the Time of blocks should be defined by correct processes".

sergio-mena: As is starts with "In other words...", consider removing it

spec/consensus/time.md
sergio-mena: Are we gonna be able to nuke it, and still support blocksyncing from arbitrarily in the past? I don't think so.
Anyway, I'm also OK without the "and later".

Issue #2320, opened 2024-02-13
PBTS: improve the logging of untimely messages


cason:
Messages rejected for being untimely should be logged in a more visible (e.g. Info level) and clear way.

Currently, the logs produced are in debug level ([source](https://github.com/cometbft/cometbft/blob/feature/pbts/internal/consensus/state.go#L1385-L1398)):

```
% go test -v -run TestPBTS  | grep timely
D[2024-02-13|10:25:23.095] prevote step: Proposal is not timely; prevoting nil module=consensus height=2 round=0 proposed=2024-02-13T09:25:23.082075Z received=2024-02-13T09:25:23.095122Z msg_delay=10ms precision=1ms
D[2024-02-13|10:25:23.119] prevote step: Proposal is not timely; prevoting nil module=consensus height=2 round=0 proposed=2024-02-13T09:25:23.209133Z received=2024-02-13T09:25:23.119325Z msg_delay=10ms precision=1ms
```

It is not clear whether the proposal timestamp is in the past, in the future, and subtracting time values "by hand" is hard. Moreover, the terms `proposed`, meaning proposal's timestamp, and `received`, meaning proposal's receive time are not clear.

## Definition of Done

- [ ] Untimely proposals, leading to a nil prevote, are logger with more visibility (e.g. Info leve)
- [ ] Log messages for untimely proposals should be more clear
- [ ] Option: log (e.g., Debug level) also timely proposals

cason:
While running tests for this case, I realized that we are not handling recovery properly:

```
$ go test -v -run TestWALCrash | grep timely
I[2024-02-13|11:09:32.266] prevote step: Proposal.Timestamp is not timely; prevoting nil height=1 round=0 reason="timestamp too far in the future" timestamp=2024-02-13T10:09:32.247514Z receive_time=0001-01-01T00:00:00Z difference=-2562047h47m16.854775808s max_difference=-10ms
I[2024-02-13|11:09:32.371] prevote step: Proposal.Timestamp is not timely; prevoting nil height=1 round=0 reason="timestamp too far in the future" timestamp=2024-02-13T10:09:32.352564Z receive_time=0001-01-01T00:00:00Z difference=-2562047h47m16.854775808s max_difference=-10ms
[...]

PR #2321, opened 2024-02-13
feat(consensus): improve logging for timely and untimely messages

cason:
Addresses  #2320.

For instance, the new logging of timely/untimely messages is:

```
$ go test -v -run TestPBTSEnableHeight | grep timely
I[2024-03-01|13:36:37.072] prevote step: Proposal is not timely; prevoting nil module=consensus height=4 round=0 timestamp=2024-03-01T12:36:38.071256Z receive_time=2024-03-01T12:36:37.072355Z timestamp_difference=-998.901ms lower_bound=-500ms upper_bound=2.5s
D[2024-03-01|13:36:37.998] prevote step: Proposal is timely             module=consensus height=4 round=1 timestamp=2024-03-01T12:36:37.997463Z receive_time=2024-03-01T12:36:37.997737Z timestamp_difference=274µs
D[2024-03-01|13:36:38.014] prevote step: Proposal is timely             module=consensus height=5 round=0 timestamp=2024-03-01T12:36:38.014012Z receive_time=2024-03-01T12:36:38.01419Z timestamp_difference=178µs
```

Notice that untimely messages are logging in INFO level, while timely messages are logging in DEBUG level.

Note: I had to force push this branch because the changes were conflicting a lot with parallel changes. This PR now only addresses logging, metrics are left for another PR.

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/internal/consensus/state.go
+++ b/internal/consensus/state.go
@@ -1401,14 +1401,20 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
 		if cs.Proposal.POLRound == -1 && cs.LockedRound == -1 && !cs.proposalIsTimely() {
 			minimumDifference, maximumDifference := cs.timelyProposalMargins()
 			logger.Debug("prevote step: Proposal is not timely; prevoting nil",
-				"proposed",
-				cmttime.Canonical(cs.Proposal.Timestamp).Format(time.RFC3339Nano),
-				"received",
-				cmttime.Canonical(cs.ProposalReceiveTime).Format(time.RFC3339Nano),
+				"timestamp", cs.Proposal.Timestamp.Format(time.RFC3339Nano),
+				"receive_time", cs.ProposalReceiveTime.Format(time.RFC3339Nano),
+				"timestamp_difference", cs.ProposalReceiveTime.Sub(cs.Proposal.Timestamp),
 				"minimum_difference", minimumDifference,
 				"maximum_difference", maximumDifference)
 			cs.signAddVote(types.PrevoteType, nil, types.PartSetHeader{}, nil)
 			return
+		}
+
+		if cs.Proposal.POLRound == -1 {
+			logger.Debug("prevote step: Proposal is timely",
+				"timestamp", cs.Proposal.Timestamp.Format(time.RFC3339Nano),
+				"receive_time", cs.ProposalReceiveTime.Format(time.RFC3339Nano),
+				"timestamp_difference", cs.ProposalReceiveTime.Sub(cs.Proposal.Timestamp))
 		}
 	}
 
--- a/internal/consensus/state.go
+++ b/internal/consensus/state.go
@@ -1400,6 +1400,7 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
 
 		if cs.Proposal.POLRound == -1 && !cs.proposalIsTimely() {
 			lowerBound, upperBound := cs.timelyProposalMargins()
+			// TODO: use Warn level once available.
 			logger.Info("prevote step: Proposal is not timely; prevoting nil",
 				"timestamp", cs.Proposal.Timestamp.Format(time.RFC3339Nano),
 				"receive_time", cs.ProposalReceiveTime.Format(time.RFC3339Nano),
--- a/internal/consensus/state.go
+++ b/internal/consensus/state.go
@@ -1351,6 +1351,18 @@ func (cs *State) enterPrevote(height int64, round int32) {
 	// (so we have more time to try and collect +2/3 prevotes for a single block)
 }
 
+func (cs *State) timelyProposalMargins() (time.Duration, time.Duration) {
+	sp := types.AdaptiveSynchronyParams(
+		cs.state.ConsensusParams.Synchrony.Precision,
+		cs.state.ConsensusParams.Synchrony.MessageDelay,
+		cs.Round,
+	)
+
+	// cs.ProposalReceiveTime - cs.Proposal.Timestamp >= -1 * Precision
+	// cs.ProposalReceiveTime - cs.Proposal.Timestamp <= MessageDelay + Precision
+	return -sp.Precision, sp.MessageDelay + sp.Precision
+}
+
 func (cs *State) proposalIsTimely() bool {
 	sp := types.AdaptiveSynchronyParams(
 		cs.state.ConsensusParams.Synchrony.Precision,
@@ -1387,15 +1399,14 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
 		}
 
 		if cs.Proposal.POLRound == -1 && cs.LockedRound == -1 && !cs.proposalIsTimely() {
+			minimumDifference, maximumDifference := cs.timelyProposalMargins()
 			logger.Debug("prevote step: Proposal is not timely; prevoting nil",
 				"proposed",
 				cmttime.Canonical(cs.Proposal.Timestamp).Format(time.RFC3339Nano),
 				"received",
 				cmttime.Canonical(cs.ProposalReceiveTime).Format(time.RFC3339Nano),
-				"msg_delay",
-				cs.state.ConsensusParams.Synchrony.MessageDelay,
-				"precision",
-				cs.state.ConsensusParams.Synchrony.Precision)
+				"minimum_difference", minimumDifference,
+				"maximum_difference", maximumDifference)
 			cs.signAddVote(types.PrevoteType, nil, types.PartSetHeader{}, nil)
 			return
 		}
--- a/internal/consensus/state.go
+++ b/internal/consensus/state.go
@@ -1399,13 +1399,13 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
 		}
 
 		if cs.Proposal.POLRound == -1 && cs.LockedRound == -1 && !cs.proposalIsTimely() {
-			minimumDifference, maximumDifference := cs.timelyProposalMargins()
-			logger.Debug("prevote step: Proposal is not timely; prevoting nil",
+			lowerBound, upperBound := cs.timelyProposalMargins()
+			logger.Info("prevote step: Proposal is not timely; prevoting nil",
 				"timestamp", cs.Proposal.Timestamp.Format(time.RFC3339Nano),
 				"receive_time", cs.ProposalReceiveTime.Format(time.RFC3339Nano),
 				"timestamp_difference", cs.ProposalReceiveTime.Sub(cs.Proposal.Timestamp),
-				"minimum_difference", minimumDifference,
-				"maximum_difference", maximumDifference)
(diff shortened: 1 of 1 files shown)


sergio-mena:
Is this ready for review?

internal/consensus/metrics.go
```diff
@@ -125,11 +125,17 @@ type Metrics struct {
...
+	// of the validator at the time it receives a proposal message, and the
+	// timestamp of the received proposal message.
+	//
+	// The buckets are defined based on the default values of the Synchrony
+	// consensus parameters: Precision = 0.5s and MessageDelay = 2.0s. The
+	// first bucket represents differences less than than -Precision, while
+	// the last twobuckets represent differences greater than (MessageDelay
+	// + Precision). In both extremes, the proposal is not timely.
+	// metrics:Difference between the local time when a proposal message is received and the timestamp in the proposal message.
+	ProposalTimestampDifference metrics.Histogram `metrics_bucketsizes:"-.5, 0, .5, 1, 1.5, 2, 2.5, 3" metrics_labels:"is_timely"`
```
glnro, line 138: Is there a link to a discussion/issue where there's consensus that these values should replace the old ones?

internal/consensus/state.go
```diff
@@ -1360,6 +1360,17 @@ func (cs *State) proposalIsTimely() bool {
 	return cs.Proposal.IsTimely(cs.ProposalReceiveTime, sp)
 }
 
+func (cs *State) timelyProposalMargins() (time.Duration, time.Duration) {
```
glnro, line 1363: These margins are not adaptive according to the round, see [this pattern](https://github.com/cometbft/cometbft/pull/2431/files#diff-c27a58d5a8b3c15166169543b3cdb94da91bb60efa27609ef446719474af78c5R2696-R2702)

glnro: Perhaps we don't export `SynchronyParams` as a type and instead only export the constructor function. Can follow up in a subsequent PR

internal/consensus/metrics.go
```diff
@@ -125,11 +125,17 @@ type Metrics struct {
...
 
-	// ProposalTimestampDifference is the difference between the timestamp in
-	// the proposal message and the local time of the validator at the time
-	// that the validator received the message.
-	// metrics:Difference between the timestamp in the proposal message and the local time of the validator at the time it received the message.
-	ProposalTimestampDifference metrics.Histogram `metrics_bucketsizes:"-10, -.5, -.025, 0, .1, .5, 1, 1.5, 2, 10" metrics_labels:"is_timely"`
+	// ProposalTimestampDifference is the difference between the local time
+	// of the validator at the time it receives a proposal message, and the
+	// timestamp of the received proposal message.
+	//
```
sergio-mena, line 131:
```suggestion
```

internal/consensus/metrics.go
```diff
@@ -125,11 +125,17 @@ type Metrics struct {
...
+	// The buckets are defined based on the default values of the Synchrony
+	// consensus parameters: Precision = 0.5s and MessageDelay = 2.0s. The
+	// first bucket represents differences less than than -Precision, while
+	// the last twobuckets represent differences greater than (MessageDelay
```
sergio-mena, line 135:
```suggestion
	// the last two buckets represent differences greater than (MessageDelay
```

internal/consensus/metrics.go
sergio-mena: There's an issue for that, I think. And also we agreed yesterday that the QA folks will run a preliminary 200-node run to gather metrics and fine-tune these values

internal/consensus/state.go
cason: Yes, we need to adapt this once #2431 is merged.

Actually, let me make it the target branch.

internal/consensus/metrics.go
cason: https://github.com/cometbft/cometbft/issues/2323

We should add there this requirement.

cason: Added.

internal/consensus/state.go
```diff
@@ -1377,27 +1389,37 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
...
-			logger.Debug("prevote step: Proposal is not timely; prevoting nil",
-				"proposed",
-				cmttime.Canonical(cs.Proposal.Timestamp).Format(time.RFC3339Nano),
-				"received",
-				cmttime.Canonical(cs.ProposalReceiveTime).Format(time.RFC3339Nano),
-				"msg_delay",
-				cs.state.ConsensusParams.Synchrony.MessageDelay,
-				"precision",
-				cs.state.ConsensusParams.Synchrony.Precision)
+	if pbtsEnabled && cs.Proposal.POLRound == -1 {
```
sergio-mena, line 1401: What happened with `cs.LockedRound == -1`?

cason: I deliberately removed it. As it is not part of the PBTS algorithm, and no-one proved me (e.g., via test units) that this condition changes the operation in any way.

sergio-mena:
> Addresses  #2323

Can we say it _partially_ addresses it? We used that issue to describe our QA test cases with @hvanz, so I wouldn't like to close #2323 until those tests are run.
Or, should we move the test cases to another issue?

internal/consensus/state.go
```diff
@@ -1377,27 +1389,37 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
...
+			Observe(timestampDifference.Seconds())
+
+		if !isProposalTimely {
+			logger.Info("prevote step: Proposal is not timely; prevoting nil",
+				"proposal_timestamp_difference", timestampDifference,
+				"lower_bound", lowerBound,
+				"upper_bound", upperBound)
 			cs.signAddVote(types.PrevoteType, nil, types.PartSetHeader{}, nil)
 			return
+		} else {
```
sergio-mena, line 1418: You can get rid of the `else`. If feels weird after a return

internal/consensus/state.go
```diff
@@ -1377,27 +1389,37 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
 		return
 	}
 
+	pbtsEnabled := cs.isPBTSEnabled(height)
+	if pbtsEnabled && !cs.Proposal.Timestamp.Equal(cs.ProposalBlock.Header.Time) {
+		logger.Debug("prevote step: proposal timestamp does not match block time; prevoting nil")
```
sergio-mena, line 1394: would be good to show the two values in the log message

internal/consensus/state.go
```diff
@@ -1377,27 +1389,37 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
...
-				cs.state.ConsensusParams.Synchrony.MessageDelay,
-				"precision",
-				cs.state.ConsensusParams.Synchrony.Precision)
+	if pbtsEnabled && cs.Proposal.POLRound == -1 {
+		isProposalTimely := cs.proposalIsTimely()
+
+		// For logging and metrics
+		lowerBound, upperBound := cs.timelyProposalMargins()
+		timestampDifference := cs.ProposalReceiveTime.Sub(cs.Proposal.Timestamp)
+		cs.metrics.ProposalTimestampDifference.With(
```
cason, line 1407: We should produce this metric, and maybe print (in debug mode) this difference even when PBTS is not in use.

This does not make sense, for PBTS, when we have `PolRound >= 0`.

See: https://github.com/cometbft/cometbft/issues/2323#issuecomment-1966601461

internal/consensus/state.go
```diff
@@ -1377,28 +1389,42 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
...
-				"msg_delay",
-				cs.state.ConsensusParams.Synchrony.MessageDelay,
-				"precision",
-				cs.state.ConsensusParams.Synchrony.Precision)
-			cs.signAddVote(types.PrevoteType, nil, types.PartSetHeader{}, nil)
-			return
-		}
+		logger.Debug("prevote step: Proposal is timely",
+			"proposal_timestamp", cs.Proposal.Timestamp,
+			"receive_time", cs.ProposalReceiveTime,
```
cason, line 1424: Added the times here.

internal/consensus/state.go
cason: Added.

internal/consensus/metrics.go
```diff
@@ -125,11 +125,17 @@ type Metrics struct {
...
+	// of the validator at the time it receives a proposal message, and the
+	// timestamp of the received proposal message.
+	//
+	// The buckets are defined based on the default values of the Synchrony
+	// consensus parameters: Precision = 0.5s and MessageDelay = 2.0s. The
+	// first bucket represents differences less than than -Precision, while
+	// the last two buckets represent differences greater than (MessageDelay
+	// + Precision). In both extremes, the proposal is not timely.
+	// metrics:Difference between the local time when a proposal message is received and the timestamp in the proposal message.
+	ProposalTimestampDifference metrics.Histogram `metrics_bucketsizes:"-.5, 0, .5, 1, 1.5, 2, 2.5, 3"`
```
cason, line 138: Removed the `is_timely` label. Remember to update documentation.

sergio-mena (approved):
🚀

internal/consensus/state.go
```diff
@@ -1378,27 +1390,32 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
...
-				"proposed",
-				cmttime.Canonical(cs.Proposal.Timestamp).Format(time.RFC3339Nano),
-				"received",
-				cmttime.Canonical(cs.ProposalReceiveTime).Format(time.RFC3339Nano),
-				"msg_delay",
-				cs.state.ConsensusParams.Synchrony.MessageDelay,
-				"precision",
-				cs.state.ConsensusParams.Synchrony.Precision)
+		if cs.Proposal.POLRound == -1 && !cs.proposalIsTimely() {
+			lowerBound, upperBound := cs.timelyProposalMargins()
```
lasarojc, line 1402:
```suggestion
			lowerBound, upperBound := cs.timelyProposalMargins()
            // TODO: use Warn level once availabel.
```

lasarojc (approved):

internal/consensus/state.go
sergio-mena: I dug around a bit:
* `cs.LockedRound == -1` is present on [tip of `v0.36.x`](https://github.com/cometbft/cometbft/blob/v0.36.x/internal/consensus/state.go#L1535)
* By blaming that version, we get it was committed as part of the big branch-squasing PR (tendermint/tendermint#7605) which we forward-ported first. So this confirms our suspicions
* Finally, out of curiosity, I went to branch `wb/pbts-rebase-master` (which looks like the branch tendermint/tendermint#7605 was squashing) and it leads to tendermint/tendermint#7415. We might want to ask @ancazamfir to check we're not missing anything here when she's back.

Issue #1822, opened 2023-12-13
storage: add version to each DB


melekes:
### Summary

Add a version to each DB, so if/when the schema changes, we can rely on it

### Problem Definition

Currently, the DBs (block, evidence, state, light) are not versioned. If the schema changes, there's no way to check what version you're running besides parsing data. In #1814, I will have to assume the data format, but that's not ideal from multiple perspectives.

### Proposal

Add a version key-value pair to each DB (e.g., `version=1`). Upon a startup, check that the version in the code matches the one from the DB. If no version is found, assume 1 (initial version).

sergio-mena:
I support this. Will also likely improve troubleshooting.
The only caveat from my side: the solution's design must take into account what happens with old versions of the software that predate this. We should try to avoid breaking those.

Issue #1044, opened 2023-06-28
Benchmark current storage improvements


jmalicevic:
We need to create a testing infrastructure focused on storage in order to test and validate the storage optimizations and changes we plan to implement based on the outcome of the issues where we define metrics and understand workloads (#46),  create a baseline (#67), and evaluate database engines (#64).

In order to create this testing environment, the following tasks should be performed:

- Identify a workload that generates enough data on storage for meaningful testing (at least 10s of GB) on a single machine. 
- Containerized (Docker) environment for testing that can be run locally and in a CI environment
- Support for multiple ComeBFT binaries compiled with all the supported databases (e.g. LevelDB, RocksDB)
- Allow testing for multiple database configuration options
- Generate testing output data that can be used to generate reports that can display improvements (or not) against a baseline or any other particular branch/commit.

17.01.2024.

The scope of the issue has been changed and here is the plan on what will be completed during Q1 2024. 
- [x] Identify a workload that generates enough data for meaningful storage.
- [x] Record the size of the databases during the experiment

## Definition of Done (DoD):

- [x] Decided and fixed workload that will be used to benchmark any storage improvements. 
- [x] Running the workload with metrics done in #46 on baseline CometBFT without the optimizations
- [x] Running the workload with metrics done in #46 on CometBFT with optimizations from #1041 and #49. 
- [ ] Produce a report comparing the results obtained by all the experiments.  #2596

## **Deprioritized - left for later (open issue to continue tracking if needed):**
- Any end user (e.g. developer) can run this testing infrastructure locally without too much local configuration and can generate testing data for validating the database optimizations during the development process
- The storage testing e2e runs in a CI infrastructure (Github actions) (optional, it is not yet well defined and not clear what it would contain). 
- Generated reports that can be shared with the community that outline the improvements accomplished as a result of optimizations and new implementations

PR #2327, opened 2024-02-13
feat(store)!: Support for different key layout

### LINKED ISSUE #2057: Plan and implement database migration to support a new key layout
If the representation of keys in the database is changed based on #1041 and supported by benchmarks in #1044, we need to understand how users would migrate  their existing databases to a new layout. 

One of the main concerns raised by users was that migration on upgrading while a node is offline is not always feasibly due to downtime. 

There have been several proposals to address this : https://github.com/cometbft/cometbft/pull/1814#discussion_r1444497860 .

jmalicevic:
Closes #2057 , #1041

If we keep the current design, we are also closing https://github.com/cometbft/cometbft/issues/1822 

Supersedes #1764 

This PR implements support for an additional DB key representation. The different key layout sorts the entries by height instead of lexicographically as in the current version of Comet. 

When starting this work, we hoped that the new layout would significantly outperform the current layout. As we do not have sufficient real world evidence for this, this PR introduces a DB key layout interface that would allow Comet to easily integrate a more preferential key representation without major breaking changes. 

The layout using ordercode is introduced as experimental, allowing users to easily experiment with this.

This layout was thoroughly tested as part of #1044 and all results will be in a report closing the mentioned PR. 
 
Locally tested:
- Empty stores get initialized with v2
- Existing stores without a version key get initialized to v1 and the key is set
- When a nodes' stores are deleted and we boot it up again that node uses v2 while the rest of the nodes use v1

---

#### PR checklist

- [x] Tests written/updated


internal/store/store.go
```diff
@@ -58,6 +58,8 @@ type BlockStore struct {
 	mtx    cmtsync.RWMutex
 	base   int64
 	height int64
+
+	dbKeyLayout BlockKeyLayout
```
mzabaluev, line 62: Am I wrong in thinking that the `BlockStore` API is at a too low abstraction level to be effectively implemented in all suitable database backends?

It assumes a global key-value store and segregates relations with prefixes, while using multiple separate KV trees (if supported by the DB engine, which exist in the wild) might allow tuning specifically for different logical data tables. In particular, this API rules out an efficient SQL implementation.

This is an architectural musing, not necessarily calling for action in this PR.

internal/state/store.go
```diff
@@ -144,10 +198,45 @@ func NewStore(db dbm.DB, options StoreOptions) Store {
...
+			panic(err)
+		}
+		if options.Logger != nil {
+			options.Logger.Info("State store version ", "version", "v2")
+		}
+		return store
+	}
+
+	versionNum, err := db.Get([]byte("version"))
+	if len(versionNum) == 0 && err == nil {
```
melekes, line 221: I think we should panic if `err != nil` similar to evidence

internal/store/store.go
melekes: @mzabaluev that's something to keep in mind, for sure 👍 However, I'm skeptical about switching to an SQL DB (SQLite) in `v1` as it requires planning, execution, and evaluation (benchmarking, testing). Maybe this is something we should examine when picking a single DB storage.

internal/store/store.go
```diff
@@ -79,6 +81,30 @@ func WithMetrics(metrics *Metrics) BlockStoreOption {
...
+func setDBLayout(bStore *BlockStore) {
+	if bStore.IsEmpty() {
+		bStore.dbKeyLayout = &v2Layout{}
+		if err := bStore.db.SetSync([]byte("version"), []byte("2")); err != nil {
+			panic(err)
+		}
+		return
+	}
+	versionNum, err := bStore.db.Get([]byte("version"))
+	if len(versionNum) == 0 && err == nil {
```
melekes, line 93: same comment about if `err != nil`

internal/store/store.go
```diff
@@ -101,6 +129,16 @@ func NewBlockStore(db dbm.DB, options ...BlockStoreOption) *BlockStore {
...
 }
 
+func (bs *BlockStore) GetVersion() string {
+	switch bs.dbKeyLayout.(type) {
+	case *v1LegacyLayout:
+		return "v1"
+	case *v2Layout:
+		return "v2"
+	}
+	return "no version set"
```
melekes, line 139: Should we default to `v1`?

jmalicevic: I don't think so because at this point in the code we should have set this. So if this return false it is a scenario not supported by us. It is true someone else might have called it before but I think it's better to return this rather than assume something.

internal/store/store.go
jmalicevic: 7e4844ec8

internal/state/store.go
jmalicevic: 7e4844ec8 added it but added a warning that in case the DB starts throwing errors on key not found this should not panic.

config/toml.go
```diff
@@ -553,6 +553,17 @@ peer_query_maj23_sleep_duration = "{{ .Consensus.PeerQueryMaj23SleepDuration }}"
 # reindex events in the command-line tool.
 discard_abci_responses = {{ .Storage.DiscardABCIResponses}}
 
+# The representation of keys in the database.
```
melekes, line 556: docs update (docs/explanation/core/configuration.md) is missing

light/store/db/dbKeyLayout.go
```diff
@@ -0,0 +1,106 @@
+package db
```
melekes, line 1: Shouldn't the filename be `db_key_layout.go`?

config/toml.go
jmalicevic: Done, also the part on compaction was missing so I opened a PR for it https://github.com/cometbft/cometbft/pull/2600

docs/explanation/core/configuration.md
```diff
@@ -457,6 +457,17 @@ peer_query_maj23_sleep_duration = "2s"
...
+# The representation of keys in the database.
+# The current representation of keys in Comet's stores is considered to be v1
+# Users can experiment with a different layout by setting this field to v2.
+# Not that this is an experimental feature and switching back from v2 to v1
+# is not supported by CometBFT.
+# If the database was initially created with v1, it is necessary to migrate the DB
+# before switching to v2. The migration is not done automatically.
+# v1 - the legacy layout existing in Comet prior to v1.
+# v2 - Order preserving representation ordering entries by height.
+experimental_db_key_layout = "{{ .Storage.ExperimentalKeyLayout }}"
```
melekes, line 469:
```suggestion
experimental_db_key_layout = "v1"
```

melekes (approved):
👍

melekes:
@mergify backport v1.x

PR #2328, opened 2024-02-13
feat(e2e): Log number of sent txs (success and failed)

hvanz:
When we run e2e's `load` command there is no message telling if the txs were sent or not, and if not, for what reason (eg. mempool is full, txs is too large, wrong endpoint, etc.). This PR will log a message every ~1 second with the number of successfully sent and failed transactions.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/test/e2e/runner/load.go
+++ b/test/e2e/runner/load.go
@@ -24,6 +24,7 @@ func Load(ctx context.Context, testnet *e2e.Testnet) error {
 	initialTimeout := 1 * time.Minute
 	stallTimeout := 30 * time.Second
 	chSuccess := make(chan struct{})
+	chFailed := make(chan struct{})
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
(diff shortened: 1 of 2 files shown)


melekes (approved):
👍

test/e2e/runner/load.go
```diff
@@ -40,31 +41,39 @@ func Load(ctx context.Context, testnet *e2e.Testnet) error {
...
 		}
 	}
 
-	// Monitor successful transactions, and abort on stalls.
-	success := 0
+	// Monitor successful and failed transactions, and abort on stalls.
+	success, failed := 0, 0
 	timeout := initialTimeout
 	for {
+		rate := log.NewLazySprintf("%.1f", float64(success)/time.Since(started).Seconds())
```
andynog, line 52: shouldn't the rate be the total `float64(success+failed)/time.Since(started).Seconds()` to give a better accurate result? I'm assuming here this is the `txs/secs` rate

test/e2e/runner/load.go
```diff
@@ -40,31 +41,39 @@ func Load(ctx context.Context, testnet *e2e.Testnet) error {
...
 				return errors.New("failed to submit any transactions")
 			}
-			logger.Info("load", "msg", log.NewLazySprintf("Ending transaction load after %v txs (%.1f tx/s)...",
-				success, float64(success)/time.Since(started).Seconds()))
+			logger.Info("load", "msg", log.NewLazySprintf("Ending transaction load after %v txs (%.1f tx/s)...", success, rate))
+			return nil
+		}
+
+		// Log every ~1 second the number of sent transactions.
+		if (success+failed)%testnet.LoadTxBatchSize == 0 {
```
andynog, line 71: here you added `success+failed` which validates the comment above about the rate

test/e2e/runner/load.go
```diff
@@ -40,31 +41,39 @@ func Load(ctx context.Context, testnet *e2e.Testnet) error {
...
+			logger.Debug("load", "success", success, "failed", failed, "tx/s", rate)
+		}
+
+		if testnet.LoadMaxTxs > 0 && success >= testnet.LoadMaxTxs {
+			logger.Info("load", "msg", log.NewLazySprintf("Ending transaction load after reaching %v txs (%.1f tx/s)...", success, rate))
```
andynog, line 76: Would also be useful to log ratio of failed/success ? like a `failure rate` might be important to know I'd think

andynog (approved):
lgtm, I've added a few comments

test/e2e/runner/load.go
hvanz: `success/started` is the real load rate of transactions that were actually delivered since the load started, while `(success+failed)/started` is a constant value equal to the parameter `load_tx_batch_size` (the number of expected txs/second).

test/e2e/runner/load.go
hvanz: Here we actually want to produce a log per second, and this works because `success+failed` is equal to `LoadTxBatchSize * time.Since(started).Seconds()`, so `(success+failed)%testnet.LoadTxBatchSize` will be 0 every time a batch of txs is sent (successfully or not).

test/e2e/runner/load.go
hvanz: Ok, I added a success rate.

Issue #2324, opened 2024-02-13
PBTS should be enabled by default in consensus test units


cason:
Enabling PBTS in existing chains or via genesis is performed via consensus parameters (see https://github.com/cometbft/cometbft/issues/2197), as chains have to define safe and good values for the synchronous parameters.

However, in internal test units we should only run consensus with PBTS enabled, as it will become the de facto default for computing timestamps from `v1.x`.

Regading `e2e` tests, the default behavior will be defined by https://github.com/cometbft/cometbft/pull/2284.

PR #2329, opened 2024-02-13
test(consensus): PBTS should be enabled by default in test units

cason:
Addresses #2324.

PBTS is enabled by setting `ConsensusParams.Feature.PbtsEnableHeight` to a height > 0. To be enabled from startup, this value should be set to `1`.

The default values for consensus parameters have both the `Feature` entries set `0`, meaning that both PBTS and vote extensions are disabled. While this setup can be reasonable for in production networks, it does not allow test units to explore these features that are expected to be enabled in new versions.

This PR ensures that PBTS is enabled in all test units. As a side effect, vote extensions are also enabled in all unit tests (except when the test checks the behavior without these features).

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/internal/consensus/byzantine_test.go
+++ b/internal/consensus/byzantine_test.go
@@ -541,7 +541,7 @@ func sendProposalAndParts(
 
 	// votes
 	cs.mtx.Lock()
-	prevote, _ := cs.signVote(types.PrevoteType, blockHash, parts.Header(), block)
+	prevote, _ := cs.signVote(types.PrevoteType, blockHash, parts.Header(), nil)
 	precommit, _ := cs.signVote(types.PrecommitType, blockHash, parts.Header(), block)
 	cs.mtx.Unlock()
 	peer.Send(p2p.Envelope{
--- a/internal/consensus/byzantine_test.go
+++ b/internal/consensus/byzantine_test.go
@@ -143,7 +143,7 @@ func TestByzantinePrevoteEquivocation(t *testing.T) {
 		// allow first height to happen normally so that byzantine validator is no longer proposer
 		if height == prevoteHeight {
 			bcs.Logger.Info("Sending two votes")
-			prevote1, err := bcs.signVote(types.PrevoteType, bcs.ProposalBlock.Hash(), bcs.ProposalBlockParts.Header(), nil)
+			prevote1, err := bcs.signVote(types.PrevoteType, bcs.ProposalBlock.Hash(), bcs.ProposalBlockParts.Header(), bcs.ProposalBlock)
 			require.NoError(t, err)
 			prevote2, err := bcs.signVote(types.PrevoteType, nil, types.PartSetHeader{}, nil)
 			require.NoError(t, err)
@@ -190,8 +190,8 @@ func TestByzantinePrevoteEquivocation(t *testing.T) {
 			extCommit = &types.ExtendedCommit{}
 		case lazyProposer.LastCommit.HasTwoThirdsMajority():
 			// Make the commit from LastCommit
-			veHeightParam := types.DefaultFeatureParams()
-			veHeightParam.VoteExtensionsEnableHeight = height
+			// Vote extensions are enabled by default for test units
+			veHeightParam := lazyProposer.state.ConsensusParams.Feature
 			extCommit = lazyProposer.LastCommit.MakeExtendedCommit(veHeightParam)
 		default: // This shouldn't happen.
 			lazyProposer.Logger.Error("enterPropose: Cannot propose anything: No commit for the previous block")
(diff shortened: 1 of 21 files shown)


sergio-mena:
@cason check out #2284. There, you can control _when_ you enable PBTS via the manifest, rather than hard-coding it

cason:
I have probably to revert all changes here, with a force-push. This should not be a problem, as it is in draft mode.

internal/consensus/byzantine_test.go
```diff
@@ -540,8 +541,8 @@ func sendProposalAndParts(
 
 	// votes
 	cs.mtx.Lock()
-	prevote, _ := cs.signVote(types.PrevoteType, blockHash, parts.Header(), nil)
-	precommit, _ := cs.signVote(types.PrecommitType, blockHash, parts.Header(), nil)
+	prevote, _ := cs.signVote(types.PrevoteType, blockHash, parts.Header(), block)
+	precommit, _ := cs.signVote(types.PrecommitType, blockHash, parts.Header(), block)
```
cason, line 545: Otherwise, vote extensions don't work as expected.

lasarojc: should this be a different PR?

internal/consensus/common_test.go
```diff
@@ -920,6 +920,10 @@ func randGenesisDoc(numValidators int,
 	}
 	sort.Sort(types.PrivValidatorsByAddress(privValidators))
 
+	if consensusParams == nil {
+		consensusParams = test.ConsensusParams()
```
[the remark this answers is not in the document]
lasarojc: Why not force the callers to pass a non-nil value. They could (should?) run with different configurations.

internal/consensus/byzantine_test.go
```diff
@@ -143,7 +143,7 @@ func TestByzantinePrevoteEquivocation(t *testing.T) {
 		// allow first height to happen normally so that byzantine validator is no longer proposer
 		if height == prevoteHeight {
 			bcs.Logger.Info("Sending two votes")
-			prevote1, err := bcs.signVote(types.PrevoteType, bcs.ProposalBlock.Hash(), bcs.ProposalBlockParts.Header(), nil)
+			prevote1, err := bcs.signVote(types.PrevoteType, bcs.ProposalBlock.Hash(), bcs.ProposalBlockParts.Header(), bcs.ProposalBlock)
```
sergio-mena, line 146: Are you sure you need this change? If you do, we have a problem

internal/consensus/byzantine_test.go
```diff
@@ -190,8 +190,8 @@ func TestByzantinePrevoteEquivocation(t *testing.T) {
 			extCommit = &types.ExtendedCommit{}
 		case lazyProposer.LastCommit.HasTwoThirdsMajority():
 			// Make the commit from LastCommit
-			veHeightParam := types.DefaultFeatureParams()
-			veHeightParam.VoteExtensionsEnableHeight = height
+			// Vote extensions are enabled by default for test units
+			veHeightParam := lazyProposer.state.ConsensusParams.Feature
 			extCommit = lazyProposer.LastCommit.MakeExtendedCommit(veHeightParam)
```
sergio-mena, line 195: Very minor (I know this is not directly related to this PR).
Why are we passing the whole `Feature` struct to `MakeExtendedCommit`? Can't we just past the VE enable height?
If you agree but doing this takes much effort, just ignore.

internal/consensus/byzantine_test.go
sergio-mena: I think you can leave the `prevote` one (line 544) to `nil` and it'll still work.
As for the `precommit`... oh my! I looked into it and I think you're right, there's no other way, as the production code is calling `ExtendVote` with the block (and checking that the block we give makes sense).

@lasarojc I think it has to be this PR, otherwise the tests won't pass. @cason unless we decide to run **just this test case** with vote extensions disabled (but I'm also OK with the current solution)

internal/consensus/common_test.go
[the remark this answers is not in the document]
sergio-mena: I checked around and my understanding is that test cases that don't care, just provide `nil` as `ConsensusParams`. And those that do care, they provide their `ConsensusParams` (and we don't touch them here). This approach is fine IMO.

sergio-mena (approved):
This PR is small, but I'm aware of the thought and analysis that went into it so that all our UTs run with PBTS enabled, but also with vote extensions enabled. Something we didn't manage to do last year. Thank you!

internal/consensus/common_test.go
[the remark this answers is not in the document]
cason: There are only two calls to `randGenesisDoc` method. But it is also invoked by other methods, which together have 20+ calls. I think it is good enough if `nil` means "please use the defaults".

internal/consensus/common_test.go
```diff
@@ -933,6 +937,9 @@ func randGenesisState(
 	numValidators int,
 	consensusParams *types.ConsensusParams,
 ) (sm.State, []types.PrivValidator) {
+	if consensusParams == nil {
+		consensusParams = test.ConsensusParams()
+	}
```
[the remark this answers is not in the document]
cason: Indeed, reverting.

internal/consensus/byzantine_test.go
cason: The method uses the helper: `fp.VoteExtensionsEnabled(ec.Height)`. We could use the same logic used there here, but it does not look great.

internal/consensus/byzantine_test.go
cason: > I think you can leave the prevote one (line 544) to nil and it'll still work.

Indeed, reverting.

internal/consensus/byzantine_test.go
sergio-mena: Up to you

PR #2346, opened 2024-02-15
refactor(consensus): print err from SignAndCheckVote

melekes:
Refs https://github.com/cometbft/cometbft/issues/2357

---

#### PR checklist

- [x] Tests written/updated


cason (commented):
This is the typical double, triple checking of conditions present in all this code, unfortunately.

I think that, in this case, the problem is that the `SignAndCheckVote` is a public method that is part of the public API.

Not sure what to do in this case. In particular because we don't really have test units covering this method.

types/vote.go
```diff
@@ -410,33 +416,31 @@ func SignAndCheckVote(
...
 
-	isPrecommit := vote.Type == PrecommitType
-	if !isPrecommit && extensionsEnabled {
-		// Non-recoverable because the caller passed parameters that don't make sense
-		return false, fmt.Errorf("only Precommit votes may have extensions enabled; vote type: %d", vote.Type)
-	}
-
-	isNil := vote.BlockID.IsNil()
-	extSignature := (len(v.ExtensionSignature) > 0)
-	if extSignature == (!isPrecommit || isNil) {
```
melekes, line 425: previously this would error if 

* extSignature is empty
* recommit is true
* isNil is false

so `false == (!true || false), false == false`

types/vote.go
```diff
@@ -410,33 +416,31 @@ func SignAndCheckVote(
...
-		// Non-recoverable because the vote is malformed
-		return false, fmt.Errorf(
-			"extensions must be present IFF vote is a non-nil Precommit; present %t, vote type %d, is nil %t",
-			extSignature,
-			vote.Type,
-			isNil,
+	if extensionsEnabled {
```
cason, line 425: I suggest renaming this filed to `signExtension`, to better represent the new behavior.

types/vote.go
```diff
@@ -401,6 +402,11 @@ func VotesToProto(votes []*Vote) []*cmtproto.Vote {
 	return res
 }
 
+// SignAndCheckVote signs the vote with the given privVal and checks the vote.
+// It returns an error if the vote is invalid and a boolean indicating if the
+// error is recoverable or not.
+//
+// CONTRACT: if extensionsEnabled is true, then vote's type must be PrecommitType.
```
cason, line 409:
```suggestion
// CONTRACT: if extensionsEnabled is true, then vote's type must be PrecommitType and must not be for a nil BlockID.
```

types/vote.go
```diff
@@ -410,33 +416,31 @@ func SignAndCheckVote(
...
-	if extSignature == (!isPrecommit || isNil) {
-		// Non-recoverable because the vote is malformed
-		return false, fmt.Errorf(
-			"extensions must be present IFF vote is a non-nil Precommit; present %t, vote type %d, is nil %t",
-			extSignature,
-			vote.Type,
-			isNil,
+	if extensionsEnabled {
+		var (
+			extSignature = (len(v.ExtensionSignature) > 0)
```
cason, line 427: I suggest splitting the verification here into 3 ifs.

types/vote.go
```diff
@@ -31,6 +31,7 @@ var (
 	ErrVoteNil                       = errors.New("nil vote")
 	ErrVoteExtensionAbsent           = errors.New("vote extension absent")
 	ErrInvalidVoteExtension          = errors.New("invalid vote extension")
+	ErrVoteExtSignatureForNilBlock   = errors.New("vote for a nil block contains extension signature")
```
cason, line 34:
```suggestion
	ErrVoteExtSignatureForNilBlock   = errors.New("Precommit for a nil block contains extension signature")
```

types/vote.go
```diff
@@ -410,33 +416,31 @@ func SignAndCheckVote(
...
-	}
 
-	vote.ExtensionSignature = nil
-	if extensionsEnabled {
+		if !isPrecommit {
+			// Non-recoverable because the vote is malformed.
+			panic(fmt.Sprintf("extensions are enabled, but vote is not Precommit (vote type: %T)", vote.Type))
+		}
+
+		if extSignature && isNil {
```
cason, line 437: From the logic here, if the Precommit is for nil but the signature is empty, we are good.

Not sure how it should work in this case.

cason: From the contract, we should panic here.

types/vote.go
```diff
@@ -419,25 +417,32 @@ func SignAndCheckVote(
...
+	if !isPrecommit && extensionsEnabled {
+		// Non-recoverable because the caller passed parameters that don't make sense
+		return false, fmt.Errorf("only Precommit votes may have extensions enabled; vote type: %T", vote.Type)
+	}
+
+	isNil := vote.BlockID.IsNil()
+	extSignature := (len(v.ExtensionSignature) > 0)
+	if extSignature == (!isPrecommit || isNil) {
+		// Non-recoverable because the vote is malformed
+		return false, fmt.Errorf(
```
mzabaluev, line 431: Can this be an error struct as well?

types/vote.go
```diff
@@ -419,25 +417,32 @@ func SignAndCheckVote(
...
-			return false, ErrVoteNoSignature // Non-recoverable because the vote is malformed.
-		}
 
-		if vote.BlockID.IsNil() {
-			return false, ErrVoteExtSignatureForNilBlock // Non-recoverable because the vote is malformed.
-		}
+	isPrecommit := vote.Type == PrecommitType
+	if !isPrecommit && extensionsEnabled {
+		// Non-recoverable because the caller passed parameters that don't make sense
+		return false, fmt.Errorf("only Precommit votes may have extensions enabled; vote type: %T", vote.Type)
```
mzabaluev, line 424: Same here.

types/vote_test.go
```diff
@@ -490,3 +490,67 @@ func TestVoteProtobuf(t *testing.T) {
...
+				ValidatorIndex:     0,
+				Height:             height,
+				Round:              round,
+				Timestamp:          cmttime.Now(),
+				Type:               PrecommitType,
+				BlockID:            BlockID{make([]byte, 0), PartSetHeader{0, make([]byte, 0)}},
+				ExtensionSignature: []byte("signature"),
+			},
+			expectError: false,
+		},
```
cason, line 541: This code is not test:

if !isPrecommit && extensionsEnabled

cason: This code is also not tested:

if extSignature == (!isPrecommit || isNil) {

cason: I think it is more complex, but can we also test when signing a vote fails? Namely:

if err := privVal.SignVote(chainID, v); err != nil

types/vote.go
```diff
@@ -409,23 +412,24 @@ func SignAndCheckVote(
...
 	}
 
 	isNil := vote.BlockID.IsNil()
 	extSignature := (len(v.ExtensionSignature) > 0)
 	if extSignature == (!isPrecommit || isNil) {
 		// Non-recoverable because the vote is malformed
 		return false, fmt.Errorf(
-			"extensions must be present IFF vote is a non-nil Precommit; present %t, vote type %d, is nil %t",
+			"extensions must be present IFF vote is a non-nil Precommit; present %t, vote type %T, is nil %t",
 			extSignature,
```
cason, line 433: Is a good idea to print a byte array?

types/vote.go
cason: I would print the entire vote here. It is evident that if it is not a Precommit, it must be a Prevote. Otherwise, signing will fail.

types/vote.go
```diff
@@ -409,23 +412,24 @@ func SignAndCheckVote(
...
 			vote.Type,
```
cason, line 434: Again, I think more useful to print the vote here. Not only the type.

cason (approved):
Legit.

The actual change is minor. The added test make tests more readable (although not increasing coverage, as the failure scenarios are not really covered).

types/vote_test.go
melekes: done

PR #2364, opened 2024-02-16
fix: remove TestMempoolFIFOWithParallelCheckTx

faddat:
This test fails 100% of the time.

When new developers begin to use comet, the tests will not pass.

Because they've been taught that tests not passing is a **bad thing** indicative of problems since most new users of CometBFT are / have been software engineers for some time, this makes the product less attractive. 

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/mempool/reactor_test.go
+++ b/mempool/reactor_test.go
@@ -334,40 +334,6 @@ func TestReactorTxSendersMultiNode(t *testing.T) {
 	require.Zero(t, len(firstReactor.txSenders))
 }
 
-func TestMempoolFIFOWithParallelCheckTx(t *testing.T) {
-	if os.Getenv("CI") != "" {
-		t.Skip("FIFO is not supposed to be guaranteed and this this is just used to evidence one of the cases where it does not happen. Hence we skip this test during CI.")
-	}
-
-	config := cfg.TestConfig()
-	reactors, _ := makeAndConnectReactors(config, 4)
-	defer func() {
-		for _, r := range reactors {
-			if err := r.Stop(); err != nil {
-				require.NoError(t, err)
-			}
-		}
-	}()
-	for _, r := range reactors {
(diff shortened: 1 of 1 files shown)


melekes:
@hvanz I think @faddat has a point here because it also confused me. Can we either a) fix the test b) comment it and link to the issue c) remove it and post it in the associated issue as a comment ?

faddat:
I'd like to strongly suggest that we merge:

* https://github.com/cometbft/cometbft/pull/2364

hvanz:
I don't think we will fix this test any time soon. Finding a solution for guaranteeing FIFO ordering is not easy; it would require changes at the p2p level. But we need to keep the test somewhere. I think the priority would be to improve the documentation, specifically the spec on p2p. We now know that the order of messages is not guaranteed, it's just best-effort, but this is not documented anywhere. If this is well understood and documented, we don't need this test. In the meantime, I'd prefer to keep it in the codebase, totally disabled and better documented, so that the test is kept up to date with the rest of the code. Wdyt @lasarojc?

faddat:
@hvanz - I think that if the test is totally disabled, it won't be possible to keep it up to date, because it'll be disabled.

mzabaluev:
> I'd prefer to keep it in the codebase, totally disabled and better documented, so that the test is kept up to date with the rest of the code.

So the idea is to have it compiled so it does not go stale, but not run, correct?

faddat:
ah, I get the idea now.

lasarojc:
Observe that the test is disabled in CI and can be disabled locally by setting `CI=true`.
Or it can be changed to always invoke `t.Skip(...)`  if you prefer that.
I agree with @hvanz that it should be removed only if/when it is well documented that FIFO is best effort.

melekes (approved):

PR #2366, opened 2024-02-17
fix: remove gofumpt from pre-commit configuration 

faddat:
gofumpt is a part of golangci-lint so this entry is redundant.

Additonally, there seem to have been some issues with it.  I think it is best to remove this. 

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,19 +10,6 @@ repos:
     - id: golangci-lint-full
       args: [--timeout, 10m] # for CI
 
- # gofumpt (does not have .pre-commmit-hooks.yaml)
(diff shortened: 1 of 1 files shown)


melekes:
See https://github.com/cometbft/cometbft/pull/2049#issuecomment-1903272907

Osmosis also runs it in addition to `golangci-lint` https://github.com/osmosis-labs/osmosis/blob/7d734f670c045229572f91da6646eeaf31685049/scripts/makefiles/lint.mk#L27

I don't know why there's such a discrepancy between golangci-lint's gofumpt linter and standalone gofumpt, but until it's resolved it's better to keep gofumpt around.

PR #2388, opened 2024-02-20
feat(pbts): Adds timestamp to the `msgInfo` written into the WAL

### LINKED ISSUE #2322: PBTS: Proposal's reception time should be persisted to the WAL
Originally: https://github.com/tendermint/tendermint/issues/8954

For implementing PBTS mechanism, in particular for registering the reception time of `Proposal` message, the consensus reactor registers the `ReceiveTime` of every message in the [msgInfo](https://github.com/cometbft/cometbft/blob/feature/pbts/internal/consensus/state.go#L488-L487) instance delivered to the consensus protocol.

Messages received and accepted by the consensus protocol are written to the WAL, allowing nodes to recover from crashes. However, the [msgInfo proto](https://github.com/cometbft/cometbft/blob/feature/pbts/proto/cometbft/consensus/v1/wal.proto#L14), that is written to the WAL, does not include the `ReceiveTime` field.

As a result, when replaying messages from the WAL their `ReceiveTime` is nil (zero time), which leads nodes to reject a replayed `Proposal` message upon recovery, even though the same `Proposal` message was accepted during regular execution. This leads recovering validator nodes to equivocate.

@sergio-mena has observed this behavior in e2e tests and reported it in detail in https://github.com/tendermint/tendermint/issues/8739. He also proposed a workaround to avoid tests from failing (https://github.com/tendermint/tendermint/pull/8774), by disabling the PBTS `timely` verification of the `Proposal`'s timestamp when in replay mode. The problem of this approach is that it can also lead to equivocating behavior: if the `Proposal`'s timestamp is rejected by PBTS during regular execution, then upon recovery the same `Proposal` can be accepted, as the `timely` verification is disabled.

The straightforward solution for this issue is to write the `ReceiveTime` field of `msgInfo` struct to the WAL during regular operation, and to read the same field from the WAL when replaying messages, so that the protocol processes `Proposal` messages using the same temporal information employed before, during regular operation.

The downside of this solution is to increase the size and the amount of the information written to the WAL, with a potential performance impact. The byte-size of a timestamp is typically 13 bytes, but the main point here is that the `ReceiveTime` is not relevant for most of the received and broadcast messages. In fact, this field is only [employed](https://github.com/cometbft/cometbft/blob/feature/pbts/internal/consensus/state.go#L884) for `Proposal` messages. For all the other messages, having it set to zero (nil) during recovery does not constitute a problem.

### DoD

- [ ] Implement solution (WAL should persist the proposal receive time for proposals made)
  - #2388
- [ ] Test if the solution works using a manual`e2e` test
  - See result: https://github.com/cometbft/cometbft/pull/2388#issuecomment-1964016465
- [ ] **Optional**: test if the solution works by introducing a new test unit
   - This means that the test should fail without the fix, and pass with the fix
   - The current test units infra renders it very hard to test the recovery of a node in consensus' terms

lasarojc:
Closes #2322 

---

#### PR checklist

- [x] Tests written/updated


internal/consensus/wal_test.go
```diff
@@ -82,10 +84,37 @@ func TestWALTruncate(t *testing.T) {
 
 func TestWALEncoderDecoder(t *testing.T) {
```
andynog, line 85: Probably a good test would be to include logic in the `func TestWALMsgProto(t *testing.T)` test to ensure the conversions to proto work fine especially in the edge cases where the `ReceiveTime` can be `nil`. 

Noticed that in this unit test there is no test for this particular WAL message

andynog (approved):
lgtm, made a suggestion on testing

internal/consensus/wal_test.go
lasarojc: Thanks, @andynog 
Yeah, I need to add a test that serializes in the old version and deserializes in the new one

internal/consensus/wal_test.go
```diff
@@ -82,10 +87,37 @@ func TestWALTruncate(t *testing.T) {
...
+	err := vss[0].SignProposal(cs1.state.ChainID, pp)
+	require.NoError(t, err)
+
+	p.Signature = pp.Signature
+
 	msgs := []TimedWALMessage{
 		{Time: now, Msg: EndHeightMessage{0}},
 		{Time: now, Msg: timeoutInfo{Duration: time.Second, Height: 1, Round: 1, Step: types.RoundStepPropose}},
 		{Time: now, Msg: cmttypes.EventDataRoundState{Height: 1, Round: 1, Step: ""}},
+		{Time: now, Msg: msgInfo{Msg: &ProposalMessage{Proposal: &p}, PeerID: "Nobody", ReceiveTime: now}},
```
cason, line 118: You don't test the `ReceiveTime` below.

internal/consensus/msgs_test.go
```diff
@@ -316,6 +316,8 @@ func TestWALMsgProto(t *testing.T) {
 			}
 		})
 	}
+
+	// TODO: Test here instead of wherever?
```
cason, line 320: You could test here a `Proposal` with receive time, another without it.

cason (commented):
Some general comments. This appears to be very hard to properly test.

internal/consensus/msgs.go
```diff
@@ -269,11 +270,20 @@ func WALToProto(msg WALMessage) (*cmtcons.WALMessage, error) {
 		if err != nil {
 			return nil, err
 		}
+
+		var rtp *time.Time = nil
```
sergio-mena, line 274:
```suggestion
		var rtp *time.Time
```

internal/consensus/msgs.go
```diff
@@ -326,11 +336,16 @@ func WALFromProto(msg *cmtcons.WALMessage) (WALMessage, error) {
 		if err != nil {
 			return nil, cmterrors.ErrMsgFromProto{MessageName: "MsgInfo", Err: err}
 		}
-		pb = msgInfo{
+		msgInfo := msgInfo{
```
sergio-mena, line 339: I'm not sure I get why we need this temporary variable

internal/consensus/msgs.go
```diff
@@ -269,11 +270,20 @@ func WALToProto(msg WALMessage) (*cmtcons.WALMessage, error) {
 		if err != nil {
 			return nil, err
 		}
+
+		var rtp *time.Time = nil
+		if !msg.ReceiveTime.IsZero() {
```
sergio-mena, line 275: I see you didn't need to modify the logic where the `ReceiveTime` field is populated/consumed. I'm assuming that's because that logic is reused from the PTBS code we have previously forward-ported. Two questions:
* Is this accurate?
* Are we sure this is addressing the problem we saw initially in tendermint/tendermint#8739 ?

cason: Yes, the second point is really important.

We need to make sure that a node doesn't equivocate after accepting/rejecting a `Proposal` during normal operation, same `Proposal` that will instead be untimely (most likely) or timely (e.g., ts in the future) upon recovery.

internal/consensus/wal_test.go
```diff
@@ -82,10 +84,37 @@ func TestWALTruncate(t *testing.T) {
...
+		{Time: now, Msg: msgInfo{Msg: &ProposalMessage{Proposal: &p}, PeerID: "Nobody", ReceiveTime: time.Time{}}},
+		{Time: now, Msg: msgInfo{Msg: &ProposalMessage{Proposal: &p}, PeerID: "Nobody"}},
 	}
 
 	b := new(bytes.Buffer)
```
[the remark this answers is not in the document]
lasarojc: Equal compares data structures recursively, comparing basic type for equality. 

> Pointer variable equality is determined based on the equality of the referenced values (as opposed to the memory addresses). Function equality cannot be determined and will always fail. 
From [here](https://pkg.go.dev/github.com/stretchr/testify/assert#Equal}

So here it compares the fields of the ProposalMessage and the ReceiveTime of the `msgInfo`.

internal/consensus/wal_test.go
lasarojc: It is done by the `Equal`

sergio-mena:
### Before this PR

Managed to repro tendermint/tendermint#8739 easily with this patch:

```diff
diff --git a/internal/consensus/state.go b/internal/consensus/state.go
index f6dce0882..1faec9da4 100644
--- a/internal/consensus/state.go
+++ b/internal/consensus/state.go
@@ -2557,6 +2557,7 @@ func (cs *State) signAddVote(
 		panic(fmt.Errorf("vote extension absence/presence does not match extensions enabled %t!=%t, height %d, type %v",
 			hasExt, extEnabled, vote.Height, vote.Type))
 	}
+	time.Sleep(1 * time.Second)
 	cs.sendInternalMessage(msgInfo{&VoteMessage{vote}, "", time.Time{}})
 	cs.Logger.Debug("signed and pushed vote", "height", cs.Height, "round", cs.Round, "vote", vote)
 }
diff --git a/test/e2e/networks/simple.toml b/test/e2e/networks/simple.toml
index 96b81f79f..9bfebcf34 100644
--- a/test/e2e/networks/simple.toml
+++ b/test/e2e/networks/simple.toml
@@ -1,4 +1,6 @@
+pbts_enable_height = 1
+
 [node.validator01]
+perturb = ["kill"]
+
 [node.validator02]
-[node.validator03]
-[node.validator04]
```

You need to lauch the `e2e` steps manually, and launch "perturb" several times until it kills `validator01` at the right moment.

These are the logs when reproing:

```
validator01  | I[2024-02-26|12:10:07.766] Replay: Proposal                             module=consensus height=52 round=0 header=1:1A1DCB7F4EE4 pol=-1 peer=local receive_time="0001-01-01 00:00:00 +0000 UTC"
validator01  | I[2024-02-26|12:10:07.767] received proposal                            module=consensus proposal="Proposal{52/0 (30AD0242ABFD507702B85D8F6894C7CC4806ED4DDCCE8A98AC8A6EC0C03E2E97:1:1A1DCB7F4EE4, -1) 4FA743475BE1 @ 2024-02-26T12:09:30.615924583Z}" proposer=D3B5BFD2AEAC159C79812DD9426DE51A971C7A61
validator01  | I[2024-02-26|12:10:07.767] Replay: BlockPart                            module=consensus height=52 round=0 peer=local
validator01  | I[2024-02-26|12:10:07.767] received complete proposal block             module=consensus height=52 hash=30AD0242ABFD507702B85D8F6894C7CC4806ED4DDCCE8A98AC8A6EC0C03E2E97
validator01  | D[2024-02-26|12:10:07.767] entering prevote step                        module=consensus height=52 round=0 current=52/0/RoundStepPropose
validator01  | D[2024-02-26|12:10:07.767] prevote step: Proposal is not timely; prevoting nil module=consensus height=52 round=0 proposed=2024-02-26T12:09:30.615924583Z received=0001-01-01T00:00:00Z msg_delay=2s precision=500ms
validator01  | E[2024-02-26|12:10:07.768] failed signing vote                          module=consensus height=52 round=0 vote="Vote{1:D3B5BFD2AEAC 52/00/SIGNED_MSG_TYPE_PREVOTE(Prevote) 000000000000 000000000000 000000000000 @ 2024-02-26T12:10:07.768676795Z}" err="error signing vote: conflicting data"
```

### After this PR

Coming up

internal/consensus/msgs.go
```diff
@@ -269,11 +270,20 @@ func WALToProto(msg WALMessage) (*cmtcons.WALMessage, error) {
...
+			// Only record the `ReceiveTime` if explicitly set.
+			rt := msg.ReceiveTime
```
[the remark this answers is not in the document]
lasarojc: Yes. Wanted to not reuse the allocated value in `msg`

internal/consensus/msgs.go
lasarojc: This was tested by @sergio-mena using e2e

internal/test/state.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"fmt"
+
+	dbm "github.com/cometbft/cometbft-db"
+	"github.com/cometbft/cometbft/crypto/ed25519"
+	sm "github.com/cometbft/cometbft/internal/state"
+	"github.com/cometbft/cometbft/types"
+)
+
+func State(nVals, height int, chainID string) (sm.State, dbm.DB, map[string]types.PrivValidator) {
```
cason, line 12: NewState maybe?

lasarojc: changed it to what it was, MakeState.

lasarojc:
@cason and @sergio-mena 
Merging since all comments seem to have been addressed.

internal/consensus/common_test.go
```diff
@@ -1024,3 +1026,43 @@ func signDataIsEqual(v1 *types.Vote, v2 *cmtproto.Vote) bool {
...
+
+	for i := 1; i < height; i++ {
+		s.LastBlockHeight++
+		s.LastValidators = s.Validators.Copy()
+		if err := stateStore.Save(s); err != nil {
+			panic(err)
+		}
+	}
+
+	return s, stateDB, privVals
```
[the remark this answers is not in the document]
sergio-mena: Is this addresssed?

sergio-mena (approved):

PR #2399, opened 2024-02-21
refactor(proto): remove pubkey from valsetUpdate and multiple keys in params

tac0turtle:
This PR does two things: 

- remove multiple keys being set in validator params. 
  - this was removed because comet only supports a single key type being used at any one time. Comet may fail to verify many types of keys in a single network 
- remove pubkey from abci interface
  - this was removed since if comet forces a single type to be used for validators then the application does not need to know what type of key is present since the application does not use the pubkey to verify. If users would like to verify things in the application then we can add a pubkey type to validator update. Then its up to the application to setup key verification.  

---

#### PR checklist

- [ ] Tests written/updated


tac0turtle:
this limitation also prevents sdk from being used if a new curve is added to comet.

proto/cometbft/abci/v1/types.proto
```diff
@@ -516,8 +516,8 @@ message Validator {
 
 // ValidatorUpdate is a singular update to a validator set.
 message ValidatorUpdate {
-  cometbft.crypto.v1.PublicKey pub_key = 1 [(gogoproto.nullable) = false];
-  int64                        power   = 2;
+  bytes pub_key = 1;
+  int64  power   = 2;
```
tac0turtle, line 520: if we need type i would pass the key type to avoid the extra dep

proto/cometbft/abci/v1/types.proto
```diff
@@ -516,8 +516,11 @@ message Validator {
 
 // ValidatorUpdate is a singular update to a validator set.
 message ValidatorUpdate {
-  cometbft.crypto.v1.PublicKey pub_key = 1 [(gogoproto.nullable) = false];
-  int64                        power   = 2;
+  int64  power   = 2;
+  bytes pub_key = 3;
+  string key_type = 4;
```
melekes, line 521:
```suggestion
  string pub_key_type = 4;
```

proto/cometbft/types/v1/params.proto
```diff
@@ -58,7 +58,7 @@ message ValidatorParams {
   option (gogoproto.populate) = true;
   option (gogoproto.equal)    = true;
 
-  repeated string pub_key_types = 1;
+string pub_key_type = 1;
```
melekes, line 61:
```suggestion
  string pub_key_type = 1;
```

formatting seems off

proto/tendermint/abci/types.proto
```diff
@@ -435,7 +435,7 @@ message Validator {
 }
 
 message ValidatorUpdate {
-  tendermint.crypto.PublicKey pub_key = 1 [(gogoproto.nullable) = false];
+bytes pub_key = 1 ;
```
melekes, line 438:
```suggestion
  bytes pub_key = 1;
```

formatting seems off

types/params.go
```diff
@@ -132,10 +132,8 @@ func DefaultABCIParams() ABCIParams {
 }
 
 func IsValidPubkeyType(params ValidatorParams, pubkeyType string) bool {
-	for i := 0; i < len(params.PubKeyTypes); i++ {
-		if params.PubKeyTypes[i] == pubkeyType {
-			return true
-		}
+	if params.PubKeyType == pubkeyType {
```
melekes, line 135:
```suggestion
	return params.PubKeyType == pubkeyType
```

melekes (commented):
conceptual ACK 👍

mzabaluev:
Wouldn't it still be nice to have a marker type for the public keys (we've got domain types for this as well, but I'm talking about the level of protobuf bindings), and a structure self-describing the key type rather than relying on ambient knowledge to decode it as the correct type? How about being future-proof if Comet (or an alternative implementation using these protos) does add support for heterogeneous keys?

> this limitation also prevents sdk from being used if a new curve is added to comet.

I don't understand this comment. If a new curve is added, surely there'd be a new tag for `PublicKey` to identify it? Isn't it the whole purpose of the extensible `PublicKey` message type?

proto/cometbft/abci/v1/types.proto
```diff
@@ -516,8 +516,11 @@ message Validator {
...
+
+  reserved 1;  // pub_key is unused in ABCI
```
mzabaluev, line 523: We'll need to preserve the former definition in another beta package for 0.38 a-la #2529 because this is a breaking change in `types.v1` which currently corresponds to 0.38+.

proto/tendermint/abci/types.proto
mzabaluev, line 438: Is it OK to lose the `gogoproto.nullable` attribute here?

proto/cometbft/types/v1/params.proto
mzabaluev: This breaks the field type without changing the field number.

proto/tendermint/abci/types.proto
mzabaluev: Also breaks the type without changing the field number.

proto/tendermint/types/params.proto
```diff
@@ -56,7 +56,7 @@ message ValidatorParams {
   option (gogoproto.populate) = true;
   option (gogoproto.equal)    = true;
 
-  repeated string pub_key_types = 1;
+  string pub_key_type = 1;
```
mzabaluev, line 59: Another type breakage here.

types/params.go
```diff
@@ -57,7 +57,7 @@ type EvidenceParams struct {
 // ValidatorParams restrict the public key types validators can use.
 // NOTE: uses ABCI pubkey naming, not Amino names.
 type ValidatorParams struct {
-	PubKeyTypes []string `json:"pub_key_types"`
+	PubKeyType string `json:"pub_key_type"`
```
mzabaluev, line 60: JSON schema breakage as well. We might need to take this into account in RPC versioning, and we still need the old JSON schema for pre-1.0 nodes.

mzabaluev (changes requested):

proto/tendermint/abci/types.proto
tac0turtle: cant use nullable on a proto type

proto/tendermint/types/params.proto
tac0turtle: can revert and assumer the length will always be 1

types/params.go
tac0turtle: will revert and treat length as always 1

proto/tendermint/abci/types.proto
lasarojc: Why not? Asking because we recently made use of it in another type.

sergio-mena (commented):
Thanks @tac0turtle for this PR.

I am totally convinced about the first part; namely, passing key type info as a string rather than our current convoluted solution involving protobufs. I understand this is a necessary step if we are to eventually allow our users to use new (their own) crypto without having to fork CometBFT.

Regarding the other change, limiting the possible key types in ConsensusParams to only one, here's what I think. As I mention in some of the detailed comments, we'd need to confirm that none of our users do (or would) use that feature. In the meantime, I'd like to understand how that change is helpful to you, and whether it would be an option to add that limitation at application (SDK) level.

proto/tendermint/abci/types.proto
andynog: I think If you remove the `nullable` you will have to check for `nil` every time you access pub_key. Failure to do so might lead to a runtime panic due to a nil pointer dereference. Also tests would need to be adapted to handle this use case.

types/params.go
```diff
@@ -132,6 +132,10 @@ func DefaultABCIParams() ABCIParams {
 }
 
 func IsValidPubkeyType(params ValidatorParams, pubkeyType string) bool {
+	// disallow more than one pubkey type
+	if len(params.PubKeyTypes) > 1 {
```
tac0turtle, line 136: let me know of this is fine

proto/cometbft/abci/v1/types.proto
tac0turtle: reverted

proto/tendermint/types/params.proto
tac0turtle: reverted

proto/cometbft/abci/v1/types.proto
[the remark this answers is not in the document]
tac0turtle: reverted this change

proto/tendermint/abci/types.proto
```diff
@@ -435,7 +435,7 @@ message Validator {
 }
```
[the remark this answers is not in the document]
tac0turtle: done

proto/tendermint/types/params.proto
```diff
@@ -56,7 +56,7 @@ message ValidatorParams {
   option (gogoproto.populate) = true;
```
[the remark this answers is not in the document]
tac0turtle: done

types/protobuf.go
```diff
@@ -59,14 +59,9 @@ func (tm2pb) PartSetHeader(header PartSetHeader) cmtproto.PartSetHeader {
...
 
-// XXX: panics on unknown pubkey type.
 func (tm2pb) ValidatorUpdate(val *Validator) abci.ValidatorUpdate {
-	pk, err := cryptoenc.PubKeyToProto(val.PubKey)
-	if err != nil {
-		panic(err)
-	}
 	return abci.ValidatorUpdate{
-		PubKey: pk,
+		PubKey: val.PubKey.Address(),
```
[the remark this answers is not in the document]
tac0turtle: added

types/protobuf.go
```diff
@@ -82,12 +77,9 @@ func (tm2pb) ValidatorUpdates(vals *ValidatorSet) []abci.ValidatorUpdate {
...
 // XXX: panics on nil or unknown pubkey type.
 func (tm2pb) NewValidatorUpdate(pubkey crypto.PubKey, power int64) abci.ValidatorUpdate {
-	pubkeyABCI, err := cryptoenc.PubKeyToProto(pubkey)
-	if err != nil {
-		panic(err)
-	}
+
 	return abci.ValidatorUpdate{
-		PubKey: pubkeyABCI,
+		PubKey: pubkey.Address(),
```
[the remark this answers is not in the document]
tac0turtle: added

types/params.go
```diff
@@ -193,6 +197,10 @@ func (params ConsensusParams) ValidateBasic() error {
 		return errors.New("len(Validator.PubKeyTypes) must be greater than 0")
 	}
 
+	if len(params.Validator.PubKeyTypes) > 1 {
```
melekes, line 200: I would merge the 2 conditions together:

```suggestion
	if len(params.Validator.PubKeyTypes) != 1 {
```

types/params.go
melekes: why are we double-checking the length here again?

abci/types/pubkey.go
```diff
@@ -18,8 +18,10 @@ func Ed25519ValidatorUpdate(pk []byte, power int64) ValidatorUpdate {
 
 	return ValidatorUpdate{
 		// Address:
-		PubKey: pkp,
-		Power:  power,
+		PubKey:      pkp,
```
melekes, line 21: When are we going to drop it? Next major version? Should we create an issue?

melekes (approved):
Thanks @tac0turtle ❤️ 

Could you add a changelog entry?

tac0turtle:
any chance this can land in v1 since its additive

itsdevbear:
@melekes bump

types/params.go
```diff
@@ -189,8 +189,8 @@ func (params ConsensusParams) ValidateBasic() error {
 		return fmt.Errorf("ABCI.VoteExtensionsEnableHeight cannot be negative. Got: %d", params.ABCI.VoteExtensionsEnableHeight)
 	}
 
-	if len(params.Validator.PubKeyTypes) == 0 {
-		return errors.New("len(Validator.PubKeyTypes) must be greater than 0")
+	if len(params.Validator.PubKeyTypes) == 1 {
```
melekes, line 192:
```suggestion
	if len(params.Validator.PubKeyTypes) != 1 {
```

.changelog/unreleased/breaking-changes/2399-make-pubkey-interface-optional-in-abci.md
```diff
@@ -0,0 +1,2 @@
+- [abci/types] Allow applications to avoid implementing `PubKey` interface. ([\#2399](https://github.com/cometbft/cometbft/pull/2399))
```
melekes, line 1: this is not breaking

types/params.go
melekes: This should've been caught by the unit test

melekes:
>  In the meantime, I'd like to understand how that change is helpful to you, and whether it would be an option to add that limitation at application (SDK) level.

@tac0turtle would you mind answering ^?

> this was removed because comet only supports a single key type being used at any one time.

where's that enforced exactly in the code? I can't find the place.

tac0turtle:
> > In the meantime, I'd like to understand how that change is helpful to you, and whether it would be an option to add that limitation at application (SDK) level.
> 
> @tac0turtle would you mind answering ^?

we have a switch statement for the proto oneof, so if a new type is added we have to remember to add it there. 

> where's that enforced exactly in the code? I can't find the place.
its not something enforced in code but it is enforced in a network. Comet hasnt done testing with multiple key types in the last 5 years so there is a high chance it doesnt work unless there is a test for it.

sergio-mena:
Thanks @melekes for volunteering.
@tac0turtle, are you OK if we take it from here?

melekes:
Closing in favor of https://github.com/cometbft/cometbft/pull/2843.

> remove multiple keys being set in validator params

While it's true it's been largely untested, we will need to gather more feedback from users before we remove this feature. This is to ensure that nobody uses it now or will need it in the future.

PR #2404, opened 2024-02-21
test(consensus): test enabling PBTS at a given height

### LINKED ISSUE #2393: tests: add unit tests to check the activation of PBTS
There is a test unit `pbts_test.go` only dedicated to PBTS. Those tests, however, assume that PBTS is enabled by default, which is true in unit tests (see https://github.com/cometbft/cometbft/issues/2324). Since we need to support the legacy BFT Time algorithm, we should validate the transition from BFT Time to PBTS.

cason:
Closes #2393.

A bug was found while testing. Consensus was checking whether the `Proposal.Timestamp` matched the corresponding `ProposalBlock.Time` even when PBTS is disabled. This has been fixed.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/internal/consensus/common_test.go
+++ b/internal/consensus/common_test.go
@@ -493,12 +493,12 @@ func randStateWithAppWithHeight(
 
 func randStateWithAppWithBFTTime(nValidators int) (*State, []*validatorStub) {
 	c := test.ConsensusParams()
+	c.Feature.PbtsEnableHeight = 0 // Disable PBTS
 	return randStateWithAppImpl(nValidators, kvstore.NewInMemoryApplication(), c)
 }
 
 func randStateWithApp(nValidators int, app abci.Application) (*State, []*validatorStub) {
 	c := test.ConsensusParams()
-	c.Feature.PbtsEnableHeight = 1
 	return randStateWithAppImpl(nValidators, app, c)
 }
 
@@ -933,6 +933,9 @@ func randGenesisState(
 	numValidators int,
 	consensusParams *types.ConsensusParams,
 ) (sm.State, []types.PrivValidator) {
+	if consensusParams == nil {
+		consensusParams = test.ConsensusParams()
+	}
 	return randGenesisStateWithTime(numValidators, consensusParams, cmttime.Now())
 }
 
--- a/internal/consensus/common_test.go
+++ b/internal/consensus/common_test.go
@@ -611,29 +611,6 @@ func ensureNewTimeout(timeoutCh <-chan cmtpubsub.Message, height int64, round in
 		"Timeout expired while waiting for NewTimeout event")
 }
 
-func ensureNewProposalWithTimeout(proposalCh <-chan cmtpubsub.Message, height int64, round int32, timeout time.Duration) {
-	select {
-	case <-time.After(timeout):
-		panic("Timeout expired while waiting for NewProposal event")
-	case msg := <-proposalCh:
-		proposalEvent, ok := msg.Data().(types.EventDataCompleteProposal)
-		if !ok {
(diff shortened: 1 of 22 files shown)


lasarojc:
Good catch. Thanks for fixing. I still need to delve into the new test, but I can do that on your base branch if you want to merge.

internal/consensus/common_test.go
```diff
@@ -154,6 +157,10 @@ func signVote(vs *validatorStub, voteType types.SignedMsgType, chainID string, b
 	return v
 }
 
+func signVote(vs *validatorStub, voteType types.SignedMsgType, chainID string, blockID types.BlockID, extEnabled bool) *types.Vote {
```
glnro, line 160: I don't understand why this is necessary if it's just a wrapper to avoid calling `vs.clock.Now()`...maybe I'm overlooking something?

lasarojc (approved):

internal/consensus/common_test.go
```diff
@@ -171,6 +178,7 @@ func signVotes(
 func incrementHeight(vss ...*validatorStub) {
 	for _, vs := range vss {
 		vs.Height++
+		vs.Round = 0
```
sergio-mena, line 181: Wow, hard to believe this hadn't been necessary until now

internal/consensus/pbts_test.go
```diff
@@ -531,3 +534,163 @@ func pbtsFromHeightParams(height int64) types.FeatureParams {
...
+		} else {
+			height, round = height, round+1
+			incrementRound(vss[1:]...)
+		}
+	}
+	// Last call to FinalizeBlock
+	ensureNewRound(newRoundCh, height, round)
+}
+
+func createProposalBlock(t *testing.T, cs *State, timestamp time.Time) (*types.Block, *types.PartSet, types.BlockID) {
```
sergio-mena, line 677: Have you looked around to see if we have similar utility functions elsewhere in the consensus UTs that we could reuse? Same for `signProposal` below.

sergio-mena (approved):
Amazing 👏

internal/consensus/common_test.go
cason: This is necessary because there are tens of calls to this method with the original format. All the methods in this section are essentially this: call another method with some defaults.

internal/consensus/common_test.go
```diff
@@ -603,9 +611,9 @@ func ensureNewTimeout(timeoutCh <-chan cmtpubsub.Message, height int64, round in
 		"Timeout expired while waiting for NewTimeout event")
 }
 
-func ensureNewProposal(proposalCh <-chan cmtpubsub.Message, height int64, round int32) {
+func ensureNewProposalWithTimeout(proposalCh <-chan cmtpubsub.Message, height int64, round int32, timeout time.Duration) {
```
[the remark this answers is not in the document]
cason: What a redundancy. Re-used the existing method. However, I've kept the `ensureNewProposal` as it is useful for proposers, when you do not know the `BlockID` in advance.

internal/consensus/pbts_test.go
```diff
@@ -531,3 +534,163 @@ func pbtsFromHeightParams(height int64) types.FeatureParams {
...
+			blockID = rs.Proposal.BlockID
+		} else {
+			var ts time.Time
+			if height >= pbtsSetHeight && height < pbtsEnableHeight {
+				// Use PBTS logic while PBTS is not yet activated
+				ts = time.Now()
+				rejectProposal = true
+			} else if height >= pbtsEnableHeight {
+				// Shift timestamp to the future => not timely
+				ts = time.Now().Add(time.Second)
```
cason, line 605:
```suggestion
				ts = time.Now().Add(2 * c.Synchrony.Precision)
```

PR #2405, opened 2024-02-22
docs(qa): Update instructions on the QA methodology

hvanz:
[Rendered](https://github.com/cometbft/cometbft/blob/hvanz/qa-update-instructions/docs/references/qa/method.md)

The section on Extracting Results depends on new scripts added by https://github.com/cometbft/qa-infra/pull/61.

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):

docs/references/qa/method.md
```diff
@@ -47,54 +64,71 @@ _200 Node Test_, and _Rotating Nodes Test_.
 
 ## 200 Node Testnet
 
+This test consists in spinning up 200 nodes (175 validators + 20 full nodes + 5 seed nodes) and
+performing two of experiments:
```
sergio-mena, line 68:
```suggestion
performing two experiments:
```

docs/references/qa/method.md
```diff
@@ -47,54 +64,71 @@ _200 Node Test_, and _Rotating Nodes Test_.
...
+- First we find the [saturation point](saturation) of the network by running the script
+  [200-node-loadscript.sh][200-node-loadscript.sh].
+- Then we run several times the testnet using the saturation point to collect data.
+
+The script [200-node-loadscript.sh] runs multiple transaction load instances with all possible
+combinations of the following parameters:
+- number of transactions sent per second (the rate): 200, 400, 800, and 1600.
+- number of connections to the target node: 1, 2, and 4.
```
sergio-mena, line 76: We probably can drop 2 and 4 here, and just use 1 connections, as we never saw any impact in this test case when changing the number of connections. @lasarojc what do you think?

sergio-mena (approved):
🙏

docs/references/qa/method.md
hvanz: Totally agree. Most instances with 400 tx/s and above are anyway saturated. We could consider refining the rate values, something like from 200 to 1000, in intervals of 100.

melekes (approved):
👍

PR #2412, opened 2024-02-22
perf(state/store): avoid double-saving ABCI responses

melekes:
Manual backport of #2017


sergio-mena (approved):
According to diff-of-diffs this is good to go

PR #2425, opened 2024-02-24
refactor(protobuf): switch from gogofaster to gocosmos generator

glnro:
There is no difference between the two. `gogofaster` was renamed to `gocosmos` to be a clear distinction from `gogofaster` from the original repo, which behaves differently.

Renaming for consistency.


melekes:
@glnro what's the reason for switching?

cason:
> @glnro what's the reason for switching?

Yes, can you provide some context?

melekes:
"I've checked, turns out there is no difference. However, seems like the update in our gogoproto fork makes them both behave the same. I don't know the historical reason, but then I would guess that it was renamed as gocosmos to be a clear distinction from gogofaster from the original gogoproto, which behaves differently. So for consistency I would suggest to still switch to gocosmos."

melekes (approved):

PR #2431, opened 2024-02-26
feat(pbts): Make synchrony params adaptive

### LINKED ISSUE #2184: PBTS: should synchrony parameters be adaptive?
Original discussion: https://github.com/tendermint/spec/issues/371.

### Summary

The issue here is the possibility that the application configures bad (e.g., very conservative) time bounds, resulting in a chain that will halt as proposals will be rejected by all or most of correct nodes. This situation is hard to deal because the synchronous parameters are consensus parameters, which require committing blocks to be updated. Which might become unfeasible if consensus cannot progress due to bad parameters.

There are two synchronous parameters considered by PBTS: `PRECISION` and `MSGDELAY`. `PRECISION` should not be a problem here, as it restricts malicious nodes from proposing timestamps in the future. The `MSGDELAY` parameter, in its turn, restricts malicious nodes from proposing timestamps "too much in the past". The diagram below can be helpful to understand the use of these parameters to validate proposed timestamps:

By setting a `MSGDELAY` parameter that is unrealistic (too low), a chain might end up in a situation in which correct validators reject proposals because their timestamps are deemed "too much in the past" because the observed propagation delay for proposals is larger than the configured `MSGDELAY` parameter.

### Adaptive `MSGDELAY`

The general idea is that we should increase the `MSGDELAY` parameter adopted by PBTS when we fail to accept proposals in multiple rounds, which may indicate that the value chosen for this parameter was inadequate. 

A possible analogy here is `timeout_propose`, the maximum expected delay for a proposed value (proposal and full block) to reach correct validators. Once a round fails on reaching a decision, the next round considers an increased delay `timeout_propose(round - 1) + timeout_propose_delta`, which is linearly increased until it represents the actual delay for delivering full proposals to correct validators. The same concept could be adopted for the `MSGDELAY` parameter.

The [original issue](https://github.com/tendermint/spec/issues/371) several approaches for adaptive synchronous parameters for PBTS. They include a round from which adaptations start being adopted, how we adapt the values (linearly, exponentially) etc. The implemented version (this [PR](https://github.com/tendermint/tendermint/pull/7739)) adopted an exponential increasing of `MSGDELAY` when more than `10` rounds are needed to decide on a value.

### Open questions

1. Do we need adaptive synchrony parameters?
2. If so, how and when we increase the value of `MSGDELAY`?
3. How we specify/document this approach, making it clear to external components which synchrony parameters were used to validate the timestamp of a given block

### Definition of Done

- [x] Mechanism to relax `timely` verification under certain circumstances is defined
- [x] #2432

Documentation:
- [x] #2452
   - ADR 112 describes the implementation of this mechanism
   - ~[PBTS runbook](https://github.com/cometbft/cometbft/blob/feature/pbts/docs/tools/proposer-based-timestamps-runbook.md) is updated accordingly~
   - [PBTS documentation](https://github.com/cometbft/cometbft/blob/feature/pbts/docs/core/proposer-based-timestamps.md) is updated accordingly
- [x] PBTS spec: #2318

glnro:
Contributes to #2432

---

#### PR checklist

- [x] Tests written/updated


types/proposal_test.go
```diff
@@ -235,23 +238,50 @@ func TestIsTimely(t *testing.T) {
...
+		},
+		{
+			name:           "proposal timely for round",
+			proposalHeight: 2,
+			proposalTime:   genesisTime,
+			recvTime:       genesisTime.Add(2970 * time.Nanosecond),
+			round:          4,
+			precision:      time.Nanosecond * 50,
+			msgDelay:       time.Nanosecond * 2000,
+			expectTimely:   true,
```
sergio-mena, line 268: Are these tests based in a find-grained timing behavior? I'm concerned they'll become flaky when run in heavy-loaded systems (e.g., on GitHub runners)

types/params.go
```diff
@@ -109,6 +110,20 @@ type SynchronyParams struct {
 	MessageDelay time.Duration `json:"message_delay,string"`
 }
 
+/*
+AdaptiveSynchronyParams ensures an exponential backoff for timestamp validation by relaxing MessageDelay by a factor of
```
sergio-mena, line 114:
```suggestion
AdaptiveSynchronyParams ensures an exponential backoff for timestamp validation by increasing MessageDelay by a factor of
```

types/proposal_test.go
cason: There is no time aspect here. It is only comparing two predefined time instants. It could be any time instant.

types/params.go
```diff
@@ -109,6 +110,20 @@ type SynchronyParams struct {
...
+10% each subsequent round a proposal's timeliness is calculated. This is meant to facilitate the progression of
+consensus if bad synchrony parameters are set or become insufficient to preserve liveness.
+
+Where r is the consensus round, MSGDELAY(r) == MSGDELAY(0) * (1.1)^r.
```
cason, line 118:
```suggestion
Where round is the consensus round, MessageDelay(rround) == MessageDelay * (1.1)^round.
```

types/proposal_test.go
```diff
@@ -201,6 +201,7 @@ func TestIsTimely(t *testing.T) {
 		proposalHeight int64
 		proposalTime   time.Time
 		recvTime       time.Time
+		round          int32
```
cason, line 204: Can you update the comment on line 209?

cason (commented):
Hey, thank for this.

I think we need to test this from the consensus point of view. Namely, that a untimely proposal, because it is far in the past, eventually gets accepted once the round number increases.

types/proposal_test.go
sergio-mena: If that's the case, please disregard my comment

cason:
I think this PR may solve #2432.

I've just created this issue to separate the documentation from the implementation.

types/params.go
```diff
@@ -109,6 +110,20 @@ type SynchronyParams struct {
...
+*/
+func AdaptiveSynchronyParams(precision time.Duration, messageDelay time.Duration, round int32) SynchronyParams {
```
cason, line 120: Why we receive the two fields of `SynchronyParams` here, instead of an instance of it?

glnro: Felt cleaner and more efficient to implement a constructor instead of creating synch params and needing to ensure the message delay value had been properly adapted either before or after.

sergio-mena: @glnro I think what @cason means is that you could pass just the original `SynchronyParams` in one simgle param,  so that all callsites of `AdaptiveSynchronyParams` can be simplified. Something like this:
```go
func AdaptiveSynchronyParams(origParams SynchronyParams, round int32) SynchronyParams {
...
}
```
And then, inside the function, you can still keep the constructor as you have it now.

glnro: Actually I disagree and the more I think about it, the more I think we should not be able to create synchrony params and pass them around without ensuring that the round is accounted for. Wdyt about making `synchronyParams` unexported at the package level and only having a public constructor that additionally takes a round.

sergio-mena: Fair enough. Then I think the best solution is what you have at present.
@cason wdyt?

cason (approved):
Lets merge this.

Pending comments to be addressed were added as a task of #2432.

PR #2434, opened 2024-02-26
fix(jsonrpc): enable HTTP basic auth in WS client

MattKetmo:
Hello,

I noticed that the JSON-RPC client is compatible with URLs containing **HTTP basic auth**, but not on the WebSocket side.

For instance:

```golang
client, err := http.New("https://foo:bar@example.org:443", "/websocket")

client.Block(ctx, nil) // works

client.Start() // websocket error: the client doesn't forward auth headers
```

The HTTP client handle the username/password from the given endpoint correctly:
- https://github.com/cometbft/cometbft/blob/v0.38.5/rpc/jsonrpc/client/http_json_client.go#L140-L141
- https://github.com/cometbft/cometbft/blob/v0.38.5/rpc/jsonrpc/client/http_json_client.go#L184-L185

So this PR brings the same logic for the WS client.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/.changelog/unreleased/improvements/2434-jsonrpc-websocket-basic-auth.md
+++ b/.changelog/unreleased/improvements/2434-jsonrpc-websocket-basic-auth.md
@@ -1,0 +1,1 @@
+- `[jsonrpc]` enable HTTP basic auth in websocket client ([#2434](https://github.com/cometbft/cometbft/pull/2434))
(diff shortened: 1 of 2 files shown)


melekes (approved):
Thanks @MattKetmo ❤️ 

1. Please base this PR against the `main` branch?
2. Please add a changelog entry 🙏

MattKetmo:
@melekes done: PR based on `main` + changelog entry added (in the "improvements" section)

Issue #2432, opened 2024-02-26
PBTS: Implement adaptive synchronous parameters for consensus


cason:
Derives from the decision in  #2184 to make the `SynchronyParams.MessageDelay` PBTS parameter adaptive.

The defined mechanism is:

1. `Precision` parameter remains unchanged over consensus rounds
2. `MessageDelay` parameter increases over consensus rounds, with the following logic:

```
MessageDelay(round) = MessageDelay * (1.1)^round
```

### Definition of Done

- [x] Implement logic for `MessageDelay` to increase over rounds and associated testing
  - #2431
- [x] Adopt this this logic on the consensus implementation
  - #2431
- [x] Test that a `Proposal` with timestamps far in the past is eventually decided in a height of consensus, provided that multiple rounds are executed
  - #2447
  - Merge into #2431 branch
- [x] Address comments on #2431 
   - #2490

PR #2446, opened 2024-02-26
test(consensus): test if untimely proposal is decided after multiple rounds

cason:
This test should fail with the current implementation in `feature/pbts`.

This test should pass when changes from  #2432 are implemented.

This is a proof of concept and should not be merged.

---

#### PR checklist

- [ ] Tests written/updated


sergio-mena:
From @cason : _The test unit added there is present in feature/pbts and passes._

PR #2467, opened 2024-02-28
feat(blocksync)!: set the max number of (concurrently) downloaded blocks to {peersCount * 20}

### LINKED ISSUE #1283: Fast sync gets stuck if there's not many peers because `requestRoutine` in pool requests blocks out of order
Fast sync gets stuck (or proceeds very slowly) because `requestRoutine` in pool requests blocks in a different order because requesters get made asynchronously by `makeNextRequester` (https://github.com/tendermint/tendermint/blob/main/blockchain/v0/pool.go#L375C24-L375C41). This causes fast sync to get stuck if there's not many peers, because `maxPendingRequestsPerPeer` is 20.

Sample debugging logs from manually added debugging prints:
```
debug0: making next requester for  256513
debug0: making next requester for  256514
debug0: making next requester for  256515
debug0: making next requester for  256516
debug0: making next requester for  256517
debug0: making next requester for  256518
debug0: making next requester for  256519
debug0: making next requester for  256520
debug0: making next requester for  256521
debug0: making next requester for  256522
debug0: making next requester for  256523
debug0: making next requester for  256524
debug0: making next requester for  256525
debug0: making next requester for  256526
debug0: making next requester for  256527
debug0: making next requester for  256528
debug0: making next requester for  256529
debug0: bpr making request for  256529
debug0: bpr making request for  256521
debug0: bpr making request for  256514
debug0: bpr making request for  256516
debug0: bpr making request for  256517
debug0: bpr making request for  256515
debug0: bpr making request for  256518
debug0: bpr making request for  256527
debug0: bpr making request for  256519
debug0: bpr making request for  256520
debug0: bpr making request for  256522
debug0: bpr making request for  256513
debug0: bpr making request for  256523
debug0: bpr making request for  256524
debug0: bpr making request for  256525
debug0: bpr making request for  256526
```

Increasing `maxPendingRequestsPerPeer` to 600 fixes the problem (though obviously not a proper fix). edit: still having issues with increasing it to that param, unless I also decrease maxTotalRequesters significantly

### LINKED ISSUE #2465: blocksync: respect the `maxDiffBetweenCurrentAndReceivedBlockHeight` limit
## Report

Osmosis team: "My understanding of this value is that this implies we should only be fetching blocks 100 blocks in advance from where we currently are. I added prints and saw that even when this value was set as low as 10, we were still fetching blocks 2000+ blocks out while being starved of blocks 2 away from where we currently were

And I might be wrong about this second part, but looking at where this value is used https://github.com/osmosis-labs/cometbft/blob/490dd752eb8a1a934542f70fdc8960594e06973b/blocksync/pool.go#L264-L268, it seems like because we are fetching blocks very far out, we just error and disregard them, so its completely wasted cycles AFAICT."

## Preliminary discussion

@jmalicevic: "it seems that maxDiffBetweenCurrentAndReceivedBlockHeight is only checked when we receive blocks, it should probably also be checked when we request blocks. Otherwise it is not the peers fault we get blocks too far ahead. Perhaps avoid creating requesters too far ahead in makeRequestersRoutine() by [adapting when](https://github.com/osmosis-labs/cometbft/blob/62a97f2c9713c716ef203052eb32069f917737f9/internal/blocksync/pool.go#L126) then pool.makeNextRequester is called could help."

@cason: "Yes, but only for blocks for which we don't have a tracked request.
It is still to be understood if we don't have a tracked request because it is too "old" (i.e., we already committed this block) or too "new" (i.e.,  too far in the future). I still can't see how the latter could occur, as: (i) a correct peer only sends us a block upon receiving our request; (ii) we track all requests we send."

@jmalicevic: "So effectively this variable is never checked unless it is an error (we don’t have a requester for it ) or we removed it at some point. But we are never enforcing not requesting blocks too far ahead with this variable it seems"

melekes:
where 20 is the new max requests per peer.

Previously, the number of requesters (and consequently the number of requested blocks) was 600 and did not depend on the number of peers. There's no point in creating more than that since N of requests per peer is capped and we know the number of peers.

Closes #2465
Fixes #1283
[api-breaking]

---

#### PR checklist

- [x] Tests written/updated


czarcas7ic:
Tested and appears to respect the limit, thank you!

As mentioned in slack thread, I think this would be good to either slightly bump up or (more preferable) make this a config value that the user can set.

Once we fix the peer starving issue we will be able to determine what the most ideal number for this is. Right now its hard to determine given the consistent pauses while syncing, so benchmarking is slightly random

melekes:
This PR won't be backported to `v0.38` since it's API breaking.

melekes:
Fixes https://github.com/cometbft/cometbft/issues/1283

internal/blocksync/pool.go
```diff
@@ -154,25 +147,16 @@ func (pool *BlockPool) removeTimedoutPeers() {
...
-	pool.mtx.Lock()
-	defer pool.mtx.Unlock()
-
-	return pool.height, atomic.LoadInt32(&pool.numPending), len(pool.requesters)
-}
-
 // IsCaughtUp returns true if this node is caught up, false - otherwise.
 // TODO: relax conditions, prevent abuse.
-func (pool *BlockPool) IsCaughtUp() bool {
+func (pool *BlockPool) IsCaughtUp() (isCaughtUp bool, height, maxPeerHeight int64) {
```
hvanz, line 152: `height` and `maxPeerHeight` are only returned for logging messages. An alternative is to change their types from `int64` to `atomic.Int64` to access them safely when needed.

hvanz (approved):
LGTM, I have just a comment about the return type of `IsCaughtUp`.

internal/blocksync/pool.go
```diff
@@ -273,7 +257,8 @@ func (pool *BlockPool) AddBlock(peerID p2p.ID, block *types.Block, extCommit *ty
 		if diff < 0 {
 			diff *= -1
 		}
-		if diff > maxDiffBetweenCurrentAndReceivedBlockHeight {
+		const maxDiff = 100 // maximum difference between current and received block height
```
cason, line 260: Why is this constant?

internal/blocksync/pool.go
```diff
@@ -387,30 +371,18 @@ func (pool *BlockPool) pickIncrAvailablePeer(height int64) *bpPeer {
...
-	}
-
 	request := newBPRequester(pool, nextHeight)
-
 	pool.requesters[nextHeight] = request
-	atomic.AddInt32(&pool.numPending, 1)
 
-	err := request.Start()
-	if err != nil {
+	if err := request.Start(); err != nil {
```
cason, line 381: Not sure if we need to that while holding the mutex.

internal/blocksync/reactor.go
```diff
@@ -394,14 +395,13 @@ FOR_LOOP:
 			if missingExtension {
 				bcR.Logger.Info(
 					"no extended commit yet",
-					"height", height,
```
cason, line 397: I thinks this information is kind of useful.

internal/blocksync/reactor.go
```diff
@@ -537,9 +537,9 @@ FOR_LOOP:
 			blocksSynced++
 
 			if blocksSynced%100 == 0 {
+				_, height, maxPeerHeight := bcR.pool.IsCaughtUp()
 				lastRate = 0.9*lastRate + 0.1*(100/time.Since(lastHundred).Seconds())
```
cason, line 541: Do you know why we use this window mechanism?

internal/blocksync/pool.go
cason: I think this is a snapshot from the pool state. Namely, one should expect the information to sometimes be outdated. I don't see much point of creating another point of contention in order to return an atomic version of these counters.

internal/blocksync/pool.go
```diff
@@ -104,28 +96,29 @@ func (pool *BlockPool) OnStart() error {
...
 	for {
 		if !pool.IsRunning() {
-			break
+			return
 		}
 
-		_, numPending, lenRequesters := pool.GetStatus()
+		pool.mtx.Lock()
+		var (
+			maxRequestersCreated = len(pool.requesters) >= len(pool.peers)*maxPendingRequestsPerPeer
```
cason, line 107: So the core change of this PR is this one, am I right?

cason (approved):
Legit.

I don't understand why this would be API breaking. Are the changes in the `IsCaughtUp()` method? If so, we could have another method to gather the additional two values that it is returning now. This would make this change not API breaking.

I wonder whether the PR description could include some indications related to the performance benefits of this approach, assume there are some available.

internal/blocksync/pool.go
melekes: Correct

internal/blocksync/pool.go
melekes: I've removed it in the other PR. Please ignore.

internal/blocksync/reactor.go
cason: I was meaning the 90% of the last rate + 10% of the current. This is a form of moving window, I don't remember the precise name of this method. Anyway, just a comment.

melekes:
> I don't understand why this would be API breaking. Are the changes in the IsCaughtUp() method? If so, we could have another method to gather the additional two values that it is returning now. This would make this change not API breaking.

Yes, because of the changes to `IsCaughtUp`. Now I understand that I should have created two separate PRs, but it was easier to make changes in a single PR back then. Anyway, now that https://github.com/cometbft/cometbft/pull/2515 exists, I don't see a need for splitting the work.

melekes:
> I wonder whether the PR description could include some indications related to the performance benefits of this approach, assume there are some available.

This PR doesn't increase or decrease performance. It merely fixes the bug reported in https://github.com/cometbft/cometbft/issues/1283

PR #2469, opened 2024-02-28
docs(README): remove Terra

JekaMas:
It removes Terra from the readme file


melekes (approved):
Thanks @JekaMas

JekaMas:
Sorry, for the tiny PR. Was reading latest changes and comparing with original paper and found that line about Terra. I believe your project is much better than that thing from the past.

PR #2483, opened 2024-02-29
ci: check metrics generation in CI checks

### LINKED ISSUE #2482: CI: make sure that `make metrics` is run when metrics are updated
`make metrics` uses the program in `scripts/metricsgens` to produce `metrics.gen.go` files from `metric.go` files. If the companion `metrics.gen.go` is not produced or updated, the new or updated metrics on  `metrics.go` have no effect.

We should thus have a CI check to make sure that `metrics.gen.go` files are updated when `metric.go` files are changed.

sergio-mena:
Closes #2482

Also, spotted and fixed a bug in `.github/workflows/check-generated.yml`: if the changes in the generated code consist only in new files added, `git diff --stat --exit-code` doesn't detect them, and the check passes while it should fail. Check the CI checks in the commit history of this PR for further info.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/.github/workflows/check-generated.yml
+++ b/.github/workflows/check-generated.yml
@@ -13,7 +13,7 @@ permissions:
   contents: read
 
 jobs:
-  check-mocks:
+  check-mocks-metrics:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/setup-go@v5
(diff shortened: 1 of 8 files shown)


cason (approved):
Thank you for that :)

PR #2489, opened 2024-03-01
feat(types): improve Proposal and SynchronyParams validation tests

cason:
Contributes to #1731

---

#### PR checklist

- [x] Tests written/updated


types/canonical.go
```diff
@@ -42,9 +42,9 @@ func CanonicalizePartSetHeader(psh cmtproto.PartSetHeader) cmtproto.CanonicalPar
 func CanonicalizeProposal(chainID string, proposal *cmtproto.Proposal) cmtproto.CanonicalProposal {
 	return cmtproto.CanonicalProposal{
 		Type:      ProposalType,
-		Height:    proposal.Height,       // encoded as sfixed64
-		Round:     int64(proposal.Round), // encoded as sfixed64
-		POLRound:  int64(proposal.PolRound),
+		Height:    proposal.Height,          // encoded as sfixed64
+		Round:     int64(proposal.Round),    // encoded as sfixed64
+		POLRound:  int64(proposal.PolRound), // FIXME: not matching
```
sergio-mena, line 47: ???

types/params.go
```diff
@@ -247,11 +251,11 @@ func (params ConsensusParams) ValidateBasic() error {
 		return fmt.Errorf("Feature.PbtsEnableHeight cannot be negative. Got: %d", params.Feature.PbtsEnableHeight)
 	}
 
+	// FIXME: should we enforce something more than > 0? Maybe >= time.Millisecond?
```
sergio-mena, line 254: I would be in favor of that, for practical reasons. @lasarojc ? @glnro ?

types/params_test.go
```diff
@@ -209,27 +209,27 @@ func TestConsensusParamsValidation(t *testing.T) {
...
+		// test invalid synchrony params
 		{
-			name: "messageDelay -1",
+			name: "messageDelay 0",
 			params: makeParams(makeParamsArgs{
 				evidenceAge:  2,
-				precision:    1,
-				messageDelay: -1,
+				precision:    time.Nanosecond,
+				messageDelay: 0,
```
sergio-mena, line 223: Why this change?

types/params_test.go
```diff
@@ -209,27 +209,27 @@ func TestConsensusParamsValidation(t *testing.T) {
...
 			valid: false,
 		},
 		{
-			name: "precision -1",
+			name: "precision 0",
 			params: makeParams(makeParamsArgs{
 				evidenceAge:  2,
-				precision:    -1,
-				messageDelay: 1,
+				precision:    0,
```
sergio-mena, line 231: And this one?

internal/consensus/msgs_test.go
```diff
@@ -50,10 +50,10 @@ func TestMsgToProto(t *testing.T) {
 	proposal := types.Proposal{
 		Type:      types.ProposalType,
 		Height:    1,
-		Round:     1,
+		Round:     2,
```
lasarojc, line 53: Any particular reason for this change, or just shaking things up?

types/params.go
lasarojc: I don't see the reason. If it is large, may be too large. And if it is small, like `time.Millisecond`, then it is just as good as checking `<= 0` imho.
Unless having an increment of less than 1ms could problematic, which I don't think it is since timestamps have ns resolution.

types/params_test.go
lasarojc: Anything <= 0 is invalid, so this is the largest value that already triggers the failure.

cason: 0 should be invalid

types/params_test.go
cason: 0 should be invalid

internal/consensus/msgs_test.go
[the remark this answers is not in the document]
cason: POLRound < Round.

types/canonical.go
cason: I can remove, but the proto name does not match the type name.

types/params_test.go
```diff
@@ -252,8 +252,8 @@ func TestConsensusParamsValidation(t *testing.T) {
 				makeParamsArgs{
 					blockBytes:   1,
 					evidenceAge:  2,
-					precision:    1,
-					messageDelay: 1,
+					precision:    time.Nanosecond,
+					messageDelay: time.Nanosecond,
 					pbtsHeight:   0,
 				}),
 			valid: true,
```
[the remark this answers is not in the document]
cason: What is the difference?

types/params_test.go
```diff
@@ -654,3 +654,36 @@ func TestProtoUpgrade(t *testing.T) {
...
+	lastSP := originalSP
+	for round := int32(1); round <= 10; round++ {
+		adaptedSP := adaptiveSP(originalSP, round)
+		assert.NotEqual(t, adaptedSP, lastSP)
+		assert.Equal(t, adaptedSP.Precision, lastSP.Precision,
+			"Precision must not change over rounds")
+		assert.Greater(t, adaptedSP.MessageDelay, lastSP.MessageDelay,
+			"MessageDelay must increase over rounds")
+
+		// It should not increase a lot per round, say more than 25%
```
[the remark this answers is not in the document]
cason: It can be. I didn't want to be super precise here, we only need it to increase.

types/canonical.go
sergio-mena: You mean `POLRound` vs `PolRound` ?

cason: Yep

types/params.go
sergio-mena: OK, so let's keep it simple. Let's leave the `>0`

cason: Up to you.

types/params_test.go
sergio-mena: Then, can you test both, 0 and -1? (i.e., add an extra TC here)

types/params_test.go
sergio-mena: Same, let's test both, in case someone, 3 years from now, decides it should be `<0` in production

types/params_test.go
cason: Ok.

types/params.go
cason: So remove the FIXME?

types/params_test.go
sergio-mena: 🙏

types/canonical.go
sergio-mena: Added a bullet point in #1731

lasarojc (commented):
> What is the difference? 

PBTS is enabled

lasarojc (approved):

sergio-mena (approved):

PR #2491, opened 2024-03-01
Update proposer-based-timestamps-runbook.md

lasarojc:
Includes information for `timesync` daemon usage.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/docs/guides/tools/proposer-based-timestamps-runbook.md
+++ b/docs/guides/tools/proposer-based-timestamps-runbook.md
@@ -106,7 +106,15 @@ Re-run the `timedatectl` command and verify that the change has taken effect.
 
 #### Check if Your NTP Daemon is Synchronized
 
-Check the status of your local `chrony` NTP daemon using by running the following:
+We provide two examples here, for `chrony` and `timesync`, but these steps
+should be adapted if you are using a different daemon.
+
+If you find that the NTP is not synchronizing, remember allow NTP traffic
+(123/UDP) to your NTP servers.
+
+##### `chrony` 
+
+Check the status of your `chrony` deamon by running the following command:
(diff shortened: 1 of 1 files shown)


docs/guides/tools/proposer-based-timestamps-runbook.md
```diff
@@ -106,7 +106,15 @@ Re-run the `timedatectl` command and verify that the change has taken effect.
 
 #### Check if Your NTP Daemon is Synchronized
 
-Check the status of your local `chrony` NTP daemon using by running the following:
+We provide two examples here, for `chrony` and `timesync`, but these steps
+should be adapted if you are using a different daemon.
+
+If you find that the NTP is not synchronizing, remember allow NTP traffic
```
cason, line 112:
```suggestion
If you find that the NTP is not synchronizing, remember allow NTP inbound traffic
```

cason (approved):
Good.

In line 87-88, it mentions `timedatectl` and `chrony`. I assume we should add now `timesync`.

docs/guides/tools/proposer-based-timestamps-runbook.md
lasarojc: Not sure how to phrase this. The traffic is initiated from inside, so I should say outbound traffic. But we need to allow the returning traffic as well, it cannot be just outbound. Cannot even say "allow connecting to NTP servers" since it uses UDP. I think I will just leave it this way.

lasarojc: Yup, but since that's the nature of NTP, I will just leave it like this and let net admins decide how they implement it.

PR #2493, opened 2024-03-01
fix(pbts): block Time must be Canonical

cason:
Motivated by [this comment](https://github.com/cometbft/cometbft/pull/2205#discussion_r1491321877).

---

#### PR checklist

- [x] Tests written/updated


internal/state/validation.go
```diff
@@ -112,6 +113,10 @@ func validateBlock(state State, block *types.Block) error {
 	}
 
 	// Validate block Time
+	if block.Time != cmttime.Canonical(block.Time) {
```
sergio-mena, line 116: Is this check a sanity check?

sergio-mena (approved):
🚀

internal/state/validation.go
cason: Yes, because of the "stripping", removing the monotonic part.

sergio-mena:
@cason I found a solution. Check the two upcoming commits

internal/state/helpers_test.go
```diff
@@ -259,3 +260,7 @@ func (app *testApp) ProcessProposal(
 	}
 	return &abci.ProcessProposalResponse{Status: abci.PROCESS_PROPOSAL_STATUS_ACCEPT}, nil
 }
+
+func makeStateNilParams(nVals, height int, chainID string) (sm.State, dbm.DB, map[string]types.PrivValidator) {
```
cason, line 264: Should it be `DefaultParams`?

internal/state/state.go
```diff
@@ -331,3 +333,46 @@ func MakeGenesisState(genDoc *types.GenesisDoc) (State, error) {
 		AppHash: genDoc.AppHash,
 	}, nil
 }
+
+// This is a testing function. Had to be placed here to avoid import cycles.
+func MakeState(nVals, height int, params *types.ConsensusParams, chainID string) (State, dbm.DB, map[string]types.PrivValidator) {
```
cason, line 338: Not way to put this on a `_test.go` file? The problem is importing, right?

cason (commented):
This works but it is not really nice...

internal/consensus/common_test.go
```diff
@@ -1057,26 +1056,14 @@ func signDataIsEqual(v1 *types.Vote, v2 *cmtproto.Vote) bool {
...
-		valAddr := pk.PubKey().Address()
-		vals[i] = types.GenesisValidator{
-			Address: valAddr,
-			PubKey:  pk.PubKey(),
-			Power:   1000,
-			Name:    fmt.Sprintf("test%d", i),
-		}
-		privVals[valAddr.String()] = types.NewMockPVWithParams(pk, false, false)
-	}
+func makeStateNilParams(nVals int, chainID string) (sm.State, map[string]types.PrivValidator) {
```
lasarojc, line 1059: To reduce the changes, call this one `makeState` and teh other one `makeStateWithCP`.

lasarojc (approved):
Just a suggestion to reduce the amount of changes.

internal/consensus/common_test.go
cason: Yes, it is possible. I didn't want to undo all renaming Sergio did, but your comment makes sense.

internal/consensus/wal_test.go
```diff
@@ -360,6 +362,28 @@ func TestWALPeriodicSync(t *testing.T) {
 	}
 }
 
+// FIXME: this helper is very similar to the one in internal/state/helpers_test.go.
```
cason, line 365: Moved it here, since it is not used anywhere else.

sergio-mena (approved):
👍

PR #2494, opened 2024-03-01
merge(pbts): merge branch `main` into `feature/pbts`

sergio-mena:
Part of #1731

N.B.: This is a merge commit bringing all recent changes on `main` into feature branch `feature/pbts`, so no detailed review is needed, as the changes have already been reviewed on `main`.

---

#### PR checklist

- [ ] Tests written/updated
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments
- [ ] Title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) spec

PR #2495, opened 2024-03-02
fix: lint errors introduced with the latest merge

sergio-mena:
Contributes to #1731

After some troubleshooting, I realized that, for two `// nolint` added at the beginning of files in #2494, the `gofumpt` and the `golangci-lint-full` have contradictory rules. So the first one removes the space between `//` and `nolint` and the second adds it back 🤯. The problem has been to _find_ this, as, at the end no files where modified whereas both linters were saying they were modifying some files.

To work around this problem, I changed the way to fix the original lint issues.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/test/e2e/pkg/testnet.go
+++ b/test/e2e/pkg/testnet.go
@@ -1,4 +1,3 @@
-// nolint: goconst
 package e2e
 
 import (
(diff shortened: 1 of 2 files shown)


greg-szabo (approved):
Heroic work!

Issue #2443, opened 2024-02-26
mempool: CListMempool.Flush invokes a read-lock meanwhile it performs write/mutating operations; inadequately tested because there will be data races too


odeke-em:
## Bug Report
Noticed in a code audit that this code is erroneously invoking a read-lock meanwhile it performs write/mutating operations

https://github.com/cometbft/cometbft/blob/4ce0277b35f31985bbf2c25d3806a184a4510010/mempool/clist_mempool.go#L221-L229

which also means that this mempool implementation is not properly tested because it clearly invokes mutating code here https://github.com/cometbft/cometbft/blob/4ce0277b35f31985bbf2c25d3806a184a4510010/mempool/clist_mempool.go#L136-L147

hoanguyenkh:
thank you, i have a PR here -> https://github.com/cometbft/cometbft/pull/2497

PR #2497, opened 2024-03-02
refactor(mempool): Use a full lock instead of a read lock

### LINKED ISSUE #2444: mempool: CListMempool.CheckTx misuses read-lock to protect write operations; possible data race
## Bug Report
Noticed in an audit that there is a misuse of read-locks around mempool/CListMempool for example in #2443 and in particular for this bug report let's look at this code 
https://github.com/cometbft/cometbft/blob/4ce0277b35f31985bbf2c25d3806a184a4510010/mempool/clist_mempool.go#L251C12-L251-L254

but meanwhile in that code protected, we can see this invocation https://github.com/cometbft/cometbft/blob/4ce0277b35f31985bbf2c25d3806a184a4510010/mempool/clist_mempool.go#L280 and even later on
https://github.com/cometbft/cometbft/blob/4ce0277b35f31985bbf2c25d3806a184a4510010/mempool/clist_mempool.go#L290

because we aren't sure how guarantees that .CheckTxAsync won't be mutating.

hoanguyenkh:
Closes #2444, #2443 

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -217,6 +217,7 @@ func (mem *CListMempool) FlushAppConn() error {
 	return nil
 }
 
+// XXX: Unsafe! Calling Flush may leave mempool in inconsistent state.
 func (mem *CListMempool) Flush() {
 	mem.updateMtx.Lock()
 	defer mem.updateMtx.Unlock()
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -217,10 +217,9 @@ func (mem *CListMempool) FlushAppConn() error {
 	return nil
 }
 
-// XXX: Unsafe! Calling Flush may leave mempool in inconsistent state.
 func (mem *CListMempool) Flush() {
-	mem.updateMtx.RLock()
-	defer mem.updateMtx.RUnlock()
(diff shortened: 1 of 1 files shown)


mempool/clist_mempool.go
```diff
@@ -219,8 +219,8 @@ func (mem *CListMempool) FlushAppConn() error {
 
 // XXX: Unsafe! Calling Flush may leave mempool in inconsistent state.
 func (mem *CListMempool) Flush() {
-	mem.updateMtx.RLock()
-	defer mem.updateMtx.RUnlock()
+	mem.updateMtx.Lock()
```
melekes, line 222: // Exclusive mutex for Update method to prevent concurrent execution of
	// CheckTx or ReapMaxBytesMaxGas(ReapMaxTxs) methods.
	updateMtx cmtsync.RWMutex

melekes (commented):
The usage of mutex is explained in the comment.

mempool/clist_mempool.go
hoanguyenkh: sorry, i mistake, it is not issues when use mempool

PR #2498, opened 2024-03-02
fix(mempool): converting to uint64 before additions to avoid overflows

### LINKED ISSUE #2445: mempool: CListMempool.isFull could benefit from bullet-proofing against any integer overflows by converting to uint64 before additions
## Bug Report
If we examine this code https://github.com/cometbft/cometbft/blob/4ce0277b35f31985bbf2c25d3806a184a4510010/mempool/clist_mempool.go#L393

we see this code `int64(txSize)+txsBytes > mem.config.MaxTxsBytes`

and if at least txsBytes (total) is close to maxInt64 given that we don't have bounds guarantees or proper size/length calculations, any txSize will cause an overflow e.g.

```go
txSize=int64(100) + txsBytes=9223372036854775800 -> -9223372036854775806
```

## Suggestion
Let's take advantage of a cast to uint64 of which it would take a whole lot more of filling up the mempool to cause that overflow given that len(max) aka max(int*(64)) on 64-bit machines is maxInt64 which would firstly cause a RAM blow-up long before one can exceed maxUint64

`uint64(txSize)+uint64(txsBytes) > uint64(mem.config.MaxTxsBytes)`

minhngoc274:
Closes #2445

---

#### PR checklist

- [ ] Tests written/updated


lasarojc (approved):

mempool/clist_mempool.go
```diff
@@ -390,7 +390,7 @@ func (mem *CListMempool) isFull(txSize int) error {
 		txsBytes = mem.SizeBytes()
 	)
 
-	if memSize >= mem.config.Size || int64(txSize)+txsBytes > mem.config.MaxTxsBytes {
+	if memSize >= mem.config.Size || uint64(txSize)+uint64(txsBytes) > uint64(mem.config.MaxTxsBytes) {
```
melekes, line 393: Can we make these variables `uint64` to avoid casting?

PR #2512, opened 2024-03-04
feat(config): add a flag to skip state and block syncs

### LINKED ISSUE #1679: config: add a flag to skip blocksync if num(peers)=0 and num(vals) > 0
## Bug Report

Many times, we create state exported testnets based on mainnet state. Usually, these networks don't start unless fast sync is set to true, at which point they start. Before knowing this hack, we used "kicker nodes", or nodes we simply peered to the validator to get the network to start. When moving over to comet BFT, I found that the fast sync hack no longer works, and "kicker nodes" are once again required. Removing the following lines prevents the need for a kicker node or the fast sync hack https://github.com/cometbft/cometbft/blob/6302491554b2fa895091aeb99c448d7fc4ca0781/internal/blocksync/pool.go#L172-L176

I am not sure if just removing those lines is the best way forward, I am just pointing out that there is once again large overhead in creating these state exported testnets, as multiple nodes are once again required. Please let me know if any further information is needed.

melekes:
It's useful in testnets when a node is not a single validator, but the operators still want the nodes to proceed to consensus even if there's no guarantees that the node is caught up to the tip of the network.

Config:

```
testnet_skip_sync = false
```

Closes #1679 

---

#### PR checklist

- [ ] Tests written/updated


jmalicevic:
I think for testing it is fine but note that the last I know vote extensions require at least one blocksync iteration to be able start a node. But in the case this node is the only validator it might not be needed. Maybe best to verify with @sergio-mena

node/node.go
```diff
@@ -414,7 +424,7 @@ func NewNode(ctx context.Context,
 		}
 	}
 	// Don't start block sync if we're doing a state sync first.
```
hvanz, line 426: Maybe move this comment to where the `!stateSync` condition was moved?

hvanz (approved):
LGTM

cason:
We could target this change to `v0.34.x` and `v0.37.x`, where block sync is not mandatory (because we don't have vote extensions).

Or, if we keep this for `v0.38.x` and `main`, only enable this flag when vote extensions are disabled.

melekes:
What could happen if vote extensions are enabled but we skip blocksync?

sergio-mena:
> What could happen if vote extensions are enabled but we skip blocksync?

We (may) crash when we switch to consensus, as we wouldn't have the needed vote extensions (they are not available in the blockchain itself, they need to be retrieved via consensus or blocksync messages)

sergio-mena:
Removing the requirement to blocksync at least one block would mean a full redesign of vote extensions

melekes:
@p0mvn @sampocs @czarcas7ic could you please test this before we merge it? thanks 🙏

melekes:
Closing in favor of https://github.com/cometbft/cometbft/issues/2610

PR #2519, opened 2024-03-05
refactor(privval): rename sign_extension to skip_sign_extension

melekes:
because the former is not backwards compatible. I.e., if the remote signer runs a new version for old chain and `sign_extension` field is absent, then it will be false by default, BUT the expected behavior is different. For this reason, the field is renamed to skip_sign_extension. The remote signer might! skip signing extension bytes, but the default behavior is to sign.

Thanks to @mzabaluev for spotting this.

---

#### PR checklist

- [x] Tests written/updated


sergio-mena (approved):
👍 🙏

PR #2475, opened 2024-02-29
feat(blocksync): sort peers by download rate & multiple requests for closer blocks

melekes:
It can be reviewed commit by commit.

1. Request the block N from peer B immediately after getting `NoBlockResponse` from peer A https://github.com/cometbft/cometbft/pull/2475/commits/a718fb47913d58322e0af27167c76f5a2978be20
2. Sort peers by download rate (the fastest peer is picked first) https://github.com/cometbft/cometbft/pull/2475/commits/cf19851c7b51510e3b406c151139d40ad2e02a66
3. Request a block from peer B if we are approaching pool's height (less than 50 blocks) and the current peer A is slow in sending us the block https://github.com/cometbft/cometbft/pull/2475/commits/e06ebfe60bc162b8def8d0958a098b9d45a860f7 https://github.com/cometbft/cometbft/pull/2475/commits/e97007fd691fde41b2f362703ebe03607025fbc8

Closes #2379 

### Benchmarks

Osmosis:

without this PR (baseline): 11.45m
with this PR: 7.29m
~ 36% decrease

---

#### PR checklist

- [x] Tests written/updated
- [x] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [x] Updated relevant documentation (`docs/` or `spec/`) and code comments
- [x] Title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) spec

PR #2532, opened 2024-03-06
chore: blocksync improv settings

czarcas7ic:
This implements a peerConnWait time that the pool routine waits before sending requests. We were previously always taking 45 seconds to start, but this was because when we started sending requests for the first x blocks, we only had a pool of peers of size 1, causing wait times to occur.

I have tested this on multiple benchmarks and there is no longer a wait time.

I have also increased the minBlocksForSingleRequest to 50 and decreased maxPendingRequestsPerPeer to 10. These values have behaved more reliably than the current settings.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/internal/blocksync/pool.go
+++ b/internal/blocksync/pool.go
@@ -32,7 +32,7 @@ const (
 	requestIntervalMS         = 2
 	maxTotalRequesters        = 600
 	maxPendingRequests        = maxTotalRequesters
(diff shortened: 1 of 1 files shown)


melekes:
ported over to https://github.com/cometbft/cometbft/pull/2475. thanks ❤️

PR #2534, opened 2024-03-07
feat(crypto): adding `secp256k1` with Ethereum format

sergio-mena:
---

Added support for the secp256k1-eth crypto codec.
This includes functionality for generating and verifying
Ethereum-compatible signatures using the secp256k1 curve.

---
#### PR checklist

- [x] Tests written/updated
- [x] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [x] Updated relevant documentation (`docs/` or `spec/`) and code comments
- [x] Title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) spec

PR #2538, opened 2024-03-07
feat: bls signatures

tac0turtle:
This pr adds bls layout and pulls in a bls lib into comet. Opening this pr after discussions in slack with the comet team. 

This pr does not implement batch nor aggregation. 

If there is a want to separate the bls curve addition from the addition of it to the params and pub key proto type happy to do that as well

---

#### PR checklist

- [ ] Tests written/updated


crypto/bls/blst/doc.go
```diff
@@ -0,0 +1,6 @@
+// Package blst implements a go-wrapper around a library implementing the
```
ValarDragon, line 1: I suggest this package be `crypto/bls12381/blst` personally

itsdevbear: +1 agreed better to be specific

tac0turtle: done

sergio-mena:
You may want to add (something similar to) [this change](https://github.com/cometbft/cometbft/pull/2534/files#r1517397293) in order to test the whole node

tac0turtle:
> You may want to add (something similar to) [this change](https://github.com/cometbft/cometbft/pull/2534/files#r1517397293) in order to test the whole node

left a comment, but the change to secret connection should be reverted in that pr. It doesnt effect consensus signing

crypto/bls/bls12381/secret_key.go
```diff
@@ -0,0 +1,75 @@
...
+
+	blst "github.com/supranational/blst/bindings/go"
+)
+
+// bls12SecretKey used in the BLS signature scheme.
+type bls12SecretKey struct {
+	p *blst.SecretKey
+}
+
+// RandKey creates a new private key using a random method provided as an io.Reader.
```
melekes, line 20: doesn't look like we're providing anything

melekes (commented):
👍

melekes:
> We should move this later on to the new github.com/cosmos/crypto repo so that the sdk can use it as well.

Why not move now?

tac0turtle:
moved the curve impl here https://github.com/cosmos/crypto/pull/1

privval/file.go
```diff
@@ -185,7 +185,11 @@ func NewFilePV(privKey crypto.PrivKey, keyFilePath, stateFilePath string) *FileP
 // GenFilePV generates a new validator with randomly generated private key
 // and sets the filePaths, but does not call Save().
 func GenFilePV(keyFilePath, stateFilePath string) *FilePV {
-	return NewFilePV(ed25519.GenPrivKey(), keyFilePath, stateFilePath)
+	key, err := bls.GenPrivKey()
```
tac0turtle, line 188: for testing

faddat:
thanks!

types/signable.go
```diff
@@ -1,14 +1,14 @@
 package types
 
 import (
-	"github.com/cometbft/cometbft/crypto/ed25519"
+	"github.com/cometbft/cometbft/crypto/bls"
```
itsdevbear, line 4: let's make this bls12381 to prevent ambiguity

tac0turtle:
is there a timeline that this can be merged. i can dedicate when that time is closer otherwise ill wait

melekes:
> is there a timeline that this can be merged

I think we can merge it as long as it's under a build flag. And the PR shouldn't be in a draft mode, ofc.

crypto/bls/key.go
```diff
@@ -0,0 +1,117 @@
...
+// Cosmos SDK.
+
+// Compile-time type assertion.
+var _ cmcrypto.PrivKey = &PrivKey{}
+
+// PrivKey represents a BLS private key noop when blst is not set as a build flag and cgo is disabled.
+type PrivKey []byte
+
+func NewPrivateKeyFromBytes(bz []byte) (PrivKey, error) {
+	panic("bls12_381 is disabled")
```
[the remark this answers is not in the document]
itsdevbear: @tac0turtle we can definitely support this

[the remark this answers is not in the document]
tac0turtle: this is the noop key. key_12381 is the bls key that will be supported. This is done to avoid init magic

itsdevbear:
@tac0turtle do u need help finishing this up

melekes (approved):
Thanks @tac0turtle ❤️

melekes:
@tac0turtle @itsdevbear do we need this in v1 / v0.38?

melekes:
I can't push changes to the fork due to permissions. Closing in favor of https://github.com/cometbft/cometbft/pull/2765

PR #2539, opened 2024-03-08
feat(privval): allow different curves 

tac0turtle:
This pr allows different curves to be used with privval file.go for generation. This avoids the need for application devs to write a tool to generate the keys. While its not hard to write the tool its life improvement 

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/cmd/priv_val_server/main.go
+++ b/cmd/priv_val_server/main.go
@@ -42,7 +42,12 @@ func main() {
 		dialer = privval.DialUnixFn(address)
 	case "tcp":
 		connTimeout := 3 * time.Second // TODO
-		dialer = privval.DialTCPFn(address, connTimeout, ed25519.GenPrivKey())
+		key, err := ed25519.GenPrivKey()
+		if err != nil {
+			logger.Error("Failed to generate priv val key", "err", err)
+			os.Exit(1)
+		}
+		dialer = privval.DialTCPFn(address, connTimeout, key)
 	default:
 		logger.Error("Unknown protocol", "protocol", protocol)
 		os.Exit(1)
--- a/privval/file.go
+++ b/privval/file.go
@@ -11,7 +11,6 @@ import (
 
 	cmtproto "github.com/cometbft/cometbft/api/cometbft/types/v1"
 	"github.com/cometbft/cometbft/crypto"
-	"github.com/cometbft/cometbft/crypto/ed25519"
 	cmtos "github.com/cometbft/cometbft/internal/os"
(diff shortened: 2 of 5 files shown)


cmd/priv_val_server/main.go
```diff
@@ -42,7 +42,12 @@ func main() {
 		dialer = privval.DialUnixFn(address)
 	case "tcp":
 		connTimeout := 3 * time.Second // TODO
-		dialer = privval.DialTCPFn(address, connTimeout, ed25519.GenPrivKey())
+		key, err := ed25519.GenPrivKey()
+		if err != nil {
+			logger.Error("Failed to generate priv val key", "err", err)
```
melekes, line 47:
```suggestion
			logger.Error("Failed to generate private validator's key", "err", err)
```

melekes (commented):
Conceptual ACK 👍

tac0turtle:
should i wrap up this pr?

privval/file.go
```diff
@@ -184,8 +183,12 @@ func NewFilePV(privKey crypto.PrivKey, keyFilePath, stateFilePath string) *FileP
 
 // GenFilePV generates a new validator with randomly generated private key
 // and sets the filePaths, but does not call Save().
-func GenFilePV(keyFilePath, stateFilePath string) *FilePV {
-	return NewFilePV(ed25519.GenPrivKey(), keyFilePath, stateFilePath)
+func GenFilePV(keyFilePath, stateFilePath string, keyGen func() (crypto.PrivKey, error)) *FilePV {
```
cason, line 186: I think we could create a new method, receiving this third parameter. Then we maintain this method as it is, but calling the new method with the default implementation, i.e., ed255519.

This in particular allows us to not change the `cmd/priv_val_server` at all.

privval/file.go
```diff
@@ -184,8 +183,12 @@ func NewFilePV(privKey crypto.PrivKey, keyFilePath, stateFilePath string) *FileP
...
+	key, err := keyGen()
```
cason, line 187: Do we really need this to return an error? Again, changes in all previous files are only needed because we have this new requirement.

cason (commented):
Legit, but some stuff to fix.

privval/file.go
tac0turtle: errors are always better than panics imo. since comet is a library the caller should decide how to handle the error not the library

tac0turtle:
a bit swamped at the moment, this is a short term fix but if the cli commands stay the same then there isnt much of a point to do this. Ill close this pr

PR #2581, opened 2024-03-11
spec(gossip): Specifies the gossip layer of consensus as a CRDT

lasarojc:
Only for reviewing the current status of branch `spec/consensus-gossip` -- DO NOT MERGE

Initial reviews are captured in the branches that merged into this one. Comments were addressed as best as possible at the time of writing, but given how much time has passed and context lost, reviews should probably start on this PR and ignore the original ones.

- [ ] warn the reader this is experimental spec and it's incomplete
- [ ] cross-check with knowledge base -- if not, port it, and let it live in there


adizere:
deprioritized

PR #2597, opened 2024-03-12
fix(e2e): Add clock skew only to processes that accumulate less than 1/3 of voting power

hvanz:
Some nightly tests are failing because validators with clock skew and high stake are slowing down progress.

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

PR #2605, opened 2024-03-13
fix(e2e): Increase waiting timeout when killing, pausing, or restarting a node

hvanz:
When a node restarts and there's no state on disk to restore (because of `persist_interval = 0`), the application takes some time to replay blocks before starting. During this time, the RPC endpoints are still disabled, so the e2e's waiting function cannot check the nodes' status.

---

#### PR checklist

- [ ] Tests written/updated


lasarojc (approved):
Increasing the timeout from 20s to 100s in some cases. LGTM

PR #2606, opened 2024-03-13
fix(e2e): Increase timeout when waiting for new block

hvanz:
Some nightly test scenarios are failing because the test that checks that new blocks are being created timeouts. This happens on larger networks with many perturbations, on a few heights. Currently the timeout is 5 seconds, but in some cases it takes the network more than 20 seconds to create a block, so this PR changes it to 30 seconds.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/test/e2e/tests/app_test.go
+++ b/test/e2e/tests/app_test.go
@@ -56,7 +56,7 @@ func TestApp_Hash(t *testing.T) {
 			require.NoError(t, err)
 			require.NotZero(t, status.SyncInfo.LatestBlockHeight)
 			return status.SyncInfo.LatestBlockHeight >= requestedHeight
-		}, 5*time.Second, 500*time.Millisecond)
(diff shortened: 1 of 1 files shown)


lasarojc (approved):
The change itself looks good, but will this not cause the testing framework to timeout?

hvanz:
Is there a timeout on the whole test that could expire before 30 seconds? If we see that the tests still timeouts after this fix, we can fix it in another PR.

lasarojc:
Cool. The CI relies on `test-group` action, in the root `Makefile`. The timeout there is set to 15 minutes
`test-group-%:split-test-packages
        cat $(BUILDDIR)/packages.txt.$* | xargs go test -mod=readonly -timeout=15m -race -coverprofile=$(BUILDDIR)/$*.profile.out` 

Probably enough

Issue #1742, opened 2023-12-04
Proposer often gossips the first block parts to all peers


evan-forbes:
ref https://github.com/celestiaorg/celestia-core/issues/1145

Currently when the proposer creates a block, it will send each block part over the internal message queue. This can actually take a some time. The gossip routine doesn't account for that and will start gossiping immediately, which results in sending the same block parts to all peers instead of sending them randomly.

This is why when we trace the block parts, we often see them downloaded in order. This also defeats the purpose of using block parts entirely, as we end up gossiping the entire block to all peers.

below we can see a diagram that illustrates block propagation for a given height that shows this occuring. on the y axis we have block part indexes, and on the x axis

![broken_proposer](https://github.com/celestiaorg/celestia-core/assets/42654277/6453e488-453f-417e-b088-26cd0f909ac0)

If we [simply wait](https://github.com/celestiaorg/celestia-core/commit/b6b57b9109b0880cf67967e60a6d6209f2ce0fb1) for the block to go over the internal message queue before gossiping, it fixes the issue and we gossip all of the block parts randomly!

There's still quite a bit of room to optimize which parts we send to each peer. One simple non-breaking change would be to not gossip block parts entirely randomly. Instead, we should keep track of which parts have been sent to any peer, and prefer to send a part to a peer if it has been sent to fewer peers.

While I'm convinced that this was one of the bottlenecks, it certainly isn't the only one since this change still doesn't increase throughput on its own. ATM, I think there's quite a lot of backup occurring, because we frequently end up receiving block parts super late.

![fixed_proposer](https://github.com/celestiaorg/celestia-core/assets/42654277/75db2b12-fbc2-40d3-b182-75d1d6617342)

lasarojc:
Thanks @evan-forbes for sharing these very interesting results. A few questions

- Which version of the software are you using?
- Do you have the data for a block producer? I would like to see how the block propagates out of the now that from the the beginning already has all blocks
- Would you be able to reproduce this experiment with a branch having the `blockPartMessage` being propagated as well?

evan-forbes:
I'm using [this branch](https://github.com/celestiaorg/celestia-core/tree/evan/big-blonks-2), which is based on v0.34

what is meant by blockPartMessage, do you mean HasProposalBlockPartMessage?

melekes:
> what is meant by blockPartMessage, do you mean HasProposalBlockPartMessage?

Yes

evan-forbes:
[...]

PR #2624, opened 2024-03-16
fix: correctly calculate numFinalBits in getTrueIndices

czarcas7ic:
I was looking into the bug where seemingly random blocks are only signed by 2/3rd power of validators. The Injective devs said they saw a pattern of larger blocks causing this problem. I then started to look into block parts, and looked into the logic of PickRandom to see if there was potentially a bug in how we "randomly" select blocks. 

While I don't think this is the root cause of the issue (https://github.com/cometbft/cometbft/issues/1742 this is probably more useful in solving the problem we are seeing here), I did come across us handling the last element incorrectly when returning the true indicies. The previous calculation assumes that all previous (numElems-1) are fully accounted for with 64 bits each, but if ba.Bits in not a multiple of 64, then this must be handled differently.

This PR correctly handles the last element in the getTrueIndices method, which is used in PickRandom for propagating block parts.

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
Thanks @czarcas7ic ❤️ 

Would be great to cover this with a unit test.

melekes:
Also, this PR needs a changelog entry.

adizere:
@czarcas7ic is this needed for backporting to 0.37 for use in Osmosis mainnet? Or is it enough if we backport to 0.38?

czarcas7ic:
I think it's fine if we can't back port to v0.37. I do think the over arching issue of block parts being propagated when understood should get backported to v0.37

adizere:
> I think it's fine if we can't back port to v0.37. I do think the over arching issue of block parts being propagated when understood should get backported to v0.37

Cool, thanks!
In general we try to avoid backporting to 0.37 (ref [support policy](https://github.com/cometbft/cometbft/discussions/590)).

internal/bits/bit_array.go
```diff
@@ -277,7 +277,10 @@ func (bA *BitArray) getTrueIndices() []int {
 	}
 	// handle last element
 	lastElem := bA.Elems[numElems-1]
-	numFinalBits := bA.Bits - curBit
+	numFinalBits := bA.Bits % 64
```
melekes, line 280: I'm failing to see how it's different than `numFinalBits := bA.Bits - curBit`.

We'll need a test case.

```go
func TestPickRandomCorrectlyHandlesLastElem(t *testing.T) {
	bA := NewBitArray(128)
	bA.SetIndex(127, true)

	elem, ok := bA.PickRandom()
	require.True(t, ok)
	require.Equal(t, 127, elem)
}
```

melekes (changes requested):

czarcas7ic:
Hm, after writing tests it appears melekes is correct and these are handled the same. My appologies! Closing

PR #2659, opened 2024-03-19
refactor(consensus): separate sending votes from "pick votes" logic

sergio-mena:
This is done by @adizere , @ebuchman , and @sergio-mena as part of a refactoring spree :-)

The goal of this is to improve testability of gossipVotesRoutine by separating out the logic for picking which vote to send. We introduced a new pure function, pickVoteToSend, which returns a vote, and then call the new SendVoteSetHasVote method to send it. This greatly simplifies the loop in gossipVoteRoutine, and allows the pickVoteToSend function to be independently tested.

Note for reviewers:
* This is pure refactoring, no logic change was intended.
* This PR can be reviewed commit by commit

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/internal/consensus/reactor.go
+++ b/internal/consensus/reactor.go
@@ -838,7 +838,7 @@ func (*Reactor) gossipVotesForHeight(
 ) bool {
 	// If there are lastCommits to send...
 	if prs.Step == cstypes.RoundStepNewHeight {
-		if ps.PickSendVote(rs.LastCommit) {
+		if vote, ok := ps.PickVoteToSend(rs.LastCommit); ok {
 			logger.Debug("Picked rs.LastCommit to send")
 			return true
 		}
@@ -846,7 +846,7 @@ func (*Reactor) gossipVotesForHeight(
 	// If there are POL prevotes to send...
 	if prs.Step <= cstypes.RoundStepPropose && prs.Round != -1 && prs.Round <= rs.Round && prs.ProposalPOLRound != -1 {
 		if polPrevotes := rs.Votes.Prevotes(prs.ProposalPOLRound); polPrevotes != nil {
-			if ps.PickSendVote(polPrevotes) {
+			if vote, ok := ps.PickVoteToSend(polPrevotes); ok {
 				logger.Debug("Picked rs.Prevotes(prs.ProposalPOLRound) to send",
 					"round", prs.ProposalPOLRound)
 				return true
@@ -855,21 +855,21 @@ func (*Reactor) gossipVotesForHeight(
 	}
 	// If there are prevotes to send...
 	if prs.Step <= cstypes.RoundStepPrevoteWait && prs.Round != -1 && prs.Round <= rs.Round {
-		if ps.PickSendVote(rs.Votes.Prevotes(prs.Round)) {
+		if vote, ok := ps.PickVoteToSend(rs.Votes.Prevotes(prs.Round)); ok {
 			logger.Debug("Picked rs.Prevotes(prs.Round) to send", "round", prs.Round)
 			return true
 		}
 	}
 	// If there are precommits to send...
 	if prs.Step <= cstypes.RoundStepPrecommitWait && prs.Round != -1 && prs.Round <= rs.Round {
-		if ps.PickSendVote(rs.Votes.Precommits(prs.Round)) {
+		if vote, ok := ps.PickVoteToSend(rs.Votes.Precommits(prs.Round)); ok {
 			logger.Debug("Picked rs.Precommits(prs.Round) to send", "round", prs.Round)
 			return true
 		}
 	}
 	// If there are prevotes to send...Needed because of validBlock mechanism
 	if prs.Round != -1 && prs.Round <= rs.Round {
-		if ps.PickSendVote(rs.Votes.Prevotes(prs.Round)) {
+		if vote, ok := ps.PickVoteToSend(rs.Votes.Prevotes(prs.Round)); ok {
 			logger.Debug("Picked rs.Prevotes(prs.Round) to send", "round", prs.Round)
 			return true
 		}
@@ -877,7 +877,7 @@ func (*Reactor) gossipVotesForHeight(
 	// If there are POLPrevotes to send...
 	if prs.ProposalPOLRound != -1 {
 		if polPrevotes := rs.Votes.Prevotes(prs.ProposalPOLRound); polPrevotes != nil {
-			if ps.PickSendVote(polPrevotes) {
+			if vote, ok := ps.PickVoteToSend(polPrevotes); ok {
 				logger.Debug("Picked rs.Prevotes(prs.ProposalPOLRound) to send",
 					"round", prs.ProposalPOLRound)
 				return true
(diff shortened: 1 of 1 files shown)


internal/consensus/reactor.go
```diff
@@ -1232,12 +1246,16 @@ func (ps *PeerState) PickVoteToSend(votes types.VoteSetReader) (vote *types.Vote
 
 	psVotes := ps.getVoteBitArray(height, round, votesType)
 	if psVotes == nil {
-		return nil, false // Not something worth sending
+		return nil // Not something worth sending
 	}
 	if index, ok := votes.BitArray().Sub(psVotes).PickRandom(); ok {
-		return votes.GetByIndex(int32(index)), true
+		vote := votes.GetByIndex(int32(index))
```
andynog, line 1252: even though you're passing an index that should be in the voteSet, in the `GetByIndex` function there's no bound checks and it will return `nil` only if the `voteSet` is nil. Adding a defensive logic in the `GetByIndex` such as the one below can improve the logic

```
if valIndex >= 0 && valIndex < len(voteSet.votes) {
    return voteSet.votes[valIndex]
} else {
   return nil
}
```

right now it just returns `return voteSet.votes[valIndex]`. The logic seems OK everywhere it's invoked because you always check for `ok` and the index should be inside the boundaries but since this is an exported function, good idea to validate the boundaries of the slice 😉

andynog (commented):
just added a suggestion

internal/consensus/reactor.go
```diff
@@ -732,6 +732,62 @@ func (conR *Reactor) gossipDataForCatchup(logger log.Logger, rs *cstypes.RoundSt
...
+		} else {
+			c := conS.blockStore.LoadBlockCommit(prs.Height)
+			if c == nil {
+				return nil
+			}
+			ec = c.WrappedExtendedCommit()
+		}
+		if ec == nil {
+			return nil
+
```
cason, line 781: This empty line is breaking the linter.

internal/consensus/reactor.go
```diff
@@ -1192,34 +1209,31 @@ func (ps *PeerState) setHasProposalBlockPart(height int64, round int32, index in
...
-		}) {
-			ps.SetHasVote(vote)
-			return true
-		}
-		return false
+// SendVoteSetHasVote sends the vote to the peer.
+// Returns true and marks the peer as having the vote if the vote was sent.
+func (ps *PeerState) SendVoteSetHasVote(vote *types.Vote) bool {
+	ps.logger.Debug("Sending vote message", "ps", ps, "vote", vote)
+	if ps.peer.Send(p2p.Envelope{
```
cason, line 1216: Can we simplify this if call?

internal/consensus/reactor.go
```diff
@@ -732,6 +732,62 @@ func (conR *Reactor) gossipDataForCatchup(logger log.Logger, rs *cstypes.RoundSt
...
 
+func pickVoteToSend(
+	logger log.Logger,
+	conS *State,
+	rs *cstypes.RoundState,
+	ps *PeerState,
+	prs *cstypes.PeerRoundState,
+) *types.Vote {
+	// If height matches, then send LastCommit, Prevotes, Precommits.
+	if rs.Height == prs.Height {
```
cason, line 743: I wonder whether this logic could be outside, in the original call.

internal/consensus/reactor.go
```diff
@@ -732,6 +732,62 @@ func (conR *Reactor) gossipDataForCatchup(logger log.Logger, rs *cstypes.RoundSt
...
+		heightLogger := logger.With("height", prs.Height)
+		return pickVotesForHeight(heightLogger, rs, prs, ps)
+	}
+
+	// Special catchup logic.
```
cason, line 748: And the method can start from here, only for catching up?

internal/consensus/reactor.go
```diff
@@ -826,66 +844,65 @@ OUTER_LOOP:
 		}
 
 		time.Sleep(conR.conS.config.PeerGossipSleepDuration)
-		continue OUTER_LOOP
 	}
 }
 
-func (*Reactor) gossipVotesForHeight(
+func pickVotesForHeight(
```
cason, line 850: Is this name precise? We don't pass a height to this method. It Would it be the current node's height?

cason: Also, it returns a single vote, not multiple.

internal/consensus/reactor.go
cason: This is a good point, but if no one changes `votes` between the two calls, how can we get an invalid index? Actually, a bad index would only be produced if the size of the validator set changes, which can only happen over heights.

cason (approved):
Good refactoring. Separating the logic for picking a vote to send to a peer from the actual sending logic is a good improvement.

I left some comments with suggestions. I would separate the most common logic, pick a vote from the current height, from the catchup logic, that takes votes from previous heights.

internal/consensus/reactor.go
```diff
@@ -732,6 +732,62 @@ func (conR *Reactor) gossipDataForCatchup(logger log.Logger, rs *cstypes.RoundSt
...
+
+	// Catchup logic
+	// If peer is lagging by more than 1, send Commit.
+	blockStoreBase := conS.blockStore.Base()
+	if blockStoreBase > 0 && prs.Height != 0 && rs.Height >= prs.Height+2 && prs.Height >= blockStoreBase {
+		// Load the block's extended commit for prs.Height,
+		// which contains precommit signatures for prs.Height.
+		var ec *types.ExtendedCommit
+		var veEnabled bool
+		func() {
```
melekes, line 765: wonder if we really need this `func() {}` wrapper...

cason: Yes, this is really hard to read.

internal/consensus/reactor.go
```diff
@@ -732,6 +732,62 @@ func (conR *Reactor) gossipDataForCatchup(logger log.Logger, rs *cstypes.RoundSt
 	time.Sleep(conR.conS.config.PeerGossipSleepDuration)
 }
 
+func pickVoteToSend(
```
cason, line 735: So, for code clarity, don't you think that `pickVoteToSend` and `pickVotesForHeight` methods should be on the same place?

So either both here, namely before `gossipVotesRoutine` method, or after it.

internal/consensus/reactor.go
cason, line 781:
```suggestion
```

internal/consensus/reactor.go
sergio-mena: Closing this ATM as there seems not to be away to expose the problem. @andynog feel free to open an issue if you would like to take a closer look at this

internal/consensus/reactor.go
sergio-mena: We can, but I propose to do this in another refactoring effort, as these calls are done like this all over our code

internal/consensus/reactor.go
sergio-mena: We started calling it "bla" (no joke!), so I think it improved :-D

sergio-mena: About the height, we're using the heights inside the structs we are passing as params

sergio-mena: About `votes`, I think you got a point. Let me rename

sergio-mena: As @cason is 20 cm away from me, we agreed on this name "in person"

internal/consensus/reactor.go
sergio-mena: I won't discuss in public why this ended up being here. If can give explanations upon request.

internal/consensus/reactor.go
sergio-mena: Discussed in person. I like the fact that the three cases (same height, off-by-1, and off-by-2-or-more) are located at the same place. This was the case before this refactoring, so I'd leave it as it is

internal/consensus/reactor.go
```diff
@@ -766,52 +821,14 @@ OUTER_LOOP:
...
-				}
-				ec = c.WrappedExtendedCommit()
-			}
-			if ec == nil {
-				continue
-			}
-			if ps.PickSendVote(ec) {
-				logger.Debug("Picked Catchup commit to send", "height", prs.Height)
+		if vote := pickVoteToSend(logger, conR.conS, rs, ps, prs); vote != nil {
+			if ps.SendVoteSetHasVote(vote) {
```
cason, line 825: Why is this method public? Any reason?

sergio-mena: No reason AFAICT. Making it private

melekes:
@mergify backport v1.x

PR #2663, opened 2024-03-20
refactor(consensus): separate sending block parts from "pick part" logic

cason:
This was done by @ebuchman as part of a refactoring spree :-)

The goal of this is to improve testability of `gossipDataRoutine` by separating out the logic for picking which block part to send. We introduced a new pure function, `pickPartToSend`, which returns a block part, and then call the new `SendPartSetHasPart` method to send it and update the peer state. This greatly simplifies the loop in `gossipDataRoutine`, and allows the `pickPartToSend` function to be independently tested.

Note for reviewers:
* This PR is built on top of #2659
* This is pure refactoring, no logic change was intended.
* This PR can be reviewed commit by commit

---

#### PR checklist

- [ ] Tests written/updated


internal/consensus/reactor.go
```diff
@@ -1004,6 +836,145 @@ OUTER_LOOP:
...
+	part := blockStore.LoadBlockPart(prs.Height, index)
+	if part == nil {
+		logger.Error("Could not load part", "index", index,
+			"blockPartSetHeader", blockMeta.BlockID.PartSetHeader, "peerBlockPartSetHeader", prs.ProposalBlockPartSetHeader)
+		return nil
+	}
+	return part
+}
+
+func pickVoteToSend(
```
cason, line 920: This method was only moved here.

internal/consensus/reactor.go
```diff
@@ -1004,6 +836,145 @@ OUTER_LOOP:
...
+		part := pickPartForCatchup(heightLogger, rs, prs, blockStore)
+		if part != nil {
+			return part, true
+		}
+	}
+
+	return nil, false
+}
+
+func pickPartForCatchup(
```
cason, line 888: This logic was in `gossipDataForCatchup`.

melekes (approved):
Love this ❤️

internal/consensus/reactor.go
```diff
@@ -564,6 +564,55 @@ func (conR *Reactor) getRoundState() *cstypes.RoundState {
...
+		if prs.ProposalBlockParts == nil {
+			blockMeta := blockStore.LoadBlockMeta(prs.Height)
+			if blockMeta == nil {
+				heightLogger.Error("Failed to load block meta",
+					"blockstoreBase", blockStoreBase, "blockstoreHeight", blockStore.Height())
+				return nil, false
+			} else {
+				ps.InitProposalBlockParts(blockMeta.BlockID.PartSetHeader)
+			}
+			// continue the loop since prs is a copy and not effected by this initialization
```
sergio-mena:
```suggestion
			// continue the loop since prs is a copy and not affected by this initialization
```

internal/consensus/reactor.go
```diff
@@ -585,151 +634,124 @@ OUTER_LOOP:
...
-					ps.InitProposalBlockParts(blockMeta.BlockID.PartSetHeader)
-				}
-				// continue the loop since prs is a copy and not effected by this initialization
+		if part, continueLoop := pickPartToSend(logger, conR.conS.blockStore, rs, ps, prs); part != nil {
+			if ps.SendPartSetHasPart(part, prs) {
 				continue OUTER_LOOP
 			}
-			conR.gossipDataForCatchup(heightLogger, rs, prs, ps, peer)
+		} else if continueLoop {
+			// part is nil but we dont want to sleep
```
sergio-mena, line 601:
```suggestion
			// part is nil but we don't want to sleep
```

internal/consensus/reactor.go
```diff
@@ -1003,6 +778,198 @@ OUTER_LOOP:
...
+		// if we never received the commit message from the peer, the block parts won't be initialized
+		if prs.ProposalBlockParts == nil {
+			blockMeta := blockStore.LoadBlockMeta(prs.Height)
+			if blockMeta == nil {
+				heightLogger.Error("Failed to load block meta",
+					"blockstoreBase", blockStoreBase, "blockstoreHeight", blockStore.Height())
+				return nil, false
+			}
+			ps.InitProposalBlockParts(blockMeta.BlockID.PartSetHeader)
+			// continue the loop since prs is a copy and not effected by this initialization
```
sergio-mena, line 815:
```suggestion
			// continue the loop since prs is a copy and not affected by this initialization
```

sergio-mena (approved):
I went through all the changes, in particular examining when we sleep and when we don't (before and after this PR), and everything looks good to me. Thanks @ebuchman!
Only found a couple of typos.

internal/consensus/reactor.go
```diff
@@ -1208,6 +1175,66 @@ func (ps *PeerState) setHasProposalBlockPart(height int64, round int32, index in
...
 }
 
+// SendPartSetHasPart sends the part to the peer.
+// Returns true and marks the peer as having the part if the part was sent.
+func (ps *PeerState) SendPartSetHasPart(part *types.Part, prs *cstypes.PeerRoundState) bool {
+	// Send the part
+	ps.logger.Debug("Sending block part", "height", prs.Height, "round", prs.Round, "index", part.Index)
+	pp, err := part.ToProto()
+	if err != nil {
+		// NOTE: only returns error if part is nil, which it should never be by here
```
cason, line 1185: This was originally a panic. Are we ok dropping it in favor of a logger.Error() ?

internal/consensus/reactor.go
```diff
@@ -1003,6 +778,198 @@ OUTER_LOOP:
...
+		prs.Height >= blockStoreBase {
+		heightLogger := logger.With("height", prs.Height)
+
+		// if we never received the commit message from the peer, the block parts won't be initialized
+		if prs.ProposalBlockParts == nil {
+			blockMeta := blockStore.LoadBlockMeta(prs.Height)
+			if blockMeta == nil {
+				heightLogger.Error("Failed to load block meta",
+					"blockstoreBase", blockStoreBase, "blockstoreHeight", blockStore.Height())
+				return nil, false
```
cason, line 812: In the original code, we sleep here:

then we continue to the OUTER_LOOP.

cason: In summary, the operation is different.

cason: https://github.com/cometbft/cometbft/blob/dc4590e1c874eda7d760d14e39f656b3e2622cdf/internal/consensus/reactor.go#L617-L628

internal/consensus/reactor.go
```diff
@@ -1003,6 +778,198 @@ OUTER_LOOP:
...
+			}
+			ps.InitProposalBlockParts(blockMeta.BlockID.PartSetHeader)
+			// continue the loop since prs is a copy and not affected by this initialization
+			return nil, true // continue OUTER_LOOP
+		}
+		part := pickPartForCatchup(heightLogger, rs, prs, blockStore)
```
cason, line 818: In the original code, if we don't have the part to send, we: 
1. sleep: time.Sleep(conR.conS.config.PeerGossipSleepDuration)
2. continue OUTER_LOOP (we do that in any case, after calling gossipDataForCatchup)

If we have the part, we continue to OUTER_LOOP. If we don't, we don't continue.

Different behavior again.

cason (commented):
I think we should double-check some possible changes in operation.

internal/consensus/reactor.go
```diff
@@ -585,206 +588,36 @@ OUTER_LOOP:
...
-					ChannelID: DataChannel,
-					Message: &cmtcons.BlockPart{
-						Height: rs.Height, // This tells peer that this part applies to us.
-						Round:  rs.Round,  // This tells peer that this part applies to us.
-						Part:   *parts,
-					},
-				}) {
-					ps.SetHasProposalBlockPart(prs.Height, prs.Round, index)
-				}
-				continue OUTER_LOOP
```
cason, line 607: Observe that we `continue` even though the send operation fails (i.e., the `BlockPart` messages is not sent to the peer). Although this behavior can be discussed, this PR is not intended to change the existing behavior, but just to refactor the implementation.

internal/consensus/reactor.go
```diff
@@ -585,206 +588,36 @@ OUTER_LOOP:
...
-				Height: prs.Height, // Not our height, so it doesn't matter.
-				Round:  prs.Round,  // Not our height, so it doesn't matter.
-				Part:   *pp,
-			},
-		}) {
-			ps.SetHasProposalBlockPart(prs.Height, prs.Round, index)
-		} else {
-			logger.Debug("Sending block part for catchup failed")
-			// sleep to avoid retrying too fast
-			time.Sleep(conR.conS.config.PeerGossipSleepDuration)
```
cason, line 727: Contrary to what reported in the previous comment (line 607), if the send operation does not succeed s (i.e., the `BlockPart` message is not sent to the peer), the routine sleeps for `config.PeerGossipSleepDuration` before this method returns. After the call of this method (`gossipDataForCatchup`), on line 629, the main routine is started again (`continue OUTER_LOOP`).

When mapped to the new, refactored behavior, this means that the routine does not continue, sleeping at the end of the for loop. If the send operation does succeed, the main routine does continue, i.e., the for loop is restarted without a previous sleep step.

internal/consensus/reactor.go
cason: So, correcting the rationale above. The original code sleeps in this case before restarting the main for loop. In this refactored version, this is implemented by not continuing (`continueLoop = false`) and sleeping at the end of the main for loop.

In other words, this behavior is equivalent to the original code's.

internal/consensus/reactor.go
cason: This is indeed a different behavior. As commented for line 727, if sending the `part` to the peer fails, the main routine should sleep. This is achieved with `continueLoop = false`.

internal/consensus/reactor.go
```diff
@@ -1003,6 +778,198 @@ OUTER_LOOP:
...
+
+func pickPartForCatchup(
+	logger log.Logger,
+	rs *cstypes.RoundState,
+	prs *cstypes.PeerRoundState,
+	blockStore sm.BlockStore,
+) *types.Part {
+	index, ok := prs.ProposalBlockParts.Not().PickRandom()
+	if !ok {
+		return nil
```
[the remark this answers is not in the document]
cason: We achieve this by the caller method returning `nil, false`. This means that there is no `part` to send and that the main loop must continue (`continueLoop = false`) and the sleep at the end of the loop is triggered.

internal/consensus/reactor.go
```diff
@@ -1003,6 +778,198 @@ OUTER_LOOP:
...
+					"blockstoreBase", blockStoreBase, "blockstoreHeight", blockStore.Height())
+				return nil, false
+			}
+			ps.InitProposalBlockParts(blockMeta.BlockID.PartSetHeader)
+			// continue the loop since prs is a copy and not affected by this initialization
+			return nil, true // continue OUTER_LOOP
+		}
+		part := pickPartForCatchup(heightLogger, rs, prs, blockStore)
+		if part != nil {
+			return part, true
```
cason, line 820:
```suggestion
			return part, false
```

internal/consensus/reactor.go
```diff
@@ -585,206 +588,36 @@ OUTER_LOOP:
...
-						"blockstoreBase", blockStoreBase, "blockstoreHeight", conR.conS.blockStore.Height())
-					time.Sleep(conR.conS.config.PeerGossipSleepDuration)
-				} else {
-					ps.InitProposalBlockParts(blockMeta.BlockID.PartSetHeader)
-				}
-				// continue the loop since prs is a copy and not effected by this initialization
+		if part, continueLoop := pickPartToSend(logger, conR.conS.blockStore, rs, ps, prs); part != nil {
+			if ps.SendPartSetHasPart(part, prs) {
 				continue OUTER_LOOP
 			}
```
cason, line 599: If sending the `part` fails, we have two different behaviors:

1. In line 607, as detailed in the commented there anchored, we should continue (i.e., restart the loop without sleeping)
2. In line 727, as detailed in the commented there anchored, we should not continue (i.e, continue the loop, thus sleeping at the end of the loop)

This implementation does not allow for this distinction. In fact, in the absence of an `else` clause, we only implement the behavior 2.

cason (commented):
Me and @sergio-mena have discussed the differences in behavior of the original and the refactored code.

The conclusions were added as a comments. A minor refactoring is required, these changes will come in a following commit.

internal/consensus/reactor.go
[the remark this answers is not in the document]
sergio-mena: Can we resolve this comment?

internal/consensus/reactor.go
sergio-mena: Shall we resolve this?

melekes:
@mergify backport v1.x

melekes:
@mergify backport v1.x

Issue #2679, opened 2024-03-26
spec: hyperlink the ABCI Params entry and clean-up 


andynog:
## Bug Report

The current spec does not have a hyperlink to the ABCI Params entry [here](https://github.com/cometbft/cometbft/blob/main/spec/abci/abci%2B%2B_app_requirements.md#list-of-parameters)

Needs to be backported to v0.38.x

andynog:
Fixed in #2680

PR #2680, opened 2024-03-26
spec: added missing hyperlink and some clean-up on ABCI spec

andynog:
close: #2679 

Added hyperlink and cleaned-up hidden comments

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/spec/abci/abci++_app_requirements.md
+++ b/spec/abci/abci++_app_requirements.md
@@ -595,7 +595,7 @@ These are the current consensus parameters (as of v0.38.x):
 
 This parameter is either 0 or a positive height at which vote extensions
 become mandatory. If the value is zero (which is the default), vote
-extensions are not required. Otherwise, at all heights greater than the
+extensions are not expected. Otherwise, at all heights greater than the
 configured height `H` vote extensions must be present (even if empty).
 When the configured height `H` is reached, `PrepareProposal` will not
 include vote extensions yet, but `ExtendVote` and `VerifyVoteExtension` will


spec/abci/abci++_app_requirements.md
```diff
@@ -276,20 +276,6 @@ time.
 
 The existence of this global mutex means Go application developers can get thread safety for application state by routing all reads and writes through the ABCI system. Thus it may be unsafe to expose application state directly to an RPC interface, and unless explicit measures are taken, all queries should be routed through the ABCI Query method.
 
-<!--
```
sergio-mena, line 279: For `v0.38.x` this is good.
However, for `main` and `v1.x` you may want to explain here the new unsync client (check [here)](https://github.com/cometbft/cometbft/blob/v1.0.0-alpha.2/abci/client/unsync_local_client.go#L32) we now have. You would need to adapt the commented text: in `v0.36.x` the former Tendermint Core team decided to remove the mutex altogether, whereas `v1.x` gives the flexibility of keeping the mutex or not, but providing the extra ABCI client.

spec/abci/abci++_app_requirements.md
```diff
@@ -594,25 +580,35 @@ all full nodes have the same value at a given height.
...
+5. [EvidenceParams.MaxAgeNumBlocks](#evidenceparamsmaxagenumblocks)
+6. [EvidenceParams.MaxBytes](#evidenceparamsmaxbytes)
+7. [ValidatorParams.PubKeyTypes](#validatorparamspubkeytypes)
+8. [VersionParams.App](#versionparamsapp)
+
+##### ABCIParams.VoteExtensionsEnableHeight
+
+This parameter is either 0 or a positive height at which vote extensions
+become mandatory. If the value is zero (which is the default), vote
+extensions are not required. Otherwise, at all heights greater than the
```
sergio-mena, line 598:
```suggestion
extensions are not expected. Otherwise, at all heights greater than the
```

sergio-mena (approved):

spec/abci/abci++_app_requirements.md
andynog: I can open a separate issue for this on v1, thanks for pointing that out

sergio-mena: Thanks

PR #2682, opened 2024-03-27
fix(changelog): fix entry in `v1.x` which doesn't have extension `.md`

sergio-mena:
As per title

---

#### PR checklist

- [ ] Tests written/updated


andynog (approved):
lgtm

PR #2688, opened 2024-03-28
Removes unneeded commands and copies the right entrypoint file.

lasarojc:
The entry point file used is entrypoint-builtin, so there is no point in copying others or in specifying an entry point in this file.

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

PR #2709, opened 2024-04-03
revert: Partly undo ADR 109 by re-exporting store & state packages

### LINKED ISSUE #2610: Cosmos SDK: provide an API for accessing `state` and `store`
### Summary

The latest version of Cosmos SDK imports `state` and `store` packages (which in v1 are internal, meaning can't be imported outside of CometBFT).

### Problem Definition

Previously, the way to test something on a testnet for a single validator was by modifying the `genesis.json` file and skipping blocksync as indicated in https://github.com/cometbft/cometbft/issues/1679. A new solution was developed because exporting `genesis.json` can take minutes and requires lots of RAM. Instead of modifying the `genesis.json` file, Cosmos SDK directly modifies the chain's state and updates the store: https://github.com/cosmos/cosmos-sdk/blob/main/server/start.go#L729. For that, it needs access to `state` and `store` packages, which are currently internal.

### Proposal

Either create a subset of API needed for ^ or completely reexport `state` and `store` packages.

adizere:
This partly undoes the changes in #1605 by re-exporting the store/ package.

Closes #2610 

Paging @tac0turtle to check if this can help Cosmos SDK continue migration to Comet v1

---

#### PR checklist


adizere:
todo: remember to mark the packages as deprecated to signal we will potentially make them private in future releases

melekes (approved):
👍

PR #2713, opened 2024-04-03
refactor(consensus): remove redundant code in `State#defaultDoPrevote`

raulk:
This condition is already covered in the previous conditional block.

#### PR checklist

(Not applicable)

- [ ] ~Tests written/updated~
- [ ] ~Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)~
- [ ] ~Updated relevant documentation (`docs/` or `spec/`) and code comments~


cason:
Hey, thank you for pointing this out.

I couldn't really find where this double check was introduced, probably at merging time.

Regarding the solution, I would go for separating the two conditions of the first if in two different ifs, so that we have more precise log messages.

melekes (approved):
Thanks @raulk ❤️

Issue #779, opened 2023-05-01
spec: Clearly document the ordering within `VoteInfo` and `ExtendedVoteInfo`


adizere:
### Summary

It would be useful if the spec for ABCI 2.0 documented the expected ordering of elements within fields `VoteInfo` and `ExtendedVoteInfo` in all the methods where these fields appear. Application developers might expects that these elements are ordered deterministically by validator power; if so, the docs should capture it, and if not, we should also state it.

### Problem Definition

Developers who are integrating v0.38 can be confused about the guarantees on the ordering of elements within `VoteInfo` and `ExtendedVoteInfo`. There are three ABCI methods concerned here:
-  [`RequestPrepareProposal`](https://github.com/cometbft/cometbft/blob/v0.38.x/abci/types/types.pb.go#L1097) and
-  [`RequestProcessProposal`](https://github.com/cometbft/cometbft/blob/v0.38.x/abci/types/types.pb.go#L1197) and 
-  [`RequestFinalizeBlock`](https://github.com/cometbft/cometbft/blob/v0.38.x/abci/types/types.pb.go#L1425).

### Proposal

The method `ValidatorSet.Validators` already handles ordering:

https://github.com/cometbft/cometbft/blob/2a0c1e6c6acf00269810adce6cb0355eea787334/types/validator_set.go#L38-L53

This method is used by [`buildExtendedCommitInfo`](https://github.com/cometbft/cometbft/blob/v0.38.x/state/execution.go#L446) and [`buildLastCommitInfo`](https://github.com/cometbft/cometbft/blob/v0.38.x/state/execution.go#L396), used in the three ABCI methods described above.

Acceptance criteria:

- [x] double-check that the above is accurate
- [ ] extend the specification for the three ABCI methods concerned to document the ordering guarantees

H/t Marko B. for noticing this underspecified part of the spec!

PR #2715, opened 2024-04-03
spec: Add additional information about validator ordering (total power) 

andynog:
close: #779 

This PR adds additional information on `ExtendedCommitInfo` and `CommitInfo` data types about the validator set ordering (total power) guarantees.

#### PR checklist

- [ ] Tests written/updated


adizere (approved):
LGTM, thanks for the swift turnover Andy!

Should we backport this into 0.38 and v1 branches ?

andynog:
Thanks @adizere, I think we should backport at least to `v1`. For `v0.38.x` since it's a spec and we might have many things in v1 that is not part of v0.38 that I feel it would be better just to do a separate PR for v0.38 (which would be faster too).

PR #2754, opened 2024-04-09
refactor(consensus): defaultDoPrevote validity checks

cason:
Follow-up from #2713.

- Separate the two validity checks `Proposal == nil` and `ProposalBlock == nil`
- Improved the comments on related methods.

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

PR #2765, opened 2024-04-10
feat(crypto): add BLS12-381 curve

melekes:
[curve](https://github.com/cosmos/crypto/tree/marko/bls12381) based on [blst](https://github.com/supranational/blst).

The `bls12381` build tag was added. Also, note that BLS requires `cgo`.

Co-authored by: @tac0turtle 

---

#### PR checklist

- [x] Tests written/updated


crypto/bls/key.go
cason, line 1: Shouldn't the name of this package be `bls12381`?

types/block.go
```diff
@@ -587,9 +587,12 @@ const (
 const (
 	// Max size of commit without any commitSigs -> 82 for BlockID, 8 for Height, 4 for Round.
 	MaxCommitOverheadBytes int64 = 94
-	// Commit sig size is made up of 64 bytes for the signature, 20 bytes for the address,
+	// Commit sig size is made up of 96 bytes for the signature, 20 bytes for the address,
 	// 1 byte for the flag and 14 bytes for the timestamp.
-	MaxCommitSigBytes int64 = 109
+	MaxCommitSigBytes int64 = 141
```
cason, line 592: I am afraid the math here is wrong: 96 + 20 + 1 + 14 = 131.

Can we somehow test this?

types/block.go
```diff
@@ -587,9 +587,12 @@ const (
...
+
+	// protoEncodingOverhead represents the overhead in bytes when encoding a protocol buffer message.
+	protoEncodingOverhead int64 = 2
```
cason, line 595: Why private?

.changelog/unreleased/features/2538-bls-signatures.md
```diff
@@ -0,0 +1 @@
+- `[crypto]` Add BLS 12381
```
cason, line 1:
```suggestion
- `[crypto]` Add support to BLS 12381 keys
```

crypto/bls/key.go
```diff
@@ -0,0 +1,116 @@
...
+	cmtjson.RegisterType(PubKey{}, PubKeyName)
+	cmtjson.RegisterType(PrivKey{}, PrivKeyName)
+}
+
+// ===============================================================================================
+// Private Key
+// ===============================================================================================
+
+// PrivKey is a wrapper around the Ethereum bls12_381 private key type. This wrapper conforms to
+// crypotypes.Pubkey to allow for the use of the Ethereum bls12_381 private key type within the
```
cason, line 37: crypotypes?

crypto/bls/key_test.go
```diff
@@ -0,0 +1,39 @@
+//go:build ((linux && amd64) || (linux && arm64) || (darwin && amd64) || (darwin && arm64) || (windows && amd64)) && blst
+
+package bls_test
```
cason, line 3: Shouldn't it be `bls` only?

crypto/encoding/codec.go
```diff
@@ -53,6 +55,12 @@ func PubKeyToProto(k crypto.PubKey) (pc.PublicKey, error) {
 				Secp256K1: k,
 			},
 		}
+	case bls.PubKey:
```
cason, line 58: Are we registering it even when disabled (at compilation time)?

cason (commented):
Left some general comments.

Not sure if the solution for having this key optional, depending on the compiling flags, is ideal.

crypto/bls/key_test.go
melekes: Good catch. Fixed

types/block.go
melekes: The math was wrong before, too. I don't know why exactly.

sergio-mena:
Thanks @melekes I'll take a look as soon as I can

types/block.go
melekes: It's probably because it's an implementation detail.

cason (approved):
Great.

itsdevbear:
@melekes there are non cgo versions as well just fyi, we should support both

melekes:
> @melekes there are non cgo versions as well just fyi, we should support both

@itsdevbear, do you have a Github link? @tac0turtle, if there are non-CGO versions, why did you choose a CGO version?

itsdevbear:
I think the one in geth 

https://github.com/ethereum/go-ethereum/tree/master/crypto/bls12381

is.

Otherwise I can dig it up in the morning just in bed 

I imagine the CGO one is more performant

itsdevbear:
either way, usually good to support multiple implementations to guard against vulns

itsdevbear:
This is used in geth: 

https://github.com/protolambda/bls12-381-util

itsdevbear:
Protolambda's stuff is good

tac0turtle:
Cgo is more performant. Since there is no aggreagtion in comet, yet the performance overhead will be high for doing individual signatures.

crypto/bls12381/key.go
```diff
@@ -0,0 +1,111 @@
...
+// bls12_381 private key type.
+
+// Compile-time type assertion.
+var _ crypto.PrivKey = &PrivKey{}
+
+// PrivKey represents a BLS private key noop when blst is not set as a build flag and cgo is disabled.
+type PrivKey []byte
+
+func NewPrivateKeyFromBytes([]byte) (PrivKey, error) {
+	panic("bls12_381 is disabled")
```
sergio-mena, line 42: Would it be better to return it as an error, rather than panicking? (same comment below)

crypto/bls12381/key.go
```diff
@@ -0,0 +1,111 @@
+//go:build !bls12381 || !((linux && amd64) || (linux && arm64) || (darwin && amd64) || (darwin && arm64) || (windows && amd64))
```
sergio-mena, line 1: Could we express this as the negated version of the one appearing in `crypto/bls12381/key_bls12381.go`, so that it's easier to see they are mutually exclusive?

.changelog/unreleased/features/2765-bls12-381-curve.md
```diff
@@ -0,0 +1,2 @@
+- `[crypto]` Add support for BLS12-381 keys. Use `bls12381` build flag to enable
```
sergio-mena, line 1: Could you insert this?:

_Since it needs `cgo`, we use `bls12381` build flag to enable..._

crypto/bls12381/key.go
```diff
@@ -0,0 +1,111 @@
...
+	SignatureLength = 96
+	// KeyType is the string constant for the bls12_381 algorithm.
+	KeyType = "bls12_381"
+	// Enabled indicates if this curve is enabled.
+	Enabled = false
+)
+
+// -------------------------------------.
+const (
+	PubKeyName = "cometbft/PubKeyBLS12_381"
```
sergio-mena, line 24: To be consistent with other `PubKeyName`, it should be `cometbft/PubKeyBls12_381`

sergio-mena: Also, given the importance of this, it may be worth defining this in a common place (not under conditional build) so that it is defined only once (avoiding the risk of someone later changing it only in one place and it falling out of sync). If you go for this approach, then the consts above could also be moved.

crypto/bls12381/key_bls12381.go
```diff
@@ -0,0 +1,159 @@
...
+
+// Bytes returns the byte representation of the Private Key.
+func (privKey PrivKey) Bytes() []byte {
+	return privKey
+}
+
+// PubKey returns the private key's public key. If the privkey is not valid
+// it returns a nil value.
+func (privKey PrivKey) PubKey() crypto.PubKey {
+	secretKey, _ := bls12381.SecretKeyFromBytes(privKey)
```
sergio-mena, line 74: Please, use the error value here.
This unit test panics on line 76:
```go
func TestInvalidPrivKeyBLS12381(t *testing.T) {
	privKey, err := bls12381.GenPrivKey()
	require.NoError(t, err)
	privKey = privKey[1:] // corrupt key
	pubKey := privKey.PubKey()
	require.Nil(t, pubKey)
}
```

crypto/bls12381/key_bls12381.go
```diff
@@ -0,0 +1,159 @@
...
+
+	return PubKey(secretKey.PublicKey().Marshal())
+}
+
+// Equals returns true if two private keys are equal and false otherwise.
+func (privKey PrivKey) Equals(other crypto.PrivKey) bool {
+	return privKey.Type() == other.Type() && bytes.Equal(privKey.Bytes(), other.Bytes())
+}
+
+// Type returns eth_bls12_381.
```
sergio-mena, line 84: Comment is inaccurate

crypto/bls12381/key_bls12381.go
```diff
@@ -0,0 +1,159 @@
...
+	return KeyType
+}
+
+func (privKey PrivKey) Sign(digestBz []byte) ([]byte, error) {
+	secretKey, err := bls12381.SecretKeyFromBytes(privKey)
+	if err != nil {
+		return nil, err
+	}
+
+	bz := digestBz
```
sergio-mena, line 95: what's the point of introducing `bz`?

crypto/bls12381/key_bls12381.go
```diff
@@ -0,0 +1,159 @@
...
+	if len(bz) > 32 {
+		hash := sha256.Sum256(bz)
+		bz = hash[:]
```
sergio-mena, line 98: The `[:]` has no effect AFAICT

crypto/bls12381/key_bls12381.go
```diff
@@ -0,0 +1,159 @@
...
+// bls12_381 public key type.
+
+var _ crypto.PubKey = &PubKey{}
+
+type PubKey []byte
+
+// Address returns the address of the public key.
+// The function will panic if the public key is invalid.
+func (pubKey PubKey) Address() crypto.Address {
+	pk, _ := bls12381.PublicKeyFromBytes(pubKey)
```
sergio-mena, line 120: Since we're panicking in line 122, can we please check the error here and have a dedicated panic? Otherwise, we may have a `panic` in line 121 when trying to marshall a `nil` public key.

crypto/bls12381/key_bls12381.go
```diff
@@ -0,0 +1,159 @@
...
+	if len(pk.Marshal()) != PubKeySize {
+		panic("pubkey is incorrect size")
+	}
+	// TODO: do we want to keep this address format?
```
sergio-mena, line 124: Any commns to be done to solve this TODO?

crypto/bls12381/key_bls12381.go
```diff
@@ -0,0 +1,159 @@
...
+	return crypto.Address(tmhash.SumTruncated(pubKey))
+}
+
+func (pubKey PubKey) VerifySignature(msg, sig []byte) bool {
+	if len(sig) != SignatureLength {
+		return false
+	}
+	bz := msg
```
sergio-mena, line 132: Why do we need a new variable?

crypto/bls12381/key_bls12381.go
```diff
@@ -0,0 +1,159 @@
...
+	if len(msg) > 32 {
+		hash := sha256.Sum256(msg)
+		bz = hash[:]
+	}
+
+	pubK, _ := bls12381.PublicKeyFromBytes(pubKey)
```
sergio-mena, line 138: Please check the error

crypto/bls12381/key_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+	assert.False(t, pubKey.VerifySignature(msg, sig))
+
+	msg = crypto.CRandBytes(192)
+	sig, err = privKey.Sign(msg)
+	require.NoError(t, err)
+
+	// Test the signature
+	assert.True(t, pubKey.VerifySignature(msg, sig))
+}
```
sergio-mena, line 39: Given that this is adding a new curve, I think it is worth spending some time writing more unit tests (especially those testing error code paths) so that we have good coverage.

go.mod
```diff
@@ -42,11 +42,13 @@ require (
 	github.com/cometbft/cometbft-db v0.11.1-0.20240407063702-fa37a805b0b4
 	github.com/cometbft/cometbft-load-test v0.1.0
 	github.com/cometbft/cometbft/api v1.0.0-alpha.2
+	github.com/cosmos/crypto v0.0.0-20240309083813-82ed2537802e
```
sergio-mena, line 45: No official release?

test/e2e/docker/Dockerfile.debug
```diff
@@ -1,14 +1,14 @@
 # We need to build in a Linux environment to support C libraries, e.g. RocksDB.
 # We use Debian instead of Alpine, so that we can use binary database packages
 # instead of spending time compiling them.
-FROM cometbft/cometbft-db-testing:v0.10.0
+FROM cometbft/cometbft-db-testing:latest
```
sergio-mena, line 4: Can we pin this to a version?

types/block.go
```diff
@@ -587,9 +587,9 @@ const (
 const (
 	// Max size of commit without any commitSigs -> 82 for BlockID, 8 for Height, 4 for Round.
 	MaxCommitOverheadBytes int64 = 94
-	// Commit sig size is made up of 64 bytes for the signature, 20 bytes for the address,
+	// Commit sig size is made up of 96 bytes for the signature, 20 bytes for the address,
 	// 1 byte for the flag and 14 bytes for the timestamp.
-	MaxCommitSigBytes int64 = 109
+	MaxCommitSigBytes int64 = 131 + 10 // where's the 10 from?
```
sergio-mena, line 592: Can you open an issue whose goal is to find out where they come from?

sergio-mena:
I added a few comments, some of them important. Sorry for arriving late to this review

crypto/bls12381/key.go
melekes: Sure

crypto/bls12381/key.go
melekes: How?

.changelog/unreleased/features/2765-bls12-381-curve.md
melekes: We also use `bls12381` build flag because we don't want to bring in new dependencies if the user doesn't need bls.

crypto/bls12381/key.go
melekes: Good idea 👍

crypto/bls12381/key_bls12381.go
melekes: Good catch 👍

crypto/bls12381/key_bls12381.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func (privKey PrivKey) Sign(digestBz []byte) ([]byte, error) {
+	secretKey, err := bls12381.SecretKeyFromBytes(privKey)
+	if err != nil {
+		return nil, err
+	}
+
+	bz := digestBz
+	if len(bz) > 32 {
```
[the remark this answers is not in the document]
melekes: Unfortunately, no. But we can add it.

go.mod
melekes: Nope 

cc @tac0turtle

.changelog/unreleased/features/2765-bls12-381-curve.md
sergio-mena: Agree with you

crypto/bls12381/key_bls12381.go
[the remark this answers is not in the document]
sergio-mena: Fair enough, let's leave it like that, then

go.mod
sergio-mena: Fine, then

test/e2e/docker/Dockerfile.debug
sergio-mena: Minor. If we can't, we can't

types/block.go
melekes: https://github.com/cometbft/cometbft/issues/2838

go.mod
```diff
@@ -42,11 +42,13 @@ require (
...
 	github.com/cosmos/gogoproto v1.4.12
 	github.com/go-git/go-git/v5 v5.12.0
 	github.com/goccmack/goutil v1.2.3
 	github.com/gofrs/uuid v4.4.0+incompatible
 	github.com/google/uuid v1.6.0
+	github.com/minio/sha256-simd v1.0.1
```
cason, line 51: Why using this version instead of the standard one?

PR #2767, opened 2024-04-10
feat(config): new default value for `mempool.max_txs_bytes` parameter

### LINKED ISSUE #2756: mempool/config: reduce the default `max_txs_bytes` value
The current default value is 1GB (1024**3).

Given that the default block size (a consensus parameter, not a config one) was reduced to 4MB, we should adapt the default maximum size in bytes for the mempool accordingly. 

For instance, with maximum-size blocks, the mempool would be drained after 256 blocks by default. A factor of around 10 would be advisable.

_Originally posted by @sergio-mena in https://github.com/cometbft/cometbft/pull/2352#discussion_r1556509571_

cason:
Closes #2756.

The proposed value is 64 MiB = 16 x default `block.max_bytes` (consensus parameter).

---

#### PR checklist

- [ ] Tests written/updated


docs/references/config/config.toml.md
```diff
@@ -1256,32 +1256,41 @@ If the mempool is full, incoming transactions are dropped.
 
 The value `0` is undefined.
 
-### mempool.max_txs_bytes
-The maximum size of all transactions accepted in the mempool.
+### mempool.max_tx_bytes
```
cason, line 1259: Moved up.

.changelog/unreleased/features/2756-mempool-max-txs-bytes.md
cason, line 1: Question: we point to the issue or the PR number? In the file name and in its content?

ancazamfir: imo to the issue

ancazamfir:
> The proposed value is 64 MiB = 16 x default block.max_bytes (consensus parameter).

Looks good.

config/config.go
```diff
@@ -917,9 +917,9 @@ func DefaultMempoolConfig() *MempoolConfig {
 		// Each signature verification takes .5ms, Size reduced until we implement
 		// ABCI Recheck
 		Size:        5000,
-		MaxTxsBytes: 1024 * 1024 * 1024, // 1GB
+		MaxTxBytes:  1024 * 1024,          // 1MB
+		MaxTxsBytes: 16 * 4 * 1024 * 1024, // 64MB
```
ancazamfir, line 921: Curious why not? Trying to show `16 * max_block_size *...`?
```suggestion
		MaxTxsBytes: 64 * 1024 * 1024, // 64MB
```

ancazamfir (approved):

melekes (commented):
`docs/explanation/core/configuration.md` needs to be updated as well

hvanz:
I'd say that 64 Gib is a low number but this is just a guess. If it happens that a few big txs are submitted at the same time, that would take up a big chunk of the mempool capacity and leave a reduced space for the regular-sized txs. Cosmwasm smart contracts could take a couple of Mbs (see https://github.com/cometbft/cometbft/pull/1518#issuecomment-1793240623).

config/config.go
cason: Yes.

cason:
> Cosmwasm smart contracts could take a couple of Mbs

In this case, they have to change the default `max_tx_bytes` (1MB) as well...

Notice that this is not a consensus parameter, is something pretty easy to tune. The real point here is whether it makes sense to allow 1GB transactions by default with block size of 4MB (default again). It would take 256 rounds to drain all transactions from the mempool.

cason:
> `docs/explanation/core/configuration.md` needs to be updated as well

Good point, I will fix it.

But should we maintain this document (as well)? Another source to manually update each time we change a config param...

melekes:
> > `docs/explanation/core/configuration.md` needs to be updated as well
> 
> Good point, I will fix it.
> 
> But should we maintain this document (as well)? Another source to manually update each time we change a config param...

I would probably remove it, yeah. Now that we have `config.toml.md`

melekes (approved):

docs/explanation/core/configuration.md
```diff
@@ -27,7 +27,7 @@ like the file below, however, double check by inspecting the
 
 # The version of the CometBFT binary that created or
 # last modified the config file. Do not modify this.
-version = "0.39.0"
+version = "1.0.0-dev"
```
cason, line 30:
```suggestion
version = "1.0.0"
```

config/config.go
andynog: Agree, ideally if there’s a correlation with these parameters, e.g. tx size, # txs per block, block size, we should look into initializing some values (e.g. MaxTxsBytes = RoundsToDrain * BlockSize * MaxTxBytes) based on a calculation of things that would constrain it

docs/explanation/core/configuration.md
sergio-mena: We can't set it to exactly "1.0.0" on `main`, as we risk to confuse it with the first final release of `v1.x`

sergio-mena (approved):
🙏

config/config.go
cason: Simplified the math, added rationale as a comment

cason:
> I would probably remove it, yeah. Now that we have `config.toml.md`

#2769

config/config.go
andynog: Thanks !

docs/references/config/config.toml.md
```diff
@@ -1256,32 +1256,41 @@ If the mempool is full, incoming transactions are dropped.
...
-max_tx_bytes = 1048576
+max_txs_bytes = 67108864
 ```
 
 | Value type          | integer |
 |:--------------------|:--------|
 | **Possible values** | &gt;= 0 |
 
-This is the maximum size of a transaction allowed to be accepted into the mempool.
+This is the raw, total transaction size. For example, given 1MB
```
andynog, line 1283:
```suggestion
This is the raw, total size in bytes of all transactions in the mempool. For example, given 1MB
```
`total transaction size` might need some better clarification ?

docs/references/config/config.toml.md
```diff
@@ -1256,32 +1256,41 @@ If the mempool is full, incoming transactions are dropped.
...
+transactions and a 5MB maximum mempool byte size, the mempool will
+only accept five transactions.
+
+The maximum mempool byte size should be a factor of the network's maximum block size
+(which is a [consensus parameter](https://docs.cometbft.com/v1.0/spec/abci/abci++_app_requirements#blockparamsmaxbytes)).
+The rationale is to consider how many blocks have to be produced in order to
+drain all transactions stored in a full mempool.
```
andynog, line 1290: Should we add a blurb of text saying what happens if the mempool is "full" ? We added that to the max tx size, but probably would be important to also highlight the fact that if the mempool is considered full txs won't be accepted ?

andynog (approved):
lgtm, added a few suggestions.

Issue #2627, opened 2024-03-18
types: VerifyCommitLightTrusting should at the beginning simply invoke Commit.Block.ValidateBasic() and check that error long before invoking commit.VoteSignBytes


odeke-em:
## Bug Report
Found through fuzzing, if there is a commit with a BlockID whose value is not 32-bytes e.g. `"F4D22138F431C3167CEE09EFE638B875timestF60A1442E6A603E7B8FF2CA561"`, deep within there will be a panic yet really before doing any work we should have validated that the BlockID was valid with this crash
```shell
esting.go:1590: panic: wrong Hash: expected size to be 32 bytes, got 48 bytes
            goroutine 10402 [running]:
            runtime/debug.Stack()
            	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/runtime/debug/stack.go:26 +0x9b
            testing.tRunner.func1()
            	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:1590 +0x1c8
            panic({0x3a7da60?, 0xc00805b120?})
            	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/runtime/panic.go:759 +0x132
            github.com/tendermint/tendermint/types.FuzzVerifyCommitLightTrusting.func1.1()
            	/Users/emmanuelodeke/go/src/github.com/cometbft/cometbft/types/fuzz_test.go:217 +0x135
            panic({0x3a7da60?, 0xc00805b120?})
            	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/runtime/panic.go:759 +0x132
            github.com/tendermint/tendermint/types.CanonicalizeBlockID({{0xc0083a4cf0, 0x30, 0x30}, {0x7b, {0xc00806d280, 0x20, 0x40}}})
            	/Users/emmanuelodeke/go/src/github.com/cometbft/cometbft/types/canonical.go:21 +0x2c5
            github.com/tendermint/tendermint/types.CanonicalizeVote(...)
            	/Users/emmanuelodeke/go/src/github.com/cometbft/cometbft/types/canonical.go:62
            github.com/tendermint/tendermint/types.VoteSignBytes({0xc0083a6580, 0xa}, 0xc006d712d8)
            	/Users/emmanuelodeke/go/src/github.com/cometbft/cometbft/types/vote.go:150 +0x1b
[...]

Issue #2626, opened 2024-03-18
types: verifyBasicValsAndCommit should ensure that each validator set has a non-nil proposer and a non-nil public key


odeke-em:
## Bug Report
Found by fuzzing that if  the validatorset doesn't a proposer or a nil public key that we get this panic

### Setup
```go
package types_test
        
import (
        "testing"
        
        "github.com/cometbft/cometbft/types"
)               
                        
func TestReproVerifyCommit(t *testing.T) {
        vs := &types.ValidatorSet{
                Validators: []*types.Validator{{},{},{},{},{},{},{},{},{},{}},
        }               
        commit := &types.Commit{
                Signatures: []types.CommitSig{{},{},{},{},{},{},{},{},{},{}},
        }               
        var bid types.BlockID
        _ = vs.VerifyCommit("", bid, 0, commit)
}
```

## Fix
```diff
diff --git a/types/validation.go b/types/validation.go
index 02d1b0b56..9946a46fc 100644
--- a/types/validation.go
+++ b/types/validation.go
@@ -338,6 +338,11 @@ func verifyBasicValsAndCommit(vals *ValidatorSet, commit *Commit, height int64,
 		return errors.New("nil validator set")
 	}
 
+        proposer := vals.GetProposer()
+	if proposer == nil || proposer.PubKey == nil {
+		return errors.New("nil proposer.PubKey in validatorSet")
+	}
+
 	if commit == nil {
 		return errors.New("nil commit")
 	}
```

melekes:
See https://github.com/cometbft/cometbft/pull/2662#issuecomment-2048998246

PR #2773, opened 2024-04-11
docs(types): make contracts about commit & vals explicit

melekes:
in VerifyCommitLight* functions
Refs #2627, #2626

---

#### PR checklist

### DIFF
--- a/types/validation.go
+++ b/types/validation.go
@@ -122,6 +122,8 @@ func verifyCommitLightInternal(
 //
 // This method is primarily used by the light client and does NOT check all the
 // signatures.
+//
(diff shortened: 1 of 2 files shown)


types/validation.go
```diff
@@ -138,6 +140,8 @@ func VerifyCommitLightTrusting(
 // for this commit, but there may be some intersection.
 //
 // This method DOES check all the signatures.
+//
+// CONTRACT: must run ValidateBasic() on commit before verifying and check vals (if untrusted).
```
cason, line 144: The validator set should be always trusted here. Otherwise, the methods without `Trusting` in the name are used...

cason (approved):
Doesn't change much, shouldn't be a problem.

PR #2774, opened 2024-04-11
fix(bits): prevent BitArray.UnmarshalJSON from crashing on 0 bits in the JSON

melekes:
This change fixes a bug in which BitArray.UnmarshalJSON hadn't accounted for the fact that invoking NewBitArray(<=0) returns nil and hence when dereferenced would crash with a runtime nil pointer dereference. This bug was found by my security analysis and fuzzing too.

Author: @odeke-em 

Fixes https://github.com/cometbft/cometbft/issues/2658

---

#### PR checklist

- [x] Tests written/updated


cason (approved):
Not clear what the use case for the introduced test unit, but it fixes it.

internal/bits/bit_array_test.go
```diff
@@ -286,3 +286,18 @@ func TestBitArrayProtoBuf(t *testing.T) {
...
+func TestUnmarshalJSONDoesntCrashOnZeroBits(t *testing.T) {
+	type indexCorpus struct {
+		BitArray *BitArray `json:"ba"`
+		Index    int       `json:"i"`
+	}
+
+	ic := new(indexCorpus)
+	blob := []byte(`{"BA":""}`)
+	err := json.Unmarshal(blob, ic)
+	require.Nil(t, err, "the returned error must be nil")
```
sergio-mena, line 301: Use `require.NoError` ?

sergio-mena (approved):
👍 I agree this can (should!) be backported

faddat:
Fantastic

PR #2786, opened 2024-04-11
feat!: remove cleveldb and boltdb

melekes:
`cleveldb` and `boltdb` were deprecated in v1.

---

#### PR checklist


docs/explanation/core/running-in-production.md
```diff
@@ -8,8 +8,8 @@ order: 4
 
 By default, CometBFT uses the `syndtr/goleveldb` package for its in-process
 key-value database. If you want maximal performance, it may be best to install
-the real C-implementation of LevelDB and compile CometBFT to use that using
-`make build COMETBFT_BUILD_OPTIONS=cleveldb`. See the [install
+RocksDB and compile CometBFT to use that using
+`make build COMETBFT_BUILD_OPTIONS=rocksdb`. See the [install
```
andynog, line 12: Is this suggestion based on any evidence that RocksDB provides the maximal performance ? I understand in the past we suggested `cleveldb` because of the `C-Implementation` (not sure if that was also an accurate statement), but maybe we should re-word the `maximal performance` to something like `one of the best performance` or something like this. I don't think we have any benchmarks to support this statement (afaik we didn't test one against the other and we even included Pebble recently that might provide good performance too).

docs/tutorials/install.md
```diff
@@ -80,46 +80,46 @@ git pull origin main
 make install
 ```
 
-## Compile with CLevelDB support
+## Compile with RocksDB support
```
andynog, line 83: were these commands tested on both Ubuntu and MacOS ? Or just a rename ? I am assuming these new commands and instructions would work fine.

test/e2e/generator/generate.go
```diff
@@ -34,7 +34,7 @@ var (
 	}
 
 	// The following specify randomly chosen values for testnet nodes.
-	nodeDatabases = uniformChoice{"goleveldb", "cleveldb", "rocksdb", "boltdb", "badgerdb", "pebbledb"}
+	nodeDatabases = uniformChoice{"goleveldb", "rocksdb", "badgerdb", "pebbledb"}
```
andynog, line 37: Does it make sense to have `goleveldb` here ? In the Dockerfile (above) seems we don't build images with these option.

test/e2e/pkg/manifest.go
```diff
@@ -160,8 +160,8 @@ type ManifestNode struct {
 	// this relates to the providers the light client is connected to.
 	PersistentPeers []string `toml:"persistent_peers"`
 
-	// Database specifies the database backend: "goleveldb", "cleveldb",
-	// "rocksdb", "boltdb", "pebbledb" or "badgerdb". Defaults to goleveldb.
+	// Database specifies the database backend: "goleveldb", "rocksdb",
```
andynog, line 163: again, is `goleveldb` supported in e2e ?  just checking again based on content of Dockerfile

test/e2e/pkg/testnet.go
```diff
@@ -539,7 +539,7 @@ func (n Node) Validate(testnet Testnet) error {
 		return fmt.Errorf("invalid block sync setting %q", n.BlockSyncVersion)
 	}
 	switch n.Database {
-	case "goleveldb", "cleveldb", "boltdb", "rocksdb", "badgerdb", "pebbledb":
+	case "goleveldb", "rocksdb", "badgerdb", "pebbledb":
```
andynog, line 542: again `goleveldb` support

andynog (approved):
lgtm, added a few comments nothing major

test/e2e/generator/generate.go
melekes: CometBFT comes with goleveldb. You don't need a flag for it.

docs/explanation/core/running-in-production.md
jmalicevic: @melekes , @andynog , just seeing this. I wonder whether Andy is right and we might want to suggest people to use either rocksdb or pebble. Or we could run the benchmarks on rocksdb to be sure of this. We could also link to the storage report and report the fact that pebble does good compaction vs goleveldb.

melekes: @jmalicevic what you're suggesting sounds good to me. The changes above to the documentation were more like a quick patch.

jmalicevic: I just saw that this text was actually removed with a subsequent PR. I can open a small issue to extend this with the one line saying there are multiple options and to look for these two as well for better performance. But its low priority now that we are not explicitly suggesting rocksdb.

PR #2788, opened 2024-04-11
fix(rpc): Make websocket's log on success be in Debug, not info

ValarDragon:
This log is extreme noise in our current setup for all RPC's. I moved to Debug, but should be deleted imo.

This logging of requests/responses should come at a higher proxy level IMO, or have the request + response written elsewhere asynchronously.

---

#### PR checklist

- [ ] Tests written/updated


andynog (approved):
lgtm. I think this makes sense to move to debug. Thanks for doing that.

PR #2792, opened 2024-04-12
refactor(rpc): miss stop PingPongLatencyTimer

### LINKED ISSUE #2771: HTTP client leaks PingPongLatencyTimer
### What happened?

`rpc/jsonrpc/client.WSClient` starts a `github.com/rcrowley/go-metrics.Timer` but never closes it, thus `rpc/client/http.HTTP` leaks.

Either:

- Some method on HTTP that releases the timer; or
- The timer is released when HTTP.Stop is called and not allocated in the first place unless HTTP.Start is called.

### How to reproduce it

Call `rpc/client/http.NewClient`.

Besides the field definition and initialization, `PingPongLatencyTimer` is only referenced once: `c.PingPongLatencyTimer.UpdateSince(t)`. There are no calls to `c.PingPongLatencyTimer.Stop()`. More significantly, **it isn't used by anything**. `PingPongLatencyTimer` appears to be dead code. It could be completely removed without affecting Tendermint.

hoanguyenkh:
Closes #2771

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/rpc/jsonrpc/client/ws_client.go
+++ b/rpc/jsonrpc/client/ws_client.go
@@ -213,6 +213,7 @@ func (c *WSClient) Stop() error {
 	}
 	// only close user-facing channels when we can't write to them
 	c.wg.Wait()
+	c.PingPongLatencyTimer.Stop()
 	close(c.ResponsesCh)
 
 	return nil


rpc/jsonrpc/client/ws_client.go
```diff
@@ -213,6 +213,7 @@ func (c *WSClient) Stop() error {
 	}
 	// only close user-facing channels when we can't write to them
 	c.wg.Wait()
+	c.PingPongLatencyTimer.Stop()
```
melekes, line 216: I don't think it's used anywhere, right? If so, let's just remove it.

melekes (approved):
Thanks @hoanguyenkh ❤️

rpc/jsonrpc/client/ws_client.go
hoanguyenkh: i think used in latency stats

PR #2791, opened 2024-04-12
fix: BootstrapState use custom genesis doc provider

yihuang:
The cosmos-sdk 0.50 genesis doc marshalling not being compatible with the cometbft default one (see https://github.com/cosmos/cosmos-sdk/issues/18477), so we need to support custom genesis doc provider in this public API as well, it's called by the `bootstrap-state` command in cosmos-sdk.

Please backport to 0.38, the SDK 0.50 `bootstrap-state` command is already broken.

---

#### PR checklist

- [ ] Tests written/updated
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments
- [ ] Title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) spec

PR #2793, opened 2024-04-12
fix: BootstrapState use custom genesis doc provider (backport: #2791)

yihuang:
The cosmos-sdk 0.50 genesis doc marshalling not being compatible with the cometbft default one (see https://github.com/cosmos/cosmos-sdk/issues/18477), so we need to support custom genesis doc provider in this public API as well, it's called by the `bootstrap-state` command in cosmos-sdk.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -141,7 +141,16 @@ func StateProvider(stateProvider statesync.StateProvider) Option {
 // store are empty at the time the function is called.
 //
 // If the block store is not empty, the function returns an error.
-func BootstrapState(ctx context.Context, config *cfg.Config, dbProvider cfg.DBProvider, height uint64, appHash []byte) (err error) {
+func BootstrapState(ctx context.Context, config *cfg.Config, dbProvider cfg.DBProvider, height uint64, appHash []byte) error {
+	return BootstrapStateWithGenProvider(ctx, config, dbProvider, DefaultGenesisDocProviderFunc(config), height, appHash)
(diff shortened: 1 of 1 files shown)


melekes (approved):

node/node.go
```diff
@@ -193,7 +202,7 @@ func BootstrapState(ctx context.Context, config *cfg.Config, dbProvider cfg.DBPr
 		return fmt.Errorf("state not empty, trying to initialize non empty state")
 	}
 
-	genState, _, err := LoadStateFromDBOrGenesisDocProvider(stateDB, DefaultGenesisDocProviderFunc(config))
+	genState, _, err := LoadStateFromDBOrGenesisDocProvider(stateDB, genProvider)
```
andynog, line 205: this changes the semantics a bit, because you're saying that the `config` used in `DefaultGenesisDocProviderFunc(config)` when invoking this function is the same `config` (2nd parameter) in the `BootstrapStateWithGenProvider` function. In some respect I think in the end the original code having to pass `config` to DefaultGenesisDocProviderFunc is weird, probably it should have been a receiver in config. Anyway, I think this change is fine, I don't see any other use cases or any issues that this might cause a problem (I don't think anyone will invoke `BootstrapStateWithGenProvider` with two config :-)

andynog (approved):
lgtm

Issue #2041, opened 2024-01-15
Stopping services which contain sub-routines is a hassle


cason:
This issue was motivated by the PR #2010, that addresses #646 and #652. The problem, however, is more general and possibly causes problems when stopping multiple services/reactors.

In short, when we call `Stop()` for an implementation of `service.Service`, the `service.BaseService` implementation of this method on [service/service.go](https://github.com/cometbft/cometbft/blob/main/internal/service/service.go#L165-L191) does the following:
1. Sanity check: is the service started and not already stopped
2. Changes the status of the service to stopped, causing `IsRunning()` to return false
3. Calls `OnStop()`, which is supposed to be overridden by service implementations
4. Closes the `Quit()` channel, which also releases blocking calls to `Wait()`

Implementations of the `OnStop()` method are supposed, in particular, to stop sub-routines, typically started upon `OnStart()`. The problem is how the routines know they have to leave, when the service was or is being stopped.

The typical approach is to use the `Quit()` channel for that, since once it is closed the service has been stopped. But this channel is only closed once closed once `OnStop()` returns (see above). While `OnStop()` implementations Are supposed to stop all sub-routines. We thus have short circuit here.

The solution adopted in #2010 consists in overriding the `Stop()` method, not the `OnStop()` method as expected by the service definition. The overridden method calls `BaseService.Stop()`, so that to, in particular, close the `Quit()` channel. The sub-routines should eventually, once realizing that this channel is closed, return and notify this to a synchronized abstraction (typically, a `sync.WaitGroup`). The overridden `Stop()` method will then block on the synchronized abstraction before returning.

While this solution, despite the added complexity, mostly works, we have an additional problem here. The `Wait()` call returns once the `Quit()` channel is closed. This channel is closed by `BaseService.Stop()`, once the `OnStop()` method returns. This is the right way to implement this functionality. But when we override the `Stop()` method, so that it waits for all sub-routines to quit, we first call `BaseService.Stop()`, as above described. Which means that the `Wait()` method is released and returns. While the service is not yet completely stopped, as the overridden version of `Stop()` is still waiting for sub-routines to quit. Moreover, the same problem arises if another component or service is using `Quit()` to identify whether this service has been stopped.

[...]

PR #2804, opened 2024-04-12
refactor(rpc): allocate PingPongLatencyTimer on start

### LINKED ISSUE #2771: HTTP client leaks PingPongLatencyTimer

firelizzard18:
Followup to #2792 which closed #2771.

#2792 does not handle the case where Start is never called. If Start is not called, Stop returns an error, thus with #2792's implementation the only way to ensure that PingPongLatencyTimer is cleaned up is to call Start and Stop, even when not using any of the features provided by Start (i.e. events).

This PR moves initialization of PingPongLatencyTimer into OnStart so that it is only initialized if it is going to be used. This PR also moves cleanup of PingPongLatencyTimer into readRoutine's defer statement to align it with other cleanup (i.e. closing the connection).

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/rpc/jsonrpc/client/ws_client.go
+++ b/rpc/jsonrpc/client/ws_client.go
@@ -114,12 +114,11 @@ func NewWS(remoteAddr, endpoint string, options ...func(*WSClient)) (*WSClient, 
 	}
 
 	c := &WSClient{
-		Address:              parsedURL.GetTrimmedHostWithPath(),
-		Username:             username,
-		Password:             password,
-		Dialer:               dialFn,
-		Endpoint:             endpoint,
-		PingPongLatencyTimer: metrics.NewTimer(),
+		Address:  parsedURL.GetTrimmedHostWithPath(),
+		Username: username,
+		Password: password,
+		Dialer:   dialFn,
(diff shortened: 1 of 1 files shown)


melekes (approved):

cason:
Another example for https://github.com/cometbft/cometbft/issues/2041.

cason (approved):
+1

PR #2822, opened 2024-04-16
chore: mark state and store packages as deprecated

melekes:
Please note that I didn't use the `Deprecated` keyword because doing this would result in many warnings for us and our users.

Follow-up to https://github.com/cometbft/cometbft/pull/2709

---

#### PR checklist

### DIFF
--- a/state/doc.go
+++ b/state/doc.go
@@ -1,0 +1,3 @@
+// XXX: This package may be internalized (made private) in the future
+// releases.
+package state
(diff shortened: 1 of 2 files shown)


adizere (approved):
Looks good, guess it makes sense to avoid using the magic key word.

state/doc.go
```diff
@@ -0,0 +1,3 @@
+// XXX: This package may be internalized (made private) in the future
```
cason, line 1:
```suggestion
// XXX: This package may be internalized (made private) in future
```

cason: The same in the next file. We can also make a single-line comment.

cason (approved):
Thank you for this. 

I don't understand the usage of `// XXX:` in terms of comment, but if you find it more appropriate, good.

PR #2833, opened 2024-04-17
chore: use sized array

tac0turtle:
Since the checksum is assumed to be sha256 you might as well set the size in the struct. 

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/node/setup.go
+++ b/node/setup.go
@@ -46,7 +46,7 @@ const readHeaderTimeout = 10 * time.Second
 // SHA256 checksum.
 type ChecksummedGenesisDoc struct {
 	GenesisDoc     *types.GenesisDoc
-	Sha256Checksum [32]byte
+	Sha256Checksum [tmhash.Size]byte
 }
 
 // GenesisDocProvider returns a GenesisDoc together with its SHA256 checksum.
@@ -70,7 +70,7 @@ func DefaultGenesisDocProviderFunc(config *cfg.Config) GenesisDocProvider {
 		if err != nil {
 			return ChecksummedGenesisDoc{}, err
 		}
-		return ChecksummedGenesisDoc{GenesisDoc: genDoc, Sha256Checksum: incomingChecksum}, nil
+		return ChecksummedGenesisDoc{GenesisDoc: genDoc, Sha256Checksum: sizedByteArray(incomingChecksum)}, nil
 	}
 }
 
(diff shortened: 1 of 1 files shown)


node/setup.go
```diff
@@ -46,7 +46,7 @@ const readHeaderTimeout = 10 * time.Second
 // SHA256 checksum.
 type ChecksummedGenesisDoc struct {
 	GenesisDoc     *types.GenesisDoc
-	Sha256Checksum []byte
+	Sha256Checksum [32]byte
```
andynog, line 49:
```suggestion
	Sha256Checksum [tmhash.Size]byte
```
since we use `tmhash.Sum` we use its size, I know we named it `Sha256Checksum` but probably should have called it only `Checksum` to allow changes in the checksum in the future

andynog (changes requested):
thanks for the pointing this out, some slight change recommendation

node/setup.go
tac0turtle: happy to rename to checksum if thats preferred

tac0turtle: if you guys prefer this happy to do it.

node/setup.go
```diff
@@ -70,7 +70,7 @@ func DefaultGenesisDocProviderFunc(config *cfg.Config) GenesisDocProvider {
 		if err != nil {
 			return ChecksummedGenesisDoc{}, err
 		}
-		return ChecksummedGenesisDoc{GenesisDoc: genDoc, Sha256Checksum: incomingChecksum}, nil
+		return ChecksummedGenesisDoc{GenesisDoc: genDoc, Sha256Checksum: sizedByteArray(incomingChecksum)}, nil
```
melekes, line 73: Could you explain how the fixed-size array would be better if it introduced all these conversions?

tac0turtle: yea realised that as well. As a library user its a minor nit to be able to pass what is expected instead, sha256sum, not a unbounded checksum. if the naming is changed to checksum then this becomes mute.

cason: The fun part is that the `tmhash.Sum(jsonBlob)` above does exactly the opposite: produces a `[32]int` array and converts it to a slice...

tac0turtle:
closing as its a mute point. would be nice to use fixed size arrays but seems like a larger refactor

PR #2839, opened 2024-04-18
perf(txindex): Lower allocation overhead of txIndex matchRange

ValarDragon:
In Osmosis we see massive amounts of heap pressure/allocations coming from txIndex matchRange. (Screenshot below from ~1 hour of heap profiling)

image

This PR is expected to fully compatibly drop this down by a factor of 3. It:
- Does not get Key() twice (160GB allocation saved)
- Uses no heap allocations for isTagKey (120GB saved)
- Does not string cast or do strings.Split in parsing the value (~400GB expected saved)
- Reuses the big.Int (24GB saved)

The remaining RAM overhead from .Key() needs a cometbft-db API change. The remaining RAM overhead from extracting the value can be saved with an unsafe call for casting the output to string with no heap allocation, but we can do that in a separate PR.

---

#### PR checklist

- [x] Tests written/updated - All existing tests still apply

### DIFF
--- a/state/txindex/kv/kv.go
+++ b/state/txindex/kv/kv.go
@@ -546,9 +546,9 @@ func lookForHash(conditions []syntax.Condition) (hash []byte, ok bool, err error
 	return nil, false, nil
 }
 
-func (*TxIndex) setTmpHashes(tmpHeights map[string][]byte, it dbm.Iterator) {
-	eventSeq := extractEventSeqFromKey(it.Key())
-	tmpHeights[string(it.Value())+eventSeq] = it.Value()
+func (*TxIndex) setTmpHashes(tmpHeights map[string][]byte, key, value []byte) {
+	eventSeq := extractEventSeqFromKey(key)
+	tmpHeights[string(value)+eventSeq] = value
 }
 
 // match returns all matching txs by hash that meet a given condition and start
@@ -584,7 +584,8 @@ func (txi *TxIndex) match(
 		for ; it.Valid(); it.Next() {
 			// If we have a height range in a query, we need only transactions
 			// for this height
-			keyHeight, err := extractHeightFromKey(it.Key())
+			key := it.Key()
+			keyHeight, err := extractHeightFromKey(key)
 			if err != nil {
 				txi.log.Error("failure to parse height from key:", err)
 				continue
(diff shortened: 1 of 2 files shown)


state/txindex/kv/kv.go
```diff
@@ -749,25 +750,30 @@ func (txi *TxIndex) matchRange(
 		panic(err)
 	}
 	defer it.Close()
+	bigIntValue := new(big.Int)
 
 LOOP:
 	for ; it.Valid(); it.Next() {
-		if !isTagKey(it.Key()) {
+		// TODO: We need to make a function for getting it.Key() as a byte slice with no copies.
```
melekes, line 757: Would you mind opening an issue for this?

melekes (approved):
Thanks @ValarDragon ❤️

state/txindex/kv/kv.go
ValarDragon: Sure! Do you want me to delete the comment as well?

ValarDragon: cometbft-db issue opened: https://github.com/cometbft/cometbft-db/issues/156

melekes: > Sure! Do you want me to delete the comment as well?

no

ValarDragon:
Sorry for adding another commit after review :sweat_smile: . I noticed in the heap profile another source of memory overhead. Its much much lower in magnitude, but the fix was easy so figured I'd still add it before doing a second round of post-fix re-benchmarking. 
image

This commit removes the Key call within `setTmpHashes` and most allocations from `ExtractHeightFromKey`, yielding another ~5GB of allocation reduction

state/txindex/kv/kv.go
```diff
@@ -870,9 +873,25 @@ func isTagKey(key []byte) bool {
...
+		return 0, errors.New("separator not found")
+	}
+
+	// Find the position of the second last occurrence of tagKeySeparator
+	startPos := bytes.LastIndexByte(key[:endPos-1], tagKeySeparatorRune)
+	if startPos == -1 {
+		return 0, errors.New("second last separator not found")
+	}
 
-	return strconv.ParseInt(parts[len(parts)-2], 10, 64)
```
cason, line 861: How did this work until now?

cason (approved):
Thanks for this.

I couldn't find (no test performed, only by reading) operational differences, while the changes probably reduce the memory footprint.

PR #2841, opened 2024-04-18
perf(internal/bits): Significantly speedup bitArray.PickRandom

ValarDragon:
This PR significantly speeds up bitArray.PickRandom which is used in VoteGossip and BlockPart gossip. We saw for a query serving full node, over an hour, this was a very large amount of RAM allocations. (75GB of RAM!)

image

This PR drops it down to 0 allocations, and makes the routine 10x faster on my machine.

OLD:
```
BenchmarkPickRandomBitArray-12           1545199               846.1 ns/op          1280 B/op          1 allocs/op
```
NEW:
```
BenchmarkPickRandomBitArray-12          22192857                75.39 ns/op            0 B/op          0 allocs/op
```

I think the new tests I wrote make this more tested than the old code that was here tbh, but pls let me know if theres more tests we'd like to see!

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/internal/bits/bit_array.go
+++ b/internal/bits/bit_array.go
@@ -3,6 +3,7 @@ package bits
 import (
 	"encoding/binary"
 	"fmt"
+	"math/bits"
 	"regexp"
 	"strings"
 	"sync"
@@ -247,14 +248,61 @@ func (bA *BitArray) PickRandom() (int, bool) {
 	}
 
 	bA.mtx.Lock()
-	trueIndices := bA.getTrueIndices()
-	bA.mtx.Unlock()
-
-	if len(trueIndices) == 0 { // no bits set to true
+	numTrueIndices := bA.getNumTrueIndices()
+	if numTrueIndices == 0 { // no bits set to true
+		bA.mtx.Unlock()
 		return 0, false
 	}
+	index := bA.getNthTrueIndex(cmtrand.Intn(numTrueIndices))
+	bA.mtx.Unlock()
+	if index == -1 {
+		return 0, false
+	}
+	return index, true
+}
 
-	return trueIndices[cmtrand.Intn(len(trueIndices))], true
+func (bA *BitArray) getNumTrueIndices() int {
+	count := 0
+	numElems := len(bA.Elems)
+	for i := 0; i < numElems; i++ {
+		count += bits.OnesCount64(bA.Elems[i])
+	}
+	return count
+}
+
+// getNthTrueIndex returns the index of the nth true bit in the bit array.
+// If there is no such value, it returns 0. It is required that the caller
+// ensures that n is less than or equal to the number of true bits in the bit array
+func (bA *BitArray) getNthTrueIndex(n int) int {
+	numElems := len(bA.Elems)
+	count := 0
+
+	// Iterate over each element
+	for i := 0; i < numElems; i++ {
+		// Count set bits in the current element
+		setBits := bits.OnesCount64(bA.Elems[i])
+
+		// If the count of set bits in this element plus the count so far
+		// is greater than or equal to n, then the nth bit must be in this element
+		if count+setBits >= n {
+			// Find the index of the nth set bit within this element
+			for j := 0; j < 64; j++ {
+				if bA.Elems[i]&(1<<uint(j)) != 0 {
+					if count == n {
+						// Calculate the absolute index of the set bit
+						return i*64 + j
+					}
+					count++
+				}
+			}
+		} else {
+			// If the count is not enough, continue to the next element
+			count += setBits
+		}
+	}
+
+	// If we reach here, it means n is out of range
+	return -1
 }
 
 func (bA *BitArray) getTrueIndices() []int {
--- a/internal/bits/bit_array.go
+++ b/internal/bits/bit_array.go
@@ -272,7 +272,7 @@ func (bA *BitArray) getNumTrueIndices() int {
 
 // getNthTrueIndex returns the index of the nth true bit in the bit array.
 // If there is no such value, it returns 0. It is required that the caller
-// ensures that n is less than or equal to the number of true bits in the bit array
+// ensures that n is less than or equal to the number of true bits in the bit array.
 func (bA *BitArray) getNthTrueIndex(n int) int {
 	numElems := len(bA.Elems)
 	count := 0
--- a/internal/bits/bit_array.go
+++ b/internal/bits/bit_array.go
@@ -305,36 +305,6 @@ func (bA *BitArray) getNthTrueIndex(n int) int {
 	return -1
 }
 
-func (bA *BitArray) getTrueIndices() []int {
-	trueIndices := make([]int, 0, bA.Bits)
-	curBit := 0
-	numElems := len(bA.Elems)
-	// set all true indices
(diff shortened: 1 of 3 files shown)


melekes (approved):
Thanks @ValarDragon ❤️

internal/bits/bit_array_test.go
```diff
@@ -142,6 +148,86 @@ func TestPickRandom(t *testing.T) {
...
+		{"____________________________________________", 1, -1},                               // No true indices
+		{"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx", 63, 63},  // last index of first word
+		{"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx", 64, 64},  // first index of second word
+		{"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx", 100, -1}, // Out-of-range
+
+		// Input beyond 64 bits
+		{"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx", 99, 99}, // Last true index
+
+		// Input less than 64 bits
+		{"x_x_x_", 3, -1}, // Out-of-range
```
cason, line 206: From here a learned that the 3th one is the 4th one, right? Namely, 0 is the first?

internal/bits/bit_array.go
```diff
@@ -247,44 +248,61 @@ func (bA *BitArray) PickRandom() (int, bool) {
...
+	for i := 0; i < numElems; i++ {
+		count += bits.OnesCount64(bA.Elems[i])
+	}
+	return count
 }
 
-func (bA *BitArray) getTrueIndices() []int {
-	trueIndices := make([]int, 0, bA.Bits)
-	curBit := 0
+// getNthTrueIndex returns the index of the nth true bit in the bit array.
```
cason, line 273: I would add that n starts from 0.

internal/bits/bit_array.go
```diff
@@ -247,44 +248,61 @@ func (bA *BitArray) PickRandom() (int, bool) {
...
+// If there is no such value, it returns 0. It is required that the caller
```
cason, line 274: Returns -1, doesn't it?

internal/bits/bit_array.go
```diff
@@ -247,44 +248,61 @@ func (bA *BitArray) PickRandom() (int, bool) {
...
+// ensures that n is less than or equal to the number of true bits in the bit array.
```
cason, line 275: This is requirement is not really a requirement, right? The code support this case, but returns -1.

cason (approved):
Great work.

Left some comments regarding the documentation of the introduced methods.

cason:
I didn't add to the review, but I was thinking the following. This is only a suggestion.

We implement `PickRandom()` by 1. finding all bits set 1/true, lets say `n`, then 2. randomly taking one of these bits (random from `0` to `n-1`). So, we need 1. `n` to produce the random number used in 2.

Assuming that we know the maximum number of bits stored, lets say `m`, why we don't pick a random value from `0` to `m`, lets say `i`. Of course, probably in most cases, we have `m > n`, so potentially `i > n`. But in this case, actually in the general case, we just need to pick the index `i % n`. The problem, of course, is that we don't know `n` a priori (which can be solved in a different way, but lets keep this a bit simpler).

My point here is that we probably can implement `PickRandom()` with a single method. Of course, this method would need another loop, to handle the case where `i > n`. But in this case, after the first loop, we know `n` and we know that `i > n`. Therefore can compute `i % n`, and solve the problem in the second iteration.

internal/bits/bit_array_test.go
```diff
@@ -301,3 +387,15 @@ func TestUnmarshalJSONDoesntCrashOnZeroBits(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, ic.BitArray, &BitArray{Bits: 0, Elems: nil})
 }
+
+func BenchmarkPickRandomBitArray(b *testing.B) {
```
cason, line 391: I wonder if adding some contention to this benchmark would affect the results. I think the results are better with this PR, but maybe this is something to explore. By contention, I mean, another thread changing the bit set while we are retrieving random indexes.

internal/bits/bit_array.go
ValarDragon: Whoops yes, thank you!

ValarDragon:
I thought about the second method, but getting perfect distribution will require rejection sampling which is potentially much slower. (Doing i%n biases it)

Rejection sampling seems like unneeded edge case slowdowns right now. Targeting a single pass seems like large overkill when in the systemic view or what's going on with the bit arrays, we're very wasteful at the moment. (Two Extra full clones in the relevant vote sampling code. We clone the bit array from votes, then do a sub call that allocates another bit array, and everything here is single threaded but does mutexes. The mutex call's Cas operation is likely slower than the first pass of this codes ~16 bitwise math ops per word)

internal/bits/bit_array_test.go
ValarDragon: Where this is used, these mutexes are truly useless. Was going to make an issue to make a new bit array version that gets rid of the mutexes

internal/bits/bit_array.go
ValarDragon: True!

internal/bits/bit_array_test.go
cason: This would be ideal. In particular because I assume that the methods using this are already synchronized.

For instance, consensus state, peer consensus state should be synchronized.

cason:
I didn't fully get this rejection sampling part, but I was wondering whether `i % n` is as random as `rand(n)`, but I am not sure. Maybe it is not, but probably if we make `m` big enough, it becomes pretty much as random as.

cason:
Also agree that removing locks is much more relevant for performance than saving a few loops.

cason:
> The problem, of course, is that we don't know `n` a priori (which can be solved in a different way, but lets keep this a bit simpler).

We can have this as a private field:

-  `SetIndex` adds or subtracts 1
-  `copy()` can be adapted easily
- `IsEmpty()` and `IsFull()` become trivial
- `Update()`, I don't even know the point, same as `copy()`
- `UnmarshalJSON()` uses `SetIndex`, so nothing new here
- `FromProto()` can have this field initialized with a private method (the one added here)

I should probably create an issue from this.

ValarDragon:
Your totally right, that would work! I expect that the difference in performance for the PickRandom call won't be that high (I think were already fast enough here), but agreed that seems like a better solution!

PR #2843, opened 2024-04-18
feat(proto)!: replace `PubKey` in `abci.ValidatorUpdate` with `PubKeyType` and `PubKeyBytes`

melekes:
This PR deprecates `PubKey` in `abci.ValidatorUpdate` and adds two new fields:

- string `PubKeyType`
- bytes `PubKeyBytes`

Co-authored by: @tac0turtle 

---

#### PR checklist

- [x] Tests written/updated


hvanz:
The spec needs to be updated. For instance, here https://github.com/cometbft/cometbft/blob/main/spec/abci/abci%2B%2B_methods.md#validatorupdate

types/protobuf.go
```diff
@@ -59,15 +59,17 @@ func (tm2pb) PartSetHeader(header PartSetHeader) cmtproto.PartSetHeader {
 	}
 }
 
-// XXX: panics on unknown pubkey type.
 func (tm2pb) ValidatorUpdate(val *Validator) abci.ValidatorUpdate {
 	pk, err := cryptoenc.PubKeyToProto(val.PubKey)
```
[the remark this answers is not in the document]
melekes: Good catch 👍 Will fix here and in v1

proto/cometbft/abci/v1/types.proto
```diff
@@ -516,8 +516,10 @@ message Validator {
 
 // ValidatorUpdate is a singular update to a validator set.
 message ValidatorUpdate {
-  cometbft.crypto.v1.PublicKey pub_key = 1 [(gogoproto.nullable) = false];
-  int64                        power   = 2;
+  cometbft.crypto.v1.PublicKey pub_key = 1 [(gogoproto.nullable) = false, deprecated = true];
+  int64  power   = 2;
+  bytes pub_key_bytes = 3;
+  string pub_key_type = 4;
```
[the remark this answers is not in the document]
melekes: UPD: I've removed `pub_key`.

state/execution.go
```diff
@@ -580,14 +579,9 @@ func validateValidatorUpdates(abciUpdates []abci.ValidatorUpdate,
 		}
 
 		// Check if validator's pubkey matches an ABCI type in the consensus params
-		pk, err := cryptoenc.PubKeyFromProto(valUpdate.PubKey)
-		if err != nil {
-			return err
-		}
-
-		if !types.IsValidPubkeyType(params, pk.Type()) {
+		if !types.IsValidPubkeyType(params, valUpdate.PubKeyType) {
```
andynog, line 582: again, here we're checking only if the type is one of the ones in `ValidatorParams`, but since in `ValidatorUpdate` you have the bytes, you might be able to add some logic to check if the size is also right (matching the type size using the `crypto` library) as a defensive practice since this will be coming from an ABCI response

andynog (approved):
Looks good to me, thanks for the refactoring. Added a couple of comments, mostly defensive practices. Good job @melekes  !

abci/types/pubkey.go
```diff
@@ -1,44 +1,15 @@
...
 import (
-	"fmt"
-
-	"github.com/cometbft/cometbft/crypto/ed25519"
-	cryptoenc "github.com/cometbft/cometbft/crypto/encoding"
-	"github.com/cometbft/cometbft/crypto/secp256k1"
+	"github.com/cometbft/cometbft/crypto"
 )
 
-func Ed25519ValidatorUpdate(pk []byte, power int64) ValidatorUpdate {
```
sergio-mena, line 11: I like red 😃

internal/consensus/replay.go
```diff
@@ -304,6 +304,10 @@ func (h *Handshaker) ReplayBlocks(
 	if appBlockHeight == 0 {
 		validators := make([]*types.Validator, len(h.genDoc.Validators))
 		for i, val := range h.genDoc.Validators {
+			// Ensure that the public key type is supported.
```
sergio-mena, line 307: Why wasn't this logic here before, was it a bug, or was is elsewhere?

proto/cometbft/abci/v1/types.proto
```diff
@@ -516,8 +515,11 @@ message Validator {
 
 // ValidatorUpdate is a singular update to a validator set.
 message ValidatorUpdate {
-  cometbft.crypto.v1.PublicKey pub_key = 1 [(gogoproto.nullable) = false];
```
sergio-mena, line 519: Should it (at least) deprecate the corresponding message definition?

test/e2e/app/app.go
```diff
@@ -719,18 +718,14 @@ func (app *Application) checkHeightAndExtensions(isPrepareProcessProposal bool,
...
 func (app *Application) storeValidator(valUpdate *abci.ValidatorUpdate) error {
 	// Store validator data to verify extensions
-	pubKey, err := cryptoenc.PubKeyFromProto(valUpdate.PubKey)
+	pubKey, err := cryptoenc.PubKeyFromTypeAndBytes(valUpdate.PubKeyType, valUpdate.PubKeyBytes)
 	if err != nil {
 		return err
 	}
 	addr := pubKey.Address().String()
 	if valUpdate.Power > 0 {
-		pubKeyBytes, err := valUpdate.PubKey.Marshal()
```
sergio-mena, line 728: I still think you need to marshall here, otherwise you're losing key type information.
Alternatively you can also include the key type concatenated in in line 728 below. But marshalling is a lot simpler

test/e2e/app/app.go
```diff
@@ -852,12 +847,7 @@ func (app *Application) verifyAndSum(
 		if err != nil {
 			return 0, fmt.Errorf("could not hex-decode public key for validator address %s, err %w", valAddr, err)
 		}
-		var pubKeyProto cryptoproto.PublicKey
-		err = pubKeyProto.Unmarshal(pubKeyBytes)
-		if err != nil {
-			return 0, fmt.Errorf("unable to unmarshal public key for validator address %s, err %w", valAddr, err)
-		}
-		pubKey, err := cryptoenc.PubKeyFromProto(pubKeyProto)
+		pubKey, err := cryptoenc.PubKeyFromTypeAndBytes(app.cfg.KeyType, pubKeyBytes)
```
sergio-mena, line 850: This comment is symmetric to the one above. You shouldn't use `app.cfg.KeyType` here, but rather get it from the deserialized data.
If we use `app.cfg.KeyType` here, we won't be able to test a validator set with several keytypes

internal/consensus/replay.go
melekes: Before, we're panicking here if the key type is unknown.

proto/cometbft/abci/v1/types.proto
melekes: We're using it in other places (Validator, remote signer).

andynog (approved):
lgtm, thanks for adding the additional logic @melekes

crypto/encoding/codec.go
cason, line 1: Do you feel that tests would be useful here? We don't have any, so not that there is something to update...

internal/consensus/reactor_test.go
```diff
@@ -1132,3 +1122,7 @@ func TestMarshalJSONPeerState(t *testing.T) {
 			"block_parts":"0"}
 		}`, string(data))
 }
+
+func updateValTx(pubKey crypto.PubKey, power int64) []byte {
```
cason, line 1126: Should we move this to `common_test.go`? It is also used in the next test file.

abci/types/pubkey.go
cason, line 1: This is way simpler!

cason (approved):
Not a part of code I know, but no restrictions.

proto/cometbft/abci/v1/types.proto
sergio-mena: That doesn't look right, but doesn't belong to this PR.
Could you please open an issue to decide what to do with all other places that still use `cometbft.crypto.v1.PublicKey`?

sergio-mena: Or maybe I'm missing something...

sergio-mena (approved):
🚀

proto/cometbft/abci/v1/types.proto
sergio-mena: 👍

PR #2845, opened 2024-04-18
docs(api): Added basic readme for /api directory

adizere:
---

#### PR checklist


cason (approved):
Are we still using the `tendermint` registry, even from v1? If we are not, we should update `../proto/README.md`. 

You could take the opportunity to update `go.sum` here with `go mod tidy`.

adizere:
> Are we still using the tendermint registry, even from v1? If we are not, we should update ../proto/README.md.

Can you explain more what exactly to update? I had a look over the readme and was not clear what to update.

> You could take the opportunity to update go.sum here with go mod tidy.

Good point, thanks for pointing out! 7f9431473

melekes (approved):
👍

cason:
The link is to the tendermint registry: https://github.com/cometbft/cometbft/blob/main/proto/README.md?plain=1#L76

But we have a Comet registry: https://buf.build/cometbft/cometbft

Anyway, this shouldn't be part of this PR.

PR #2846, opened 2024-04-18
perf(libs/json): Lower heap overhead of JSON encoding

ValarDragon:
---

Many RPC methods require JSON marshalled responses. We saw this taking a notable amount of heap allocation in query serving full nodes. This PR removes some extra heap allocations that were being done. We avoided using the more efficient encoder.Encode before, because it added a newline. This PR changes the function signature for these private methods to be using *bytes.Buffer, and then uses the in-buffer methods (rather than a second copy). We then just truncate the final byte after each such call, which does not waste any allocations.

I added a benchmark for the most complex test case. 

OLD:
```
BenchmarkJsonMarshalStruct-12              78992             15542 ns/op            4487 B/op        191 allocs/op
```
New:
```
BenchmarkJsonMarshalStruct-12              93346             11132 ns/op            3245 B/op         58 allocs/op
```

Roughly a 3-4x reduction in the number of allocations, and 20% speedup.

#### PR checklist

- [x] Tests written/updated - Existing tests cover this

### DIFF
--- a/libs/json/encoder.go
+++ b/libs/json/encoder.go
@@ -238,7 +238,7 @@ func encodeReflectInterface(w *bytes.Buffer, rv reflect.Value) error {
 }
 
 func encodeStdlib(w *bytes.Buffer, v any) error {
-	// Stream the output of the JSON marshalling directly into the buffer.
+	// Stream the output of the JSON marshaling directly into the buffer.
 	// The stdlib encoder will write a newline, so we must truncate it,
 	// which is why we pass in a bytes.Buffer throughout, not io.Writer.
 	enc := json.NewEncoder(w)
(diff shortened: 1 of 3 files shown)


libs/json/encoder.go
```diff
@@ -237,14 +237,17 @@ func encodeReflectInterface(w io.Writer, rv reflect.Value) error {
 	return writeStr(w, "}")
 }
 
-func encodeStdlib(w io.Writer, v any) error {
-	// Doesn't stream the output because that adds a newline, as per:
-	// https://golang.org/pkg/encoding/json/#Encoder.Encode
-	blob, err := json.Marshal(v)
+func encodeStdlib(w *bytes.Buffer, v any) error {
+	// Stream the output of the JSON marshalling directly into the buffer.
```
ValarDragon, line 241:
```suggestion
	// Stream the output of the JSON marshaling directly into the buffer.
```

andynog (approved):
logic looks good to me. Thanks!

PR #2855, opened 2024-04-20
perf: TxSearch pagination

czarcas7ic:
Since moving to faster blocks, Osmosis public RPC nodes have noticed massive RAM spikes, resulting in nodes constantly crashing:

Screenshot 2024-04-20 at 11 25 36 AM

After heap profiling, the issue was clearly coming from TxSearch, showing that it was unmarshaling a huge amount of data.

Screenshot 2024-04-20 at 11 28 29 AM

After looking into the method, the issue is that txSearch retrieves all hashes (filtered by the query condition), but we call Get (and therefore unmarshal) every filtered transaction from the transaction index store, regaurdless whether or not the transactions are within the pagination request. Therefore, if one were to call txSearch on an event that happens on almost every transaction, this causes the node to unmarshal essentially every transaction.

We have all the data we need in the key though to sort the transaction hashes without unmarshaling the transactions at all! This PR filters and sorts the hashes, paginates them, and then only retrieves the transactions that fall in the page being requested.

We have run this patch on two of our RPC nodes, and have seen zero spikes on the patched ones thus far!

Screenshot 2024-04-20 at 11 33 11 AM

#### PR checklist

- [x] Tests written/updated


czarcas7ic:
I don't have perms to add labels, but I would imagine this would be a backport to 1, .38, and .37

state/txindex/kv/kv.go
```diff
@@ -923,3 +1003,24 @@ func startKey(fields ...any) []byte {
 	}
 	return b.Bytes()
 }
+
+func validatePage(pagePtr *int, perPage, totalCount int) (int, error) {
+	if perPage < 1 {
+		panic(fmt.Sprintf("zero or negative perPage: %d", perPage))
```
melekes, line 1009: shouldn't we error here?

melekes (commented):
Thanks @czarcas7ic ❤️

melekes:
I am not sure if we can backport this to v0.37 and v0.38 since it's API-breaking.

state/txindex/kv/kv.go
czarcas7ic: I am actually just copying the logic from here https://github.com/cometbft/cometbft/blob/ca4090ea8147496d945b06bd11f907cc305666cf/rpc/core/env.go#L98-L100

Since we cant make the function public and import due to cyclic deps, I had to duplicate here. 

We called this method previously (one level higher in Search), and I wanted to keep the same expectations as there was previously, just essentially switching the order of events.

After seeing this lmk if you would still like me to change to error and I will do so!

melekes: better change to error

czarcas7ic: Done here https://github.com/cometbft/cometbft/pull/2855/commits/0e7ac0a44d88b0586450f63114dc82153a2701b8

CHANGELOG.md
```diff
@@ -213,6 +213,8 @@
 - `[privval]` Retry accepting a connection ([\#2047](https://github.com/cometbft/cometbft/pull/2047))
 - `[state]` Fix rollback to a specific height
   ([\#2136](https://github.com/cometbft/cometbft/pull/2136))
+- `[state]` Fix txSearch performance issue
```
melekes, line 216: the entry should be a separate file in `.changelog/improvements/2855-fix-txsearch-performance.md`.

see https://github.com/informalsystems/unclog

melekes (approved):

CHANGELOG.md
czarcas7ic: My bad, first PR here, will get that in tomorrow morning!

czarcas7ic: Done https://github.com/cometbft/cometbft/pull/2855/commits/140bb5c2adbc947a22ec08bd992a06e84dac57c3

state/txindex/indexer.go
```diff
@@ -7,6 +7,7 @@ import (
 	abci "github.com/cometbft/cometbft/abci/types"
 	"github.com/cometbft/cometbft/internal/pubsub/query"
 	"github.com/cometbft/cometbft/libs/log"
+	ctypes "github.com/cometbft/cometbft/rpc/core/types"
```
cason, line 10: We use `ctypes` for consensus types in the code. Can we rename this to something recalling rpc?

rpc/core/tx.go
```diff
@@ -68,47 +65,28 @@ func (env *Environment) TxSearch(
...
-		sort.Slice(results, func(i, j int) bool {
-			if results[i].Height == results[j].Height {
-				return results[i].Index < results[j].Index
-			}
-			return results[i].Height < results[j].Height
-		})
-	default:
-		return nil, ErrInvalidOrderBy{orderBy}
+	pagSettings := ctypes.Pagination{
+		OrderDesc:   orderBy == "desc",
```
cason, line 77: So we assume that everything that is not `desc` is `asc`.

rpc/core/types/responses.go
```diff
@@ -254,3 +254,13 @@ type ResultEvent struct {
 	Data   types.TMEventData   `json:"data"`
 	Events map[string][]string `json:"events"`
 }
+
+// Pagination provides pagination information for queries.
+// This allows us to use the same TxSearch API for pruning to return all relevant data,
+// while still limiting public queries to pagination.
+type Pagination struct {
```
cason, line 261: Shouldn't we add this type to where the `Search()` is declared?

state/txindex/indexer.go
```diff
@@ -26,7 +27,7 @@ type TxIndexer interface {
 	Get(hash []byte) (*abci.TxResult, error)
 
 	// Search allows you to query for transactions.
-	Search(ctx context.Context, q *query.Query) ([]*abci.TxResult, error)
+	Search(ctx context.Context, q *query.Query, pagSettings ctypes.Pagination) ([]*abci.TxResult, int, error)
```
cason, line 30: So, we are importing rpc/core/types here, does it make sense?

cason: Notice that all implementations of this method already import this package, if I am not wrong. So we could move the `Pagination` type to here?

cason (approved):
Legit.

I didn't test if the resulting responses work as expected, and how the pagination works, but this part looks more like a refactoring.

rpc/core/tx.go
czarcas7ic: Yeah, thought it would be better than to just add more logic gates to error on an invalid orderBy, but happy to change if desired (sort of like how we order asc in the previous code if nothing is provided)

rpc/core/types/responses.go
czarcas7ic: Done here https://github.com/cometbft/cometbft/pull/2855/commits/ee94683447e945094726f70a0b73790a0b7a7bb9

state/txindex/indexer.go
czarcas7ic: Removed need for it from your suggestion here https://github.com/cometbft/cometbft/pull/2855/commits/ee94683447e945094726f70a0b73790a0b7a7bb9

rpc/core/tx.go
czarcas7ic: Added orderBy check, which mirrors the case logic that was previously in place https://github.com/cometbft/cometbft/pull/2855/commits/b7946c4bb7a4ea40f59b3d956a1ee79e58849f92

rpc/core/tx.go
```diff
@@ -68,47 +66,28 @@ func (env *Environment) TxSearch(
...
-			if results[i].Height == results[j].Height {
-				return results[i].Index < results[j].Index
-			}
-			return results[i].Height < results[j].Height
-		})
-	default:
-		return nil, ErrInvalidOrderBy{orderBy}
+	pagSettings := txindex.Pagination{
+		OrderDesc:   orderBy == "desc",
+		IsPaginated: true,
```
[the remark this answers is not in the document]
czarcas7ic: I am specifically talking about here https://github.com/osmosis-labs/cometbft/blob/b7946c4bb7a4ea40f59b3d956a1ee79e58849f92/state/txindex/kv/kv.go#L76-L84

Notice was pass in a blank txindex.Pagination{}, which implies is paginated is false, so it returns all the results. This behavior is only necessary for this method, callers from RPC should only have access to paginated results.

rpc/core/tx.go
```diff
@@ -63,52 +61,38 @@ func (env *Environment) TxSearch(
 		return nil, ErrQueryLength{len(query), maxQueryLength}
 	}
 
-	q, err := cmtquery.New(query)
-	if err != nil {
-		return nil, err
+	// if orderBy is not "asc", "desc", or blank, return error
+	if orderBy != "" && orderBy != "asc" && orderBy != "desc" {
```
melekes, line 65: I think this condition is incorrect.

```suggestion
	switch orderBy {
	    case "", "asc", "desc":
	    default:
	    
```

czarcas7ic: I am not seeing an issue with the condition

czarcas7ic: This logic branch confirms that it is either blank, "asc", or "desc". We then set OrderDesc here

https://github.com/osmosis-labs/cometbft/blob/d4ed02ce870faa8b5a76f163d7026920ad18c75d/rpc/core/tx.go#L82-L87

Which is equal to the previous logic (if "desc", its in descending order, if "" or "asc", its in ascending order)

melekes:
```
// if orderBy is not "asc", "desc", or blank, return error
	if orderBy != "" && orderBy != "asc" && orderBy != "desc" {
	```
	
**You say if OR OR
But your code is if AND AND**

Comment (czarcas7ic) [reply rpc/core/tx.go:65]: Im still not seeing the problem 😥

We want to check if the value is any one of "", "asc", or "desc". 

If the value was "bar", the result would be true true true, causing the error

If the value is "", the result would be false true true, continuing on with the method, etc etc.

Comment (melekes) [reply rpc/core/tx.go:65]: Right, sorry for confusion.

**Comment by melekes** [comment]
```
rpc/core/blocks.go:224:7: string `desc` has 3 occurrences, make it a constant (goconst)
	case "desc", "":
```

PR #2868, opened 2024-04-22
docs(mempool): skeleton of QoS ADR

sergio-mena:
This PR reflects the initial discussions, both internal and with users on introducing QoS to the CometBFT mempool.

We will work on top of these changes until we have a problem definition and preliminary design to circulate with our users.

This ADR is not complete -- it's just a skeleton -- so it will not be merged to `main` but rather to the newly created `feature/mempool-qos` feature branch.

---

#### PR checklist


adizere (approved):

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -0,0 +1,312 @@
...
+Latency(tx) --- important
+
+Properties / Requirements (similar to abci++ requirement)
+
+[Examples]
+* Property 1: tx1 --> tc1, tx2 --> tc1 ---> FIFO
+
+* Property 2: tx1 --> tc1, tx2 --> tc2; tc1 < tc2 --> tx1 first, then tx2
+
+* Property 3: tx1 --> tc1, tx2 --> tc2; tc < tc2 --> latency(tx1)<=latency(tx2)
```
ancazamfir, line 36:
```suggestion
* Property 3: tx1 --> tc1, tx2 --> tc2; tc1 < tc2 --> latency(tx1)<=latency(tx2)
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -0,0 +1,312 @@
...
+        the broadcast goroutine itself could manage the priority of lanes within the existing `select` at the end of the loop
+      * if we're going for (future-proof) solution where max lanes can be configurable,
+        we need to refactor method `TxWaitChan` (or the code that sends the data to that channel) to manage the lane priority there.
+        In this case, the broadcast goroutine will be mainly unmodified.
+* Reap flow
+  * TXs are reaped from higher-priority lanes first, in FIFO order.
+  * We go from the reap loop (currently just one), to probably a nested one
+    * the outer loop goes through all TX lists (remember, one per lane), in decreasing priority order
+    * the current `break` statements in the inner loop (limit of bytes or gas reached), should also break from the outer loop
+* Exit flow (Update, Re-Check). Update: unconditionally; Re-Check: only if App says TX is now invalid
```
ancazamfir, line 179: `Re-Check: only if App says TX is now invalid` 
what does this mean? doesn't recheck need to run in order to determine "now invalid"?

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -0,0 +1,312 @@
...
+
+### Who defines lanes and priorities
+
+The list of lanes and their priorities: are consensus parameters? are defined by the app?
+* Discussion on _where_ the lane configuration is set up. Three options
+  1. `config.toml` / `app.toml`
+  2. ConsensusParams
+  3. Hardcoded in the application. Info passed to CometBFT during handshake
+* Outcome of discussion
+  * 1. vs. 2.
```
ancazamfir, line 205: for me this doesn't render properly (`a. vs. 2.`)

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -0,0 +1,312 @@
...
+What is clear is that each lane will need its own `CList` data structure, right from the beginning.
+* `CList` **need to be different**
+  * if using a common `CList`, the broadcast routine becomes overly complicated,
+    whereas each lane having its own `CList`, the broadcast routine is very simple
+    (advance the highest priority, non-empty lane)
+* cache **can be shared** in MVP
+  * lock contention: we think it won't be a big deal performance-wise for an MVP
+  * cache eviction of entries: lanes with a lot of traffic may force eviction on TXs
+    on lanes with little traffic.
+    * For an SDK application this is irrelevant: SDK implements replay protection (nonce)
```
ancazamfir, line 108: Is it relevant for non-SDK? In general we should also think about non-SDK, ideally we would have a non-SDK "customer" or reviewer.

ancazamfir (approved):

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -0,0 +1,312 @@
...
+  * For Osmosis, a two-lane solution ("native" and "priority") would probably be enough.
+    * This means that we could defer the lane and priority configuration, and not tackle it as part of the MVP
+  * For Injective, their use case seems to be more demanding: many lanes (4?, 9?),
+    with one lane even having less prio than the native lane
+    * If we want to address that use case fully, our MVP _does need_ to tackle the design of lane & prio configuration
+    * so, a more ambitious MVP, implying more energy, time, and resources. We must be careful with what we promise for when
+
+* Byzantine nodes may gossip transactions on a high-priority lane to get preferential treatment.
+  * we may consider banning peers that send transactions with incorrect lane information.
+  * Decision was not to ship lane info, but we're still using different channels for different lanes
```
cason, line 83: So we have a variable number of channels? Notice that channel IDs must match in all instances, and they can't collide.

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -0,0 +1,312 @@
...
+"These are our decisions on what to implement and what not to.
+The rest of the document only tackles the items reported as "to implement" here.
+
+### Lane coordination
+
+There should be some kind of Lane Coordinator that receives uncategorised txs and assigns them a lane.
+This could be implemented in the reactor or inside `CListMempool`.
+* For an MVP, probably it's easier to put the coordination logic in `CListMempool`.
+* There will be a _native_ or _default_ lane, similar to the _native VLAN_ in an 802.1Q network.
+* `CheckTx` will (optionally) return lane information, and only then it can be put in a lane.
```
cason, line 140: So how can we know when a lane is used to broadcast txs from other lanes?

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -0,0 +1,312 @@
...
+2. other are general considerations of all flows, or architectural.
+   These are likely to go in a first, introductory subsection of [Data flow](#data-flow) section
+
+### Data flow
+
+* Entry flow:
+  * [Rough description]: the steps when a Tx enters the node are unmodified, except
+    * `CheckTx` (optionally) returns lane information (if TX valid)
+    * The lane information is used to decide to which transaction list the Tx will be added
+* Broadcast flow:
```
cason, line 165: Using the same routine means, in particular, that a lane with a full send queue would prevent the other lane from being used. Notice that we use a blocking `Send()` on the routine.

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -0,0 +1,312 @@
...
+  * Higher-priority lanes get to be gossiped first, and gossip within a lane is still in FIFO order.
+  * [Rough description]: we need to extend the loop in the broadcast goroutine. Several channels now, not just one
+    * To decide
+      * if we're going for (simpler) solution where with max 2 lanes,
+        the broadcast goroutine itself could manage the priority of lanes within the existing `select` at the end of the loop
+      * if we're going for (future-proof) solution where max lanes can be configurable,
+        we need to refactor method `TxWaitChan` (or the code that sends the data to that channel) to manage the lane priority there.
+        In this case, the broadcast goroutine will be mainly unmodified.
+* Reap flow
+  * TXs are reaped from higher-priority lanes first, in FIFO order.
```
cason, line 175: There is a logic for this in `func (c *MConnection) sendPacketMsg() bool` in `p2p/conn` package. The highest priority first approach may lead to starvation in low priority lanes.

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -0,0 +1,312 @@
...
+* Exit flow (Update, Re-Check). Update: unconditionally; Re-Check: only if App says TX is now invalid
+  * Transactions in higher-priority lanes are processed (updated) first.
+  * Depends on how the multiple TX lists are implemented (see discussion above)
+  * Update contains Re-check at the end
+  * Remember Update is done while holding the mempool lock (no new CheckTx calls... they have to wait)
+  * Changes:
+    * We could update the different TX lists in parallel
+    * `Update` method (implementing `Update` method in `Mempool` interface) is implemented by `CListMempool` so
+      * If we go for several `CListMempool`s, then we'll have to call `Update` on each of them with all the TXs every time
+        * or move `Update` method out of `CListMempool` (where?)
```
cason, line 188: We probably need an actor to manage the multiple lanes (or mempools).

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -0,0 +1,312 @@
...
+      * Also, the way we will do `ReCheck` would be complicated if we had N `Update` calls
```
cason, line 189: Are we keeping this blocking, while holding the mempool lock?

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -0,0 +1,312 @@
...
+        * We will do `ReCheck` lane-by-lane, in FIFO order within a lane
+      * So, this seems to make us lean toward: one `CListMempool` containing N `txs` lists
+* Each lane informs consensus when there are txs available.
```
cason, line 192: Not sure, the current method tries to add an element to the unbuffered channel. Re-using the same logic here seems simpler.

docs/references/architecture/adr-113-mempool-lanes.md
cason: Yes, should be `i.`

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -0,0 +1,312 @@
...
+          * then upgrade the lanes via `ConsensusParams`
+        * also, not clear in which order: community should be careful not to break performance between the passing of both proposals
+        * the `gov` module may allow the two things to be shipped in the same gov proposal,
+          but, if we need to do it that way, what's the point in having lanes in `ConsensusParams` ?
+  * Conclusion: 3.
+    * lane info is "hardcoded" in the app's logic
+    * lane info is passed to CometBFT during handshake
+  * Advantages:
+    * Simple lane update cycle: one software update gov proposal
+    * CometBFT doesn't need to deal with dynamic lane changes: it just needs to set up lanes when starting up (whether afresh, or recovery)
```
cason, line 223: Would updating the app and the lanes render two nodes incommunicable?

I think we should consider the case where two nodes don't fully share their lanes setup.

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -0,0 +1,312 @@
...
+    * currently, mempool is p2p channel ID is `0x30`, which would be a special case: native lane.
+  * How to deal with nodes that are late? So, lane info at mempool level (and thus p2p channels to establish) does not match.
+    * Two cases
+    * A node that is up to date and falls behind (e.g., network instability, etc.)
+      * Not a problem. For lanes to change we **require** a coordinated update.
+        * Reason: if we allow upgrade changing lane info **not** to be a coordinated upgrage,
+          then we won't have consistent lane info across nodes (each node would be running a particular version of the software)
+    * A node that is blocksyncing from far away in the past (software upgrades in the middle)
+      * Normal channels (including `0x30`): same as before
+      * Lane channels (>=`0x80`), not declared until `SwitchToConsensus`
```
cason, line 254: Ok, this is a problem. Channels declaration is _static_. We can't add/remove channels once the p2p layer (namely, the `Switch`) is started.

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -0,0 +1,312 @@
...
+        * If we're not at the latest version (e.g., Cosmovisor-drive blocksync)
+          * channel info likely to be wrong
+          * but we Cosmovisor will kill the node before switching to consensus
+    * We will need to extend the channel-related handshake between nodes to be able to add channels after initial handshake
```
cason, line 258: 😵‍💫

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -0,0 +1,312 @@
...
+    * TODO: this needs a bit more thought/discussion
```
cason, line 259: We should always accept txs in the native lane (using the current standard channel). This solves most of the problems surfacing here.

cason (commented):
We need to talk about channels...

docs/references/architecture/adr-113-mempool-lanes.md
sergio-mena: Yes, we have a solution to that below, which we will have to present to users to make sure it addresses their use cases

docs/references/architecture/adr-113-mempool-lanes.md
sergio-mena: It's irrelevant to any _serious_ chain, as the problem of replay protection MUST be solved at app level, so serious apps should have a mechanism. The cache in the mempool is just a partial solution that helps with performance but that doesn't offer absolute guarantees.

> ideally we would have a non-SDK "customer" or reviewer.

Totally agree, although TTBOMK no non-SDK apps have come up so far with TX dissemination problems that would need QoS

docs/references/architecture/adr-113-mempool-lanes.md
sergio-mena: With the design we're currently considering (the one sketched in these notes) you would detect a cheating peer because:
* you receive a TX on the p2p channel corresponding to lane A
* you then checktx the TX and your app tells you the TX belongs to lane B (lane A being "more privileged" than B) 

@hvanz Something we haven't discussed, and that @cason's question makes me think of. We will need some property that requires the app at _all nodes_ to provide consistent lane information. Otherwise we might end up in the same unfortunate situation we have now, where the outcome of `CheckTx` might depend on some `app.toml` config (e.g., min fees) and therefore `CheckTx` is not consistent across nodes.

docs/references/architecture/adr-113-mempool-lanes.md
sergio-mena: You're very right. We'll have to add this as a caveat. However, to me, this is a problem to tackle at the p2p level. I would go with what we have, and if our users, when validating the implementation (when it's completed) find the same problems as Evan from Celestia, then we'll have to tackle the problem you mention with top prio.

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -0,0 +1,312 @@
...
+   These are likely to go in a first, introductory subsection of [Data flow](#data-flow) section
+
+### Data flow
+
+* Entry flow:
+  * [Rough description]: the steps when a Tx enters the node are unmodified, except
+    * `CheckTx` (optionally) returns lane information (if TX valid)
+    * The lane information is used to decide to which transaction list the Tx will be added
+* Broadcast flow:
+  * Higher-priority lanes get to be gossiped first, and gossip within a lane is still in FIFO order.
```
sergio-mena, line 166:
```suggestion
  * Higher-priority lanes get to be gossiped first, and gossip within a lane is still in FIFO order.
  * We will need a warning about channels sharing the send queue at p2p level (will need data early on to see if this is a problem for lanes in practice
```
Thanks @cason

docs/references/architecture/adr-113-mempool-lanes.md
sergio-mena: I added a bullet point with this. 🙏 @cason

docs/references/architecture/adr-113-mempool-lanes.md
sergio-mena: > The highest priority first approach may lead to starvation in low priority lanes

Yes, just like QoS traffic classes in IP networks. When speaking about QoS, it comes with the territory.

sergio-mena: About the logic in p2p implementing priority in channels, thanks for pointing out; @hvanz we'll need to take a look to see how we can leverage that in our design.

docs/references/architecture/adr-113-mempool-lanes.md
sergio-mena: The text is a bit cryptic 😄 
I means that we only need to manage the mempool lanes (i.e., remove the TX) is Re-checktx returned "no longer valid". Whereas, in the case of `Update`, it unconditionally removes all TXs included in the block just finalized.

docs/references/architecture/adr-113-mempool-lanes.md
sergio-mena: Maybe... we didn't get to that level of detail yet

docs/references/architecture/adr-113-mempool-lanes.md
sergio-mena: Not discussed yet. We may reach out to you if we decide to remove that lock. ATM, I see it as orthogonal, but I haven't given it enough thought.

docs/references/architecture/adr-113-mempool-lanes.md
sergio-mena: We want to get rid of it... remember it's in our laundry list in Notion

docs/references/architecture/adr-113-mempool-lanes.md
sergio-mena: Thanks for pointing out. We'll double-check when the text becomes more mature.

docs/references/architecture/adr-113-mempool-lanes.md
sergio-mena: In the current design, updating lane info requires a coordinated update. That simplifies this design a lot. We need to check with users whether this fits their use cases (mainly, that the don't want to change lanes every day). But if have to drop the "updating lane info requires a coordinated update" requirement, the design will become _way_ more complex. As in every design exercise, it's a tradeoff we must carefully consider.

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -0,0 +1,312 @@
...
+    * If it is absent, it is equivalent to having it as the last element (lowest priority)
+  * Channel ID is a byte
+    * Current channel distribution (among reactors) goes up to `0x61`
+    * Proposal: reserve for mempool lanes channel ID `0x80` and all channels above (so, all channels whose MSB is 1)
+      * max of 128 lanes. Big enough?
+    * currently, mempool is p2p channel ID is `0x30`, which would be a special case: native lane.
+  * How to deal with nodes that are late? So, lane info at mempool level (and thus p2p channels to establish) does not match.
+    * Two cases
+    * A node that is up to date and falls behind (e.g., network instability, etc.)
+      * Not a problem. For lanes to change we **require** a coordinated update.
```
sergio-mena, line 249:
```suggestion
      * Not a problem. For lanes to change we **require** a coordinated upgrade.
```

docs/references/architecture/adr-113-mempool-lanes.md
sergio-mena, line 254:
```suggestion
      * Lane channels (>=`0x80`), not declared. The node can't send/receive mempool-related info. Not a problem, since it's blocksyncing
```

docs/references/architecture/adr-113-mempool-lanes.md
sergio-mena: Don't worry @cason, this line is obsolete 😄

docs/references/architecture/adr-113-mempool-lanes.md
sergio-mena: Yeah. You're probably right, although I'm not sure you even need that. The goal of this part of the design is to allow very late nodes to be able to blocksync, without confusing other nodes with stale lane info. But when you're blocksyncing, you shouldn't even care about TX gossipping (specially when the Cosmovisor model of doing _forces_ you to use outdated versions of the app to do blocksync).

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -0,0 +1,313 @@
...
+
+### Lane coordination
+
+There should be some kind of Lane Coordinator that receives uncategorised txs and assigns them a lane.
+This could be implemented in the reactor or inside `CListMempool`.
+* For an MVP, probably it's easier to put the coordination logic in `CListMempool`.
+* There will be a _native_ or _default_ lane, similar to the _native VLAN_ in an 802.1Q network.
+* `CheckTx` will (optionally) return lane information, and only then it can be put in a lane.
+  (So a tx in a lane has been validated at least once by some node.)
+  * If `CheckTx` provides no lane information: native lane
```
sergio-mena, line 142:
```suggestion
  * If `CheckTx` provides no lane information: native lane
  * Note: if we punish peers (e.g., close connection to them) that send messages on wrong lanes, we will need a property that requires all node to produce consistent lane info in `CheckTx`
```

sergio-mena:
Thanks folks for all your comments! I updated the text with some of your ideas. I am merging this now, to keep working on it

PR #2894, opened 2024-04-25
fix(mempool): panic when the app returns error on CheckTx request

### LINKED ISSUE #2225: mempool: panic when app's CheckTx returns an error while rechecking

hvanz:
Closes #2225

If the app returns an error on an ABCI call (in particular CheckTx), CometBFT should stop, because the error is unrecoverable.

---

#### PR checklist

- [X] Tests written/updated

### DIFF
--- a/.changelog/unreleased/bug-fixes/2225-fix-checktx-request-returns-error.md
+++ b/.changelog/unreleased/bug-fixes/2225-fix-checktx-request-returns-error.md
@@ -1,0 +1,2 @@
+- [`mempool`] Panic when a CheckTx request to the app returns an error
+  ([\#2225](https://github.com/cometbft/cometbft/pull/2225))
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -278,22 +278,19 @@ func (mem *CListMempool) CheckTx(tx types.Tx) (*abcicli.ReqRes, error) {
 	}
 
 	if added := mem.addToCache(tx); !added {
-		mem.logger.Debug("Not cached", "tx", tx.Hash())
 		mem.metrics.AlreadyReceivedTxs.Add(1)
 		// TODO: consider punishing peer for dups,
 		// its non-trivial since invalid txs can become valid,
 		// but they can spam the same tx with little cost to them atm.
 		return nil, ErrTxInCache
 	}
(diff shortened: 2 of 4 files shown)


melekes (approved):
👍

PR #2895, opened 2024-04-25
feat(config)!: merge `timeout_prevote` and `timeout_precommit`

### LINKED ISSUE #2893: Merge `timeout_prevote_delta` & `timeout_precommit_delta` into a single `timeout_vote_delta`
`timeout_prevote` & `timeout_precommit` into a single `timeout_vote`

melekes:
`timeout_prevote_delta` and `timeout_precommit_delta`

into `timeout_vote` and `timeout_vote_delta` accordingly

Closes #2893

---

#### PR checklist


config/config.go
```diff
@@ -1102,14 +1102,10 @@ type ConsensusConfig struct {
...
-	// How long we wait after receiving +2/3 prevotes for “anything” (ie. not a single block or nil)
-	TimeoutPrevote time.Duration `mapstructure:"timeout_prevote"`
-	// How much the timeout_prevote increases with each round
-	TimeoutPrevoteDelta time.Duration `mapstructure:"timeout_prevote_delta"`
-	// How long we wait after receiving +2/3 precommits for “anything” (ie. not a single block or nil)
-	TimeoutPrecommit time.Duration `mapstructure:"timeout_precommit"`
-	// How much the timeout_precommit increases with each round
-	TimeoutPrecommitDelta time.Duration `mapstructure:"timeout_precommit_delta"`
+	// How long we wait after receiving +2/3 prevotes/precommits for “anything” (ie. not a single block or nil)
+	TimeoutRound time.Duration `mapstructure:"timeout_round"`
```
sergio-mena, line 1106: `TimeoutVote`?

docs/references/config/config.toml.md
```diff
@@ -1584,79 +1584,53 @@ round of consensus will adopt increased timeout durations.
...
 observe +2/3 precommits for a block, which would allow the node to commit that
 block in the current round.
 
-Setting `timeout_precommit` to `0s` means that the validator will not wait
-for additional precommits (other than the mandatory +2/3) before moving to the
-next round.
-This has important liveness implications and should be avoided.
+#### Warning
+
+Setting `timeout_round` to `0s` means that the validator will not wait for
```
sergio-mena, line 1623: We need a good way to minimize disruption for operators upgrading from current config (and not thinking of adapting this, so the new field will have value 0).
An idea can be that the new timeout, when 0, means some sensible default value (e.g. our default value: 1s).
Anyway, disabling these timeouts never made any sense to me in production. And if need to disable it in text, we can just set it to `time.Hour`.

Also, this design would enable us to (later on) add those timeouts as ConsensusParams:
* If the value is 0, the node uses the consensus param value
* But if the operator wants to override the consensus param value, they can by setting a value != 0 in the `config.toml`

Let me know what you think

melekes: While this would work in this case, it won't in the case of `skip_timeout_commit`. If the user had `skip_timeout_commit=true` and `timeout_commit=1s`, in v1 if he/she doesn't do anything, the node will be running with `1s` timeout commit.

What if instead we version config and write a CLI tool for upgrading it? If the manual cmd is not desirable, then we can attempt upgrade upon a startup. How does that sound?

sergio-mena: Fair enough, but it risks being quite some effort. Maybe, a middle ground would be:
* We version the config as you say
* We include all changes to the config in a paragraph in `UPGRADING.md`
* Upon start-up, if the version received is
  * In the future, we panic saying this is an old version of CometBFT
  * In the past, we warn the user that they may have some fields that became obsolete
    * we include a link to the paragraph in `UPGRADING.md` to explain what to change
    * alternatively, for safely, we panic and ask the user to
      * bump the version manually
      * while doing that, to pay attention at the config fields that changed

If we decide to go the config tool path (I think it's overkill for the changes we have ATM), there is already one, called Confix, developed by the previous Tendermint Core team, so we should probably reuse it. It's on `v0.36.x` and `master`. AFAICT the SDK folks are also using it.

sergio-mena: Anyway, this discussion should block this PR (although the value 0 to `timeout_round` when upgrading isn't solved)

sergio-mena (approved):

docs/references/config/config.toml.md
sergio-mena: @melekes FYI, I'm adding a point in the community call, to see what users think

melekes: I will address this in a separate PR.

PR #2899, opened 2024-04-26
fix(bits): handle last element in PickRandom

melekes:
Follow-up to #2841

---

#### PR checklist

- [x] Tests written/updated


hvanz (approved):
LGTM, though I'm not familiar with this code. Could we expand the test to include this scenario?

ValarDragon:
Thank you for fixing this bug :pray:

PR #2909, opened 2024-04-27
feat(crypto): Upgrade all usage of `crypto/sha256` to `minio/sha256-simd`

itsdevbear:
Free performance boost, used in hundreds of production projects.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/crypto/hash.go
+++ b/crypto/hash.go
@@ -1,7 +1,7 @@
 package crypto
 
 import (
-	"crypto/sha256"
+	"github.com/minio/sha256-simd"
 )
 
 func Sha256(bytes []byte) []byte {
--- a/crypto/merkle/bench_test.go
+++ b/crypto/merkle/bench_test.go
@@ -1,9 +1,10 @@
 package merkle
 
 import (
-	"crypto/sha256"
 	"strings"
 	"testing"
+
+	sha256 "github.com/minio/sha256-simd"
 )
 
 var sink any
--- a/crypto/merkle/tree.go
+++ b/crypto/merkle/tree.go
@@ -1,9 +1,10 @@
 package merkle
 
 import (
-	"crypto/sha256"
 	"hash"
 	"math/bits"
+
+	sha256 "github.com/minio/sha256-simd"
 )
 
 // HashFromByteSlices computes a Merkle tree where the leaves are the byte slice,
--- a/crypto/secp256k1/secp256k1.go
+++ b/crypto/secp256k1/secp256k1.go
@@ -2,11 +2,12 @@ package secp256k1
 
 import (
 	"bytes"
-	"crypto/sha256"
 	"crypto/subtle"
 	"fmt"
 	"io"
 	"math/big"
+
+	sha256 "github.com/minio/sha256-simd"
 
 	secp256k1 "github.com/btcsuite/btcd/btcec/v2"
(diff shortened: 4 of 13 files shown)


itsdevbear:
Will be CPU architecture dependant, ensure test environment is setup correctly

andynog:
Thanks @itsdevbear. 

@sergio-mena do you know if we have a mixed environment in our tests ? Do we test on Intel and ARM ? 

Looks like the benefits are dependent on message size according to the repo. Might be also interesting to see if this could be tested with different input sizes.

andynog:
One concern that I have though is that this library seems to be a bit stale, [last commit](https://github.com/minio/sha256-simd/commits/master/) was February 2023. Maybe it's stable enough. But since we're replacing it everywhere where a Sha256 is needed, changing it on hash, merkletree, keys, etc, 

It is critical that this library is vetted. I understand this is used in production in many projects, but the concern is if there's a vulnerability how fast it can be addressed as opposed to using the official one that would be patched possibly faster.

sergio-mena:
We are in the process of having a golang crypt expert vet this library change.

itsdevbear:
I can also hide behind an optional build flag is desired.

itsdevbear:
@andynog yeah intel and arm, but notably will see the greatest improvement on cpus with AVX-512 support

itsdevbear:
This being said, we want to optimize some places in the code to use the AVX-512 scheduler directly. But that's a bigger change

cason:
This package was added by #2765. Not sure why they decided to use this precise implementation.

cason:
Overall the code now is kind of confusing with the use of:

-  `crypto/sha256`: standard lib
- `minio/sha256-simd`: possibly improved lib
- `crypto/tmhash`: our own lib, which is a wrapper to `crypto/sha256`

We might consider simplifying this setup.

itsdevbear:
Personally, I like having everywhere in the codebase important from tmhash where tmhash is an alias it whatever library we like. This alias could be changed via build flags for instance.


many to one to few relationship

sergio-mena:
We discussed about this in our internal meeting today. These are the conclusions:

1. We currently don't have performance data that shows an obvious gain in the production setups where CometBFT is used
2. We could gather such performance data ourselves as a due-diligence activity, but we don't have cycles ATM
3. There is an ongoing longer term solution, described in #2241, whereby an app can have CometBFT use custom hashing logic.
    * This longer term solution will render our efforts in point 2 unnecessary

Therefore, we will:

* Use this issue to move all imports of `minio/sha256-simd` (BLS-related code) to `crypto/sha256`
* Wait for #2241 to be completed so that users can do what is suggested in this issue on app-side code

cason:
> Use this issue to move all imports of minio/sha256-simd (BLS-related code) to crypto/sha256

Opened #3071.

faddat:
> We currently don't have performance data that shows an obvious gain in the production setups where CometBFT is used

IDK how much realworldish gain there is, but:

> Support for Intel SHA Extensions

hehe this would make it a lot faster on my homebrew routers which use the J3455 cpu

> Support for AVX512

That's basically "every validator"

melekes:
> IDK how much realworldish gain there is, but:

That's the culprit of this PR 😅

melekes:
Closing as per https://github.com/cometbft/cometbft/pull/2909#issuecomment-2107944974.

Issue #2869, opened 2024-04-23
Speedup EventBus.PublishEventTx 


ValarDragon:
### Summary

In a block sync of 5000 blocks on Osmosis mainnet, we see 23 seconds out of 1050 in EventBus.PublishEventTx (`2.1%` of the ApplyBlock time), this should likely also impact live syncing blocks as it affects ApplyBlock.

![image](https://github.com/cometbft/cometbft/assets/6440154/f2eb6389-a4c5-4c5e-ba4c-7b7d78a3778c)

Most of this time is time that doesn't even need to be spent!

### Proposal

- [x] First notice that we spend 10.4 seconds in zerologger.With which is only [here](https://github.com/cometbft/cometbft/blob/main/types/event_bus.go#L182)! However in the function we pass it into [here](https://github.com/cometbft/cometbft/blob/main/types/event_bus.go#L117), that time is _only_ used in Debug logs. So we should just refactor the API's to only pay that cost if we are in debug mode. We just need to refactor that API to pass the base logger in, and only do the With call as a LazyLogger I think, that also only gets called if we are in that error case.
- [ ] The fmt.Sprintf call taking 3.5 seocnds [here](https://github.com/cometbft/cometbft/blob/main/types/event_bus.go#L127) should be able to live in a cache, since there should be a small number of composite tags. Perhaps we add an LRU cache of size (say) 20000 to cache these? 

We should be able to speed this up  notably more beyond these two improvements, but figured it may be worth writing up these two and getting it deployed in benchmarks first

adizere:
closed by #2911

ValarDragon:
No we should re-open this, that PR only closed the first checkbox of this issue!

andynog:
Re-opened as per @ValarDragon comment

PR #2911, opened 2024-04-27
perf(event bus): Remove expensive Logger debug call in PublishEventTx

ValarDragon:
Component of #2869 

This on its own is an expected 1% speedup to blocksync on osmosis mainnet right now.

I originally considered keeping the log lines but with only creating the logger cost if there is an error, but these are debug logs I've never seen. I think its better to just remove these debug logs directly, rather than worry about maintaining them. These aren't even that concerning scenarios I feel like, as more of the stack moves away from these. 

---

#### PR checklist

- [x] Tests written/updated - Covered by existing tests


melekes (approved):
Thanks @ValarDragon ❤️

andynog (approved):
Looks good to me, I don't see any value on logging these messages and seems that removing them will only benefit performance. Thanks @ValarDragon

PR #2924, opened 2024-04-29
perf(state): Cache the block hash

### LINKED ISSUE #2923: Cache computation of the block hash in state
### Summary

We re-compute the block hash several times during consensus. We should just cache this block hash computation. E.g. each `HashesTo` call recreates the block's hash.

Here is the time spent in block hashing in a node synced to osmosis mainnet for 1 hour: 
image

This is by no means the bottleneck, but is part of the sequential bottleneck of consensus and would be good to reduce, via only computing it once per block.

## Proposal

I suggest that we cache this by adding a `verifiedHash` field to the internal `Block` type.

ValarDragon:
Closes #2923 

Caches the block hash to ensure we only compute it once in consensus execution.

---

#### PR checklist

- [x] Tests written/updated - not sure what test I should write, any suggestions?

### DIFF
--- a/types/block.go
+++ b/types/block.go
@@ -44,10 +44,11 @@ const (
 type Block struct {
 	mtx cmtsync.Mutex
 
-	Header     `json:"header"`
-	Data       `json:"data"`
-	Evidence   EvidenceData `json:"evidence"`
-	LastCommit *Commit      `json:"last_commit"`
+	verifiedHash cmtbytes.HexBytes // Verified block hash (not included in the struct hash)
+	Header       `json:"header"`
+	Data         `json:"data"`
+	Evidence     EvidenceData `json:"evidence"`
+	LastCommit   *Commit      `json:"last_commit"`
 }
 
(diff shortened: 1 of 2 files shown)


melekes (approved):
Thanks @ValarDragon ❤️

hvanz (approved):

cason (commented):
Do you know why we need a mutex here?

types/block.go
```diff
@@ -44,10 +44,11 @@ const (
 type Block struct {
 	mtx cmtsync.Mutex
 
-	Header     `json:"header"`
-	Data       `json:"data"`
-	Evidence   EvidenceData `json:"evidence"`
-	LastCommit *Commit      `json:"last_commit"`
+	verifiedHash cmtbytes.HexBytes // Verified block hash (not included in the struct hash)
```
cason, line 47: You could skip a line, to separate private from public field, just a nit.

ValarDragon:
I don't know why the mutex is needed

Issue #2854, opened 2024-04-20
enterPrecommit is almost always double verifying the block


ValarDragon:
### Summary

enterPrecommit almost always runs `ValidateBlock`, however in most cases this is duplicate logic as we have already prevoted the block. We only enter a lock after this line! This is the line where we run validate block: https://github.com/cometbft/cometbft/blob/main/internal/consensus/state.go#L1667

I suggest that we change this line [here](https://github.com/cometbft/cometbft/blob/main/internal/consensus/state.go#L1667), to be one of two things:
- Only validate the block after checking some cache for whether we have already validated this block.
- Check  "did we prevote this block", and only ran ValidateBlock if we did not.

 ### Problem Definition

We see we have duplicate VerifyBlock logic on synced full nodes:
![image](https://github.com/cometbft/cometbft/assets/6440154/f7f980ec-e6b9-4155-9299-99c003db4651)

This operation is expensive, especially in consensus critical contexts!

### Proposal

I think that making a cache for checking if we've already validated a block is better, as it seems easier to also avoid extra redundant validations for `finalizeCommit`. I haven't read that logic, but guessing by how all 3 of these validate block calls are pretty similar in the amount of time spent, I'm guessing its a redundant ValidateBlock as well.

cason:
We had a discussion at some point regarding the multiple validations of the same block. I found this comment:

https://github.com/cometbft/knowledge-base/pull/14#issuecomment-1708062228

cason:
A similar discussion also here: https://github.com/cometbft/cometbft/pull/1391

cason:
We need to separate the validation of a block by the state-machine/blockchain, and the validation of a block by the application. The first validation, whether a block can be appended to the blockchain, could indeed be cached.

ValarDragon:
Just checked, we are indeed also doing an extra validate Block call in FinalizeCommit, that would be mitigated with a simple cache :)

ValarDragon:
Ah we are actually double validating the block in FinalizeCommit -- though this one is super easy to fix using ApplyVerifiedBlock:
![image](https://github.com/cometbft/cometbft/assets/6440154/19029008-7156-4959-b1cc-66c12e0f9ba5)

sergio-mena:
Hi @ValarDragon @cason chiming in here

First of all, the caching mechanism that @ValarDragon is proposing shouldn't be hard to do, and would probably do the trick. So this is still and option.

[...]

PR #2928, opened 2024-04-29
perf(consensus/state): Change finalizeCommit to use applyVerifiedBlock

ValarDragon:
Simplest component of #2854 

We already run ValidateBlock in finalizeCommit, so this PR removes one extra redundant call by using ApplyVerifiedBlock. (The other call is also redundant, but that likely requires a more complex caching strategy as noted in #2854 to remedy)

From my understanding of these benchmarks, at Osmosis 150 validators, this should be saving ~13ms of execution time per block. 

---

#### PR checklist

- [X] Tests written/updated - Theres no test to update, as its impossible to reach the difference in codepaths! 


melekes (approved):
Thanks @ValarDragon ❤️

ValarDragon:
Thank you for all the reviews @melekes !

cason:
I've just noticed that we don't use `ApplyBlock` any longer in the code, except for `internal/consensus/replay.go`. Maybe it is the case of updating its documentation and usage?

Issue #2867, opened 2024-04-22
jsonrpc: add a batch size limit for RPC requests


cason:
Our homemade RPC implementation, [`jsonrpc`](https://github.com/cometbft/cometbft/tree/main/rpc/jsonrpc), allows users to invoke multiple RCP methods using a single request, using batching. While this is not a problem, it may constitute a form of attack vector.

By investigating the code, batching is implemented by this method:

https://github.com/cometbft/cometbft/blob/6da2cf396c145081ab534cd9b72810a531812a14/rpc/jsonrpc/server/http_json_handler.go#L41

Not sure if this is the only method implementing batched requests, but this one was identified by running a test client.

We should consider adding a configuration parameter to limit the size/length of batched RPC requests submitted to Comet. To preserver backwards compatibility, the default value for this parameter can be `0`, meaning no limit.

cason:
This solution was adopted in a fork, using a hard-coded batch size limit:

https://github.com/sei-protocol/sei-tendermint/blob/5e8e9a2c60a3989e48b600cebe524c475d34bed4/rpc/jsonrpc/server/http_json_handler.go#L46-L50

ValarDragon:
I'm very supportive of doing this w/ a config option!

andynog:
I'll look into this and try to push a PR in the next couple of days

PR #2939, opened 2024-04-29
feat: add a configurable maximum batch size for RPC requests

andynog:
close: #2867 

Adds middleware for the JSONRPC server to enforce a configurable maximum batch size for RPC requests. 

---

#### PR checklist

- [ ] Tests written/updated


andynog:
@melekes, do you think that it would be helpful if I implemented some refactoring for this handler
https://github.com/cometbft/cometbft/blob/ea1f89633903ac0ef3d02127e0c684cc7fa58e23/rpc/jsonrpc/server/http_server.go#L263

I thought of combining this into just one middleware. I think doing this in `pre-check` middleware might also be more beneficial since if the body is over the max size then you break before processing anything. please let me know what you think.

melekes:
> @melekes, do you think that it would be helpful if I implemented some refactoring for this handler
> 
> https://github.com/cometbft/cometbft/blob/ea1f89633903ac0ef3d02127e0c684cc7fa58e23/rpc/jsonrpc/server/http_server.go#L263
> 
> I thought of combining this into just one middleware. I think doing this as a `pre-check` middleware might also be more beneficial since if the body is over the max size then it breaks before processing anything. please let me know what you think.

sound good to me 👍

andynog:
This should be ready to review @melekes, implemented a test too.

I've tagged it to be back ported to `v1`, not sure if it should be back ported to others. I think it could without breaking changes. Please let me know what you think.

rpc/jsonrpc/server/http_server.go
```diff
@@ -277,3 +280,49 @@ func Listen(addr string, maxOpenConnections int) (listener net.Listener, err err
...
+				if len(requests) > config.MaxRequestBatchSize {
+					res := types.RPCInvalidRequestError(nil, fmt.Errorf("batch request exceeds maximum (%d) allowed number of requests", config.MaxRequestBatchSize))
+					responses = append(responses, res)
+					_ = WriteRPCResponseHTTP(w, responses...)
+					return
+				}
+			}
+
+			// ensure the request body can be read again by other handlers
+			r.Body = io.NopCloser(bytes.NewBuffer(data))
```
melekes, line 322: shouldn't that be `r.Body = http.MaxBytesReader(w, r.Body, config.MaxBodyBytes)` otherwise we lose max body bytes restriction, no?

config/config.go
```diff
@@ -442,8 +446,9 @@ func DefaultRPCConfig() *RPCConfig {
 		TimeoutBroadcastTxCommit:  10 * time.Second,
 		WebSocketWriteBufferSize:  defaultSubscriptionBufferSize,
 
-		MaxBodyBytes:   int64(1000000), // 1MB
-		MaxHeaderBytes: 1 << 20,        // same as the net/http default
+		MaxRequestBatchSize: 0,              // no maximum requests in a batch request
```
melekes, line 449: is this a good default?

melekes (commented):
Thanks @andynog ❤️

config/config.go
andynog: I had the default initially set to 10, but revert to 0 because @cason in the issue for this PR mentioned we should keep it at 0 for backwards compatibility (no restrictions). 

But that might not do much good, probably we can enforce a maximum and people will have to change to 0 if they don't want the limit. In my experience, most operators just leave the default values and that might not be beneficial. 

Should I change the default value to something like 10 ? Usually there's another limit for max body request bytes but that's also very high by default so it's hard to trigger. Wdyt?

rpc/jsonrpc/server/http_server.go
andynog: I think the first line will catch that because it's outside the `if`, but I will do some test to confirm. Thanks for pointing this out.

config/config.go
adizere: Agree we should constrain the value by default, say to `10`. Since we're introducing a new parameter to configure this behavior, operators are free to mimic the old behavior by setting `0` if they want.

adizere:
Makes sense to backport to 0.38 IMO.

config/config.go
andynog: OK, I'll revert back to 10, makes sense.

andynog: reverted

andynog:
If you want to test this locally you can:

- `make install`
- `cometbft init`
- `vim ~/.cometbft/config/config.toml`
- change `max_request_batch_size = 1`
- `cometbft start --proxy_app=kvstore`
- on another terminal, make a `curl` JSON-RPC batch request sending two requests in the batch:
```
curl --location --request GET 'http://localhost:26657/' \
--header 'Content-Type: application/json' \
--data '[
    {
        "jsonrpc": "2.0",
        "method": "net_info",
        "params": {},
        "id": 1
    },
    {
        "jsonrpc": "2.0",
        "method": "net_info",
        "params" : {},
        "id": 2
    }
]'
```
- you should get an error:
```
{"jsonrpc":"2.0","error":{"code":-32600,"message":"Invalid Request","data":"batch request exceeds maximum (1) allowed number of requests"}}%
```

melekes (approved):
👍

PR #2949, opened 2024-04-30
perf(p2p/channel): Speedup NewDelimitedWriter

ValarDragon:
Speeds up 5% of the non-IO time overhead from `channel.WritePacketMsgTo`. The CPU time overhead in this function is quite significant, CPU time is more than 3 times the syscall time for writing to the net buffer. Working on a github issue for more substantial refactor / time eliminations, but this 3s is easy enough.

We don't even use this codepath, so this make slice is entirely wasted. However we should do things that reduce this overhead further.

image

---

#### PR checklist

- [x] Tests written/updated


melekes (approved):
Thanks @ValarDragon ❤️

melekes:
> We don't even use this codepath, so this make slice is entirely wasted

Are you saying we're creating delimited writer, but do not call WriteMsg? That doesn't make sense

PR #2950, opened 2024-04-30
perf(p2p): Remove unnecessary atomic read

ValarDragon:
This PR is a driveby change. We were doing an atomic read here, but this is unnecessary. Nothing in the codebase modified this variable after instantiation.

Doesn't really feel changelog worthy, but I can add one if we want. Felt like a minor code nit

### DIFF
--- a/p2p/conn/connection.go
+++ b/p2p/conn/connection.go
@@ -512,7 +512,7 @@ func (c *MConnection) sendSomePacketMsgs() bool {
 	// Block until .sendMonitor says we can write.
 	// Once we're ready we send more than we asked for,
 	// but amortized it should even out.
-	c.sendMonitor.Limit(c._maxPacketMsgSize, atomic.LoadInt64(&c.config.SendRate), true)
+	c.sendMonitor.Limit(c._maxPacketMsgSize, c.config.SendRate, true)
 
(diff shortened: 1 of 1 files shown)


melekes (approved):
Thanks @ValarDragon ❤️

sergio-mena:
Hi, getting here a bit late...

I've checked the call hierarchy of `sendSomePacketMsgs` and I see that, ultimately, it's being called from two different goroutines (potentially for the same `MConnection`):

* It's called from `MConnection.FlushStop` <-- `pex.Reactor.Receive` (actually, an ad-hoc goroutine there)
* It's called from `MConnection.sendRoutine`, which is run as a goroutine

In principle, we could suspect there might be a race between `FlushStop` and `sendRoutine`. However, `FlushStop` waits on `MConnection.doneSendRoutine` before calling sendSomePacketMsgs, so we should be good.

Posting this here to further document this change.

cason:
Just for context, calling `FlushStop` more than once is not a problem. The method startswith a call to `stopServices`. This method is protected by a lock and once the first execution of this method completes, all the following will return `true`. When this method returns `true`, the `FlushStop` method is a no-op.

PR #2953, opened 2024-04-30
refactor(p2p/connection): Slight refactor to sendManyPackets that helps highlight performance improvements

ValarDragon:
Fully compatible refactor to the `sendSomePacketMsgs` loop, that I think makes it easier to:
1) Understand the code
2) See how we improve it for serial time performance
3) Puts my notes for how to improve some consensus gossip bottlenecks.

---

#### PR checklist

- [x] Tests written/updated - N/A, this is fully compatible so no new test updates needed
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)  - I dont think refactors get changelogs?
- [x] Updated relevant documentation (`docs/` or `spec/`) and code comments
- [x] Title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) spec

Issue #2955, opened 2024-04-30
Increase the Priority weight for Vote and BlockPart channels


ValarDragon:
### Summary

We should increase the priority weight for Vote and BlockPart channels, to ensure these packets at consensus time get prioritized over mempool tx gossip.

Right now the priorities are:
- Vote: 7
- Data: 10 (Aka Blockpart)
- Mempool: 5

I suggest starting such a move by making it Vote: 21, and Data: 30.

This unfortunately would increase the prioritization of catching up peers who are behind, until that is split out of the Data channel. I think thats worthwhile for increasing prioritization of consensus messages.

PR #2957, opened 2024-04-30
perf(p2p): Change channel.recentlySent to be based on num_messages

### LINKED ISSUE #2954: Make channel prioritization based on number of messages in channel (vs number of bytes sent)
### Summary

`Channel.RecentlySent` is used for choosing priority of a channel. Currently channel selection is done as channel with the minimum `channel.RecentlySent / channel.Priority` ratio. Every packet message increases `channel.RecentlySent` by the proto-encoded packet size. `channel.RecentlySent` gets lowered by 20% at every `UpdateStats` window (which is just a 2 second timer). 

BlockGossip is consensus time critical, and has very large packets. Its current priority weight is 10, vs vote's 7, hasVote's 1, and mempool's 5. Its packets are block parts which are large. This guarantees block gossip will always be the lowest prioritized packet out of all packets, since its recently sent will always be very large (in the 100kb's), which is problematic for fast block gossip. Any of the vote messages or mempool messages to a peer will outcompete block part gossip in the prioritization. 

This matches behavior observed that we never hit our bandwidth flow rates in production and that block gossip just occurs slowly.

Instead channel.RecentlySent should first off be based on the number of packets on a channel, not the bandwidth size.

### Problem Definition

This should greatly increase the prioritization of block gossip.

### Proposal

channel.RecentlySent should first off be based on the number of packets on a channel, not the bandwidth size.

ValarDragon:
Closes #2954 

This should make block part gossip not be bottom priority. (though more work is needed to make it match the prioritization expectation)

---

#### PR checklist

- [x] Tests written/updated - there is actually no tests here for prioritization, just eventual delivery tests. Should be easy to add tests after #2953 

### DIFF
--- a/.changelog/unreleased/improvements/2957-improve-channel-prioritization.md
+++ b/.changelog/unreleased/improvements/2957-improve-channel-prioritization.md
@@ -1,0 +1,3 @@
+- `[p2p]` Improve the prioritization of which channel to gossip from, by changing channel deprioritization
+  to be based on the number of messages in the channel, rather than the size of messages communicated on the channel.
+  ([\#2928](https://github.com/cometbft/cometbft/pull/2928))
--- a/p2p/conn/connection.go
+++ b/p2p/conn/connection.go
@@ -31,7 +31,9 @@ const (
 	numBatchPacketMsgs = 10
 	minReadBufferSize  = 1024
 	minWriteBufferSize = 65536
-	updateStats        = 2 * time.Second
+
+	updateStats                  = 2 * time.Second
+	recentlySentExponentialDecay = 0.8
 
 	// some of these defaults are written in the user config
 	// flushThrottle, sendRate, recvRate
@@ -755,7 +757,10 @@ type Channel struct {
 	sendQueueSize int32 // atomic.
 	recving       []byte
 	sending       []byte
-	recentlySent  int64 // exponential moving average
+	// recentlySent is an exponential moving average of the number of messages sent.
+	// This is used by the mconnection for choosing what channel to send from next.
+	// The exponential decay is done in updateStats.
+	recentlySent int64
 
 	maxPacketMsgPayloadSize int
 
@@ -856,7 +861,8 @@ func (ch *Channel) writePacketMsgTo(w io.Writer) (n int, err error) {
 		err = ErrPacketWrite{Source: err}
 	}
 
-	atomic.AddInt64(&ch.recentlySent, int64(n))
+	// increment recentlySent by 1, to denote we sent another message.
+	atomic.AddInt64(&ch.recentlySent, 1)
 	return n, err
 }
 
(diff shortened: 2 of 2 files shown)


melekes (approved):
Thanks @ValarDragon ❤️

melekes:
We'll need to test this somehow through our e2e framework or local net to avoid relying on pure reasoning.

ValarDragon:
Agreed, I also think that in practice this may also need #2955

.changelog/unreleased/improvements/2957-improve-channel-prioritization.md
```diff
@@ -0,0 +1,3 @@
+- `[p2p]` Improve the prioritization of which channel to gossip from, by changing channel deprioritization
```
cason, line 1:
```suggestion
- `[p2p]` Change the prioritization of which channel to send packets from, by having de-prioritization
```

.changelog/unreleased/improvements/2957-improve-channel-prioritization.md
```diff
@@ -0,0 +1,3 @@
...
+  to be based on the number of messages in the channel, rather than the size of messages communicated on the channel.
```
cason, line 2:
```suggestion
  to be based on the number of messages rather than the byte size of packets recently sent on the channel.
```

p2p/conn/connection.go
```diff
@@ -755,7 +757,10 @@ type Channel struct {
 	sendQueueSize int32 // atomic.
 	recving       []byte
 	sending       []byte
-	recentlySent  int64 // exponential moving average
+	// recentlySent is an exponential moving average of the number of messages sent.
+	// This is used by the mconnection for choosing what channel to send from next.
```
cason, line 761:
```suggestion
	// This is used for choosing from which channel to send the next packet.
```

p2p/conn/connection.go
```diff
@@ -755,7 +757,10 @@ type Channel struct {
...
+	// The exponential decay is done in updateStats.
```
cason, line 762:
```suggestion
	// The exponential decay is applied every updateStats inverval.
```

p2p/conn/connection.go
```diff
@@ -856,7 +861,8 @@ func (ch *Channel) writePacketMsgTo(w io.Writer) (n int, err error) {
 		err = ErrPacketWrite{Source: err}
 	}
 
-	atomic.AddInt64(&ch.recentlySent, int64(n))
+	// increment recentlySent by 1, to denote we sent another message.
```
cason, line 864: But this is not a message, but a packet. A message is composed of multiple packets.

p2p/conn/connection.go
```diff
@@ -31,7 +31,9 @@ const (
 	numBatchPacketMsgs = 10
 	minReadBufferSize  = 1024
 	minWriteBufferSize = 65536
-	updateStats        = 2 * time.Second
+
+	updateStats                  = 2 * time.Second
```
cason, line 35:
```suggestion
	recentlySentDecayInterval                  = 2 * time.Second
```

cason: Need to rename elsewhere in the code.

cason (commented):
The problem is not solved by this PR.

The original code compute priorities based on the number of bytes sent in each channel.

The proposed changes update the computation to be based on the number of packets sent in each channel. While it is true that not every packet is full (i.e., contains `MaxPacketMsgPayloadSize`, default to 1024 bytes), when full packets are exchanged, there is not difference in behavior.

In any case, messages are split into multiple packets to be sent in the channel. If a message is 10x larger than the other, it would be split into about 10x more packets than the other, and the proposed changes won't have any real effect.

PR #2959, opened 2024-05-01
perf(internal/bits): Speedup extended commit.BitArray()

ValarDragon:
Speedup ExtendedCommit.BitArray() by making a direct constructor that does not go through mutexes. I expect this to be a 10x performance improvement. (It also removes the duffcopies and reduces setIndex proportion of time here)

This removes this time (which is mostly coming from mutex calls): 
image

Later on we should instead make an API that lets us randomly sample from a bit array with no bit array copying needed. (But that makes the interface messier) 

---

#### PR checklist

- [x] Tests written/updated


melekes (approved):
Thanks @ValarDragon ❤️

PR #2960, opened 2024-05-01
perf(consensus): align calls to `ValidateBlock`, `ProcessProposal` and `isTimely`

sergio-mena:
~~This change is an implementation of the alignment of `ValidateBlock` to `ProcessProposal` explained [here](https://github.com/cometbft/cometbft/issues/2854#issuecomment-2088261899).~~

We need to:

* Check if that helps, perf-wise
* Carefully make we're not compromising algorithm's correctness

EDIT:

In the end, we are implementing this as agreed [here](https://github.com/cometbft/cometbft/issues/2854#issuecomment-2107159921)

EDIT2:

This PR will just deal with alignment of `ValidateBlock`, `ProcessProposal` and `isTimely` (so it's no longer closing #2854)

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/internal/consensus/state.go
+++ b/internal/consensus/state.go
@@ -1427,7 +1427,7 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
 					return
 				}
 
-				if cs.Proposal.POLRound == -1 && !cs.proposalIsTimely() {
+				if !cs.proposalIsTimely() {
 					lowerBound, upperBound := cs.timelyProposalMargins()
 					// TODO: use Warn level once available.
 					logger.Info("prevote step: Proposal is not timely; prevoting nil",
@@ -1440,12 +1440,10 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
 					return
 				}
 
-				if cs.Proposal.POLRound == -1 {
-					logger.Debug("prevote step: Proposal is timely",
-						"timestamp", cs.Proposal.Timestamp.Format(time.RFC3339Nano),
-						"receive_time", cs.ProposalReceiveTime.Format(time.RFC3339Nano),
-						"timestamp_difference", cs.ProposalReceiveTime.Sub(cs.Proposal.Timestamp))
-				}
+				logger.Debug("prevote step: Proposal is timely",
+					"timestamp", cs.Proposal.Timestamp.Format(time.RFC3339Nano),
+					"receive_time", cs.ProposalReceiveTime.Format(time.RFC3339Nano),
+					"timestamp_difference", cs.ProposalReceiveTime.Sub(cs.Proposal.Timestamp))
 			}
 
 			// Validate proposal block, from consensus' perspective
--- a/internal/consensus/state.go
+++ b/internal/consensus/state.go
@@ -1414,16 +1414,6 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
 		}
 	}
 
-	// Validate proposal block, from consensus' perspective
-	err := cs.blockExec.ValidateBlock(cs.state, cs.ProposalBlock)
-	if err != nil {
-		// ProposalBlock is invalid, prevote nil.
-		logger.Error("prevote step: consensus deems this block invalid; prevoting nil",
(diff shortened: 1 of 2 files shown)


internal/consensus/state.go
```diff
@@ -1458,6 +1448,16 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
 				return
 			}
 
+			// Validate proposal block, from consensus' perspective
+			err := cs.blockExec.ValidateBlock(cs.state, cs.ProposalBlock)
```
sergio-mena, line 1452: @cason if you agree that moving this here aligns the code with the algorithm (so the code stays correct), would it also be worth moving PBTS's `isTimely` check (the whole _if_ in line 1389 above) here?

I know that it wouldn't probably change anything perf-wise, as checking for `isTimely` is cheap, but I now think the code is clearer if all three checks (`ValidateBlock`, `ProcessProposal`, and `isTimely`) are at the same place

adizere:
Removed from project board & Milestone b/c it's tracked already under https://github.com/cometbft/cometbft/issues/2854

internal/consensus/state.go
cason: The corner case that remains is when  `cs.Proposal.POLRound == -1` and `cs.LockedRound == -1` and `cs.ProposalBlock.HashesTo(cs.ValidBlock.Hash())`. This is an optimization for when a block has became valid but we didn't lock it (we receive the 2/3+ prevotes after issuing our precommit).

Actually, I don't even see how this situation can happen. The situation is that a block has received a Polka in a round, then in a future round a propose that _hasn't_ seen this Polka, proposes the same block as a new block (since  `cs.Proposal.POLRound == -1`). This is then the case of two nodes producing the same block "by chance" in different rounds...

sergio-mena:
See [this comment](https://github.com/cometbft/cometbft/issues/2854#issuecomment-2107159921) for an explanation of the latest commits

state/execution.go
```diff
@@ -194,9 +197,11 @@ func (blockExec *BlockExecutor) ProcessProposal(
...
-	err := validateBlock(state, block)
-	if err != nil {
-		return err
+	if !blockExec.lastValidatedBlock.HashesTo(block.Hash()) {
+		if err := validateBlock(state, block); err != nil {
+			return err
+		}
+		blockExec.lastValidatedBlock = block
 	}
 	return blockExec.evpool.CheckEvidence(block.Evidence.Evidence)
```
cason, line 206: We are still running this all the times.

state/execution.go
```diff
@@ -41,6 +41,9 @@ type BlockExecutor struct {
 	mempool mempool.Mempool
 	evpool  EvidencePool
 
+	// 1-element cache of validated blocks
+	lastValidatedBlock *types.Block
```
cason, line 45: I wonder whether we should reset this field (to nil) when committing a block.

cason (commented):
This PR now has 3 contributions:

1. On prevote step, only validate blocks when they are new (`POLRound == -1`) and we are not locked
2. The same for the timely validation performed as part of PBTS
3. Minimal chaching mechanism for block validation

I think the PR should implement 1. and 2., with description updated to reflect this change. In this case it is not really a performance improvement, as there is no change or benefit in the good case (single round).

I would leave 3. for another PR.

state/execution.go
sergio-mena: Yes, but there's no evidence in the happy path, which is what we are concerned with here

cason (approved):
Good refactoring :)

internal/consensus/state.go
```diff
@@ -1842,10 +1837,6 @@ func (cs *State) finalizeCommit(height int64) {
 		panic("cannot finalize commit; proposal block does not hash to commit hash")
 	}
 
-	if err := cs.blockExec.ValidateBlock(cs.state, block); err != nil {
```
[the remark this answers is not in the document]
sergio-mena: 👍 changed approach to simplistic cache

PR #2962, opened 2024-05-01
refactor(state): reuse method `proposalIsTimely`

sergio-mena:
Small refactor to simplify the calls to "is timely"

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

cason:
Shouldn't this go to v1 too?

sergio-mena:
> Shouldn't this go to v1 too?

It's just a small refactor. I don't think it's worth the hassle

PR #2964, opened 2024-05-01
perf(consensus/blockstore): Remove validate basic call from LoadBlockMeta

ValarDragon:
Our gossip block parts routine calls `Blockstore.LoadBlockMeta` (as do things in blocksync). This currently takes some time due to ValidateBasic:
image

However note that we only save validated data to the blockstore. We only do it in:
- Commit: https://github.com/cometbft/cometbft/blob/main/internal/consensus/state.go#L1867
- Blocksync after validation: https://github.com/cometbft/cometbft/blob/main/internal/consensus/state.go#L1867

Hence the validate basic time is wasted.

This should eventually just go to an LRU cache to even avoid the proto unmarshalling (as should ~everything in blockstore) but we want this anyway to reduce the computational overhead,

WRT live consensus, in main this only helps with catchup for live syncing nodes, I haven't checked on v0.47.x, but the cpuprofile suggests it may help with active block gossip time as well?

---

#### PR checklist

- [x] Tests written/updated


melekes (approved):
Thanks @ValarDragon ❤️

Issue #2655, opened 2024-03-18
PBTS: Allow Application to Set BlockTimestamp


itsdevbear:
### Summary

Useful in cases where the application logic relies on block times adhering to some property. 

### Proposal

Create a "PreparePrepareProposal" hook of some kinda.

adizere:
hey Dev, thanks for opening this!

is it fair to say that the underlying concern you'd like to address is ensuring that CometBFT produces blocks at a predictable (ideally fixed) rate, say 1 block every 5 second. so even if blocks are empty, Comet should produce a block every 5 seconds, say.

this model was discussed by the previously team (ref https://github.com/tendermint/tendermint/issues/5911) and it's part of a broader block timeouts UX improvements -- discussion here https://github.com/cometbft/cometbft/discussions/2266

itsdevbear:
yeah @adizere you got it!

cason:
So, the underlying goal here is consensus to wait until the time returned by the application before proposing the block?

lasarojc:
What if prepareProposal also let the block time to be modified? And prepareProposal could just wait to return when it decides the time is good

itsdevbear:
^ this behaviour would actually work for us very well

itsdevbear:
We should be able to do this in a similar way to how we did the timeout commit.

PR #2966, opened 2024-05-02
docs: ADR-115 predictable block times

melekes:
[RENDERED](https://github.com/cometbft/cometbft/blob/2655-predictable-block-times/docs/references/architecture/adr-115-predictable-block-times.md#adr-115-predictable-block-times)

Refs #2655

---

#### PR checklist


ValarDragon:
Making this a network parameter is great! I fully support doing that!

I actually think we can do one-step better as well though. Right now a block's time is roughly set by 4 steps
---------- Start time counting
* Proposer(N+1) receives 2/3rd precommits for block N
* Proposer Executes + Commits block N
* Proposer waits Timeout_commit
* Proposer builds block N + 1 and gossips it
* 2 rounds of 2/3rds voting amongst validator set
* Proposer(N+2) recieves 2/3rds precommits for block N+1
---------- Stop counting

Each of the steps could have variable delay for whatever reason. It would be nice if we made timeout_commit react to the time-since-last-block.

In PBTS, we could do this by saying, block N+1 proposer should wait `max(true_minimum_timeout_commit, current_time - block N timestamp - expected_block_time)`.

And then we lower the `true_minimum_timeout_commit` to perhaps something like `.5s` by default, and have networks parameterize around the expected block time.

tac0turtle (commented):
super glad to see this work taking shape.

melekes:
> In PBTS, we could do this by saying, block N+1 proposer should wait max(true_minimum_timeout_commit, current_time - block N timestamp - expected_block_time).

What I like about the current proposal is that it doesn't restrict the ABCI dev in any way. There's no `max(x, y)`, no min. The ABCI dev has access to all the info (I believe), so no one is stopping him/her from calculating `max(true_minimum_timeout_commit, current_time - block N timestamp - expected_block_time)` if he/she desires to do so.

ValarDragon:
Oh true, that sounds great to me!

ValarDragon (approved):

docs/references/architecture/adr-115-predictable-block-times.md
```diff
@@ -0,0 +1,113 @@
...
+Let's say you're an ABCI application developer and you want to have
+predictable block times: 1 block each 6s. How would you do that?
+
+You found out that proposing a block and voting on it takes roughly 2s in your
+network. You change `timeout_commit` (_how long a node waits for additional
+precommits after committing a block, before starting on the next height_) from
+1s to 4s in the default config and ship your app with that modified config. Do
+you have predictable block times now?
+
+No - you don't. The block time will be around 6s, but it won't be exactly 6s. A
```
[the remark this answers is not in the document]
cason: The idea behind "favorable runs" is that we are in the good case, no additional delays, no failures, namely timely decisions in round 0.

docs/references/architecture/adr-115-predictable-block-times.md
```diff
@@ -0,0 +1,113 @@
...
+   block time.
+2. Network latency.
+3. Clock drifts.
+
+By now, many teams (Celestia, Berachain, Osmosis, etc.) expressed their
+interest in either a) having predictable block times OR b) having the ability
+to alter the next block time.
+
+In case of Osmosis, committing a big block might take longer than usual. They
+could increase the block time for big blocks to give the state machine some
```
cason, line 39: At proposal time, you mean?

docs/references/architecture/adr-115-predictable-block-times.md
```diff
@@ -0,0 +1,113 @@
...
+extra time to finish execution if that was possible.
+
+In case of Berachain, they want to have a predictable block time to match
+Ethereum's slots, which are equal to 12s.
+
+## Decision
+
+Remove `timeout_commit` in favor of a new field in `FinalizeBlockResponse`:
+`next_block_delay`. This field's semantics stays essentially the same: delay
+between this block and the time when the next block is proposed.
```
cason, line 49: What you mean by "between this block". When this interval starts? When `FinalizeBlock` returns?

docs/references/architecture/adr-115-predictable-block-times.md
```diff
@@ -0,0 +1,113 @@
...
+
+If Comet goes into multiple rounds of consensus, the ABCI application can react
```
cason, line 51: Not sure if the application is aware of the commit round. In any case, different nodes may commit at different rounds.

cason (commented):
I think that the ADR can be extended/reworded as follows:

1. The problem description should make it more precise what we mean by predictable block times. If I understand correctly, we want the medium-to-long term average block time to converge to a desired value. To achieve that, we need to define a form of minimum block time, but also be able to changing it dynamically. Namely, if a block takes longer than expected, we should be able to render the next block(s) faster in order to converge to the desired (average) block time.
1. Properly define how `timeout_commit` works, at which extent it can be adopted to achieve the predictable block times goal, and what are its limitations. In addition to the one explained in the text (it is not a global parameter), we have the fact that it is static.
1. There is a confusion in the text, that I assume also comes from the confusing in the operation of `timeout_commit`, between block time and the delay between committing a block and proposing the next block. And this is a main concern: we cannot precisely predict the consensus delay for a block, but we can establish a minimum duration for the processing step of a block, i.e., the time from which consensus decides it and the next consensus instance effectively starts.
1. We need to make it clear that this approach is always best-effort, not only because delays are variable (by block size, by system size, by network conditions), but also because we have nodes that don't follow the protocol. By the way, running a validator node with a `timeout_commit` parameter that differs from the one defined for the network is a form of misbehavior.

docs/references/architecture/adr-115-predictable-block-times.md
```diff
@@ -0,0 +1,113 @@
...
+validator could set `timeout_commit`  to 0s because it's a local parameter,
+which is not enforced by Comet. That means whenever this validator proposes a
+block, the overall block time will be 2s (not 6s!).
+
+There are other reasons why the block time might not be predictable:
+
+1. CometBFT can go into multiple rounds of consensus, which will increase the
+   block time.
+2. Network latency.
+3. Clock drifts.
```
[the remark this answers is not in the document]
melekes: Yes

docs/references/architecture/adr-115-predictable-block-times.md
melekes: > different nodes may commit at different rounds.

could you please explain?

melekes: decided_last_commit https://docs.cometbft.com/v0.38/spec/abci/abci++_methods#finalizeblock ?

cason: A node can commit at round 0, other might decide the same block on block 1. This happens if part of the nodes don't see the precommits of round 0 timely and proceed to round 1.

The canonical commit, in fact, is the one observed by the proposer (for the previous block) and added to the proposed block.

melekes: > Not sure if the application is aware of the commit round.

does the ABCI dev need it? They can simply use real time, no?

cason: Yes, it should use real time, in particular because the duration of rounds can be pretty variable.

cason: But the problem of real time is, of course, that different nodes will process a block at (slightly) different times.

docs/references/architecture/adr-115-predictable-block-times.md
```diff
@@ -76,8 +89,8 @@ message FinalizeBlockResponse {
 }
 ```
 
-Block proposals received before last block time + `next_block_delay` MUST be rejected.
-A proposer MUST wait until last block time + `next_block_delay` to propose a block.
+Block proposals received before the last block is committed + `next_block_delay` MUST be rejected.
+A proposer MUST wait until the last block is committed + `next_block_delay` to propose a block.
```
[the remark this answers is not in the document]
melekes: > Plus, if the application ignores this field, would it have the same effect as setting `timeout_commit = 0` right now?

yes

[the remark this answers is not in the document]
melekes: > So, and this should be clearer in this document, the idea is literally to have the same behavior as `timeout_commit`, while allowing the application to pick a different value for each height?

yes

docs/references/architecture/adr-115-predictable-block-times.md
```diff
@@ -76,8 +89,8 @@ message FinalizeBlockResponse {
 }
 ```
 
-Block proposals received before last block time + `next_block_delay` MUST be rejected.
-A proposer MUST wait until last block time + `next_block_delay` to propose a block.
+Block proposals received before the last block is committed + `next_block_delay` MUST be rejected.
```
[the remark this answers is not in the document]
melekes: What if we use some delta?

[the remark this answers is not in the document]
melekes: I don't think we need it. Please ignore my comment. Correct validators will wait for the required time anyway before moving onto the next height.

docs/references/architecture/adr-115-predictable-block-times.md
```diff
@@ -0,0 +1,129 @@
...
+In case of Berachain, they want to have a constant block time to match
+Ethereum's slots, which are equal to 12s.
+
+## Proposal
+
+Move `timeout_commit` into `FinalizeBlockResponse` as `next_block_delay`. This
+field's semantics stays essentially the same: delay between the time when the
+current block is committed and the next height is started. The idea is
+literally to have the same behavior as `timeout_commit`, while allowing the
+application to pick a different value for each height.
```
evan-forbes, line 64: fwiw, if I'm interpreting this correctly we've tried this [here](https://github.com/celestiaorg/celestia-core/pull/965)

in our network tests, everything worked as expected and we have very consistent block times. However when we shipped this to a testnet, we frequently failed to reach consensus.

replicating and debugging this issue by changing the clocks of the network instances is next on our list of todos after we switch network testing backends https://github.com/celestiaorg/celestia-core/issues/1333

melekes: Thanks for flagging this. We will keep this in mind when implementing this feature.

evan-forbes (approved):
lgtm fwiw

docs/references/architecture/adr-115-predictable-block-times.md
sergio-mena: I have read up to here, so what I say here might be redundant with text below.
My impression is that most applications, in order to leverage this feature:
* will need to use real --wallclock-- time (having Berachain's use case in mind, where you want to conform to regular wall-clock deadlines)
* will mandate it's nodes to have synchronized clocks (NTP, or other). This is not a big deal since we also require this for PBTS

docs/references/architecture/adr-115-predictable-block-times.md
```diff
@@ -0,0 +1,133 @@
...
+
+If Comet goes into multiple rounds of consensus, the ABCI application can react
+by lowering `next_block_delay`. Of course, nobody could guarantee that there
+won't be 100000 rounds of consensus, so **it's still best effort** when it
+comes to individual block times.
+
```
sergio-mena, line 74: I think that an important thing that has not been mentioned (or I badly missed it) is that `next_block_delay` is a non-deterministic field (unlike most fields in `FinalizeBlock`), that is: it is not part of the replicated data (replicated in SMR sense). This means that each node may provide a different value, which is supposed to depend on how longs things are taking at the local node.

docs/references/architecture/adr-115-predictable-block-times.md
```diff
@@ -0,0 +1,133 @@
...
+## Alternative Approaches
+
+1. Do nothing. The block time will stay largely unpredictable.
+2. Make `timeout_commit` global (= consensus parameter). It doesn't solve the
+   problem of multiple rounds of consensus + the execution is delayed by one
+   block (as with other consensus parameters).
+3. Add `next_block_delay`, but keep `timeout_commit`. Individual validators can
```
sergio-mena, line 81: I think we'd need to keep `timeout_commit` around for at least one more release (but we deprecate it now). The reason is UX of upgrade. If an appchain doesn't care about this at the time of upgrading, they shouldn't _feel_ any pain.
A possible way of achieving this would be:
* keep `timeout_commit` deprecated
* if a value for `timeout_commit` is detected at process start-up, warn the user (WARNING log?) that they are using a deprecated field
* if the app provides a value for `next_block_delay`, then `timeout_commit` is ignored
* if the app does not provide a value for `next_block_delay`, then CometBFT falls back to `timeout_commit`

docs/references/architecture/adr-115-predictable-block-times.md
```diff
@@ -0,0 +1,133 @@
...
+### Positive
+
+- ABCI application developers will have more control over block times.
+
+### Neutral
+
+### Negative
+
+- ABCI application developers will need to set `next_block_delay` to `1s`
+  to preserve the old behavior. If the application ignores this field, it
```
sergio-mena, line 124: Check my comment on upgrading above

sergio-mena (approved):
Thanks @melekes for this ADR!

I support this way of dealing with `timeout_commit` over the alternative that has been discussed internally in the past months: moving `timeout_commit` to `ConsensusParams`. Your solution is more versatile and powerful for apps and I believe the upgrade path can be made painless (unlike the "move `timeout_commit` to `ConsensusParams`" solution).

docs/references/architecture/adr-115-predictable-block-times.md
```diff
@@ -0,0 +1,133 @@
...
+To achieve that, we need to define a form of variable block time. Namely, if a
+block takes longer than expected, we should be able to render the next block(s)
+faster in order to converge to the desired (average) block time.
+
+### Use Cases
+
+In case of Osmosis, committing a big block is expected to take longer than
+usual. If we know the approximate size of the next block, we could increase
+`timeout_commit` dynamically (if such feature existed) to give the state
+machine some extra time to finish execution.
```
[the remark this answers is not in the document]
melekes: Agree. I copied this from https://github.com/tendermint/tendermint/issues/5911#issuecomment-760968413. @ValarDragon @sunnya97 can you please clarify what you're hoping to archive in https://github.com/tendermint/tendermint/issues/5911?

docs/references/architecture/adr-115-predictable-block-times.md
melekes: Good proposal 👍

docs/references/architecture/adr-115-predictable-block-times.md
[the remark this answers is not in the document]
cason: Because `timeout_commit` encompasses the block execution.

So, faster nodes will start next height (proposing blocks or expecting for proposals) before slower nodes.

PR #2986, opened 2024-05-02
perf(p2p/connection): Lower wasted re-allocations in sendRoutine

ValarDragon:
This PR makes sending packet messages re-use the protowriter for writing to the channel, rather than remaking it in `writePacketMsgTo`. 

In a 1 hour sync benchmark, this saves 10% of the time spent in the `sendRoutine` (6s), and saves 13GB of heap allocation.

This is a simple fix, so I think its worth doing. Later on, I think we should move this proto-marshalling to `mConnection.Send`, but that change will require more robust testing, as it would be a tradeoff of increasing the CPU time of gossipVotesRoutine and gossipBlockPartRoutine. (I personally think it will be worth it / were anyways lowering the CPU time of these routines in total) Will be writing this later direction idea into an issue.

---

#### PR checklist

- [x] Tests written/updated - should be covered by existing tests
- [x] Updated relevant documentation (`docs/` or `spec/`) and code comments - I actually this is simpler/dont see anything to update


melekes (approved):

melekes:
Thanks @ValarDragon ❤️

Issue #2847, opened 2024-04-18
Don't pay vote Debug logging cost in normal operation


ValarDragon:
### Summary

Currently we spend lots of RAM allocation in Debug logs across votes. Here are some profiles of a query serving full node in its operation over 1-2 hours. We see:

addVote's StringShort calls taking 5GB of allocation (compared to the state machines total 48GB):
![Screenshot from 2024-04-18 20-55-22](https://github.com/cometbft/cometbft/assets/6440154/b4cfe07b-aac6-4df1-9f2d-c2255e334b62)
These are both debug line calls:
- https://github.com/cometbft/cometbft/blob/d27732ad9a992e321e440417a2a8220f90fe0044/internal/consensus/state.go#L2289
- https://github.com/cometbft/cometbft/blob/d27732ad9a992e321e440417a2a8220f90fe0044/internal/consensus/state.go#L2378

Peerstate.setHasVote taking 15.4GB, this is entirely in debug logging:
https://github.com/cometbft/cometbft/blob/d27732ad9a992e321e440417a2a8220f90fe0044/internal/consensus/reactor.go#L1444-L1449

### Problem Definition

This is certainly adding notable load to the system, and its not even providing utility most of the time!

### Proposal

I'm not sure what the most well abstracted way to do the logic for "if debug mode, run this log logic" is. Just wanted to point out that there is a lot were paying right now

cason:
I think this can do the job: https://github.com/cometbft/cometbft/blob/d27732ad9a992e321e440417a2a8220f90fe0044/libs/log/lazy.go#L14-L19

ValarDragon:
Heres a screenshot for SetHasVote, sorry for not including that in OP: 
![image](https://github.com/cometbft/cometbft/assets/6440154/ce4046fb-20fb-485d-a404-b0bbbb14081e)

cason:
But for the first 2 examples, we can do what we already have for block hashes, with `LazyBlockHash`. 

In both cases it is a vote set. The logic is that we pass the vote set (reference), and only call the `StringShort()` if the logger produces the log message.

andynog:
I agree that there are too many allocations in this logic, especially on `consensus/state`. Unfortunately the current logging logic allocates heap even if debug mode is not enabled. 

[...]

PR #2989, opened 2024-05-02
perf: log debug only if it is enabled to avoid heap allocations

andynog:
close: #2847 

**NOTE**: Please **don't** merge this yet. This is an experimental WIP so @ValarDragon can do some testing for performance profiling. Depending on the outcome of these changes we then can decide to merge it or not. But it would still need a bit of work in the logging library.

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/internal/consensus/reactor.go
+++ b/internal/consensus/reactor.go
@@ -238,7 +238,9 @@ func (conR *Reactor) RemovePeer(p2p.Peer, any) {
 // NOTE: blocks on consensus state for proposals, block parts, and votes.
 func (conR *Reactor) Receive(e p2p.Envelope) {
 	if !conR.IsRunning() {
-		conR.Logger.Debug("Receive", "src", e.Src, "chId", e.ChannelID)
+		if conR.Logger.DebugOn() {
+			conR.Logger.Debug("Receive", "src", e.Src, "chId", e.ChannelID)
+		}
 		return
 	}
 	msg, err := MsgFromProto(e.Message)
@@ -254,8 +256,9 @@ func (conR *Reactor) Receive(e p2p.Envelope) {
 		return
 	}
 
-	conR.Logger.Debug("Receive", "src", e.Src, "chId", e.ChannelID, "msg", msg)
-
+	if conR.Logger.DebugOn() {
+		conR.Logger.Debug("Receive", "src", e.Src, "chId", e.ChannelID, "msg", msg)
+	}
 	// Get peer states
 	ps, ok := e.Src.Get(types.PeerStateKey).(*PeerState)
 	if !ok {
@@ -660,18 +663,22 @@ OUTER_LOOP:
 			if ps.sendVoteSetHasVote(vote) {
 				continue OUTER_LOOP
 			}
-			logger.Debug("Failed to send vote to peer",
-				"height", prs.Height,
-				"vote", vote,
-			)
+			if logger.DebugOn() {
+				logger.Debug("Failed to send vote to peer",
+					"height", prs.Height,
+					"vote", vote,
+				)
+			}
 		}
 
 		if sleeping == 0 {
 			// We sent nothing. Sleep...
 			sleeping = 1
-			logger.Debug("No votes to send, sleeping", "rs.Height", rs.Height, "prs.Height", prs.Height,
-				"localPV", rs.Votes.Prevotes(rs.Round).BitArray(), "peerPV", prs.Prevotes,
-				"localPC", rs.Votes.Precommits(rs.Round).BitArray(), "peerPC", prs.Precommits)
+			if logger.DebugOn() {
+				logger.Debug("No votes to send, sleeping", "rs.Height", rs.Height, "prs.Height", prs.Height,
+					"localPV", rs.Votes.Prevotes(rs.Round).BitArray(), "peerPV", prs.Prevotes,
+					"localPC", rs.Votes.Precommits(rs.Round).BitArray(), "peerPC", prs.Precommits)
+			}
 		} else if sleeping == 2 {
 			// Continued sleep...
 			sleeping = 1
@@ -876,7 +883,9 @@ func pickVoteToSend(
 	// If peer is lagging by height 1, send LastCommit.
 	if prs.Height != 0 && rs.Height == prs.Height+1 {
 		if vote := ps.PickVoteToSend(rs.LastCommit); vote != nil {
-			logger.Debug("Picked rs.LastCommit to send", "height", prs.Height)
+			if logger.DebugOn() {
+				logger.Debug("Picked rs.LastCommit to send", "height", prs.Height)
+			}
 			return vote
 		}
 	}
@@ -907,7 +916,9 @@ func pickVoteToSend(
 			return nil
 		}
 		if vote := ps.PickVoteToSend(ec); vote != nil {
-			logger.Debug("Picked Catchup commit to send", "height", prs.Height)
+			if logger.DebugOn() {
+				logger.Debug("Picked Catchup commit to send", "height", prs.Height)
+			}
 			return vote
 		}
 	}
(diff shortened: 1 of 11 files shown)


andynog:
Hi @ValarDragon, any updates on profiling these changes ?

Please let us know when you have some results so we can decide if we will continue with this PR or not (currently it's on draft mode until we get some feedback on these changes). Thanks again !

ValarDragon:
Yes! Been mildly delayed due to breaking an api in the sdk, and our tooling messed up on both sdk and comet patches at once. Going to try again today

andynog:
> Yes! Been mildly delayed due to this breaking an api in the sdk, and our tooling messed up on both sdk and comet patches at once. Going to try again today

Amazing, keep us posted ! 👍

ValarDragon:
I just did this on some known hot loop locations I saw due to merge conflicts, and saw a very significant improvement: 
https://github.com/osmosis-labs/cometbft/commit/30a125b6f4702c417d67422e37643597048b9e45

Savings were over 20GB of RAM. We see that `setHasVote`'s heap and time went away entirely, same for StringShort. I would need to reprofile to get something more apples to apples to understand improvements outside of these two functions. Would that be helpful? Anyways going to re-profile to see where performance with all merged improvements jointly together is.

internal/consensus/reactor.go
```diff
@@ -923,47 +934,59 @@ func pickVoteCurrentHeight(
 	// If there are lastCommits to send...
 	if prs.Step == cstypes.RoundStepNewHeight {
 		if vote := ps.PickVoteToSend(rs.LastCommit); vote != nil {
-			logger.Debug("Picked rs.LastCommit to send")
+			if logger.DebugOn() {
+				logger.Debug("Picked rs.LastCommit to send")
```
cason, line 938: So, this kind of debug log does not constitute a problem. The string is a constant, included in the binary (whether debug is enabled or not). And the cost of checking the level can be paid inside the `Debug()` call.

internal/consensus/reactor.go
```diff
@@ -923,47 +934,59 @@ func pickVoteCurrentHeight(
...
 	}
 	// If there are POL prevotes to send...
 	if prs.Step <= cstypes.RoundStepPropose && prs.Round != -1 && prs.Round <= rs.Round && prs.ProposalPOLRound != -1 {
 		if polPrevotes := rs.Votes.Prevotes(prs.ProposalPOLRound); polPrevotes != nil {
 			if vote := ps.PickVoteToSend(polPrevotes); vote != nil {
-				logger.Debug("Picked rs.Prevotes(prs.ProposalPOLRound) to send",
-					"round", prs.ProposalPOLRound)
+				if logger.DebugOn() {
+					logger.Debug("Picked rs.Prevotes(prs.ProposalPOLRound) to send",
+						"round", prs.ProposalPOLRound)
```
cason, line 949: In this case, the only non constant field passed to the logger is an integer. Not sure whether this code around is need also in this case.

internal/consensus/reactor.go
```diff
@@ -1165,13 +1190,14 @@ func (ps *PeerState) SetHasProposalBlockPart(height int64, round int32, index in
...
 func (ps *PeerState) setHasProposalBlockPart(height int64, round int32, index int) {
-	ps.logger.Debug("setHasProposalBlockPart",
-		"peerH/R",
-		log.NewLazySprintf("%d/%d", ps.PRS.Height, ps.PRS.Round),
-		"H/R",
-		log.NewLazySprintf("%d/%d", height, round),
-		"index", index)
-
+	if ps.logger.DebugOn() {
+		ps.logger.Debug("setHasProposalBlockPart",
```
cason, line 1194: Are you sure that the lazy print does not solve the problem here?

internal/consensus/state.go
```diff
@@ -1188,16 +1211,20 @@ func (cs *State) enterPropose(height int64, round int32) {
 
 	// if not a validator, we're done
 	if !cs.Validators.HasAddress(addr) {
-		logger.Debug("propose step; not proposing since node is not in the validator set",
-			"addr", addr,
-			"vals", cs.Validators)
+		if logger.DebugOn() {
+			logger.Debug("propose step; not proposing since node is not in the validator set",
+				"addr", addr,
+				"vals", cs.Validators)
```
cason, line 1217: This is a case that is worth considering, as `cs.Validators` is a complex struct with a string method etc. etc.

internal/consensus/state.go
```diff
@@ -1407,10 +1444,12 @@ func (cs *State) defaultDoPrevote(height int64, round int32) {
 		}
 
 		if cs.Proposal.POLRound == -1 {
-			logger.Debug("prevote step: Proposal is timely",
-				"timestamp", cs.Proposal.Timestamp.Format(time.RFC3339Nano),
-				"receive_time", cs.ProposalReceiveTime.Format(time.RFC3339Nano),
-				"timestamp_difference", cs.ProposalReceiveTime.Sub(cs.Proposal.Timestamp))
+			if logger.DebugOn() {
+				logger.Debug("prevote step: Proposal is timely",
+					"timestamp", cs.Proposal.Timestamp.Format(time.RFC3339Nano),
```
cason, line 1449: This is another case in which makes sense, as we expend time formatting timestamps.

internal/consensus/state.go
```diff
@@ -1560,10 +1613,12 @@ func (cs *State) enterPrevoteWait(height int64, round int32) {
...
 
 	if cs.Height != height || round < cs.Round || (cs.Round == round && cstypes.RoundStepPrevoteWait <= cs.Step) {
-		logger.Debug(
-			"entering prevote wait step with invalid args",
-			"current", log.NewLazySprintf("%v/%v/%v", cs.Height, cs.Round, cs.Step),
-		)
+		if logger.DebugOn() {
+			logger.Debug(
+				"entering prevote wait step with invalid args",
+				"current", log.NewLazySprintf("%v/%v/%v", cs.Height, cs.Round, cs.Step),
```
cason, line 1619: We don't need the `log.NewLazySprintf` in this case, right? The idea of this feature was to avoid paying the cost of printing when we are not printing because the logger level is not the desired one (debug).

cason (commented):
This idea is great, but we don't need to test the log level for every produced log message. I left some comments where I think there is no gain at all in using this approach (actually, we check the log level twice).

internal/consensus/reactor.go
ValarDragon: Yes, the lazy print is adding new objects to heap

andynog:
That's amazing @ValarDragon, if it save 20GB of RAM 👀, please let us know if you re-profile

I've added the code in all debugs. I've seen @cason comments that we might not need in all the places and I've seen in your commit the hot spot ones. If you confirm these are the ones causing issues I can revert some of the logic for `DebugOn` on places that wouldn't make much of a difference in performance.

Thanks again!

ValarDragon:
Yeah I think just doing it on the hot spot ones sounds great to me!

melekes:
> Yeah I think just doing it on the hot spot ones sounds great to me!

agree 👍

andynog:
Hi @ValarDragon, just following up here. Are these debug logs something you had a chance to validate in performance gains? Should we proceed with these changes? Please let us know. Thanks !

ValarDragon:
Ah sorry missed this ping (please feel free to ping on slack if I miss something in the future)

Definitely noticing the cost of logging in:
- `peerstate.SetHasVote` (which got worse when we lowered the gossip routines timeout!), 
- `mconnection.Send`
- `StringShort` in `AddVote`

andynog:
Thanks for the follow-up @ValarDragon. I'll try to constrain this PR only on those places that are hot spots that you mentioned above and then mark it ready for review.

ValarDragon:
I've removed several more that are bottlenecks in the p2p stack. Forgot to post here, will reply with more as I remember / re-see them. (Maybe when we do our SDK v50 upgrade, since I'll have to move all the commits vo v0.38.x line)

ch *Channel) recvPacketMsg is a big one to remove

andynog:
I'll close this PR. The implementation here is not the best way to solve the problem in #2847. I'm working on a new PR that has a better solution to the problem.

PR #2999, opened 2024-05-03
perf(p2p): lower `flush_throttle_timeout`

### LINKED ISSUE #2988: Lower default flush timeout
### Summary

This issue is based on conversations with @melekes , and experimentation in production done on both Osmosis and Injective.

After you "send a message to a peer", it doesn't actually send until either:
* You hit a 64kb chunk of total messages to send
* `flush_throttle_timeout` elapses

The default flush_throttle_timeout is `100ms`. Both osmosis and injective have found great success in performance lowering this to `10ms`, and have not tested lower values.

On the given profile I'm looking at, of an Osmosis synced-to-tip full node at `10ms`, we see that 70% of outbound packets are stuck waiting for `flush_throttle_timeout`! 

image

We argue this by looking at the ratio of the top call from SendRoutine (happens on timer timeout), vs call from .Write() which only writes to file if you hit buffer size. This is a time ratio of 56s in top-level flush, to 23s in the .Write() call. IMO this suggests it is worth lowering this buffer timeout much further. (To be honest, I'm unclear if we have any evidence this flush timeout is even helpful)

One nuance that could be happening, but I think is unlikely, is that during consensus time we always hit the buffer size, and only do flush timeouts during idle periods. I think its unlikely for consensus time, because we saw that lowering this flush timeout had direct impacts on validators staying synced/ getting votes propogated to the network. (E.g. when there is no blockparts or many mempool txs filing the buffer)

### Problem Definition

We wait far too long on networks, vs flushing faster.

### Proposal

Lower default flush timeout from 100ms to 5ms

melekes:
Closes #2988

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/.changelog/unreleased/improvements/2988-flush-throttle-timeout.md
+++ b/.changelog/unreleased/improvements/2988-flush-throttle-timeout.md
@@ -1,0 +1,2 @@
+- [`p2p`] Lower `flush_throttle_timeout` to 10ms
+  ([\#2988](https://github.com/cometbft/cometbft/issues/2988)
--- a/config/config.go
+++ b/config/config.go
@@ -762,7 +762,7 @@ func DefaultP2PConfig() *P2PConfig {
 		MaxNumInboundPeers:           40,
 		MaxNumOutboundPeers:          10,
 		PersistentPeersMaxDialPeriod: 0 * time.Second,
-		FlushThrottleTimeout:         100 * time.Millisecond,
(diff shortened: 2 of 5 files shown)


andynog (approved):
lgtm, the default values and spec/docs have been updated properly.

sergio-mena (approved):
LGTM, apart from the comment I added above

ValarDragon (approved):

PR #3003, opened 2024-05-03
perf(blockstore): Add LRU caches to blockstore operations used in consensus

### LINKED ISSUE #2844: Add lru cache for loading commit from blockstore
### Summary

Whenever a peer is lagging by more than 1 block, we load the commit for the height they were on and send them the whole commit (per code comment. Admittedly I haven't looked long, but I don't see how it's currently sending the whole commit vs just a single vote from it). [Source](https://github.com/cometbft/cometbft/blob/d27732ad9a992e321e440417a2a8220f90fe0044/internal/consensus/reactor.go#L885)

Loading the commit requires doing a db read and a proto unmarshal. [Source](https://github.com/cometbft/cometbft/blob/main/store/store.go#L331) In benchmarks we see the proto unmarshal taking a large number of heap allocations, and this is likely duplicate work. (Unmarshalling the same block commit)

We should keep a small LRU cache for height to unmarshalled commit to reduce the memory pressure here. I suspect the same maybe helpful for block parts as well. 

In v37 we saw that this was around 70GB of heap allocation on a query serving node over 1-2 hours. (Will paste memprofile later)

I do not know the total system impact it has, but a 70GB ram allocation is very high. 

### Proposal

Add a simple lru cache in blockstore for loading commit and block parts

ValarDragon:
Closes #2844

We are seeing that the blockstore loading operations get used in hot loops within gossip routines, and queryMaj23 routines. This PR reduces that overhead using an LRU cache.

The LRU cache does have a mutex on every get, but the time with the LRU cache is 9x faster than without (before even adding in DB overheads), due to the proto unmarshalling saved. We could imagine a setup where we avoided a lock there entirely. I don't think this is worth right now, since the new code is 9x faster, and these mostly appear in catchup code which should not be highly contended for across peers at the same time.

With the new benchmark I added:
OLD:
```
BenchmarkRepeatedLoadSeenCommit-12         24447             54691 ns/op           46495 B/op        319 allocs/op
```
NEW:
```
BenchmarkRepeatedLoadSeenCommit-12        224131              6401 ns/op            8320 B/op          2 allocs/op
```

It turns out these gossip routines don't need mutative copies, so we could optimize out the large allocation in the future if we want.

1 hour cpu profile that shows this appearing in prod: 
image

The state machine execution time here for context is 92 seconds. So this is adding up in system load (and GC! The GC load is 52GB, the entire trace is 200GB, with other parts being optimized down from recent PRs)

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/store/bench_test.go
+++ b/store/bench_test.go
@@ -13,7 +13,7 @@ import (
 // TestLoadBlockExtendedCommit tests loading the extended commit for a previously
 // saved block. The load method should return nil when only a commit was saved and
 // return the extended commit otherwise.
-func BenchmarkRepeatedLoadSeenCommit(b *testing.B) {
+func BenchmarkRepeatedLoadSeenCommitSameBlock(b *testing.B) {
 	state, bs, _, _, cleanup, _ := makeStateAndBlockStoreAndIndexers()
 	defer cleanup()
 	h := bs.Height() + 1
--- a/store/bench_test.go
+++ b/store/bench_test.go
@@ -1,0 +1,35 @@
+package store
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/cometbft/cometbft/internal/test"
+	"github.com/cometbft/cometbft/types"
+	cmttime "github.com/cometbft/cometbft/types/time"
+)
+
+// TestLoadBlockExtendedCommit tests loading the extended commit for a previously
+// saved block. The load method should return nil when only a commit was saved and
+// return the extended commit otherwise.
+func BenchmarkRepeatedLoadSeenCommit(b *testing.B) {
+	state, bs, _, _, cleanup, _ := makeStateAndBlockStoreAndIndexers()
+	defer cleanup()
+	h := bs.Height() + 1
+	block := state.MakeBlock(h, test.MakeNTxs(h, 10), new(types.Commit), nil, state.Validators.GetProposer().Address)
+	seenCommit := makeTestExtCommitWithNumSigs(block.Header.Height, cmttime.Now(), 100).ToCommit()
+	ps, err := block.MakePartSet(types.BlockPartSizeBytes)
+	require.NoError(b, err)
+	bs.SaveBlock(block, ps, seenCommit)
+
+	// sanity check
+	res := bs.LoadSeenCommit(block.Height)
+	require.Equal(b, seenCommit, res)
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		res := bs.LoadSeenCommit(block.Height)
+		require.NotNil(b, res)
+	}
+}
--- a/store/store.go
+++ b/store/store.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/cosmos/gogoproto/proto"
 	"github.com/go-kit/kit/metrics"
+	lru "github.com/hashicorp/golang-lru/v2"
 
 	dbm "github.com/cometbft/cometbft-db"
 	cmtstore "github.com/cometbft/cometbft/api/cometbft/store/v1"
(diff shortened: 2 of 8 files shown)


ValarDragon:
Seems like the PR is failing due to the use of the lru import from hashicorp. Open question I guess if thats safe as an import? As far as I understand its pretty standard, and doesn't have licensing issues: https://pkg.go.dev/github.com/hashicorp/golang-lru/v2 . Its used in many of the big blockchain repos as well: https://pkg.go.dev/github.com/hashicorp/golang-lru/v2?tab=importedby (Filecoin, libp2p, Polygon Hermez' bridge server )

melekes (approved):
Thanks @ValarDragon ❤️

melekes:
> Seems like the PR is failing due to the use of the lru import from hashicorp. Open question I guess if thats safe as an import? As far as I understand its pretty standard, and doesn't have licensing issues: https://pkg.go.dev/github.com/hashicorp/golang-lru/v2 . Its used in many of the big blockchain repos as well: https://pkg.go.dev/github.com/hashicorp/golang-lru/v2?tab=importedby (Filecoin, libp2p, Polygon Hermez' bridge server )

You need to add it to https://github.com/cometbft/cometbft/blob/8f8e4db1922d1558185440bddd8caa29ba3e2ab1/.golangci.yml#L92

melekes (approved):

store/store.go
```diff
@@ -329,6 +352,10 @@ func (bs *BlockStore) LoadBlockMetaByHash(hash []byte) *types.BlockMeta {
 // and it comes from the block.LastCommit for `height+1`.
 // If no commit is found for the given height, it returns nil.
 func (bs *BlockStore) LoadBlockCommit(height int64) *types.Commit {
+	comm, ok := bs.blockCommitCache.Get(height)
+	if ok {
+		return comm.Clone()
```
cason, line 357: Do we need to return a clone here? I assume this data never changes...

store/bench_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/cometbft/cometbft/internal/test"
+	"github.com/cometbft/cometbft/types"
+	cmttime "github.com/cometbft/cometbft/types/time"
+)
+
+// TestLoadBlockExtendedCommit tests loading the extended commit for a previously
+// saved block. The load method should return nil when only a commit was saved and
+// return the extended commit otherwise.
+func BenchmarkRepeatedLoadSeenCommit(b *testing.B) {
```
cason, line 16:
```suggestion
func BenchmarkRepeatedLoadSeenCommitSameBlock(b *testing.B) {
```

cason: To make it clearer.

store/store.go
```diff
@@ -139,6 +145,23 @@ func NewBlockStore(db dbm.DB, options ...BlockStoreOption) *BlockStore {
 	return bStore
 }
 
+func (bs *BlockStore) addCaches() {
+	var err error
+	// err can only occur if the argument is non-positive, so is impossible in context.
+	bs.blockCommitCache, err = lru.New[int64, *types.Commit](100)
```
cason, line 151: I know that it is somehow arbitrary, but why 100? :D

cason (approved):
Thank you for this.

Should we add a changelog entry?

store/store.go
ValarDragon: Returning the Clone replicates existing behavior. (A caller mutation would pollute the cache)

We can change the API to remove the clone, which I think is a good idea, we would just have to look for all usages and ensure they don't mutate it

store/store.go
ValarDragon: It was very much arbitrary, it has to be bigger than like 3 for live consensus + peer catchup. Was hoping that if its a bigger value, we wouldn't get too much thrashing while also serving blocksync

store/store.go
melekes: @ValarDragon would you prefer changing the API here or in another PR?

ValarDragon: Another PR

PR #3006, opened 2024-05-05
Make every gossip thread use its own randomness instance, reducing contention

### LINKED ISSUE #3005: Delete usage of comet's libs/rand, make each gossip routine maintain its own rand instance.
### Summary

CometBFT's libs/rand was always a really questionable choice that never made sense to me. (Documented from before the cosmos hub launch: https://github.com/tendermint/tendermint/issues/2956 )

We should just delete all usage of our libs/rand, and make each gossiping thread maintain its own randomness instance. This will reduce mutex contention, and speedup core gossip routines.

Each rand instance has only 50% doing its actual rand operation. And we see `lockSlow` getting called, which per my understanding of the go runtime, means that this is actually a contended for mutex, hence this is justified.
image

ValarDragon:
Closes #3005 

As noted in that issue, we currently are doing extra CPU overhead and mutex contention for getting a random number. This PR removes this overhead by making every performance sensitive thread have its own rand instance.

In a subsequent PR, we can cleanup all the testing usages, and likely just entirely delete our internal randomness package. I didn't do that in this PR to keep it straightforward to verify.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/internal/consensus/reactor.go
+++ b/internal/consensus/reactor.go
@@ -2,6 +2,7 @@ package consensus
 
 import (
 	"fmt"
+	"math/rand"
 	"reflect"
 	"sync"
 	"sync/atomic"
@@ -569,6 +570,7 @@ func (conR *Reactor) getRoundState() *cstypes.RoundState {
 
 func (conR *Reactor) gossipDataRoutine(peer p2p.Peer, ps *PeerState) {
 	logger := conR.Logger.With("peer", peer)
+	r := cmtrand.NewStdlibRand()
 
 OUTER_LOOP:
 	for {
@@ -581,7 +583,7 @@ OUTER_LOOP:
 		// so we can reduce the amount of redundant block parts we send
 		if conR.conS.config.PeerGossipIntraloopSleepDuration > 0 {
 			// the config sets an upper bound for how long we sleep.
-			randDuration := cmtrand.Int63n(int64(conR.conS.config.PeerGossipIntraloopSleepDuration))
+			randDuration := r.Int63n(int64(conR.conS.config.PeerGossipIntraloopSleepDuration))
 			time.Sleep(time.Duration(randDuration))
 		}
 
@@ -593,7 +595,7 @@ OUTER_LOOP:
 		// (Note these can match on hash so round doesn't matter)
 		// --------------------
 
-		if part, continueLoop := pickPartToSend(logger, conR.conS.blockStore, rs, ps, prs); part != nil {
+		if part, continueLoop := pickPartToSend(logger, conR.conS.blockStore, rs, ps, prs, r); part != nil {
 			// part is not nil: we either succeed in sending it,
 			// or we were instructed not to sleep (busy-waiting)
 			if ps.SendPartSetHasPart(part, prs) || continueLoop {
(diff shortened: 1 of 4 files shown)


internal/rand/random.go
```diff
@@ -302,6 +314,8 @@ func (r *Rand) Perm(n int) []int {
 // NOTE: This relies on the os's random number generator.
 // For real security, we should salt that with some seed.
 // See github.com/cometbft/cometbft/crypto for a more secure reader.
```
ValarDragon, line 316: I disagree with this comment, but will leave to the general cleanup PR for this package to remove.

internal/consensus/reactor.go
```diff
@@ -569,6 +570,7 @@ func (conR *Reactor) getRoundState() *cstypes.RoundState {
 
 func (conR *Reactor) gossipDataRoutine(peer p2p.Peer, ps *PeerState) {
 	logger := conR.Logger.With("peer", peer)
+	r := cmtrand.NewStdlibRand()
```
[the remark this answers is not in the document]
ValarDragon: Makes sense, will do!

internal/rand/random.go
```diff
@@ -37,14 +37,28 @@ func NewRand() *Rand {
...
 }
 
-func (r *Rand) init() {
+// Make a new stdlib rand source. Its up to the caller to ensure
+// that the rand source is not called in parallel.
+// The failure mode of calling the returned rand multiple times in parallel is
+// repeated values across threads.
+func NewStdlibRand() *mrand.Rand {
+	// G404: Use of weak random number generator (math/rand instead of crypto/rand)
+	//nolint:gosec
```
ValarDragon, line 46: Note that the way we create the local Rand object also has this nolint, so its not newly introduced.

internal/rand/random_test.go
```diff
@@ -87,6 +87,24 @@ func TestRngConcurrencySafety(_ *testing.T) {
 	wg.Wait()
 }
 
+// Makes a new stdlib random instance 100 times concurrently.
```
cason, line 90: Should we test that `cRandBytes` is thread safe instead?

internal/rand/random.go
cason: This comment does not make sense given the preamble of this source file...

internal/consensus/reactor.go
cason, line 1: Somehow impressed that the bit array method is only used here. I wonder whether implementing part of this logic here, instead of there, could be a better solution.

At the end, we are counting the 1 bits, then getting a random index from zero to this number, then retrieving the i-th 1 bit. The random part we could implement outside the bit array library.

cason (approved):
Using a synchronized random source here does not make sense.

The solution, however, is pretty complex, but it is what we can do right now. I would also check other internal uses of this internal random library to understand when we actually need a thread-safe version of it.

ValarDragon:
Every other usage from what I can tell can just be replaced with math.Rand, but wanted to do that in a separate library. (Math.rand's global instance is thread safe, by internally doing CAS operations and a retry loop with no wait time)

cason:
The problem of `math/rand` is the setup/seed used. Paying the synchronization costs does not make much sense in most of our use cases.

Issue #2925, opened 2024-04-29
Mempool Rechecking all txs blocks consensus


ValarDragon:
### Summary

Rechecking all txs in the mempool seems to block consensus. We see this from:
- How pprof's look
- How the code appears

This is problematic as it means larger mempools will delay consensus longer. (Also IBC has a change that creeped in that is causing overly large expenses in RecheckTx)

### Problem Definition

We should make the mempool rechecking not block `BlockExecutor.ApplyBlock`. 

Ideally it should only be blocking `ProposeBlock` until either everything in the mempool is rechecked or `blockGas`  worth of txs are rechecked. It should never be blocking for `timeout_prevote`

andynog:
There are some refactoring on the re-checking logic happened in this PR #2268, maybe this will not solve the problem from the perspective @ValarDragon reported but just adding here for visibility. @hvanz might have a better opinion in this case.

The mempool recheck can also be controlled with a configuration parameter
https://docs.cometbft.com/v1.0/references/config/config.toml#mempoolrecheck

hvanz:
Thanks @ValarDragon for reporting this problem! I agree that rechecking should not block consensus, though it will still need to block the mempool for checking new incoming transactions. This is for not breaking the FIFO ordering when rechecking txs.

Currently `BlockExecutor.Commit` calls `mempool.Lock`, `mempool.FlushAppConn`, `mempool.Update` and `mempool.Unlock`. It is the only function that calls these mempool methods. We have it in our list to refactor these calls to make the mempool code cleaner and more modular. We'll keep into account the problem you report when addressing the refactoring.

hvanz:
Also, #2268 is about a corner case in the rechecking logic, not related to current issue.

ValarDragon:
I think what we should do here is:

- Maintain statistics via Atomic instructions in the mempool for "how many bytes", "how many gas" is rechecked, and "are we done rechecking"
- Always run Update asynchronously. (Mempool updating should not block any consensus timeouts)
  - Update returns once the trackers are rechecked
- Make the Reap function poll/block until oneof:
  - "We are done rechecking" = true
  - "how many bytes rechecked" < Reap Bytes limit
  - "How much gas is rechecked" < Reap gas limit

cason:
Should we add this to https://github.com/cometbft/cometbft/issues/2803?

cason:
> If you look into the relevant code, each recheck call is actually synchronous due to how the callback's are structured.

[...]

PR #3008, opened 2024-05-06
perf: Make mempool update async from block.Commit

ValarDragon:
First step to fixing #2925 

PR'ing this to see if we have any test failures. Note that this is safe in the happy path, as Reap and CheckTx both share this same lock. The functionality behavior is that:

- Full nodes and non-proposers `timeout_prevote` beginning should not block on  updating the mempool
- Block proposers get _very slight_ increased concurrency before reaping their next block. (Should be significantly fixed in subsequent PR's in #2925) 
  - Reap takes a lock on the mempool mutex, so there is no concurrency safety issues right now.
- Mempool errors will not halt consensus, instead they just log an error and call mempool flush. I actually think this may be better behavior? If we want to preserve the old behavior, we can thread a generic "consensus halt error" channel perhaps?

I'm not sure how/where to best document this. Please also let me know if tests need creating. Seems like the create empty block tests sometimes hit failures, I'll investigate tmrw

Also please feel free to take over this PR, just thought I"d make it to help us with performance improvements. Happy to get this into an experimental release to test on mainnets.

---

#### PR checklist

- [ ] Tests written/updated


state/execution.go
```diff
@@ -417,15 +416,44 @@ func (blockExec *BlockExecutor) Commit(
...
+		updateData.block.Txs,
+		updateData.abciResponse.TxResults,
+		TxPreCheck(updateData.state),
+		TxPostCheck(updateData.state),
 	)
-
-	return res.RetainHeight, err
+	if err != nil {
+		blockExec.logger.Error("client error during mempool.Update, flushing mempool", "err", err)
+		blockExec.mempool.Flush()
```
melekes, line 454: This is new, right? Why it's important to flush the mempool?

ValarDragon: Before we would halt, sorry forgot to flag in a comment. I feel like we only have two options,  dump the mempool or replicate existing functionality and make an error channel to halt consensus

melekes (approved):

cason:
> Full nodes and non-proposers timeout_prevote beginning should not block on updating the mempool

What do you mean? In any case, it is probably `timeout_commit`.

cason:
One problem of this approach is that this breaks the contract that we have regarding the ABCI `Commit` method:

https://docs.cometbft.com/v0.38/spec/abci/abci++_app_requirements#commit

> Before invoking Commit, CometBFT locks the mempool and flushes the mempool connection. This ensures that no new messages will be received on the mempool connection during this processing step, providing an opportunity to safely update all four connection states to the latest committed state at the same time.
>
> When Commit returns, CometBFT unlocks the mempool.

By updating the mempool in parallel, this contract is clearly broken.

state/execution.go
```diff
@@ -417,15 +416,44 @@ func (blockExec *BlockExecutor) Commit(
...
+
+// updates the mempool with the latest state asynchronously.
+// it begins flushing the mempool immediately, and then updates the mempool
+// after the new block state is ready.
+func (blockExec *BlockExecutor) asyncUpdateMempool(ch chan mempoolUpdate) {
+	blockExec.mempool.Lock()
+	defer blockExec.mempool.Unlock()
+
+	// while mempool is Locked, flush to ensure all async requests have completed
+	// in the ABCI app before Commit.
```
cason, line 432: We are supposed to do that _before_ calling ABCI Commit.

state/execution.go
```diff
@@ -417,15 +416,44 @@ func (blockExec *BlockExecutor) Commit(
...
+		blockExec.mempool.Flush()
+		return
+	}
+
+	updateData := <-ch
+	if updateData.err != nil {
+		return
+	}
+
 	err = blockExec.mempool.Update(
```
cason, line 445: This part we can do in parallel.

Once the new state is committed, the application can start performing the re-check of all transactions remaining in the mempool, using the updated state.

cason: Moreover, it is not clear to me why we need to call `Update` with a locked mempool...

cason (changes requested):
I would prefer to block this change while we address how we are handling the contract with the application as part of the ABCI Commit call.

state/execution.go
ValarDragon: As far as I understand, this is due to the current FIFO guarantee, before you add a new tx. (You want to know your done rechecking, before you add more txs to the end)

This really should be a lock owned within the mempool though, rather than here.

state/execution.go
ValarDragon: Your right, thank you!

ValarDragon:
> One problem of this approach is that this breaks the contract that we have regarding the ABCI `Commit` method:
> 
> https://docs.cometbft.com/v0.38/spec/abci/abci++_app_requirements#commit
> 
> > Before invoking Commit, CometBFT locks the mempool and flushes the mempool connection. This ensures that no new messages will be received on the mempool connection during this processing step, providing an opportunity to safely update all four connection states to the latest committed state at the same time.
> > When Commit returns, CometBFT unlocks the mempool.
> 
> By updating the mempool in parallel, this contract is clearly broken.

After your fix of flushing the mempool before Commit, we are only breaking the contract in "When Commit returns, CometBFT unlocks the mempool.". We do flush and take a lock during Commit. We just now hold the lock after Commit terminates. I'm not sure that application behavior would depend on this.

ValarDragon:
I'm still stuck on the create empty blocks tests. Going to try more tomorrow. But somehow playing around with the timeouts / expanding leniency isn't fixing this. (Testing via running `go test -timeout 30s -run ^TestMempoolProgressAfterCreateEmptyBlocksInterval$ github.com/cometbft/cometbft/internal/consensus -count=100 -p=1`) 

Apriori I'd just guess maybe we shaved 10ms or something in this configuration in edge cases. Going to try more with a debugger tmrw.

hvanz (approved):
Looks good. I've also run the tests and they all pass.

Regarding the ABCI spec on Commit, I don't think this change will pose a problem. The important part of that paragraph is "Before invoking Commit, CometBFT locks the mempool and flushes the mempool connection.". This is not broken.

But the line "When Commit returns, CometBFT unlocks the mempool." should be updated to reflect the new behaviour. I think it doesn't matter for the ABCI connections that the mempool is still locked after Commit returns.

ValarDragon:
> Looks good. I've also run the tests and they all pass.

Oh interesting, I am getting failures, I've debugged the cause though! We sometimes get RoundStepNewHeight scheduled first, with a negative time. That timeout then gets processed first before the create empty blocks one. I think this failure is independent of this PR, and is just a bug thats present (and would cause failures sometimes in main).

Maybe the chance of hitting this bug is just raised with this PR somehow?
```
1715931925716 round step new height
1715931925716 scheduled timeout create empty blocks 3 0
Scheduled timeout dur -595.496µs height 3 round 0 step RoundStepNewHeight
Scheduled timeout dur 200ms height 3 round 0 step RoundStepNewRound

```

ValarDragon:
I think the failing test cases is due to some bug in the timeout ticker. I added some logs, and most of the times I see logs like:
```
1715936989720 Scheduled timeout dur 200ms height 2 round 0 step RoundStepNewRound
1715936989921 Timed out dur 200ms height 2 round 0 step RoundStepNewRound
```
where the left number is `time.Now().UnixMilli()`

In the failing test cases I see:
```
1715936989928 Scheduled timeout dur 200ms height 2 round 0 step RoundStepNewRound
1715936989928 Timed out dur 200ms height 2 round 0 step RoundStepNewRound
```

(So were not getting the 200ms wait)
Trying to figure out whats the cause of this edge case.

ValarDragon:
Ok just confirmed I get this same failure on main when I run the test in a loop, so its not new to this PR (and is an edge case failure in the timeout ticker)

sergio-mena:
> Or perhaps we merge that in first, then rebase into this PR?

I favour that approach

state/execution.go
```diff
@@ -417,15 +425,28 @@ func (blockExec *BlockExecutor) Commit(
...
 	// Update mempool.
-	err = blockExec.mempool.Update(
-		block.Height,
-		block.Txs,
-		abciResponse.TxResults,
-		TxPreCheck(state),
-		TxPostCheck(state),
-	)
+	memUpdate := mempoolUpdate{block: block, state: state.Copy(), abciResponse: abciResponse}
+	go blockExec.asyncUpdateMempool(unlockMempool, memUpdate)
```
sergio-mena, line 429: Wouldn't it be simpler to just pass all the arguments to `asyncUpdateMempool` directly (and so, we wouldn't need to define `mempoolUpdate` struct`.

Bear in mind that the actual parameters passed to `asyncUpdateMempool` are evaluated in the [caller's goroutine](https://go.dev/tour/concurrency/1).

state/execution.go
sergio-mena: Any caller of function `Commit` in line 395 above will panic if `Commit` doesn't return a `nil` error. This is for safety. I'd feel more comfortable if we didn't modify that behavior here, so `err != nil` in line 445, I'd panic here with the error text of line 446.

sergio-mena (approved):
Thanks for this. Seems like a simple change but it's actually an advanced one!

state/execution.go
ValarDragon: Sounds good!

state/execution.go
ValarDragon: Great point, done!

ValarDragon:
Great points! I think I addressed everything in your comments :)

ValarDragon:
Yay all tests passing now that the timeout ticker bugfix got merged in!

sergio-mena:
I'm removing the "backport to `v1.x`" label for now, as that branch is about to reach RC status, and is past out QA process. @ValarDragon if you include this in production and we see it's not causing unexpected side-effects, let us know and we can backport it at that moment.

ValarDragon:
Sounds great, planning to get it into a patch release soon

sergio-mena:
An update regarding [this](https://github.com/cometbft/cometbft/pull/3008#issuecomment-2124892339).

We have finally decided to backport this to `v1.x` because:

* Our `e2e` nightlies are failing, and that is blocking our process for cutting `v1.x` RC. We are convinced this is part of the fix for those failing tests. So I added the "bug" label
* @ValarDragon reported that this fix has been in production on Osmosis for a few weeks now without issue.

So, benefits clearly outweigh the risks at this point.

PR #3016, opened 2024-05-06
perf: Speedup flowrate by removing time.Now() from every update call

### LINKED ISSUE #2958: Change flowrate.clock() to be based on a variable that is scheduled to update every 20ms
### Summary

Currently the flowrate package has every single call to it do a time.Now() call, and then round it to the nearest 20ms, and uses that as a "flowrate epoch" internally for its computations.

This leads to lots of time spent in time.Now() and rounding logic. Result from a 1 hour live osmosis node profile (~1 month ago), that overlapped with the epoch (60s long):
image

10 seconds spent in time.Now() checking, when were trying to just send the packet, feels very significant. Unfortunately I don't have statistics for the number of packets we sent in this time period to be able to bound it though. Assuming the current flowrate functionality is what we want, we actually have a much more efficient way to implement this function. We can have a global "currentClock" var. Every clock() function is an atomic read of this var. We spin up another goroutine at initialization that atomically writes this var every 20ms. This should change this clock time to then be a tiny fraction of what it is now, with very little consensus delay.

Right now the code is currently
```
// czero is the process start time rounded down to the nearest clockRate
// increment.
var czero = time.Now().Round(clockRate)

// clock returns a low resolution timestamp relative to the process start time.
func clock() time.Duration {
	return time.Now().Round(clockRate).Sub(czero)
}
```
so this would change to roughly
```
var curClockTime time.Duration = time.Duration(0)

func clockTimeUpdateLoop() {
   for {
          time.Sleep(clockRate)
          newClockTime = oldclockfn()
          Atomic.SetInt64(newClockTime, &curClockTime)
    }
}

// clock returns a low resolution timestamp relative to the process start time.
func clock() time.Duration {
	return time.Duration(Atomic.LoadInt64(curClockTime))
}
```

I don't think we care about high fidelity time keeping of the flow rate, but if we did, then we could do the epoching goroutine do a time.Sleep for say 19.9ms, and check for the final boundary point with a .1ms sleep loop.

### Problem Definition

Lowers CPU contention in trying to send packets.

ValarDragon:
Closes #2958 
Speeds up flowrate by removing time.Now() calls from every `clock` call, instead making one goroutine responsible for all clock reads. On my most recent mainnet benchmark, this should shave 14% off of the synchronous time in sending packets: 

image

There may be slight differences at the boundary of these 20ms windows, but I do not think that is important.

---

#### PR checklist

- [x] Tests written/updated - existing tests pass


internal/flowrate/util.go
```diff
@@ -7,24 +7,50 @@ package flowrate
...
+	clockStartTime = time.Now().Round(clockRate)
+	go runClockUpdates()
+}
+
+// increments the current clock value every clockRate interval.
+func runClockUpdates() {
+	// sleep, then increment the clock value
+	// This is the only place the clock value is updated.
+	// Ensures that the clock value starts at 0.
+	for {
```
melekes, line 38: not sure how I feel about us not ever exiting here. I guess it's fine.

melekes (approved):
Thanks @ValarDragon ❤️

internal/flowrate/util.go
ValarDragon: I can use some more CAS instructions to make it exit

cason:
Not sure if I agree with this approach.

Using an atomic int field to keep time at higher granularity is good. Using sleeps to increase it is not great, we can't trust that sleep sleeps what it is supposed to.

An alternative is to use something like `time.Tick`, which also returns a time, and adjust the increment of the atomic counter, in case the tick is longer than expected. In other words, if we expect a sleep duration of 10ms, but it for some reason (and it will do that), sleeps for 30ms, we increase the atomic counter by 3.

internal/flowrate/util.go
ValarDragon: EDIT didn't work

melekes:
> Using an atomic int field to keep time at higher granularity is good. Using sleeps to increase it is not great, we can't trust that sleep sleeps what it is supposed to.

Good point 👍

ValarDragon:
Great point! Fixed in latest commit

internal/flowrate/util.go
```diff
@@ -7,24 +7,50 @@ package flowrate
...
+// increments the current clock value every clockRate interval.
+func runClockUpdates() {
+	// Create a ticker that sends the current time on the channel every clockRate interval
+	ticker := time.Tick(clockRate)
+
+	// First tick happens after clockrate, therefore initial value is 0.
+	for range ticker {
+		curValue := time.Duration(currentClockValue.Load())
+		nextValue := curValue + clockRate
+		currentClockValue.Store(int64(nextValue))
```
melekes, line 42:
```suggestion
		currentClockValue.Add(int64(clockRate))
```

internal/flowrate/util.go
```diff
@@ -7,24 +7,50 @@ package flowrate
...
+}
+
+// increments the current clock value every clockRate interval.
+func runClockUpdates() {
+	// Create a ticker that sends the current time on the channel every clockRate interval
+	ticker := time.Tick(clockRate)
+
+	// First tick happens after clockrate, therefore initial value is 0.
+	for range ticker {
+		curValue := time.Duration(currentClockValue.Load())
```
melekes, line 40:
```suggestion
```

internal/flowrate/util.go
```diff
@@ -7,24 +7,50 @@ package flowrate
...
+		nextValue := curValue + clockRate
```
melekes, line 41:
```suggestion
```

internal/flowrate/util.go
```diff
@@ -7,24 +7,50 @@ package flowrate
...
+	go runClockUpdates()
+}
+
+// increments the current clock value every clockRate interval.
+func runClockUpdates() {
+	// Create a ticker that sends the current time on the channel every clockRate interval
+	ticker := time.Tick(clockRate)
+
+	// First tick happens after clockrate, therefore initial value is 0.
+	for range ticker {
```
melekes, line 39: @cason @ValarDragon should we use time from the ticker? Because ticks might be dropped or delayed...

melekes: like 

```go
for t := <-ticker {
  currentClockValue.Store(t.Round(clockRate))
}
```

ValarDragon: Your right, that is better! Updating, didn't realize the scheduler was unreliable at 20ms gaps

melekes (approved):
👍

ValarDragon:
I think this PR is good to go!

sergio-mena (approved):
LGTM

adizere:
Should we backport anything from here into .37 or .38 ?

ValarDragon:
Yes please, lets backport to 38!

melekes:
@mergify backport v1.x

Issue #3002, opened 2024-05-03
txs_results not returned properly on dYdX nodes


penso:
### Setup

dYdX node, currently using `v0.38.5` ([code](https://github.com/dydxprotocol/v4-chain/blob/419549843fc9e093638a75008939b7c79c33f3e2/protocol/go.mod#L9))

### What happened?

Fetching older `block_results` doesn't return `txs_results` as it should. [This block/block_results](http://blocks.constellations.zone/dydx_mainnet/0/0/0/10000.json) was saved in November and properly has 4 txs and 4 txs_results (this JSON combines both block and block_results)

That same block_results fetched today at [https://dydx-ops-archive-rpc.kingnodes.com/block_results?height=10000](https://dydx-ops-archive-rpc.kingnodes.com/block_results?height=10000) only has 1 `txs_results`, and missing 3 `txs_results`. I feel like `txs_results` created with `v0.37.2` aren't returned properly when upgraded to `v0.38`. I also tried a few archive nodes and they all behave the same way.

I'm not affiliated with dYdX, I just happen to try running my cosmos indexer (Constellations) on dYdX and noticed it failed because the amount of `txs_results` is expected to be the same as `txs`.

penso:
Someone on a channel said there is the same issue on sei devnet

andynog:
we are currently investigating the problem and we will add more context here

andynog:
I could reproduce the problem only on CometBFT side. There is a way to reproduce it using only the kvstore example app.

Here are the steps:

First you need to run a cometbft `v0.37`. On a terminal in a cometbft repo folder, run:

```
git checkout v0.37.x
make install
cometbft init --home /tmp/cometbft-kvstore
cometbft start --proxy_app=kvstore --home=/tmp/cometbft-kvstore
```

Then submit a tx:
```
curl --location 'localhost:26657/broadcast_tx_commit?tx=%22tendermint%3Drocks%22'
```

Which will output a JSON
```
[...]

PR #3031, opened 2024-05-07
fix: invalid `txs_results` returned for legacy ABCI responses

andynog:
close: #3002 

This PR fixes the issue reported above.

This is not a storage issue in particular, the results are still in storage after an upgrade, but not returned properly by the RPC endpoint. The fix is to make the `/block_results` endpoint in `v0.38` to return properly a legacy ABCI response created with `v0.37`. 

Once this fix is merged on `v0.38` and a patch release is cut, any node on `v0.38` (e.g. an archive node) that applies the patch release, should have the results returned properly by the RPC `/block_results` endpoint. 

---

#### PR checklist

- [X] Tests written/updated


melekes (commented):
👍

state/errors.go
```diff
@@ -171,3 +171,7 @@ func (e ErrCannotLoadState) Error() string {
 func (e ErrCannotLoadState) Unwrap() error {
 	return e.Err
 }
+
+var ErrFinalizeBlockResponseUnmarshalError = errors.New("problem retrieving the finalize block response")
+
+var ErrFinalizeBlockResponseCorruptedError = errors.New("cannot retrieve finalize block response, corrupted data or spec change")
```
sergio-mena, line 177: Can we add these in line 164-165?

state/store.go
```diff
@@ -651,14 +651,22 @@ func (store dbStore) LoadFinalizeBlockResponse(height int64) (*abci.FinalizeBloc
...
+	if resp.AppHash == nil {
 		// The data might be of the legacy ABCI response type, so
 		// we try to unmarshal that
 		legacyResp := new(cmtstate.LegacyABCIResponses)
-		rerr := legacyResp.Unmarshal(buf)
-		if rerr != nil {
-			cmtos.Exit(fmt.Sprintf(`LoadFinalizeBlockResponse: Data has been corrupted or its spec has
-					changed: %v\n`, err))
+		rErr := legacyResp.Unmarshal(buf)
+		if rErr != nil {
```
sergio-mena, line 666:
```suggestion
		if err := legacyResp.Unmarshal(buf); err != nil {
```

state/store.go
```diff
@@ -651,14 +651,22 @@ func (store dbStore) LoadFinalizeBlockResponse(height int64) (*abci.FinalizeBloc
...
+			// only return an error, this method is only invoked through the `/block_results` not for state logic and
+			// some tests, so no need to exit cometbft if there's an error, just return it.
+			return nil, ErrFinalizeBlockResponseCorruptedError
```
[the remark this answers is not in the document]
sergio-mena: 💯

state/store.go
```diff
@@ -651,14 +651,22 @@ func (store dbStore) LoadFinalizeBlockResponse(height int64) (*abci.FinalizeBloc
 
 	resp := new(abci.FinalizeBlockResponse)
 	err = resp.Unmarshal(buf)
+	// if an error is returned, then it needs to be handled here
 	if err != nil {
+		return nil, ErrFinalizeBlockResponseUnmarshalError
+	}
```
sergio-mena, line 657: So, the bug is that a `LegacyABCIResponses` struct _can_ be unmarshalled as `FinalizeBlockResponse` with no error (albeit with incomplete information), is that right?

Also, if seems that this logic wasn't really unit-tested (this is coming from famous #9468), otherwise we would have caught it. Could you please add a UT that ... ?
* fails without this PR
* passes with this PR
🙏

andynog: OK, I will add a test

state/errors.go
andynog: I will wrap this so these errors will change and will have Error() and Wrap() so will keep them separate

state/store.go
```diff
@@ -651,14 +651,22 @@ func (store dbStore) LoadFinalizeBlockResponse(height int64) (*abci.FinalizeBloc
 
 	resp := new(abci.FinalizeBlockResponse)
 	err = resp.Unmarshal(buf)
+	// if an error is returned, then it needs to be handled here
 	if err != nil {
+		return nil, ErrFinalizeBlockResponseUnmarshalError
```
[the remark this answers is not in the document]
andynog: I've refactored the errors a bit. I didn't implement a `Wrap()` because it would surface the internal storage error to the API response call which is usually not a best practice. I think it's better to log the error from the store and then just return additional context (height) as the other errors are doing

state/store.go
andynog: I've implemented a test under `api/api_test.go`, this test tries to create a message at the v0.37 level using the same proto definitions and then convert it to a legacy ABCI response. I think this is a good way to test this use cases across proto versions.

I look into creating a test using the store methods SaveABCIResponseInfo but the way the code is structured in store a lot of things are internal and cannot be used in the test, we'd need to refactor things quite a bit. So focusing on the proto messages I think is a good way to test. I might look into testing something generating a message in v0.37 and using the byte buffer in a test, that would simulate some the storage too.

andynog:
I am working on some additional tests to ensure the logic holds and it makes sense

.changelog/unreleased/bug-fixes/3002-invalid-txs-results.md
```diff
@@ -0,0 +1,3 @@
+- `[rpc]` Fix an issue where a legacy ABCI response, created on `v0.37` or before, is not returned properly in `v0.38` and up
+- on the `/block_results` RPC endpoint.
```
cason, line 2:
```suggestion
on the `/block_results` RPC endpoint.
```

api/api_test.go
cason, line 1: I'd say that this test belongs to the `state` package. In any case, I would not keep it in this package...

rpc/core/blocks.go
```diff
@@ -193,6 +193,7 @@ func (env *Environment) BlockResults(_ *rpctypes.Context, heightPtr *int64) (*ct
 		FinalizeBlockEvents:   results.Events,
 		ValidatorUpdates:      results.ValidatorUpdates,
 		ConsensusParamUpdates: results.ConsensusParamUpdates,
+		AppHash:               results.AppHash,
```
cason, line 196: Is this field returned via RPC? Or at least was it before (e.g., v0.38.x)?

cason (commented):
No reserve against this change.

I would move the added test unit elsewhere thought.

andynog:
Just a quick update on this. I'm still doing some additional testing. The one I mentioned above was giving different results from other tests I'm performing with an "enhanced" kvstore (returns more information in a block_results). It seems the original logic was OK at least going from a `v0.37` to `v0.38`. Will add more context here once I have more information.

sergio-mena:
👍 let us know when you're done with the extra tests and I'll take another look

andynog:
> 👍 let us know when you're done with the extra tests and I'll take another look

Hi @sergio-mena, I've added an extra test `Test_V1Beta2_ABCIResponses_To_FinalizeBlockResponse` which un-marshalls the binary file (persisted ABCI Responses) as a FinalizeBlockResponse and if you check the test it is returning an error, so the old logic (`err != nil` instead of the suggested one I changed to `resp.AppHash == nil`) should still be good. 

I think in my troubleshooting above I was not getting the error because the `kvstore` response doesn't include data for all the fields, but once I generated a `v0.37` response (the `.bin` file) including DeliverTx, BeginBlock events I got the error when un-marshalling the binary as a `FinalizeBlockResponse`.

rpc/core/blocks.go
melekes: What about `LastResultsHash`? Should we include it too?

state/store.go
```diff
@@ -651,14 +651,23 @@ func (store dbStore) LoadFinalizeBlockResponse(height int64) (*abci.FinalizeBloc
 
 	resp := new(abci.FinalizeBlockResponse)
 	err = resp.Unmarshal(buf)
+	// if an error is returned, then it needs to be handled here
 	if err != nil {
+		store.Logger.Debug("failed to unmarshall ABCI response: %s", err.Error())
```
melekes, line 656: Why not include err in `ErrABCIResponseResponseUnmarshalForHeight`?

state/store.go
```diff
@@ -651,14 +651,23 @@ func (store dbStore) LoadFinalizeBlockResponse(height int64) (*abci.FinalizeBloc
...
 		// we try to unmarshal that
 		legacyResp := new(cmtstate.LegacyABCIResponses)
-		rerr := legacyResp.Unmarshal(buf)
-		if rerr != nil {
-			cmtos.Exit(fmt.Sprintf(`LoadFinalizeBlockResponse: Data has been corrupted or its spec has
-					changed: %v\n`, err))
+		if err := legacyResp.Unmarshal(buf); err != nil {
+			// only return an error, this method is only invoked through the `/block_results` not for state logic and
+			// some tests, so no need to exit cometbft if there's an error, just return it.
+			store.Logger.Debug("failed to unmarshall legacy ABCI response: %s", err.Error())
```
melekes, line 669: same comment

api/api_test.go
andynog: moved inside state in a `compatibility_test` file

state/store.go
```diff
@@ -651,14 +651,23 @@ func (store dbStore) LoadFinalizeBlockResponse(height int64) (*abci.FinalizeBloc
...
```
sergio-mena, line 669:
```suggestion
			store.Logger.Debug("failed to unmarshall FinalizeBlockResponse (also tried as legacy ABCI response): %s", err.Error())
```

sergio-mena: I would also include the error you get in line 662. The reason is, if it was indeed _not_ a legacy response, but it failed to deserialise, the error you get when you later try to deserialise as legacy is not useful.

state/store.go
```diff
@@ -651,14 +651,23 @@ func (store dbStore) LoadFinalizeBlockResponse(height int64) (*abci.FinalizeBloc
...
+			return nil, ErrABCIResponseCorruptedOrSpecChangeForHeight{Height: height}
```
sergio-mena, line 670: Also, if we don't break any API (not sure) it'd be good to include the error inside this error, so the caller can print it if needed

state/compatibility_test.go
```diff
@@ -0,0 +1,225 @@
...
+	statev1 "github.com/cometbft/cometbft/api/cometbft/state/v1"
+	statev1beta2 "github.com/cometbft/cometbft/api/cometbft/state/v1beta2"
+	statev1beta3 "github.com/cometbft/cometbft/api/cometbft/state/v1beta3"
+	typesv1beta1 "github.com/cometbft/cometbft/api/cometbft/types/v1beta1"
+	typesv1beta2 "github.com/cometbft/cometbft/api/cometbft/types/v1beta2"
+)
+
+// Compatibility test across different state proto versions
+
+// This test unmarshal a v1beta2.ABCIResponses as a statev1.LegacyABCIResponses
```
sergio-mena, line 22:
```suggestion
// This test unmarshals a v1beta2.ABCIResponses as a statev1.LegacyABCIResponses
```

state/compatibility_test.go
```diff
@@ -0,0 +1,225 @@
...
+		Value: "value",
+	}
+
+	deliverTxEvent := abciv1beta2.Event{
+		Type:       "deliver_tx_event",
+		Attributes: []abciv1beta2.EventAttribute{eventAttr},
+	}
+
+	responseDeliverTx := abciv1beta2.ResponseDeliverTx{
+		Code:   abci.CodeTypeOK,
```
sergio-mena, line 136: Should you populate the rest of the fields here? (`GasWanted`, `Codespace`, etc)

state/compatibility_test.go
```diff
@@ -0,0 +1,225 @@
...
+	// un-marshall a v1beta2 ABCI Response as a LegacyABCIResponse
+	legacyABCIResponse := new(statev1.LegacyABCIResponses)
+	err = legacyABCIResponse.Unmarshal(v1b2Resp)
+	require.NoError(t, err)
+	require.NotNil(t, legacyABCIResponse.DeliverTxs)
+	require.NotNil(t, legacyABCIResponse.BeginBlock)
+	require.NotNil(t, legacyABCIResponse.EndBlock)
+	require.Equal(t, 1, len(legacyABCIResponse.DeliverTxs))
+	require.Equal(t, 1, len(legacyABCIResponse.BeginBlock.Events))
+	require.Equal(t, 1, len(legacyABCIResponse.EndBlock.Events))
```
sergio-mena, line 41: I think we should be more exhaustive checking the values of these

state/compatibility_test.go
```diff
@@ -0,0 +1,225 @@
...
+	v1beta2ABCIResponses := newV1Beta2ABCIResponses()
+	data, err := v1beta2ABCIResponses.Marshal()
+	require.NoError(t, err)
+	require.NotNil(t, data)
+
+	// This cannot work since they have different schemas, a wrong wireType error is generated
+	finalizeBlockResponse := new(abci.FinalizeBlockResponse)
+	err = finalizeBlockResponse.Unmarshal(data)
+	require.Error(t, err)
+	require.ErrorContains(t, err, "unexpected EOF")
```
sergio-mena, line 61: I don't understand, by looking at the production changes below, I had understood that the bug is that this unmarshalling actually doesn't error, but leaves some fields unfilled...

state/compatibility_test.go
sergio-mena: Same in comments below

state/compatibility_test.go
```diff
@@ -0,0 +1,225 @@
...
+	require.NotNil(t, data)
+
+	// This works because they are equivalent protos and the fields are populated
+	// even if a field is null, it will be converted properly
+	legacyResponseWithNull := new(statev1beta3.LegacyABCIResponses)
+	err = legacyResponseWithNull.Unmarshal(data)
+	require.NoError(t, err)
+	require.NotNil(t, legacyResponseWithNull.DeliverTxs)
+	require.Nil(t, legacyResponseWithNull.EndBlock)
+	require.NotNil(t, legacyResponseWithNull.BeginBlock)
```
sergio-mena, line 120: Shouldn't we do the same checks here as in `TestStateProtoV1Beta2ToV1` above?

state/compatibility_test.go
```diff
@@ -0,0 +1,225 @@
+package state_test
```
sergio-mena, line 1: What I am missing here is a test case that exercises the (production) changes you made to `LoadFinalizeBlockResponse`.
In other words, if I take this PR and only apply the testing changes (without the production changes) is any test failing?

mbreithecker:
Please also add the `mode` attribute to the legacy response. Otherwise it is impossible to tell the difference between an BeginBlock and EndBlock event:

`state/store.go`

```go
// responseFinalizeBlockFromLegacy is a convenience function that takes the old abci responses and morphs
// it to the finalize block response. Note that the app hash is missing
func responseFinalizeBlockFromLegacy(legacyResp *cmtstate.LegacyABCIResponses) *abci.ResponseFinalizeBlock {

	// Add BeginBlock attribute to BeginBlock events
	for idx, event := range legacyResp.BeginBlock.Events {
		legacyResp.BeginBlock.Events[idx].Attributes = append(event.Attributes, abci.EventAttribute{
			Key:   "mode",
			Value: "BeginBlock",
			Index: false,
		})
	}

	// Add EndBlock attribute to BeginBlock events
	for idx, event := range legacyResp.EndBlock.Events {
		legacyResp.EndBlock.Events[idx].Attributes = append(event.Attributes, abci.EventAttribute{
			Key:   "mode",
			Value: "EndBlock",
			Index: false,
		})
	}

	return &abci.ResponseFinalizeBlock{
		TxResults:             legacyResp.DeliverTxs,
		ValidatorUpdates:      legacyResp.EndBlock.ValidatorUpdates,
		ConsensusParamUpdates: legacyResp.EndBlock.ConsensusParamUpdates,
		Events:                append(legacyResp.BeginBlock.Events, legacyResp.EndBlock.Events...),
		// NOTE: AppHash is missing in the response but will
		// be caught and filled in consensus/replay.go
	}
}
```

johnletey:
Noble would love to see this released as well!

andynog:
Just an update, still working on this, need to add some additional testing to ensure these changes are good especially backporting to old releases. But hopefully should be ready to be merged sooner than later.

andynog:
Just an update. The fix is still on-going. We believe we have a solution but might be hard to test with live production data (archive nodes). Would anyone that benefits from this fix be able to test against real data before we cut an official release ? 

Once the fix is in, we could backport to v0.38.x branch (but we would not cut a release until we have some confirmation the fix works). Then you can test using a binary compiled from that branch and let us know.

state/compatibility_test.go
andynog: Fixed it.

jhernandezb:
hi @andynog we could give it a try in our testnet but we will need what @mbreithecker suggested in https://github.com/cometbft/cometbft/pull/3031#issuecomment-2152095271 to be able to identify being/end block events

andynog:
Hi @mbreithecker and @jhernandez,

I made some changes a few days ago, including the attributes logic suggested. Do you need to test this on a v0.38 level branch, or are you okay with cherry-picking the changes? Please let me know. Thanks for your help!

mmsqe:
Thanks @andynog, patch works for v0.38.x.

melekes (approved):
👍

state/store.go
```diff
@@ -653,14 +653,22 @@ func (store dbStore) LoadFinalizeBlockResponse(height int64) (*abci.FinalizeBloc
...
 		// we try to unmarshal that
 		legacyResp := new(cmtstate.LegacyABCIResponses)
-		rerr := legacyResp.Unmarshal(buf)
-		if rerr != nil {
-			cmtos.Exit(fmt.Sprintf(`LoadFinalizeBlockResponse: Data has been corrupted or its spec has
-					changed: %v\n`, err))
+		if err := legacyResp.Unmarshal(buf); err != nil {
+			// only return an error, this method is only invoked through the `/block_results` not for state logic and
+			// some tests, so no need to exit cometbft if there's an error, just return it.
+			store.Logger.Debug("failed to unmarshall FinalizeBlockResponse (also tried as legacy ABCI response): %s", err.Error())
```
SpicyLemon, line 670: The log methods (e.g. `Debug`) take in a string message, then alternating key, value args. You should either wrap the two args in a `fmt.Sprintf` or else remove the `: %s` from the message and add an `"error"` key argument before `err.Error()`.


```suggestion
			store.Logger.Debug(fmt.Sprintf("failed to unmarshall FinalizeBlockResponse (also tried as legacy ABCI response): %v", err))
```

or 

```suggestion
			store.Logger.Debug("failed to unmarshall FinalizeBlockResponse (also tried as legacy ABCI response)", "error", err)
```

melekes:
@andynog can we merge this now since this patch was verified?

andynog:
> @andynog can we merge this now since this patch was verified?

Hi @melekes, I believe we can merge since we got a couple of confirmations that the patch works. One confirmation here and another [here](https://github.com/cometbft/cometbft/issues/3354#issuecomment-2201563686)

state/compatibility_test.go
```diff
@@ -98,81 +140,157 @@ func TestSaveLegacyAndLoadFinalizeBlock(t *testing.T) {
 	multiStore := NewMultiStore(stateDB, options, store)
 
 	// try with a complete ABCI Response
-	abciResponses := newV1Beta2ABCIResponses()
-	require.Equal(t, 1, len(abciResponses.DeliverTxs))
-	require.Equal(t, 1, len(abciResponses.BeginBlock.Events))
-	require.Equal(t, 1, len(abciResponses.EndBlock.Events))
```
sergio-mena: Could you keep these three tests? This way we make sure that e.g. check in line 157 isn't passing because 0 == 0... in this case line 158 will panic

state/compatibility_test.go
```diff
@@ -98,81 +140,157 @@ func TestSaveLegacyAndLoadFinalizeBlock(t *testing.T) {
...
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].Log, finalizeBlockResponse.TxResults[0].Log)
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].GasWanted, finalizeBlockResponse.TxResults[0].GasWanted)
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].GasUsed, finalizeBlockResponse.TxResults[0].GasUsed)
+	require.Equal(t, len(v1beta2ABCIResponses.DeliverTxs[0].Events), len(finalizeBlockResponse.TxResults[0].Events))
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].Events[0].Type, finalizeBlockResponse.TxResults[0].Events[0].Type)
+	require.Equal(t, len(v1beta2ABCIResponses.DeliverTxs[0].Events[0].Attributes), len(finalizeBlockResponse.TxResults[0].Events[0].Attributes))
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].Events[0].Attributes[0].Key, finalizeBlockResponse.TxResults[0].Events[0].Attributes[0].Key)
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].Events[0].Attributes[0].Value, finalizeBlockResponse.TxResults[0].Events[0].Attributes[0].Value)
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].Codespace, finalizeBlockResponse.TxResults[0].Codespace)
```
sergio-mena, line 168: Excellent! Just one question (not sure about this): if you call
```
require.Equal(v1beta2ABCIResponses.DeliverTxs, finalizeBlockResponse.TxResults)
```
Won't that check all the values and all the fields in the values?

sergio-mena: Actually, that's what you're doing in line 211 below (with a different value for `v1beta2ABCIResponses`)

state/compatibility_test.go
```diff
@@ -98,81 +140,157 @@ func TestSaveLegacyAndLoadFinalizeBlock(t *testing.T) {
...
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].Code, legacyABCIResponses.DeliverTxs[0].Code)
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].Data, legacyABCIResponses.DeliverTxs[0].Data)
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].Log, legacyABCIResponses.DeliverTxs[0].Log)
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].GasWanted, legacyABCIResponses.DeliverTxs[0].GasWanted)
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].GasUsed, legacyABCIResponses.DeliverTxs[0].GasUsed)
+	require.Equal(t, len(v1beta2ABCIResponses.DeliverTxs[0].Events), len(legacyABCIResponses.DeliverTxs[0].Events))
+	require.Equal(t, len(v1beta2ABCIResponses.DeliverTxs[0].Events[0].Attributes), len(legacyABCIResponses.DeliverTxs[0].Events[0].Attributes))
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].Events[0].Attributes[0].Key, legacyABCIResponses.DeliverTxs[0].Events[0].Attributes[0].Key)
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].Events[0].Attributes[0].Value, legacyABCIResponses.DeliverTxs[0].Events[0].Attributes[0].Value)
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].Codespace, legacyABCIResponses.DeliverTxs[0].Codespace)
```
sergio-mena, line 249: Same comment as above: wouldn't `require.Equal(t, v1beta2ABCIResponses.DeliverTxs, legacyABCIResponses.DeliverTxs)` be checking all elements and all fields one by one?

state/compatibility_test.go
```diff
@@ -98,81 +140,157 @@ func TestSaveLegacyAndLoadFinalizeBlock(t *testing.T) {
...
+
+	require.Equal(t, len(v1beta2ABCIResponses.BeginBlock.Events), len(legacyABCIResponses.BeginBlock.Events))
```
sergio-mena, line 251: Same here, with `require.Equal(t, v1beta2ABCIResponses.BeginBlock.Events, legacyABCIResponses.BeginBlock.Events`

state/compatibility_test.go
```diff
@@ -98,81 +140,157 @@ func TestSaveLegacyAndLoadFinalizeBlock(t *testing.T) {
...
+	require.Equal(t, v1beta2ABCIResponses.BeginBlock.Events[0].Type, legacyABCIResponses.BeginBlock.Events[0].Type)
+	require.Equal(t, len(v1beta2ABCIResponses.BeginBlock.Events[0].Attributes), len(legacyABCIResponses.BeginBlock.Events[0].Attributes))
+	require.Equal(t, v1beta2ABCIResponses.BeginBlock.Events[0].Attributes[0].Key, legacyABCIResponses.BeginBlock.Events[0].Attributes[0].Key)
+	require.Equal(t, v1beta2ABCIResponses.BeginBlock.Events[0].Attributes[0].Value, legacyABCIResponses.BeginBlock.Events[0].Attributes[0].Value)
```
sergio-mena, line 255: Same here, with `require.Equal(t, v1beta2ABCIResponses.BeginBlock.Events, legacyABCIResponses.BeginBlock.Events`

state/compatibility_test.go
```diff
@@ -98,81 +140,157 @@ func TestSaveLegacyAndLoadFinalizeBlock(t *testing.T) {
...
+}
+
+// This test unmarshal a v1beta2.ABCIResponses as a v1beta3.ResponseFinalizeBlock
```
sergio-mena, line 258:
```suggestion
// This test unmarshals a v1beta2.ABCIResponses as a v1beta3.ResponseFinalizeBlock
```

state/compatibility_test.go
```diff
@@ -204,15 +324,36 @@ func TestStateV1Beta2ABCIResponsesAsStateV1Beta3FinalizeBlockResponse(t *testing
...
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].Code, legacyABCIResponses.DeliverTxs[0].Code)
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].Data, legacyABCIResponses.DeliverTxs[0].Data)
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].Log, legacyABCIResponses.DeliverTxs[0].Log)
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].GasWanted, legacyABCIResponses.DeliverTxs[0].GasWanted)
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].GasUsed, legacyABCIResponses.DeliverTxs[0].GasUsed)
+	require.Equal(t, len(v1beta2ABCIResponses.DeliverTxs[0].Events), len(legacyABCIResponses.DeliverTxs[0].Events))
+	require.Equal(t, len(v1beta2ABCIResponses.DeliverTxs[0].Events[0].Attributes), len(legacyABCIResponses.DeliverTxs[0].Events[0].Attributes))
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].Events[0].Attributes[0].Key, legacyABCIResponses.DeliverTxs[0].Events[0].Attributes[0].Key)
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].Events[0].Attributes[0].Value, legacyABCIResponses.DeliverTxs[0].Events[0].Attributes[0].Value)
+	require.Equal(t, v1beta2ABCIResponses.DeliverTxs[0].Codespace, legacyABCIResponses.DeliverTxs[0].Codespace)
```
sergio-mena, line 344: Same comment

state/compatibility_test.go
```diff
@@ -204,15 +324,36 @@ func TestStateV1Beta2ABCIResponsesAsStateV1Beta3FinalizeBlockResponse(t *testing
...
+
+	require.Equal(t, len(v1beta2ABCIResponses.BeginBlock.Events), len(legacyABCIResponses.BeginBlock.Events))
+	require.Equal(t, v1beta2ABCIResponses.BeginBlock.Events[0].Type, legacyABCIResponses.BeginBlock.Events[0].Type)
+	require.Equal(t, len(v1beta2ABCIResponses.BeginBlock.Events[0].Attributes), len(legacyABCIResponses.BeginBlock.Events[0].Attributes))
+	require.Equal(t, v1beta2ABCIResponses.BeginBlock.Events[0].Attributes[0].Key, legacyABCIResponses.BeginBlock.Events[0].Attributes[0].Key)
+	require.Equal(t, v1beta2ABCIResponses.BeginBlock.Events[0].Attributes[0].Value, legacyABCIResponses.BeginBlock.Events[0].Attributes[0].Value)
```
sergio-mena, line 350: Same comment

state/compatibility_test.go
```diff
@@ -226,6 +367,89 @@ func TestStateV1Beta2ABCIResponsesWithNullAsStateV1Beta3FinalizeBlockResponse(t
 	require.NotNil(t, legacyResponseWithNull.DeliverTxs)
 	require.Nil(t, legacyResponseWithNull.EndBlock)
 	require.NotNil(t, legacyResponseWithNull.BeginBlock)
+
+	require.Equal(t, len(v1beta2ABCIResponsesWithNull.BeginBlock.Events), len(legacyResponseWithNull.BeginBlock.Events))
+	require.Equal(t, v1beta2ABCIResponsesWithNull.BeginBlock.Events[0].Type, legacyResponseWithNull.BeginBlock.Events[0].Type)
+	require.Equal(t, len(v1beta2ABCIResponsesWithNull.BeginBlock.Events[0].Attributes), len(legacyResponseWithNull.BeginBlock.Events[0].Attributes))
+	require.Equal(t, v1beta2ABCIResponsesWithNull.BeginBlock.Events[0].Attributes[0].Key, legacyResponseWithNull.BeginBlock.Events[0].Attributes[0].Key)
+	require.Equal(t, v1beta2ABCIResponsesWithNull.BeginBlock.Events[0].Attributes[0].Value, legacyResponseWithNull.BeginBlock.Events[0].Attributes[0].Value)
```
sergio-mena, line 375: Same

state/compatibility_test.go
```diff
@@ -226,6 +367,89 @@ func TestStateV1Beta2ABCIResponsesWithNullAsStateV1Beta3FinalizeBlockResponse(t
...
+	require.Equal(t, v1beta2ABCIResponsesWithNull.DeliverTxs[0].Data, legacyResponseWithNull.DeliverTxs[0].Data)
+	require.Equal(t, v1beta2ABCIResponsesWithNull.DeliverTxs[0].Log, legacyResponseWithNull.DeliverTxs[0].Log)
+	require.Equal(t, v1beta2ABCIResponsesWithNull.DeliverTxs[0].GasWanted, legacyResponseWithNull.DeliverTxs[0].GasWanted)
+	require.Equal(t, v1beta2ABCIResponsesWithNull.DeliverTxs[0].GasUsed, legacyResponseWithNull.DeliverTxs[0].GasUsed)
+	require.Equal(t, len(v1beta2ABCIResponsesWithNull.DeliverTxs[0].Events), len(legacyResponseWithNull.DeliverTxs[0].Events))
+	require.Equal(t, len(v1beta2ABCIResponsesWithNull.DeliverTxs[0].Events[0].Attributes), len(legacyResponseWithNull.DeliverTxs[0].Events[0].Attributes))
+	require.Equal(t, v1beta2ABCIResponsesWithNull.DeliverTxs[0].Events[0].Attributes[0].Key, legacyResponseWithNull.DeliverTxs[0].Events[0].Attributes[0].Key)
+	require.Equal(t, v1beta2ABCIResponsesWithNull.DeliverTxs[0].Events[0].Attributes[0].Value, legacyResponseWithNull.DeliverTxs[0].Events[0].Attributes[0].Value)
+	require.Equal(t, v1beta2ABCIResponsesWithNull.DeliverTxs[0].Codespace, legacyResponseWithNull.DeliverTxs[0].Codespace)
```
sergio-mena, line 387: Same

state/compatibility_test.go
```diff
@@ -226,6 +367,89 @@ func TestStateV1Beta2ABCIResponsesWithNullAsStateV1Beta3FinalizeBlockResponse(t
...
+	require.Equal(t, v1beta3ResponseFinalizeBlock.TxResults[0].Log, finalizeBlockResponse.TxResults[0].Log)
+	require.Equal(t, v1beta3ResponseFinalizeBlock.TxResults[0].GasWanted, finalizeBlockResponse.TxResults[0].GasWanted)
+	require.Equal(t, v1beta3ResponseFinalizeBlock.TxResults[0].GasUsed, finalizeBlockResponse.TxResults[0].GasUsed)
+	require.Equal(t, len(v1beta3ResponseFinalizeBlock.TxResults[0].Events), len(finalizeBlockResponse.TxResults[0].Events))
+	require.Equal(t, v1beta3ResponseFinalizeBlock.TxResults[0].Events[0].Type, finalizeBlockResponse.TxResults[0].Events[0].Type)
+	require.Equal(t, len(v1beta3ResponseFinalizeBlock.TxResults[0].Events[0].Attributes), len(finalizeBlockResponse.TxResults[0].Events[0].Attributes))
+	require.Equal(t, v1beta3ResponseFinalizeBlock.TxResults[0].Events[0].Attributes[0].Key, finalizeBlockResponse.TxResults[0].Events[0].Attributes[0].Key)
+	require.Equal(t, v1beta3ResponseFinalizeBlock.TxResults[0].Events[0].Attributes[0].Value, finalizeBlockResponse.TxResults[0].Events[0].Attributes[0].Value)
+	require.Equal(t, v1beta3ResponseFinalizeBlock.TxResults[0].Codespace, finalizeBlockResponse.TxResults[0].Codespace)
```
sergio-mena, line 424: same

state/compatibility_test.go
```diff
@@ -226,6 +367,89 @@ func TestStateV1Beta2ABCIResponsesWithNullAsStateV1Beta3FinalizeBlockResponse(t
...
+
+	require.Equal(t, len(v1beta3ResponseFinalizeBlock.Events), len(finalizeBlockResponse.Events))
+	require.Equal(t, v1beta3ResponseFinalizeBlock.Events[0].Type, finalizeBlockResponse.Events[0].Type)
+	require.Equal(t, len(v1beta3ResponseFinalizeBlock.Events[0].Attributes), len(finalizeBlockResponse.Events[0].Attributes))
+	require.Equal(t, v1beta3ResponseFinalizeBlock.Events[0].Attributes[0].Key, finalizeBlockResponse.Events[0].Attributes[0].Key)
+	require.Equal(t, v1beta3ResponseFinalizeBlock.Events[0].Attributes[0].Value, finalizeBlockResponse.Events[0].Attributes[0].Value)
```
sergio-mena, line 430: same

state/compatibility_test.go
```diff
@@ -226,6 +367,89 @@ func TestStateV1Beta2ABCIResponsesWithNullAsStateV1Beta3FinalizeBlockResponse(t
...
+	require.Equal(t, v1beta3ResponseFinalizeBlock.ConsensusParamUpdates.Evidence.MaxBytes, finalizeBlockResponse.ConsensusParamUpdates.Evidence.MaxBytes)
+	require.Equal(t, v1beta3ResponseFinalizeBlock.ConsensusParamUpdates.Validator.PubKeyTypes, finalizeBlockResponse.ConsensusParamUpdates.Validator.PubKeyTypes)
+	require.Equal(t, v1beta3ResponseFinalizeBlock.ConsensusParamUpdates.Version.App, finalizeBlockResponse.ConsensusParamUpdates.Version.App)
+	require.Equal(t, v1beta3ResponseFinalizeBlock.ConsensusParamUpdates.Synchrony.Precision, finalizeBlockResponse.ConsensusParamUpdates.Synchrony.Precision)
+	require.Equal(t, v1beta3ResponseFinalizeBlock.ConsensusParamUpdates.Synchrony.MessageDelay, finalizeBlockResponse.ConsensusParamUpdates.Synchrony.MessageDelay)
+	require.Equal(t, v1beta3ResponseFinalizeBlock.ConsensusParamUpdates.Feature.VoteExtensionsEnableHeight.Value, finalizeBlockResponse.ConsensusParamUpdates.Feature.VoteExtensionsEnableHeight.Value)
+	require.Equal(t, v1beta3ResponseFinalizeBlock.ConsensusParamUpdates.Feature.PbtsEnableHeight.Value, finalizeBlockResponse.ConsensusParamUpdates.Feature.PbtsEnableHeight.Value)
```
sergio-mena, line 442: Same: should comparing `v1beta3ResponseFinalizeBlock.ConsensusParamUpdates` and `finalizeBlockResponse.ConsensusParamUpdates` do the same check as these?

state/compatibility_test.go
andynog: I can't check them for equality straight away because they are based on different protos so the check fails (I tried that first), because even though structurally they might look the same their fields might come from a different version and the `Equals` fails

state/compatibility_test.go
andynog: Same answer as above, but here's the error if you try to test for equality

```
=== RUN   TestStateV1Beta2ABCIResponsesAsStateV1LegacyABCIResponse
    compatibility_test.go:239: 
        	Error Trace:	/Users/andynogueira/Development/Go/src/github.com/cometbft/cometbft/state/compatibility_test.go:239
        	Error:      	Not equal: 
        	            	expected: []*v1beta2.ResponseDeliverTx([]*v1beta2.ResponseDeliverTx{(*v1beta2.ResponseDeliverTx)(0x14000255b00)})
        	            	actual  : []*v1.ExecTxResult([]*v1.ExecTxResult{(*v1.ExecTxResult)(0x14000255b80)})
        	Test:       	TestStateV1Beta2ABCIResponsesAsStateV1LegacyABCIResponse
--- FAIL: TestStateV1Beta2ABCIResponsesAsStateV1LegacyABCIResponse (0.00s)

Expected :[]*v1beta2.ResponseDeliverTx([]*v1beta2.ResponseDeliverTx{(*v1beta2.ResponseDeliverTx)(0x14000255b00)})
Actual   :[]*v1.ExecTxResult([]*v1.ExecTxResult{(*v1.ExecTxResult)(0x14000255b80)})```

File: state/compatibility_test.go (hunk 23 above)
Comment (andynog) [reply state/compatibility_test.go:255]: same as above

File: state/compatibility_test.go (hunk 25 above)
Comment (andynog) [reply state/compatibility_test.go:344]: same as above

Comment (andynog) [reply state/compatibility_test.go:344]:
```
=== RUN   TestStateV1Beta2ABCIResponsesAsStateV1Beta3LegacyABCIResponse
    compatibility_test.go:335: 
        	Error Trace:	/Users/andynogueira/Development/Go/src/github.com/cometbft/cometbft/state/compatibility_test.go:335
        	Error:      	Not equal: 
        	            	expected: []*v1beta2.ResponseDeliverTx([]*v1beta2.ResponseDeliverTx{(*v1beta2.ResponseDeliverTx)(0x14000316400)})
        	            	actual  : []*v1beta3.ExecTxResult([]*v1beta3.ExecTxResult{(*v1beta3.ExecTxResult)(0x14000316480)})
        	Test:       	TestStateV1Beta2ABCIResponsesAsStateV1Beta3LegacyABCIResponse
--- FAIL: TestStateV1Beta2ABCIResponsesAsStateV1Beta3LegacyABCIResponse (0.00s)

Expected :[]*v1beta2.ResponseDeliverTx([]*v1beta2.ResponseDeliverTx{(*v1beta2.ResponseDeliverTx)(0x14000316400)})
Actual   :[]*v1beta3.ExecTxResult([]*v1beta3.ExecTxResult{(*v1beta3.ExecTxResult)(0x14000316480)})```

state/compatibility_test.go
andynog:
```
=== RUN   TestStateV1Beta2ABCIResponsesAsStateV1LegacyABCIResponse
    compatibility_test.go:252: 
        	Error Trace:	/Users/andynogueira/Development/Go/src/github.com/cometbft/cometbft/state/compatibility_test.go:252
        	Error:      	Not equal: 
        	            	expected: []v1beta2.Event([]v1beta2.Event{v1beta2.Event{Type:"begin_block_event", Attributes:[]v1beta2.EventAttribute{v1beta2.EventAttribute{Key:"key", Value:"value", Index:false}}}})
        	            	actual  : []v1.Event([]v1.Event{v1.Event{Type:"begin_block_event", Attributes:[]v1.EventAttribute{v1.EventAttribute{Key:"key", Value:"value", Index:false}}}})
        	Test:       	TestStateV1Beta2ABCIResponsesAsStateV1LegacyABCIResponse
--- FAIL: TestStateV1Beta2ABCIResponsesAsStateV1LegacyABCIResponse (0.00s)

Expected :[]v1beta2.Event([]v1beta2.Event{v1beta2.Event{Type:"begin_block_event", Attributes:[]v1beta2.EventAttribute{v1beta2.EventAttribute{Key:"key", Value:"value", Index:false}}}})
Actual   :[]v1.Event([]v1.Event{v1.Event{Type:"begin_block_event", Attributes:[]v1.EventAttribute{v1.EventAttribute{Key:"key", Value:"value", Index:false}}}})```

File: state/compatibility_test.go (hunk 20 above)
Comment (andynog) [reply state/compatibility_test.go:168]:
```
=== RUN   TestSaveLegacyAndLoadFinalizeBlock
    compatibility_test.go:158: 
        	Error Trace:	/Users/andynogueira/Development/Go/src/github.com/cometbft/cometbft/state/compatibility_test.go:158
        	Error:      	Not equal: 
        	            	expected: []*v1beta2.ResponseDeliverTx([]*v1beta2.ResponseDeliverTx{(*v1beta2.ResponseDeliverTx)(0x140002e0380)})
        	            	actual  : []*v1.ExecTxResult([]*v1.ExecTxResult{(*v1.ExecTxResult)(0x140002e0400)})
        	Test:       	TestSaveLegacyAndLoadFinalizeBlock
--- FAIL: TestSaveLegacyAndLoadFinalizeBlock (0.00s)

Expected :[]*v1beta2.ResponseDeliverTx([]*v1beta2.ResponseDeliverTx{(*v1beta2.ResponseDeliverTx)(0x140002e0380)})
Actual   :[]*v1.ExecTxResult([]*v1.ExecTxResult{(*v1.ExecTxResult)(0x140002e0400)})
```

File: state/compatibility_test.go (hunk 28 above)
Comment (andynog) [reply state/compatibility_test.go:387]:
```
=== RUN   TestStateV1Beta2ABCIResponsesWithNullAsStateV1Beta3LegacyABCIResponse
    compatibility_test.go:380: 
        	Error Trace:	/Users/andynogueira/Development/Go/src/github.com/cometbft/cometbft/state/compatibility_test.go:380
        	Error:      	Not equal: 
        	            	expected: []*v1beta2.ResponseDeliverTx([]*v1beta2.ResponseDeliverTx{(*v1beta2.ResponseDeliverTx)(0x14000326600)})
        	            	actual  : []*v1beta3.ExecTxResult([]*v1beta3.ExecTxResult{(*v1beta3.ExecTxResult)(0x14000326680)})
        	Test:       	TestStateV1Beta2ABCIResponsesWithNullAsStateV1Beta3LegacyABCIResponse
--- FAIL: TestStateV1Beta2ABCIResponsesWithNullAsStateV1Beta3LegacyABCIResponse (0.00s)

Expected :[]*v1beta2.ResponseDeliverTx([]*v1beta2.ResponseDeliverTx{(*v1beta2.ResponseDeliverTx)(0x14000326600)})
Actual   :[]*v1beta3.ExecTxResult([]*v1beta3.ExecTxResult{(*v1beta3.ExecTxResult)(0x14000326680)})```

state/compatibility_test.go
andynog:
```
=== RUN   TestStateV1Beta3ResponsesFinalizeBlockAsV1FinalizeBlockResponse
    compatibility_test.go:416: 
        	Error Trace:	/Users/andynogueira/Development/Go/src/github.com/cometbft/cometbft/state/compatibility_test.go:416
        	Error:      	Not equal: 
        	            	expected: []*v1beta3.ExecTxResult([]*v1beta3.ExecTxResult{(*v1beta3.ExecTxResult)(0x1400030ff00)})
        	            	actual  : []*v1.ExecTxResult([]*v1.ExecTxResult{(*v1.ExecTxResult)(0x14000316000)})
        	Test:       	TestStateV1Beta3ResponsesFinalizeBlockAsV1FinalizeBlockResponse
--- FAIL: TestStateV1Beta3ResponsesFinalizeBlockAsV1FinalizeBlockResponse (0.00s)

Expected :[]*v1beta3.ExecTxResult([]*v1beta3.ExecTxResult{(*v1beta3.ExecTxResult)(0x1400030ff00)})
Actual   :[]*v1.ExecTxResult([]*v1.ExecTxResult{(*v1.ExecTxResult)(0x14000316000)})
```

state/compatibility_test.go
andynog:
```
=== RUN   TestStateV1Beta3ResponsesFinalizeBlockAsV1FinalizeBlockResponse
    compatibility_test.go:429: 
        	Error Trace:	/Users/andynogueira/Development/Go/src/github.com/cometbft/cometbft/state/compatibility_test.go:429
        	Error:      	Not equal: 
        	            	expected: []v1beta2.Event([]v1beta2.Event{v1beta2.Event{Type:"tx_event", Attributes:[]v1beta2.EventAttribute{v1beta2.EventAttribute{Key:"key", Value:"value", Index:false}}}, v1beta2.Event{Type:"one_event", Attributes:[]v1beta2.EventAttribute{v1beta2.EventAttribute{Key:"key", Value:"value", Index:false}}}, v1beta2.Event{Type:"two_event", Attributes:[]v1beta2.EventAttribute{v1beta2.EventAttribute{Key:"key", Value:"value", Index:false}}}})
        	            	actual  : []v1.Event([]v1.Event{v1.Event{Type:"tx_event", Attributes:[]v1.EventAttribute{v1.EventAttribute{Key:"key", Value:"value", Index:false}}}, v1.Event{Type:"one_event", Attributes:[]v1.EventAttribute{v1.EventAttribute{Key:"key", Value:"value", Index:false}}}, v1.Event{Type:"two_event", Attributes:[]v1.EventAttribute{v1.EventAttribute{Key:"key", Value:"value", Index:false}}}})
        	Test:       	TestStateV1Beta3ResponsesFinalizeBlockAsV1FinalizeBlockResponse
--- FAIL: TestStateV1Beta3ResponsesFinalizeBlockAsV1FinalizeBlockResponse (0.00s)

Expected :[]v1beta2.Event([]v1beta2.Event{v1beta2.Event{Type:"tx_event", Attributes:[]v1beta2.EventAttribute{v1beta2.EventAttribute{Key:"key", Value:"value", Index:false}}}, v1beta2.Event{Type:"one_event", Attributes:[]v1beta2.EventAttribute{v1beta2.EventAttri ...

Actual   :[]v1.Event([]v1.Event{v1.Event{Type:"tx_event", Attributes:[]v1.EventAttribute{v1.EventAttribute{Key:"key", Value:"value", Index:false}}}, v1.Event{Type:"one_event", Attributes:[]v1.EventAttribute{v1.EventAttribute{Key:"key", Value:"value", Index:fal ...
```

state/compatibility_test.go
andynog: That can work. I'll add the test. It just works because `v1beta3ResponseFinalizeBlock.ConsensusParamUpdates` uses a `v1.ConsensusParams` the same used by `finalizeBlockResponse.ConsensusParamUpdates` but this might change in the future because of #3090. So I'll add the equality test but will leave the individual ones too.

state/compatibility_test.go
andynog: good point, I've added the tests

state/store.go
andynog: also logging as `Error()` not `Debug()`

state/compatibility_test.go
sergio-mena: Thanks for giving it a try

sergio-mena (approved):
Thanks a lot for this!

andynog:
Hi @mbreithecker, @mmsqe, @johnletey, @jhernandezb and @SpicyLemon 

This fixed has been back ported to `v0.38.x` (#3435) and it will be available in the next release in the `v0.38` line. Thanks for all feedback and testing. Please let us know if you encounter any issues while using the fix.

Thanks again,

Andy.

mbreithecker:
Thank you very much!

PR #3036, opened 2024-05-08
feat(confix): copy confix from cosmos sdk

melekes:
# Confix

`Confix` is a configuration management tool that allows you to manage your configuration via CLI.

It is based on the [CometBFT RFC 019](https://github.com/cometbft/cometbft/blob/5013bc3f4a6d64dcc2bf02ccc002ebc9881c62e4/docs/rfc/rfc-019-config-version.md).

### Get

Get a configuration value, e.g.:

```shell
cometbft config get pruning # gets the value pruning
cometbft config get chain-id # gets the value chain-id
```

### Set

Set a configuration value, e.g.:

```shell
cometbft config set pruning "enabled" # sets the value pruning
cometbft config set chain-id "foo-1" # sets the value chain-id
```
### Migrate

Migrate a configuration file to a new version:

```shell
cometbft config migrate v0.38 # migrates defaultHome/config/config.toml to the latest v0.38 config
```

### Diff

Get the diff between a given configuration file and the default configuration
file, e.g.:

```shell
cometbft config diff v0.38 # gets the diff between defaultHome/config/config.toml and the latest v0.38 config
```

### View

View a configuration file, e.g:

```shell
cometbft config view # views the current config
```

## Credits

This project is based on the [CometBFT RFC 019](https://github.com/cometbft/cometbft/blob/5013bc3f4a6d64dcc2bf02ccc002ebc9881c62e4/docs/rfc/rfc-019-config-version.md) and their own implementation of [confix](https://github.com/cometbft/cometbft/blob/v0.36.x/scripts/confix/confix.go).
Most of the code is copied over from [Cosmos SDK](https://github.com/cosmos/cosmos-sdk/tree/main/tools/confix).

---

#### PR checklist


melekes:
Tested it manually starting from `v0.34` up to `v1.0`.

ValarDragon (approved):
utACK

ValarDragon:
Anything blocking this being merged :eyes:

cason:
> Anything blocking this being merged 👀

Having another person testing it, lots of code to review...

internal/confix/data/v0.34.toml
```diff
@@ -0,0 +1,487 @@
...
+#######################################################################
+###                   Main Base Config Options                      ###
+#######################################################################
+
+# TCP or UNIX socket address of the ABCI application,
+# or the name of an ABCI application compiled in with the CometBFT binary
+proxy_app = "tcp://127.0.0.1:26658"
+
+# A custom human readable name for this node
+moniker = "test"
```
cason, line 18: This will always differ, should we keep or comment it?

internal/confix/data/v0.34.toml
```diff
@@ -0,0 +1,487 @@
...
+
+# If this node is many blocks behind the tip of the chain, FastSync
+# allows them to catchup quickly by downloading blocks in parallel
+# and verifying their commits
+fast_sync = true
```
cason, line 23: This is a tag not used from v0.37.x, I think. Should we report that it is non-existing, invalid on newer versions?

cason (approved):
This is a nice tool to have. I think we are only observing behaviors not considered when people start using it.

General comments:
- The command creates a CometBFT config directory (as `init`) when it is not present. Is this the behavior we want?
- Migrate only works when increasing the version number. Not a problem, but probably we should document this.

internal/confix/data/v0.34.toml
melekes: > This is a tag not used from v0.37.x, I think. Should we report that it is non-existing, invalid on newer versions?

It will be removed during the upgrade.

internal/confix/data/v0.34.toml
melekes: Keeping it is fine imho

PR #3044, opened 2024-05-08
chore: simplify addvote

faddat:
This PR simplifies the AddVote function, working towards the goal of enabling the maintidx linter

---

#### PR checklist

- [ ] Tests written/updated


internal/consensus/state.go
```diff
@@ -2348,124 +2400,122 @@ func (cs *State) addVote(vote *types.Vote, peerID p2p.ID) (added bool, err error
...
+func (cs *State) updateMetricsAndEvents(vote *types.Vote) {
 	if vote.Round == cs.Round {
 		vals := cs.state.Validators
 		_, val := vals.GetByIndex(vote.ValidatorIndex)
 		cs.metrics.MarkVoteReceived(vote.Type, val.VotingPower, vals.TotalVotingPower())
 	}
 
 	if err := cs.eventBus.PublishEventVote(types.EventDataVote{Vote: vote}); err != nil {
-		return added, err
+		cs.Logger.Error("failed to publish event vote", "err", err)
```
melekes, line 2425: Are we ignoring this error now?

melekes (approved):
Thanks @faddat ❤️

internal/consensus/state.go
```diff
@@ -2254,7 +2254,54 @@ func (cs *State) tryAddVote(vote *types.Vote, peerID p2p.ID) (bool, error) {
...
+		cs.handlePrecommit(vote)
+	default:
+		panic(fmt.Sprintf("unexpected vote type %v", vote.Type))
+	}
+
+	return true, nil
+}
+
+// logVoteDetails logs the details of a vote.
+func (cs *State) logVoteDetails(vote *types.Vote) {
```
cason, line 2304: I don't see the point of moving this to a method.

internal/consensus/state.go
```diff
@@ -2264,54 +2311,59 @@ func (cs *State) addVote(vote *types.Vote, peerID p2p.ID) (added bool, err error
 		"extLen", len(vote.Extension),
 		"extSigLen", len(vote.ExtensionSignature),
 	)
+}
 
-	if vote.Height < cs.Height || (vote.Height == cs.Height && vote.Round < cs.Round) {
-		cs.metrics.MarkLateVote(vote.Type)
-	}
+// isLateVote checks if a vote is a late vote.
+func (cs *State) isLateVote(vote *types.Vote) bool {
```
cason, line 2317: This is harder to read than the original version.

The same for the similar helpers. I think that the content of the ifs should be maintained for better readability. A comment can do the work of explaining what is being evaluated.

internal/consensus/state.go
```diff
@@ -2264,54 +2311,59 @@ func (cs *State) addVote(vote *types.Vote, peerID p2p.ID) (added bool, err error
...
-		added, err = cs.LastCommit.AddVote(vote)
-		if !added {
-			// If the vote wasn't added but there's no error, its a duplicate vote
-			if err == nil {
-				cs.metrics.DuplicateVote.Add(1)
-			}
-			return added, err
-		}
+// handlePrecommitForPreviousHeight handles a precommit vote for the previous height.
+func (cs *State) handlePrecommitForPreviousHeight(vote *types.Vote) (bool, error) {
```
cason, line 2327: This could be something like handleLastCommitVote.

internal/consensus/state.go
```diff
@@ -2264,54 +2311,59 @@ func (cs *State) addVote(vote *types.Vote, peerID p2p.ID) (added bool, err error
...
+	if cs.Step != cstypes.RoundStepNewHeight {
```
cason, line 2328: I know this was in the original code, but I don't see much point in this test with debug-only purposes.

We only add votes to `cs.LastCommit` when it is a precommit from the previous height **and** we are in the new_height round step.

internal/consensus/state.go
```diff
@@ -2254,7 +2254,54 @@ func (cs *State) tryAddVote(vote *types.Vote, peerID p2p.ID) (bool, error) {
...
+
+	// A precommit for the previous height?
+	// These come in while we wait timeoutCommit
+	if cs.isPrecommitForPreviousHeight(vote) {
+		return cs.handlePrecommitForPreviousHeight(vote)
+	}
+
+	// Height mismatch is ignored.
+	// Not necessarily a bad peer, but not favorable behavior.
+	if cs.isHeightMismatch(vote) {
```
cason, line 2273: How is this code easier to read than `vote.Height != cs.Height`?

internal/consensus/state.go
```diff
@@ -2264,54 +2311,59 @@ func (cs *State) addVote(vote *types.Vote, peerID p2p.ID) (added bool, err error
...
+		cs.Logger.Debug("precommit vote came in after commit timeout and has been ignored", "vote", vote)
+		return false, nil
+	}
 
-		cs.Logger.Debug("added vote to last precommits", "last_commit", cs.LastCommit.StringShort())
-		if err := cs.eventBus.PublishEventVote(types.EventDataVote{Vote: vote}); err != nil {
-			return added, err
+	added, err := cs.LastCommit.AddVote(vote)
+	if !added {
+		// If the vote wasn't added but there's no error, it's a duplicate vote
```
cason, line 2336: There is a `handleDuplicateVote` method...

cason (commented):
Thank you for this.

While I agree that some refactoring would increase readability, I would prefer moving outside from `addVote` only auxiliary logic.

faddat:
@cason - appreciate your comments, and will give this a proper go-through.

internal/consensus/state.go
faddat: will check that.

internal/consensus/state.go
faddat: Moving this out makes a lot of sense to me.  It takes something that isn't "core" and moves it out.  What am I missing?

internal/consensus/state.go
faddat: @melekes it's safe to say, that if we are ignoring that error, it should be found by a test, right?

Would it make sense to address this by adding a test?

internal/consensus/state.go
faddat: english.  This version calls a method whose name is self descriptive.  

But to change it back is nbd

faddat:
Proposal: we make addvote its own file

internal/consensus/state.go
faddat: gotcha, makes sense.

internal/consensus/state.go
```diff
@@ -2346,124 +2398,122 @@ func (cs *State) addVote(vote *types.Vote, peerID p2p.ID) (added bool, err error
...
+			cs.enterPrevote(cs.Height, cs.Round)
+		}
 	}
+}
 
-	return added, err
+// handlePrecommit handles a precommit.
+func (cs *State) handlePrecommit(vote *types.Vote) {
+	precommits := cs.Votes.Precommits(vote.Round)
+	cs.Logger.Debug("added vote to precommit",
```
faddat, line 2492: this here could also be its own method

internal/consensus/state.go
faddat: we definitely are.  Nice catch!  Fixing...

internal/consensus/state.go
faddat: unresolve if you really want this one changed.  Some of the others were very good catches, thank you!

ValarDragon:
I recommend we close this and do some simpler PR's to reduce complexity. It may be helpful to examine some of the simplification as what can be re-used between the reactor and state.go

melekes:
> I recommend we close this and do some simpler PR's to reduce complexity. It may be helpful to examine some of the simplification as what can be re-used between the reactor and state.go

Agree. @faddat please let us know if you want to break this work into smaller PRs as per @ValarDragon comment.

PR #3070, opened 2024-05-13
perf(consensus): add simplistic block validation cache

### LINKED ISSUE #2854: enterPrecommit is almost always double verifying the block

sergio-mena:
Closes #2854

Follow up from #2960

This PR adds a simplistic 1-element block validation cache in `BlockExecutor`. This addresses a performance bottleneck raised as part of #2854

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/state/execution.go
+++ b/state/execution.go
@@ -41,6 +41,9 @@ type BlockExecutor struct {
 	mempool mempool.Mempool
 	evpool  EvidencePool
 
+	// 1-element cache of validated blocks
+	lastValidatedBlock *types.Block
+
 	logger log.Logger
 
 	metrics *Metrics
@@ -194,9 +197,11 @@ func (blockExec *BlockExecutor) ProcessProposal(
 // Validation does not mutate state, but does require historical information from the stateDB,
 // ie. to verify evidence from a validator at an old height.
 func (blockExec *BlockExecutor) ValidateBlock(state State, block *types.Block) error {
-	err := validateBlock(state, block)
-	if err != nil {
-		return err
+	if !blockExec.lastValidatedBlock.HashesTo(block.Hash()) {
+		if err := validateBlock(state, block); err != nil {
+			return err
+		}
+		blockExec.lastValidatedBlock = block
 	}
 	return blockExec.evpool.CheckEvidence(block.Evidence.Evidence)
 }


cason (approved):
Simple yet powerful. Thanks :)

ValarDragon (approved):
Awesome, looks great to me!

melekes:
Do we want to backport this?

sergio-mena:
@melekes AFAIU, @ValarDragon and team need this, but they are cherry-picking to their fork (they just want to know we're up-streaming these changes, so they'll eventually disappear from their fork; @ValarDragon please chime in if that's not the case). So I see no point in backporting to `v0.37.` or `v0.38.x`. As for `v1.x` we are about to cut RC1 (we are long past "code freeze", which we reached when we carried out QA), so backports to `v1.x` should be done under the same conditions as the older branches at this point.

Changes to consensus a delicate, I'd prefer to backport only if needed (bug, security issue, etc) or if totally harmless (doc, comments, logs, etc).

ValarDragon:
I'd like to get this backported to v0.38.x if folks are fine with this. Our standing plan was to do an SDK 50 upgrade and be pretty on par with v0.38.x . (I am not caught up on whats the latency / breaking behavior of v1.x vs v0.38.x)

sergio-mena:
@ValarDragon For backporting changes -- that are not bugs, or security related -- to release branches, our policy is to backport to the corresponding `*-experimental` branch. In the case of `v0.38.x`, the branch is `v0.38.x-experimental`. The difference is that we do not _guarantee_ that the code changes landing on the `*-experimental` branch underwent any quality assurance (IOW, the `*-experimental` branches are not considered code-frozen). The advantage is that it's us who carry out the backport (solving conflicts, adapting where necessary) rather than the users.

Is backporting this to `v0.38.x-experimental` good enough for you?

ValarDragon:
Yes!

ValarDragon:
In a patch release and working great!

PR #3074, opened 2024-05-14
chore(deps): remove `minio/sha256-simd`

### LINKED ISSUE #3071: Remove dependency from `minio/sha256-simd` library
The dependency to this library was introduced by https://github.com/cometbft/cometbft/pull/2765.

It is present on `main` and `v1.x` branches (which is still unreleased).

We should at the moment stick to the standard `crypto/sha256` library, with the same purpose.

The adoption of `minio/sha256-simd` was proposed in https://github.com/cometbft/cometbft/pull/2909, but at the moment we are not planning to switch to this non-standard library, as summarized in the following comment:

> We discussed about this in our internal meeting today. These are the conclusions:
> 
> 1. We currently don't have performance data that shows an obvious gain in the production setups where CometBFT is used
> 2. We could gather such performance data ourselves as a due-diligence activity, but we don't have cycles ATM
> 3. There is an ongoing longer term solution, described in #2241, whereby an app can have CometBFT use custom hashing logic.
>     * This longer term solution will render our efforts in point 2 unnecessary
> 
> Therefore, we will:
> 
> * Use this issue to move all imports of `minio/sha256-simd` (BLS-related code) to `crypto/sha256`
> * Wait for #2241 to be completed so that users can do what is suggested in this issue on app-side code

_Originally posted by @sergio-mena in https://github.com/cometbft/cometbft/issues/2909#issuecomment-2107944974_

melekes:
Closes #3071

---

#### PR checklist

### DIFF
--- a/crypto/bls12381/key_bls12381.go
+++ b/crypto/bls12381/key_bls12381.go
@@ -4,8 +4,7 @@ package bls12381
 
 import (
 	"bytes"
-
-	sha256 "github.com/minio/sha256-simd"
+	"crypto/sha256"
 
 	"github.com/cometbft/cometbft/crypto"
 	bls12381 "github.com/cosmos/crypto/curves/bls12381"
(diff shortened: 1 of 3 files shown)


sergio-mena (approved):
👍

PR #2890, opened 2024-04-25
feat(proto)!: replace `pub_key` with `pub_key_type`

melekes:
and `pub_key_bytes` in `privval.PubKeyResponse`, `types.Validator` and `types.SimpleValidator`.

Closes #2878
Follow-up to https://github.com/cometbft/cometbft/pull/2843

---

#### PR checklist

- [x] Tests written/updated
- [x] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments
- [x] Title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) spec

PR #3077, opened 2024-05-14
fix(proto): deprecate `pub_key` in `Validator` instead of removing it

### LINKED ISSUE #3075: proto: make #2890 non-breaking
PR: https://github.com/cometbft/cometbft/pull/2890

Refs https://github.com/cosmos/cosmos-sdk/issues/20318

melekes:
which makes the change non-breaking (i.e., chains using v0.38 can use v1 without a hard fork)

* revert changes to `SimpleValidator` because it breaks consensus (`v1 valSet hash != v0.38 valSet hash`).

Closes #3075

---

#### PR checklist

- [ ] Tests written/updated


proto/cometbft/types/v1/validator.proto
```diff
@@ -30,12 +31,11 @@ message ValidatorSet {
 // Validator represents a node participating in the consensus protocol.
 message Validator {
   bytes  address           = 1;
+  cometbft.crypto.v1.PublicKey pub_key = 2 [(gogoproto.nullable) = false, deprecated = true];
```
sergio-mena, line 34: `(gogoproto.nullable) = false`, should it rather be `true`?

types/validator.go
```diff
@@ -159,7 +159,10 @@ func ValidatorFromProto(vp *cmtproto.Validator) (*Validator, error) {
 
 	pk, err := ce.PubKeyFromTypeAndBytes(vp.PubKeyType, vp.PubKeyBytes)
 	if err != nil {
-		return nil, err
+		pk, err = ce.PubKeyFromProto(vp.PubKey)
+		if err != nil {
+			return nil, err
+		}
```
sergio-mena, line 165: In this case, before continuing I think we should:
* Extract the keytype (string) from the public key itself
* Assign the keytype to `pub_key_type`
* Nullify pub_key

sergio-mena:
Thanks for this @melekes !  Are there other possible structs using `cometbft.crypto.v1.PublicKey` that might also have been persisted?

types/validator.go
melekes: When the state is updated, it will be saved in a new format. The error is only returned upon loading the state from the state store (i.e., when we're loading the old state using Comet v1).

melekes:
> Thanks for this @melekes ! Are there other possible structs using `cometbft.crypto.v1.PublicKey` that might also have been persisted?

we are calculating `validators_hash` and `next_validators_hash` differently in https://github.com/cometbft/cometbft/pull/2890.

melekes:
- [ ] Add tests that check if we're breaking the header's hash

melekes:
> * [ ]  Add tests that check if we're breaking the header's hash

I'm going to add tests in a separate PR.

melekes:
Patch has been verified manually in https://github.com/cosmos/cosmos-sdk/issues/20318#issuecomment-2111833775

sergio-mena (approved):
Changes in this PR LGTM.
I have a question: given our current understanding on how difficult it is to get rid of `PublicKey` from the `Validator` structure (if possible at all), is it really a good move to introduce `pub_key_type`?

melekes:
> I have a question: given our current understanding on how difficult it is to get rid of PublicKey from the Validator structure (if possible at all), is it really a good move to introduce pub_key_type?

PublicKey will stay in `SimpleValidator` until we decide to break Header's hash. Note it's not user-facing. Even though SimpleValidator is exposed in proto, it's not used to communicate with the ABCI app. I would say it's still worth it from the usability perspective. @tac0turtle please comment if you disagree.

sergio-mena:
@melekes So, if I want to add a new curve in the near future, will I still need to modify protobufs (the `oneof`) to add the new case? IIRC that was the main advantage of introducing `pub_key_type` and removing `public_key`

melekes:
> @melekes So, if I want to add a new curve in the near future, will I still need to modify protobufs (the `oneof`) to add the new case? IIRC that was the main advantage of introducing `pub_key_type` and removing `public_key`

Yeah. We need to break the hash in order to remove it from `SimpleValidator`.

PR #3089, opened 2024-05-16
feat(abci): move `timeout_commit` into `FinalizeBlockResponse`

### LINKED ISSUE #2655: PBTS: Allow Application to Set BlockTimestamp

melekes:
as `next_block_delay`

ADR-115: https://github.com/cometbft/cometbft/pull/2966
Closes #2655 

---

#### PR checklist


docs/tutorials/go.md
```diff
@@ -438,7 +438,8 @@ func (app *KVStoreApplication) FinalizeBlock(_ context.Context, req *abcitypes.F
     }
 
     return &abcitypes.FinalizeBlockResponse{
-        TxResults:        txs,
+        TxResults:      txs,
```
sergio-mena, line 441: If you decided to add a sentence in previous file, please add it here as well

abci/types/application.go
```diff
@@ -115,5 +118,7 @@ func (BaseApplication) FinalizeBlock(_ context.Context, req *FinalizeBlockReques
 	}
 	return &FinalizeBlockResponse{
 		TxResults: txs,
+		// Safeguard devs from forgetting to set this field.
+		NextBlockDelay: 1 * time.Second,
```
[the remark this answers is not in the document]
melekes: I guess so. I will remove it.

proto/cometbft/abci/v1/types.proto
```diff
@@ -436,6 +437,12 @@ message FinalizeBlockResponse {
   // that execution of the transactions was deterministic.
   // It is up to the application to decide which algorithm to use.
   bytes app_hash = 5;
+  // delay between the time when this block is committed and the next height is started.
```
sergio-mena, line 440: @andynog and I had a discussion last night about how we are modifying protobufs on `main` and/or `v1.x`. We have added a point to discuss this in next week's meeting, as it is complex and we'd like everyone in the team to be aware of it, and provide feedback.
If you want to know what this is all about, take a look at #3090.
This shouldn't be blocking this PR, it's just to raise awareness.

proto/cometbft/abci/v1/types.proto
```diff
@@ -436,6 +437,12 @@ message FinalizeBlockResponse {
...
+  // previously `timeout_commit` in config.toml
+  google.protobuf.Duration next_block_delay = 6 [
+    (gogoproto.nullable) = false,
```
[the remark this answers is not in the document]
melekes: The current upgrade path should also work.

* if `0` and `timeout_commit=0s`, then skip timeout
* if `0` and `timeout_commit` is not zero, then use `timeout_commit`
* if `not-0`, then use it

spec/abci/abci++_methods.md
```diff
@@ -663,6 +664,18 @@ without calling `VerifyVoteExtension` to verify it.
       already passed on to the Application via `PrepareProposalRequest` or `ProcessProposalRequest`.
     * When calling `FinalizeBlock` with a block, the consensus algorithm run by CometBFT guarantees
       that at least one non-byzantine validator has run `ProcessProposal` on that block.
+    * `FinalizeBlockResponse.next_block_delay` - how long CometBFT waits after
+      committing a block, before starting on the new height (this gives the
+      proposer a chance to receive some more precommits, even though it
+      already has +2/3). Set to 0 if you want a proposer to make progress as
+      soon as it has all the precommits. Previously `timeout_commit` in
+      CometBFT config. **Set to constant 1s to preserve the old (v0.34 - v1.0) behavior**.
```
sergio-mena, line 672: > Set to constant 1s to preserve the old (v0.34 - v1.0) behavior

See my comments above about an operator -- oblivious to this -- that wants to upgrade, and has e.g. their `timeout_commit` set to 0 (or to 3).
Also see my comment above about the possibility of using `nil` in this field to allow an app to signal "I'm not aware of this, please fall back to `timeout_commit`"

spec/abci/abci++_methods.md
```diff
@@ -663,6 +664,18 @@ without calling `VerifyVoteExtension` to verify it.
...
+    * `FinalizeBlockResponse.next_block_delay` is a non-deterministic field.
+      This means that each node MAY provide a different value, which is
+      supposed to depend on how longs things are taking at the local node. It's
```
sergio-mena, line 675:
```suggestion
      supposed to depend on how long things are taking at the local node. It's
```

proto/cometbft/abci/v1/types.proto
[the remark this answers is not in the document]
sergio-mena: Yeah, I was thinking about this and I think you're right, no need to "leverage" `nil` here. This is my reasoning:

If the app (and app's community) is aware of this feature, they have to be aware of _the whole_ feature (including its upgrade path!).
So if you are setting `next_block_delay` to 0, you have to make sure `timeout_commit` is also 0. If you don't do that, you can't blame us for bad UX.
Do you agree? If yes, then, this thread can be dismissed.

spec/abci/abci++_methods.md
sergio-mena: Update: I'm now convinced we don't need to use `nil` here

sergio-mena (approved):
Awesome work! 🙏

proto/cometbft/abci/v1/types.proto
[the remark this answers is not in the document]
melekes: > Do you agree? If yes, then, this thread can be dismissed.

Exactly.

state/state.go
```diff
@@ -329,5 +337,8 @@ func MakeGenesisState(genDoc *types.GenesisDoc) (State, error) {
 		LastHeightConsensusParamsChanged: genDoc.InitialHeight,
 
 		AppHash: genDoc.AppHash,
+
+		// NextBlockDelay is set to 0 because the genesis block is committed.
+		NextBlockDelay: 0,
```
melekes, line 342: NOTE that previously, we've been waiting for 1s even for the genesis block, which doesn't make sense since validators don't vote for it.

melekes:
@cason please review once you're back 🙏

state/state.go
sergio-mena: Just has a quick chat. We agreed:

* This hunk is correct: it doesn't make sense to put something different from 0 here
* If the user starts the chain with a (stale) value of (now deprecated) `timeout_commit` in `config.toml` bigger than zero, if value will override the here for the first block of the chain. But that's up to the user to fix, and anyway not a bug deal.

I'm happy we now have a common understanding. I think we can resolve this comment thread

cason (commented):
Just triple-checking, from an orthogonal view looks good.

cason:
What I maybe miss, from a documentation point of view, is to make it clear that is the commit_time that is considered to "start" the configured delay. Not when `FinalizeBlock` is called. Although the time difference should not be that relevant.

sergio-mena:
@cason thanks for triple-checking. Regarding your last comment? If you create a PR with the improved doc, I can take a look and approve

VAIBHAVJINDAL3012:
Where do we set the next_block_delay? Is it done in the application and returned to cometBFT?

melekes:
> Where do we set the next_block_delay? Is it done in the application and returned to cometBFT?

Remember, it's not in `v1` nor `v0.38`. It's currently on `main`. From `v1.1` it will be in `FinalizeBlockResponse` https://docs.cometbft.com/v1.0/spec/abci/abci++_methods#finalizeblock

VAIBHAVJINDAL3012:
Thanks @melekes for your quick reply. I am trying to cherry pick these changes as I want const block time. Btw is it safe to do so as I can see there is no consensus breaking change. My question is who set the next_block_delay in FinalizeBlockResponse?

Issue #3091, opened 2024-05-17
Timeout ticker has incorrect behavior when two timeouts placed in quick succession


ValarDragon:
### Summary

I found that we can get negative timeout duration at least in tests, and subsequently incorrect behavior while working on: #3008 . I have made a PR that can reproduce this failing test (at least locally)

Suppose we have code like (We have effectively this in create empty blocks tests)
```
                startTime := time.Now()
		negTimeout := timeoutInfo{Duration: -1 * time.Millisecond, Height: height, Round: 0, Step: types.RoundStepNewHeight}
		timeout := timeoutInfo{Duration: 5 * time.Millisecond, Height: height, Round: 0, Step: types.RoundStepNewRound}
		ticker.ScheduleTimeout(negTimeout)
		ticker.ScheduleTimeout(timeout)

		// Wait for the timeout to be received
		to := <-c
                elapsedTime := time.Now().Sub(startTime)
```

The intended behavior is oneof:
- to := negTimeout, elapsedTime ~= 0
- to := timeout, elapsedTime ~= 5ms

Unfortunately we also get the behavior
- to := timeout, elapsedTime ~= 0

Which is our bug. This should be causing test flakyness. unfortunately I don't know enough about our timeout usages to know if this poses any safety problems.

ValarDragon:
After understanding the problem more, this will happen _whenever_ we have two timeouts scheduled near each other in time. It ~appears from spot checking, near each other in time translates to (within an order of magnitude) 20000 ns.

ValarDragon:
After understanding the race condition more, this will occur whenever we schedule two timeouts at ~the same time. We can get timeout 1 timer delay, but with timeout 2's value.

PR #3092, opened 2024-05-17
fix: TimeoutTicker returns wrong value/timeout pair when timeouts are scheduled at ~approximately the same time

ValarDragon:
#3091 

The problem is we have an edge case where we should drain the timer channel, but we "let it slide" in certain race conditions when two timeouts are scheduled near each other. This means we can have unsafe timeout behavior as demonstrated in the github issue, and likely more spots in consensus.

Notice that aside from NewTimer and OnStop, all timer accesses are from the same thread. In NewTimer we can block until the timer is drained (very quickly up to goroutine scheduling). In OnStop we don't need to guarantee draining before the method ends, we can just launch something into the channel that will kill it.

In the main timer goroutine, we can safely maintain this "timerActive" variable, and force drain when its active. This removes the edge case.

The test I created does fail on main.

---

#### PR checklist

- [X] Tests written/updated


internal/consensus/ticker.go
```diff
@@ -29,17 +29,23 @@ type TimeoutTicker interface {
...
-		timer:    time.NewTimer(0),
-		tickChan: make(chan timeoutInfo, tickTockBufferSize),
-		tockChan: make(chan timeoutInfo, tickTockBufferSize),
+		timer: time.NewTimer(0),
+		// An indicator variable to check if the timer is active or not.
+		// Concurrency safe because the timer is only accessed by a single goroutine.
+		timerActive: true,
+		tickChan:    make(chan timeoutInfo, tickTockBufferSize),
+		tockChan:    make(chan timeoutInfo, tickTockBufferSize),
+		stopChan:    make(chan struct{}),
```
melekes, line 48: don't think we need it. after `OnStop` is called, `t.Quit()` is closed.

ValarDragon: Thank you! Wasn't familiar with thats how stop works, I'll remove that

melekes (approved):
Thanks @ValarDragon ❤️

internal/consensus/ticker.go
```diff
@@ -29,17 +29,21 @@ type TimeoutTicker interface {
...
-		tockChan: make(chan timeoutInfo, tickTockBufferSize),
+		timer: time.NewTimer(0),
+		// An indicator variable to check if the timer is active or not.
+		// Concurrency safe because the timer is only accessed by a single goroutine.
+		timerActive: true,
+		tickChan:    make(chan timeoutInfo, tickTockBufferSize),
+		tockChan:    make(chan timeoutInfo, tickTockBufferSize),
 	}
 	tt.BaseService = *service.NewBaseService(nil, "TimeoutTicker", tt)
 	tt.stopTimer() // don't want to fire until the first scheduled timeout
```
hvanz, line 49: Since timerActive is true, stopTimer will return immediately, so this line could be removed?

hvanz (approved):
LGTM

internal/consensus/ticker_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+		// Wait for the timeout to be received
+		to := <-c
+		endTime := time.Now()
+		elapsedTime := endTime.Sub(startTime)
+		if timeout == to {
+			require.True(t, elapsedTime >= timeout.Duration, "We got the 5ms timeout. However the timeout happened too quickly. Should be >= 5ms. Got %dms (start time %d end time %d)", elapsedTime.Milliseconds(), startTime.UnixMilli(), endTime.UnixMilli())
+		} else {
+			require.True(t, elapsedTime <= timeout.Duration, "We got the -1ms timeout. However the timeout happened too slowly. Should be ~near instant. Got %dms (start time %d end time %d)", elapsedTime.Milliseconds(), startTime.UnixMilli(), endTime.UnixMilli())
+		}
```
sergio-mena, line 40: Aren't the conditions reversed? E.g. `elapsedTime >= timeout.Duration` means "more time has elapsed than expected" so it happened too slowly. Same for the _else_.

internal/consensus/ticker_test.go
```diff
@@ -0,0 +1,42 @@
...
+		ticker.ScheduleTimeout(timeout)
+
+		// Wait for the timeout to be received
+		to := <-c
+		endTime := time.Now()
+		elapsedTime := endTime.Sub(startTime)
+		if timeout == to {
+			require.True(t, elapsedTime >= timeout.Duration, "We got the 5ms timeout. However the timeout happened too quickly. Should be >= 5ms. Got %dms (start time %d end time %d)", elapsedTime.Milliseconds(), startTime.UnixMilli(), endTime.UnixMilli())
+		} else {
+			require.True(t, elapsedTime <= timeout.Duration, "We got the -1ms timeout. However the timeout happened too slowly. Should be ~near instant. Got %dms (start time %d end time %d)", elapsedTime.Milliseconds(), startTime.UnixMilli(), endTime.UnixMilli())
```
sergio-mena, line 39:
```suggestion
			require.True(t, elapsedTime <= to.Duration, "We got the -1ms timeout. However the timeout happened too slowly. Should be ~near instant. Got %dms (start time %d end time %d)", elapsedTime.Milliseconds(), startTime.UnixMilli(), endTime.UnixMilli())
```

sergio-mena (approved):
Excellent! Thanks @ValarDragon 🙏

ValarDragon:
@sergio-mena re your comment yes, but the timeout here used is 0s, which is supposed to execute ~near instantly (whereas its budgeted 5ms). I believe this has different scheduler logic than things budgeted further into the future. 

Will remove though, since as you note its covered by the then branch regardless!

cason:
Hey, just to mention that timeout scheduling/events are totally ordered. That is why when we receive a command to schedule timeout for a height (H), round (S), and step (RS), we compare it to the currently scheduled. If it is "larger", we replace the old timeout by the new one. This ad-hoc solution prevents a lot of timeouts to expire when they are not needed anymore.

In your test case you use two timeouts that are equal. This implementation does not support this, because of the total order assumption, that is why the test fails.

Issue #2873, opened 2024-04-23
`timeout` params [Tracking issue]


melekes:
### Summary

```
timeout_propose = "3s"
timeout_propose_delta = "500ms"
timeout_prevote = "1s"
timeout_prevote_delta = "500ms"
timeout_precommit = "1s"
timeout_precommit_delta = "500ms"
timeout_commit = "1s"

skip_timeout_commit = false
```

When speaking about `timeout` parameters, there are multiple concerns.

1. We don't need that many parameters
2. App developers want consistency across validators, even though params may vary depending on validator's network topology and runtime
3. App developers want a way to set block intervals (think of it as if all validators set the same `timeout_commit` value).

Note there are other potential ways to align parameters in a network https://github.com/cometbft/cometbft/issues/1617.

```[tasklist]
### Tasks
- [ ] Remove `skip_timeout_commit` in favor of setting `timeout_commit` to `0`
- [ ] Remove `timeout_prevote` `timeout_precommit` in favor of calculating them https://github.com/tendermint/tendermint/issues/2920#issuecomment-445874592
- [ ] Merge `timeout_propose_delta`, `timeout_prevote_delta`, `timeout_precommit_delta` into a single `round_delta`
- [ ] Make `timeout_commit` a consensus parameter
```

So (1) will be solved by 1,2,3. (3) - by 4. (2) - by social consensus & external tools https://github.com/cometbft/cometbft/issues/1617

To protect against greedy proposers, we will likely need to punish validators who propose a block those timestamp is less than `LastBlock.Time + TimeoutCommit + SomeDelta`.

sergio-mena:
These are my thoughts:
* Task 1. Fully agree
* Task 2. Makes sense in principle, although the relations pointed out in [this comment](https://github.com/tendermint/tendermint/issues/2920#issuecomment-445874592) are _inequations_, so they don't really tell you the value of some params in terms of others, but rather bounds (i.e., min or max values). I recently heard @cason say something along these lines.
  * Something I do see clearly here is that I don't see any practical reason not to merge `timeout_prevote` and `timeout_precommit` into one single value. As the event these timeouts relate to are vote reception (same number, similar size)
* Task 3. Fully agree with `timeout_prevote_delta` and `timeout_precommit_delta` (both relating to votes). Not 100% sure about `timeout_propose_delta`, as it has to do with a proposal and its block, with can be way bigger in size than votes.
* Task 4. Fully agree

melekes:
[...]

PR #3093, opened 2024-05-17
feat(config): deprecate skip_timeout_commit

melekes:
instead of removing it
Refs #2926

https://github.com/cometbft/cometbft/issues/2873#issuecomment-2117172714

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/docs/references/config/config.toml.md
+++ b/docs/references/config/config.toml.md
@@ -1720,6 +1720,21 @@ But it may have implications in the way the application rewards validators.
 Setting `timeout_commit` to `0s` means that the node will start the next height
 as soon as it gathers all the mandatory +2/3 precommits for a block.
 
+### consensus.skip_timeout_commit
+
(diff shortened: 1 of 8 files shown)


docs/references/config/config.toml.md
```diff
@@ -1720,6 +1720,21 @@ But it may have implications in the way the application rewards validators.
...
+```toml
+skip_timeout_commit = false
+```
+
+| Value type          | boolean |
+|:--------------------|:--------|
+| **Possible values** | `false` |
+|                     | `true`  |
+
+Setting `skip_timeout_commit` to `true` has the same effect as setting [`timeout_commit`](#consensustimeout_commit) to `"0s"`.
```
sergio-mena, line 1736: Can we add a sentence (or reword the current one) saying something like "do no use, it's deprecated and may be removed in future releases"

sergio-mena (approved):
Wow, that was fast! 🚀

PR #3104, opened 2024-05-22
refactor: unlock logic

lilasxie:
```golang
{
        // limit the scope of the lock
	ss.handlerMtx.Lock()
	defer ss.handlerMtx.Unlock()
	res, err = ss.validationRequestHandler(ss.privVal, req, ss.chainID)
	ss.handlerMtx.Unlock()
}
```
If I understood the comment `limit the scope of the lock` correctly: wrapping this logic in the block aims to `unlock` early.

`defer` won't execute after the block end. We should wrap it in the closure `func(){}()`, or just unlock directly after `ss.validationRequestHandler`

---

#### PR checklist

- [x] Tests written/updated


hvanz (approved):
lgtm

melekes (approved):
Thanks @lilasxie ❤️

privval/signer_server.go
```diff
@@ -72,16 +72,12 @@ func (ss *SignerServer) servicePendingRequest() {
...
-		ss.handlerMtx.Lock()
-		defer ss.handlerMtx.Unlock()
-		res, err = ss.validationRequestHandler(ss.privVal, req, ss.chainID)
-		if err != nil {
-			// only log the error; we'll reply with an error in res
-			ss.Logger.Error("SignerServer: handleMessage", "err", err)
-		}
+	ss.handlerMtx.Lock()
+	res, err := ss.validationRequestHandler(ss.privVal, req, ss.chainID)
+	ss.handlerMtx.Unlock()
```
sergio-mena, line 77: `validationRequestHandler` is a functor whose default value is `DefaultValidationRequestHandler`

I can see that `DefaultValidationRequestHandler` can panic. It is not go-idiomatic to leave a mutex locked in case of panic (the panic might be "recovered" up in the call stack).

This should rather read:
```go
func () {
	ss.handlerMtx.Lock()
	defer ss.handlerMtx.Unlock()
	res, err := ss.validationRequestHandler(ss.privVal, req, ss.chainID)
}()
```
Check [this](https://go.dev/play/p/OrlUdJ5ZZ05) for an example.

lilasxie: Thanks for the knowledge, sir. I have opened a new PR #3135 for this😀

sergio-mena: Thanks a lot! Taking a look

adizere:
This should be backported to v1.x

PR #3109, opened 2024-05-23
feat: add api to allow get unconfirmed_tx by hash

mmsqe:
---

#### PR checklist

- [ ] Tests written/updated


mmsqe:
Hi @andynog, I use this api for cross-checking, can we link to https://github.com/cometbft/cometbft/pull/3008? Since adding hashes seems conflict with limit param, I added a new api instead. We could also support multi hashes like `/unconfirmed_tx?hash[0]=0xA&hash[1]=0xB`

mempool/clist_mempool.go
```diff
@@ -539,6 +539,20 @@ func (mem *CListMempool) ReapMaxTxs(max int) types.Txs {
 	return txs
 }
 
+// FilterTx returns the types.Tx with the given hash if found in the mempool, otherwise returns nil.
+func (mem *CListMempool) FilterTx(hash []byte) types.Tx {
+	mem.updateMtx.RLock()
+	defer mem.updateMtx.RUnlock()
+
+	for e := mem.txs.Front(); e != nil; e = e.Next() {
```
hvanz, line 547: There is no need to traverse the whole clist to find the transaction. (This is locking the mempool and it could be a DoS attack vector). The function `InMempool` should be used instead.

melekes (approved):
Thanks @mmsqe ❤️

mempool/mempool.go
```diff
@@ -44,6 +44,10 @@ type Mempool interface {
 	// (~ all available transactions).
 	ReapMaxTxs(max int) types.Txs
 
+	// FilterTx returns the types.Tx with the given hash if found in the mempool,
+	// otherwise returns nil.
+	FilterTx(hash []byte) types.Tx
```
hvanz, line 49: This is just a personal preference but I would change the name of the function to something more specific like `GetTxByHash`.

hvanz (approved):
Thank you @mmsqe!

Issue #3108, opened 2024-05-22
Add callback for publishing pending transactions from mempool


mmsqe:
### Summary

Add [PublishEventPendingTx](https://github.com/mmsqe/cometbft/blob/4d00e28bcfec05a0405489cdf7ff91883c744bbf/types/event_bus.go#L171) in EventBus

### Problem Definition

Currently we rely on [EventTx](https://github.com/mmsqe/cometbft/blob/4d00e28bcfec05a0405489cdf7ff91883c744bbf/types/events.go#L23) for [newPendingTransactions](https://geth.ethereum.org/docs/interacting-with-geth/rpc/pubsub), but transactions are published after they are mined in a block. There is no way to subscribe to pending transactions.

### Proposal

1. add callback for new transaction that added to the mempool
2. publish pending tx event for each newly inserted transaction

andynog:
Hi @mmsqe, usually the events are triggered from the state package, after a block has been committed. Emitting events for each transaction added to the mempool might be a new use case. I see that in your fork you seem to have implemented that though. 

Using the `/unconfirmed_txs` endpoint not an alternative ?  I think it's more a pull that push behavior but just trying to understand better the use cases for implementing a new model.

mmsqe:
Thanks @andynog, you mean long poll that block until new data available and return the response, which works similar as real-time push for web socket subscribe right? But `/unconfirmed_txs` is just normal API which won't block, client need keep pulling.

andynog:
yes, the mempool txs are not correlated with new block events, so you would need to keep polling for data on that endpoint.

hvanz:
The mempool does not use the event bus and we actually don't want to add this dependency. It goes against our effort of making the mempool and other components more modular. It would be better to understand the use case for issue so we can find an alternative implementation.

cason:
The transactions that are added to the mempool are transactions that have been accepted by the local instance of the application using the ABCI `CheckTx` method. This is the best point, in the current architecture, to identify when new transactions are available, and multiple applications use this information for their own goals (e.g., derive some information, produce some order, compute something from all transactions accepted). Adding a listener there for new transactions is a way to go.

[...]

Issue #2803, opened 2024-04-12
Mempool Lanes: introduce QoS to the mempool [tracking issue]


sergio-mena:
### Summary

In order to enable better management of transaction traffic at P2P level, we need to allow the app to classify transactions. For that, we introduce the concept of Quality of Service (QoS), well known in IP networks, which we realize as **mempool lanes**. [ADR-118](https://github.com/cometbft/cometbft/blob/main/docs/references/architecture/adr-118-mempool-lanes.md) describes this feature more in detail.

This issue will track:
* the discussion with our users
* design work
* implementation work
* any follow up

### Teams interested

1. Osmosis
2. Injective
3. dYdX
4. Astria

### Execution Steps

As with previous sizeable features, the design and implementation work will be carried out in a feature branch -- `feature/mempool-qos`, bifurcating from `main` -- to avoid the risk of last-minute serious problems. A feature branch enables the (management) decision to defer shipping a feature in a new release if it is considered not ready.

If we need our users to validate our design/implementation while this feature is in progress, we will probably need to maintain some temporary branch off a released version (likely `v0.38.x`) that reflects our work in `feature/mempool-qos`. We will work out the details as we start creating PRs with code.

#### Initial steps

- [x] Create branch `feature/mempool-qos`
- [x] Initial ADR skeleton with structure and discussions
  - #2868

#### Other related pending improvements on mempool

- [x] #2317
- [x] #895
- [ ] ~~Remove `TxsAvailable` and `CreateEmptyBlocks` (laundry list)~~ (not actually needed)

#### Design: core work

These tasks are mostly sequential, although some of them can be done in parallel

- [x] #2934
- [x] #3027
- [x] #3110
  - [2 days left]
- [x] #3088
- [x] ADR: Iterate on the design with output from users
- [x] #3138
  - [2 days left]
- [x] Decide where the design/spec will land [1 day]
  - Shall we keep the design/spec in the ADR? Or do we need to create a section in the spec?
  - We decided to keep it all in the ADR because it makes sense to understand the whole text. Later we will copy the Properties section to its own file in the spec folder.
- [x] Detailed implementation plan: flesh out tasks in section below
- [x] #3499
- [x] #4120
- [X] ~~Finalize spec/design document (ADR or move elsewhere) [7 days]~~
  - We decided to leave the current mini spec as part of the ADR.

Initial estimate (this section): 12 days

#### Implementation

- [x] #3084
- [x] #3303
- [x] #3296
[...]

PR #3111, opened 2024-05-23
feat: add callback for new transaction that added to the mempool 

mmsqe:
Closes: https://github.com/cometbft/cometbft/issues/3108

---

#### PR checklist

- [ ] Tests written/updated


hvanz (changes requested):
Currently the mempool does not have the event bus as a dependency. Adding the event bus goes against our efforts to make the system more modular. 

The `TxsAvailable` channel could be modified to notify which new tx is available in the mempool, instead of just sending `struct{}{}`. But it's in our plans to remove it (https://github.com/cometbft/cometbft/issues/2803).

As commented in the issue, it would be better to understand the use case.

mmsqe:
Actually eventbus is added in node, mempool doesn't need to know eventbus and no extra dependency. And `mempl.WithNewTxCallback` is optional, can add if need, it's still decoupled. Seems also fine without`TxsAvailable` which only notify once per height.

types/event_bus.go
```diff
@@ -169,6 +169,15 @@ func (b *EventBus) PublishEventValidBlock(data EventDataRoundState) error {
 	return b.Publish(EventValidBlock, data)
 }
 
+func (b *EventBus) PublishEventPendingTx(data EventDataPendingTx) error {
+	// no explicit deadline for publishing events
```
melekes, line 173: I don't like that this is blocking. Note that `notifyTxsAvailable` above is non-blocking. The consequence of potentially waiting for `eventBus` is slower transaction processing.

melekes (commented):
Nice work @mmsqe 👍 However, I have some concerns related to the newly introduced callback.

mempool/clist_mempool.go
```diff
@@ -434,6 +441,9 @@ func (mem *CListMempool) resCbFirstTime(tx types.Tx, res *abci.CheckTxResponse)
 		tx:        tx,
 	}) {
 		mem.notifyTxsAvailable()
+		if mem.onNewTx != nil {
+			go mem.onNewTx(tx)
```
melekes, line 445:
```suggestion
			go mem.onNewTx(tx) // non-blocking to not slow down txs processing
```

melekes (approved):

mempool/clist_mempool.go
```diff
@@ -434,6 +441,9 @@ func (mem *CListMempool) resCbFirstTime(tx types.Tx, res *abci.CheckTxResponse)
 		tx:        tx,
 	}) {
 		mem.notifyTxsAvailable()
+		if mem.onNewTx != nil {
+			go mem.onNewTx(tx) // non-blocking to not slow down txs processing
```
sergio-mena, line 445: There is no backpressure mechanism here.
In our latest [QA tests](https://github.com/cometbft/cometbft/blob/v1.x/docs/references/qa/CometBFT-QA-v1.md#blocks-produced-per-minute-and-transactions-processed-per-minute) in version `v1.x`, with 200 nodes, 175 of them being validators, CometBFT is able to sustain a workload of around 25000 (1k-long) transactions per minute.
We are concerned that allowing the number or goroutines to grow here with no further mechanism to limit goroutine creation might:

* negatively affect the results we get on [memory used](https://github.com/cometbft/cometbft/blob/v1.x/docs/references/qa/CometBFT-QA-v1.md#memory-resident-set-size) by CometBFT in future versions
* or even worse, that the CometBFT node might run out of memory and crash

mmsqe: You're right, do you have any suggestion?

hvanz (commented):
I'm unblocking the PR as I see that the callback does not really introduce a dependency. Still, as @sergio-mena pointed above, I'm a bit concerned about performance under high tx load, on the number of go routines and on the number of mempool events that may interfere with consensus.

melekes (changes requested):
Thanks for the PR @mmsqe ❤️ I'm going to request changes so this PR doesn't get merged. We do want it to get merged eventually, but I'm concerned about performance and lock contention. It'll take some extra work hours to address these concerns in a proper way.

We need to limit the queue for this type of messages, so we don't block consensus and other more important messages PLUS ensure the lock contention doesn't negatively affect mempool performance.

mmsqe:
@melekes Thanks for the detail, but this PR is unnecessary when we handle in sdk level. It's more flexible to add eth tx hash in [decorator](https://github.com/crypto-org-chain/ethermint/pull/491/files#diff-f0f6e8137667cb37566fc8981a6a06576dadd945aac2cca28aa9efa459f693abR28) without extra decode in [listener](https://github.com/crypto-org-chain/ethermint/pull/491/files#diff-76d70485a06c8576e4b02647c361257651640fc15a380db6644b674bdfcf63b7R116). The current stream would drop the oldest segment of messages if the maximum [capacity](https://github.com/crypto-org-chain/ethermint/blob/bb1441462e3c784f55902950ce28022a16c233bd/rpc/stream/stream.go#L48) is reached.

.changelog/unreleased/features/3108-pending-tx-event.md
cason, line 1: The changelog, as well as the PR description, should be more precision. The feature added is an optional event produced whenever a new tx is added to the mempool. The callback approach is how it is implemented, not what is implemented.

config/config.go
```diff
@@ -914,6 +914,9 @@ type MempoolConfig struct {
 	// performance results using the default P2P configuration.
 	ExperimentalMaxGossipConnectionsToPersistentPeers    int `mapstructure:"experimental_max_gossip_connections_to_persistent_peers"`
 	ExperimentalMaxGossipConnectionsToNonPersistentPeers int `mapstructure:"experimental_max_gossip_connections_to_non_persistent_peers"`
+
+	// PublishEventPendingTx indicates whether publishing a new transaction is added to the mempool.
+	PublishEventPendingTx bool `mapstructure:"publish-event-pending-tx"`
```
cason, line 919: We could come with a better name here.

types/event_bus.go
```diff
@@ -169,6 +169,19 @@ func (b *EventBus) PublishEventValidBlock(data EventDataRoundState) error {
 	return b.Publish(EventValidBlock, data)
 }
 
+func (b *EventBus) PublishEventPendingTx(data EventDataPendingTx) error {
```
cason, line 172: I would expect this method to be very similar to `PublishEventTx`.

node/setup.go
```diff
@@ -251,13 +253,38 @@ func createMempoolAndMempoolReactor(
 	// allow empty string for backward compatibility
 	case cfg.MempoolTypeFlood, "":
 		logger = logger.With("module", "mempool")
+		options := []mempl.CListMempoolOption{
+			mempl.WithMetrics(memplMetrics),
+			mempl.WithPreCheck(sm.TxPreCheck(state)),
+			mempl.WithPostCheck(sm.TxPostCheck(state)),
+		}
+		if config.Mempool.PublishEventPendingTx {
```
cason, line 261: All this logic is intended to bash transactions and produce events from batches of transactions?

If you check how the `PublishEventTx` is used: 

https://github.com/cometbft/cometbft/blob/2dd5209177270fde125c71aeedda9819046f92f0/state/execution.go#L749-L758

you will see that no batching is performed, as this is the role of the event bus by itself. It expect to take multiple events, put them in a queue and make sure that the queue is consumed fast enough, otherwise events are dropped.

cason (changes requested):
Conceptually ok, some discussion is worth regarding naming, etc.

However, the implementation should, except if really explained why not, mimic the implementation of the `Tx` event, already present in the code base.

node/setup.go
mmsqe: It is also used to prevent potential blocking of event publish, which can affect the performance of mempool.

cason: Why?

`PublishWithEvents` adds the event to the `s.cmds` channel of the pubsub server. This server contains an infinite `loop` that for publish events calls a `send` method with:

https://github.com/cometbft/cometbft/blob/2dd5209177270fde125c71aeedda9819046f92f0/libs/pubsub/pubsub.go#L420-L432

So, essentially, if the subscription has a buffer, there is no blocking here. Otherwise, any slow client subscribing to the `Tx` event would crash the node.

cason: Again, when a block is committed, all txs in the block are published using the `Tx` event. This is more publishing calls than any solution inside the mempool. And this does not break, provided it is properly used.

If you say that this method is not efficient and can be improved, I will very likely agree. But this PR is not about improving the pubsub service, is about creating a new event to be published using the existing service.

mmsqe: That's the [origin](https://github.com/cometbft/cometbft/blob/fba9b434cdd8b0a355291b2d10086fe3e6c5037b/node/setup.go#L263) way but there's a [concern](https://github.com/cometbft/cometbft/pull/3111#discussion_r1615819167) about blocking behaviour.

cason: Yes, I saw @sergio-mena comment. But I still wouldn't solve the problem in this way, namely, by adding a sub-routine inside the `node` package to batch notifications about txs added to the mempool.

mmsqe: Any suggestion is welcome.

cason: I gave my suggestion: let `pubsub` handle this and set a decent buffer length for this subscription. Then warns users that with this option they can observe performance issues. We attempt to measure the impact in performance using some test/benchmark units.

@sergio-mena does not agree with me, so I will probably discuss this with him synchronously.

cason:
Hey @mmsqe, we've discussed how to proceed with this PR and our decision was the following:

1. We are accepting this feature as `experimental`, since we are not able to perform comprehensive QA/performance evaluation in order to assess its impact on performance. Experimental features are disabled by default (ok), the config flag starts with `experimental_` and the documentation in the file lets it clear that it should be used with caution.
2. When providing an experimental feature, we do not want to make important changes in core code, in this case there are relevant changes in the eventbus/pubsub logic, that we should not include in this PR. Like I mentioned in previous comments, we should rely on the current logic to buffer/flow control the production of events, as we already do with other subscriptions. However, if we get strong evidence of important performance impact, we should consider optimizing the event bus behavior in a **different** PR, provided that we have measured and attested the negative impact.

cason:
@melekes is blocking this PR from being approved, he is off this week.

melekes (approved):
Thanks @mmsqe ❤️

melekes:
> @melekes is blocking this PR from being approved, he is off this week.

no

PR #3112, opened 2024-05-23
perf: Make mempool recheck state only use an atomic for num rechecked txs

ValarDragon:
This PR makes the mempool recheck only use an atomic instruction for the number of succesfully rechecked txs, not the "unbalanced" number of txs rechecked.

There are two reasons for this:
- As part of #2925 , I want to make an alternate information communication channel between Reap and Update. Recheck will maintain variables for rechecked # txs, # gas, # bytes. If Reap is called, and recheck is active, it will setup a channel fpr 
- This is technically a minor speedup, by making the initiating thread not use atomics, only the callbacks.

---

#### PR checklist

- [x] Tests written/updated
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog) - omitted since its new to this version / very minor, and about to be edited more in a subsequent PR

### DIFF
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -618,7 +618,7 @@ func (mem *CListMempool) recheckTxs() {
 	// because this function has the lock (via Update and Lock).
 	for e := mem.txs.Front(); e != nil; e = e.Next() {
 		tx := e.Value.(*mempoolTx).tx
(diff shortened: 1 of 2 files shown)


ValarDragon:
Closing this. not obvious this is going to be the end state, let me get the full thing working first.

PR #3115, opened 2024-05-24
perf: Make mempool reap be able to occur once Update has succesfully updated enough txs

### LINKED ISSUE #2925: Mempool Rechecking all txs blocks consensus

ValarDragon:
If this PR is the right design/  taken to completion, will close #2925 

This PR is trying to handle communication between rechecking and Updating processes, and allow recheck to begin concurrently once enough txs have been succesfully updated.

The way it works:
- Adapts the recheck struct @hvanz recently introduced, to also have a new field for "shared state between recheck <> update", that has a mutex.
- We only allow one reap at a time, while update is going on.
- When we need to reap, we update this shared state to say:
  - "We are waiting for reap"
  - "Use this function to determine if we are ready for reap"
  - "communicate reap readyness on this channel"
  - (And we check if were already ready to reap)
- When we process the callback for each rechecked tx, we update some stats.
  - "if we are waiting for reap", then check if we have updated enough txs to begin reaping. If so signal the reap channel and mark ourself as no longer waiting for reap
  - (Potentially a micro-optimization) if the mempool is really small, don't bother with updating any stats. Just reap once update ends. 

Aside from the one thing I noted about "compute tx len", I don't think this should be a real performance drawdown. We already have many mutex lock/unlocks here 

I'm currently looking for feedback on is this communication structure between Reap and Update, please let me know! Then I will proceed with tests!

---

#### PR checklist

- [ ] Tests written/updated


ValarDragon:
My last commit protects against an edge case for triggering Update, while Reap is still in progress. This seems impossible to actually construct, and even if we were in such a situation, does not seem like it would be a problem.

Why its impossible to construct:
```
// We are still reaping, while we are committing the next block.
// However reaping only occurs during propose, which is under the consensus mutex.
// Update only happens during Commit which is also under the consensus mutex.
// Furthermore Reap is very quick once unlocked.
// If we are in this edge case that seems impossible, we just retry in a loop until recheck is complete.
```

Even if we were in such a situation it should be fine:
* Reap proceeds faster than Update
* Reap is Read-only
* Reap has a head start on Update

Thoughts on whether we should delete it?

mempool/clist_mempool.go
```diff
@@ -524,12 +534,18 @@ func (mem *CListMempool) ReapMaxBytesMaxGas(maxBytes, maxGas int64) types.Txs {
...
 func (mem *CListMempool) ReapMaxTxs(max int) types.Txs {
-	mem.updateMtx.RLock()
-	defer mem.updateMtx.RUnlock()
-
+	memTxsLen := mem.txs.Len() // concurrency safe
 	if max < 0 {
-		max = mem.txs.Len()
+		max = memTxsLen
 	}
+	canExecuteConcurrentlyWithRecheck := max < memTxsLen
```
melekes, line 541: if `max = memTxsLen` on line 539, then this condition will be false, correct?

mempool/clist_mempool.go
```diff
@@ -664,10 +712,135 @@ func (mem *CListMempool) recheckTxs() {
...
+	doneCh chan struct{}   // to signal that rechecking has finished successfully (for async app connections)
+
+	// This is only read in the recheck threads
+	initialMempoolSize int32        // initial size of the mempool before rechecking started
+	numRecheckedTxs    atomic.Int32 // number of transactions still pending to recheck
+
+	recheckReapSharedState recheckReapSharedState
+}
+
+var minimumMempoolSizeForConcurrentRecheck int32 = 50
```
melekes, line 727:
```suggestion
const minimumMempoolSizeForConcurrentRecheck int32 = 50
```

mempool/clist_mempool.go
```diff
@@ -664,10 +712,135 @@ func (mem *CListMempool) recheckTxs() {
...
+// If we are in this edge case that seems impossible, we just retry in a loop until recheck is complete.
+// Unclear if this method is needed in the first place.
+func (r *recheckReapSharedState) waitForReapingToFinish() {
+	r.mtx.Lock()
+	isReaping := r.isReaping
+	r.mtx.Unlock()
+	if !isReaping {
+		return
+	}
+	r.mtx.Unlock()
```
melekes, line 780: ?

mempool/clist_mempool.go
```diff
@@ -664,10 +712,135 @@ func (mem *CListMempool) recheckTxs() {
...
+	for {
+		time.Sleep(5 * time.Millisecond)
```
melekes, line 782: let's extract this into a const

mempool/clist_mempool.go
```diff
@@ -676,23 +849,28 @@ func newRecheck() *recheck {
...
+func (rc *recheck) init(first, last *clist.CElement, mempoolLen int32) {
+	// set to active, and check if
+	isAlreadyActive := !rc.active.CompareAndSwap(false, true)
+	if isAlreadyActive {
 		panic("Having more than one rechecking process at a time is not possible.")
 	}
 	rc.cursor = first
 	rc.end = last
-	rc.numPendingTxs.Store(0)
+	rc.initialMempoolSize = (mempoolLen)
```
melekes, line 860:
```suggestion
	rc.initialMempoolSize = mempoolLen
```

melekes (commented):
Thanks @ValarDragon ❤️

mempool/clist_mempool.go
ValarDragon: Woops, thank you!

mempool/clist_mempool.go
ValarDragon: Yes! (That is the intention, as we would have to wait until recheck is done by definition)

melekes (approved):

ValarDragon:
The goal of this change is that when we have a large mempool, we do not block the block proposer on rechecking the entire mempool, instead just a sufficient number of txs. This reduces the issues we will get on rechecking a large mempool, as it won't delay the next block of consensus. This means that a recheck times would have to be a full block time before having any impact on the system. (As then it would block the next block;s Update)

Right now (prior to this PR) any time spent in Update, would block reap and delay the proposers block proposal.

melekes:
We would like to have a synchronous review session for this PR.

mempool/clist_mempool.go
```diff
@@ -664,10 +712,136 @@ func (mem *CListMempool) recheckTxs() {
...
+type recheckReapSharedState struct {
+	mtx                    sync.Mutex
+	successfullyUpdatedTxs int64
+	bytesUpdated           int64
+	gasUpdated             int64
+
+	// indicator variable needed to ensure we don't start a new recheck process
+	// while reaping is in progress. Since reap's only have read-only access,
+	// we could relax this to an "active reaps" counter if we wanted.
+	isReaping bool
```
hvanz, line 738: Change type to `atomic.Bool`?

cason:
A question that came to mind while trying to understand the implementation: why not to use a semaphore?

We want to reap X bytes and Y gas. So we try to acquire one of both permits.

The re-check, once the response for a tx is received, releases tx size to the first semaphore and gas to the second.

Does it solve the problem?

mempool/clist_mempool.go
```diff
@@ -664,10 +712,136 @@ func (mem *CListMempool) recheckTxs() {
...
+	r.reapSignal <- struct{}{}
+	r.reapSignal = nil
+	r.isReaping = true
+}
+
+func (rc *recheck) updateForRecheckSuccess(tx types.Tx, res *abci.CheckTxResponse) {
+	if rc.initialMempoolSize > minimumMempoolSizeForConcurrentRecheck {
+		rc.recheckReapSharedState.mtx.Lock()
+		rc.recheckReapSharedState.successfullyUpdatedTxs++
+		// TODO: compute proto size for txs may be slow.
```
cason, line 836: This is super inefficient and of course the tx size should be part of the tx entry on the mempool, computed once it is accepted by the application at first.

But this is a problem for another issue/PR.

cason:
It is important to recall here that this change would only have effect if the `ReCheck` call is _indeed_ asynchronous.

ValarDragon:
Per call, we agreed to put focus elsewhere! Thanks for the reviews and great convos! This may become relevant after some other improvement ideas

melekes:
@ValarDragon: "we shouldn't bet on the app keeping recheck times minimal, so this still matters. (Block SDK accidentally re-introduced a ~4x overhead to our recheck time)"

ValarDragon:
After increased learnings/thinkings, we think that https://github.com/cometbft/cometbft/pull/3314 is a better approach now!

Issue #3110, opened 2024-05-23
ADR: Due-diligence on the three items in "Alternative Approaches" section


sergio-mena:
Alternative approaches to look into:

* [x] Priority mempool (as of `v0.37.x`): why is it not a good fit?
  * #3116
* [x] Solana's Gulf Stream
  * #3139
* [x] Skip's Block SDK
  * #3141
* Other alternatives:
  * [x] Ethereum pre-confirmations
    * #4093
* Also check [this issue](https://github.com/cosmos/ibc-go/issues/6159)

These tasks are to be time-boxed: 1-2 days each

sergio-mena:
@adizere still reading on ethereum preconfirmations. Will close this when I'm back from vacation. This shouldn't block any other work on Mempool QoS

PR #3116, opened 2024-05-24
docs(mempool): added network consistency properties and Priority Mempool

sergio-mena:
Partially addresses #3110

Added subsection "Priority mempool" in "Alternative approaces" section.

I also added extra properties in the previous section: those properties are one of the core points in our current discussions with our users, and I wanted to relate it to the priority mempool.

---

#### PR checklist

- ~~[ ] Tests written/updated~~

### DIFF
--- a/docs/references/architecture/adr-113-mempool-lanes.md
+++ b/docs/references/architecture/adr-113-mempool-lanes.md
@@ -182,9 +182,41 @@ there will be at least a period where the these two properties may not hold for 
 
 ## Alternative Approaches
 
-TODO
+### Priority Mempool
 
-Prio mempool: why is it not a good fit?
+CometBFT used to have a `v1` mempool, specified in Tendermint Core [ADR067][adr067] and deprecated as of `v0.37.x`,
+which supported per-transaction priority assignment.
+The key point of the priority mempool's design was that `CheckTxResponse` was extended with a few fields,
+one of which being an `int64` that the application could use to provide a priority to the transaction being checked.
+
+This design can be seen as partially addressing the specification of a Mempool with QoS
+presented in the previous section. Every possible value of the `int64` priority field returned by the application
+can be understood as a _different_ traffic class.
+Let us examine whether the properties specified above are fulfilled by the priority mempool design
+as described in [ADR067][adr067]:
+
+1. Partial ordering of all transactions is maintained because the design still keeps a FIFO queue for gossiping transactions.
+  Also, transactions are reaped according to non-decreasing priority first, and then in FIFO order
+  for transactions with equal priority (see this `ReapMaxBytesMaxGas`'s [docstring][reapmaxbytesmaxgas]).
+1. Since the priority mempool uses FIFO for transactions of equal priority, it also fulfills FIFO ordering per class.
+  The problem here is that, since every value of the priority `int64` field is considered a different transaction class,
+  there are virtually unlimited traffic classes.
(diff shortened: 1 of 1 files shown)


docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -139,11 +143,80 @@ the mempool, regardless of their classes, will have a *partial order*.
...
+to be useful and predictable for the whole appchain network.
+These properties are expressed in terms of consistency of the information, configuration and behaviour
+across nodes in the network.
+
+:parking: _Property_ **Consistent transaction classes**: For any transaction `tx`,
+and any two correct nodes $p$ and $q$ that receive `tx` for the first time,
+$p$ and $q$ MUST have the same the set of transaction classes and their relative priority and configuration,
+as long as `tx` has not been included in a block.
+
+> TODO: Do we really need to require that the `tx` hasn't been decided? Isn't it adding complexity for nothing?
```
hvanz, line 159: This property needs to cover the cases when the node actually makes a CheckTx request to the app. This happens when the tx arrives for the first time (meaning tx first passes the cache check) and when rechecking. So both scenarios are before the tx is decided in a block. We have to keep in mind that a tx may be decided in a block and still be in the cache. And that a node may never receive a tx from a peer, thus not having the tx in the mempool, but still add the tx to the cache (when updating the mempool with a block containing the tx). So I'd say that the property is correct as it is.

hvanz (approved):

docs/references/architecture/adr-113-mempool-lanes.md
sergio-mena: Let's maybe briefly discuss it verbally on Mon

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -139,11 +143,80 @@ the mempool, regardless of their classes, will have a *partial order*.
...
 
+#### Network-wide consistency
+
+The properties presented so far may be interpreted as per-node properties.
+However, we need to define some network-wide properties in order for a mempool QoS implementation
+to be useful and predictable for the whole appchain network.
+These properties are expressed in terms of consistency of the information, configuration and behaviour
+across nodes in the network.
+
+> TODO: Define "for the first time" here: receveived from RPC, peer, or `Update`
```
sergio-mena, line 154:
```suggestion
> TODO: Define "for the first time" here: received from RPC, peer, or `Update`
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -139,11 +143,80 @@ the mempool, regardless of their classes, will have a *partial order*.
...
+Even if we ensure consistent transaction classes, if this property does not hold, a given transaction
+may not receive the same classification across the network and it will thus be impossible to reason
+about any network-wide guarantees we want to provide that transaction with.
+
+Additionally, it is important to note that these two properties also constrain the way transaction
+classes and transaction classification logic can evolve in an existing implementation.
+If either transaction classes or classification logic are not modified in a coordinated manner in a working system,
+there will be at least a period where the these two properties may not hold for all transactions.
+
+> TODO: Need to find somewhere in the text to say: "ReCheckTx" doesn't classify, it's mempool information is disregarded"
```
sergio-mena, line 181:
```suggestion
> TODO: Need to find somewhere in the text to say: "ReCheckTx" doesn't classify, its mempool information is disregarded"
```

PR #3117, opened 2024-05-26
perf(types): 3x speedup MakePartSet

ValarDragon:
This PR adds some benchmarks, and significantly speeds up types.MakePartSet, and Partset.AddPart. (Used by the block proposer, and every consensus instance) It does so by doing two things:
- Saving mutexes on the newly created bit array, by defaulting every value to True (rather than setting it in a loop that goes through a mutex)
- Uses the same hash object throughout, and avoids an extra copy of every leaf. (main speedup)

I do the same hash optimization for proof.Verify, which is used in the add block part codepath for both the proposer and every full node.

New:
```
BenchmarkMakePartSet/nParts=1-12         	   38616	     29817 ns/op	     568 B/op	      12 allocs/op
BenchmarkMakePartSet/nParts=2-12         	   19888	     59866 ns/op	    1000 B/op	      22 allocs/op
BenchmarkMakePartSet/nParts=3-12         	   12979	     95691 ns/op	    1528 B/op	      33 allocs/op
BenchmarkMakePartSet/nParts=4-12         	    8688	    128192 ns/op	    2024 B/op	      44 allocs/op
BenchmarkMakePartSet/nParts=5-12         	    7308	    155224 ns/op	    2888 B/op	      57 allocs/op
```

Old:
```
BenchmarkMakePartSet/nParts=1-12         	   16647	    106545 ns/op	   74169 B/op	      12 allocs/op
BenchmarkMakePartSet/nParts=2-12         	   10000	    106361 ns/op	  148329 B/op	      23 allocs/op
BenchmarkMakePartSet/nParts=3-12         	    6992	    337644 ns/op	  222587 B/op	      35 allocs/op
BenchmarkMakePartSet/nParts=4-12         	    3488	    480109 ns/op	  296811 B/op	      47 allocs/op
BenchmarkMakePartSet/nParts=5-12         	    2228	    557768 ns/op	  371404 B/op	      61 allocs/op
```

System wide, this is definitely not our issue (looks like roughly .1ms per blockpart), but still definitely useful time to remove

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/.changelog/unreleased/improvements/3117-significantly-speedup-make-partset.md
+++ b/.changelog/unreleased/improvements/3117-significantly-speedup-make-partset.md
@@ -1,2 +1,2 @@
-- [`types`] Significantly speedup types.MakePartSet, which is used in creating a block proposal
+- [`types`] Significantly speedup types.MakePartSet and types.AddPart, which are used in creating a block proposal
   ([\#3117](https://github.com/cometbft/cometbft/issues/3117)
--- a/.changelog/unreleased/improvements/3117-significantly-speedup-make-partset.md
+++ b/.changelog/unreleased/improvements/3117-significantly-speedup-make-partset.md
@@ -1,0 +1,2 @@
+- [`types`] Significantly speedup types.MakePartSet, which is used in creating a block proposal
+  ([\#3117](https://github.com/cometbft/cometbft/issues/3117)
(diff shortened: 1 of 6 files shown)


melekes (approved):
Thanks @ValarDragon ❤️

sergio-mena (approved):
Amazing 🎉

ValarDragon:
Ok to backport to 1.x and 38.x ?

melekes:
@sergio-mena ^

PR #3129, opened 2024-05-27
perf(types): Speedup valset.GetByAddress by noticing we do not need to copy the validator object

ValarDragon:
Speedup valset.GetByAddress by noticing we do not need to copy the validator object in ~every single usage, as it operates on read-only instances.

This acts as a very small speedup to ValidateBlock, and calculating statistics for enter prevote. In the future we can change data structures for bigger impact.

If we think this needs more tests to feel comfortable, I think we should close this  PR. Was a driveby change, and there are more impactful things we can do

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
Thanks @ValarDragon ❤️

types/validator_set.go
```diff
@@ -269,9 +269,21 @@ func (vals *ValidatorSet) HasAddress(address []byte) bool {
...
+
+// GetByAddressMut returns an index of the validator with address and the
+// direct validator object if found. Mutations on this return value affect the validator set.
+// This method should be used by callers who will not mutate Val.
+// Otherwise, -1 and nil are returned.
+func (vals *ValidatorSet) GetByAddressMut(address []byte) (index int32, val *Validator) {
 	for idx, val := range vals.Validators {
 		if bytes.Equal(val.Address, address) {
-			return int32(idx), val.Copy()
+			return int32(idx), val
```
sergio-mena, line 286: Are 100% sure that no callsite will try to modify the validator structure we are returning here?

sergio-mena: Hmm... anyway, inside `Validator`, field `Address` is a slice, and field `PubKey` is an interface, so method `Copy()` in the original code wasn't really _copying_ these.

So, I like the idea of removing the `Copy()` because it makes the code less messy, but we need to make sure the other fields (`VotingPower` and `ProposerPriority`) are not modified at any callsite of `GetByAddressMut`.

andynog: I've check the calls to the `GetByAddressMut` and seems not code invoking it are mutating the val returned. 

Afaiu, we're not eliminating the `Copy()` but just providing an alternative method to return a val that is not a `Copy()`

But I agree, that the `Copy()` is probably not doing a deep copy properly. One enhancement could be outside this PR is to add unit tests to ensure this fact doesn't have additional consequences.

andynog (approved):
lgtm, there's not way to prevent someone using the `GetByAddressMut()` and mutating the val by accident but I think having that information in the methods description should be good enough to prevent erroneous uses.

andynog:
Should this be back ported to `v1` ?

melekes:
> Should this be back ported to `v1` ?

Since the change is a bit risky, we'd like to test this out more thoroughly and also hear the feedback from Osmosis (who had this changes backported to v0.37.x branch of their fork). cc @ValarDragon

melekes:
@mergify backport v1.x

PR #3131, opened 2024-05-27
fix(mempool): Move senders back from reactor to clist_mempool entries

### LINKED ISSUE #3084: mempool: tx could be sent back to sender
On [ADR-105](https://github.com/cometbft/cometbft/pull/1032), it was decided to move the list of senders from the `mempoolTx` entries in `CList` to the mempool reactor. This was implemented in #1010. The interface of `CheckTx` was changed from
```golang
CheckTx(tx types.Tx, cb func(*abci.ResponseCheckTx), txInfo TxInfo) error
```
to
```golang
CheckTx(tx types.Tx) (abcicli.ReqRes, error)
```
where `TxInfo` used to contain the sender information. With this modification, the reactor now calls `CheckTx` and record the sender like this:
```golang
reqRes, err := memR.mempool.CheckTx(tx)
...
reqRes.SetCallback(func(res *abci.Response) {
	if res.GetCheckTx().Code == abci.CodeTypeOK {
        	memR.addSender(tx.Key(), sender.ID())
        }
})
```

The problem with this design is that the callback may introduce a delay when recording the sender, and `broadcastTxRoutine` may read the transactions as soon as they are added to clist. So, when `broadcastTxRoutine` tries to read the list of senders in the reactor, the sender is not stored yet, so it sends back the transaction to the sender.

A potential solution would be to move back the sender to the `CList` entries. Then the sender would be stored together with the transaction in `mempoolTx` and would be available at the moment the transaction is broadcasted. The drawback of this solution is that it reintroduces the mixing of the pool implementation and the gossip protocol.

Also note that:
- Removing the `cb` parameter from the original `CheckTx` signature does not pose a problem and it's still useful to keep it out of `CheckTx`. This callback was only used by the RPC endpoints `broadcast_sync` and `broadcast_commit` to handle the CheckTx response. Now this is done with `abcicli.ReqRes`.
- ADR-105 was not backported to `v0.38.x`.

hvanz:
Close #3084 

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/abci/client/client.go
+++ b/abci/client/client.go
@@ -36,6 +36,11 @@ type Client interface {
 	// with the exception of `CheckTxAsync` which we maintain
 	// for the v0 mempool. We should explore refactoring the
 	// mempool to remove this vestige behavior.
+	//
+	// SetResponseCallback is not actually used anymore. It was used only by the mempool on CheckTx
+	// requests. Now the callback for processing the response is set on the ReqRes returned by
+	// CheckTxAsync. This is more flexible as it allows to pass other tx information such as the
+	// sender.
 	SetResponseCallback(cb Callback)
 	CheckTxAsync(ctx context.Context, req *types.CheckTxRequest) (*ReqRes, error)
 }
--- a/abci/client/local_client.go
+++ b/abci/client/local_client.go
@@ -61,7 +61,9 @@ func (app *localClient) CheckTxAsync(ctx context.Context, req *types.CheckTxRequ
 }
 
 func (app *localClient) callback(req *types.Request, res *types.Response) *ReqRes {
-	app.Callback(req, res)
+	if app.Callback != nil {
+		app.Callback(req, res)
+	}
 	rr := newLocalReqRes(req, res)
 	rr.callbackInvoked = true
 	return rr
--- a/docs/references/architecture/adr-105-refactor-mempool-senders.md
+++ b/docs/references/architecture/adr-105-refactor-mempool-senders.md
@@ -5,10 +5,11 @@
 - 2023-07-19: Choose callbacks option and mark as accepted (@hvanz)
 - 2023-07-10: Add callback alternative (@hvanz)
 - 2023-06-26: Initial draft (@hvanz)
+- 2024-05-22: Reverted (@hvanz)
 
 ## Status
 
-Accepted
+Reverted
 
 ## Context
 
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -8,7 +8,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	abcicli "github.com/cometbft/cometbft/abci/client"
 	abci "github.com/cometbft/cometbft/abci/types"
 	"github.com/cometbft/cometbft/config"
 	"github.com/cometbft/cometbft/internal/clist"
@@ -31,10 +30,6 @@ type CListMempool struct {
 	// notify listeners (ie. consensus) when txs are available
 	notifiedTxsAvailable atomic.Bool
 	txsAvailable         chan struct{} // fires once for each height, when the mempool is not empty
-
-	// Function set by the reactor to be called when a transaction is removed
-	// from the mempool.
-	removeTxOnReactorCb func(txKey types.TxKey)
 
 	config *config.MempoolConfig
 
@@ -92,8 +87,6 @@ func NewCListMempool(
 		mp.cache = NopTxCache{}
 	}
 
-	proxyAppConn.SetResponseCallback(mp.globalCb)
-
 	for _, option := range options {
 		option(mp)
 	}
@@ -138,7 +131,6 @@ func (mem *CListMempool) removeAllTxs() {
 
 	mem.txsMap.Range(func(key, _ any) bool {
 		mem.txsMap.Delete(key)
-		mem.invokeRemoveTxOnReactor(key.(types.TxKey))
 		return true
 	})
 }
@@ -146,18 +138,6 @@ func (mem *CListMempool) removeAllTxs() {
 // NOTE: not thread safe - should only be called once, on startup.
 func (mem *CListMempool) EnableTxsAvailable() {
 	mem.txsAvailable = make(chan struct{}, 1)
-}
-
-func (mem *CListMempool) SetTxRemovedCallback(cb func(txKey types.TxKey)) {
-	mem.removeTxOnReactorCb = cb
-}
-
-func (mem *CListMempool) invokeRemoveTxOnReactor(txKey types.TxKey) {
-	// Note that the callback is nil in the unit tests, where there are no
-	// reactors.
-	if mem.removeTxOnReactorCb != nil {
-		mem.removeTxOnReactorCb(txKey)
-	}
 }
 
 // SetLogger sets the Logger.
@@ -245,7 +225,11 @@ func (mem *CListMempool) TxsWaitChan() <-chan struct{} {
 
 // It blocks if we're waiting on Update() or Reap().
 // Safe for concurrent use by multiple goroutines.
-func (mem *CListMempool) CheckTx(tx types.Tx) (*abcicli.ReqRes, error) {
+func (mem *CListMempool) CheckTx(
+	tx types.Tx,
+	externalCb func(*abci.CheckTxResponse),
+	txInfo TxInfo,
+) error {
 	mem.updateMtx.RLock()
 	// use defer to unlock mutex because application (*local client*) might panic
 	defer mem.updateMtx.RUnlock()
@@ -253,11 +237,11 @@ func (mem *CListMempool) CheckTx(tx types.Tx) (*abcicli.ReqRes, error) {
 	txSize := len(tx)
 
 	if err := mem.isFull(txSize); err != nil {
-		return nil, err
+		return err
 	}
 
 	if txSize > mem.config.MaxTxBytes {
-		return nil, ErrTxTooLarge{
+		return ErrTxTooLarge{
 			Max:    mem.config.MaxTxBytes,
 			Actual: txSize,
 		}
@@ -265,21 +249,37 @@ func (mem *CListMempool) CheckTx(tx types.Tx) (*abcicli.ReqRes, error) {
 
 	if mem.preCheck != nil {
 		if err := mem.preCheck(tx); err != nil {
-			return nil, ErrPreCheck{Err: err}
+			return ErrPreCheck{Err: err}
 		}
 	}
 
 	// NOTE: proxyAppConn may error if tx buffer is full
 	if err := mem.proxyAppConn.Error(); err != nil {
-		return nil, ErrAppConnMempool{Err: err}
+		return ErrAppConnMempool{Err: err}
 	}
 
 	if added := mem.addToCache(tx); !added {
 		mem.metrics.AlreadyReceivedTxs.Add(1)
+		if txInfo.sender != "" {
+			// Record a new sender for a tx we've already seen.
+			// Note it's possible a tx is still in the cache but no longer in the mempool
+			// (eg. after committing a block, txs are removed from mempool but not cache),
+			// so we only record the sender for txs still in the mempool.
+			if elem, ok := mem.getCElement(tx.Key()); ok {
+				memTx := elem.Value.(*mempoolTx)
+				if found := memTx.addSender(txInfo.sender); found {
+					// It should not be possible to receive twice a tx from the same sender.
+					mem.logger.Error("tx already received from peer", "tx", tx.Hash(), "sender", txInfo.sender)
+				}
+				// TODO: consider punishing peer for dups,
+				// its non-trivial since invalid txs can become valid,
+				// but they can spam the same tx with little cost to them atm.
+			}
+		}
 		// TODO: consider punishing peer for dups,
 		// its non-trivial since invalid txs can become valid,
 		// but they can spam the same tx with little cost to them atm.
-		return nil, ErrTxInCache
+		return ErrTxInCache
 	}
 
 	reqRes, err := mem.proxyAppConn.CheckTxAsync(context.TODO(), &abci.CheckTxRequest{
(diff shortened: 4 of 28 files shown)


mempool/reactor.go
```diff
@@ -276,7 +253,7 @@ func (memR *Reactor) broadcastTxRoutine(peer p2p.Peer) {
 		// NOTE: Transaction batching was disabled due to
 		// https://github.com/tendermint/tendermint/issues/5796
 
-		if !memR.isSender(memTx.tx.Key(), peer.ID()) {
+		if !memTx.isSender(peer.ID()) {
```
hvanz, line 256: This is the main reason of this PR. When reading the sender, it must be available immediately to not send back the tx. By storing the sender in the reactor, a small delay was introduced by the callback.

mempool/clist_mempool.go
```diff
@@ -92,8 +87,6 @@ func NewCListMempool(
 		mp.cache = NopTxCache{}
 	}
 
-	proxyAppConn.SetResponseCallback(mp.globalCb)
```
hvanz, line 95: The `globalCb` callback was only used for Recheck responses. Now the response is simply handled by a callback in the `ReqRes` structure.

mempool/mempool.go
```diff
@@ -25,7 +25,7 @@ const (
 type Mempool interface {
 	// CheckTx executes a new transaction against the application to determine
 	// its validity and whether it should be added to the mempool.
-	CheckTx(tx types.Tx) (*abcicli.ReqRes, error)
+	CheckTx(tx types.Tx, externalCb func(*abci.CheckTxResponse), txInfo TxInfo) error
```
hvanz, line 28: Now `CheckTx` has the same signature as before (also the same as in v0.38).

mempool/clist_mempool.go
hvanz: There are three structures with a callback:
- `proxyAppConn`, implementing the `Client` interface, with a callback of type `func(*types.Request, *types.Response)`. Before this PR, this callback was used only on CheckTx ABCI responses. It is set only once, during the creation of `CListMempool`.
- `ReqRes`, with a callback of type `func(*types.Response)`. This is more flexible because we can set it when `CheckTxAsync` returns the request/response, allowing us to customise the function by passing extra information, such as the sender.
- the external callback passed to `CListMempool.CheckTx`, reintroduced in this PR, and invoked at the end of handling the CheckTx response. This is used for processing the response on the RPC endpoints `broadcast_tx_sync` and `broadcast_tx_commit`.

mempool/clist_mempool.go
```diff
@@ -289,51 +283,81 @@ func (mem *CListMempool) CheckTx(tx types.Tx) (*abcicli.ReqRes, error) {
...
+			gasWanted: res.GasWanted,
+			tx:        tx,
+		}
+		if mem.addTx(&memTx, txInfo) {
+			mem.notifyTxsAvailable()
 
-	default:
-		// ignore other messages
+			// update metrics
+			mem.metrics.Size.Set(float64(mem.Size()))
```
hvanz, line 338: Now, update metrics only when the mempool is modified.

mempool/clist_mempool.go
```diff
@@ -289,51 +283,84 @@ func (mem *CListMempool) CheckTx(tx types.Tx) (*abcicli.ReqRes, error) {
 	if err != nil {
 		panic(fmt.Errorf("CheckTx request for tx %s failed: %w", log.NewLazySprintf("%v", tx.Hash()), err))
 	}
+	reqRes.SetCallback(mem.handleCheckTxResponse(tx, txInfo))
```
hvanz, line 286: The CheckTx response is handled again by the `ReqRes` callback, as in v0.38.x.

mempool/clist_mempool.go
```diff
@@ -622,21 +628,21 @@ func (mem *CListMempool) recheckTxs() {
...
-		_, err := mem.proxyAppConn.CheckTxAsync(context.TODO(), &abci.CheckTxRequest{
+		// Send CheckTx request to the app to re-validate transaction.
+		resReq, err := mem.proxyAppConn.CheckTxAsync(context.TODO(), &abci.CheckTxRequest{
 			Tx:   tx,
 			Type: abci.CHECK_TX_TYPE_RECHECK,
 		})
 		if err != nil {
 			panic(fmt.Errorf("(re-)CheckTx request for tx %s failed: %w", log.NewLazySprintf("%v", tx.Hash()), err))
 		}
+		resReq.SetCallback(mem.handleRecheckTxResponse(tx))
```
hvanz, line 645: The CheckTx response for rechecking transactions is handled now by `ReqRes` callbacks set right sending the request, instead of `globalCb`, which was set at as the `proxyApp` callback.

abci/client/local_client.go
```diff
@@ -70,6 +72,7 @@ func (app *localClient) callback(req *types.Request, res *types.Response) *ReqRe
 func newLocalReqRes(req *types.Request, res *types.Response) *ReqRes {
 	reqRes := NewReqRes(req)
 	reqRes.Response = res
+	reqRes.Done() // release waiters on response
```
hvanz, line 75: This is a fix to the local and unsync clients. Now we can do `reqRes.Wait()`.

melekes (approved):
👍

UPGRADING.md
```diff
@@ -142,23 +142,26 @@ The `Mempool` interface was modified on `CheckTx`. Note that this interface is
...
+`CheckTx`'s signature changed from 
+`CheckTx(tx types.Tx, cb func(*abci.ResponseCheckTx), txInfo TxInfo) error` to 
+`CheckTx(tx types.Tx, txInfo *TxInfo) (*abcicli.ReqRes, error)`.
+The method used to take a callback function `cb` to be applied to the
+ABCI `CheckTx` response. Now `CheckTx` returns the ABCI response of
+type `*abcicli.ReqRes`, on which one can apply any callback manually.
+For example:
+```golang
+reqRes, err := CheckTx(tx, &txInfo)
+cb(reqRes.Response.GetCheckTx())
```
sergio-mena, line 154:
```suggestion
// check `err` here
cb(reqRes.Response.GetCheckTx())
```

UPGRADING.md
```diff
@@ -142,23 +142,26 @@ The `Mempool` interface was modified on `CheckTx`. Note that this interface is
...
-- The second parameter was `txInfo`, which essentially contained
-  information about the sender of the transaction. Now that information
-  is stored in the mempool reactor instead of the data structure, so it
-  is no longer needed in this method.
+The `*abcicli.ReqRes` structure that `CheckTx` returns has a callback to 
+process the response already set (namely, the function `handleCheckTxResponse`).
+The callback can be invoked manually; for example:
+```golang
+reqRes, err := CheckTx(tx, &txInfo)
+reqRes.Wait()
```
sergio-mena, line 162:
```suggestion
// check `err` here
reqRes.Wait()
```

abci/client/client.go
```diff
@@ -36,6 +36,11 @@ type Client interface {
 	// with the exception of `CheckTxAsync` which we maintain
 	// for the v0 mempool. We should explore refactoring the
 	// mempool to remove this vestige behavior.
+	//
+	// SetResponseCallback is not used anymore. The callback was invoked only by the mempool on
+	// CheckTx responses, only during rechecking. Now the responses are handled by the callback of
+	// the *ReqRes struct returned by CheckTxAsync. This callback is more flexible as it allows to
+	// pass other information such as the sender.
 	SetResponseCallback(cb Callback)
```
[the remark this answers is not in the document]
sergio-mena: I just checked, and:

* This no longer used anywhere in CometBFT (except for a proxy class, which is in turn not used
* On `main` branch of the SDK, they don't use it for anything.

Since CometBFT `v1.x` will be a major version. I think we could even remove it in `v1.x`. What do you think? If you find it to reckless, we can deprecate in on `v1.x`, as @melekes suggests, and remove it altogether on `main`.

abci/client/local_client.go
```diff
@@ -61,7 +61,9 @@ func (app *localClient) CheckTxAsync(ctx context.Context, req *types.CheckTxRequ
 }
 
 func (app *localClient) callback(req *types.Request, res *types.Response) *ReqRes {
-	app.Callback(req, res)
+	if app.Callback != nil {
```
sergio-mena, line 64: How come we weren't checking the functor for `nil` before?

abci/client/local_client.go
sergio-mena: I don't see this change on unsync client... my understanding is that this function is also used by the unsync client. If I'm right, could you add a comment on top of the function saying the unsync client also uses it?

docs/references/architecture/adr-105-refactor-mempool-senders.md
```diff
@@ -147,6 +148,11 @@ chances of concurrent accesses to the list of senders and it removes the
 transaction immediately, keeping the mempool and the list of senders better
 synchoronized.
 
+Update: We have decided to revert this change as it may cause in some cases that the transaction is
+sent back to the sender. Storing the sender in the reactor after storing the transaction in `CList`
+introduces a small delay resulting in the transaction having no sender during a very small window of
```
sergio-mena, line 153:
```suggestion
introduces a race condition resulting in the transaction having no sender during a very small window of
```

rpc/core/mempool.go
```diff
@@ -42,16 +42,16 @@ func (env *Environment) BroadcastTxSync(ctx *rpctypes.Context, tx types.Tx) (*ct
...
 		return nil, err
 	}
-	reqRes.SetCallback(func(_ *abci.Response) {
-		select {
-		case <-ctx.Context().Done():
-		case resCh <- reqRes.Response.GetCheckTx():
-		}
-	})
+	go func() {
+		reqRes.Wait()
```
sergio-mena, line 50: Shouldn't you still the select with a `case` with `ctx.Context().Done()` so it doesn't on `Wait()` when the context is cancelled?

abci/client/local_client.go
hvanz: Yes, `newLocalReqRes` is called in `unsync_local_client`. I will add a comment.

abci/client/local_client.go
hvanz: It was never the case that app.Callback was not set.

mempool/mempool.go
```diff
@@ -147,3 +144,9 @@ func PostCheckMaxGas(maxGas int64) PostCheckFunc {
 
 // TxKey is the fixed length array key used as an index.
 type TxKey [sha256.Size]byte
+
+// TxInfo is a parameter passed to CheckTx when attempting to add a new tx to the mempool.
+type TxInfo struct {
```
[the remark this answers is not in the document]
hvanz: It's mainly to maintain this part of the signature of CheckTx as before. True that now it's `*TxInfo`, while before it was just `TxInfo`, so it's not exactly the same.

mempool/clist_mempool.go
```diff
@@ -289,51 +283,84 @@ func (mem *CListMempool) CheckTx(tx types.Tx) (*abcicli.ReqRes, error) {
...
-//   - resCbFirstTime (lock not held) if tx is valid
-func (mem *CListMempool) addTx(memTx *mempoolTx) bool {
+//   - handleCheckTxResponse (lock not held) if tx is valid
+func (mem *CListMempool) addTx(memTx *mempoolTx, txInfo *TxInfo) bool {
 	tx := memTx.tx
+	txKey := tx.Key()
+
+	// Check if the transaction is already in the mempool.
+	if elem, ok := mem.getCElement(txKey); ok {
+		if txInfo != nil {
```
sergio-mena, line 355: Can we have a case where `txInfo` is `nil`?

mempool/clist_mempool.go
sergio-mena: > The globalCb callback was only used for Recheck responses.

I see there a `case` for `CheckTx` in addition to the one for `ReCheckTx`

sergio-mena: Should this useful explanation be somewhere in the form of a comment?

rpc/core/mempool.go
hvanz: Yes. Fixed.

mempool/mempoolTx.go
```diff
@@ -11,9 +13,29 @@ type mempoolTx struct {
...
 }
+
+func (memTx *mempoolTx) isSender(peerID p2p.ID) bool {
+	_, ok := memTx.senders.Load(peerID)
+	return ok
+}
+
+// Add the peer ID to the list of senders. Return true iff it exists already in the list.
+func (memTx *mempoolTx) addSender(peerID p2p.ID) bool {
+	if peerID == "" {
```
sergio-mena, line 34:
```suggestion
	if len(peerID) == 0 {
```

sergio-mena (approved):
Thanks for this!

mempool/clist_mempool.go
hvanz: This is a corner case, allowed by `config.KeepInvalidTxsInCache`. If we remove this option, we could simplify this code and panic if there's no sender.

mempool/mempool.go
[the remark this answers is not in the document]
hvanz: Ok, so I simplied to CheckTx signature by passing a p2p.ID for the sender instead of *TxInfo.

abci/client/client.go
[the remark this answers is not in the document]
hvanz: I marked it as deprecated.

mempool/reactor.go
```diff
@@ -163,24 +152,11 @@ func (memR *Reactor) Receive(e p2p.Envelope) {
...
 		for _, txBytes := range protoTxs {
 			tx := types.Tx(txBytes)
-			reqRes, err := memR.mempool.CheckTx(tx)
-			switch {
-			case errors.Is(err, ErrTxInCache):
+			_, err := memR.mempool.CheckTx(tx, e.Src.ID())
+			if errors.Is(err, ErrTxInCache) {
 				memR.Logger.Debug("Tx already exists in cache", "tx", tx.Hash())
-			case err != nil:
+			} else {
```
jmalicevic, line 158: @hvanz  Is this not outputting this error even when there is no error?

jmalicevic: Sorry this was fixed!

Issue #1247, opened 2023-08-14
Support for applications to detect if node is in sync with network or replaying blocks


adizere:
## Feature Request

We got feedback from multiple users that the application might have specific logic that is not supposed to run in case the node is not in sync with the rest of the network.

### Problem Definition

_What problems may be addressed by introducing this feature?_
_What benefits does CometBFT stand to gain by including this feature?_

User feedback:

> I need to know if the blockchain is doing a block sync or not. Because when syncing, there is some application logic that I don't need to execute.
> How can I know if the consensus engine is synchronizing or not?
> Before [_in v0.36_] I did `app.Service.(*node.Node).ConsensusReactor().WaitSync()`. I'm not sure what would be ideal. For now, I'm just missing the ConsensusReactor access. So I can check if the blockchain is still syncing or not.

For ref. on the above discussion, we had trimmed the Go API surface in v0.38 to remove accessor to `ConsensusReactor`. Since a user highlighted they are using this API, we [reverted that change in v0.38 RC3](https://github.com/cometbft/cometbft/pull/1122/files).

> we have a similar use case
> basically there is some node-local logic (e.g. not part of consensus itself) which needs to know whether the node is syncing or not as it doesn't make sense to run it during sync

> is there a way for the application to know that cometbft is currently replaying blocks and not at the head of the chain taking part of conssensus with the other nodes?
> Every of our validator nodes are sourcing log events from ethereum to find deposits [for example]. And when doing this, sending an event on our chain, from there all nodes are also starting procedure to verify the event really happen on ethereum, from there if they can confirm it, they will send a vote on our chain, and when 2/3 of the validator vote for a deposit, an account on our chain is credited. Now the issue is that when replaying, the node will be emitting event / votes for things which have been acknowledge by our chain for a while. So while it is not a big deal in term of consistency as the chain/app protects against using transaction twice, it's a bit of an issue as the node end up filling their mempool with lots of event which are far in the past, and unnecessary, but also increase the processing. 

For ref on this discussion, it seems like https://github.com/cometbft/cometbft/issues/785 is a potential fix.

### Proposal

Adjust the `FinalizeBlock` call to add height of tip of chain.

PR #3134, opened 2024-05-27
feat(proto): add `SyncingToHeight` to `FinalizeBlockRequest`

melekes:
to let the ABCI app know if CometBFT is syncing or not. If the node is syncing/replaying blocks, the value is the maximum known height. If not, `SyncingToHeight` equals `Height` in `FinalizeBlockRequest`.

Closes https://github.com/cometbft/cometbft/issues/1247

---

#### PR checklist

- [ ] Tests written/updated


spec/abci/abci++_methods.md
```diff
@@ -611,6 +611,7 @@ without calling `VerifyVoteExtension` to verify it.
     | time                 | [google.protobuf.Timestamp][protobuf-timestamp] | Timestamp of the finalized block.                                                         | 6            |
     | next_validators_hash | bytes                                           | Merkle root of the next validator set.                                                    | 7            |
     | proposer_address     | bytes                                           | [Address](../core/data_structures.md#address) of the validator that created the proposal. | 8            |
+    | syncing_to_height    | int64                                           | If the node is syncing/replaying blocks - target height. If not, syncing_to == height.    | 9            |
```
andynog, line 614:
```suggestion
    | syncing_to_height    | int64                                           | If the node is syncing/replaying blocks then syncing_to_height == target height. If not, syncing_to_height == height.    | 9            |  
```

internal/blocksync/reactor.go
```diff
@@ -586,7 +586,7 @@ func (bcR *Reactor) processBlock(first, second *types.Block, firstParts *types.P
 
 	// TODO: same thing for app - but we would need a way to
 	// get the hash without persisting the state
-	state, err = bcR.blockExec.ApplyVerifiedBlock(state, firstID, first)
+	state, err = bcR.blockExec.ApplyVerifiedBlock(state, firstID, first, bcR.pool.MaxPeerHeight())
```
andynog, line 589: just want to check, the intention here is to use `bcR.pool.MaxPeerHeight()` because it sets a lock on the pool, but if you're just reading the value, could the `bcR.pool.maxPeerHeight` be used instead to avoid contention ? (but not too sure about the mechanism to update the max peer height, so that might be why)

andynog (approved):
looks good to me, minor suggestions/comments

internal/blocksync/reactor.go
melekes: I don't like the lock either, but we need it (or atomic var) to prevent data races.

state/execution.go
```diff
@@ -258,15 +259,14 @@ func (blockExec *BlockExecutor) applyBlock(state State, blockID types.BlockID, b
...
+		"syncing_to_height", syncingToHeight,
 	)
 
 	// Assert that the application correctly returned tx results for each of the transactions provided in the block
 	if len(block.Data.Txs) != len(abciResponse.TxResults) {
 		return state, fmt.Errorf("expected tx results length to match size of transactions in block. Expected %d, got %d", len(block.Data.Txs), len(abciResponse.TxResults))
 	}
 
-	blockExec.logger.Info("executed block", "height", block.Height, "app_hash", fmt.Sprintf("%X", abciResponse.AppHash))
-
```
sergio-mena, line 269: Why are we removing this?

internal/blocksync/pool.go
```diff
@@ -78,9 +79,10 @@ type BlockPool struct {
 	requesters map[int64]*bpRequester
 	height     int64 // the lowest key in requesters.
 	// peers
-	peers         map[p2p.ID]*bpPeer
-	sortedPeers   []*bpPeer // sorted by curRate, highest first
-	maxPeerHeight int64     // the biggest reported height
+	peers       map[p2p.ID]*bpPeer
+	sortedPeers []*bpPeer // sorted by curRate, highest first
+
+	maxPeerHeight atomic.Int64 // the biggest reported height
```
sergio-mena, line 85: Why do we need to turn this into an atomic? Don't we have a lock already? I mean, is there evidence we're improving contention?

sergio-mena (approved):
Thanks @melekes  for implementing this! 🙏

state/execution.go
melekes: Duplicate log line. See 3 lines above.

internal/blocksync/pool.go
melekes: No hard evidence. https://github.com/cometbft/cometbft/pull/3134#discussion_r1617918338

Do you think I should revert?

Issue #3130, opened 2024-05-27
Don't re-use the internal message queue in block proposal


ValarDragon:
### Summary

Currently when you propose a block, you build and chunk up the proposal block, and sign it. Then we put our "Proposal" message onto the internal message queue, and then add every block part into the proposal queue.

https://github.com/cometbft/cometbft/blob/main/internal/consensus/state.go#L1250-L1261

This causes us to suffer the following latencies:
- We have to process everything ahead of the proposal & blockpart in the queue. Which could be very many peer votes gathered during proposal creation process.
- We have to go through extra validation logic (unlikely to be the bottleneck short term)

### Problem Definition

This + #3007 is likely contributing lots of delay to our block gossip times from the block proposer.

### Proposal

Instead of sending these messages onto the internal message queue's, lets handle writing the proposal data + block parts to WAL here (doing all the encoding in parallel, which is >75% of WAL write time), and then run the internal methods directly.

ValarDragon:
After looking at the logic more carefully in #3136, I think what we should do is:

On Propose:
- Run the SetProposal method immediately
- Add all block parts immediately
- Make a new message on the internal message queue for "handleCompleteProposal" (which will take us into prevote).

We want to skip the delay for adding all block parts for gossip routines, but we want to unblock gossiping block data before doing our prevote.

I suggest the following implementation strategy:
- [ ] PR #1: Run SetProposal immediately
- [ ] PR #2: Make a new message for "handleCompleteProposal". Do each "addProposalBlockPart" and then put this new message into the internal message queue.
- [ ] PR #3: Replace add proposal block part with more efficient methods when its coming from us.

The over-arching claim is that after we sign a proposal, getting ourselves ready to gossip all block parts _should_ be instant, but right now its very much not, so we need to move to that :)

cason:
> After looking at the logic more carefully in #3136, I think what we should do is:
> 
> On Propose:
> 
>     * Run the SetProposal method immediately
> 
>     * Add all block parts immediately
> 
>     * Make a new message on the internal message queue for "handleCompleteProposal" (which will take us into prevote).
> 
> 
> We want to skip the delay for adding all block parts for gossip routines, but we want to unblock gossiping block data before doing our prevote.
> 
> I suggest the following implementation strategy:
> 
>     * [ ]  PR 1: Run SetProposal immediately
> 
[...]

PR #3137, opened 2024-05-28
perf(consensus): Make the proposer run setProposal immediately

ValarDragon:
Component of #3130 . This PR makes the proposer, upon building a block, directly update its local Proposal setting, rather than going through the internal message queue. This is progressing towards the goal, of minimizing the time between "proposer has locally built a block", and "proposer is ready to gossip block parts".

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/internal/consensus/byzantine_test.go
+++ b/internal/consensus/byzantine_test.go
@@ -176,7 +176,7 @@ func TestByzantinePrevoteEquivocation(t *testing.T) {
 	// proposed will have a valid timestamp
 	lazyProposer := css[1]
 
-	lazyProposer.decideProposal = func(height int64, round int32) {
+	lazyProposer.decideProposal = func(height int64, round int32) func() {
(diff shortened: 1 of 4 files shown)


ValarDragon:
Hrmm, I clearly don't understand something about how the WAL works. I can't tell if its just the crashing WAL tests, or the real WAL replay that has an issue here. I suspect its the crashing WAL tests?

If thats not true, only thing I can think of is that the WAL writes need to have the time in ascending order?

ValarDragon:
Ah I understand the problem. Its that with this change, we first write to WAL
```
{[Proposal Proposal{1/0 ({hash}:1:C028B329B066, -1) C2DDC7EDA7A0 @ 2024-05-28T07:13:04.786963728Z}]  2024-05-28 07:13:04.804440718 +0000 UTC}
```
Then
```
{1 0 RoundStepPropose}
```

Whereas before it went RoundStepPropose then the proposal (and blockparts)

PR #3145, opened 2024-05-29
fix(log): fix panic when log with nil val

0xfocus:
replace `Stringer.String` with `fmt.Sprintf` to for log val stringer

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
Thanks @0xfocus ❤️

andynog (approved):
lgtm. Did some diff testing on cometbft log output with the new code and old code, even though the log might not come in the same order, everything seems to be logging the same.

andynog:
Thanks @0xfocus, could you please take a look at the lint check?

0xfocus:
Thank you @melekes @andynog , I'm sorry, I am not very familiar with this development process.

andynog (dismissed):
Could you please provide a changelog entry in this PR? If you check other PRs you will be able to see how to create one in the unreleased folder.

andynog:
Thanks @0xfocus, this is ready to merge but noticed there's no changelog entry.

Also, for just for additional context, Since there's no issue associated with this, I'd like more context on the issue in particular when you would get a panic for logging a nil val. 

Just trying to understand the use case. We might backport this change to older releases but I'm find interesting the use cases when this panic would be caused. I'd assume that since this piece of code is present even in very older releases, if something would cause a panic when logging I'd assume that this would have surfaced a long time ago. Please let us know. Thanks !

0xfocus:
In my case, I want to log a Hash struct pointer, sometimes the pointer is nil,  so the panic happend
list a example
```
package main

import (
    "encoding/hex"
    "os"

    cmtlog "github.com/cometbft/cometbft/libs/log"
)

type Hash [32]byte

func (h Hash) String() string {
    return "0x" + hex.EncodeToString(h[:])
}

func main() {
    logger := cmtlog.NewTMLogger(cmtlog.NewSyncWriter(os.Stdout))
    var h Hash
    logger.Info("Hash object OK", "hash", h)
    p := &h
    logger.Info("Hash pointer OK", "hash", p)
    p = nil
    logger.Info("nil Hash pointer panic", "hash", p)
}
```

0xfocus:
Hi, @andynog, Is there any problem in this PR?

0xfocus (commented):
add changelog entry for #3145

libs/log/cmtfmt_logger.go
```diff
@@ -92,7 +92,8 @@ func (l tmfmtLogger) Log(keyvals ...any) error {
 
 		// Realize stringers
 		if s, ok := keyvals[i+1].(fmt.Stringer); ok {
-			keyvals[i+1] = s.String()
```
melekes, line 95: Please help me understand one thing: if `s` is `nil`, how come it passes the check for `fmt.Stringer` (`ok`) and then panics?

0xfocus: `keyvals[i+1]` is nil  but `s` isn't a `nil`. you can run the above example to get the panic.

PR #3156, opened 2024-05-31
perf(consensus): Make one peer state mutex call in AddVote

### LINKED ISSUE #3155: Make one peer state mutex call in AddVote
### Summary

There was a regression at some point, and now adding a vote makes 3 mutex locks and unlocks to peerstate: https://github.com/cometbft/cometbft/blob/v0.37.6/consensus/reactor.go#L348-L350

We should just make one method, that takes a lock once, and then does all the relevant operations. I don't have system impact benchmarks, as this 3 mutex-taking is not in v0.37.x

### Proposal

Make one method that just takes the peerstate lock once and does all 3 internal functions

ValarDragon:
Closes #3155 

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

ValarDragon:
Ah correction this is in v0.37.x, and does cause some very slight overhead.

melekes:
@mergify backport v1.x

PR #3157, opened 2024-05-31
perf(consensus): Make late votes outside of last block commits not get to peerMsgQueue

### LINKED ISSUE #3154: Make late votes not enter consensus peer msg queue
### Summary

Related to #2927. We should make the consensus reactor, upon seeing a vote message, check if the vote is late and ignore it, outside of consensus state's processing.

### Problem Definition

On Osmosis RPC nodes, we see 2 million late votes per hour, which I think should mean considerably more work load under the consensus mutex.

### Proposal

Add logic right here (https://github.com/cometbft/cometbft/blob/main/internal/consensus/reactor.go#L360-L361 ) of the form:

```
if vote.Height < height {
// add to late vote metrics
  return
}
```

ValarDragon:
Closes #3154 . Make late votes outside of last blocks precommits not get into the peer msg queue. This lowers the delay for processing more important messages from the peerMsgQueue, and lowers WAL write delays. As noted in the github issue, the number of late votes is actually quite significant.

I don't think theres any test updates needed (all relevant behavior is covered under existing tests. And if you look into the consensus logic, aside from WAL write, all we'd otherwise do is:

```
		cs.Logger.Debug("vote ignored and not added", "vote_height", vote.Height, "cs_height", cs.Height, "peer", peerID)
                return
```

We could still do that if we wanted? But doesn't seem that important, since we have the mark late vote still being sent.

---

#### PR checklist

- [ ] Tests written/updated


internal/consensus/reactor.go
```diff
@@ -352,12 +352,22 @@ func (conR *Reactor) Receive(e p2p.Envelope) {
...
-			ps.EnsureVoteBitArrays(height, valSize)
-			ps.EnsureVoteBitArrays(height-1, lastCommitSize)
+			ps.EnsureVoteBitArrays(csHeight, valSize)
+			ps.EnsureVoteBitArrays(csHeight-1, lastCommitSize)
 			ps.SetHasVote(msg.Vote)
 
+			// if vote is late, and is not a precommit for the last block, mark it late and return.
+			isLate := msg.Vote.Height < csHeight || (msg.Vote.Height == csHeight && msg.Vote.Round < csRound)
+			if isLate {
+				isLastBlockPrecommit := msg.Vote.Type != types.PrecommitType || msg.Vote.Height != csHeight-1
```
melekes, line 364:
```suggestion
				notLastBlockPrecommit := msg.Vote.Type != types.PrecommitType || msg.Vote.Height != csHeight-1
```

ValarDragon: ah ty!

melekes (approved):
👍

ValarDragon:
Ah correction this is in v0.37.x, and does cause some very slight overhead.

cason (commented):
We should not spend much resources on messages (any type) which are old  and useless for the consensus protocol. So, the goal here is perfectly legit.

But adding the solution to the consensus _reactor_ is not really the best approach. In fact, the consensus reactor should not take the consensus _state_ lock for every received vote, which is super inefficient in any case. While this PR hasn't introduced this behavior, if we really want to decouple reactor and state, this solution goes towards the wrong direction.

I would instead add this check to the _state_ implementation. The goal is anyway to prevent writing this useless vote to the WAL and keeping the consensus lock for longer than needed.

ValarDragon:
> But adding the solution to the consensus reactor is not really the best approach. In fact, the consensus reactor should not take the consensus state lock for every received vote, which is super inefficient in any case. While this PR hasn't introduced this behavior, if we really want to decouple reactor and state, this solution goes towards the wrong direction.

My thinking right now is that we can get the lock removed with #3211 , and then have this able to proceed concurrently (not under the single threaded state.go workload)

internal/consensus/reactor.go
```diff
@@ -352,9 +352,19 @@ func (conR *Reactor) Receive(e p2p.Envelope) {
...
 			cs.mtx.RUnlock()
-			ps.SetHasVoteFromPeer(msg.Vote, height, valSize, lastCommitSize)
+			ps.SetHasVoteFromPeer(msg.Vote, csHeight, valSize, lastCommitSize)
+
+			// if vote is late, and is not a precommit for the last block, mark it late and return.
+			isLate := msg.Vote.Height < csHeight || (msg.Vote.Height == csHeight && msg.Vote.Round < csRound)
+			if isLate {
+				notLastBlockPrecommit := msg.Vote.Type != types.PrecommitType || msg.Vote.Height != csHeight-1
+				if notLastBlockPrecommit {
+					cs.metrics.MarkLateVote(msg.Vote.Type)
```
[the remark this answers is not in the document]
ValarDragon: cs.metrics calls are not, they are all concurrency safe. This is a guarantee provided by go-kit/metrics, which cs.metrics uses

.changelog/unreleased/improvements/3157-late-votes-dont-take-cs-mtx.md
```diff
@@ -0,0 +1,2 @@
+- [`types`] Significantly speedup types.MakePartSet and types.AddPart, which are used in creating a block proposal
```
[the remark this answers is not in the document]
cason: I mean, this is the wrong chngelog entry in my opinion.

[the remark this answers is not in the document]
ValarDragon: Yes your right, my bad!

.changelog/unreleased/improvements/3157-late-votes-dont-take-cs-mtx.md
```diff
@@ -0,0 +1,4 @@
+- [`consensus`] Check for late votes in the reactor, preventing them from entering the
```
cason, line 1:
```suggestion
- [`consensus`] Check for late votes in the reactor, preventing them from being delivered to the consensus logic, as a performance optimization.
```

.changelog/unreleased/improvements/3157-late-votes-dont-take-cs-mtx.md
```diff
@@ -0,0 +1,4 @@
+- [`consensus`] Check for late votes in the reactor, preventing them from entering the 
+single-threaded consensus logic. This change is a performance optimization that reduces the number 
+of redundant votes that are processed by the consensus logic.
+  ([\#3157](https://github.com/cometbft/cometbft/issues/3157)
```
cason, line 4: I think adding the issue here is the right approach.

internal/consensus/reactor.go
```diff
@@ -352,9 +352,19 @@ func (conR *Reactor) Receive(e p2p.Envelope) {
 		case *VoteMessage:
 			cs := conR.conS
 			cs.mtx.RLock()
-			height, valSize, lastCommitSize := cs.Height, cs.Validators.Size(), cs.LastCommit.Size()
+			csHeight, csRound, valSize, lastCommitSize := cs.Height, cs.Round, cs.Validators.Size(), cs.LastCommit.Size()
 			cs.mtx.RUnlock()
-			ps.SetHasVoteFromPeer(msg.Vote, height, valSize, lastCommitSize)
+			ps.SetHasVoteFromPeer(msg.Vote, csHeight, valSize, lastCommitSize)
+
+			// if vote is late, and is not a precommit for the last block, mark it late and return.
```
cason, line 359:
```suggestion
			// If a vote is late, and is not a precommit for the last block, mark it late and return.
			// This mimics the logic of `State.addVote()` method.
```

cason (commented):
While I don't really like this solution, I do agree it can have good consequences in terms of performance.

My suggestions would render this code as similar as possible as the code in `addVote()` method of `state.go`.

cason (changes requested):
We need to check a corner case, though.

If a vote is from the current height but from a previous round, we might still need it. We can use it for deciding a block, if its a Precommit, and for accepting a new Proposal, if it is a Prevote for the `PolRound`.

See test units `TestState_PrevotePOLFromPreviousRound` and `TestCommitFromPreviousRound`, for instance.

melekes:
> We need to check a corner case, though.
> 
> If a vote is from the current height but from a previous round, we might still need it. We can use it for deciding a block, if its a Precommit, and for accepting a new Proposal, if it is a Prevote for the `PolRound`.
> 
> See test units `TestState_PrevotePOLFromPreviousRound` and `TestCommitFromPreviousRound`, for instance.

@cason could you check this? Then, if we need these votes, we can close this PR. If not, merge this PR.

cason:
> @cason could you check this? Then, if we need these votes, we can close this PR. If not, merge this PR.

We would have to make a copy of those tests that run with a reactor, instead of manually providing messages to the consensus implementation. This is not so trivial, unfortunately.

But in general I am pretty convinced that we need in some cases messages from previous rounds, these tests are just some examples.

cason:
We cannot discard votes from previous rounds because the algorithm may require those votes to progress.

The problem here is that our test units test consensus `State` where we deliver the messages we want it to process. Adapting these tests to rely on a reactor, i.e., messages received from peers is not simple.

If you manage to adapt the two mentioned tests to run with a consensus `Reactor` and they do not fail I will be surprised but accept this change.

PR #3159, opened 2024-05-31
perf(consensus): Fix some peer messages taking write locks on the cs mtx

ValarDragon:
Some clear, read-only spots in receive routine took Write locks instead of read locks. Seems like a bug.

This PR fixes it.

---

#### PR checklist

- [x] Tests written/updated - N/A imo?


melekes (approved):

melekes:
@mergify backport v1.x

Issue #2127, opened 2024-01-25
consensus: extract block gossiping into its own reactor


melekes:
spawned by the discussion in https://github.com/cometbft/cometbft/issues/1742

Block gossiping is currently a big part of the consensus reactor, but it actually might not be that difficult to refactor it out, so the consensus reactor just has a `blockAvailable` signaling mechanism (like it has for transactions), and that thing has to be triggered before you can vote on a proposal. 

The thing that populates `blockAvailable` could be another reactor that propagates the block parts however it wants, including having most of it from the mempool.

cason:
Related to https://github.com/cometbft/cometbft/issues/14

cason:
We should discuss this topic, possibly synchronously.

I agree that block propagation should be a component by itself, maybe a new reactor, although this is not a requirement. This enables, in particular, some optimizations for consensus heights with multiple rounds (currently we drop the block from the previous round once a new round is started, even thought the same block can be re-proposed).

This component that handles block propagation should be informed, in some way, by consensus about the blocks it has to retrieve. One of the problems of a "blind" block propagation component is that it can be flooded with an arbitrary and possibly invalid number of blocks.

We also have the problem of persisting the blocks we accept (or propose). This is necessary, as otherwise, a node recovering from a crash will likely misbehave, since it does not "remember" actions taken before crashing.

faddat:
@cason 

So I've got to agree with you, I guess that it fits at all possible it's probably good to source what a block is directly from the proposer. However if we look at CAT, maybe there's a solution here.

The proposer could reference the transaction hashes and I think we might be able to reduce bandwidth quite a bit. Overall I'm still of the opinion that we are unlikely to be able to comprehensively fix the mempool, but....

I would like to be wrong about this because I do actually like having it and the other changes that are necessary will in my opinion require a great deal of work from infrastructure operators just to get into production.

ebuchman:
Copying over what I just wrote in #1776, which really belongs here. 

The block gossip is currently an integral part of the consensus reactor, but its really orthogonal to the core logic. In principle it could be separated out, even into its own reactor. This would clean up the consensus reactor and allow it to align even further with the core consensus specification, which would be nice.

However, there's a few complications to consider in the interface between the consensus reactor and such a new block gossip reactor.

[...]

PR #3161, opened 2024-05-31
perf(consensus): Make reactor check for duplicate/old block parts

### LINKED ISSUE #3158: Make late block parts not enter consensus peer message queue
### Summary

Same as #3154 , but for block parts. Here saving the WAL writes is important.

Osmosis blocks are most of the time 1-2 block parts, but we sometimes get spikes to much higher. 

Here is a graph of late block parts (green), and num_block_parts * 1000 (yellow). The *1000, was to make these visible on the same axes
image
Ignore the resets to 0 for the green, thats us resetting the node entirely.

So the late block parts seems to have odd trends, e.g. for 7 days we were at 17k parts per day, in a range hwere we only had 1-2 block parts. Sometimes its better, sometimes its worse. After the spike to 120 parts in one block, we had many ~5k late parts received!

Precisely when were getting all these late parts, is likely when its most important for us to save WAL load of writing these to disk, and times under mutex.

### Problem Definition

Progress to lowering our contention on the consensus mutex when under load.

### Proposal

Essentially same implementation as #3157 but for block parts

ValarDragon:
Closes #3158 

I made a common method between the reactor and consensus. In the future may add another argument to indicate if it should check the block part, but decided to not try that until I got vote signature checking into the reactor, as thats the bottleneck for Osmosis right now. (I suspect WAL is dominated from here though, will be straightforward to check once this is merged / in a patch release that I can A/B test)

The consensus state.go logic should be unchanged.

---

#### PR checklist

- [ ] Tests written/updated


internal/consensus/state.go
```diff
@@ -2084,34 +2084,60 @@ func (cs *State) defaultSetProposal(proposal *types.Proposal, recvTime time.Time
...
+
+	return true
+}
+
+// NOTE: block is not necessarily valid.
+// Asynchronously triggers either enterPrevote (before we timeout of propose) or tryFinalizeCommit,
+// once we have the full block.
+func (cs *State) addProposalBlockPart(msg *BlockPartMessage, peerID p2p.ID) (added bool, err error) {
+	height, round, part := msg.Height, msg.Round, msg.Part
+	// TODO: better handle block parts for future heights, by saving them and processing them later.
```
melekes, line 2133: do you mind opening an issue for this?

melekes (approved):
Thanks @ValarDragon ❤️

melekes (approved):

cason (changes requested):
We could change our mind regarding the Block we are looking for at any given instant.

We usually initialize `cs.ProposalBlockParts` upon receiving the first valid `Proposal` message for that round. But we might reset it upon receiving 2/3+ Prevotes or Precommits for a different block.

As a result, we might drop a block part that matches the new Block we are looking for, because we didn't change the block we are looking for because we are still processing votes that might made us to change our mind.

melekes:
@cason are you saying that because we can change our minds about `cs.ProposalBlockParts` at any moment, we should still write "invalid" block parts to WAL because between the time we write them and try to add them to `cs.ProposalBlockParts`, `cs.ProposalBlockParts` might have changed?

I'm in favor of this PR because, in most cases, we don't change our minds, and preventing duplicate parts from entering WAL as a perf optimization makes sense to me.

cason:
> @cason are you saying that because we can change our minds about `cs.ProposalBlockParts` at any moment, we should still write "invalid" block parts to WAL because between the time we write them and try to add them to `cs.ProposalBlockParts`, `cs.ProposalBlockParts` might have changed and "invalid" block part may "become valid"?
> 
> I'm in favor of this PR because, in most cases, we don't change our minds, and preventing duplicate parts from entering WAL as a perf optimization makes sense to me.

So, first, I don't think we should write block parts to the WAL at all. Except if we are using the full block. Namely, we either write the full block, or nothing. The best way to achieve this is to move block propagation outside of the consensus logic, which is the right direction to go.

At the moment, what can happen, and we might even write a test unit for that is:

- I am looking for block `X`, because we received a `Proposal` for `id(X)`
- But I am receiving votes (either type) for `id(Y) != id(X)`
- When I get 2/3+ of such votes, I will drop the block parts and part set of `X` and replace the block part set structure to accept block `Y`
- Meanwhile, I start receiving parts of block `Y`

So, if I receive a part of block `Y` at the reactor level but the votes that will make me change my mind are still on the queue to be processed, we are rejecting, basing on our current state, something we should accept, that we need by the way. Of course, if we don't mark that part as received, we should eventually get it again. But this is something to be tested, in my opinion.

melekes:
> Of course, if we don't mark that part as received, we should eventually get it again. But this is something to be tested, in my opinion.

@ValarDragon did you roll this out in production? Are you observing any slowness or blocks failing to commit as a consequence of this PR?

cason:
This should be done as part of this https://github.com/cometbft/cometbft/issues/2127

melekes:
> > Of course, if we don't mark that part as received, we should eventually get it again. But this is something to be tested, in my opinion.
> 
> @ValarDragon did you roll this out in production? Are you observing any slowness or blocks failing to commit as a consequence of this PR?

cc @ValarDragon

ValarDragon:
This has been in prod on Osmosis for ~2 months, no issues observed :)

cason:
We are probably not seen issues derived from this optimization in stable systems. This happens in a real corner case when we change the block we are looking at after receiving prevotes for that block, that does not match the proposed block. Which essentially requires Byzantine agents.

PR #3162, opened 2024-05-31
perf(consensus): Reuse an internal buffer for block building

ValarDragon:
Makes an internal buffer that we can re-use when building blocks. This seems to save on average 75 microseconds per block for osmosis blocks. (Ranging between 1-2 block parts in the relevant time range) This should be scaling roughly linear with block size.

This lets us remove one allocation cost per complete block coming in. We only need to re-allocate on the next "biggest ever seen" block we see. 

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/internal/consensus/state.go
+++ b/internal/consensus/state.go
@@ -137,6 +137,10 @@ type State struct {
 
 	// offline state sync height indicating to which height the node synced offline
 	offlineStateSyncHeight int64
+
+	// a buffer to store the concatenated proposal block parts (serialization format)
+	// should only be accessed under the cs.mtx lock
+	serializedBlockBuffer []byte
 }
 
 // StateOption sets an optional parameter on the State.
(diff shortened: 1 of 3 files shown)


internal/consensus/state.go
```diff
@@ -2084,6 +2088,23 @@ func (cs *State) defaultSetProposal(proposal *types.Proposal, recvTime time.Time
...
+	// reuse a serialized block buffer from cs
+	var serializedBlockBuffer []byte
+	if len(cs.serializedBlockBuffer) < int(cs.ProposalBlockParts.ByteSize()) {
+		serializedBlockBuffer = make([]byte, cs.ProposalBlockParts.ByteSize())
+		cs.serializedBlockBuffer = serializedBlockBuffer
+	} else {
+		serializedBlockBuffer = cs.serializedBlockBuffer[:cs.ProposalBlockParts.ByteSize()]
+	}
+
+	_, err := io.ReadFull(cs.ProposalBlockParts.GetReader(), serializedBlockBuffer)
```
melekes, line 2101: should we check that `n == len(serializedBlockBuffer)` just to be sure?

melekes (approved):

melekes:
@mergify backport v1.x

PR #3180, opened 2024-06-04
perf(consensus): Run broadcast routines out of process

ValarDragon:
Run broadcast routines out of process. Right now each broadcast routine blocks the consensus mutex for roughly `num_peers * process_creation_time`, which is genuinely notable! This PR reduces the consensus blocking overhead to just be `process_creation_time`. 

On the latest osmosis branch with improvements, thats 20s of blocking time out of 140s (over the course of 1 hour. This 140s includes block execution!) 
image

Note that WAL write time should go significantly down with open PR's. For `HasVote`, this is a meaningful increase to consensus mutex lock time, so its worth reducing.

---

#### PR checklist

- [x] Tests written/updated - I can't think of any test to add
- [x] Updated relevant documentation (`docs/` or `spec/`) and code comments - I don't know of any related docs here

### DIFF
--- a/.changelog/unreleased/improvements/3180-lower-broadcasts-consensus-overhead.md
+++ b/.changelog/unreleased/improvements/3180-lower-broadcasts-consensus-overhead.md
@@ -1,0 +1,2 @@
+- [`consensus`] Lower the consensus blocking overhead of broadcasts from `num_peers * process_creation_time` to `process_creation_time`.
+  ([\#3180](https://github.com/cometbft/cometbft/issues/3180)
--- a/internal/consensus/reactor.go
+++ b/internal/consensus/reactor.go
@@ -419,28 +419,28 @@ func (conR *Reactor) subscribeToBroadcastEvents() {
 	const subscriber = "consensus-reactor"
 	if err := conR.conS.evsw.AddListenerForEvent(subscriber, types.EventNewRoundStep,
 		func(data cmtevents.EventData) {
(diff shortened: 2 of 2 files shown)


melekes (approved):
Thanks @ValarDragon ❤️

cason:
So we are creating a Go routine to create `num_peers` Go routines to send a message to each peer? I am really impressed (and somehow suspicious) by the improvement this can provide.

ValarDragon:
Yes! This was surprisingly costly to me as well, but this is what the CPU profiles were saying! Post this change, this fireEvent time is ~entirely removed

cason:
Can't it be the contention of the for loop in the `Switch()` method? Because it has to retrieve the list of connected peers, and it is synchronized....

melekes:
> Can't it be the contention of the for loop in the `Switch()` method? Because it has to retrieve the list of connected peers, and it is synchronized....

I doubt that since Broadcast only further spawns `{N peers}` goroutines to `Send` an envelope https://github.com/cometbft/cometbft/blob/fd2a1abc59549a2cbbbb127f99682119a1cb25ee/p2p/switch.go#L276

ValarDragon:
No, this CPU profile is directly saying its runtime.NewProc, which is not based on mutex contentions

cason (approved):
Lets go with this.

I think the best solution is to improve the `Broadcast` implementation, but the benefits of this minimal change seem to be relevant.

melekes:
@mergify backport v1.x

Issue #3151, opened 2024-05-30
Change broadcastHasVote and broadcastHasBlockpart to use TrySend not Send


ValarDragon:
### Summary

Currently both BroadcastHasVote and BroadcastHasBlockpart use Broadcast, which does a Send message to every peer. 

While its not clear that BroadcastHasVote is needed, it definitely should be using TrySend not Send! TrySend makes it ignore the channel buffering capacity, and depend on time.Sleep's for cancelling. This increases the overhead for the go scheduler, which does actually take a lot of the time for the system. (Just creating the timers takes 30s of CPU time across a 1hour cpu profile, independent of the scheduler overhead)

Theres a number of things worth revisting about this `BroadcastHas*`, but it seems like we should be using TrySend everytime not Send. The channel capacity for hasVote is sent to 1, so were not really respecting that right now with the correct behavior. Its likely to me this is an accident in implementation, since the Broadcast method already existed, and used Send, with no corresponding trySend method.

### Problem Definition

Reduce our CPU and scheduling overheads, and respect the channel capacities, so we don't "overgossip" from `Has*` channels. 

### Proposal

For now, just make a BroadcastTrySend method, and switch the consensus gossip routines to using that!

If for some reason we really wanted Send (which doesn't make sense to me) then we should raise the channel capacities.

cason:
Yes, I partially agree.

The problem is that `TrySend` does not guarantee delivery to correct peers. If we me miss the Has* messages, ok, they are there to save redundancy. 

But the `NewRoundStepMessage` appears to me to be required, I am afraid that dropping them would have unpredictable outputs.

cason:
The easiest solution here is to clone `Switch.Broadcast()` to some method inside the consensus reactor that does the same while using `TrySend()` instead of `Send()`.

PR #3182, opened 2024-06-04
perf(p2p): Remove broadcast return channel

ValarDragon:
Remove the return channel and waitgroup from switch.Broadcast as its unused. This was costing us some CPU overhead (2% of broadcast routine as it stands right now, but broadcast routine time should be dropping) and one extra goroutine. This removes more overhead than what #3180 may introduce.

It also better highlights the point that maybe all broadcast usecases should consider using TrySend (And we potentially adapt channel buffers), ala the point in #3151 . (This would have significant system-wide impact, by significantly lowering the number of timers/scheduler overhead we have)

---

#### PR checklist

- [x] Tests written/updated - Simplifies code, the behavior being removed was never tested!


ValarDragon:
I recommend backporting this, as this is basically performance enhancing deadcode elimination

melekes (approved):
Thanks @ValarDragon ❤️

melekes:
> I recommend backporting this, as this is basically performance enhancing deadcode elimination

It's Go API breaking.

melekes:
@mergify backport v1.x

Issue #2840, opened 2024-04-18
Lower allocation overhead of metrics in peer.Send


ValarDragon:
### Summary

In Memory profiles of full nodes, we see large amounts of heap allocation in p2p.send coming from Prometheus logging, that is ~entirely duplicate work. 
![image](https://github.com/cometbft/cometbft/assets/6440154/b4547d5e-6d43-4465-a923-92169ec20c4e)

This is coming from these lines: https://github.com/cometbft/cometbft/blob/b8466398aacfc6f491ac54e741fce6750fa1cb9e/p2p/peer.go#L284-L292

We should be able to have the labels already cached (or at minimum the peer ID label cached) in the existing peer.Metrics instance.

### Problem Definition

Lower heap allocation and time overhead of sending something to peers. I have no idea how much this is a dominant term in wall clock time, though I do see recv time and send time routinely in CPU profiles.

### Proposal

Cache the peer ID label in p.metrics.PeerSendBytesTotal, so this is re-done on every packet send

cason:
There is a cache though: https://github.com/cometbft/cometbft/blob/d27732ad9a992e321e440417a2a8220f90fe0044/p2p/metrics.go#L43

cason:
But it doesn't apply to channel ID (which is immutable after startup) and to peer ID (this changes over time, so a cache would be good).

But in general, have per peer ID metrics looks an overkill to me.

ValarDragon:
That makes sense! I'm down for deleting per peer id metrics as well

cason:
> We should be able to have the labels already cached (or at minimum the peer ID label cached) 

I am wondering why we need to cache the peer ID. It is a string, so are you sure we are creating copies here?

ValarDragon:
Lets measure after that gets in! The memory overhead is coming from the prometheus `With` and `Add` calls I believe, so the creation of the PeerId as a string isn't the issue

cason:
Updated the PR. It is covering both send and receive side metrics.

ValarDragon:
I think we should re-open this issue. I think from the issue we were leaning towards just removing these peer metrics entirely. Or should I make a new issue for that?

cason:
Adding here the comment that indicates the next step to improve this overhead:

https://github.com/cometbft/cometbft/pull/2851#discussion_r1571202734

adizere:
Beside #3298, what is left to close this issue?

PR #3184, opened 2024-06-04
perf(p2p)!: Remove PeerSendBytesTotal metric

ValarDragon:
Component of #2840. As we discussed in the issue, we want to deprecate the peer send bytes total metric because:
- This is a costly metric (see pprof)
- This increases the cardinality of labels in Prometheus, which can slow things down in Prometheus significantly.
- Its not really useful at this level of abstraction. We don't have reported usecases of this being used yet. If we want to do it in the future, we should more likely architect it as updating a live go variable, and then sending to our metrics instance on a polling ticker (ala PeerSendQueue metric). The right levels that are interesting for outbound traffic could be the consensus gossip routines or blocksync.

To see its costly, 21% of all CPU time for live osmosis nodes in the latest Osmosis patch release is in peer.Send. over half of that time is in these metrics: 
image

Please advise for how we should go about deprecating / removing this metric? Feel free to close this PR / say you want to handle it separately.

---

#### PR checklist

- [X] Tests written/updated - N/A
- [x] Updated relevant documentation (`docs/` or `spec/`) and code comments - I didn't find any relevant docs


melekes (approved):
Thanks @ValarDragon ❤️

andynog (changes requested):
Thanks, the docs need to be updated too (p2p_peer_send_bytes_total)
https://github.com/cometbft/cometbft/blob/main/docs/explanation/core/metrics.md

andynog:
Concerning the fact that this is going to be removed and not available in the future, it would be interesting to ensure this is not used in our QA or Infra tests, I briefly looked at the logic there and couldn't find references to this particular metric, but @hvanz or @sergio-mena might have more insights. 

If we find out that some logic critically depends on this metric, probably a better approach would be to make it a configurable option if possible (e.g. disable it through a config value).

ValarDragon:
ah thanks, didn't think to search the docs with underscores! (Explains why I didn't find it earlier haha)

melekes:
> Thoughts on merging, and then reverting if we discover some metric consumer used this?

That's a bad dev UX 😅 

> If we find out that some logic critically depends on this metric, probably a better approach would be to make it a configurable option if possible (e.g. disable it through a config value).

please no more config options. IMO we have too many config options

melekes:
> If we find out that some logic critically depends on this metric

Even if some external software depends on this metric, it shouldn't prevent us from removing it. CometBFT's mission and what it should do best in the world is fast and reliable consensus. And if some metric is significantly slowing CometBFT down, we should eliminate it!

ValarDragon:
Ive tested removing it, and the prometheus time here went down ~70%! TBH that still justifies us removing the other metric or doing that more "asynchronously", e.g. via the `updateStatsRoutine`

melekes:
@mergify backport v1.x

PR #3201, opened 2024-06-06
fix(e2e): add TTL for txs in kvstore to remove txs stuck in mempool

hvanz:
The e2e test `TestApp_Tx` is sometimes failing on testnets with full nodes doing state sync. This is the scenario:
- Node A starts with state sync, receives a snapshot at height H, switches to consensus mode, and enables the mempool.
- Then node A receives from a peer a transaction that was included in a block at height lower than H. This transaction will get stuck in the mempool forever. 
- The reason is that in e2e transactions are always valid, so it cannot be removed from the mempool during rechecking. It can only be removed when committing a block that includes the transaction. But the transaction was already committed at height H before node A started.

This fix implements a TTL in the app side as a weak form of replay protection. Real applications should implement replay protections so they shouldn't have this problem.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/abci/example/kvstore/code.go
+++ b/abci/example/kvstore/code.go
@@ -7,4 +7,5 @@ const (
 	CodeTypeInvalidTxFormat uint32 = 2
 	CodeTypeUnauthorized    uint32 = 3
 	CodeTypeExecuted        uint32 = 5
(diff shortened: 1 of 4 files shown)


melekes (approved):
👍

PR #3204, opened 2024-06-07
fix(mempool/rpc): remove duplicate invokeCallback

hvanz:
On sync RPC endpoints `broadcast_tx_sync|commit`, it is wrong to call `reqRes.InvokeCallback()` because the callback was executed before. This PR removes these lines. 

The reason is that the callback was already executed, either inside the remote clients (socket or grpc) or when setting the callback on the local client.
- The socket client executes the callback explicitly [when receiving the response](https://github.com/cometbft/cometbft/blob/e731a3fdf9187fe07b2385463320669fcac569b6/abci/client/socket_client.go#L234C9-L234C15).
- The local client, on receiving the response, sets `reqRes.callbackInvoked = true`. Then it's the function `SetCallback` that actually executes the callback [outside of the client](https://github.com/cometbft/cometbft/blob/e731a3fdf9187fe07b2385463320669fcac569b6/mempool/clist_mempool.go#L287), when `CheckTxAsync` returns the reqRes object.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -158,12 +158,12 @@ cb(reqRes.Response.GetCheckTx())
 
 The `*abcicli.ReqRes` structure that `CheckTx` returns has a callback to 
 process the response already set (namely, the function `handleCheckTxResponse`).
-The callback can be invoked manually; for example:
+The callback will be invoked internally when the response is ready. We need only 
+to wait for it; for example:
 ```golang
 reqRes, err := CheckTx(tx, sender)
 // check `err` here
 reqRes.Wait()
-reqRes.InvokeCallback()
 ```
 
 ### Protobufs and Generated Go Code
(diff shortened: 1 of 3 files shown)


UPGRADING.md
```diff
@@ -158,12 +158,12 @@ cb(reqRes.Response.GetCheckTx())
 
 The `*abcicli.ReqRes` structure that `CheckTx` returns has a callback to 
 process the response already set (namely, the function `handleCheckTxResponse`).
-The callback can be invoked manually; for example:
+The callback will be invoked internally when the response is ready. We need only 
+to wait for it; for example:
 ```golang
 reqRes, err := CheckTx(tx, sender)
 // check `err` here
 reqRes.Wait()
```
hvanz, line 166: We only need to wait that the response is ready. When this happens, the callback will be executed on another place, as explained in the description.

andynog (approved):
changes looks good to me, ran tests locally and they passed.

Issue #3199, opened 2024-06-06
Make reactor.onReceive not block unmarshalling more packets from peer


ValarDragon:
### Summary

We want to be able to ingest packets from peers faster. There is some evidence (cc @evan-forbes) that we have blocking behavior on our ability to ingest data from peers. But as my explanation goes on, it should become clear based on intuition this is a problem.

Here is a profile from osmosis mainnet taken last night over one hour, from recvRoutine: 
![image](https://github.com/cometbft/cometbft/assets/6440154/17ef29cb-c9cb-49c1-994e-2015a5a6eb32)

p2p.createMConnection.func1 is `reactor.onReceive`.

The flow we currently have is:
```
recvRoutine:
- Checks if flowrate is satisfied for max packet size (1024 bytes)
- Tries to read the next packet
- Protobuf Decodes packet
- Find corresponding packets channel
- Buffer the proto-decoded packet data
- If buffer corresponds to a full logical packet
  - Run reactor.OnReceive
- Go back to beginning
```

The issue is reactor.OnReceive blocks reading and proto-decoding more data for _all channels_ to that peer. Some messages, e.g. IBC txs, take over 5ms. This means that if a peer gossips you an IBC tx, it will take at least 5ms before you even attempt to decode the subsequent packets they send you. (Leading to IBC enabled-DOS', amongst many others)

We see under low-load, CheckTx is a dominant item here.

Its clear we need to split these into different processes.

### Proposal

If we need to guarantee in-order delivery across each channel, then I think the answers are:
- Short term: Use a go channel within each channel, for buffering incoming packets and processing them in a different thread than the recv routine.
- Long term, change the channel.onReceive API

If we don't need to guarantee in order delivery, and just achieve it in ~almost every case, then we can just call this function in a goroutine: https://github.com/cometbft/cometbft/blob/main/p2p/conn/connection.go#L671

ValarDragon:
I propose a two-PR short-term fix, and then decide on a longer term API breaking fix later.

PR #1:

First notice that the key property is that we only need in-order delivery per peer<>channel pair, but we have no in-order delivery guarantees across channels. So we can achieve this by making every `onReceive` call in a goroutine https://github.com/cometbft/cometbft/blob/main/p2p/conn/connection.go#L671, and do a mutex per channel.

PR #2:

Make each of the main reactors have an incoming buffer, and a separate goroutine that proccesses the inbound. This would eventually be an API break on reactor, but short term should drop these incoming costs notably.

[...]

PR #3209, opened 2024-06-07
perf(p2p): Make ingesting packets not block on channel unmarshal or channel receive logic

ValarDragon:
Handles a significant part of #3199

This separates out the entire left hand side of into its own goroutine that won't block further reading:
image

That is unblocking us from reading/ingesting packets 73% of the CPU time on a very light period of Osmosis traffic, and likely far more real-time as much of this is mutex blocked! (Separate PR's will improve I/O usage, and other non-I/O terms)

We guarantee that every channel receives packets in-order from a peer.

Furthermore this will allow concurrency across channels for getting their own packets unmarshalled and sent to respective reactors. (Meaning a mempool tx gossip will now not block on a consensus packet receive, unless the per-channel unmarshal/recv buffer is full. Right now its parameterized to be low, but @evan-forbes has a PR that composes with this, for adding a buffer into every reactor, that should fix this from being remotely likely)

To me, I think the main questions for this PR being merged are:
- Are we fine w/ my API change of: `mconn._recover` to `mconn.PanicRecover` (publicly exporting it)? I do like where I ended up putting the panic recover. 
- Is it fine to put "setup channels fn" as a field element in peer? 
- Do we want any new behavior tested? I'll test this on mainnet osmosis. Due to matching the panic recovery path the old code, I don't think theres concern over that being handled differently.
- Any other tests we think are missing? Nothing comes to mind

I'll get this tested on mainnet osmosis soon, to help build confidence :)

---

#### PR checklist

- [ ] Tests written/updated - Some of the behavior here appears to never have been tested, so I we could introduce a test for sending an undecodable packet on a channel that does exist.

### DIFF
--- a/p2p/peer.go
+++ b/p2p/peer.go
@@ -217,6 +217,9 @@ func (p *peer) OnStart() error {
 // NOTE: it is not safe to call this method more than once.
 func (p *peer) FlushStop() {
 	p.mconn.FlushStop() // stop everything and close the conn
+	for i := 0; i < len(p.channels); i++ {
+		p.recvListenersQuitChan <- struct{}{}
+	}
 }
 
 // OnStop implements BaseService.
--- a/p2p/peer.go
+++ b/p2p/peer.go
@@ -213,13 +213,11 @@ func (p *peer) OnStart() error {
 
 // FlushStop mimics OnStop but additionally ensures that all successful
 // .Send() calls will get flushed before closing the connection.
+// Correct API usage requires FlushStop to terminate, and then OnStop getting called.
 //
 // NOTE: it is not safe to call this method more than once.
 func (p *peer) FlushStop() {
 	p.mconn.FlushStop() // stop everything and close the conn
-	for i := 0; i < len(p.channels); i++ {
-		p.recvListenersQuitChan <- struct{}{}
-	}
 }
 
 // OnStop implements BaseService.
@@ -230,6 +228,7 @@ func (p *peer) OnStop() {
 	for i := 0; i < len(p.channels); i++ {
 		p.recvListenersQuitChan <- struct{}{}
 	}
+	close(p.recvListenersQuitChan)
 }
 
 // ---------------------------------------------------
--- a/p2p/peer.go
+++ b/p2p/peer.go
@@ -199,7 +199,6 @@ func (p *peer) OnStart() error {
 		return err
 	}
 
-	defer p.mconn.PanicRecover()
 	p.recvListenersQuitChan = make(chan struct{}, len(p.channels))
 	p.setupRecvListeners()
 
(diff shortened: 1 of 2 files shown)


evan-forbes (commented):
we should totally not block on marshaling, that make a ton of sense to me! However we might be able to improve on this further. 

I could definitely be missing something, but it looks like we are breaking the previous guarantee where messages are processed in the order that they were sent by the peer. In most cases, I don't think this should have any effect. There could be some weird difficult to debug edge cases however. 

If we measure the time required to unmarshal messages and queue them in an envelop buffer, we can see that we have an average of around .03ms and a std of around .15ms. Doing the unmarshaling in parallel would be a nice bonus, but how much time do we expect to cut off?

msg unmarshaling time


With the other envelope buffering approach, we can see that in some circumstances we see those envelope buffer reaching over 1000 and frequently over 250, so I think if we don't want to block here, we should increase to beyond 250.

one slight modification that could be interesting would be to combine it with the previous attempt, so instead of only passing the envelopes in a buffer, we stick the msgBytes in the buffer. 

by default, it could look like [this](https://github.com/celestiaorg/celestia-core/commit/8e8a0da6a161bdd38d6b6754a520584608f11b95), but the neat thing about this approach is that we can leave parallelism to the reactor. instead of just unmarshaling in parallel, in reactors where its perfectly safe (such as the mempool), we can do all processing steps in parallel.

ValarDragon:
> I could definitely be missing something, but it looks like we are breaking the previous guarantee where messages are processed in the order that they were sent by the peer. In most cases, I don't think this should have any effect. There could be some weird difficult to debug edge cases however.

We aren't breaking this guarantee within a channel. We are only potentially breaking this guarantee across channels. Though your right this could cause complicated edge cases -- You may receive a vote for a block before the proposal msg, even though the peer sent both.

Thats a good point though, maybe we should do this per-reactor. 

> one slight modification that could be interesting would be to combine it with the previous attempt, so instead of only passing the envelopes in a buffer, we stick the msgBytes in the buffer.

I think thats an amazing idea. If we want to do this without breaking changes so we can ship faster, maybe the move is:
- Change my code to one thread per reactor
- Do logic where "If reactor supports async receive bytes, use that. Else do whats in this PR"

That way we can ship with no breakage, and always get the guarantee we need?

p2p/peer.go
```diff
@@ -211,6 +224,10 @@ func (p *peer) OnStop() {
 	if err := p.mconn.Stop(); err != nil { // stop everything and close the conn
 		p.Logger.Debug("Error while stopping peer", "err", err)
 	}
+	for i := 0; i < len(p.channels); i++ {
+		p.recvListenersQuitChan <- struct{}{}
```
melekes, line 228: Why do we need to write `len(channels)` messages? Isn't `close(chan)` won't result in all goroutines receiving a msg?

melekes (commented):
Thanks @ValarDragon ❤️

cason (commented):
I have conflicting feelings regarding this solution:

1. We should pay the cost of unmarshalling at some point, but is it fair do that channel by channel?
2. Aren't we artificially pretending that we are able to consume 256 messages per channel, while this is not true, clearly? How we impose back-pressure in this case?

Maybe the better solution here is quite simpler: have a thread for unmarshalling data, which is a costly part. Notice that in the original design were Reactors to do unmarshalling. Most of these performance penalties are result of unmarshalling/marshalling in the p2p layer, instead of in the Reactors.

PR #3211, opened 2024-06-08
perf(consensus): Remove reactor rlocks on Consensus

ValarDragon:
PR with @ebuchman !!!

Remove (most) of the reactor RLock's on consensus! (Remaining one is in the gossip routine, should be easy to fix, but as a separate PR)

The consensus reactor on ingesting messages takes RLock's on Consensus mutex, preventing us from adding things to the queue. (And therefore blocking behavior)

This can cause the peer message queue to be blocked. Now it won't be blocked because we update the round state directly from the cs state update routine (via event bus) when:
- A vote is added
- A block part is added
- We enter a new consensus step. (Receiving a full block triggers this for us, we will enter prevote or precommit round)

This shouldn't change reactor message validity, because Reactor.Receive could always view a cs.RoundState that will be old when the packet reaches the cs.HandleMsg queue. Every update to consensus' roundstate pushes the update into the reactor's view, so we avoid locks.

I don't think any new tests are required!

---

#### PR checklist

- [x] Tests written/updated


internal/consensus/reactor.go
```diff
@@ -267,9 +268,10 @@ func (conR *Reactor) Receive(e p2p.Envelope) {
 	case StateChannel:
 		switch msg := msg.(type) {
 		case *NewRoundStepMessage:
-			conR.conS.mtx.RLock()
-			initialHeight := conR.conS.state.InitialHeight
-			conR.conS.mtx.RUnlock()
+			// TODO: Set this up on load
```
ebuchman, line 271:
```suggestion
```

ebuchman (approved):

ValarDragon:
We profiled the number of mutex contentions before/after this change on Osmosis. This is the number of contentions on the recvRoutine over a 500s profile. (200 blocks)
image

We had 137000 contentions over 500s!

This became 0 contentions in the mconnection.RecvRoutine after this change :)

internal/consensus/state.go
```diff
@@ -255,6 +255,12 @@ func (cs *State) GetRoundState() *cstypes.RoundState {
 	return &rs
 }
 
+// getRoundState returns a shallow copy of the internal consensus state.
```
melekes, line 258: Having both `GetRoundState` and `getRoundState` with the same description is confusing. Can we add a comment to `GetRoundState` about why we need an RLock there and to `getRoundState` why we don't need an RLock there?

melekes (commented):
Thanks @ValarDragon ❤️

internal/consensus/state.go
melekes: Because `GetRoundState` also gets shallow copy, but with the lock.

cason:
I need to go through in more detail, but why not send this shallow copy with the synchronous message sent from state to the reactor?

This PR works (without locks) because the calls are synchronous, so nothing happens in the protocol while it is called.

internal/consensus/state.go
ValarDragon: Oh ok, can do. I thought the comet convention was that capitalized `Get`'s take the RLock, and lowercase don't

ValarDragon:
> I need to go through in more detail, but why not send this shallow copy with the synchronous message sent from state to the reactor?

I thought it was easier to handle copying/pointers to a struct in an independent `getRoundState` method (e.g. reasoning about shallow copy)

Or do you mean something different? Maybe I don't understand

cason:
> Or do you mean something different? Maybe I don't understand

I meant that adding the shallow copy of the consensus state to the `data` in the multiple `func(data cmtevents.EventData)` will have the same effect.

internal/consensus/state.go
cason: A solution here is to have a private `getRoundState()` method, with a comment saying that it is not thread safe, and a private `GetRoundState()` that acquires the lock and calls the former.

.changelog/unreleased/improvements/3211-make-cs-reactor-no-longer-takes-cs-locks.md
```diff
@@ -0,0 +1,4 @@
+- `[consensus]` Make the consensus reactor no longer have packets on receive take the consensus lock.
+Consensus will now update the reactor's view after every relevant change through the existing event
```
cason, line 2: This is not really an event bus.

I would say that we use the synchronous events produced by the state to the reactor to transmit a fresh/updated copy of the consensus state to the reactor.

internal/consensus/reactor.go
```diff
@@ -46,8 +46,9 @@ type Reactor struct {
 	waitSync atomic.Bool
 	eventBus *types.EventBus
 
-	rsMtx cmtsync.Mutex
-	rs    *cstypes.RoundState
+	rsMtx         cmtsync.RWMutex
+	rs            *cstypes.RoundState
+	initialHeight int64 // under rsMtx
```
cason, line 51: Why do we need this value for every received message? Not questioning the change, but the rationale.

cason:
Moreover, we still have the `func (conR *Reactor) updateRoundStateRoutine()` that does this shallow copy from the consensus state.

cason:
So, for context here.

The `Reactor` needs the consensus state, the infamous `rs` that is the way that the consensus logic "sends" information to the reactor (shared memory communication).

At the beginning of the times, each time we needed `rs` in the reactor, either in the `Receive()` method or in the multiple sending routines, we copied it from the consensus state (`conR.rs = conR.conS.GetRoundState()`). 

This of course was not efficient, so at some point we replaced this by creating a `updateRoundStateRoutine` infinite for loop that every 100 micros (I don't know why this granularity) does the same.

However, for some reason (also not 100% clear to me), in some points you have touched in this PR, in the `Receive()` method, we didn't trust in the `rs` copy (from 100 micros ago), but instead retrieved the current value.

From what I understood, you are proposing to still use a (potentially) outdated `rs` copy in the `Receive()` method but, instead of only updating it every 100 micros, also "forcedly" update it each time that the consensus logic produces a synchronous event to the reactor, the three that you mentioned and touched.

My general question here is: if we are pretty sure we don't need a fresh copy of `rs` on each `Receive()` call, because relevant changes on `rs` lead us to update our copy, why do we still need the `updateRoundStateRoutine()` loop?

cason:
In summary, I suggest we open an issue to understand better this synchronization/shared memory design and link the issues and PRs (like this one) that improves performance by saving useless lock acquisition and data copy.

ValarDragon:
> My general question here is: if we are pretty sure we don't need a fresh copy of rs on each Receive() call, because relevant changes on rs lead us to update our copy, why do we still need the updateRoundStateRoutine() loop?

Talked about this with Bucky IRL as well. Left it in within this PR, since its a nice fallback that may not be costing us. It would let us get this PR in, without risking any events updating the round state that we missed. I feel like we can remove that in a second PR, after getting this one tested on more robust networks. (As this PR should only help, it took us to 0 lock conflicts!)

cason (commented):
Left some general comments here.

I think this change is in the right direction. The description of the PR and, specially, the changelog entry are imprecise and misleading.

Once they are fixed or reworded, we are good to go.

.changelog/unreleased/improvements/3211-make-cs-reactor-no-longer-takes-cs-locks.md
ValarDragon: Ok! I thought this event system was called the event bus, thats my bad. Thank you!

internal/consensus/reactor.go
ValarDragon: We need it to be updated on every prune, since that can change our initial Height. I didn't see a synchronous point to update for prunes

cason: I think this is the first height of the blockchain, it never changes. In any case, the first block (height) stored is something we can get from the block store, not from the state.

ValarDragon:
> I meant that adding the shallow copy of the consensus state to the data in the multiple func(data cmtevents.EventData) will have the same effect.

That makes sense to me!

.changelog/unreleased/improvements/3211-make-cs-reactor-no-longer-takes-cs-locks.md
cason: No problem, it is described like that:

https://github.com/cometbft/cometbft/blob/2dd5209177270fde125c71aeedda9819046f92f0/internal/consensus/state.go#L131-L133

cason: And the comment is outdated, as we have a third event, eheheheheh

ValarDragon:
Anything needed for this PR to progress?

internal/consensus/state.go
```diff
@@ -248,6 +248,7 @@ func (cs *State) GetLastHeight() int64 {
 }
 
 // GetRoundState returns a shallow copy of the internal consensus state.
+// This function is thread-safe.
 func (cs *State) GetRoundState() *cstypes.RoundState {
 	cs.mtx.RLock()
 	rs := cs.RoundState // copy
```
cason, line 254:
```suggestion
	rs := cs.getRoundState()
```

cason: And change `&rs` to `rs` in the return of lines 255/256.

cason (approved):
Good.

The discussion, to have in another issue/PR, is whether we still, after this changes, have the `updateRoundStateRoutine()`, which does the same as this PR proposes.

internal/consensus/reactor.go
```diff
@@ -557,6 +560,14 @@ func (conR *Reactor) updateRoundStateRoutine() {
 	}
 }
 
+func (conR *Reactor) updateRoundStateNoCsLock() {
+	rs := conR.conS.getRoundState()
+	conR.rsMtx.Lock()
+	conR.rs = rs
+	conR.initialHeight = conR.conS.state.InitialHeight
```
melekes, line 567:
```suggestion
	conR.conS.mtx.RLock()
	conR.initialHeight = conR.conS.state.InitialHeight
	conR.conS.mtx.RULock()
```

how come we don't need a mutex to access state anymore?

melekes (changes requested):

internal/consensus/reactor.go
ValarDragon: This is in the `noCsLock` function!

melekes (approved):

Issue #2685, opened 2024-03-28
Make sure that reactors' implementation of `Receive()` method are non-blocking


cason:
Motivated by #2533.

The [`Reactor`](https://github.com/cometbft/cometbft/blob/main/p2p/base_reactor.go#L15) interface defines methods that a component must implement in order in order to use the communication services provided by the p2p layer. A core method that a reactor has to be implement is `Receive()`, invoked by the p2p layer to deliver a message to the reactor. The implementation of this method should be non-blocking (see details [here](https://github.com/cometbft/cometbft/blob/main/spec/p2p/reactor-api/reactor.md#receive-messages)).

If the implementation of `Receive()` in some reactor blocks when processing a message received from a given peer, the receive routine (`recvRoutine()`) of the p2p connection with that peer also blocks until the `Receive()` call returns:

https://github.com/cometbft/cometbft/blob/0147e634d51f5b27c60f914b02dad348e1ad96d9/p2p/conn/connection.go#L654-L655

This scenario has some consequences:

1. Messages received from the same peer are not handled by the p2p connection. This means that other reactors are affected, since they will not receive new messages until the blocking `Receive()` call returns;
2. As described in #2533, ping/pong messages, that play the role of protocol-level keepalive messages, are not handled either. As a result, the p2p connection pong timeout expires, then the connection produces an error and quits, which leads the switch to disconnect from that peer.

### Possible solutions

This situation could be avoided if the `Receive()` call was not executed by the receive routine of the p2p connection. This change, however, is far from trivial, given the design of the p2p multiples connections. Moreover, even if the `Receive()` call is executed in a different routine, this other routine will eventually block if the `Receive()` method of a reactor blocks for a long period of time.

The other approach, suggested in this issue, is to review the implementation of each reactor to make sure that the `Receive()` method is not blocking. This approach is also not trivial, as ultimately the way to prevent this is to buffer received messages that cannot be immediately processed. Since buffers are always finite in size, this would eventually lead to dropping messages.

evan-forbes:
one simple solution that we played around with was

https://github.com/celestiaorg/celestia-core/blob/146746b114624029432720cfa74e0a4a55a87b9b/p2p/base_reactor.go#L98-L105

the envelop buffers have to be quite large in order not to block for the consensus reactor, at least when blocks are large. As one might expect, most of messages that end up blocking are small messages, not larger messages such as block parts. The most common blocking channels are consensus state and votes, then block parts, at least from some recent experiments. 

[...]

PR #3230, opened 2024-06-10
feat: don't block on receives

evan-forbes:
Draft PR atm, since this is blocked by https://github.com/cometbft/cometbft/issues/3237

In collaboration with @ValarDragon, this PR combines the solution in https://github.com/cometbft/cometbft/pull/3209 with https://github.com/celestiaorg/celestia-core/commit/ab38bddcc57aaefe4049e861b6fc9a8c810553dc to fix https://github.com/cometbft/cometbft/issues/3199 and https://github.com/cometbft/cometbft/issues/2685

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/internal/evidence/reactor.go
+++ b/internal/evidence/reactor.go
@@ -39,7 +39,7 @@ func NewReactor(evpool *Pool) *Reactor {
 	evR := &Reactor{
 		evpool: evpool,
 	}
-	evR.BaseReactor = *p2p.NewBaseReactor("Evidence", evR)
+	evR.BaseReactor = *p2p.NewBaseReactor("Evidence", evR, p2p.WithIncomingQueueSize(1))
 	return evR
 }
 
--- a/internal/evidence/reactor.go
+++ b/internal/evidence/reactor.go
@@ -208,6 +208,10 @@ func (evR Reactor) prepareEvidenceMessage(
 	return []types.Evidence{ev}
 }
 
+func (evR *Reactor) OnStop() {
+	evR.BaseReactor.OnStop()
+}
+
 // PeerState describes the state of a peer.
 type PeerState interface {
 	GetHeight() int64
--- a/p2p/base_reactor.go
+++ b/p2p/base_reactor.go
@@ -1,8 +1,13 @@
 package p2p
 
 import (
+	"fmt"
+	"reflect"
+
 	"github.com/cometbft/cometbft/libs/service"
 	"github.com/cometbft/cometbft/p2p/conn"
+	"github.com/cometbft/cometbft/types"
+	"github.com/cosmos/gogoproto/proto"
 )
 
 // Reactor is responsible for handling incoming messages on one or more
@@ -41,6 +46,15 @@ type Reactor interface {
 	// Receive is called by the switch when an envelope is received from any connected
 	// peer on any of the channels registered by the reactor
 	Receive(e Envelope)
+
+	// QueueUnprocessedEnvelop is called by the switch when an unprocessed
+	// envelope is received. Unprocessed envelopes are immediately buffered in a
+	// queue to avoid blocking. Incoming messages are then passed to a
+	// processing function. The default processing function unmarshals the
+	// messages in the order the sender sent them and then calls Receive on the
+	// reactor. The queue size and the processing function can be changed by
+	// passing options to the base reactor.
+	QueueUnprocessedEnvelope(e UnprocessedEnvelope)
 }
 
 // --------------------------------------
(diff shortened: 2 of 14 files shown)


p2p/base_reactor.go
```diff
@@ -41,25 +46,151 @@ type Reactor interface {
...
+func ParallelProcessor(impl Reactor, threads int) func(<-chan UnprocessedEnvelope) error {
+	return func(incoming <-chan UnprocessedEnvelope) error {
+		for i := 0; i < threads; i++ {
+			go func() {
+				DefaultProcessor(impl)(incoming)
+			}()
+		}
+		return nil
+	}
+}
```
evan-forbes, line 192: tbh, I'm not sure if there's a ton of usecases for this atm. The thinking was that we could use this for reactors that don't require processing msgs in the original send order. Initially, I thought the mempool reactor would be a good candidate, and it might!, however it could result in more issues for users that are sending multiple txs within the same block for chains that use the default replay protection in the cosmos-sdk. 

there is a different flavor of this that could be where each channel is processed in order, but the others aren't. We could do this, but we might be better off just refactoring the entire p2p stack to use QUIC streams, since that builds on primitives designed for that.

p2p/base_reactor.go
```diff
@@ -41,25 +46,151 @@ type Reactor interface {
...
+		processor:   DefaultProcessor(impl),
+	}
+
+	for _, opt := range opts {
+		opt(base)
+	}
+
+	go func() {
+		base.processor(base.incoming)
+	}()
```
evan-forbes, line 89: ahh the reason for not calling `go base.processor(base.incoming)` was that I wasn't sure how to do error handling here. if we want to do the same thing that we are doing with https://github.com/cometbft/cometbft/pull/3209, then we need to stop the peer after catching the panic. 

we still need to fix this before merging. There are few different ways we could handle this, but I haven't found not super hacky way of doing this yet

cason:
See: https://github.com/cometbft/cometbft/issues/2685

p2p/base_reactor.go
```diff
@@ -41,25 +46,151 @@ type Reactor interface {
...
+		for _, chDesc := range implChannels {
+			chIDs[chDesc.ID] = chDesc.MessageType
+		}
+
+		var (
+			err error
+			msg proto.Message
+		)
+
+		for ue := range incoming {
```
evan-forbes, line 139: had some time to debug this tonight, and we're seeing a really weird issue. The messages are that are queued in this channel are unique, but they come out here, they aren't always :fearful: 

interesting... will continue to debug tmrw

evan-forbes:
blocked by https://github.com/cometbft/cometbft/issues/3237

p2p/base_reactor.go
```diff
@@ -41,25 +46,151 @@ type Reactor interface {
...
 
-func NewBaseReactor(name string, impl Reactor) *BaseReactor {
-	return &BaseReactor{
+type ReactorOptions func(*BaseReactor)
+
+func NewBaseReactor(name string, impl Reactor, opts ...ReactorOptions) *BaseReactor {
+	base := &BaseReactor{
 		BaseService: *service.NewBaseService(nil, name, impl),
 		Switch:      nil,
+		incoming:    make(chan UnprocessedEnvelope, 100),
```
melekes, line 79:
```suggestion
		incoming:    make(chan UnprocessedEnvelope, defaultUnprocessedQueueSize),
```

p2p/base_reactor.go
```diff
@@ -41,25 +46,151 @@ type Reactor interface {
...
+		processor:   DefaultProcessor(impl),
+	}
+
+	for _, opt := range opts {
+		opt(base)
+	}
+
+	go func() {
```
melekes, line 87: Shouldn't this be in `OnStart`?

p2p/base_reactor.go
```diff
@@ -41,25 +46,151 @@ type Reactor interface {
...
 	}
 }
 
 func (br *BaseReactor) SetSwitch(sw *Switch) {
 	br.Switch = sw
 }
+
+// QueueUnprocessedEnvelope is called by the switch when an unprocessed
+// envelope is received. Unprocessed envelopes are immediately buffered in a
+// queue to avoid blocking. The size of the queue can be changed by passing
```
melekes, line 117: But this will block if the queue is full, no?

internal/evidence/reactor.go
```diff
@@ -39,7 +39,7 @@ func NewReactor(evpool *Pool) *Reactor {
 	evR := &Reactor{
 		evpool: evpool,
 	}
-	evR.BaseReactor = *p2p.NewBaseReactor("Evidence", evR)
+	evR.BaseReactor = *p2p.NewBaseReactor("Evidence", evR, p2p.WithIncomingQueueSize(1))
```
melekes, line 42: How did you choose the sizes of the queues?

melekes (commented):
Nice work @evan-forbes ❤️

Issue #2719, opened 2024-04-04
bug: don't start `cometbft` if proxy_app is not specified or invalid


andynog:
### Setup

**CometBFT version**: latest on `main` (should be the same problem in all releases)

**Have you tried the latest version**: `yes`

**ABCI app** (name for built-in, URL for self-written if it's publicly available): `N/A`

**Environment**:
- **OS** (e.g. from /etc/os-release): `MacOS`
- **Install tools**:
- **Others**:

**node command runtime flags**: `cometbft start`

### Config

```
#######################################################################
###                   Main Base Config Options                      ###
#######################################################################

# TCP or UNIX socket address of the ABCI application,
# or the name of an ABCI application compiled in with the CometBFT binary
proxy_app = ""
```

### What happened?

Tried to run cometbft after removing the default `proxy_app` default address `tcp://127.0.0.1:26658` and without specifying the `--proxy_app` parameter. Also tried specifying an invalid address.

### What did you expect to happen?

CometBFT should abort the start since there's not a valid proxy application for it to connect to.

### How to reproduce it

**Empty parameter**
1.  Removed the `proxy_app` parameter from the config (leave it blank e.g. "")
2. Start comet: `cometbft start`

**Invalid parameter**
1.  Specified invalid address in `proxy_app` parameter in the config
2. Start comet: `cometbft start`

### Logs

Cometbft output messages that `tcp: missing address`. Probably should abort the start in this case in case the `proxy_app` is an empty string 

**Empty string `proxy_app` logs**:
```
 cometbft start                    
I[2024-04-04|18:13:41.994] State store key layout version               module=main version=vv1
I[2024-04-04|18:13:42.007] Blockstore version                           module=main version=v1
I[2024-04-04|18:13:42.007] WARNING: deleting genesis file from database if present, the database stores a hash of the original genesis file now module=main 
I[2024-04-04|18:13:42.007] service start                                module=proxy msg="Starting multiAppConn service" impl=multiAppConn
I[2024-04-04|18:13:42.007] service start                                module=abci-client connection=query msg="Starting socketClient service" impl=socketClient
E[2024-04-04|18:13:42.007] abci.socketClient failed to connect to .  Retrying after 3s... module=abci-client connection=query err="dial tcp: missing address"
E[2024-04-04|18:13:45.009] abci.socketClient failed to connect to .  Retrying after 3s... module=abci-client connection=query err="dial tcp: missing address"
[...]

PR #3232, opened 2024-06-10
feat(config): validate `proxy_app`

melekes:
Closes https://github.com/cometbft/cometbft/issues/2719

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -4,6 +4,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -324,6 +325,37 @@ func (cfg BaseConfig) ValidateBasic() error {
...
+		"noop":
+	default: // network address
+		parts := strings.SplitN(cfg.ProxyApp, "://", 2)
+		if len(parts) != 2 { // TCP address
+			_, err := net.ResolveTCPAddr("tcp", cfg.ProxyApp)
+			if err != nil {
+				return fmt.Errorf("failed to resolve TCP proxy_app %s: %w", cfg.ProxyApp, err)
+			}
+		} else {
+			switch parts[0] { // other protocol
```
andynog, line 348: here's the validation is only in the protocol (`parts[0]`) but would it be possible to have a validation on the address (`parts[1]`) ? at least from a well form perspective ?

config/config.go
```diff
@@ -324,6 +325,37 @@ func (cfg BaseConfig) ValidateBasic() error {
 	default:
 		return errors.New("unknown log_format (must be 'plain' or 'json')")
 	}
+
+	switch cfg.ProxyApp {
+	case "kvstore",
```
andynog, line 330: These values are the ones in `DefaultClientCreator` right ? Would be possible to have a more idiomatic `enum` way to check them? Like

type ProxyAppType int

const (
    KVStore ProxyAppType = iota
    KVStoreUnsync
    E2E
    ...
)

config/config_test.go
```diff
@@ -70,6 +70,15 @@ func TestBaseConfigValidateBasic(t *testing.T) {
 	require.Error(t, cfg.ValidateBasic())
 }
 
+func TestBaseConfigValidateBasic2(t *testing.T) {
```
andynog, line 73: Could we also test more examples here ? Like the `default` use case where we assume it's `tcp` ?

andynog (approved):
lgtm, added a few comments but those are just additional suggested enhancements.

cason:
I think we create the connection with the client in `DefaultClientCreator` from `proxy/client.go`.

Wouldn't make sense to add a kind of validation method there instead of checking this in the `config` package?

I always think in the case we change anything in the `proxy`, who will remember and do the super boring step of updating the `config` package as well?

config/config.go
cason: This can be a constant []string in any case, just check if it matches any of them.

config/config.go
```diff
@@ -324,6 +325,37 @@ func (cfg BaseConfig) ValidateBasic() error {
...
+		"kvstore_connsync",
+		"kvstore_unsync",
+		"persistent_kvstore",
+		"persistent_kvstore_connsync",
+		"persistent_kvstore_unsync",
+		"e2e",
+		"e2e_connsync",
+		"e2e_unsync",
+		"noop":
+	default: // network address
```
cason, line 340: I wonder if we don't have anything on this style inside the `p2p` package. We have to validate addresses there too...

melekes: 90% is done by the `net.Dial` function.

melekes:
> Wouldn't make sense to add a kind of validation method there instead of checking this in the config package?

This would defeat the purpose of this PR - validate proxy_app early. There are little to no checks in proxy package. net.Dial does all of them internally.

> I always think in the case we change anything in the proxy, who will remember and do the super boring step of updating the config package as well?

added comments in the code which might help

PR #3234, opened 2024-06-10
fix(e2e): reproduction an fixing of missing evidence in `e2e` nightlies 

### LINKED ISSUE #3233: `e2e` nightly fail sometimes because some injected evidence is lost
### Setup

**CometBFT version** (use `cometbft version` or `git rev-parse --verify HEAD` if installed from source): `v1.x`

e2e app

**Environment**:
N/A

N/A

### Config

According to nightly `e2e` manifest: networks/nightly/gen-group01-0004.toml

### What happened?

Some evidences are lost during nightly execution:

```
--- FAIL: TestEvidence_Misbehavior (0.00s)
    evidence_test.go:20: 
        	Error Trace:	/home/runner/work/cometbft/cometbft/test/e2e/tests/evidence_test.go:20
        	Error:      	Not equal: 
        	            	expected: 10
        	            	actual  : 4
        	Test:       	TestEvidence_Misbehavior
        	Messages:   	difference between the amount of evidence produced and committed
FAIL
FAIL	github.com/cometbft/cometbft/test/e2e/tests	23.159s
FAIL
```

No failure, all evidences accepted

### How to reproduce it

See first commit of PR fixing the problem

### Logs

See above

N/A

Root cause identified, fix exists, PR coming up

sergio-mena:
Closes #3233 

This PR is is to be interpreted commit by commit.

* Commit 1: Changes in `e2e runner to repro the problem (these changes will stay).
* Commit 2. Making the problem repro appear in this PR (look at the ❌ next to commit 2 below, and check the `e2e` logs)
* Commit 3: Fix. (look at the ✅ next to commit 3 below)
* Commit 4: Revert commit 2
* Commit 5: Adapt nightlies to start running all regression manifests in (newly introduced) `test/e2e/networks_regressions` directory. Also, adapt manifest generator to create testnets with more evidences.
* Further commits: fix e2e infra

The reproduction consists in coming up with a testnet where some evidences can be delayed in their transit to the next validator, long enough to become invalid (too old). The delay is achieved by having a validator with skewed clock, in such a way that every time it proposes (untimely proposal) other peers shut down their p2p connection with that validator. The validator having the skewed clock is the one receiving all evidences.

The root cause is that, in `e2e`, the max age of evidence is too restrictive (500ms in time, and 7 heights in terms blocks). So, if the gossiping of transactions is delayed, some of them are too old when they arrive to the first validator that can propose them.

The fix consists in relaxing the max age of evidences in `e2e`: 1500ms, or 14 heights.

Without the fix, the e2e run fails almost every time. With the fix, I haven't been able to repro it after multiple attempts.

---

#### PR checklist

- [x] Tests written/updated
- ~~[ ] Updated relevant documentation (`docs/` or `spec/`) and code comments~~

### DIFF
--- a/.github/workflows/e2e-manual-multiversion.yml
+++ b/.github/workflows/e2e-manual-multiversion.yml
@@ -34,12 +34,12 @@ jobs:
         # based on the current branch as compared to the latest version.
         run: ./build/generator -g 5 -m "latest:1,local:2" -d networks/nightly/ -p
 
-      - name: Run ${{ matrix.p2p }} p2p testnets
+      - name: Run p2p testnets (${{ matrix.group }})
         if: matrix.group != 5
         working-directory: test/e2e
         run: ./run-multiple.sh networks/nightly/*-group${{ matrix.group }}-*.toml
 
-      - name: Run ${{ matrix.p2p }} p2p regression testnets
+      - name: Run p2p testnets (regression)
         if: matrix.group == 5
         working-directory: test/e2e
         run: ./run-multiple.sh networks_regressions/*.toml
--- a/.github/workflows/e2e-manual-multiversion.yml
+++ b/.github/workflows/e2e-manual-multiversion.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        group: ['00', '01', '02', '03', '04']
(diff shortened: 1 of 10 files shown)


melekes (approved):
👍

hvanz (approved):

PR #3241, opened 2024-06-11
chore: correctly break from for loops

### LINKED ISSUE #2647: chore: investigate breaks from #2232
During #2232 we've identified multiple `break` points, which are not effective inside a `select` statement. https://github.com/cometbft/cometbft/pull/2232#discussion_r1522786397 

Let's verify if that was the intended behavior (as opposed to breaking from an enclosing `for` loop, for example).

melekes:
Closes #2647

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/test/e2e/runner/load.go
+++ b/test/e2e/runner/load.go
@@ -134,10 +134,12 @@ func createTxBatch(ctx context.Context, txCh chan<- types.Tx, testnet *e2e.Testn
 			}
 		}()
 	}
+FOR_LOOP:
 	for i := 0; i < testnet.LoadTxBatchSize; i++ {
 		select {
 		case genCh <- struct{}{}:
 		case <-ctx.Done():
+			break FOR_LOOP
 		}
 	}
 	close(genCh)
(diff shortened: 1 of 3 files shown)


test/e2e/runner/load.go
```diff
@@ -134,10 +134,12 @@ func createTxBatch(ctx context.Context, txCh chan<- types.Tx, testnet *e2e.Testn
 			}
 		}()
 	}
+FOR_LOOP:
 	for i := 0; i < testnet.LoadTxBatchSize; i++ {
 		select {
 		case genCh <- struct{}{}:
 		case <-ctx.Done():
+			break FOR_LOOP
```
ValarDragon, line 142: can we just do `break`?

melekes: `break` would only break from `select`, not the enclosing `for` loop.

cason (approved):
This is a safe-guard, no behavioral change expected.

But it would be nice to prove that the breaks were wrong, via test units, but maybe it is just too complex.

PR #3242, opened 2024-06-11
docs(mempool/lanes): Add technical design section to ADR

### LINKED ISSUE #3138: mempool QoS ADR: Writeup of detailed design
DoD: Complete the section Detailed Design in the ADR.

hvanz:
Closes #3138 

[Rendered](https://github.com/cometbft/cometbft/blob/hvanz/mempool-qos-technical-design/docs/references/architecture/adr-113-mempool-lanes.md)

---

#### PR checklist


docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -96,10 +97,11 @@ Before stating this property, we need to make some definitions.
 characteristics as defined by the application.
 
 A transaction may only have one class. If it is not assigned any specific class, it will be assigned
-a *default class*, which is a special class always present in any set of classes. Because no
-transaction can belong to two or more classes, transaction classes form disjoint sets, that is, the
-intersection between classes is empty. Also, all transactions in the mempool are the union of the
-transactions in all classes.
+a *default class*, which is a special class always present in any set of classes. This is analogous
+to the _native VLAN_ for untagged traffic in an 802.1Q network. Because no transaction can belong to
+two or more classes, transaction classes form disjoint sets, that is, the intersection between
```
sergio-mena, line 102:
```suggestion
two or more classes, transaction classes form disjoint sets, that is, the intersection between any two
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,285 @@ TODO
...
-  * So, at least in a first version, the lane info won't be shipped with TXs when broadcasting them
-  * TODO: we leave this bullet point here ATM, but we'll probably just deleted it
-* Before implementing lanes, is it better to first modularise the current mempool?
-  * we currently think improving modularisation is not gating for a mempool lanes MVP
-    * BUT, some things in the
-      _[modularity laundry list](https://www.notion.so/informalsystems/8a887f27e40b45dead689f2d1762b778?v=30a6fd81673b483fb16f4b7a9fc311fb)_
-      will be gating (e.g., `TxsAvailable`)
+This sections describes the architectural changes needed to implement an MVP of lanes in the
+mempool. The following is a summary of the key design decisions:
+- [[1](#lanes-definition)] The list of lanes and their corresponding priorities
```
sergio-mena, line 301:
```suggestion
- [[Lanes definition](#lanes-definition)] The list of lanes and their corresponding priorities
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,285 @@ TODO
...
+  will be hardcoded in the application logic.
+- [[2](#internal-data-structures)] There will be one concurrent list (CList) data structure per
```
sergio-mena, line 303:
```suggestion
- [[Internal data structures](#internal-data-structures)] There will be one concurrent list (CList) data structure per
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,285 @@ TODO
...
+  lane.
+- [[3](#configuration)] All lanes will share the same mempool configuration.
```
sergio-mena, line 305:
```suggestion
- [[Configuration](#configuration)] All lanes will share the same mempool configuration.
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,285 @@ TODO
...
+- [[4](#adding-transactions-to-the-mempool)] When validating a transaction via CheckTx, the
```
sergio-mena, line 306:
```suggestion
- [[Adding transactions to the mempool](#adding-transactions-to-the-mempool)] When validating a transaction via CheckTx, the
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,285 @@ TODO
...
+  application will optionally return a lane for the transaction.
+- [[5](#reaping-transactions-for-creating-blocks)] Transactions will be reaped from higher-priority
```
sergio-mena, line 308:
```suggestion
- [[Reaping transactions for creating blocks](#reaping-transactions-for-creating-blocks)] Transactions will be reaped from higher-priority
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,285 @@ TODO
...
+  lanes first, preserving the FIFO ordering.
+- [[6](#transaction-dissemination)] We will continue to use the current P2P channel for
```
sergio-mena, line 310:
```suggestion
- [[Transaction dissemination](#transaction-dissemination)] We will continue to use the current P2P channel for
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,285 @@ TODO
...
+mempool. The following is a summary of the key design decisions:
+- [[1](#lanes-definition)] The list of lanes and their corresponding priorities
+  will be hardcoded in the application logic.
+- [[2](#internal-data-structures)] There will be one concurrent list (CList) data structure per
+  lane.
+- [[3](#configuration)] All lanes will share the same mempool configuration.
+- [[4](#adding-transactions-to-the-mempool)] When validating a transaction via CheckTx, the
+  application will optionally return a lane for the transaction.
+- [[5](#reaping-transactions-for-creating-blocks)] Transactions will be reaped from higher-priority
+  lanes first, preserving the FIFO ordering.
```
sergio-mena, line 309:
```suggestion
  lanes first, preserving intra-lane FIFO ordering.
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,285 @@ TODO
...
-    * not trust info from peers, since we anyway need to run `CheckTx` to check the TX's validity
-      * so not trusting lane info from peers is aligned with our current model on checking validity of TXs received from peers
-  * So, at least in a first version, the lane info won't be shipped with TXs when broadcasting them
-  * TODO: we leave this bullet point here ATM, but we'll probably just deleted it
-* Before implementing lanes, is it better to first modularise the current mempool?
-  * we currently think improving modularisation is not gating for a mempool lanes MVP
-    * BUT, some things in the
-      _[modularity laundry list](https://www.notion.so/informalsystems/8a887f27e40b45dead689f2d1762b778?v=30a6fd81673b483fb16f4b7a9fc311fb)_
-      will be gating (e.g., `TxsAvailable`)
+This sections describes the architectural changes needed to implement an MVP of lanes in the
```
sergio-mena, line 299:
```suggestion
This section describes the architectural changes needed to implement an MVP of lanes in the
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,285 @@ TODO
...
+priority is a value of type `uint32`, with 0 being a reserved value (see below). The application
+also needs to define which of the lanes it defines is the default lane. 
+
+To obtain the lane information from the application, we need to extend the ABCI `Info` response to
+include the following fields. These fields need to be filled by the application only in case it
+wants to implement lanes.
+```protobuf
+message InfoResponse {
+  ...
+  repeated uint32 lanes = 6;
```
sergio-mena, line 326: `lane_priorities` maybe?

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,285 @@ TODO
...
+  uint32 default_lane = 7;
+}
+```
+Internally, the application may use `string`s to name lanes, and then map those names to priorities.
+The mempool does not care about the names, only about the priorities. That is why the lane
+information returned by the application only contains priorities.
+
+The lowest priority a lane may have is 1. The value 0 is reserved for two cases: when there are no
+priorities and for `CheckTx` responses of invalid transactions.
```
sergio-mena, line 335:
```suggestion
The lowest priority a lane may have is 1. The value 0 is reserved for two cases: when the application does not classify the transaction (i.e. no priority returned) and for `CheckTx` responses of invalid transactions.
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,285 @@ TODO
...
+   order in which to send transactions.
+
+### Lanes definition
+
+The list of lanes and their corresponding priorities will be hardcoded in the application logic. A
+priority is a value of type `uint32`, with 0 being a reserved value (see below). The application
+also needs to define which of the lanes it defines is the default lane. 
+
+To obtain the lane information from the application, we need to extend the ABCI `Info` response to
+include the following fields. These fields need to be filled by the application only in case it
```
sergio-mena, line 321:
```suggestion
include the following fields.
Remember CometBFT calls `Info` when the node is starting up (or recovering) as part of CometBFT handshake with the application.
These fields need to be filled by the application only in case it
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,285 @@ TODO
...
+Internally, the application may use `string`s to name lanes, and then map those names to priorities.
+The mempool does not care about the names, only about the priorities. That is why the lane
+information returned by the application only contains priorities.
+
+The lowest priority a lane may have is 1. The value 0 is reserved for two cases: when there are no
+priorities and for `CheckTx` responses of invalid transactions.
+
+On receiving the information from the app, CometBFT will validate that:
+- `lanes` has no duplicates (values in `lanes` don't need to be sorted),
+- `default_value` is in `lanes` (the default lane is not necessarily the lane with the lowest
```
sergio-mena, line 339:
```suggestion
- `default_lane` is in `lanes` (the default lane is not necessarily the lane with the lowest
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,285 @@ TODO
...
+```
+Internally, the application may use `string`s to name lanes, and then map those names to priorities.
+The mempool does not care about the names, only about the priorities. That is why the lane
+information returned by the application only contains priorities.
+
+The lowest priority a lane may have is 1. The value 0 is reserved for two cases: when there are no
+priorities and for `CheckTx` responses of invalid transactions.
+
+On receiving the information from the app, CometBFT will validate that:
+- `lanes` has no duplicates (values in `lanes` don't need to be sorted),
```
sergio-mena, line 338: Would it make our life easier to require that the lanes are sorted (to simplify the validation code)?

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,285 @@ TODO
...
+  ...
+  uint32 lane = 12;
+}
+```
+The callback that handles the first-time CheckTx response will append the new mempool entry to the
+corresponding `CList`, namely `lanes[lane]`, and update the other auxiliary variables accordingly.
+If `lane` is 0, it means that the application did not set any lane in the response message, so the
+transaction will be assigned to the default lane.
+
+### Removing transactions from the mempool
```
sergio-mena, line 424: Should this also be present in the list at the beginning of the section?

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,285 @@ TODO
...
+In this proposed approach, updating the lane definitions will require a single governance proposal
+for updating the software. CometBFT will not need to deal with dynamic lane changes: it will just
+need to set up the lanes when starting up (whether afresh or in recovery mode).
+
+Different nodes also need to agree on the lanes they use. When a node connects to a peer, they both
+perform a handshake to agree on some basic information (see `DefaultNodeInfo`), including the
+version of the application they are executing. Since the application includes the lane definitions,
+both nodes will also agree on the lanes they implement. Therefore we don't need to modify the P2P
+handshake.
+
```
sergio-mena, line 360: We could add a paragraph here to explain compatibility with CosmoVisor-like approaches.

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,285 @@ TODO
...
+// Fixed variables set during initialization.
+defaultLane Lane
+sortedLanes []Lane // Lanes sorted by priority
+```
+The auxiliary fields `txsMap` and `txLanes` are, respectively, for direct access to the mempool
+entries, and for direct access to the lane of a given transaction.
+
+If the application does not implement lanes (that is, it responds with empty values in
+`InfoResponse`), then `defaultLane` will be set to 1, and `lanes` will have only one entry for the
+default lane.
```
sergio-mena, line 390:
```suggestion
default lane. In this case, the new mempool's behaviour will be equivalent to that of the current mempool.
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,285 @@ TODO
...
+
+### Configuration
+
+For an MVP, all lanes can share the same mempool configuration. In this scenario, all lanes will be
+capped by the total mempool capacities as currently defined in the configuration. Namely, these are:
+- `Size`, the total number of transactions allowed in the mempool, 
+- `MaxTxsBytes`, the maximum total number of bytes of the mempool, and
+- `MaxTxBytes`, the maximum size in bytes of a single transaction accepted into the mempool.
+
+With lanes, the total size of the mempool will be the sum of the sizes of all lanes.
```
sergio-mena, line 403: Do we want to divide the total size evenly across lanes? (or are we gonna expunge a TX from a lower lane when a full mempool receives a TX?)

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,285 @@ TODO
...
+
+Additionally, the `Recheck` and `Broadcast` flags will apply to all lanes or to none. Remember that,
+with `PrepareProposal`, it becomes _always_ mandatory to recheck remaining transactions in the
+mempool, so there is no point in disabling `Recheck` per lane.
```
sergio-mena, line 407:
```suggestion
if `PrepareProposal`'s app logic can ever add a new transaction, it becomes _always_ mandatory to recheck remaining transactions in the
mempool, so there is no point in disabling `Recheck` per lane.
```

docs/references/architecture/adr-113-mempool-lanes.md
hvanz: Not really. It's easy to sort the list and we wouldn't need to trust the app to do it.

docs/references/architecture/adr-113-mempool-lanes.md
hvanz: Yes. That's important and I forgot.

docs/references/architecture/adr-113-mempool-lanes.md
hvanz: I'd say it's not needed for the MVP but it could be an interesting improvement for the future. Equal sizes for all lanes may not make sense when lanes are defined by transaction size. For example, a lane for only very big txs. Probably we won't see this issue when we add custom configurations per lane.

docs/references/architecture/adr-113-mempool-lanes.md
hvanz: I didn't add it because there's nothing important about removing transactions.

docs/references/architecture/adr-113-mempool-lanes.md
sergio-mena: Will add a paragraph

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,284 @@ TODO
...
+We choose the second option for its flexibility, allowing us to start with a simple scheduling
+algorithm that can be refined over time. The desired algorithm must satisfy the properties
+"Priorities between classes" and "FIFO ordering per class". This requires supporting selection by
+_weight_, ensuring each lane gets a fraction of the P2P channel capacity proportional to its
+priority. Moreover, it should be _fair_, to prevent starvation of low-priority lanes. Given the
+extensive research on scheduling algorithms in operating systems and networking, we propose for the
+MVP implementing a variant of the [Weighted Round Robin (WRR)](wrr) algorithm, which meets these
+requirements and is straightforward to implement.
+
+Before modifying the dessemination logic, we need to refactor the current implementation and the
```
sergio-mena, line 471:
```suggestion
Before modifying the dissemination logic, we need to refactor the current implementation and the
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,284 @@ TODO
...
+just wrappers around the methods `WaitChan` and `Front` of the `CList` implementation. In
+particular, `TxsFront` is leaking implementation details outside the `Mempool` interface.
+
+### Checking received transactions
+
+Transactions are transmitted without lane information because peers cannot be trusted to send the
+correct data. A node may take advantage of the network by sending lower-priority transactions before
+higher-priority ones. Although the receiving node could easily verify the priority of a transaction
+when it calls `CheckTx`, it cannot detect if a peer is sending transactions out of order over a
+single P2P channel. For the moment, we leave out of the MVP any mechanism to detect and possible
```
sergio-mena, line 484:
```suggestion
single P2P channel. For the moment, we leave out of the MVP any mechanism to detect and possibly
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,284 @@ TODO
...
+Nodes would also need to agree on the channel assignments during the P2P handshake. Currently, one
+of the conditions for the handshake to succeed is that there must exist an intersection of P2P
+channels. Since lanes are defined in the application logic, the nodes only need to agree on the
+application version, as it already happens in the current implementation.
+
+### Duality lane/priority
+
+The duality lane/priority could introduce a powerful indirection. The app could just define the lane
+of a transaction in `CheckTx`, but the priority of the lane itself could be configured (and
+fine-tuned) elsewhere. For example, by the app itself or by node operators. The proposed design does
```
sergio-mena, line 519:
```suggestion
fine-tuned) elsewhere. For example, by the app itself or by node operators. The proposed design for the MVP does
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,284 @@ TODO
...
+A straightforward, future improvement that we leave for after the MVP is to allow customized
+configuration of the lanes instead of sharing the current mempool configuration among lanes. The
+application would need to define new configuration values per lane and pass them to CometBFT during
+the handshake.
+
+### Where to define lanes and priorities
+
+We have considered two alternative approaches for _where_ to configure lanes and priorities:
+1. In `config.toml` or `app.toml`. We have discarded this option as it does not make sense for
+   different nodes to have different lane configurations.
```
sergio-mena, line 533:
```suggestion
   different nodes to have different lane configurations. The properties defined in the specification above
   are end-to-end, and so, the lane configuration has to be consistent across the network.
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,284 @@ TODO
...
+1. In `ConsensusParams`. There are several disadvantages with this approach. If we allow changing
+   lane information via `ConsensusParams`, the mempool would need to update lanes dynamically. The
+   updating process would be very complex and cumbersome, and not really appealing for an MVP. Two
+   governance proposals would be required to pass to update the lane definitions. A first proposal
```
sergio-mena, line 537:
```suggestion
   governance proposals would be required to pass to update (a) the lane definitions, and (b) a new version of the app with adapted classification logic.
   A first proposal
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,284 @@ TODO
...
+   would be required for upgrading the application, because the lane classification logic (thus the
+   application's code) needs to know the lane configuration beforehand. And a second proposal would
+   be needed for upgrading the lanes via `ConsensusParams`.
```
sergio-mena, line 540:
```suggestion
   be needed for upgrading the lanes via `ConsensusParams`.
   While it is true that SDK applications could pass a governance proposal with both elements together,
   it would be something to _always_ do, and it is not clear what the situation would be for non-SDK applications.
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,284 @@ TODO
...
 ### Positive
 
+- Application developers will be able to better predict when transactions will be disseminated and
+  processed to be included in a block. This has direct impact on block creation and transaction
+  latency.
+- The mempool will be able to offer Quality of Service guarantees, which does not exist in the
+  current implementation. 
+- Lanes will preserve the "FIFO ordering of transactions" property within the same class (with a
+  best effort approach, as the current implementation).
+
```
sergio-mena, line 563:
```suggestion
- Applications that are unaware of this feature, and therefore not classifying transactions in `CheckTX`
  will observe the same behavior from the mempool as the current implementation.  
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,284 @@ TODO
...
 ### Negative
 
+- The total FIFO ordering of transaction dissemination and block creation may be broken when
+  multiple lanes are used. Since FIFO ordering is important within the same class of transactions,
+  we expect this will not be a real problem.
+
```
sergio-mena, line 569:
```suggestion
- Increased complexity in the logic of `CheckTx` (ante handlers) in order to classify transactions,
  with a possibility of introducing bugs in the classification logic.
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,284 @@ TODO
...
 ### Neutral
 
+- Lanes are optional. Current applications do not need to make any change to their code. Future
+  applications will not be forced to use the lanes feature.
+- The proposed prioritization logic (WRR) for the dissemination algorithm is fair, so low-priority
+  transactions will not get stuck in the mempool long periods of time.
```
sergio-mena, line 575:
```suggestion
  transactions will not get stuck in the mempool for long periods of time.
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,284 @@ TODO
...
+   allows for an unlimited number of lanes, constrained only by the nodes’ capacity to store the
+   lane data structures.
+
+We choose the second option for its flexibility, allowing us to start with a simple scheduling
+algorithm that can be refined over time. The desired algorithm must satisfy the properties
+"Priorities between classes" and "FIFO ordering per class". This requires supporting selection by
+_weight_, ensuring each lane gets a fraction of the P2P channel capacity proportional to its
+priority. Moreover, it should be _fair_, to prevent starvation of low-priority lanes. Given the
+extensive research on scheduling algorithms in operating systems and networking, we propose for the
+MVP implementing a variant of the [Weighted Round Robin (WRR)](wrr) algorithm, which meets these
```
sergio-mena, line 468: In QoS for IP networks there is [WRED](https://en.wikipedia.org/wiki/Weighted_random_early_detection). I need check if that would be relevant here.

melekes (approved):
Did a cursory review. Great work 🚀

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,289 @@ TODO
...
+Since the broadcast goroutines are constantly reading the list of transactions to disseminate them,
+it's important to prioritize the removal of transactions from high-priority lanes first.
+
+When updating the mempool, there is potential for a slight optimization by removing transactions
+from different lanes in parallel. To achieve this, we would first need to preprocess the list of
+transactions to determine the lane of each transaction. However, this optimization has minimal
+impact if the committed block contains few transactions. Therefore, we decided to exclude it from
+the MVP.
+
+### Reaping transactions for creating blocks
```
cason, line 439: Any discussion regarding starvation?

Namely, no guarantee that a tx in a low priority lane will eventually be included in the output?

I mentioned that to you guys at some point, and I still think that the way the `p2p/MConnection` does that is probably more adequate. I think that considering this approach should at least be discussed here, as it is an ADR.

docs/references/architecture/adr-113-mempool-lanes.md
cason: The link is broken.

cason: Again, is this what the p2p channel does?

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,289 @@ TODO
...
+higher-priority ones. Although the receiving node could easily verify the priority of a transaction
+when it calls `CheckTx`, it cannot detect if a peer is sending transactions out of order over a
+single P2P channel. For the moment, we leave out of the MVP any mechanism to detect and possibly
+punish nodes for this kind of behaviour.
+
+## Alternative designs
+
+### One CList for all lanes
+
+We briefly considered sharing one CList for all lanes, changing the internal logic of CList to
```
cason, line 491: This is the priority mempool, in other words.

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,289 @@ TODO
...
+accommodate lane requirements. However, this design significantly increases code complexity,
+particularly in the transaction dissemination logic.
+
+### One P2P channel per lane
```
cason, line 495: Also, we would need a different routine per lane/channel, otherwise there is no real difference between the approaches.

Also consider that the assignment of channels is **static**.

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,289 @@ TODO
...
+could reserve a range of channels for the mempool, such as starting from channel ID `0x80` and above
+(all channels with the most significant byte set to 1). This would provide a maximum of 128 lanes,
+which should suffice for most users.
+
+Nodes would also need to agree on the channel assignments during the P2P handshake. Currently, one
+of the conditions for the handshake to succeed is that there must exist an intersection of P2P
+channels. Since lanes are defined in the application logic, the nodes only need to agree on the
+application version, as it already happens in the current implementation.
+
+### Duality lane/priority
```
cason, line 515: To confirm, now lineID == priority? And the higher the greater?

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,289 @@ TODO
...
+   not to break performance between the passing of both proposals. The `gov` module could be
+   modified to allow the two changes to be shipped in the same gov proposal, but this does not seem
+   a feasible solution.
+
+Moreover, these two alternatives have a common problem which is how to deal with nodes that are
+late, possibly having lane definitions that do not match with those of nodes at the latest heights.
 
 ## Consequences
 
 TODO
```
cason, line 555: Remove?

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,289 @@ TODO
...
 
 ### Positive
 
+- Application developers will be able to better predict when transactions will be disseminated and
+  processed to be included in a block. This has direct impact on block creation and transaction
+  latency.
+- The mempool will be able to offer Quality of Service guarantees, which does not exist in the
+  current implementation. 
+- Lanes will preserve the "FIFO ordering of transactions" property within the same class (with a
```
cason, line 564: This is neutral to me.

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,289 @@ TODO
...
+  best effort approach, as the current implementation).
+- Applications that are unaware of this feature, and therefore not classifying transactions in `CheckTX`
+  will observe the same behavior from the mempool as the current implementation.  
 ### Negative
 
+- The total FIFO ordering of transaction dissemination and block creation may be broken when
```
cason, line 570: Total FIFO? Eheheh

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,289 @@ TODO
...
+- Application developers will be able to better predict when transactions will be disseminated and
+  processed to be included in a block. This has direct impact on block creation and transaction
+  latency.
+- The mempool will be able to offer Quality of Service guarantees, which does not exist in the
+  current implementation. 
+- Lanes will preserve the "FIFO ordering of transactions" property within the same class (with a
+  best effort approach, as the current implementation).
+- Applications that are unaware of this feature, and therefore not classifying transactions in `CheckTX`
+  will observe the same behavior from the mempool as the current implementation.  
 ### Negative
```
cason, line 568: On negative, again, are we discussing the starvation problem of low priority lanes. I refer to the `Reap()` part, not the broadcast part, which seems to be implemented using a starvation-free algorithm.

cason:
> The first part of this PR is clearly a spec. 

And the use of FIFO order, partial order there is something to review. Lets come up with new names, lets having a subsection about ordering, please, because this is a source of confusion.

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,289 @@ TODO
...
+optionally return its lane in the response.
+```protobuf
+message CheckTxResponse {
+  ...
+  uint32 lane = 12;
+}
+```
+The callback that handles the first-time CheckTx response will append the new mempool entry to the
+corresponding `CList`, namely `lanes[lane]`, and update the other auxiliary variables accordingly.
+If `lane` is 0, it means that the application did not set any lane in the response message, so the
```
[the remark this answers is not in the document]
hvanz: The application implicitly returns 0 when it ignores lanes in the CheckTx response. We cannot force the application to return a specific value when they don't want to implement lanes, so the mempool needs to take care of assigning the default lane.

docs/references/architecture/adr-113-mempool-lanes.md
hvanz: That's right.

docs/references/architecture/adr-113-mempool-lanes.md
hvanz: Not what I meant :stuck_out_tongue_closed_eyes:. I rephrased.

docs/references/architecture/adr-113-mempool-lanes.md
hvanz: The Reap section did not mention the starvation-free property. I corrected it.

docs/references/architecture/adr-113-mempool-lanes.md
hvanz: Changed to neutral.

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,289 @@ TODO
...
+If the application does not implement lanes (that is, it responds with empty values in
+`InfoResponse`), then `defaultLane` will be set to 1, and `lanes` will have only one entry for the
+default lane. In this case, the new mempool's behaviour will be equivalent to that of the current mempool.
+
+`CListMempool` also contains the cache, which is only needed before transactions have a lane
+assigned. Since the cache is independent of the lanes, we do not need to modify it.
+
+### Configuration
+
+For an MVP, all lanes can share the same mempool configuration. In this scenario, all lanes will be
```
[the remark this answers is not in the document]
hvanz: You're right. I rephrased it.

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,289 @@ TODO
...
+
+On receiving the information from the app, CometBFT will validate that:
+- `lanes` has no duplicates (values in `lanes` don't need to be sorted),
+- `default_lane` is in `lanes` (the default lane is not necessarily the lane with the lowest
+  priority), and
+- the list `lanes` is empty if and only if `default_value` is 0.
+
+### Initialization
+
+Currently, querying the app for `Info` happens during the handshake between CometBFT and the app,
```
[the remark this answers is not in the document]
hvanz: Not specifically this scenario but section "Validating lanes of received transactions" is about what to do when a node decides to broadcast txs in whatever order.

docs/references/architecture/adr-113-mempool-lanes.md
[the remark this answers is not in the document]
hvanz: Agree, this is not clear or correct. I rephrased it.

docs/references/architecture/adr-113-mempool-lanes.md
hvanz: It was not mentioned so I added it. In essence it's the same as for tx dissemination. I agree we should discuss what P2P does and compare it to the proposed scheduling algorithm.

docs/references/architecture/adr-113-mempool-lanes.md
hvanz: I wouldn't say so because in the priority mempool every tx has a priority. The idea of the "one CList for all lanes" solution was something like adding a lane field to each entry and then other auxiliary data structures to support traversing the entries per lane.

docs/references/architecture/adr-113-mempool-lanes.md
sergio-mena: > The link is broken.

Which link do you refer to? The one on the text, or the one in my comment (both seem to be working for me)

> Again, is this what the p2p channel does?

Yeah, I had the same question when I read this part of Thu. Is priority in p2p channels implementing `wrr` or something else?

docs/references/architecture/adr-113-mempool-lanes.md
sergio-mena: Working on it now...

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,285 @@ TODO
...
+The lowest priority a lane may have is 1. The value 0 is reserved for two cases: when there are no
+priorities and for `CheckTx` responses of invalid transactions.
+
+On receiving the information from the app, CometBFT will validate that:
+- `lanes` has no duplicates (values in `lanes` don't need to be sorted),
+- `default_value` is in `lanes` (the default lane is not necessarily the lane with the lowest
+  priority), and
+- the list `lanes` is empty if and only if `default_value` is 0.
+
+### Initialization
```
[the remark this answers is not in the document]
sergio-mena: Taking care of this...

docs/references/architecture/adr-113-mempool-lanes.md
sergio-mena: Done. Please take a look

docs/references/architecture/adr-113-mempool-lanes.md
hvanz: I added a section Prioritization logic that includes the current scheduling algorithm in the P2P reactor.

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -481,19 +478,42 @@ particular, `TxsFront` is leaking implementation details outside the `Mempool` i
...
+which we could easily reimplement in the mempool. It works as follows:
+- On each P2P channel, the variable `recentlySent` keeps track of how many bytes were recently sent
+  over the channel. Every time data is sent, increase `recentlySent` with the number of bytes
+  written to the channel. Every 2 seconds, decrease `recentlySent` by 20% on all channels (these
+  values are fixed).
+- When sending the next message, [pick the channel][selectChannelToGossipOn] whose ratio
+  `recentlySent/Priority` is the least.
+
+From the extensive research in operating systems and networking, we can choose for the MVP an
+existing scheduling algorithm that meet these requirements and are straightforward to implement,
```
sergio-mena, line 514:
```suggestion
existing scheduling algorithm that meets these requirements and are straightforward to implement,
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -481,19 +478,42 @@ particular, `TxsFront` is leaking implementation details outside the `Mempool` i
...
+such as a variant of the [Weighted Round Robin][wrr] (WRR) algorithm. Another one to consider is
+[Weighted Random Early Detection][wred].
```
sergio-mena, line 516:
```suggestion
such as a variant of the [Weighted Round Robin][wrr] (WRR) algorithm.
```

I was wrong w.r.t. WRED, it's rather used to help congestion control algorithms (in the presence of QoS), so not really related here.

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -610,7 +630,9 @@ late, possibly having lane definitions that do not match with those of nodes at
 - Cosmos SDK's [gas prices][sdk-gas-prices]
 - Cosmos SDK's [application-side mempool][sdk-app-mempool]
 - Skip's [Block SDK][skip-block-sdk]
+- P2P's [selectChannelToGossipOn][selectChannelToGossipOn] function
 - [Weighted Round Robin][wrr]
+- [Weighted Random Early Detection][wred]
```
sergio-mena, line 635:
```suggestion
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -621,5 +643,7 @@ late, possibly having lane definitions that do not match with those of nodes at
 [sdk-gas-prices]: https://docs.cosmos.network/v0.50/learn/beginner/tx-lifecycle#gas-and-fees
 [sdk-app-mempool]: https://docs.cosmos.network/v0.47/build/building-apps/app-mempool
 [skip-block-sdk]: https://github.com/skip-mev/block-sdk/blob/v2.1.3/README.md
+[cosmovisor]: https://docs.cosmos.network/v0.50/build/tooling/cosmovisor
+[selectChannelToGossipOn]: https://github.com/cometbft/cometbft/blob/6d3ff343c2d5a06e7522344d1a4e17d24ce982ad/p2p/conn/connection.go#L542-L563
 [wrr]: https://en.wikipedia.org/wiki/Weighted_round_robin
-[cosmovisor]: https://docs.cosmos.network/v0.50/build/tooling/cosmovisor
\ No newline at end of file
+[wred]: https://en.wikipedia.org/wiki/Weighted_random_early_detection
```
sergio-mena, line 649:
```suggestion
```

docs/references/architecture/adr-113-mempool-lanes.md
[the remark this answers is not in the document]
sergio-mena: Done

docs/references/architecture/adr-113-mempool-lanes.md
hvanz: Changed to `lane_priorities` and `default_lane_priority`.

docs/references/architecture/adr-113-mempool-lanes.md
hvanz: Thanks!

docs/references/architecture/adr-113-mempool-lanes.md
hvanz: The link on WRR was broken and then I fixed it. I added a section called Prioritization Logic with a description of the algorithm in P2P channels.

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,298 @@ TODO
...
+sync CometBFT with the app. The lane information is needed regardless of whether state sync is
+enabled, so one option is to query the app information outside of the `Handshaker`.
+
+In this proposed approach, updating the lane definitions will require a single governance proposal
+for updating the software. CometBFT will not need to deal with dynamic lane changes: it will just
+need to set up the lanes when starting up (whether afresh or in recovery mode).
+
+Different nodes also need to agree on the lanes they use. When a node connects to a peer, they both
+perform a handshake to agree on some basic information (see `DefaultNodeInfo`), including the
+version of the application they are executing. Since the application includes the lane definitions,
```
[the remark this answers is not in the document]
hvanz: Oops I assumed wrongly. I changed this paragraph proposing to "modify the P2P handshake process to validate that the application versions are compatible". Does this change break something?

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -294,254 +296,289 @@ TODO
...
+
+The list of lanes and their corresponding priorities will be hardcoded in the application logic. A
+priority is a value of type `uint32`, with 0 being a reserved value (see below). The application
+also needs to define which of the lanes it defines is the default lane. 
+
+To obtain the lane information from the application, we need to extend the ABCI `Info` response to
+include the following fields. These fields need to be filled by the application only in case it
+wants to implement lanes.
+```protobuf
+message InfoResponse {
```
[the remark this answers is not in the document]
cason: The one from lines 327 to 331.

By the conditions on lines 342 and 343.

docs/references/architecture/adr-113-mempool-lanes.md
cason: So it is not one clist for all lanes, but one clist and multiple clists, one per lane, using the same entries as the one big clist.

In other words, a linked-list that contains other abstract linked-lists inside it.

docs/references/architecture/adr-113-mempool-lanes.md
cason: I assume this is written somewhere...

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -300,44 +300,49 @@ This section describes the architectural changes needed to implement an MVP of l
...
+also needs to define which of the lanes in the list it defines is the **default lane**. 
 
 To obtain the lane information from the application, we need to extend the ABCI `Info` response to
 include the following fields. These fields need to be filled by the application only in case it
 wants to implement lanes.
 ```protobuf
 message InfoResponse {
   ...
-  repeated uint32 lanes = 6;
-  uint32 default_lane = 7;
```
cason: Is it correct to set the default lane to 7 if it is not in the list of lanes?

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -300,44 +300,49 @@ This section describes the architectural changes needed to implement an MVP of l
...
+application does not classify the transaction (i.e. no priority returned) and for `CheckTx`
+responses of invalid transactions.
 
 On receiving the information from the app, CometBFT will validate that:
-- `lanes` has no duplicates (values in `lanes` don't need to be sorted),
-- `default_lane` is in `lanes` (the default lane is not necessarily the lane with the lowest
-  priority), and
-- the list `lanes` is empty if and only if `default_value` is 0.
+- `lane_priorities` has no duplicates (values in `lane_priorities` don't need to be sorted),
+- `default_lane_priority` is in `lane_priorities` (the default lane is not necessarily the lane with
```
cason, line 343: This is violated in the example above.

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -352,10 +357,22 @@ for updating the software. CometBFT will not need to deal with dynamic lane chan
...
-handshake.
+perform a handshake to agree on some basic information (see `DefaultNodeInfo`). Since the
+application includes the lane definitions, it suffices to ensure that both nodes agree on the
+version of the application. Although the application version is included in `DefaultNodeInfo`, there
+is currently no check for compatibility between the versions. To address this, we would need to
+modify the P2P handshake process to validate that the application versions are compatible.
+
+Finally, this approach is compatible with applications that need to swap binaries when
+catching up or upgrading, such as SDK applications using [Cosmovisor][cosmovisor].
+When a node is catching up (i.e. BlockSyncing), its peers will detect
```
cason, line 368:
```suggestion
When a node is catching up (i.e., state or block syncing), its peers will detect
```

cason (approved):
Reviewing the changes from my previous review, most questions were addressed, great work :)

docs/references/architecture/adr-113-mempool-lanes.md
hvanz: `6` and `7` are not default values. They are the field number. Currently `InfoResponse` has 5 fields.

docs/references/architecture/adr-113-mempool-lanes.md
hvanz: Same answer as above.

spec/mempool/cache.md
```diff
@@ -0,0 +1,36 @@
+# Mempool cache
+
+The *cache* stores the hashes of recently seen transactions. Its purpose is to avoid processing and
```
cason, line 3: Does it only store it nowadays?

spec/mempool/cache.md
```diff
@@ -0,0 +1,36 @@
...
+adding duplicate transactions to the mempool.
```
cason, line 4: Also broadcasting them again.

spec/mempool/cache.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Adding transactions to the cache
+
+A node that receives a transaction, before trying to add it to the mempool, puts it in the cache, if
```
cason, line 8: I tend to understand "receives a transaction" via network. Also via RPC, i.e., from clients?

cason: I would rephrase in a simpler way:

1. New transaction, check if it is on the cache
2. If it is, does nothing (what error message is produced?)
3. If it is not, add it to the cache to then start the process of adding it to the mempool.

spec/mempool/cache.md
```diff
@@ -0,0 +1,36 @@
...
+it is not already there. Only if it is not in the cache, the node will attempt to add it to the
+mempool. If the transaction exists in the cache, it is discarded with an error message.
+
+Another way a transaction may enter the cache is when the node's execution module updates the
```
cason, line 12: This is another use we have from the cache.

We want to avoid keeping in the mempool transactions that were already committed. Because the mempool is a list of _pending_ or _oustanding_ transactions.

This should be clear also in the introduction.

cason: Again, to rephrase in a simpler way:

1. We want to make sure that we don't process transactions that were already committed
2. So we keep on the cache (hashes) of recently committed transactions
3. Upon the update call, if some committed and valid transaction is not yet present in the cache, we add it to the cache.

spec/mempool/cache.md
```diff
@@ -0,0 +1,36 @@
...
+mempool. A valid transaction that was included in a block by another node may have never been seen
+by the current node, so we added to the cache.
+
+:memo: _Definition_: We say that a node receives a transaction `tx` _for the first time_ when the
+node receives `tx` and `tx` is not in the cache.
+
+Because of how transactions are evicted from the cache (see below), it is possible by this
+definition that a transaction is received for the first time on multiple occasions.
```
cason, line 20: This sentence.... first time multiple times?

spec/mempool/cache.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Capacity
+
+The cache has a *capacity*, a maximum number of transactions that it can hold. The cache's capacity
```
cason, line 24: This mixes definitions with recommendations?

spec/mempool/cache.md
```diff
@@ -0,0 +1,36 @@
...
+MUST be bigger than the mempool's capacity (in number of transactions). For example, in the current
+implementation, the cache doubles the capacity of the mempool. 
+
+The cache enforces a FIFO policy for keeping transactions: a transaction will be evicted from the
```
cason, line 28: I thought it was a LRU cache. If it is was, there is not FIFO.

spec/mempool/cache.md
```diff
@@ -0,0 +1,36 @@
...
+cache if it is the cache's oldest entry and a new transaction is being added. Therefore, a
```
cason, line 29: Define oldest, namely, the cache eviction: last added, last accessed, last modified?

cason: This can be helpful: https://en.wikipedia.org/wiki/Cache_replacement_policies

spec/mempool/cache.md
```diff
@@ -0,0 +1,36 @@
...
+transaction stays in the cache longer than it stays in the mempool.
+
+## Removing transactions from the cache
+
+A transaction is removed from the cache also when the application deems the transaction as invalid.
```
cason, line 34: So, the assumption is that the cache keeps _valid_ transactions only? Maybe rephrasing in this way helps the reader?

spec/mempool/cache.md
```diff
@@ -0,0 +1,36 @@
...
+Because transactions are evicted when the cache is full, a transaction could be removed from the
```
cason, line 35: This has nothing to do with being invalid or not.

Plus, I thought this behavior was configurable, with a config flag.

cason (commented):
A new spec file popped up, some (several) comments regarding it.

Maybe a different PR would be a better approach.

hvanz:
@cason Thanks for all the feedback. I removed the cache spec; it does not belong here and I will leave it for another PR. For now, I simplified the explanation of the "received for the first time" definition, which was the reason to include the cache spec.

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -53,32 +55,43 @@ is an orthogonal problem to transaction classification, although the latter may
...
-happen that `tx2` is sent to a peer before `tx1` is sent to a different peer.
+By this definition, it is possible that a node receives a transaction "for the first time", then
+gets the transaction evicted from the cache, and at a later time receives it "for the first time"
+again. The cache implements a Least-Recently Used (LRU) policy for removing entries when the
+cache is full.
+
+:memo: _Definition_: Given any two different transactions `tx1` and `tx2`, in a given node, we say that:
+1. `tx1` is *validated before* `tx2`, when `tx1` and `tx2` are received for the first time, and `tx1`
+is validated against the application (via `CheckTx`) before `tx2`,
+1. `tx1` is *rechecked before* `tx2`, when `tx` and `tx2` are in the mempool and `tx1` is
```
sergio-mena, line 74:
```suggestion
1. `tx1` is *rechecked before* `tx2`, when `tx1` and `tx2` are in the mempool and `tx1` is
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -53,32 +55,43 @@ is an orthogonal problem to transaction classification, although the latter may
...
+re-validated (rechecked via `CheckTx`) before `tx2`,
```
sergio-mena, line 75: What if `tx1` and `tx2` are re-checked in different heights? Does that case fit into this definition, or we're leaving such a case out on purpose?

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -53,32 +55,43 @@ is an orthogonal problem to transaction classification, although the latter may
...
+1. `tx1` is *reaped before* `tx2`, when `tx1` is reaped from the mempool to be included in a block
+  proposal before `tx2`,
+1. `tx1` is *disseminated before* `tx2`, when `tx1` is sent to a given peer before `tx2`.
+
+In 4, note that in the current implementation there is one dissemination routine per peer, so it
+could happen that `tx2` is sent to a peer before `tx1` is sent to a different peer.
```
sergio-mena, line 81:
```suggestion
could happen that `tx2` is sent to a peer before `tx1` is sent to a different peer.
Hence the importance of expression "to a given peer" in that definition.
```

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -53,32 +55,43 @@ is an orthogonal problem to transaction classification, although the latter may
...
 
 ### Current mempool
 
 As stated above, the current mempool offers a best-effort FIFO ordering of transactions. We state
 this property as follows.
 
 :parking: _Property_ **FIFO ordering of transactions**: We say that the mempool makes a best effort
-in maintaining the FIFO ordering of transactions when transactions are validated, processed, and
-disseminated in the same order in which the mempool has received them.
+in maintaining the FIFO ordering of transactions when transactions are validated, rechecked, reaped,
```
sergio-mena, line 89: Excellent!

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -96,27 +109,32 @@ Before stating this property, we need to make some definitions.
...
+a *default class*, which is a special class always present in any set of classes. This is analogous
+to the _native VLAN_ for untagged traffic in an 802.1Q network. Because no transaction can belong to
+two or more classes, transaction classes form disjoint sets, that is, the intersection between any two
+classes is empty. Also, all transactions in the mempool are the union of the transactions in all
+classes.
 
 :memo: _Definition_: Each class has a *priority* and two classes cannot have the same priority.
 Therefore all classes can be ordered by priority.
 
+When a transaction is received for the first time and validated via `CheckTx`, the application MUST
```
sergio-mena, line 121: But, we say somewhere that, if the add doesn't return a class, then the TX belongs to the default class?

docs/references/architecture/adr-113-mempool-lanes.md
```diff
@@ -126,22 +144,22 @@ timestamp of the block in which the transaction finally was included.
...
 
 As a consequence, given that classes of transactions have a sequential ordering, and that classes do
 not have elements in common, we can state the following property:
 
 :parking: _Property_ **Partial ordering of all transactions**: The set of all the transactions in
 the mempool, regardless of their classes, will have a *partial order*.
 
-This means that some pairs of
-transactions are comparable and, thus, have and order, while others not.
+This means that some pairs of transactions are comparable and, thus, have and order, while others
```
sergio-mena, line 161:
```suggestion
This means that some pairs of transactions are comparable and, thus, have an order, while others
```

sergio-mena (approved):
🚀

docs/references/architecture/adr-113-mempool-lanes.md
hvanz: If both transactions are in the mempool, that means that they are rechecked at the same height. I'll clarify in the text.

docs/references/architecture/adr-113-mempool-lanes.md
hvanz: True. I fixed it.

PR #3246, opened 2024-06-12
fix(privval): eliminate dead code

### LINKED ISSUE #2598: privval: double check error handling logic
From @mzabaluev:

> I wonder what this was supposed to check, because in the current code it's never true: https://github.com/cometbft/cometbft/blame/44ffd6769076c1a2a6e2c66382808139e36abe50/privval/signer_requestHandler.go#L45

> It seems there are two levels of returning errors from the handler: the Go-conventional error tuple member is maybe meant for internal errors, and for "protocol-legit" errors there's wrapping an error into the response. Which one is more appropriate for this case?

> This commit opted for the Go error reporting, but left or copied over a protocol-wrapping error handling branch that became dead code: [https://github.com/cometbft/cometbft/commit/f6243d8b9ec5dde8b0fe951d3726d77520ea1195#diff-33451325467f4a3ca89fcb1b5[…]904ecf9a875953e48d7214134ec8f0](https://github.com/cometbft/cometbft/commit/f6243d8b9ec5dde8b0fe951d3726d77520ea1195#diff-33451325467f4a3ca89fcb1b580ae7485c904ecf9a875953e48d7214134ec8f0)

> This looks sketchy as well: the error is logged, but then the potentially nil response is attempted to be written and err is lost.
https://github.com/cometbft/cometbft/blob/44ffd6769076c1a2a6e2c66382808139e36abe50/privval/signer_server.go#L80

melekes:
Closes #2598

---

#### PR checklist

### DIFF
--- a/privval/signer_requestHandler.go
+++ b/privval/signer_requestHandler.go
@@ -22,20 +22,12 @@ func DefaultValidationRequestHandler(
 	switch r := req.Sum.(type) {
 	case *pvproto.Message_PubKeyRequest:
 		if r.PubKeyRequest.GetChainId() != chainID {
-			res = mustWrapMsg(&pvproto.PubKeyResponse{
-				PubKeyType: "", PubKeyBytes: []byte{}, Error: &pvproto.RemoteSignerError{
-					Code: 0, Description: "unable to provide pubkey type and bytes",
-				},
-			})
-			return res, fmt.Errorf("want chainID: %s, got chainID: %s", r.PubKeyRequest.GetChainId(), chainID)
+			return chainIDMismatchError(r.PubKeyRequest.GetChainId(), chainID)
 		}
 
 		var pubKey crypto.PubKey
-		pubKey, err = privVal.GetPubKey()
-		if err != nil {
-			return res, err
-		}
 
+		pubKey, err = privVal.GetPubKey()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


privval/signer_requestHandler.go
```diff
@@ -22,20 +22,12 @@ func DefaultValidationRequestHandler(
...
-					Code: 0, Description: "unable to provide pubkey type and bytes",
-				},
-			})
-			return res, fmt.Errorf("want chainID: %s, got chainID: %s", r.PubKeyRequest.GetChainId(), chainID)
+			return chainIDMismatchError(r.PubKeyRequest.GetChainId(), chainID)
 		}
 
 		var pubKey crypto.PubKey
-		pubKey, err = privVal.GetPubKey()
-		if err != nil {
```
melekes, line 35: My assumption is that this `if` condition was introduced by mistake. The correct response would be to return a wrapped error below.

privval/signer_requestHandler.go
```diff
@@ -22,20 +22,12 @@ func DefaultValidationRequestHandler(
...
-			return res, err
-		}
 
+		pubKey, err = privVal.GetPubKey()
 		if err != nil {
```
melekes, line 31: This was dead code.

ValarDragon (approved):

PR #3248, opened 2024-06-12
feat(crypto)!: remove unnecessary `Sha256` wrapper

melekes:
We are essentially removing a wrapper that took the [32]byte value produced by the standard library and transforming it into a []byte slice.

---

#### PR checklist


melekes:
@andynog fixed it, thanks!

ValarDragon (approved):

crypto/xsalsa20symmetric/symmetric_test.go
```diff
@@ -27,7 +26,9 @@ func TestSimpleWithKDF(t *testing.T) {
 	if err != nil {
 		t.Error(err)
 	}
-	secret = crypto.Sha256(secret)
+	sum := sha256.Sum256(secret)
+	copy(secret, sum[:])
```
cason, line 30: Just for curiosity, if you don't copy here, the test fails?

crypto/doc.go
```diff
@@ -1,42 +1,29 @@
 // crypto is a customized/convenience cryptography package for supporting
```
cason, line 1:
```suggestion
// crypto is a customized/convenience cryptography package for CometBFT.
```

crypto/doc.go
```diff
@@ -1,42 +1,29 @@
...
 // CometBFT.
```
cason, line 2:
```suggestion

```

crypto/doc.go
cason, line 1: We could maybe also document the `PubKey` interface here.

crypto/hash.go
cason, line 1: Essentially, this wrapper does transformation of array in slice?

cason (approved):
Good.

Maybe just to document in the PR/final commit that we are essentially removing a wrapper that took the `[32]byte` value produced by the standard library and transforming it into a `[]byte` slice.

crypto/hash.go
melekes: Yes

Issue #2424, opened 2024-02-23
crypto: provide simple way to add new curves 


tac0turtle:
### Summary

currently there is not a way to add new curves to comet unless you fork the software. This causes many forks for simple changes.

### Problem Definition

adding new curves forces users to fork the software. (unless im missing something)

### Proposal

add simple way to add new curves to be used for consensus key signing

Zondax is proposing an api to help with this in the sdk. we could look at hosting the new api in its own module to enable usage in comet as well

tac0turtle:
to add on it would be nice to have the file system privval also take arbitrary curves. This way users dont also have to fork privval file once pluggable curves land

cason:
I think we would need a method such as `RegisterKey` at some place.

Now one can implement the `crypto.PrivKey` and `crypto.PrivKey` interfaces, but not sure how to set the new key for use by Comet.

tac0turtle:
i think the pubkey proto type may need some changing as the current design wouldnt allow register key without redoing the proto types

melekes:
refs https://github.com/cometbft/cometbft/pull/2399

PR #3249, opened 2024-06-12
docs(adr-116): modular crypto

melekes:
Refs #2424

---

#### PR checklist

### DIFF
--- a/docs/references/architecture/adr-116-modular-crypto.md
+++ b/docs/references/architecture/adr-116-modular-crypto.md
@@ -1,0 +1,46 @@
+# ADR 116: Modular crypto
+
+## Changelog
+
+ - June 12 2024: Created by @melekes
+
+## Status
+
+**Proposed**
+
+## Context
+
+Currently there's no way to add a new cryptographic curve (used for consensus
+signing) to CometBFT without forking the codebase. Maintaining a fork is a
+significant burden and makes it hard to keep up with the latest changes in the
+main codebase.
+
+On the other hand, **adding a new curve to the main codebase is a significant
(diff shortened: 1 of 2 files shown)


zmanian:
Many of the advantages of alternate cryptography require more invasive changes to the code base and even the header structure.

For instance to get advantages from BLS-12-381, you also need to be able use aggregated public keys in the commit certificates.

There are wins like a more EVM verification friendly comet that uses keccack and ethsign for signatures that can maybe be done without a hard fork.

docs/references/architecture/adr-116-modular-crypto.md
```diff
@@ -0,0 +1,68 @@
...
+    GenPrivKey() PrivKey
+}
+
+type SignatureScheme uint
+
+const (
+	Ed25519 SignatureScheme = 1 + iota
+	Secp256k1
+	Sr25519
+	Bls12381 // import github.com/cometbft/bls12381
```
sergio-mena, line 42: I'd rather use the _KeyType_ (a string) that every crypto package must define, and which is supposed to be unique. Examples [here](https://github.com/cometbft/cometbft/blob/10bfd89d/crypto/secp256k1/secp256k1.go#L24), [here](https://github.com/cometbft/cometbft/blob/10bfd89d/crypto/ed25519/ed25519.go#L62), or [here](https://github.com/cometbft/cometbft/blob/d27732ad/crypto/bls12381/const.go#L11).

sergio-mena (commented):
Although still very preliminary, I support this approach

melekes:
I will reopen once this work is restarted.

PR #3256, opened 2024-06-13
refactor: align tx hash debug information in mempool logging

mmsqe:
---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -267,7 +267,7 @@ func (mem *CListMempool) CheckTx(tx types.Tx, sender p2p.ID) (*abcicli.ReqRes, e
 				memTx := elem.Value.(*mempoolTx)
 				if found := memTx.addSender(sender); found {
 					// It should not be possible to receive twice a tx from the same sender.
-					mem.logger.Error("tx already received from peer", "tx", tx.Hash(), "sender", sender)
+					mem.logger.Error("tx already received from peer", "tx", fmt.Sprintf("%X", tx.Hash()), "sender", sender)
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


mempool/clist_mempool.go
```diff
@@ -267,7 +267,7 @@ func (mem *CListMempool) CheckTx(tx types.Tx, sender p2p.ID) (*abcicli.ReqRes, e
 				memTx := elem.Value.(*mempoolTx)
 				if found := memTx.addSender(sender); found {
 					// It should not be possible to receive twice a tx from the same sender.
-					mem.logger.Error("tx already received from peer", "tx", tx.Hash(), "sender", sender)
+					mem.logger.Error("tx already received from peer", "tx", fmt.Sprintf("%X", tx.Hash()), "sender", sender)
```
hvanz, line 270: Here `log.NewLazySprintf` is better than `fmt.Sprintf`. This also applies to many of the other changes below. Also, `tx.Hash()` is just `tx.Key()[:]`. `Key` is called four lines above.

mempool/clist_mempool.go
```diff
@@ -263,11 +263,12 @@ func (mem *CListMempool) CheckTx(tx types.Tx, sender p2p.ID) (*abcicli.ReqRes, e
...
 			// (eg. after committing a block, txs are removed from mempool but not cache),
 			// so we only record the sender for txs still in the mempool.
-			if elem, ok := mem.getCElement(tx.Key()); ok {
+			txKey := tx.Key()
+			if elem, ok := mem.getCElement(txKey); ok {
 				memTx := elem.Value.(*mempoolTx)
 				if found := memTx.addSender(sender); found {
 					// It should not be possible to receive twice a tx from the same sender.
-					mem.logger.Error("tx already received from peer", "tx", tx.Hash(), "sender", sender)
+					mem.logger.Error("tx already received from peer", "tx", log.NewLazySprintf("%X", txKey), "sender", sender)
```
hvanz, line 271: Thanks for the update. In the logs we should use `Hash` because it is displayed different than `Key`, and the key is internal while the hash is for the users, even when they are equivalent.

melekes:
Thanks for your work @mmsqe ❤️ I'm going to close this in favor of the upcoming https://github.com/cometbft/cometbft/blob/main/docs/references/architecture/adr-113-modular-transaction-hashing.md#detailed-design solution. Please let us know your thoughts on this matter.

PR #3257, opened 2024-06-13
docs(rpc): add info about encoding

melekes:
---

#### PR checklist


rpc/openapi/openapi.yaml
```diff
@@ -54,6 +54,17 @@ info:
     However, the support for unsuffixed methods will be dropped in future releases and all users are
     encouraged to migrate as soon as possible.
 
+    ## Encoding
```
cason, line 57: Is it not base64?

melekes: What are you referring to? If you're speaking about byte arrays, then yes.

cason: I know we have to encode txs in base64 format, this is the amino JSON format?

cason (approved):
+1

PR #3260, opened 2024-06-13
test: allow the v1.x nightly to be manually triggered

andynog:
In order to allow testing the nightly on `v1.x` without having to wait a full day

#### PR checklist

- [X] Tests written/updated


andynog:
No need to change this workflow. We can use the `e2e-manual` workflow that performs the same tests.

PR #3262, opened 2024-06-13
test(e2e): fix gRPC tests on nightly

andynog:
This PR changes some gRPC test on nightly. Instead of using a mix of RPC client (to fetch the height) and gRPC to fetch the block and results, the new logic leverages the gRPC get latest height

#### PR checklist

- [X] Tests written/updated


test/e2e/tests/grpc_test.go
```diff
@@ -36,31 +36,28 @@ func TestGRPC_Version(t *testing.T) {
...
 		ctx, ctxCancel := context.WithTimeout(context.Background(), time.Minute)
 		defer ctxCancel()
 		gRPCClient, err := node.GRPCClient(ctx)
 		require.NoError(t, err)
 		defer gRPCClient.Close()
 
+		latestHeightCh, err := gRPCClient.GetLatestHeight(ctx)
+		require.NoError(t, err)
+
+		latest, ok := <-latestHeightCh
```
cason, line 49: Can't we block forever here?

I would say to use a select with a timeout.

test/e2e/tests/grpc_test.go
```diff
@@ -36,31 +36,28 @@ func TestGRPC_Version(t *testing.T) {
...
+		require.Equal(t, ok, true)
+		require.Greater(t, latest.Height, int64(0))
```
cason, line 51: Can we make this stronger? Greater than 0 seems obvious here, since we don't have block 0, eheh.

cason (approved):
For me is not really clear what those tests are actually testing.

Apparently, we ask a node for its status, to know its last block. Then use private GRPC to get this block and check that everything is ok. 

This is flaky, I assume, because the node status gets outdated pretty fast (in particular on GitHub cloud/CI) and we at the same time prune old heights.

The solution is to use the private GRPC to test GRPC looks not ideal to me, honestly. But I don't think we want to have all those errors and this solution probably works.

Suggestion: add comments to the code showing the problem of the original version and/or create an issue to define the right way to test what we want to test here (which is also not clear due to the lack of comments).

test/e2e/tests/grpc_test.go
andynog: addressed on last commit 899e1f486eb4413dbec570ae69ee8858d37e3a5c

test/e2e/tests/grpc_test.go
andynog: addressed on commit 899e1f486eb4413dbec570ae69ee8858d37e3a5c

andynog:
tests are passing now https://github.com/cometbft/cometbft/actions/runs/9554601144

@melekes feel free to merge and backport if you approve the changes. This should fix the nigthly tests

PR #3264, opened 2024-06-13
fix(mempool): wrong request type when rechecking

### LINKED ISSUE #3261: v0.38.x: spammy "discard late recheck response" for initial check
We were evaluating the latest `v0.38.x` branch since it contains a number of promising performance improvements over `v0.38.7`, and in testing we are observing new error logs, which seem to be a result of the backport of https://github.com/cometbft/cometbft/pull/2268 to `v0.38.x` (done [here](https://github.com/cometbft/cometbft/pull/3020)).  This _seems_ to differ from the intent of the original pull request in that it applies to initial check as well as recheck.  When #2268 was merged to `main`, there was a nested `switch checkType {` to target only recheck for that error log, but on `v0.38.x` that didn't exists.

The effect we are observing in evaluating `v0.38.x` is that this `"discard late recheck response"` error log happens the instant a transaction is broadcast to the node, on the initial check and without any delay as the message suggests.

Is this the expected behavior, or a regression in `v0.38.x` after `v0.38.7`?
https://github.com/cometbft/cometbft/blame/ee556e375891db6bfbd64ea54e5671c1a0944858/mempool/clist_mempool.go#L290

More recently in https://github.com/cometbft/cometbft/pull/3178, some refactoring created an explicit recheck function: https://github.com/cometbft/cometbft/commit/9710d69f65804f346a88b15829f0c5fc05711e15#diff-f71975385bf5bc21c04eb08f496140cc8cca405b8dbd7b628342e6d8afb61cc5R429

We didn't see this log before, but we could be doing something wrong.  I can't tell.  Thanks!

hvanz:
Fixes #3261 

This bug only affects `v0.38.x`.

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

PR #3269, opened 2024-06-16
refactor: add miss defer and remove defer in the loop

hoanguyenkh:
---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/internal/blocksync/pool.go
+++ b/internal/blocksync/pool.go
@@ -785,23 +785,24 @@ OUTER_LOOP:
 			bpr.pickSecondPeerAndSendRequest()
 		}
 
-		retryTimer := time.NewTimer(requestRetrySeconds * time.Second)
-		defer retryTimer.Stop()
-
 		for {
+			retryTimer := time.NewTimer(requestRetrySeconds * time.Second)
 			select {
 			case <-bpr.pool.Quit():
 				if err := bpr.Stop(); err != nil {
 					bpr.Logger.Error("Error stopped requester", "err", err)
 				}
+				retryTimer.Stop()
 				return
 			case <-bpr.Quit():
+				retryTimer.Stop()
 				return
 			case <-retryTimer.C:
 				if !gotBlock {
 					bpr.Logger.Debug("Retrying block request(s) after timeout", "height", bpr.height, "peer", bpr.peerID, "secondPeerID", bpr.secondPeerID)
 					bpr.reset(bpr.peerID)
 					bpr.reset(bpr.secondPeerID)
+					retryTimer.Stop()
 					continue OUTER_LOOP
 				}
 			case peerID := <-bpr.redoCh:
@@ -835,6 +836,7 @@ OUTER_LOOP:
 				// We got a block!
 				// Continue the for-loop and wait til Quit.
 			}
+			retryTimer.Stop()
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


ValarDragon:
Why would you do this? This seems significantly worse? (More lines of code and complexity, and initiating more Timer objects in the main loop)

hoanguyenkh:
> Why would you do this? This seems significantly worse? (More lines of code and complexity, and initiating more Timer objects in the main loop)
> 
> Your supposed to only create a timer once, and listen on receives on the channel

thank @ValarDragon for feedback, I updated logic

ValarDragon:
Why is removing defer a goal?

hoanguyenkh:
> Why is removing defer a goal?

https://stackoverflow.com/a/45617791

internal/blocksync/pool.go
```diff
@@ -786,8 +786,6 @@ OUTER_LOOP:
 		}
 
 		retryTimer := time.NewTimer(requestRetrySeconds * time.Second)
-		defer retryTimer.Stop()
```
melekes, line 789: please revert this change

internal/blocksync/pool.go
```diff
@@ -796,12 +794,14 @@ OUTER_LOOP:
 				}
```
melekes, line 794: What about this place? We should also stop timer here.

internal/blocksync/pool.go
```diff
@@ -796,12 +794,14 @@ OUTER_LOOP:
...
 				return
 			case <-bpr.Quit():
+				retryTimer.Stop()
 				return
 			case <-retryTimer.C:
 				if !gotBlock {
 					bpr.Logger.Debug("Retrying block request(s) after timeout", "height", bpr.height, "peer", bpr.peerID, "secondPeerID", bpr.secondPeerID)
 					bpr.reset(bpr.peerID)
 					bpr.reset(bpr.secondPeerID)
+					retryTimer.Stop()
```
melekes, line 804: What's the point of stopping the timer when it's already fired?

rpc/jsonrpc/client/ws_client.go
```diff
@@ -214,7 +214,7 @@ func (c *WSClient) Stop() error {
 	// only close user-facing channels when we can't write to them
 	c.wg.Wait()
 	close(c.ResponsesCh)
-
+	c.PingPongLatencyTimer.Stop()
```
melekes, line 217: see https://github.com/cometbft/cometbft/pull/2804/files

melekes: please revert

melekes (changes requested):
Thanks @hoanguyenkh ❤️ I would ask you to revert some of the changes, as we deem them unnecessary / misplaced.

internal/blocksync/pool.go
hoanguyenkh: yeah, I have removed it and added manual retryTimer.Stop()

internal/blocksync/pool.go
hoanguyenkh: thank you, my misstake

rpc/jsonrpc/client/ws_client.go
cason, line 1: No change here.

cason (commented):
I would reject this PR, as changes in test units does not really produce improvements in the code or its testability.

PR #3291, opened 2024-06-17
perf(consensus): Make save ABCI responses and Commit occur in parallel

### LINKED ISSUE #3290: Run SaveABCIResponses and Commit in parallel
### Summary

Currently both SaveABCIResponses and Commit are ran in sequence during `ApplyBlock`. I think we should run these in parallel,  as they both have I/O and CPU logic. Especially as SaveABCIResponses can increase in time. 

### Problem Definition

Certain things like large event loads or Beginblock/Endblock loads, can make the `SaveABCIResponses` time quite expensive. So we want to lower the amount of time spent on this in the critical path.

### Proposal

Start SaveABCIResponses and app.Commit in parallel in blockexecutor.ApplyBlock

ValarDragon:
Close #3290 

This helps improve the performance of live consensus and block syncing, as we do not suffer synchronous delay for SaveABCIResponses. This takes time that blocks Commit

We see this time being particularly expensive with IBC txs.

---

#### PR checklist

- [ ] Tests written/updated


ValarDragon:
It was surprising to me, that under some high load osmosis blocks, SaveABCIResponses took 33% of the time that app.Commit was taking. (These blocks had Commit taking 4x longer than they normally took)

melekes (approved):
Thanks @ValarDragon ❤️

melekes:
This needs a changelog entry

cason:
I have doubts regarding this solution, in terms of the crash-recovery mechanisms of Comet, for instance: https://github.com/cometbft/cometbft/issues/3290#issuecomment-2178694838

PR #3313, opened 2024-06-21
feat(mempool): add simple back-pressure mechanism on peers

hvanz:
Currently, when attempting to add a tx received from a peer, if the mempool is full, the transaction will just be discarded.

This PR implements a simple backpressure mechanism on the mempool by adding a new channel `MempoolControlChannel` and a new control message `MempoolIsFull`:
- If node A, with a full mempool, receives a tx from B, it will reply `MempoolIsFull`.
- Node B on receiving `MempoolIsFull` from A will stop sending transactions to A during a predefined time.

Note that nodes perform a P2P handshake when they connect, so they will exchange messages only on the channels they have in common. A node that doesn't have `MempoolControlChannel` will just ignore messages received on this channel. And a node will not send messages on `MempoolControlChannel` if it knows that the peer doesn't have the corresponding channel.

cc @sergio-mena @ValarDragon 

---

#### PR checklist

- [ ] Tests written/updated


zmanian:
So I think it might be smart to put an integer in the mempool full message and basically the idea would be an application specific priority value and it would indicate that " I don't want to receive messages lower than priority n where n is the integer in the message". 

I think that priority value could be integrated into the mempool in a. number of ways.

cason:
If we are adding an additional channel for control, my approach would be more proactive than reactive.

Think about TCP. We don't say "stop sending us stuff" because I can't consume data fast enough. We instead send acks (implicitly or not) informing what we already have processed and thus allowing the sender to adapt its sending rate.

In other words, we say to the other side: I can afford to handle X messages or bytes. And if it does not comply with my affirmation, it is a bad peer, and I'll probably drop his messages.

Not to mention here that bad/attackers peers will completely ignore this message. Also, once we have space in our mempool we will have peers not sending us anything because they are waiting for an arbitrary time to be defined. Also, we cannot punish anyone for sending stuff to us during this arbitrary time, because we just don't known how long it takes to this "stop sending me stuff" message to arrive.

Again, all this discussion should occur on an issue.

itsdevbear:
instead of "MempoolIsFull" do we want to generalize this to some sort of "Stop sending me stuff" message

hvanz:
Closing for now as currently we don't have the capacity to work on this. The brainstorming should continue in an issue or discussion.

PR #3314, opened 2024-06-21
fix(mempool)!: stop accepting TXs in the mempool if we can't keep up with reCheckTX

sergio-mena:
This PR is a combination of ideas from @ValarDragon, @hvanz and @sergio-mena to alleviate nodes that, while not having their mempool full "officially", they have too many TXs lingering in the mempool which causes them to fall behind.
The mechanism works as follows:

* We mark when we start and end reChecking
* If, by the time a new block is decided we are still running the previous `reCheckTx`, we declare the mempool as rechecktx-full
* Otherwise, we declare the mempool as not rechecktx-full

We have tested this and it fixes the failing nightlies that are blocking us from cutting `v1.0.0-rc1`.

Some UTs need to be fixed, hence posting as draft for the moment.

---

#### PR checklist

- [ ] Tests written/updated


mempool/errors.go
```diff
@@ -25,10 +25,11 @@ func (e ErrTxTooLarge) Error() string {
...
 type ErrMempoolIsFull struct {
-	NumTxs      int
-	MaxTxs      int
-	TxsBytes    int64
-	MaxTxsBytes int64
+	NumTxs           int
+	MaxTxs           int
+	TxsBytes         int64
+	MaxTxsBytes      int64
+	RecheckSaturated bool
```
sergio-mena, line 32: @andynog please rename to `RecheckFull`

mempool/clist_mempool.go
```diff
@@ -176,6 +181,15 @@ func (mem *CListMempool) Unlock() {
 	mem.updateMtx.Unlock()
 }
 
+// Safe for concurrent use by multiple goroutines.
+func (mem *CListMempool) PreUpdate() {
+	rechecking := mem.rechecking.Load()
+	reCheckFull := mem.recheckFull.Swap(rechecking)
+	if rechecking != reCheckFull {
+		mem.logger.Info("the state of recheckTooLong has flipped", "before", reCheckFull, "after", rechecking)
```
sergio-mena, line 189:
```suggestion
		mem.logger.Info("the state of recheckFull has flipped", "before", reCheckFull, "after", rechecking)
```

mempool/clist_mempool.go
```diff
@@ -45,6 +45,11 @@ type CListMempool struct {
 
 	// Keeps track of the rechecking process.
 	recheck *recheck
+	// It is true while TXs are being rechecked
+	rechecking atomic.Bool
```
hvanz, line 49: This variable is redundant. It's equivalent to `recheck.done()`.

mempool/clist_mempool.go
```diff
@@ -45,6 +45,11 @@ type CListMempool struct {
...
+	// whether rechecking TXs cannot be completed before
+	// a new block is decided
+	recheckFull atomic.Bool
```
hvanz, line 52: I would define this variable inside the `recheck` struct, to keep all the recheck state in one place.

mempool/clist_mempool.go
```diff
@@ -176,6 +176,15 @@ func (mem *CListMempool) Unlock() {
 	mem.updateMtx.Unlock()
 }
 
+// Safe for concurrent use by multiple goroutines.
+func (mem *CListMempool) PreUpdate() {
+	rechecking := mem.recheck.done()
```
greg-szabo, line 181: This is confusing for me. If rechecking is done, shouldn't recheck say false?

mempool/clist_mempool.go
```diff
@@ -176,6 +176,15 @@ func (mem *CListMempool) Unlock() {
...
+	reCheckFull := mem.recheck.recheckFull.Swap(rechecking)
+	if rechecking != reCheckFull {
+		mem.logger.Info("the state of recheckFull has flipped", "before", reCheckFull, "after", rechecking)
```
greg-szabo, line 184: I think this is enough as a debug message. Is it necessary to print it as info?

greg-szabo: If anything, an info could be printed when reCheckFull == true to inform the validator that their mempool has problematic transactions that slow down transaction checks.

greg-szabo (approved):
I'd rather have Debug level for the newly introduced log message, because it's a bit cryptic for validators. For n end-user, I could consider a "mempool is soft full" message useful, although there's not much they can do about it.

mempool/clist_mempool.go
```diff
@@ -407,17 +416,17 @@ func (mem *CListMempool) RemoveTxByKey(txKey types.TxKey) error {
 }
 
 func (mem *CListMempool) isFull(txSize int) error {
-	var (
-		memSize  = mem.Size()
-		txsBytes = mem.SizeBytes()
-	)
+	memSize := mem.Size()
+	txsBytes := mem.SizeBytes()
+	recheckFull := mem.recheck.recheckFull.Load()
```
melekes, line 421:
```suggestion
	recheckFull := mem.config.Recheck && mem.recheck.recheckFull.Load()
```

mempool/clist_mempool.go
```diff
@@ -176,6 +176,15 @@ func (mem *CListMempool) Unlock() {
 	mem.updateMtx.Unlock()
 }
 
+// Safe for concurrent use by multiple goroutines.
+func (mem *CListMempool) PreUpdate() {
+	rechecking := !mem.recheck.done()
```
melekes, line 181:
```suggestion
	if !mem.config.Recheck { return }
	
	rechecking := !mem.recheck.done()
```

melekes (commented):
Shouldn't we respect the `Recheck` config setting?

mempool/clist_mempool.go
```diff
@@ -673,6 +682,7 @@ type recheck struct {
 	end           *clist.CElement // last entry in the mempool to recheck
 	doneCh        chan struct{}   // to signal that rechecking has finished successfully (for async app connections)
 	numPendingTxs atomic.Int32    // number of transactions still pending to recheck
+	recheckFull   atomic.Bool     // whether rechecking TXs cannot be completed before a new block is decided
```
hvanz, line 685: rename to `isFull`

mempool/mempool.go
```diff
@@ -56,6 +56,10 @@ type Mempool interface {
 	// Unlock unlocks the mempool.
 	Unlock()
 
+	// PreUpdate Signal that a new update is coming, before acquiring the mempool lock.
```
hvanz, line 59:
```suggestion
	// PreUpdate signals that a new update is coming, before acquiring the mempool lock.
```

hvanz:
The unit tests need to be updated as well. Every `mempool.Lock()` in a test should be preceded by `mempool.PreUpdate()` (probably this means that the logic of PreUpdate should be part of Lock?) With these changes now I see that `TestReactorConcurrency` fails.

mempool/clist_mempool.go
hvanz: After suggesting the above I realized that they're not equivalent because `recheck.Done()` is equal to `recheck.cursor == nil`, which is not thread safe. We may have a scenario where recheck is updating ``recheck.cursor` and `PreUpdate` is reading it. Still, I don't see any race condition on the unit tests, even after fixing them (see comment below). Probably we need new tests.

mempool/clist_mempool.go
hvanz: or something more meaningful

mempool/clist_mempool.go
sergio-mena: How about leaving it as it is now (even if it suboptimal in terms of variables), so we're confident we can backport it, and then, in a furhter PR, we improve this on `main` and possibly backport it to `v1.x` (or not).
WDYT?

mempool/clist_mempool.go
sergio-mena: I don't think it's necessary... if `Recheck` is disabled in config, it will never be fired and so, `recheckFull` can never become full.

mempool/clist_mempool.go
sergio-mena: We don't need to consider the config. If it's disabled, the pre-existing code handles that case correctly

mempool/clist_mempool.go
sergio-mena: I see this has been changed. Shall we revert it?

hvanz: That's an option. But it could be confusing because we'll have two ways to check if rechecking is happening. Another option is #3333.

sergio-mena:
I realised I forgot to add a changelog in Fri. @hvanz could you add it please?

mempool/clist_mempool.go
```diff
@@ -753,3 +766,17 @@ func (rc *recheck) findNextEntryMatching(tx *types.Tx) bool {
...
+func (rc *recheck) setRecheckFull() bool {
+	rechecking := !rc.done()
+	recheckFull := rc.recheckFull.Swap(rechecking)
+	return rechecking != recheckFull
+}
+
+// consideredFull returns true iff the mempool should be considered as full while rechecking is in
+// progress.
+func (rc *recheck) consideredFull() bool {
+	return rc.recheckFull.Load()
```
andynog, line 781: I was going to comment that it would be interesting to also check if the rechecking is happening here, but from the `setRecheckFull()` logic seems that the `recheckFull` will be only true when it's is rechecking (`!rc.done()`) and that is covered in that logic I believe

andynog (approved):
looks good to me

PR #3317, opened 2024-06-23
chore(ticker): Prevent ticker from blocking

itsdevbear:
---

#### PR checklist

- [ ] Tests written/updated


internal/consensus/ticker.go
```diff
@@ -71,7 +71,11 @@ func (t *timeoutTicker) Chan() <-chan timeoutInfo {
 // The timeoutRoutine is always available to read from tickChan, so this won't block.
 // The scheduling may fail if the timeoutRoutine has already scheduled a timeout for a later height/round/step.
 func (t *timeoutTicker) ScheduleTimeout(ti timeoutInfo) {
-	t.tickChan <- ti
+	select {
```
melekes, line 74: What triggered this change? Did you encounter a case where Comet blocks here for a prolonged period?

internal/consensus/ticker.go
```diff
@@ -121,7 +128,11 @@ func (t *timeoutTicker) timeoutRoutine() {
 			// Determinism comes from playback in the receiveRoutine.
 			// We can eliminate it by merging the timeoutRoutine into receiveRoutine
 			//  and managing the timeouts ourselves with a millisecond ticker
-			go func(toi timeoutInfo) { t.tockChan <- toi }(ti)
+			select {
```
melekes, line 131: same question

melekes (commented):
Thanks @itsdevbear ❤️

ValarDragon:
I'm confused whats the cause of this. inputs to TickChan should be single threaded, so we won't get blocking behavior adding to the queue.

If theres a problem here, imo we just raise the Tick channel buffer to like 100. That should basically cost us nothing, and lower any probability of blocking behavior anyway

cason:
I agree that just increasing the size of the buffer should be enough. But the question is: how many threads are scheduling timeouts at a time? I thing there is a single one.

cason (changes requested):
I will block this PR because:

- We probably cannot afford _not_ scheduling a timeout for some internal channel problems in the ticker
- We don't have any evidence that this situation may actually occur and affect performance

melekes:
Given the above comments and no response from the PR's author, I am closing this. @itsdevbear, if/when you find some time to respond to the comments, please do so, and we'll happily reconsider the changes.

PR #3318, opened 2024-06-24
chore(consensus): Logs should be capitalize for consistency across repo

itsdevbear:
---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

melekes:
```
internal/consensus/state.go:2619:9: ST1005: error strings should not be capitalized (stylecheck)
		panic(fmt.Errorf("Vote extension absence/presence does not match extensions enabled %t!=%t, height %d, type %v",
```

internal/consensus/state.go
```diff
@@ -2602,17 +2602,17 @@ func (cs *State) signAddVote(
...
 	if err != nil {
-		cs.Logger.Error("failed signing vote", "height", cs.Height, "round", cs.Round, "vote", vote, "err", err)
+		cs.Logger.Error("Failed signing vote", "height", cs.Height, "round", cs.Round, "vote", vote, "err", err)
 		return
 	}
 	hasExt := len(vote.ExtensionSignature) > 0
 	extEnabled := cs.state.ConsensusParams.Feature.VoteExtensionsEnabled(vote.Height)
 	if vote.Type == types.PrecommitType && !vote.BlockID.IsNil() && hasExt != extEnabled {
-		panic(fmt.Errorf("vote extension absence/presence does not match extensions enabled %t!=%t, height %d, type %v",
+		panic(fmt.Errorf("Vote extension absence/presence does not match extensions enabled %t!=%t, height %d, type %v",
```
melekes, line 2611:
```suggestion
		panic(fmt.Errorf("vote extension absence/presence does not match extensions enabled %t!=%t, height %d, type %v",
```

internal/consensus/state.go
```diff
@@ -1568,20 +1568,20 @@ func (cs *State) enterPrevoteWait(height int64, round int32) {
...
+			"Entering prevote wait step with invalid args",
 			"current", log.NewLazySprintf("%v/%v/%v", cs.Height, cs.Round, cs.Step),
 		)
 		return
 	}
 
 	if !cs.Votes.Prevotes(round).HasTwoThirdsAny() {
 		panic(fmt.Sprintf(
-			"entering prevote wait step (%v/%v), but prevotes does not have any +2/3 votes",
+			"Entering prevote wait step (%v/%v), but prevotes does not have any +2/3 votes",
```
melekes, line 1579:
```suggestion
			"entering prevote wait step (%v/%v), but prevotes does not have any +2/3 votes",
```

internal/consensus/state.go
```diff
@@ -1668,19 +1668,19 @@ func (cs *State) enterPrecommit(height int64, round int32) {
 	// the proposed block, update our locked block to this block and issue a
 	// precommit vote for it.
 	if cs.ProposalBlock.HashesTo(blockID.Hash) {
-		logger.Debug("precommit step: +2/3 prevoted proposal block; locking", "hash", blockID.Hash)
+		logger.Debug("Precommit step: +2/3 prevoted proposal block; locking", "hash", blockID.Hash)
 
 		// Validate the block.
 		if err := cs.blockExec.ValidateBlock(cs.state, cs.ProposalBlock); err != nil {
-			panic(fmt.Sprintf("precommit step; +2/3 prevoted for an invalid block: %v; relocking", err))
+			panic(fmt.Sprintf("Precommit step; +2/3 prevoted for an invalid block: %v; relocking", err))
```
melekes, line 1675:
```suggestion
			panic(fmt.Sprintf("precommit step; +2/3 prevoted for an invalid block: %v; relocking", err))
```

melekes:
I can't push to your fork @itsdevbear. Please resolve the linter issues above.

internal/consensus/state.go
itsdevbear: errors shouldn't be capitalized, just log messages

internal/consensus/state.go
itsdevbear: errors shouldn't be capitalized

itsdevbear:
@melekes fixed, are you okay to put into v1.x as well? >

sergio-mena:
@melekes @itsdevbear sorry not to notice this PR before, but this goes in the opposite direction of what we want to achieve with logs in our repository. I agree it's currently inconsistent, but we want to remove the initial capital letter from all our logs (unless it's justified as an acronym, etc).

The reason for this is that errors tend to concatenated in one single line (separated by colons or semicolons), and it's not aesthetic to have each chunk of the error capitalised.

sergio-mena:
Please ignore my latest message. I was incorrectly conflating log messages and errors. Sorry for the noise 🙏

itsdevbear:
np @sergio-mena I've always operated under the assumption with go:
Log messages -> Capitalized
Errors -> lower case, for chaining, exactly as you said

PR #3332, opened 2024-06-24
refactor: use go ticker instead of timer

hoanguyenkh:
use timer.NewTicker, will remove the need for the manual retryTimer.Reset(requestRetrySeconds * time.Second) call

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/internal/blocksync/pool.go
+++ b/internal/blocksync/pool.go
@@ -785,8 +785,8 @@ OUTER_LOOP:
 			bpr.pickSecondPeerAndSendRequest()
 		}
 
-		retryTimer := time.NewTimer(requestRetrySeconds * time.Second)
-		defer retryTimer.Stop()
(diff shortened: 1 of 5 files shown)


melekes (changes requested):
Thanks @hoanguyenkh ❤️ 

However, the proposed logic is different from the original behavior. The timer gets reset (or prolonged) in the existing code under a specific condition. In the proposed changes, we Stop the ticker, but when do we restart it? I'm inclined to close this PR as it doesn't improve the code in any way.

hoanguyenkh:
> Thanks @hoanguyenkh ❤️
> 
> However, the proposed logic is different from the original behavior. The timer gets reset (or prolonged) in the existing code under a specific condition. In the proposed changes, we Stop the ticker, but when do we restart it? I'm inclined to close this PR as it doesn't improve the code in any way.

thank you for the feedback, in this case, NewTimer is better than, sorry

PR #3334, opened 2024-06-24
feat(crypto)!: modular transaction hashing

melekes:
Closes https://github.com/cometbft/cometbft/issues/3383

#go-api-breaking

Removes unused `HashFromByteSlicesIterative` and `SumMany` in favor of simple implementation.

---

#### PR checklist

- [x] Tests written/updated


cason:
Are the changes under `crypto/merkle` a necessary part of this PR, or just a further contribution?

crypto/txhash/doc.go
cason, line 1: We probably need to reinforce that this change can only be made **before** the node is created and/or started, and never while a node is running.

melekes:
> Are the changes under `crypto/merkle` a necessary part of this PR, or just a further contribution?

further contribution

crypto/txhash/doc.go
melekes: Do you have any ideas of how this could be done?

cason: I have some ideas, but none of them are great.

I would just write down in the documentation this information: if you use this method when a node is running, the behavior is unpredictable.

cason:
> further contribution

Would you mind to create another issue with those contributions? I can help with, if you want.

crypto/merkle/hash.go
```diff
@@ -12,33 +10,26 @@ var (
...
-// returns tmhash(0x00 || leaf).
-func leafHashOpt(s hash.Hash, leaf []byte) []byte {
-	s.Reset()
-	s.Write(leafPrefix)
-	s.Write(leaf)
-	return s.Sum(nil)
+// returns h(<empty>).
+func emptyHash(h hash.Hash) []byte {
+	h.Reset()
+	h.Write([]byte{})
```
sergio-mena, line 16: Are we sure we need this line?

sergio-mena: Actually, I checked the `crypto/sha256` implementation and it somehow expects to process empty slices. So please just ignore my comment

crypto/merkle/proof_value.go
```diff
@@ -11,6 +11,10 @@ import (
 
 const ProofOpValue = "simple:v"
 
+// ErrTooManyArgs is returned when the input to [ValueOp.Run] has length
+// exceeding 1.
+var ErrTooManyArgs = errors.New("merkle: len(args) != 1")
```
sergio-mena, line 16: The message doesn't really match the error's name. Which one is right?

sergio-mena: Oh, I see you just moved it. You can just leave it like that, but it would be good to make error name and message consistent

crypto/merkle/bench_test.go
```diff
@@ -26,9 +26,10 @@ var innerHashTests = []*innerHashTest{
 func BenchmarkInnerHash(b *testing.B) {
 	b.ReportAllocs()
 
+	h := sha256.New()
```
sergio-mena, line 29: Why not `tmhash.New()` (just trying to understand the usage).
This also applies to other callsites of `sha256.New()` in this PR.

crypto/tmhash/bench_test.go
```diff
@@ -1,52 +0,0 @@
...
-		want: sha256.Sum256([]byte("aaaa😎aaaa")),
-	},
-	{
-		name: "composite joined",
-		in:   [][]byte{bytes.Repeat([]byte("a"), 1<<10), []byte("AA"), bytes.Repeat([]byte("z"), 100)},
-		want: sha256.Sum256([]byte(strings.Repeat("a", 1<<10) + "AA" + strings.Repeat("z", 100))),
-	},
-}
-
-func BenchmarkSHA256Many(b *testing.B) {
```
sergio-mena, line 34: Not that I'm against removing code (I actually love the idea 😄), but what is the reason for deleting this benchmark?

crypto/txhash/doc.go
sergio-mena: Careful, it's way stricter than that.
It's unacceptable to change the hashing algorithm after the chain itself has been started (and this will continue to apply until we deliver "soft updates").

crypto/txhash/doc.go
```diff
@@ -0,0 +1,9 @@
+// Package txhash provides a way to change the hash function used for
+// transaction hashing.
+//
+// The default hash function is SHA256.
+// You can change the hash function by calling txhash.Set.
+//
+// You can also change how a checksum is converted to a string by calling
+// txhash.SetFmtHash.
```
sergio-mena, line 8:
```suggestion
// txhash.SetFmtHash.
//
// WARNING: Currently, CometBFT does NOT support changing the hash function
// after a chain has been started.
// A chain is expected to agree on the hash function to be used before genesis time,
// and stick to it.
// If a chain changes the hash function after it was started (e.g., as part of a coordinated
// upgrade) many things will stop working (block validation, block sync, light clients, etc).
```

crypto/txhash/hash.go
```diff
@@ -0,0 +1,61 @@
...
+// Bytes is a wrapper around a byte slice that implements the fmt.Stringer.
+type Bytes []byte
+
+func (bz Bytes) String() string {
+	return fmtHash(bz)
+}
+
+func (bz Bytes) Unwrap() []byte {
+	return bz
+}
```
sergio-mena, line 34: We could do this at the callsite with just a type cast, couldn't we?

crypto/txhash/hash_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+func TestSetFmtHash(t *testing.T) {
+	bz := []byte("hello")
+	assert.Equal(t, "9B71D224BD62F3785D96D46AD3EA3D73319BFBC2890CAADAE2DFF72519673CA72323C3D99BA5C11D7C7ACC6E14B8C5DA0C4663475C2E5C3ADEF46F73BCDEC043", Sum(bz).String())
+
+	SetFmtHash(func(bz []byte) string {
+		return hex.EncodeToString(bz)
+	})
+	assert.Equal(t, "9b71d224bd62f3785d96d46ad3ea3d73319bfbc2890caadae2dff72519673ca72323c3d99ba5c11d7c7acc6e14b8c5da0c4663475c2e5c3adef46f73bcdec043", Sum(bz).String())
+}
```
sergio-mena, line 43: Can you add a test where we hash the same bytes with two different hash functions and make sure the result differs (so, testing that the `Set` function isn't a noop)

types/tx.go
```diff
@@ -25,11 +25,14 @@ type (
...
+// Hash computes the hash of the transaction.
+//
+// See txhash.Sum.
+func (tx Tx) Hash() txhash.Bytes {
+	return txhash.Sum(tx)
 }
 
+// Key returns the sha256 hash of the transaction.
 func (tx Tx) Key() TxKey {
 	return sha256.Sum256(tx)
```
sergio-mena, line 37: Why aren't we changing this one to use the hash the user has set?

sergio-mena (commented):
A general comment: doesn't package `txhash` render `tmhash.New()` and `tmhash.Sum(...)`obsolete?
Moreover, should we be migrating all calls to `tmhash.New()` and `tmhash.Sum(...)` to `txhash`?

types/tx.go
melekes: Because ` TxKey is the fixed-length array key used as an index, ' we only use it internally, afaik, in the mempool reactor. Do you think we should be using `txhash` here as well?

melekes:
> A general comment: doesn't package `txhash` render `tmhash.New()` and `tmhash.Sum(...)` obsolete? Moreover, should we be migrating all calls to `tmhash.New()` and `tmhash.Sum(...)` to `txhash`?

No, because if we change consensus params or Merkle tree hashing, we lose IBC compatibility. The destination chain needs to know the hashing algorithm to verify hashes. So, we must keep `tmhash` until we have a field in the header specifying the hashing algorithm.

melekes:
@fedekunze @itsdevbear, does this satisfy your requirements? Would you please test this PR and see if it helps with your use case?

types/tx.go
```diff
@@ -42,11 +45,13 @@ func (tx Tx) String() string {
...
-// Hash returns the Merkle root hash of the transaction hashes.
-// i.e. the leaves of the tree are the hashes of the txs.
+// Hash returns the Merkle root hash.
+// The leaves of the tree are the  hashes of the txs.
+//
+// See merkle.HashFromByteSlicesWithHash.
 func (txs Txs) Hash() []byte {
 	hl := txs.hashList()
-	return merkle.HashFromByteSlices(hl)
+	return merkle.HashFromByteSlicesWithHash(txhash.New(), hl)
```
melekes, line 54: We break `Header#DataHash` here. I.e. another chain won't be able to compute this without the knowledge of the hashing algorithm used by this chain.

types/tx.go
fedekunze: Shouldn't the hash be the key?

crypto/txhash/hash.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+)
+
+// Hash is an interface for transaction hashing.
+type Hash interface {
+	// New returns a new hash.Hash.
+	New() gohash.Hash
+}
+
+// Bytes is a wrapper around a byte slice that implements the fmt.Stringer.
```
fedekunze, line 25: unrelated to this PR, but would it be possible to use `Bytes` for the block hash formatting too? Would be nice to ensure full compatibility with ethereum formatting

melekes:
There is no feedback here.

PR #3335, opened 2024-06-24
test(e2e): reverting PR3211 since it is making e2e nightly to fail

andynog:
This PR reverts #3211 since it is making the e2e nightly to fail in reproducible ways. This PR was identified as the reason for the recent e2e nightly failure on `main` and doing a bi-sect test of all recent commits it was determined that this PR introduced a behavior that makes the tests to fail and indicate a bug or unknown behavior has been introduced. 

Even though we are reverting this logic for now, we'd be happy to consider it in the future again once more tests are performed and we can ensure it passes all tests.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/.changelog/unreleased/improvements/3211-make-cs-reactor-no-longer-takes-cs-locks.md
+++ b/.changelog/unreleased/improvements/3211-make-cs-reactor-no-longer-takes-cs-locks.md
@@ -1,4 +1,0 @@
-- `[consensus]` Make the consensus reactor no longer have packets on receive take the consensus lock.
-Consensus will now update the reactor's view after every relevant change through the existing 
-synchronous event bus subscription.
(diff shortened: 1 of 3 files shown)


sergio-mena (approved):
Checked, via git fiddling, that this is the reverse patch. ✅

cason (approved):
Diffs match :)

ValarDragon:
Reading the logs, I think this issue should be easy for us to solve. 

It seems that the reactor's view of initial height on some nodes is incorrectly being set to 0 (rather than the true initial height). I think theres only two causes for this:
- updateToState hasn't been called yet
- (unlikely) initialHeight hasn't propogated to state yet. PR'ing a fix.

cason:
So lets revert the reversion!

PR #3341, opened 2024-06-26
perf(consensus): Undo revert #3211. (Remove reactor Consensus RLocks)

ValarDragon:
Pretty simple bug fix for the e2e failure on #3211. There was a race condition at iniitialization for initial height, because we didn't initialize it early on enough.

The error in the E2E logs was:
```
validator03    | E[2024-06-21|21:13:20.744] Stopping peer for error                      module=p2p peer="Peer{MConn{10.186.73.2:34810} 4fe295e4cfad69f1247ad85975c6fd87757195db in}" err="invalid field LastCommitRound can only be negative for initial height 0"
validator03    | I[2024-06-21|21:13:20.744] service stop                                 module=p2p peer=4fe295e4cfad69f1247ad85975c6fd87757195db@10.186.73.2:34810 msg="Stopping Peer service" impl="Peer{MConn{10.186.73.2:34810} 4fe295e4cfad69f1247ad85975c6fd87757195db in}"
```
hinting at initial height not being set rpoperly.

---

#### PR checklist

- [ ] Tests written/updated


internal/consensus/reactor.go
```diff
@@ -57,10 +58,11 @@ type ReactorOption func(*Reactor)
...
 func NewReactor(consensusState *State, waitSync bool, options ...ReactorOption) *Reactor {
 	conR := &Reactor{
-		conS:     consensusState,
-		waitSync: atomic.Bool{},
-		rs:       consensusState.GetRoundState(),
-		Metrics:  NopMetrics(),
+		conS:          consensusState,
+		waitSync:      atomic.Bool{},
+		rs:            consensusState.GetRoundState(),
+		initialHeight: consensusState.state.InitialHeight,
```
ValarDragon, line 64: This line is what was missing in the original PR, and is what should fix the bug

ValarDragon:
e2e-manual run: https://github.com/cometbft/cometbft/actions/runs/9671388479

melekes (approved):
Thanks @ValarDragon ❤️

melekes:
_Blocked until we release an RC._

internal/consensus/reactor.go
cason: From the family, always initialize any variable you add, ehehe

cason (approved):
The diffs match, the only new part is line 64, where we initialize `initialHeight`.

I really hope we don't need to revert the reversion of the reversion of it.

melekes:
@mergify backport v1.x

PR #3342, opened 2024-06-26
perf(blockstore)!: Cache more in blockstore, which speedsup the gossip routines

ValarDragon:
We speedup the gossip routines by using caches for LoadBlockPart, and making LoadBlockCommit no longer call Clone, as the caller does read-only operations to it.

This is a followup to https://github.com/cometbft/cometbft/pull/3003

This results in a net performance improvement, based on a 2hr sync on latest osmosis release, of:
- 20% less RAM allocated over program lifetime. (76GB no longer allocated, 380 GB alloc total)
- 50% speedup to gossipDataRoutine
- ~10% speedup to gossipVotesRoutine
- 1% less overall system CPU usage. (If GC proportional to bytes allocated, then 5% less)

I have checked that every usage of `LoadBlockPart` and `LoadBlockCommit` do not modify the return value, so returning this underlying cache copy is safe for the current codebase

I used an `!` in the title, since this is technically API breaking, even though all current usages are safe.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -69,6 +69,7 @@ type BlockStore struct {
 	seenCommitCache          *lru.Cache[int64, *types.Commit]
 	blockCommitCache         *lru.Cache[int64, *types.Commit]
 	blockExtendedCommitCache *lru.Cache[int64, *types.ExtendedCommit]
+	blockPartCache           *lru.Cache[blockPartIndex, *types.Part]
 }
 
 type BlockStoreOption func(*BlockStore)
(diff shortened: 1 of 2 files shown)


store/store.go
```diff
@@ -512,14 +530,17 @@ func (bs *BlockStore) PruneBlocks(height int64, state sm.State) (uint64, int64,
 			if err := batch.Delete(bs.dbKeyLayout.CalcBlockCommitKey(h)); err != nil {
 				return 0, -1, err
 			}
+			bs.blockCommitCache.Remove(h)
```
melekes, line 533: Do we need to do this? If we don't, new block commits will inevitably replace old / pruned commits anyway.

melekes (approved):
Thanks @ValarDragon ❤️

store/store.go
ValarDragon: We may not need to do it, got a test failure for not doing it on block parts (because the pruning tests were relying on checking that it got removed)

melekes:
@mergify backport v1.x

Issue #3340, opened 2024-06-25
Block sync problem in short block time chain.


beer-1:
### Setup

**CometBFT version** (use `cometbft version` or `git rev-parse --verify HEAD` if installed from source): v0.38.x

**Have you tried the latest version**: yes

**ABCI app** (name for built-in, URL for self-written if it's publicly available): initia, injective, sei

### What happened?

We are currently running single validator network with 0.5s block time setting and trying to keep syncing from other full node.
But periodically, the full node keeps failing to keep sync and fall behind. and wired thing is that restarting the node solves the problem.

And also got some reports like they also having this problem in Sei and Injective chain and solve the issue by restarting the node.

### What did you expect to happen?

no problem in block sync

### How to reproduce it

Prease

ValarDragon:
Yeah, I think our problem is that in consensus mode:
- Nodes don't save data sent from future heights/rounds
- Peers as a consequence try to not gossip things too far in the future
- Once a peer has sent a vote/block part they will never send it again.

cason:
I think one of the problems here is the following. To decide a block we need 2/3+ Precommits for the block, they contain the `BlockID`, and we need the full block matching that `BlockID`.

When a node realizes that a peer is lagging behind it does the follow, in different routines:

1. Sends Precommits for the `BlockID` of the block committed at the reported peer height
2. Sends block Parts for the `Block` committed at the reported peer height

We only accept a block Part if it matches our `cs.ProposalBlockParts`, which is a Merkle tree. We setup this Merkle tree upon receiving a `Proposal`, this is the ordinary operation, **or** when we see 2/3+ Prevotes or Precommits for a block (that is not the block we are building or have built).

What matters here is the second case with Precommit messages. Only when we get the 2/3+ Precommits for a `BlockID`, we start accepting the Parts of the associated block:

https://github.com/cometbft/cometbft/blob/4d98dc1b2f611e8f2545172e66799a7bd9d7a317/internal/consensus/state.go#L1774-L1786

In summary, we are receiving block parts before we receive the 2/3+ Precommits that would allow us to accept the associated block parts. This means that the block parts that we have received, although good ones, were dropped, and we are waiting for their duplicates, eventually coming from other peers.

To attest my thesis, you could look to the INFO message of the code above and increase the log level of the message produced when we reject a block part:

[...]

PR #3343, opened 2024-06-26
perf(mempool): disable rechecking when node is catching up

hvanz:
When CheckTx is slow (on the application), a node may fail to catch up to reach the latest height, as seen in #2925. In particular, rechecking may take a lot of time to finish, blocking the mempool from receiving new transactions, and not allowing to node to ever catch up. #3008 has improved the performance but we can still reproduce the problem with [this e2e testnet](https://github.com/cometbft/cometbft/blob/5d6b2b5481b15c149e7b0e49295253cb38ca15b8/test/e2e/networks_regressions/slow_recheck.toml).

This PR improves the performance of nodes that are lagging behind by temporarily disabling rechecking. A node is considered to be lagging behind when at least one of its peers is at a later height. In the code "later" is a difference of at least two blocks. Nodes keep track of the latest height of its peers, which send their height in proposal and vote messages.

If a node is not at the latest height, I consider that it is **safe to not recheck transactions during some blocks**, until the node catches up. The transactions in its mempool are either old transactions, which were already included in a committed block by some other node, or new transactions received via RPC, meaning that most probably no other node has these transactions. The node won't be able to create a block with the new transactions until it catches up but it can disseminate them to be committed by other validators. Also, the node won't receive new transactions from peers because nodes don't gossip transactions to peers that are lagging behind ([code](https://github.com/cometbft/cometbft/blob/67aa24b4892027393bbb15570d1d950557159dfb/mempool/reactor.go#L240-L250), RFC-103).

Might be related to #3340

---

#### PR checklist

- [ ] Tests written/updated


p2p/peer.go
```diff
@@ -43,6 +44,9 @@ type Peer interface {
 	Set(key string, value any)
 	Get(key string) any
 
+	SetHeight(h int64)
```
hvanz, line 47: These two methods overlap with `Get` and `Set` but I couldn't used them in `execution.go` without introducing a circular dependency with consensus. In mempool and evidence we call `peer.Get(types.PeerStateKey).(*PeerState)` to obtain the height. In the future we could replace this call by `peer.Height()`.

internal/consensus/reactor.go
cason, line 1: All routines contain:

```
ps, ok := e.Src.Get(types.PeerStateKey).(*PeerState)
```

This is the peer state known by the node, which is stored in the `Peer` object. The last known height is just `ps.PRS.Height`. No need for new data structures to keep this information.

p2p/peer.go
cason, line 1: The data you are setting is redundant, see comments on the consensus reactor. There is a keystore on each peer, it is synchronized, and we keep the peer state, with a summary of the information we have about it, on this store.

state/execution.go
```diff
@@ -450,12 +459,23 @@ func (blockExec *BlockExecutor) asyncUpdateMempool(
 ) {
 	defer unlockMempool()
 
+	recheck := true
```
cason, line 462: We know the peer height, from the same already existing database, in the mempool protocol.

I don't see a point to the block executor to interact with the p2p layer.

cason (changes requested):
I disagree with this approach.

We are adding to the peer keystore a redundant information, which is already there, produced by the consensus reactor and read by the mempool and evidence reactors. See details here: https://github.com/cometbft/cometbft/blob/main/spec/p2p/reactor-api/p2p-api.md#key-value-store

internal/consensus/reactor.go
hvanz: Thanks for pointing that out. Now I'm using this approach, which is much simpler.

p2p/peer.go
hvanz: Changed the approach to get the info from the mempool reactor.

state/execution.go
hvanz: Same as above. Now I'm getting the peer info from the mempool reactor.

mempool/clist_mempool.go
```diff
@@ -20,6 +20,8 @@ import (
 	"github.com/cometbft/cometbft/types"
 )
 
+const maxHeightDiffForRechecking = 2
```
andynog, line 23: Should this be a config parameter? Are we OK having this value hard coded?

mempool/clist_mempool.go
```diff
@@ -96,6 +100,10 @@ func NewCListMempool(
 	return mp
 }
 
+func (mem *CListMempool) SetSwitch(sw *p2p.Switch) {
```
andynog, line 103: The p2p switch only needed in the list mempool right? Are any other mempool that would need this? If so we could add to the interface but that might be breaking

mempool/clist_mempool.go
```diff
@@ -608,7 +616,25 @@ func (mem *CListMempool) Update(
...
+						return
+					}
+				}
+			})
+		}
+
+		if doRecheck {
+			mem.recheckTxs()
+		} else {
+			mem.logger.Debug("recheck disabled at this height", "height", height)
```
andynog, line 636: This condition for recheck=false is when node is behind by a max based on the logic above so the debug message could be more related to that too

test/e2e/networks_regressions/slow_recheck.toml
```diff
@@ -0,0 +1,217 @@
+ipv6 = false
+initial_height = 0
+disable_pex = false
+key_type = ""
+abci_protocol = "tcp"
+prepare_proposal_delay = "200ms"
+process_proposal_delay = "200ms"
+check_tx_delay = "120ms"
```
andynog, line 8: It would be nice to start adding comments in these manifests, like say this parameters is responsible for the slow rechecks

andynog (commented):
lgtm, minor suggestions and questions.

p2p/peer.go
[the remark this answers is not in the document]
hvanz: This code was removed. Now it's not needed.

mempool/clist_mempool.go
hvanz: I'm asking this myself. As an example in the current code, the mempool reactor doesn't send transactions if there's a difference with the peer of more than one block.

test/e2e/networks_regressions/slow_recheck.toml
hvanz: Good remark. I added a comment to the manifest file.

mempool/clist_mempool.go
```diff
@@ -608,7 +616,25 @@ func (mem *CListMempool) Update(
 
 	// Recheck txs left in the mempool to remove them if they became invalid in the new state.
 	if mem.config.Recheck {
-		mem.recheckTxs()
+		// Don't perform recheck if one of the peers is known to be at least at our height +
+		// maxHeightDiffForRechecking.
+		doRecheck := true
+		if mem.sw != nil {
+			mem.sw.Peers().ForEach(func(peer p2p.Peer) {
```
melekes, line 623: This could be a method on the Switch.

melekes: Possible attack: I construct a node, which reports height=10000000, but otherwise appears to be valid.

melekes (changes requested):
I appreciate the effort here, but I wouldn't say I like that:

a) `Switch` is added to the mempool
b) New attack vectors are introduced (see my comment)

The general direction is okay, IMHO. We need to inform Comet components that the node is syncing (either via block syncing or consensus). Previously, we did similar things via pubsub to avoid tight coupling of components.

cason:
> In particular, rechecking may take a lot of time to finish, blocking the mempool from receiving new transactions, and not allowing to node to ever catch up.

Well, the first is non-problem, except for the fact that the mempool will be blocking the connections with peers, if they continue to send us transactions (while they shouldn't). The problem here is the mempool blocking the connection with the peers.

The second problem probably derives partially from the first. If the mempool can't process messages, thus blocking the channel, consensus/block sync messages are not processed as well. There is also the fact that block execution takes longer, but this should not be the core problem here, in my opinion.

From this analysis, I think the most general solution is not to block the mempool channel, namely, to drop transactions received while the node is busy with recheck or other stuff. Or do not block the mempool while rechecking is happening.

In any case, this something to discuss on an issue, not a PR.

mempool/clist_mempool.go
cason: Agree with the attack, we cannot trust individual reports of height.

Notice that we use the reported height to decide what to send to a peer, not to define our own behavior as a node.

mempool/clist_mempool.go
cason: This is pretty arbitrary, making a config flag probably won't help much, although it is not a bad idea.

In any case, 2 is a lower value. We are talking about a node that is slightly behind and it can catch-up. The problem is when it is 10 or more heights behind, I would say.

mempool/clist_mempool.go
cason: I wouldn't change the interface, as this is to me a hopefully temporary workaround.

mempool/clist_mempool.go
```diff
@@ -608,7 +616,25 @@ func (mem *CListMempool) Update(
 
 	// Recheck txs left in the mempool to remove them if they became invalid in the new state.
 	if mem.config.Recheck {
-		mem.recheckTxs()
+		// Don't perform recheck if one of the peers is known to be at least at our height +
```
cason, line 619: This is a case for creating a new method to do this verification and clear the code of this method from this workaround. In this way, if we find a better way to implement this, we change the new method, not this one.

mempool/clist_mempool.go
cason: And this should be an INFO message.

cason (commented):
Still consider this a workaround, agree that it can be abused.

hvanz:
Thanks for all the feedback! I agree with all what was said above. I also don't like the approach of passing the switch to the mempool and depending on the height that peers inform. This was mostly a rushed attempt to try to fix the nightly tests. What we can get from this is that disabling rechecking temporarily could be a solution to some of the known problems. At least it makes the nightly tests pass.

hvanz:
Results of nightly tests: https://github.com/cometbft/cometbft/actions/runs/9696318374

mempool/clist_mempool.go
hvanz: The main problem with this attack is that transactions may get stuck in the mempool, because with PrepareProposal rechecking is needed to do garbage collection. Otherwise the transactions will be validated when proposing.

Issue #3198, opened 2024-06-06
Secret Connection: Buffer the underlying connection


ValarDragon:
### Summary

A secret connection read call, no matter how many bytes its serving, is only reading 1024 bytes from disk as a syscall. Many packets are much larger, so we end up doing something like:

1) sc.Read(10kb)
2) Syscall read 1048 bytes (encrypted data frame)
3) SecretConn decrypt
4) Return 1024 bytes (data-component of a data frame)
5) sc.Read(9kb)

So we will net do 10 syscalls for a 10kb read. Note that things like a full block part are (encoded) > 64kb, and therefore require 64 syscalls! 

### Proposal

Use a bufferred reader over the connection within secret connection. I tested it out on osmosis mainnet with this code: https://github.com/osmosis-labs/cometbft/commit/3e2254f58cfaffc612e721e3246c4d53ca070df3 and saw a 10% speedup to secret connection file open times. (There were relatively small blocks on osmosis at this time) TBH I was expecting a higher speedup than 10%

I'd make a PR, but the evil secret connection tests are really messy and will take some time to fix.

ValarDragon:
This also matters on the write side

PR #3346, opened 2024-06-26
perf(p2p/conn): Buffer secret connection writes

ValarDragon:
component of #3198 , this PR buffers writes. What happens is secret conn will often receive a write of (say) 65kb. But it will then split this into 64 1024 byte frames. It does a write on each frame, which is a syscall write. Then starts the next frame. Instead here, we buffer these writes, and do a single syscall write at the end.

I'll come back with benchmarks from running on mainnet. The baseline for this is netconn.Write taking 33% of the time in sendroutine. 
image
image

We should do something similar for reads, but due to how the evil_secret_connection test makes non-black-box use of conn, that will require notable test refactors.

---

#### PR checklist

- [ ] Tests written/updated


zmanian (approved):

zmanian:
I think at very least we should also make a breaking change to increase the frame size from 1k.

If we aren't going to adaptive frames....

p2p/conn/secret_connection.go
```diff
@@ -42,6 +43,9 @@ const (
 	labelEphemeralUpperPublicKey = "EPHEMERAL_UPPER_PUBLIC_KEY"
 	labelDHSecret                = "DH_SECRET"
 	labelSecretConnectionMac     = "SECRET_CONNECTION_MAC"
+
+	defaultWriteBufferSize = 128 * 1024
+	defaultReadBufferSize  = 65536
```
melekes, line 48: Unused?

melekes (approved):
Thanks @ValarDragon ❤️

p2p/conn/evil_secret_connection_test.go
```diff
@@ -222,12 +223,14 @@ func (c *evilConn) signChallenge() []byte {
 
 	b := &buffer{}
 	c.secretConn = &SecretConnection{
-		conn:       b,
-		recvBuffer: nil,
-		recvNonce:  new([aeadNonceSize]byte),
-		sendNonce:  new([aeadNonceSize]byte),
-		recvAead:   recvAead,
-		sendAead:   sendAead,
+		underlyingConn: b,
```
cason, line 226: I wouldn't change the name of this variable, because it does make much sense.

p2p/conn/evil_secret_connection_test.go
```diff
@@ -222,12 +223,14 @@ func (c *evilConn) signChallenge() []byte {
...
+		connReader:     b,
+		connWriter:     bufio.NewWriterSize(b, 65536),
```
cason, line 228: This is the new part, agree, the size of the buffer is something to have a as constant or at least discuss?

p2p/conn/secret_connection.go
```diff
@@ -140,12 +147,14 @@ func MakeSecretConnection(conn io.ReadWriteCloser, locPrivKey crypto.PrivKey) (*
 	}
 
 	sc := &SecretConnection{
-		conn:       conn,
```
cason, line 143: The previous comment applies to here: keep `conn` as underlying connection, please.

p2p/conn/secret_connection.go
```diff
@@ -280,16 +290,19 @@ func (sc *SecretConnection) Read(data []byte) (n int, err error) {
 }
 
 // Implements net.Conn.
-func (sc *SecretConnection) Close() error                  { return sc.conn.Close() }
```
cason, line 283: This changes only derive from the renaming `conn` -> `underlyingConn`. They don't change behaviour.

cason (commented):
I would request to make this PR simple by just adding the `connWriter` to buffer writes, which is good in general.

ValarDragon:
> I think at very least we should also make a breaking change to increase the frame size from 1k.
> If we aren't going to adaptive frames....

Agreed! I'm down to (in separate PRs) increase frame size, at least to have that in next coordinate release if putting in another secret transport layer doesn't work out. (Or at minimum raising frame size) The amount of chacha20poly1305 overhead was really surprising to me though, I hope that improves proportionately with larger frames

p2p/conn/secret_connection.go
```diff
@@ -42,6 +43,8 @@ const (
 	labelEphemeralUpperPublicKey = "EPHEMERAL_UPPER_PUBLIC_KEY"
 	labelDHSecret                = "DH_SECRET"
 	labelSecretConnectionMac     = "SECRET_CONNECTION_MAC"
+
+	defaultWriteBufferSize = 128 * 1024
```
ValarDragon, line 47: Re suggested buffer size: The maximum single message size is the encoding of a block part, which is 65kb. I set it to the size of two block parts, in case there is something else in the queue to be written

.changelog/unreleased/improvements/3346-buffer-secret-connection-writes.md
```diff
@@ -0,0 +1,2 @@
+- `[p2p/secretconn]` Speedup secretconnection large writes, by buffering the write to the underlying connection.
```
cason, line 1:
```suggestion
- `[p2p/conn]` Speedup secret connection large writes, by buffering the write to the underlying connection.
```

p2p/conn/secret_connection.go
cason: The default buffer size for a writer is 4k:

https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/bufio/bufio.go;l=19

I wonder whether is not better to the let the Writer to increase the buffer size on demand....

cason (approved):
Good.

Can we really measure the improvements that this provides?

ValarDragon:
Yes will get it! Sorry for delay, two silly issues got in the way. (Snapshot service was down when I first did it, and I didn't download result in time during my second profile so it get retented. Re-running mainnet benchmarks now to get the ratio here)

ValarDragon:
This successfully eliminates the overhead coming from the write packet, but not flush (which makes sense as flush is dealing with the case where we can't fill one frame anyway)

We should expect the ratio of seal to file.Write in the `sendPacketMsg` call (the non-flush throttle case) to be indicative of what we are speeding up. So on this latest benchmark, its a ratio of 2.5 seal : 1.5 file write. Normalized, 1s of sealing needed .6s of net con write.

Originally this case was: 1s of sealing, needed 1.73s of net conn write. So this is an almost 3x speedup to this part of the bottleneck

p2p/conn/secret_connection.go
```diff
@@ -280,10 +287,13 @@ func (sc *SecretConnection) Read(data []byte) (n int, err error) {
...
 
 // Implements net.Conn.
-func (sc *SecretConnection) Close() error                  { return sc.conn.Close() }
-func (sc *SecretConnection) LocalAddr() net.Addr           { return sc.conn.(net.Conn).LocalAddr() }
-func (sc *SecretConnection) RemoteAddr() net.Addr          { return sc.conn.(net.Conn).RemoteAddr() }
-func (sc *SecretConnection) SetDeadline(t time.Time) error { return sc.conn.(net.Conn).SetDeadline(t) }
+func (sc *SecretConnection) Close() error         { return sc.conn.Close() }
+func (sc *SecretConnection) LocalAddr() net.Addr  { return sc.conn.(net.Conn).LocalAddr() }
+func (sc *SecretConnection) RemoteAddr() net.Addr { return sc.conn.(net.Conn).RemoteAddr() }
+func (sc *SecretConnection) SetDeadline(t time.Time) error {
```
cason, line 293:
```suggestion
func (sc *SecretConnection) SetDeadline(t time.Time) error { return sc.conn.(net.Conn).SetDeadline(t) }
```

p2p/conn/secret_connection.go
```diff
@@ -280,10 +287,13 @@ func (sc *SecretConnection) Read(data []byte) (n int, err error) {
...
+	return sc.conn.(net.Conn).SetDeadline(t)
```
cason, line 294:
```suggestion

```

p2p/conn/secret_connection.go
```diff
@@ -280,10 +287,13 @@ func (sc *SecretConnection) Read(data []byte) (n int, err error) {
...
+}
```
cason, line 295:
```suggestion
```

cason (commented):
Minor suggestions. Lets merge this.

p2p/conn/secret_connection.go
ValarDragon: The buffer's don't increase buffer size on demand though, after its too big, it just goes directly to the underlying writer (which is what we'd like to avoid!)

melekes:
@mergify backport v1.x

PR #3353, opened 2024-06-27
test(e2e): add option to the 'runner logs' command to output logs separately

andynog:
When conducting end-to-end testing, it can be difficult to effectively analyze or visualize the output of docker logs when all the containers are merged into a single log file.

To address this issue, I have introduced a new parameter (`--split`) to the runner `logs` command. When this parameter is used, the command will output the log for each container separately.

Having separate logs makes it much easier to troubleshoot individual containers, particularly when the log output is extensive and detailed. This feature boosts efficiency by simplifying the process of identifying the root cause of failures.

Usage:
```sh
 ./build/runner logs -f ./networks/simple.toml --split
```
#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/test/e2e/runner/main.go
+++ b/test/e2e/runner/main.go
@@ -287,11 +287,10 @@ func NewCLI() *CLI {
 	var splitLogs bool
 	logCmd := &cobra.Command{
 		Use:   "logs",
-		Short: "Shows the testnet logs",
+		Short: "Shows the testnet logs. Use `--split` to split logs into separate files",
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			splitLogs, _ = cmd.Flags().GetBool("split")
 			if splitLogs {
-				fmt.Println("Output separate logs for each node")
 				for _, node := range cli.testnet.Nodes {
 					fmt.Println("Log for", node.Name)
 					err := docker.ExecComposeVerbose(context.Background(), cli.testnet.Dir, "logs", node.Name)
@@ -301,7 +300,6 @@ func NewCLI() *CLI {
 				}
 				return nil
 			}
-			fmt.Println("Output combined log for all nodes")
 			return docker.ExecComposeVerbose(context.Background(), cli.testnet.Dir, "logs")
 		},
 	}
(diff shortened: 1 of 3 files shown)


test/e2e/runner/main.go
```diff
@@ -284,13 +284,27 @@ func NewCLI() *CLI {
...
+				for _, node := range cli.testnet.Nodes {
+					fmt.Println("Log for", node.Name)
+					err := docker.ExecComposeVerbose(context.Background(), cli.testnet.Dir, "logs", node.Name)
+					if err != nil {
+						return err
+					}
+				}
+				return nil
+			}
 			return docker.ExecComposeVerbose(context.Background(), cli.testnet.Dir, "logs")
```
melekes, line 303: This will still produce one big log, right? Shouldn't we only execute this if `!splitLogs`

andynog: What do you mean it will produce one big log ? If you run the command it will output the logs for each container. So you'd get something like:

```
...
full01  | I[2024-06-28|18:58:02.484] Committed state                              module=state height=32 block_app_hash=52FED6073B1AAB6D6ED76E81B14CD00A158C87D8591413C6C47C4091939B0E86
full01  | E[2024-06-28|18:58:02.484] Failed to set application block retain height module=state retainHeight=5 err="invalid height value"
full01  | I[2024-06-28|18:58:02.484] indexed block events                         module=txindex height=32
Log for full02
full02  | I[2024-06-28|18:57:34.700] Application started!                         
full02  | I[2024-06-28|18:57:34.701] start app                                    msg="Server listening on unix:///var/run/app.sock (socket protocol)"
full02  | I[2024-06-28|18:57:35.661] deprecated usage found in configuration file usage="[consensus.timeout_commit] is deprecated. Use `next_block_delay` in the ABCI `FinalizeBlockResponse`."
full02  | I[2024-06-28|18:57:35.685] State store key layout version               module=main version=vv1
full02  | I[2024-06-28|18:57:35.686] Blockstore version                           module=main version=v1`
```

The single one (line 303) will not be executed because of the `nil` return if that's what you mean.

andynog: the output is more like a log but sorted by containers instead of having a mixed one where it's harder to find the logs just for a particular container in case you want to troubleshoot.

melekes (approved):
👍

hvanz (approved):

PR #3355, opened 2024-06-28
fix(rpc): fix nil pointer error in `/tx` and `/tx_search`

### LINKED ISSUE #3352: Nil Pointer Dereference Error When Querying Transactions with Proof in CometBFT
## Bug Report

**ABCI app**: any CosmosSDK chain

### What happened?

When querying a transaction from LCD, that has min_retain_blocks set, the result is nil pointer error for transactions from blocks that had been deleted away.

for example:

https://api-injective-ia.cosmosia.notional.ventures/cosmos/tx/v1beta1/txs/71B0034D463EC79CB0011C7F4EA9EE3B39342AF6BE40AA0402C672A43B21DA3B

`{"code":2,"message":"runtime error: invalid memory address or nil pointer dereference: panic","details":[]}`

The transaction itself is present in the CometBFT, which can be proved by calling the TX  RPC endpoint directly with and without a proof.

https://rpc-injective-ia.cosmosia.notional.ventures/tx?hash=0x71B0034D463EC79CB0011C7F4EA9EE3B39342AF6BE40AA0402C672A43B21DA3B&prove=true

```
{
  "jsonrpc": "2.0",
  "id": -1,
  "error": {
    "code": -32603,
    "message": "Internal error",
    "data": "runtime error: invalid memory address or nil pointer dereference"
  }
}

```

https://rpc-injective-ia.cosmosia.notional.ventures/tx?hash=0x71B0034D463EC79CB0011C7F4EA9EE3B39342AF6BE40AA0402C672A43B21DA3B&prove=false

```
{
  "jsonrpc": "2.0",
  "id": -1,
  "result": {
    "hash": "71B0034D463EC79CB0011C7F4EA9EE3B39342AF6BE40AA0402C672A43B21DA3B",
    "height": "75002060",
    "index": 5,
    "tx_result": {
      "code": 0,
      "data": 
omitted
```

Root cause for the nil pointer error can be found here. When block is no longer available, LoadBlock returns nil Block, which triggers the panic on the next line.
https://github.com/cometbft/cometbft/blob/7f24fd2b41c3fc734e6020b4d9a46fa07bebc4db/rpc/core/tx.go#L116-L117

I expect an error stating that proof is not available, since block had been pruned away.

melekes:
when block is absent
Closes #3352

---

#### PR checklist

### DIFF
--- a/.changelog/unreleased/bug-fixes/3352-nil-pointer-tx-search.md
+++ b/.changelog/unreleased/bug-fixes/3352-nil-pointer-tx-search.md
@@ -1,0 +1,2 @@
+- `[rpc]` Fix nil pointer error in `/tx` and `/tx_search` when block is
+  absent ([\#3352](https://github.com/cometbft/cometbft/issues/3352))
--- a/rpc/core/tx.go
+++ b/rpc/core/tx.go
@@ -36,7 +36,9 @@ func (env *Environment) Tx(_ *rpctypes.Context, hash []byte, prove bool) (*ctype
 	var proof types.TxProof
 	if prove {
 		block, _ := env.BlockStore.LoadBlock(r.Height)
-		proof = block.Data.Txs.Proof(int(r.Index))
+		if block != nil {
+			proof = block.Data.Txs.Proof(int(r.Index))
+		}
 	}
 
 	return &ctypes.ResultTx{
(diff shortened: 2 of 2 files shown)


cason (approved):
+1

PR #3360, opened 2024-06-28
fix(p2p/pex): respect MaxNumOutboundPeers limit while dialing peers provided by a seed node

### LINKED ISSUE #486: p2p: dialing addresses provided by a seed node disregards `MaxNumOutboundPeers` parameter
Migrated from tendermint/tendermint#9548.

The method [`ensurePeers`](https://github.com/cometbft/cometbft/blob/main/p2p/pex/pex_reactor.go#L453) defines when a node should dial peer addresses in order to establish outbound peer connections. It dials new peers when the number of outbound peers plus the number of ongoing connections attempts is lower than the configured `MaxNumOutboundPeers` parameter. If the node defines it has to dial peers but it cannot find, in its address book, any suitable peer to dial, this method will submit a PEX request to the configured seed nodes in order to retrieve candidate peer addresses.

The `ensurePeers` method is invoked periodically, every `ensurePeersPeriod = 30s`. This means that between requesting peer addresses to a seed node and actually attempting to dial them, a node will wait for this interval, which was considered too long (tendermint/tendermint#2093). A fix was proposed in tendermint/tendermint#2115, where `ensurePeers` was immediately invoked when a node receives peer addresses provided by a seed node. This solution introduced another issue (tendermint/tendermint#3338): by executing `ensurePeers` in intervals below `ensurePeersPeriod`, a node could end up submitting PEX requests too often, which consists on a misbehavior.

To solve this issue, tendermint/tendermint#3762 replaced the invocation of `ensurePeers`, when receiving addresses from a seed node, by go routines that attempt connecting to _all_ peer addresses received from a seed node. The introduced code [excerpt](https://github.com/cometbft/cometbft/blob/main/p2p/pex/pex_reactor.go#L385), however, does not takes into consideration the number of connected peers or of connection attempts, nor the `MaxNumOutboundPeers` parameter. In other words, the logic implemented in `ensurePeers` to limit the number of outbound peers is ignored. As a result, upon resorting on a seed node to retrieve candidate peers to dial, a node may end up connected to _all_ peers informed by the seed node, which tends to be more than the configured `MaxNumOutboundPeers`, with default value of 10.

This breach in the limits of outbound peers was identified when writing the spec of the p2p layer (tendermint/tendermint#9089). In testnet experiments (tendermint/tendermint#9499), it was observed that nodes had more peers than the maximum configured number (`MaxNumOutboundPeers + MaxNumInboundPeers`, 50 by default). This issue represents the suspect that the reason for this is the breach above detailed. Reinforces this suspect the fact that nodes in the testnet resort by default to seed nodes to learn addresses of potential peers.

Suggested fix: restore the solution introduced by tendermint/tendermint#2115, while creating another way to prevent sending PEX requests to often to connected peers or seed nodes.

cason:
Closes #486.

Commits:

1. Make `TestConnectionSpeedForPeerReceivedFromSeed` to fail, as it will make the node to connect to more than the configured `MaxNumOutboundPeers`
2. Solve the issue by introducing a channel to wake up `ensurePeersRoutine()`. In this way, the node will not to wait until the next `defaultEnsurePeersPeriod` (30s) before attempting again to connect to new peers (`ensurePeers()` method) whose addresses were provide by the seed node.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/p2p/pex/pex_reactor.go
+++ b/p2p/pex/pex_reactor.go
@@ -367,9 +367,7 @@ func (r *Reactor) ReceiveAddrs(addrs []*p2p.NetAddress, src Peer) error {
 		if seedAddr.Equals(srcAddr) {
 			select {
 			case r.ensurePeersCh <- struct{}{}:
-				// Wake up ensurePeersRoutine()
 			default:
-				// But do not block
 			}
 			break
 		}
--- a/p2p/pex/pex_reactor.go
+++ b/p2p/pex/pex_reactor.go
@@ -66,6 +66,7 @@ type Reactor struct {
 
 	book              AddrBook
 	config            *ReactorConfig
+	ensurePeersCh     chan struct{} // Wakes up ensurePeersRoutine()
 	ensurePeersPeriod time.Duration // TODO: should go in the config
 	peersRoutineWg    sync.WaitGroup
 
@@ -115,6 +116,7 @@ func NewReactor(b AddrBook, config *ReactorConfig) *Reactor {
 	r := &Reactor{
 		book:                 b,
 		config:               config,
+		ensurePeersCh:        make(chan struct{}),
 		ensurePeersPeriod:    defaultEnsurePeersPeriod,
 		requestsSent:         cmap.NewCMap(),
 		lastReceivedRequests: cmap.NewCMap(),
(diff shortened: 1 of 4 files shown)


p2p/pex/pex_reactor.go
```diff
@@ -366,21 +360,18 @@ func (r *Reactor) ReceiveAddrs(addrs []*p2p.NetAddress, src Peer) error {
...
-						r.Logger.Debug(err.Error(), "addr", addr)
-					}
-				}
-			}(netAddr)
+	// Try to connect to addresses coming from a seed node without waiting (#2093)
+	for _, seedAddr := range r.seedAddrs {
+		if seedAddr.Equals(srcAddr) {
+			select {
+			case r.ensurePeersCh <- struct{}{}:
+				// Wake up ensurePeersRoutine()
```
melekes, line 370:
```suggestion
```

melekes (approved):
👍

melekes:
> Not sure whether we should backport this PR. It is not API breaking, but it changes the way that new nodes connect to peers. For instance, the number of peers of nodes in our QA experiments will change with this PR, as all nodes, except seed nodes, will have at most 10 outbound connections and at most 50 peers in total.

But the previous behavior is unintended, isn't it? If so, I think we should backport **the fix**.

cason:
Yes, it was unintended.

This also needs a changelog, which folder should I use?

cason:
Another point is that the removed code was also used when the node is in seed mode. In this case, it runs the `crawlPeersRoutine()` instead of the `ensurePeersRoutine()`. I can add the same logic in the `crawlPeersRoutine()` routine, although no test is failing at the moment.

Actually, when running in seed mode, apparently there is no maximum number of connections or stuff like that.

czarcas7ic (approved):
Thanks for addressing this so quickly, implementation looks great :)

sergio-mena (approved):
LGTM 👍

PR #3361, opened 2024-06-28
refactor(mempool)!: remove PreUpdate from Mempool interface and move its logic to Lock

hvanz:
This PR partially reverts the backport of #3314 into the recently released `v0.38.8` (and `v0.37.7`). With this change the `Mempool` interface is the same as in previous versions.

The reason is that we do not want to break the public API. We still keep in the code the feature that #3314 introduced by moving it inside the existing `Lock` method. We also keep the `RecheckFull bool` field that we added to `ErrMempoolIsFull`.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -152,19 +152,15 @@ func WithMetrics(metrics *Metrics) CListMempoolOption {
 
 // Safe for concurrent use by multiple goroutines.
 func (mem *CListMempool) Lock() {
+	if mem.recheck.setRecheckFull() {
+		mem.logger.Debug("the state of recheckFull has flipped")
+	}
 	mem.updateMtx.Lock()
 }
 
 // Safe for concurrent use by multiple goroutines.
 func (mem *CListMempool) Unlock() {
 	mem.updateMtx.Unlock()
-}
-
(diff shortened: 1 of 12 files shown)


mempool/clist_mempool.go
```diff
@@ -152,6 +152,9 @@ func WithMetrics(metrics *Metrics) CListMempoolOption {
 
 // Safe for concurrent use by multiple goroutines.
 func (mem *CListMempool) Lock() {
+	if mem.recheck.setRecheckFull() {
+		mem.logger.Debug("the state of recheckFull has flipped")
```
sergio-mena, line 156: I think this message is not very useful if it doesn't say from X to Y, as we need to trace back all occurrences of this message for a node to find out if it is now full, or no longer full.
I think this message would be better inside `recheck.setRecheckFull` where it can provide this info.

In any case, I don't know if it should be done here, or in a different PR from `main`, and then backported all the way down.

hvanz: Agree. I will do it from main in a different PR.

hvanz:
Just realised the changelog file is in the wrong directory. It should be `.changelog/unreleased/breaking-changes/3361-mempool-preupdate.md`.

PR #3363, opened 2024-06-28
refactor(mempool)!: remove PreUpdate from Mempool interface and move its logic to Lock

hvanz:
Complements #3361 

This PR partially reverts the backport of https://github.com/cometbft/cometbft/pull/3314 into the recently released v0.37.7. With this change the Mempool interface is the same as in previous versions.

The reason is that we do not want to break the public API. We still keep in the code the feature that https://github.com/cometbft/cometbft/pull/3314 introduced by moving it inside the existing Lock method. We also keep the RecheckFull bool field that we added to ErrMempoolIsFull.

---

#### PR checklist

- [ ] Tests written/updated


sergio-mena:
Checking diff-of-diffs, I see this comment, added to `Lock()`, has not been backported, 
```go
	// Before acquiring the lock, it signals the mempool that a new update is coming.
	// If the mempool is still rechecking at this point, it should be considered full.
```

sergio-mena (approved):
Other than the comment about the missing text on `Lock()`, LGTM 👍

hvanz:
Thanks, I added the missing comment.

PR #3371, opened 2024-06-29
feat(p2p): Filter ensurePeers addresses; chose larger toDial pool

czarcas7ic:
Filters addresses chosen prior to dialing them in order to increase likelyhood of successful connection. Also increases toDial addresses chosen in ensurePeers to greater than the defined numToDial. This way in the even of a failed dial, we have some reserve addresses to try that makes the most out of a single ensurePeers routine call (every 30s).

Through testing on Osmosis mainnet, the worst case runs have sped up connecting to defined max peers by over 5x.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/.changelog/unreleased/enhancements/3371-ensure-peers-improvements.md
+++ b/.changelog/unreleased/enhancements/3371-ensure-peers-improvements.md
@@ -1,2 +1,2 @@
-- `[p2p]` Improve various logic in the ensurePeers flow
+- `[p2p]` EnsurePeers improvements that speed up connection to the maximum defined number of peers
   ([\#3371](https://github.com/cometbft/cometbft/pull/3371))
--- a/.changelog/unreleased/enhancements/3371-ensure-peers-improvements.md
+++ b/.changelog/unreleased/enhancements/3371-ensure-peers-improvements.md
@@ -1,2 +1,2 @@
-- `[p2p]` EnsurePeers improvements that speed up connection to the maximum defined number of peers
+- `[p2p]` Filter ensurePeers addresses, and chose larger toDial pool to increase the chance of successful connection.
   ([\#3371](https://github.com/cometbft/cometbft/pull/3371))
--- a/.changelog/unreleased/enhancements/3371-ensure-peers-improvements.md
+++ b/.changelog/unreleased/enhancements/3371-ensure-peers-improvements.md
@@ -1,0 +1,2 @@
+- `[p2p]` Improve various logic in the ensurePeers flow
+  ([\#3371](https://github.com/cometbft/cometbft/pull/3371))
--- a/p2p/pex/addrbook.go
+++ b/p2p/pex/addrbook.go
@@ -58,7 +58,7 @@ type AddrBook interface {
 	Empty() bool
 
 	// Pick an address to dial
-	PickAddress(biasTowardsNewAddrs int) *p2p.NetAddress
+	PickAddress(biasTowardsNewAddrs int, filter func(*knownAddress) bool) *p2p.NetAddress
 
 	// Mark address
 	MarkGood(id p2p.ID)
@@ -269,7 +269,7 @@ func (a *addrBook) Empty() bool {
 // and determines how biased we are to pick an address from a new bucket.
 // PickAddress returns nil if the AddrBook is empty or if we try to pick
 // from an empty bucket.
-func (a *addrBook) PickAddress(biasTowardsNewAddrs int) *p2p.NetAddress {
+func (a *addrBook) PickAddress(biasTowardsNewAddrs int, filter func(*knownAddress) bool) *p2p.NetAddress {
 	a.mtx.Lock()
 	defer a.mtx.Unlock()
 
@@ -306,13 +306,23 @@ func (a *addrBook) PickAddress(biasTowardsNewAddrs int) *p2p.NetAddress {
 			bucket = a.bucketsNew[a.rand.Intn(len(a.bucketsNew))]
 		}
 	}
-	// pick a random index and loop over the map to return that index
-	randIndex := a.rand.Intn(len(bucket))
+
+	// Create a slice of known addresses from the bucket
+	addresses := make([]*knownAddress, 0, len(bucket))
 	for _, ka := range bucket {
-		if randIndex == 0 {
+		addresses = append(addresses, ka)
+	}
+
+	// Shuffle the addresses
+	a.rand.Shuffle(len(addresses), func(i, j int) {
+		addresses[i], addresses[j] = addresses[j], addresses[i]
+	})
+
+	// Iterate through the shuffled addresses and apply the filter until we find a suitable address
+	for _, ka := range addresses {
+		if filter == nil || filter(ka) {
 			return ka.Addr
 		}
-		randIndex--
 	}
 	return nil
 }
--- a/p2p/pex/addrbook.go
+++ b/p2p/pex/addrbook.go
@@ -58,7 +58,7 @@ type AddrBook interface {
 	Empty() bool
 
 	// Pick an address to dial
-	PickAddress(biasTowardsNewAddrs int, filter func(*knownAddress) bool) *p2p.NetAddress
+	PickAddress(biasTowardsNewAddrs int) *p2p.NetAddress
 
 	// Mark address
 	MarkGood(id p2p.ID)
@@ -269,7 +269,7 @@ func (a *addrBook) Empty() bool {
 // and determines how biased we are to pick an address from a new bucket.
 // PickAddress returns nil if the AddrBook is empty or if we try to pick
 // from an empty bucket.
-func (a *addrBook) PickAddress(biasTowardsNewAddrs int, filter func(*knownAddress) bool) *p2p.NetAddress {
+func (a *addrBook) PickAddress(biasTowardsNewAddrs int) *p2p.NetAddress {
 	a.mtx.Lock()
 	defer a.mtx.Unlock()
 
@@ -306,23 +306,13 @@ func (a *addrBook) PickAddress(biasTowardsNewAddrs int, filter func(*knownAddres
 			bucket = a.bucketsNew[a.rand.Intn(len(a.bucketsNew))]
 		}
 	}
-
-	// Create a slice of known addresses from the bucket
-	addresses := make([]*knownAddress, 0, len(bucket))
+	// pick a random index and loop over the map to return that index
+	randIndex := a.rand.Intn(len(bucket))
 	for _, ka := range bucket {
-		addresses = append(addresses, ka)
-	}
-
-	// Shuffle the addresses
-	a.rand.Shuffle(len(addresses), func(i, j int) {
-		addresses[i], addresses[j] = addresses[j], addresses[i]
-	})
-
-	// Iterate through the shuffled addresses and apply the filter until we find a suitable address
-	for _, ka := range addresses {
-		if filter == nil || filter(ka) {
+		if randIndex == 0 {
 			return ka.Addr
 		}
+		randIndex--
 	}
 	return nil
 }
--- a/p2p/pex/pex_reactor.go
+++ b/p2p/pex/pex_reactor.go
@@ -458,7 +458,7 @@ func (r *Reactor) ensurePeers() {
 	// Calculate reserve size as 50% of numToDial with a minimum of 10
 	reserveSize := cmtmath.MaxInt(numToDial/2, 10)
 
-	// Try maxAttempts times to pick numToDial addresses to dial
+	// Try maxAttempts times to pick numToDial + reserveSize addresses to dial
 	maxToDialAttempts := numToDial * 3
 
 	for i := 0; i < maxToDialAttempts && len(toDial) < numToDial+reserveSize; i++ {
@@ -487,7 +487,8 @@ func (r *Reactor) ensurePeers() {
 			wg        sync.WaitGroup
 		)
 
-		// Dial a subset of toDial, specifically aiming to respect the max connections set by the user.
+		// Dial a subset of toDial, in order to respect the max connections set by the user
+		// while still having a reserve of addresses to dial in case of failures.
 		subsetSize := cmtmath.MinInt(numToDial-successCount, len(toDial))
 		subset := make([]*p2p.NetAddress, 0, subsetSize)
 		for _, addr := range toDial {
--- a/p2p/pex/pex_reactor.go
+++ b/p2p/pex/pex_reactor.go
@@ -788,6 +788,10 @@ func markAddrInBookBasedOnErr(addr *p2p.NetAddress, book AddrBook, err error) {
 // If we are above the maximum number of dial attempts, the peer is marked as bad.
 // If either of the above conditions are met, an error is returned.
 func (r *Reactor) CheckDialStatus(addr *p2p.NetAddress, attempts int, lastDialedTime time.Time) error {
+	if !r.Switch.IsPeerPersistent(addr) && attempts > maxAttemptsToDial {
+		r.book.MarkBad(addr, defaultBanTime)
+		return ErrMaxAttemptsToDial{Max: maxAttemptsToDial}
+	}
 	if attempts > 0 {
 		jitter := time.Duration(cmtrand.Float64() * float64(time.Second)) // 1s == (1e9 ns)
 		backoffDuration := jitter + ((1 << uint(attempts)) * time.Second)
@@ -797,10 +801,6 @@ func (r *Reactor) CheckDialStatus(addr *p2p.NetAddress, attempts int, lastDialed
 			r.Logger.Debug("Skipping peer due to backoff", "addr", addr, "backoffDuration", backoffDuration)
 			return ErrTooEarlyToDial{backoffDuration, lastDialedTime}
 		}
-	}
-	if !r.Switch.IsPeerPersistent(addr) && attempts > maxAttemptsToDial {
-		r.book.MarkBad(addr, defaultBanTime)
-		return ErrMaxAttemptsToDial{Max: maxAttemptsToDial}
 	}
 	return nil
 }
--- a/p2p/pex/pex_reactor.go
+++ b/p2p/pex/pex_reactor.go
@@ -474,10 +474,10 @@ func (r *Reactor) ensurePeers() {
 
 	filter := func(ka *knownAddress) bool {
 		attempts, lastDialedTime := r.dialAttemptsInfo(ka.Addr)
-		if r.IsTooEarlyToDial(ka.Addr, attempts, lastDialedTime) {
+		if r.TooEarlyToDial(ka.Addr, attempts, lastDialedTime) != nil {
 			return false
 		}
-		if r.IsMaxAttemptsToDial(ka.Addr, attempts) {
+		if r.MaxAttemptsToDial(ka.Addr, attempts) != nil {
 			r.book.MarkBad(ka.Addr, defaultBanTime)
 			return false
 		}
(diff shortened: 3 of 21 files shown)


p2p/pex/addrbook.go
```diff
@@ -306,13 +306,23 @@ func (a *addrBook) PickAddress(biasTowardsNewAddrs int) *p2p.NetAddress {
...
+	}
+
+	// Shuffle the addresses
+	a.rand.Shuffle(len(addresses), func(i, j int) {
+		addresses[i], addresses[j] = addresses[j], addresses[i]
+	})
+
+	// Iterate through the shuffled addresses and apply the filter until we find a suitable address
+	for _, ka := range addresses {
+		if filter == nil || filter(ka) {
```
czarcas7ic, line 323: Instead of picking a peer from a random index, we shuffle the bucket and return the first peer that satisfies the filter.

p2p/pex/pex_reactor.go
```diff
@@ -464,45 +464,112 @@ func (r *Reactor) ensurePeers() {
...
-	maxAttempts := numToDial * 3
+	maxToDialAttempts := numToDial * 3
 
-	for i := 0; i < maxAttempts && len(toDial) < numToDial; i++ {
-		if !r.IsRunning() || !r.book.IsRunning() {
-			return
-		}
+	// Calculate reserve size as 50% of numToDial with a minimum of 10
+	reserveSize := cmtmath.MaxInt(numToDial/2, 10)
+	maxReserveAttempts := reserveSize * 3
```
czarcas7ic, line 473: There are many times when dialing the peer, we instantly get rejected. This PR includes filtered peer addresses that we can quickly try in the event a toDial address errors on dialing.

p2p/pex/pex_reactor.go
```diff
@@ -464,45 +464,112 @@ func (r *Reactor) ensurePeers() {
...
+		}
+		if _, selected := toDial[ka.Addr.ID]; selected {
+			return false
 		}
-		if _, selected := toDial[try.ID]; selected {
+		if _, selected := reserve[ka.Addr.ID]; selected {
+			return false
+		}
+		return true
+	}
```
czarcas7ic, line 494: Callback function for all desired filters when selecting the address from the address book.

p2p/pex/pex_reactor.go
```diff
@@ -536,25 +603,9 @@ func (r *Reactor) dialAttemptsInfo(addr *p2p.NetAddress) (attempts int, lastDial
...
-	// exponential backoff if it's not our first attempt to dial given address
-	if attempts > 0 {
-		jitter := time.Duration(cmtrand.Float64() * float64(time.Second)) // 1s == (1e9 ns)
-		backoffDuration := jitter + ((1 << uint(attempts)) * time.Second)
-		backoffDuration = r.maxBackoffDurationForPeer(addr, backoffDuration)
-		sinceLastDialed := time.Since(lastDialed)
-		if sinceLastDialed < backoffDuration {
-			return ErrTooEarlyToDial{backoffDuration, lastDialed}
-		}
-	}
```
czarcas7ic, line 554: This is now done in the callback function.

p2p/pex/pex_reactor.go
```diff
@@ -464,45 +464,113 @@ func (r *Reactor) ensurePeers() {
...
+		}
+		if _, selected := reserve[ka.Addr.ID]; selected {
+			return false
 		}
-		if _, selected := toDial[try.ID]; selected {
+		return true
+	}
+
+	var (
+		successCount int
```
melekes, line 497: I think in the idiomatic Go, these should be channels in which we put numbers, and on line 564, we sum up the results.

p2p/pex/pex_reactor.go
```diff
@@ -464,45 +464,113 @@ func (r *Reactor) ensurePeers() {
...
+	for i := 0; i < maxReserveAttempts && len(reserve) < reserveSize; i++ {
+		prospectivePeer := r.book.PickAddress(newBias, filter)
+		if prospectivePeer == nil {
 			continue
 		}
-		// TODO: consider moving some checks from toDial into here
-		// so we don't even consider dialing peers that we want to wait
-		// before dialing again, or have dialed too many times already
-		toDial[try.ID] = try
+		reserve[prospectivePeer.ID] = prospectivePeer
```
melekes, line 514: Is this set of more addresses to dial?

melekes (commented):
Nice work @czarcas7ic 👍

p2p/pex/addrbook.go
czarcas7ic: Before what we would do is, pick a random index and stick with it, even if it doesn't satisfy filter requirements. By shuffling, we can start at the top of the list and work our way down to the first address that satisfies our filter. It essentially keeps the randomness while allow our selection pool to be much bigger. 

I don't think we need to shuffle, it just felt like an easier way to keep the randomness while keeping the pool of possibilities high. Happy to switch it out for something else if it makes sense!

p2p/pex/pex_reactor.go
czarcas7ic: Imagine it as a subset. If a toDial address fails to dial (happens a LOT, maybe 60 percent of the time), we get a peer from the reserve pool, until the reserve pool is depleted.

I guess we could have just made toDial bigger, but it made more sense to me to imagine it this way, and hopefully we improve our addrbook heuristics more to where we can make the reserve pool smaller and smaller.

p2p/pex/addrbook.go
czarcas7ic: (in fact if I had to bet, I would say shuffling is less performant. Can you think of a better way to choose a random index, but keep choosing more and more random indexes until we pick a good candidate? I could think of a way if bucket wasn't a map and was instead an array. I guess I could make it an array and do many choices of random index, lmk what you think)

p2p/pex/pex_reactor.go
czarcas7ic: Done here, thanks! https://github.com/cometbft/cometbft/pull/3371/commits/40aaa94bbb89d71f5a0cf1e6d7b6b9f022b4814a

cason:
Two short comments here:

1. This is a `feat!` as it is a breaking change, for updating the `AddrBook` interface
2. If we need multiple addresses, from which to select some to dial, why not to use the `GetSelectionWithBias(biasTowardsNewAddrs int)` method of this interface, instead of changing the  `PickAddress(biasTowardsNewAddrs int)` method?

p2p/pex/pex_reactor.go
```diff
@@ -752,10 +818,33 @@ func (r *Reactor) attemptDisconnects() {
...
+	// TODO, blacklist peer if fmt.Errorf("peer is on a different network. Got %v, expected %v", other.Network, info.Network)
 	switch err.(type) {
 	case p2p.ErrSwitchAuthenticationFailure:
 		book.MarkBad(addr, defaultBanTime)
 	default:
 		book.MarkAttempt(addr)
 	}
 }
+
+func (r *Reactor) TooEarlyToDial(addr *p2p.NetAddress, attempts int, lastDialedTime time.Time) error {
```
cason, line 830: I would merge these two methods, right. All of them are about number of dial attempts.

p2p/pex/addrbook.go
melekes: Why not pick a random element and then check if it satisfies `filter`:

* if yes, then proceed with this element
* if not, pick another random element and check again

p2p/pex/pex_reactor.go
melekes: > I guess we could have just made toDial bigger

This is the part I don't understand imho

czarcas7ic: You don't understand as in you would have made toDial larger instead or you don't understand why toDial needs to be larger in the first place?

melekes: you would have made toDial larger instead

czarcas7ic: Sorry I'm still not following, you are implying we should not use reserve peers and instead just increase the size of to dial?

melekes: Yes, I am sorry for the confusion. My question is, why not increase the size of the `trial` set?

p2p/pex/addrbook.go
cason: Shuffle has the cost of a lock too, because of the `internal/rand` implementation.

By the way, this code also imports `math/rand`, so someone is not using our own implementation.

cason: The solution of taking random indexes in sequence is good if the probability of a random choice is good enough. If not, you are pretty likely to pick that index again. If you reject a lot of random choices, waiting for the random generator to pick a still not picked one is pretty low...

cason: But, again, why not to use the other method `GetSelection(bias)`. It does the shuffling, it returns a list of address with no repetitions, and you only need to call it once from the PEX reactor.

p2p/pex/pex_reactor.go
```diff
@@ -464,45 +464,116 @@ func (r *Reactor) ensurePeers() {
...
-			return
-		}
+	// Calculate reserve size as 50% of numToDial with a minimum of 10
+	reserveSize := cmtmath.MaxInt(numToDial/2, 10)
+	maxReserveAttempts := reserveSize * 3
 
-		try := r.book.PickAddress(newBias)
-		if try == nil {
-			continue
+	filter := func(ka *knownAddress) bool {
```
cason, line 475: This is something we could transform into a method. Notice that persistent peers should not be dialed here as well. We need to establish the conditions that allows the PEX reactor to dial a peer.

p2p/pex/pex_reactor.go
```diff
@@ -464,45 +464,116 @@ func (r *Reactor) ensurePeers() {
...
+			return false
+		}
+		if _, selected := reserve[ka.Addr.ID]; selected {
+			return false
 		}
-		if _, selected := toDial[try.ID]; selected {
+		return true
+	}
+
+	for i := 0; i < maxToDialAttempts && len(toDial) < numToDial; i++ {
```
cason, line 496: This becomes super inefficient. If we need to dial 10 peers, we are calling this method 30 times. 30 shuffles, 30 copies, 30 random selections, some of them just returning nil.

Again, I'd go for `GetSelection()`.

p2p/pex/pex_reactor.go
```diff
@@ -464,45 +464,116 @@ func (r *Reactor) ensurePeers() {
...
+		prospectivePeer := r.book.PickAddress(newBias, filter)
+		if prospectivePeer == nil {
 			continue
 		}
-		if r.Switch.IsDialingOrExistingAddress(try) {
+		toDial[prospectivePeer.ID] = prospectivePeer
+	}
+
+	for i := 0; i < maxReserveAttempts && len(reserve) < reserveSize; i++ {
```
cason, line 504: Again, another 30 calls to the same method....

p2p/pex/pex_reactor.go
```diff
@@ -464,45 +464,116 @@ func (r *Reactor) ensurePeers() {
...
 			if err != nil {
-				switch err.(type) {
-				case ErrMaxAttemptsToDial, ErrTooEarlyToDial:
-					r.Logger.Debug(err.Error(), "addr", addr)
-				default:
-					r.Logger.Debug(err.Error(), "addr", addr)
+				errorCh <- 1
+				r.Logger.Debug(err.Error(), "addr", addr)
+				// If reserve addresses are available, try to dial them due to this toDial address erroring out
+				for id, reserveAddr := range reserve {
```
cason, line 532: Are you aware that this is running in parallel, multiple threads accessing and updating `reserve` at the same time?

p2p/pex/pex_reactor.go
```diff
@@ -464,45 +464,116 @@ func (r *Reactor) ensurePeers() {
...
 		}(addr)
 	}
 
+	go func() {
+		wg.Wait()
+		close(successCh)
+		close(errorCh)
+	}()
+
+	go func() {
```
cason, line 559: I have the impression that, 30 seconds after, when we run `ensurePeers` again, this method could still be running. Not ideal...

p2p/pex/pex_reactor.go
```diff
@@ -752,10 +818,33 @@ func (r *Reactor) attemptDisconnects() {
 
 func markAddrInBookBasedOnErr(addr *p2p.NetAddress, book AddrBook, err error) {
 	// TODO: detect more "bad peer" scenarios
+	// TODO, blacklist peer if fmt.Errorf("peer is on a different network. Got %v, expected %v", other.Network, info.Network)
```
cason, line 821: This is an `ErrSwitchAuthenticationFailure` as below, isn't it?

.changelog/unreleased/enhancements/3371-ensure-peers-improvements.md
cason, line 1: This is a super generic content. We are changing the address book, breaking its API.

cason (changes requested):
I don't see a strong reason for breaking the `AddrBook` API to perform an operation (filter) that I am pretty sure we can still perform at the PEX reactor size. In particular, I would consider using address book methods that return a list of unique addresses, such as `GetSelection(bias)`.

Moreover, we don't have tests attesting that this works as expected. And we really need tests for this feature.

czarcas7ic:
Hey, thanks for the in depth review! I'm out of office for the next 3 days and then have a project to sprint on, but will for sure be able to make the requested changes. My apologies if some of the decisions didn't make a lot of sense, still trying to get better context of all the options available to me in comet.

p2p/pex/pex_reactor.go
czarcas7ic: Yes I can do this! Just implemented it here as I was visualizing it, have no issues moving it all to toDial

p2p/pex/addrbook.go
czarcas7ic: I did not know the other method existed, am looking into using this now!

czarcas7ic: I both removed the separate reserve map in favor of a larger toDial, as well as used GetSelection here https://github.com/cometbft/cometbft/pull/3371/commits/6cf054c314136ca0088bca05ac266bb85b8ee1cf

Still need to test it out, but seems like I definitely don't need to break the API to filter, thanks for the recommendation :)

p2p/pex/pex_reactor.go
czarcas7ic: Increased the size of toDial instead of the separate sets here https://github.com/cometbft/cometbft/pull/3371/commits/6cf054c314136ca0088bca05ac266bb85b8ee1cf

p2p/pex/pex_reactor.go
czarcas7ic: Combined here https://github.com/cometbft/cometbft/pull/3371/commits/7c12ee7fbbb1dfc38bf2f285060410693d3cfd39

p2p/pex/pex_reactor.go
czarcas7ic: Agreed, made its own method w/o breaking PickAddress API here https://github.com/cometbft/cometbft/pull/3371/commits/6cf054c314136ca0088bca05ac266bb85b8ee1cf#diff-8672eb1f866ffb54f7f8fed2f3bca88569ab29d9f9db1d74af08880677fa4eb9R808

p2p/pex/pex_reactor.go
czarcas7ic: https://github.com/cometbft/cometbft/pull/3371/commits/6cf054c314136ca0088bca05ac266bb85b8ee1cf

p2p/pex/pex_reactor.go
czarcas7ic: https://github.com/cometbft/cometbft/pull/3371/commits/6cf054c314136ca0088bca05ac266bb85b8ee1cf

p2p/pex/pex_reactor.go
czarcas7ic: Thanks for calling out, resolving as it is no longer relevant due to implementing your prior suggestion https://github.com/cometbft/cometbft/pull/3371/commits/6cf054c314136ca0088bca05ac266bb85b8ee1cf

p2p/pex/pex_reactor.go
czarcas7ic: I don't think so right? The error I am referring to is defined as ErrDifferentNetwork, I don't see where it would get wrapped as ErrSwitchAuthenticationFailure. That error appears as "failed to authenticate peer. Dialed %s"

.changelog/unreleased/enhancements/3371-ensure-peers-improvements.md
czarcas7ic: While we are no longer breaking API, I slightly expanded here https://github.com/cometbft/cometbft/pull/3371/commits/b936c7993513e3dacfdda715529dc02168ed0592

Feel free to resolve if that works

p2p/pex/pex_reactor.go
```diff
@@ -453,47 +453,91 @@ func (r *Reactor) ensurePeers() {
...
 
-		try := r.book.PickAddress(newBias)
-		if try == nil {
-			continue
-		}
-		if _, selected := toDial[try.ID]; selected {
+	// Try maxAttempts times to pick numToDial + reserveSize addresses to dial
+	maxToDialAttempts := numToDial * 3
+
+	for i := 0; i < maxToDialAttempts && len(toDial) < numToDial+reserveSize; i++ {
```
cason, line 464: I am trying to understand the logic of this for.

In the worst case, we run it `numToDial * 3` times.

Or we wait until `len(toDial)` to reach `numToDial+reserveSize`.

We add to `toDial` good peer candidates, according to the filtering method.

But how many peers does `r.book.GetSelectionWithBias(newBias)` return? I that up to 23% of the number of stored addresses, up to 250 addresses.

p2p/pex/pex_reactor.go
```diff
@@ -453,47 +453,91 @@ func (r *Reactor) ensurePeers() {
...
-			if err != nil {
-				switch err.(type) {
-				case ErrMaxAttemptsToDial, ErrTooEarlyToDial:
-					r.Logger.Debug(err.Error(), "addr", addr)
-				default:
+	toDialCount := len(toDial)
+	successCount := 0
+	errorCount := 0
+
+	for successCount < numToDial && len(toDial) > 0 {
```
cason, line 483: This for is now blocking, right? We don't leave `ensurePeers` until the conditions are observed, namely, we dialed everyone we were supposed to.

p2p/pex/pex_reactor.go
```diff
@@ -453,47 +453,91 @@ func (r *Reactor) ensurePeers() {
...
+		prospectivePeers := r.book.GetSelectionWithBias(newBias)
+		if prospectivePeers == nil {
 			continue
 		}
-		if r.Switch.IsDialingOrExistingAddress(try) {
-			continue
+		for _, prospectivePeer := range prospectivePeers {
+			if r.filterAddress(prospectivePeer, toDial) {
+				toDial[prospectivePeer.ID] = prospectivePeer
+				if len(toDial) >= numToDial+reserveSize {
```
cason, line 472: Since we have this break here, I think the for condition is redundant.

cason (commented):
Thank you for the changes, really, great effort here.

I am just considering the impact of having a blocking `ensurePeers` method, that waits for all the connections to be established.

Plus, we need test units to really understand the impact of this change.

cason (approved):

cason:
Now there is a PEX reactor test failing, this is not good.

czarcas7ic:
I can look into it, just am on a rare OOO so might be slightly delayed

czarcas7ic:
I will for sure have unit tests prior to merging this, as well as testing against mainnet. If you would like I can set this to draft to reduce noise to the comet team and just ping you when 

1. Tests are created
2. Tested against mainnet
3. Currently failing test passes

cason:
We can take it over from here, if you don't mind.

But the test that is failing is due to some go routine leaked or some synchronization problem.

czarcas7ic:
Yeah totally don't mind if you take it over due to me being out. I feel bad for putting more work on you all, hope this contribution was at least somewhat helpful!

cason:
There is a test failing. The same happens locally, when testing this new code.

Essentially, we cannot stop the reactor/switch because we are probably blocking in the introduced code. In other words, `ensurePeers()` is blocking now, as it waits for all dialing go routines to return. While it does not return, the reactor cannot be stopped.

cason:
Created  #3960 to address the problem. Closing this PR for the moment.

PR #3382, opened 2024-07-01
perf(p2p): Only update send monitor once per batch packet msg send

ValarDragon:
Small optimization to outbound packet gossip, I expect this to be a 1-2% speedup to outbound packet gossip as is right now. Will test on mainnet soon

This is safe as outbound packet gossip is single threaded per peer as is right now. Technically makes the send monitor marginally less real time, but this is irrelevant as the send monitor works on 20ms sliding windows anyway

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/.changelog/unreleased/improvements/3382-single-send-monitor-per-packet.md
+++ b/.changelog/unreleased/improvements/3382-single-send-monitor-per-packet.md
@@ -1,2 +1,2 @@
-- `p2p` perf(p2p): Only update send monitor once per batch packet msg send
+- `[p2p]` Only update send monitor once per batch packet msg send
   ([\#3382](https://github.com/cometbft/cometbft/pull/3382))
--- a/.changelog/unreleased/improvements/3382-single-send-monitor-per-packet.md
+++ b/.changelog/unreleased/improvements/3382-single-send-monitor-per-packet.md
@@ -1,2 +1,2 @@
-- `[p2p]` Only update send monitor once per batch packet msg send
+- `[p2p/conn]` Update send monitor, used for sending rate limiting, once per batch of packets sent
   ([\#3382](https://github.com/cometbft/cometbft/pull/3382))
--- a/.changelog/unreleased/improvements/3382-single-send-monitor-per-packet.md
+++ b/.changelog/unreleased/improvements/3382-single-send-monitor-per-packet.md
@@ -1,0 +1,2 @@
+- `p2p` perf(p2p): Only update send monitor once per batch packet msg send
+  ([\#3382](https://github.com/cometbft/cometbft/pull/3382))
--- a/p2p/conn/connection.go
+++ b/p2p/conn/connection.go
@@ -522,6 +522,11 @@ func (c *MConnection) sendSomePacketMsgs(w protoio.Writer) bool {
 func (c *MConnection) sendBatchPacketMsgs(w protoio.Writer, batchSize int) bool {
 	// Send a batch of PacketMsgs.
 	totalBytesWritten := 0
+	defer func() {
+		if totalBytesWritten > 0 {
+			c.sendMonitor.Update(totalBytesWritten)
+		}
+	}()
 	for i := 0; i < batchSize; i++ {
 		channel := selectChannelToGossipOn(c.channels)
 		// nothing to send across any channel.
--- a/p2p/conn/connection.go
+++ b/p2p/conn/connection.go
@@ -539,7 +539,6 @@ func (c *MConnection) sendBatchPacketMsgs(w protoio.Writer, batchSize int) bool 
 		}
 		totalBytesWritten += bytesWritten
 	}
-	c.sendMonitor.Update(totalBytesWritten)
 	return false
 }
 
@@ -579,7 +578,7 @@ func (c *MConnection) sendPacketMsgOnChannel(w protoio.Writer, sendChannel *Chan
 		c.stopForError(err)
 		return n, true
 	}
-	// TODO: Change this to only do one update for the entire bawtch.
+	// TODO: Change this to only add flush signals at the start and end of the batch.
 	c.flushTimer.Set()
 	return n, false
 }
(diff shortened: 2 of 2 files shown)


.changelog/unreleased/improvements/3382-single-send-monitor-per-packet.md
```diff
@@ -0,0 +1,2 @@
+- `p2p` perf(p2p): Only update send monitor once per batch packet msg send
```
melekes, line 1:
```suggestion
- `[p2p]` Only update send monitor once per batch packet msg send
```

p2p/conn/connection.go
```diff
@@ -562,18 +565,18 @@ func selectChannelToGossipOn(channels []*Channel) *Channel {
 	return leastChannel
 }
 
-func (c *MConnection) sendPacketMsgOnChannel(w protoio.Writer, sendChannel *Channel) bool {
+// returns (num_bytes_written, error_occurred).
+func (c *MConnection) sendPacketMsgOnChannel(w protoio.Writer, sendChannel *Channel) (int, bool) {
 	// Make & send a PacketMsg from this channel
 	_n, err := sendChannel.writePacketMsgTo(w)
```
melekes, line 571:
```suggestion
	n, err := sendChannel.writePacketMsgTo(w)
```

p2p/conn/connection.go
```diff
@@ -562,18 +565,18 @@ func selectChannelToGossipOn(channels []*Channel) *Channel {
...
 	if err != nil {
 		c.Logger.Error("Failed to write PacketMsg", "err", err)
 		c.stopForError(err)
-		return true
+		return _n, true
```
melekes, line 575:
```suggestion
		return n, true
```

p2p/conn/connection.go
```diff
@@ -562,18 +565,18 @@ func selectChannelToGossipOn(channels []*Channel) *Channel {
...
 	}
 	// TODO: Change this to only do one update for the entire bawtch.
-	c.sendMonitor.Update(_n)
 	c.flushTimer.Set()
-	return false
+	return _n, false
```
melekes, line 579:
```suggestion
	return n, false
```

melekes (approved):
Thanks @ValarDragon ❤️

cason (commented):
Good. By I wondering if we are preserving the same behavior as before, in particular in terms of the error handling.

In short, if we return at line 533 we are not updating the monitor at all, even with the bytes sent in previously and successful calls. The same is valid for line 529, isn't it?

The solution I suggest is to update the `sendMonitor` as part of a `defer` clause, it is done in any case.

p2p/conn/connection.go
```diff
@@ -521,17 +521,20 @@ func (c *MConnection) sendSomePacketMsgs(w protoio.Writer) bool {
 // Returns true if messages from channels were exhausted.
 func (c *MConnection) sendBatchPacketMsgs(w protoio.Writer, batchSize int) bool {
 	// Send a batch of PacketMsgs.
+	totalBytesWritten := 0
 	for i := 0; i < batchSize; i++ {
 		channel := selectChannelToGossipOn(c.channels)
 		// nothing to send across any channel.
 		if channel == nil {
 			return true
```
cason, line 529: We might have sent data in previous iterations of this loop. But since we don't have new messages to send we return, not updating the monitor at all.

p2p/conn/connection.go
```diff
@@ -521,17 +521,20 @@ func (c *MConnection) sendSomePacketMsgs(w protoio.Writer) bool {
...
 		}
-		err := c.sendPacketMsgOnChannel(w, channel)
+		bytesWritten, err := c.sendPacketMsgOnChannel(w, channel)
 		if err {
 			return true
 		}
+		totalBytesWritten += bytesWritten
 	}
+	c.sendMonitor.Update(totalBytesWritten)
```
cason, line 537: Move to before the `for` loop and make something like:

```
if totalBytesWritten > 0 {
    c.sendMonitor.Update(totalBytesWritten)
}
```

ValarDragon:
Great catch, fixed!

p2p/conn/connection.go
cason: Now you have to remove this line.

p2p/conn/connection.go
```diff
@@ -521,17 +521,25 @@ func (c *MConnection) sendSomePacketMsgs(w protoio.Writer) bool {
...
```
cason, line 542:
```suggestion

```

p2p/conn/connection.go
```diff
@@ -562,18 +570,18 @@ func selectChannelToGossipOn(channels []*Channel) *Channel {
...
 	// Make & send a PacketMsg from this channel
-	_n, err := sendChannel.writePacketMsgTo(w)
+	n, err := sendChannel.writePacketMsgTo(w)
 	if err != nil {
 		c.Logger.Error("Failed to write PacketMsg", "err", err)
 		c.stopForError(err)
-		return true
+		return n, true
 	}
 	// TODO: Change this to only do one update for the entire bawtch.
```
cason, line 582:
```suggestion
```

cason: TODO solved, right?

.changelog/unreleased/improvements/3382-single-send-monitor-per-packet.md
```diff
@@ -0,0 +1,2 @@
+- `[p2p]` Only update send monitor once per batch packet msg send
```
cason, line 1:
```suggestion
- `[p2p/conn]` Update send monitor, used for sending rate limiting, once per batch of packets sent
```

p2p/conn/connection.go
cason: Otherwise we update it twice.

cason (approved):
Good.

There is a mandatory fix, line 542 to be removed, otherwise the behavior will be wrong.

I think we would need to have some test units for this rate limiting implementation...

p2p/conn/connection.go
ValarDragon: This still applies for the flush monitor

p2p/conn/connection.go
ValarDragon: Sorry about that :sweat_smile:

melekes:
@mergify backport v1.x

PR #3389, opened 2024-07-01
fix(e2e): test case `TestBlock_Range` is flaky because of pruning

sergio-mena:
This PR is an attempt to solve -- once and for all -- the flakiness in `e2e` test case `TestBlock_Range`. The sequence of commits in this PR shows the repro steps, the fix, and the final contents of the patch. It's not necessary to review it commit by commit, unless the reviewer wants to check the steps taken to reproduce the flakiness.

* Commits 1 to 3: Infrastructure + temporary modifications to have the flakiness of `TestBlock_Range` show at every run. Also, some further fixes in the checks of `TestBlock_Range` are improved.
* Commit 4: Reduce `e2e` tests to be run so that the GitHub action doesn't time out. This commit reproduces the flakiness: click the ❌ next to this commit below.
* Commit 5: The fix. The test now passes, even with all the infra to repro in place. Check the ✅ next to the commit below.
* Commit 6: revert commit 4.
* Commit 7: Remove all the repro infra.
* Commit 8: Fix a typo that makes `e2e` CI tests fail.

All other commits: addressing issues in workflows and comments.

---

#### PR checklist

- [x] Tests written/updated


test/e2e/tests/block_test.go
```diff
@@ -55,14 +57,46 @@ func TestBlock_Header(t *testing.T) {
...
+			return first > first0
+		}, 1*time.Minute, 3*time.Second, "node %v is not pruning", node.Name)
+		require.Eventually(t, func() bool {
+			status, err := client.Status(ctx)
+			require.NoError(t, err)
+			first := status.SyncInfo.EarliestBlockHeight
+			last := status.SyncInfo.LatestBlockHeight
+			return last-first+1 < int64(node.RetainBlocks)+10 // Margin
+		}, 1*time.Minute, 3*time.Second, "node %v is not pruning enough", node.Name)
+	})
```
sergio-mena, line 89: I'm thinking of merging these two checks into 1, to make it more efficient

test/e2e/tests/block_test.go
```diff
@@ -55,14 +57,46 @@ func TestBlock_Header(t *testing.T) {
...
+}
+
 // Tests that the node contains the expected block range.
 func TestBlock_Range(t *testing.T) {
 	t.Helper()
 	testNode(t, func(t *testing.T, node e2e.Node) {
 		t.Helper()
 		// We do not run this test on seed nodes or nodes with data
```
melekes, line 97: comment needs to be updated

melekes (approved):
👍

test/e2e/tests/block_test.go
sergio-mena: Good catch! 🚀

sergio-mena:
d5d8082 is an attempt at making `e2e` workflow more efficient

test/e2e/tests/block_test.go
```diff
@@ -55,14 +57,41 @@ func TestBlock_Header(t *testing.T) {
...
 }
 
+// Tests that the node configured to prune are actually pruning.
+func TestBlock_Pruning(t *testing.T) {
+	t.Helper()
+	testNode(t, func(t *testing.T, node e2e.Node) {
+		t.Helper()
+		// We do not run this test on seed nodes or nodes with data
+		// companion-related pruning enabled or nodes not pruning.
+		if node.Mode == e2e.ModeSeed || node.EnableCompanionPruning || node.RetainBlocks == 0 {
```
andynog, line 67: Should this also exclude light nodes ?

```suggestion
		if node.Mode == e2e.ModeSeed || node.Mode == e2e.ModeLight || node.EnableCompanionPruning || node.RetainBlocks == 0 {
```

or use the `node.Stateless()` to exclude `light` and `seed` nodes
```suggestion
		if node.Mode == node.Stateless() || node.EnableCompanionPruning || node.RetainBlocks == 0 {
```

test/e2e/tests/block_test.go
```diff
@@ -55,14 +57,41 @@ func TestBlock_Header(t *testing.T) {
...
+			return
+		}
+		client, err := node.Client()
+		require.NoError(t, err)
+		status, err := client.Status(ctx)
+		first0 := status.SyncInfo.EarliestBlockHeight
+		require.NoError(t, err)
```
andynog, line 74: doesn't change the logic, more semantics to test right after the `err` return
```suggestion
		require.NoError(t, err)
		first0 := status.SyncInfo.EarliestBlockHeight
```

andynog (approved):
Looks good to me, made some suggestions and one comment about the regex in the workflow file

.github/workflows/e2e.yml
```diff
@@ -8,10 +8,15 @@ on:
   push:
     branches:
       - main
-      - release/**
+      - v[0-9].[0-9]+.x
```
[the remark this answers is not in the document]
sergio-mena: Thanks @andynog you're totally right!

sergio-mena, line 11:
```suggestion
      - v[0-9]+(\.[0-9]+)*\.x
```

test/e2e/tests/block_test.go
sergio-mena: Thing is, I don't fully understand why we were excluding seed nodes in the first place, so I just left it there. On the other hand, light nodes seem to be passing the test anyway, so I'd leave them in.
Maybe we should try removing the check for seed, and see if the test still passes. WDYT?

test/e2e/tests/block_test.go
sergio-mena: It's actually a bug. It will panic if it returns an error. Thanks for spotting!

test/e2e/networks_regressions/nodes_pruning.toml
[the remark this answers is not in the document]
sergio-mena: Do you have a suggested text?

.github/workflows/e2e.yml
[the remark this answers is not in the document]
sergio-mena: Mhhhh, actually, these are not regexes, the rules are different (check [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet)). Let me adapt the pattern to the actual rules

.github/workflows/e2e.yml
```diff
@@ -8,10 +8,15 @@ on:
   push:
     branches:
       - main
-      - release/**
+      - v[0-9]+(\.[0-9]+)*\.x
```
sergio-mena, line 11:
```suggestion
      - v0.3[478].x
      - v[1-9][0-9]?.x
```

.github/workflows/e2e.yml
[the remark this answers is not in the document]
andynog: Interesting, TIL, I thought it would be a regex but looks like it has its own filter pattern 👍

test/e2e/tests/block_test.go
sergio-mena: I tried removing the seed on one test case, and followed your advice on the other, and used `Stateless()`. I tried the CI manifeests several times and they are passing. Let me know if you don't agree with my last commit (01a504787).

andynog (approved):
lgtm

.github/workflows/e2e.yml
```diff
@@ -8,10 +8,16 @@ on:
   push:
     branches:
       - main
-      - release/**
+      - v0.3[478].x
+      - v[1-9][0-9]?.x
```
andynog, line 12: You might need to enclose them in double-quotes ?

This should work `"v[0-9]+.[0-9]+.[0-9]+" `

See an example here:   https://github.com/cometbft/cometbft/blob/8bfd04a6f23eec1b41393f0d67228b3f5942511e/.github/workflows/testapp-docker.yml#L9C9-L9C36

andynog (commented):
one more comment on the filter pattern

.github/workflows/e2e.yml
sergio-mena: As far as I read, the quotes are only necessary if the string starts with a special character ('*', '!') that may break the yaml syntax

melekes (approved):

jmalicevic:
I am triggering a  backport of this to v1.x to resolve the flakiness of the tests.

PR #3400, opened 2024-07-02
docs: ADR-117 `PrivValidator` should implement `cosmos.CryptoProvider`

raynaudoe:
This PR introduces **ADR-117** which implements the `CryptoProvider` [ADR](https://github.com/cosmos/crypto/blob/main/docs/architecture/adr-001-crypto-provider.md) ("aka pluggable cryptography")

The ADR in this repo is only a shortened version of the original one posted [here ](https://github.com/cosmos/cosmos-sdk/pull/18824) containing only the relevant concepts and code for cometBFT.

For the full text explaining all concepts please check this [file](https://github.com/cosmos/crypto/blob/main/docs/architecture/adr-001-crypto-provider.md)  (links are included in ADR-117)

---

#### PR checklist

- [ ] Tests written/updated


raynaudoe:
@sergio-mena here's the PR

docs/references/architecture/adr-117-multi-curve-support.md
```diff
@@ -0,0 +1,245 @@
...
+The implementation of ADR-001 in CometBFT will require a medium-level refactor of the codebase. Below we describe the proposed strategy to perform such implementation, which is:
+
+* Add a new `PrivValidator` implementation that uses the `CryptoProvider` interface underneath.
+
+* New directories reorganization:
+Implementations of crypto providers (previously `Privval` implementations) should be in their own directory. See the [Directories reorganization](#directories-reorganization) section for more details.
+
+* **[Optional]** Adding [Keyring](https://github.com/cosmos/cosmos-sdk/blob/main/crypto/keyring/keyring.go) and [Record](https://github.com/cosmos/cosmos-sdk/blob/main/proto/cosmos/crypto/keyring/v1/record.proto) for storing and loading providers.
+
+* **[Optional]** Use `Keyring` to load and instantiate validators when booting up a node.
```
melekes, line 42: Why would we want to load validators from the keyring? For additional safety?

docs/references/architecture/adr-117-multi-curve-support.md
```diff
@@ -0,0 +1,245 @@
...
+##### Code snippets for `CryptoProviderPV`
+
+As mentioned above, instead of having several implementations of `PrivValidator`, the proposal is to have only one that, by dependency injection, loads the corresponding `CryptoProvider` that offers the same functionality as the previous implementations of `PrivValidator`.
+
+Below is an example of how `CryptoProviderPV` would look like. Note that in this particular case, since the PrivateKey is managed inside the corresponding implementation, we're not passing that value to the signer. This is to avoid having to significantly change the code for `FilePV`. This is also valid for all implementations that manage their private keys in their own logic.
+
+```go
+// CryptoProviderPV is the implementation of PrivValidator using CryptoProvider's methods
+type CryptoProviderPV struct {
+    provider CryptoProvider
```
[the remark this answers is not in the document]
raynaudoe: Hi, thanks for the review! In the abstract you'll find the link, I'll post it here anyways:
https://github.com/Zondax/cometbft/blob/feat/adr-multi-curve/docs/references/architecture/adr-117-multi-curve-support.md#abstract
https://github.com/cosmos/crypto/blob/main/docs/architecture/adr-001-crypto-provider.md#crypto-provider

docs/references/architecture/adr-117-multi-curve-support.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Abstract
+
+This ADR proposes the implementation of [ADR-001-crypto-provider](https://github.com/cosmos/crypto/blob/main/docs/architecture/adr-001-crypto-provider.md). By refactoring the current `PrivValidator` implementations and replacing them with `CryptoProvider`s, we will allow for "pluggable cryptography" that enables a more modular design and future-proofs crypto tools.
+
+We will be extensively referencing concepts like CryptoProviders in this ADR. Please check the full ADR describing them [here](https://github.com/cosmos/crypto/blob/main/docs/architecture/adr-001-crypto-provider.md#crypto-provider).
+
+## Context
+
+The introduction of multi-curve support, by implementing the `CryptoProvider` interface and its components, in the CometBFT cryptographic package offers significant advantages. By not being restricted to a single cryptographic curve, developers can choose the most appropriate curve based on security, performance, and compatibility requirements. This flexibility enhances the application's ability to adapt to evolving security standards and optimizes performance for specific use cases, helping to future-proof the codebase's cryptographic capabilities.
```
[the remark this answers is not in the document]
raynaudoe: Please check the ADR defining those concepts [here](https://github.com/cosmos/crypto/blob/main/docs/architecture/adr-001-crypto-provider.md#crypto-provider)

docs/references/architecture/adr-117-multi-curve-support.md
raynaudoe: In short, yes. Is an optional feature to improve security

educlerici-zondax:
hey @melekes how are you? were you able to review this? 🙏 this will be very helpful for us to continue with the implementation of the related ADR that Eze mentioned previously. let us know if we can help you

sergio-mena:
Hi @educlerici-zondax, @melekes is off and will be back next Thu. I'm going through the PR now but I'd ask you folks to wait until he's back so that all his concerns are addressed. Thanks 🙏

docs/references/architecture/adr-117-multi-curve-support.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Decision
+
+We will:
```
sergio-mena, line 23: I know you folks define `CryptoProvider` in [ADR-001-crypto-provider](https://github.com/cosmos/crypto/blob/main/docs/architecture/adr-001-crypto-provider.md), but for readability purposes, would you mind adding the `CryptoProvider` definition also here? Thanks!

docs/references/architecture/adr-117-multi-curve-support.md
```diff
@@ -0,0 +1,245 @@
...
+│   ├── priv_validator.go
+```
+
+#### Other considerations
+
+##### Node startup
+
+Node's configuration file should include the name of the `CryptoProvider` to be loaded at startup. Also, the startup logic will need to be changed from creating a validator to loading a `CryptoProvider`
+
+Not: It's important to note that during the migration path, both configurations (priv validators and crypto providers) will coexist. This means that if a `CryptoProvider` ID is passed in the configuration, the loader will give priority to that config and load the corresponding provider. However, if the provider ID is nil, it will load the priv validators as it always has. This approach ensures a smooth transition and maintains backward compatibility while allowing for the gradual adoption of the new `CryptoProvider` system.
```
sergio-mena, line 225:
```suggestion
Note: It's important to note that during the migration path, both configurations (priv validators and crypto providers) will coexist. This means that if a `CryptoProvider` ID is passed in the configuration, the loader will give priority to that config and load the corresponding provider. However, if the provider ID is nil, it will load the priv validators as it always has. This approach ensures a smooth transition and maintains backward compatibility while allowing for the gradual adoption of the new `CryptoProvider` system.
```

sergio-mena:
I went through the text again, and there is a question I'd like ask regarding the whole design (sorry I didn't realise this the first time I reviewed this while still in Cosmos SDK repo).

In your design, you seek to implement _privvals_ using _crypto providers_. In your code snippets, I see how you'll be using crypto providers (looks like there will only be one _privval_ implementation based on crypto providers) to implement `PrivValidator` interface's methods.

Here's my question: are you not _effectively_ replacing CometBFT's `PrivKey` and `PubKey` interfaces (which will no longer be used in the crypto-provider-based implementation of `PrivValidator`) ? Not saying this is bad (likely, all the contrary), just checking my understanding. If I am correct, **I think it explaining this somewhere in the text would help people better understand the planned changes**.

One more comment on this: bear in mind that _privvals_' main _raison d'être_ is to prevent double-signing (_privvals_ keep track of the last consensus height, round, and signature that the local node issued). You can check e.g. [this part](https://github.com/cometbft/cometbft/blob/4d98dc1b/privval/file.go#L397-L412) of `FilePV`'s `SignProposal` method. So the new crypto-provider-based implementation of `PrivValidator` MUST contain this logic as well. Also, **worth mentioning in the text and/or adapting your code snippets** to make this explicit.

raynaudoe:
> Here's my question: are you not _effectively_ replacing CometBFT's `PrivKey` and `PubKey` interfaces (which will no longer be used in the crypto-provider-based implementation of `PrivValidator`) by crypto providers? Not saying this is bad (likely, all the contrary), just checking my understanding. If I am correct, **I think explaining this somewhere in the text would help people better understand the planned changes**.

The `PrivKey` and `PubKey` interfaces would indeed still be present but wrapped inside the `CryptoProvider`. However, these interfaces would need to change slightly to be more "modular". I've recently been working on this aspect [here](https://github.com/cosmos/crypto/blob/main/types/keys.go), which avoids having to copy and paste those interfaces just to match the desired "flavor" of the keys. While this may not be the final design, to ensure compatibility with as many codebases as possible, these interfaces need to be composable.


> One more comment on this: bear in mind that _privvals_' main _raison d'être_ is to prevent double-signing (_privvals_ keep track of the last consensus height, round, and signature that the local node issued). You can check e.g. [this part](https://github.com/cometbft/cometbft/blob/4d98dc1b/privval/file.go#L397-L412) of `FilePV`'s `SignProposal` method. So the new crypto-provider-based implementation of `PrivValidator` MUST contain this logic as well. Also, **worth mentioning in the text and/or adapting your code snippets** to make this explicit.

Sure, that was taken into consideration, the state will be kept as it is in the current code. 
The idea is not to rewrite from scratch the PVs but only change the minimal code to implement `CryptoProvider` instead of `PrivVal`, so all the logic of keeping the state will be intact.

sergio-mena (approved):
Approving this, after our in-person conversation (please add a couple of sentences in the text, explaining the two point in my last comment in the review)
(also please wait till Thu, so that @melekes can take a look).
Thanks 🙏

docs/references/architecture/adr-117-multi-curve-support.md
```diff
@@ -0,0 +1,245 @@
...
+
+```go
+// CryptoProviderPV is the implementation of PrivValidator using CryptoProvider's methods
+type CryptoProviderPV struct {
+    provider CryptoProvider
+    pubKey   PubKey
+}
+
+// NewCryptoProviderPV creates a new instance of CryptoProviderPV
+func NewCryptoProviderPV(provider CryptoProvider, pk PubKey) (*CryptoProviderPV, error) {
```
melekes, line 134: Where is `pk` coming from in this use case?

docs/references/architecture/adr-117-multi-curve-support.md
```diff
@@ -0,0 +1,245 @@
...
+
+#### Directories reorganization
+
+Implementations of crypto providers (previously `Privval` implementations) should be in their own directory:
+
+```plaintext
+cometbft/
+├── privval/
+│   ├── providers/
+│   │   ├── file_provider.go
```
melekes, line 211:
```suggestion
│   │   ├── file.go
```

docs/references/architecture/adr-117-multi-curve-support.md
```diff
@@ -0,0 +1,245 @@
...
+│   │   ├── signer_provider.go
```
melekes, line 212:
```suggestion
│   │   ├── signer.go
```

by Go convention

docs/references/architecture/adr-117-multi-curve-support.md
```diff
@@ -0,0 +1,245 @@
...
+
+
+#### Directories reorganization
+
+Implementations of crypto providers (previously `Privval` implementations) should be in their own directory:
+
+```plaintext
+cometbft/
+├── privval/
+│   ├── providers/
```
melekes, line 210:
```suggestion
│   ├── provider/
```

https://go.dev/doc/effective_go#package-names

docs/references/architecture/adr-117-multi-curve-support.md
```diff
@@ -0,0 +1,245 @@
...
+│   │   ├── mock_provider.go
+├── types/
+│   ├── priv_validator.go
+```
+
+#### Other considerations
+
+##### Node startup
+
+Node's configuration file should include the name of the `CryptoProvider` to be loaded at startup. Also, the startup logic will need to be changed from creating a validator to loading a `CryptoProvider`
```
melekes, line 223: Will we be loading entire crypto providers from the disk or just a name?

raynaudoe: The idea is to load the entire provider at startup, by loading the id we can then search for the proper file (the json that describes the provider). Once the json is loaded, it is then passed to the "Providers Factory" and its output is the provider that will be doing the signing/verification ops.

melekes: So, the ID is a file path. Got it.

docs/references/architecture/adr-117-multi-curve-support.md
raynaudoe: That's an error, thanks for pointing it out: keys are intended to be managed inside the crypto provider.
I see that in the code, most of the times, the pubkeys are retrieved to get the proper address. We can eventually add a `GetAddress() string` or `GetPubKey []byte` to the crypto provider interface to be able to still get that data.

docs/references/architecture/adr-117-multi-curve-support.md
```diff
@@ -0,0 +1,258 @@
...
+
+    // GetHasher returns an instance of Hasher.
+    GetHasher() Hasher
+
+    // Metadata returns metadata for the crypto provider.
+    Metadata() ProviderMetadata
+}
+```
+
+## Decision
```
cason, line 40: This is a Proposal, I would say.

cason (commented):
Our ADR typically include an "Alternative Approaches" session. By analyzing the alternatives and the proposal, we define our decision.

cason:
As an additional comment. Consensus and state packages rely on a `PrivVal` implementation to sign messages and verify signatures. How would them be changed?

raynaudoe:
> As an additional comment. Consensus and state packages rely on a `PrivVal` implementation to sign messages and verify signatures. How would them be changed?

Every piece of code that uses `PrivVal` should not be changed by this ADR since the PrivVal interface will still work in the same way, with the only difference being that under the hood, a crypto provider is performing all the cryptographic operations.

cason:
> > As an additional comment. Consensus and state packages rely on a `PrivVal` implementation to sign messages and verify signatures. How would them be changed?
> 
> Every piece of code that uses `PrivVal` should not be changed by this ADR since the PrivVal interface will still work in the same way, with the only difference being that under the hood, a crypto provider is performing all the cryptographic operations.

Is this included in the ADR content? This is super relevant, on my view.

raynaudoe:
> > > As an additional comment. Consensus and state packages rely on a `PrivVal` implementation to sign messages and verify signatures. How would them be changed?
> > 
> > 
> > Every piece of code that uses `PrivVal` should not be changed by this ADR since the PrivVal interface will still work in the same way, with the only difference being that under the hood, a crypto provider is performing all the cryptographic operations.
> 
> Is this included in the ADR content? This is super relevant, on my view.

Sure, just added a clarification [here](https://github.com/Zondax/cometbft/blob/feat/adr-multi-curve/docs/references/architecture/adr-117-multi-curve-support.md#privvalidator--users)

raynaudoe:
Hi @melekes @cason did you have a chance to take a look at the latest version of the document?

melekes (commented):
It's still unclear to me how `we will allow for "pluggable cryptography" that enables a more modular design and future-proofs crypto tools.` is achieved. We already have a remote signer, which allows devs to use any curve, so what will change with this ADR? 

We will be loading crypto providers from disk (or better from keyring), which is great; I get that. But how is multi-curve coming into play here? Even if we're loading crypto providers from disk (FilePV), we still need the curve itself available (like if you're loading ed25519 from disk, you need the ed25519 curve available in CometBFT), no?

cason (changes requested):
The title of the PR is not precise.

We are not adding multi-curve support, as we already have it. If I understood correctly, the proposal is to implement our `PrivVal` interface using the functionalities provided by the introduced `cosmos.CryptoProvider` interface.

So we are keeping our `PrivVal` logic, but allowing for multiple implementations of it by adopting the potential multiple implementation of the `CryptoProvider` interface.

After reading the PR, this intuition was not clear to me. I had to meet with @sergio-mena to get this understanding.

I am asking changes because this, which is the main point of this PR, should be made clear in the title and in the documentation.

docs/references/architecture/adr-117-multi-curve-support.md
```diff
@@ -0,0 +1,263 @@
...
+
+#### Loading and storing
+
+When we refer to storing a `CryptoProvider`, we're essentially preserving two key elements: the type of crypto provider (e.g., file-based, hardware security module, remote signer, etc) and its associated configuration parameters.
+
+These configuration parameters can vary widely depending on the provider type. For instance, a file-based provider might need to store file paths, while a hardware security module provider could require device identifiers or network addresses. Remote signers might need endpoint URLs and authentication details. By storing both the type and these specific configurations, we ensure that the `CryptoProvider` can be correctly instantiated.
+
+Alternatives:
+
+* *Low impact / minimal dependencies*: The corresponding `CryptoProvider` can be directly stored on disk in a dedicated directory in an encoding format of choice (text, JSON)
```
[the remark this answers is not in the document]
raynaudoe: It is an alternative, but we think the best approach is using `keyring`

cason (approved):
Thank you for the changes.

cason:
We are having some issues with `make proto-gen` that are blocking merging some PRs. Once we have it solved, IMO we can merge this ADR.

PR #3403, opened 2024-07-02
perf(p2p/conn): Remove unneeded global pool buffers in secret connection

### LINKED ISSUE #3197: Secret Connection: Remove sync.Pool, and put those buffers in the connection struct
### Summary

In secret connection we use sync.Pool's for some 1024 byte buffers. However these buffers are only ever needed locally within the same thread, so this is a wasted synchronization primitive that does cost us. All reads and writes in secret connection are under a mutex, so we should just make a new buffer in each secret conn struct. 

Impact is ~8% of the current secret connection read time (Get and Put in following pprof): 
image

### Problem Definition

Slight speedup to secret conn

ValarDragon:
Closes #3197 

Remove unneeded global pool buffers in secret connection, instead just use a byte slice per secret connection, and just once for all packet sends. This is a 15% improvement to Secret connection's CPU time in Write, and 8% to read

We have much bigger system bottlenecks to fix (in the same RecvRoutine and SendRoutine calls), but this felt very low-lift, hence making the PR.

---

#### PR checklist

- [X] Tests written/updated  - N/A


ValarDragon:
(The govulncheck failure is unrelated, just a dependency update that needs to happen)

cason (approved):
Great work, simple change, probably high improvement.

But from the code it seems that we did not even need a pool of buffers, synchronized or not, but just a single buffer? Amazing that who implemented that did not realize it.

cason:
I renamed the changelog entry and PR title to `p2p/conn` because this is the name of the package where the secret connection is implemented.

cason:
Ah, you forgot to run `go mod tidy` on the new code, since we have removed the libp2p dependency:

```
$ git diff go.mod
diff --git a/go.mod b/go.mod
index 13b2b9f26..a85559654 100644
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,6 @@ require (
        github.com/google/orderedcode v0.0.1
        github.com/gorilla/websocket v1.5.3
        github.com/lib/pq v1.10.9
-       github.com/libp2p/go-buffer-pool v0.1.0
```

ValarDragon:
> But from the code it seems that we did not even need a pool of buffers, synchronized or not, but just a single buffer? Amazing that who implemented that did not realize it.

Yup, haha.

> I renamed the changelog entry and PR title to p2p/conn because this is the name of the package where the secret connection is implemented.

Thank you! I forgot its based on package, hopefully will correct better for subsequent PRs!

> Ah, you forgot to run go mod tidy on the new code, since we have removed the libp2p dependency

Wow, nice consequence of this

adizere:
Should we backport to v1 also?

ValarDragon:
yes please!

melekes:
@mergify backport v1.x

melekes:
@mergify backport v0.37.x

PR #3405, opened 2024-07-03
refactor(mempool): add method `TryAddTx` to reactor

### LINKED ISSUE #3296: add new method `TryAddTx` to mempool reactor
This new method will serve as a single point of access to the mempool reactor for incoming transactions. 

Currently to add a transaction we call `CListMempool.CheckTx(...)` directly, either from the `broadcast_tx_*` RPC endpoints or from the `Receive(...)` method in the reactor. The goal of `TryAddTx` is to have a single method in the reactor where in the future we can put common logic for adding transactions, such as handling errors or sending new messages. In particular, this refactoring will be needed by the implementations of mempool lanes and the DOG protocol.

hvanz:
Closes #3296 

---

#### PR checklist

- [X] Tests written/updated


cason:
https://github.com/cometbft/cometbft/issues/3296#issuecomment-2205469441

mempool/reactor.go
```diff
@@ -168,6 +164,27 @@ func (memR *Reactor) Receive(e p2p.Envelope) {
...
+func (memR *Reactor) TryAddTx(tx types.Tx, sender p2p.Peer) (*abcicli.ReqRes, error) {
+	senderID := noSender
+	if sender != nil {
+		senderID = sender.ID()
+	}
+
+	reqRes, err := memR.mempool.CheckTx(tx, senderID)
+	switch {
+	case errors.Is(err, ErrTxInCache):
+		memR.Logger.Debug("Tx already exists in cache", "tx", log.NewLazySprintf("%v", tx.Hash()))
```
hvanz, line 178: There's a change in behaviour with this code. These messages were logged only in `Receive`, for transactions coming from peers. Now we log them also when transactions coming via RPC endpoints.

sergio-mena, line 178:
```suggestion
		memR.Logger.Debug("tx already exists in cache", "tx", log.NewLazySprintf("%v", tx.Hash()))
```

sergio-mena: You might want to add the sender to the log

mempool/reactor.go
```diff
@@ -168,6 +164,27 @@ func (memR *Reactor) Receive(e p2p.Envelope) {
...
+		return nil, err
+	case err != nil:
+		memR.Logger.Info("Could not check tx", "tx", log.NewLazySprintf("%v", tx.Hash()), "err", err)
```
sergio-mena, line 181:
```suggestion
		memR.Logger.Info("could not check tx", "tx", log.NewLazySprintf("%v", tx.Hash()), "err", err)
```

sergio-mena (approved):
🎸🎸

mempool/reactor.go
melekes: why? https://github.com/cometbft/cometbft/blob/a3ec97e3b15afafd57b44305d4cc40a801d4bc35/CONTRIBUTING.md#log-messages Unlike Go errors, log messages must have the first letter of the message capitalized.

mempool/reactor.go
melekes: why? https://github.com/cometbft/cometbft/blob/a3ec97e3b15afafd57b44305d4cc40a801d4bc35/CONTRIBUTING.md#log-messages Unlike Go errors, log messages must have the first letter of the message capitalized.

melekes (approved):
👍

mempool/reactor.go
hvanz: I'll leave it as it is.

mempool/reactor.go
hvanz: I'll leave it as it is.

sergio-mena: @melekes you're right. I thought our policy was to "lowercase" both errors and log messages, but I see I was wrong.

PR #3406, opened 2024-07-03
fix(blocksync)!: don't block in blocksync if our voting power is blocking the chain

sergio-mena:
Partially addresses #3415

The a node has no peers, blocksync gets stuck without switching to consesnus, because it needs info from other peers to have an idea of maximum height.

However, there is an edge case (mainly when testing) where a validator might have >2/3 of the voting power and other validators are not started. In this case, we know we are blocking the chain, so we don't need to stay in blockchain if the only condition is that we don't have peers.

Moreover, in order to block a chain, 1/3 of the voting power is enough, so the reasoning of this fix is the following:

* _I am a node and I am starting... shall I run blocksync?_
* _Well, looks like I have 1/3 of the voting power (or more) at my current height... so there's no way the chain could advance in my absence... so I don't need to blocksync"_

Explanation of commits:

* Commit 1: `e2e` testbed reproducing the issue
* Commit 2: commit with a trivial change to trigger `e2e` tests. Check the error: ❌ next to the commit hash (3fb1057)
* Commit 3: Tentative fix. Although there is a  ❌ next to the commit hash (16a46ea), if you click on it, you'll see that `e2e` are passing now.
* Commit 4: revert commit2
* Commit 5: Move the check for "local node is blocking the chain" outside the pool, as suggested by @cason 
* Commit 6: Fixed unit tests

All further commits: addressing other comments and tidying up the code

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/.changelog/unreleased/bug-fixes/3406-blocksync-dont-stall-if-blocking-chain.md
+++ b/.changelog/unreleased/bug-fixes/3406-blocksync-dont-stall-if-blocking-chain.md
@@ -1,0 +1,3 @@
+- `[blocksync]` Do not stay in blocksync if the local node's voting power
+  is big enough to block the chain while not online
+  ([\#3406](https://github.com/cometbft/cometbft/pull/3406))
--- a/.changelog/unreleased/bug-fixes/3406-blocksync-dont-stall-if-blocking-chain.md
+++ b/.changelog/unreleased/bug-fixes/3406-blocksync-dont-stall-if-blocking-chain.md
@@ -1,3 +1,3 @@
-- `[blocksync]` Do not stay in blocksync if the local node's voting power
-  is big enough to block the chain while not online
+- `[blocksync]` Do not stay in blocksync if the node's validator voting power
+  is high enough to block the chain while it is not online
   ([\#3406](https://github.com/cometbft/cometbft/pull/3406))
--- a/internal/blocksync/pool.go
+++ b/internal/blocksync/pool.go
@@ -7,13 +7,11 @@ import (
 	"sort"
 	"time"
 
-	"github.com/cometbft/cometbft/crypto"
 	flow "github.com/cometbft/cometbft/internal/flowrate"
 	"github.com/cometbft/cometbft/libs/log"
 	"github.com/cometbft/cometbft/libs/service"
 	cmtsync "github.com/cometbft/cometbft/libs/sync"
 	"github.com/cometbft/cometbft/p2p"
-	sm "github.com/cometbft/cometbft/state"
 	"github.com/cometbft/cometbft/types"
 )
 
@@ -177,16 +175,6 @@ func (pool *BlockPool) removeTimedoutPeers() {
 	}
 
 	pool.sortPeers()
-}
-
-func weBlockTheChain(state sm.State, myAddr crypto.Address) bool {
-	_, val := state.Validators.GetByAddress(myAddr)
-	if val == nil {
-		return false
-	}
-	total := state.Validators.TotalVotingPower()
-	power := val.VotingPower
-	return power > total*2/3
 }
 
 // IsCaughtUp returns true if this node is caught up, false - otherwise.
--- a/internal/blocksync/pool.go
+++ b/internal/blocksync/pool.go
@@ -191,13 +191,9 @@ func weBlockTheChain(state sm.State, myAddr crypto.Address) bool {
 
 // IsCaughtUp returns true if this node is caught up, false - otherwise.
 // TODO: relax conditions, prevent abuse.
-func (pool *BlockPool) IsCaughtUp(state sm.State, myAddr crypto.Address) (isCaughtUp bool, height, maxPeerHeight int64) {
+func (pool *BlockPool) IsCaughtUp() (isCaughtUp bool, height, maxPeerHeight int64) {
 	pool.mtx.Lock()
 	defer pool.mtx.Unlock()
-
-	if weBlockTheChain(state, myAddr) {
-		return true, pool.height, pool.maxPeerHeight
-	}
 
 	// Need at least 1 peer to be considered caught up.
 	if len(pool.peers) == 0 {
--- a/internal/blocksync/pool.go
+++ b/internal/blocksync/pool.go
@@ -7,11 +7,13 @@ import (
 	"sort"
 	"time"
 
+	"github.com/cometbft/cometbft/crypto"
 	flow "github.com/cometbft/cometbft/internal/flowrate"
 	"github.com/cometbft/cometbft/libs/log"
 	"github.com/cometbft/cometbft/libs/service"
 	cmtsync "github.com/cometbft/cometbft/libs/sync"
 	"github.com/cometbft/cometbft/p2p"
+	sm "github.com/cometbft/cometbft/state"
 	"github.com/cometbft/cometbft/types"
 )
 
@@ -177,11 +179,25 @@ func (pool *BlockPool) removeTimedoutPeers() {
 	pool.sortPeers()
 }
 
+func weBlockTheChain(state sm.State, myAddr crypto.Address) bool {
+	_, val := state.Validators.GetByAddress(myAddr)
+	if val == nil {
+		return false
+	}
+	total := state.Validators.TotalVotingPower()
+	power := val.VotingPower
+	return power > total*2/3
+}
+
 // IsCaughtUp returns true if this node is caught up, false - otherwise.
 // TODO: relax conditions, prevent abuse.
-func (pool *BlockPool) IsCaughtUp() (isCaughtUp bool, height, maxPeerHeight int64) {
+func (pool *BlockPool) IsCaughtUp(state sm.State, myAddr crypto.Address) (isCaughtUp bool, height, maxPeerHeight int64) {
 	pool.mtx.Lock()
 	defer pool.mtx.Unlock()
+
+	if weBlockTheChain(state, myAddr) {
+		return true, pool.height, pool.maxPeerHeight
+	}
 
 	// Need at least 1 peer to be considered caught up.
 	if len(pool.peers) == 0 {
--- a/internal/blocksync/reactor.go
+++ b/internal/blocksync/reactor.go
@@ -512,7 +512,7 @@ func (bcR *Reactor) localNodeBlocksTheChain(state sm.State) bool {
 		return false
 	}
 	total := state.Validators.TotalVotingPower()
-	return val.VotingPower > total*2/3
+	return val.VotingPower >= total/3
 }
 
 func (bcR *Reactor) isCaughtUp(state sm.State, blocksSynced uint64, stateSynced bool) bool {
--- a/internal/blocksync/reactor.go
+++ b/internal/blocksync/reactor.go
@@ -512,8 +512,7 @@ func (bcR *Reactor) localNodeBlocksTheChain(state sm.State) bool {
 		return false
 	}
 	total := state.Validators.TotalVotingPower()
-	power := val.VotingPower
-	return power > total*2/3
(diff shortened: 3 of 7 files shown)


internal/blocksync/pool.go
```diff
@@ -177,12 +179,26 @@ func (pool *BlockPool) removeTimedoutPeers() {
 	pool.sortPeers()
 }
 
+func weBlockTheChain(state sm.State, myAddr crypto.Address) bool {
```
cason, line 182: I think this could stay at the reactor side, not at the pool.

internal/blocksync/reactor.go
```diff
@@ -504,7 +507,7 @@ func (bcR *Reactor) isMissingExtension(state sm.State, blocksSynced uint64) bool
 }
 
 func (bcR *Reactor) isCaughtUp(state sm.State, blocksSynced uint64, stateSynced bool) bool {
-	if isCaughtUp, height, _ := bcR.pool.IsCaughtUp(); isCaughtUp {
+	if isCaughtUp, height, _ := bcR.pool.IsCaughtUp(state, bcR.myAddr); isCaughtUp {
```
cason, line 510: We can add the addition condition here, actually. We have the state, etc.

internal/blocksync/reactor.go
```diff
@@ -109,6 +111,7 @@ func NewReactor(state sm.State, blockExec *sm.BlockExecutor, store *store.BlockS
 		store:        store,
 		pool:         pool,
 		blockSync:    blockSync,
+		myAddr:       myAddr,
```
cason, line 114: I can't believe that we don't know our validator address in this protocol...

cason (commented):
Not sure about this workaround.

We should now whether we should run block sync outside the protocol. But, ok, it works. But by changing the block Reactor constructor, we breaking a lot of code.

internal/blocksync/pool.go
sergio-mena: I'm thinking of making it part of `ValidatorSet`, as we may need it in node as well (to get rid of `onlyValidatorIsUs` later)

internal/blocksync/reactor.go
sergio-mena: We never needed it so far

internal/blocksync/reactor.go
sergio-mena: I don't understand your comment

cason: We can check this condition here, at the reactor, before checking the pool.

internal/blocksync/reactor.go
cason: When I am not a validator, what do we get here?

node/setup.go
```diff
@@ -233,12 +233,12 @@ func logNodeStartupInfo(state sm.State, pubKey crypto.PubKey, logger, consensusL
 	}
 }
 
-func onlyValidatorIsUs(state sm.State, pubKey crypto.PubKey) bool {
+func onlyValidatorIsUs(state sm.State, myAddr crypto.Address) bool {
```
cason, line 236: We still don't cover the scenario with more than one validator, one of which with more than 2/3.

internal/blocksync/reactor.go
sergio-mena: All nodes have a `privValidator`: notice that this [line](https://github.com/cometbft/cometbft/blob/9742bf6e1/node/node.go#L349) is run by all full nodes. The only difference is that if you are not a validator, that address doesn't appear in the validator set

internal/blocksync/reactor.go
sergio-mena: Oh I see, let me explore that possibility

internal/blocksync/pool.go
cason: This is good.

node/setup.go
sergio-mena: I'm pretty convinced that `onlyValidatorIsUs` is no longer necessary after this PR. We'll probably get rid of it. Not doing it here out of caution.

node/setup.go
```diff
@@ -305,13 +305,14 @@ func createBlocksyncReactor(config *cfg.Config,
 	blockExec *sm.BlockExecutor,
 	blockStore *store.BlockStore,
 	blockSync bool,
+	myAddr crypto.Address,
```
cason, line 308:
```suggestion
	validatorAddr crypto.Address,
```

internal/blocksync/reactor.go
```diff
@@ -61,6 +62,7 @@ type Reactor struct {
 	store         sm.BlockStore
 	pool          *BlockPool
 	blockSync     bool
+	myAddr        crypto.Address
```
cason, line 65:
```suggestion
	validatorAddr        crypto.Address
```

sergio-mena: But we're not sure we're a validator...

cason: But it is our validatorAddress, right, not network address.

cason: It could be empty for a full node and we are still good

sergio-mena: I still think `validatorAddr` is misleading here

internal/blocksync/reactor.go
sergio-mena: Man... you deserve a bonus for this comment! Everything is so much easier with the check outside the pool! 🚀

internal/blocksync/reactor.go
sergio-mena: How about `localAddr`?

internal/blocksync/reactor.go
```diff
@@ -503,8 +506,18 @@ func (bcR *Reactor) isMissingExtension(state sm.State, blocksSynced uint64) bool
...
 }
 
+func (bcR *Reactor) localNodeBlocksTheChain(state sm.State) bool {
+	_, val := state.Validators.GetByAddress(bcR.myAddr)
+	if val == nil {
+		return false
+	}
+	total := state.Validators.TotalVotingPower()
+	power := val.VotingPower
+	return power > total*2/3
```
sergio-mena, line 516:
```suggestion
	return val.VotingPower > total*2/3
```

internal/blocksync/reactor_test.go
```diff
@@ -164,7 +165,8 @@ func newReactor(
 		blockStore.SaveBlockWithExtendedCommit(thisBlock, thisParts, seenExtCommit)
 	}
 
-	bcReactor := NewReactor(state.Copy(), blockExec, blockStore, fastSync, NopMetrics(), 0)
+	// As the tests only support one validator in the valSet, we pass a different address to disable the `weBlockTheChain` check
```
sergio-mena, line 168:
```suggestion
	// As the tests only support one validator in the valSet, we pass a different address to disable the `localNodeBlocksTheChain` check
```

internal/blocksync/reactor.go
```diff
@@ -503,8 +506,17 @@ func (bcR *Reactor) isMissingExtension(state sm.State, blocksSynced uint64) bool
 	return missingExtension
 }
 
+func (bcR *Reactor) localNodeBlocksTheChain(state sm.State) bool {
+	_, val := state.Validators.GetByAddress(bcR.myAddr)
+	if val == nil {
+		return false
+	}
+	total := state.Validators.TotalVotingPower()
+	return val.VotingPower > total*2/3
```
sergio-mena, line 515: Wait a minute... I think this should be 1/3 (!)
@cason  ?

internal/blocksync/reactor.go
cason: Where are looking at the validator set where we expect to find a ... validator address.

But this is naming of variables, please feel free to choose.

internal/blocksync/reactor_test.go
```diff
@@ -164,7 +165,8 @@ func newReactor(
 		blockStore.SaveBlockWithExtendedCommit(thisBlock, thisParts, seenExtCommit)
 	}
 
-	bcReactor := NewReactor(state.Copy(), blockExec, blockStore, fastSync, NopMetrics(), 0)
+	// As the tests only support one validator in the valSet, we pass a different address to bypass the `localNodeBlocksTheChain` check
```
cason, line 168:
```suggestion
	// As the tests only support one validator in the valSet, we pass a different address to bypass the `localNodeBlocksTheChain` check. Namely, the tested node is not an active validator.
```

cason: Good solution!

internal/blocksync/reactor.go
cason: This is a good questions, and depends on what behavior we want here:

1. The validator runs network by itself, in other words, no other validator is needed for progress: `val.VotingPower > total*2/3`
2. The network cannot run without this validator, in other words, this validator is required for progress: `val.VotingPower >= total/3`

cason: From the name of the method, I would say 2.

Because if the validator is needed for progress and it crashes, no new block was produced since its crash, therefore block sync is useless. Do you agree?

.changelog/unreleased/bug-fixes/3406-blocksync-dont-stall-if-blocking-chain.md
```diff
@@ -0,0 +1,3 @@
+- `[blocksync]` Do not stay in blocksync if the local node's voting power
```
cason, line 1:
```suggestion
- `[blocksync]` Do not stay in blocksync if the node's validator voting power
```

.changelog/unreleased/bug-fixes/3406-blocksync-dont-stall-if-blocking-chain.md
```diff
@@ -0,0 +1,3 @@
...
+  is big enough to block the chain while not online
```
cason, line 2:
```suggestion
  is high enough to block the chain while it is not online
```

cason (commented):
I would approve, but the >=1/3 vs >2/3 question remains open.

See associated comment (line 515).

internal/blocksync/reactor.go
sergio-mena: After some internal discussion @cason and I have settled on option 2. Commit coming up

cason (approved):
+1

internal/blocksync/reactor.go
```diff
@@ -503,8 +506,17 @@ func (bcR *Reactor) isMissingExtension(state sm.State, blocksSynced uint64) bool
 	return missingExtension
 }
 
+func (bcR *Reactor) localNodeBlocksTheChain(state sm.State) bool {
+	_, val := state.Validators.GetByAddress(bcR.localAddr)
+	if val == nil {
+		return false
+	}
+	total := state.Validators.TotalVotingPower()
+	return val.VotingPower >= total/3
```
ValarDragon, line 515: Great observation that this is 1/3, not 2/3! Didn't realize that

ValarDragon (approved):
Great job!

adizere:
Is this safe to backport to 0.38? Or better not to because it's too significant behavior change?

ValarDragon:
I think it's definitely safe to backport, it can't really affect mainnets as you need one Val w/ over 1/3 to do anything. (And it only helps users right now if it's on the 38 line)

cason:
> Is this safe to backport to 0.38/v1 (post-rc1 v1)? 

We need to find a solution for v0.37.x too...

sergio-mena:
> Is this safe to backport to 0.38/v1 (post-rc1 v1)?

To me, it's a bug, so unless there is big risk identified I'd backport it. Besides, this is clearly holding teams back, which are on `v0.38.x`/`v0.37.x`. Please reply if you don't agree.

PR #3407, opened 2024-07-03
perf(consensus): Use TrySend for hasVote/HasBlockPart messages

### LINKED ISSUE #3151: Change broadcastHasVote and broadcastHasBlockpart to use TrySend not Send

ValarDragon:
Closes #3151 

The broadcast routine in its entirely is the source of ~75% of all created channels in the system, all coming from the `NewTimer` call in Send. (and 75% of all NewTimer)  Furthermore, the NewTimer call from Broadcast is 2% of all CPU time across the system. However I think this is actually more impact, due to the following. 

We spend 11% of all CPU time in the scheduler, and 4% of all CPU time is `CheckTimers` in the scheduler logic. I'd expect reducing most of the Timer calls reduces this. We also reduce total `mallocgc` time here by 5% by removing NewTimer, which should hopefully reduce GC load as well.

---

#### PR checklist

- [x] Tests written/updated


ValarDragon:
Going to test this in osmosis prod to see the scheduler improvement, mostly because I'm curious if it reflects my understanding. 

The test I added tests TryBroadcast to same extent Broadcast is tested within switch. (I didn't see any other test)

ValarDragon:
I fixed the old Go Benchmark (It hasn't been working for months).


```
BenchmarkSwitchBroadcast-12    	  347770	      4304 ns/op	    1238 B/op	      16 allocs/op
BenchmarkSwitchTryBroadcast-12     2268736	       482.7 ns/op	     381 B/op	      11 allocs/op
```

So a significant CPU cost reduction!

ValarDragon:
This appears to have been an empiric 20% reduction in scheduler time! (At consensus state and mempool times being roughly equal)

And a 5-10% GC time reduction!

cason:
I will take a detailed look maybe tomorrow, but conceptually I agree, in particular because missing those two messages is not a problem: they are optimizations, preventing duplication, they are not core messages of the protocol.

cason (approved):
Great!

Please fix the changelog before committing.

hvanz (approved):
🚀

.changelog/unreleased/improvements/3407-make-hasvote-broadcast-use-trysend.md
```diff
@@ -0,0 +1,4 @@
+- `[consensus]` Make broadcasting HasVote and HasBlockpart control messages
+    use TrySend instead of Send. This saves notable amounts of performance,
+    while at the same time those messages are for preventing redundancy, not critical, and may be dropped without risks for the protocol.
+  ([\#3342](https://github.com/cometbft/cometbft/issues/3342))
```
cason, line 4:
```suggestion
  ([\#3151](https://github.com/cometbft/cometbft/issues/3151))
```

p2p/switch.go
```diff
@@ -273,14 +273,26 @@ func (sw *Switch) Broadcast(e Envelope) {
...
 		}(p)
 	})
 }
 
+// TryBroadcast runs a go routine for each attempted send.
+// If the send queue of the destination channel and peer are full, the message will not be sent. To make sure that messages are indeed sent to all destination, use `Broadcast`.
+//
+// NOTE: TryBroadcast uses goroutines, so order of broadcast may not be preserved.
+func (sw *Switch) TryBroadcast(e Envelope) {
+	sw.Logger.Debug("TryBroadcast", "channel", e.ChannelID)
```
melekes, line 287:
```suggestion
```

.changelog/unreleased/improvements/3407-make-hasvote-broadcast-use-trysend.md
```diff
@@ -0,0 +1,4 @@
+- `[consensus]` Make broadcasting HasVote and HasBlockpart control messages
```
melekes, line 1:
```suggestion
- `[consensus]` Make broadcasting `HasVote` and `HasBlockpart` control messages
```

.changelog/unreleased/improvements/3407-make-hasvote-broadcast-use-trysend.md
```diff
@@ -0,0 +1,4 @@
...
+    use TrySend instead of Send. This saves notable amounts of performance,
```
melekes, line 2:
```suggestion
    use `TrySend` instead of `Send`. This saves notable amounts of performance,
```

melekes (approved):
Thanks @ValarDragon ❤️

p2p/switch.go
melekes: I don't think we need this one. I'd also remove `Debug` from `Broadcast`

melekes:
```
--- FAIL: TestStateOversizedBlock (0.38s)
    --- FAIL: TestStateOversizedBlock/max_size,_correct_block (0.38s)
        state_test.go:276: Detected real max data size for block; size 65255 softMaxDataBytes 64805
        state_test.go:307: Block Sizes; Limit 65536 Current 65535
        state_test.go:308: Proposal Parts; Maximum 1 Current 1
panic: Timeout expired while waiting for NewVote event [recovered]
	panic: Timeout expired while waiting for NewVote event

goroutine 37363 [running]:
testing.tRunner.func1.2({0x1359020, 0x175c150})
	/opt/hostedtoolcache/go/1.22.5/x64/src/testing/testing.go:1631 +0x3f7
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.22.5/x64/src/testing/testing.go:1634 +0x6b6
panic({0x1359020?, 0x175c150?})
	/opt/hostedtoolcache/go/1.22.5/x64/src/runtime/panic.go:770 +0x132
github.com/cometbft/cometbft/internal/consensus.ensureVote(0xc008015b00, 0x1, 0x1, 0x2)
	/home/runner/work/cometbft/cometbft/internal/consensus/common_test.go:740 +0x47e
github.com/cometbft/cometbft/internal/consensus.ensurePrecommit(...)
	/home/runner/work/cometbft/cometbft/internal/consensus/common_test.go:728
github.com/cometbft/cometbft/internal/consensus.TestStateOversizedBlock.func1(0xc004f48d00)
	/home/runner/work/cometbft/cometbft/internal/consensus/state_test.go:331 +0xcec
testing.tRunner(0xc004f48d00, 0xc0031f2300)
	/opt/hostedtoolcache/go/1.22.5/x64/src/testing/testing.go:1689 +0x21f
created by testing.(*T).Run in goroutine 37362
	/opt/hostedtoolcache/go/1.22.5/x64/src/testing/testing.go:1742 +0x826
FAIL	github.com/cometbft/cometbft/internal/consensus	20.370s
FAIL
```

ValarDragon:
I'm at conferences this week, I can work on debugging on weekend! Sorry for delay on replies

cason:
This error is odd, are you sure it was introduced by this PR?

cason:
I mean,  `TestStateOversizedBlock` don't use use a reactor.

melekes:
> This error is odd, are you sure it was introduced by this PR?

not sure

PR #3412, opened 2024-07-03
perf(p2p): Delete expensive debug log already slated for deletion

ValarDragon:
Delete an expensive debug log that has already been slated for deletion. On the osmosis fork (with many other debug logs removed), this debug log ended up taking 1.5% of all CPU time when we had many large blocks. (All from outbound gossip curiously enough) This caused 12GB of total memory allocation over an hour in that benchmark (out of 180GB total. With other optimizations in PR's / merged, would have been 12GB / 120GB total)

I am just making a PR to delete it, since it already had a code comment indicating we should delete it later. (Rather than waiting for #2989 

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/p2p/peer.go
+++ b/p2p/peer.go
@@ -311,15 +311,6 @@ func (p *peer) hasChannel(chID byte) bool {
 			return true
 		}
 	}
-	// NOTE: probably will want to remove this
-	// but could be helpful while the feature is new
-	p.Logger.Debug(
-		"Unknown channel for peer",
-		"channel",
-		chID,
-		"channels",
-		p.channels,
-	)
 	return false
 }
 


hvanz (approved):
👍

cason:
How often this code is hit? Because in most-cases we assume that the channel is present in both sender or receivers.

ValarDragon:
This is used in every single packet send!

cason:
Actually, this code is only run if we don't return true, which means that the channel does not exist.

ValarDragon:
Ah your right. Here is a profile where we are seeing it taking a large amount of total osmosis load:
image

This is the same profile where we saw gossipDataRoutine taking shockingly long amounts of time (and I suspected busy waiting) #3414 , maybe this helps us tell us the problem?

melekes (approved):
Thanks @ValarDragon ❤️

melekes:
> Actually, this code is only run if we don't return true, which means that the channel does not exist.

I don't understand it either. If all nodes have the same set of channels, then this debug line shouldn't ever be reached.

hvanz:
I would also change `channels []byte` to `channels map[byte]struct{}` so we don't have to traverse the array every time we send.

cason:
> I would also change `channels []byte` to `channels map[byte]struct{}` so we don't have to traverse the array every time we send.

I am pretty sure that traversing a byte array is faster than using a map. Almost 100% sure. We could use the `bytes` package though.

melekes:
I'm going to merge this, but we still need to understand where inefficiency is coming from https://github.com/cometbft/cometbft/pull/3412#issuecomment-2209262925

melekes:
@mergify backport v1.x

PR #3419, opened 2024-07-03
perf(p2p/conn): Use a read buffer on the secret connection

### LINKED ISSUE #3198: Secret Connection: Buffer the underlying connection

ValarDragon:
Closes #3198 

Similar to #3346 , buffers the secret connection reads. This is a notable savings to CPU time. (25% of recvRoutine time, on Osmosis' version)

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/p2p/conn/evil_secret_connection_test.go
+++ b/p2p/conn/evil_secret_connection_test.go
@@ -259,6 +259,8 @@ func TestMakeSecretConnection(t *testing.T) {
 		{"share bad ethimeral key", newEvilConn(true, true, false, false), func(err error) bool { return assert.Contains(t, err.Error(), "wrong wireType") }},
 		{"refuse to share auth signature", newEvilConn(true, false, false, false), func(err error) bool { return errors.Is(err, io.EOF) }},
 		{"share bad auth signature", newEvilConn(true, false, true, true), func(err error) bool { return errors.As(err, &ErrDecryptFrame{}) }},
+		// fails with the introduction of changes PR #3419
+		// {"all good", newEvilConn(true, false, true, false), func(err error) bool { return err == nil }},
 	}
 
 	for _, tc := range testCases {
--- a/p2p/conn/evil_secret_connection_test.go
+++ b/p2p/conn/evil_secret_connection_test.go
@@ -225,6 +225,7 @@ func (c *evilConn) signChallenge() []byte {
 	c.secretConn = &SecretConnection{
 		conn:            b,
 		connWriter:      bufio.NewWriterSize(b, defaultWriteBufferSize),
+		connReader:      b,
 		recvBuffer:      nil,
 		recvNonce:       new([aeadNonceSize]byte),
 		sendNonce:       new([aeadNonceSize]byte),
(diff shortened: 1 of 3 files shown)


p2p/conn/evil_secret_connection_test.go
```diff
@@ -258,7 +259,6 @@ func TestMakeSecretConnection(t *testing.T) {
 		{"share bad ethimeral key", newEvilConn(true, true, false, false), func(err error) bool { return assert.Contains(t, err.Error(), "wrong wireType") }},
 		{"refuse to share auth signature", newEvilConn(true, false, false, false), func(err error) bool { return errors.Is(err, io.EOF) }},
 		{"share bad auth signature", newEvilConn(true, false, true, true), func(err error) bool { return errors.As(err, &ErrDecryptFrame{}) }},
-		{"all good", newEvilConn(true, false, true, false), func(err error) bool { return err == nil }},
```
ValarDragon, line 261: I have spent over 12 hours attempting to fix this test, and it is incredibly complicated to do so. (The hardest test change out of everything I've done in comment) I do not think its worth investing in fixing the test. 

And this is because this is mocking an invalid secret connection, and then a "final check" that we mocked the whole evil secret connection correctly. Which involves simulating all reads/writes to something internal.

After talking with @zmanian we both think this test is unneeded. This test is basically testing an evil secret connection peer, who sent bad signatures /deviated from protocol could not authenticate. We check the exact error at every step, so we are still matching every error case. So it seems fine to delete the test that the evil secret connection mocking, would work for the full connection. 

We test that good handshakes complete in many  ways (All E2E tests, other secret connection tests, many switch tests, etc) I have also tested this on mainnet on multiple occasions.

ValarDragon: cc @melekes if this sounds good to you. Your the one who wrote this set of tests!

ValarDragon:
This is an improvement, but not nearly as good as the write buffer.

Encryption open time vs IO.Read time
OLD: 5.6 to 14.5
NEW: 5 to 12.6

So an ~8% speedup?

p2p/conn/evil_secret_connection_test.go
cason: The produced error, which should be nil is:

```
SecretConnection: failed to decrypt the frame: chacha20poly1305: message authentication failed
```

cason: It is produced here: 

https://github.com/cometbft/cometbft/blob/033f7bbf4f5f8a97a78d82275d84f8deb9f8eecc/p2p/conn/secret_connection.go#L437

cason: Anyway, do you mind commenting this line out instead of removing it? And put a comment "fails with the introduction of changes PR #3419"?

cason (approved):
Ok with the changes, buffering is good in general.

melekes (approved):
Thanks @ValarDragon ❤️

p2p/conn/evil_secret_connection_test.go
melekes: It's okay to comment out the test.

melekes:
@mergify backport v1.x

PR #3423, opened 2024-07-04
perf(p2p/conn): Remove p2p allocations for wrapping outbound packets

### LINKED ISSUE #2951: Speedup CPU time of channel.WritePacketMsgTo

ValarDragon:
Closes #2951 

Remove p2p allocations for wrapping outbound packets. Every packet send would require 3 new heap objects due to how we used pointers / did this wrapping. This costs us: 13% of all system heap allocations, and 20% of the CPU time of writing packets (under "normal" osmosis load. I don't have big blocks benchmarks after the secret connection improvement got in).

This should be a 25% speedup to `channel.writePacketMsgTo`, making writing getting closer to the bound of just proto-marshalling a packet.

This solution is simple (just make a buffer in the struct) since these packet sends are all single threaded.

---

#### PR checklist

- [X] Tests written/updated - N/A, existing tests cover safety here


ValarDragon:
This did succesfully lower the CPU time as expected, and also appears to have lowered GC by ~10%. (Very interestingly, after this change the largest CPU time segment on Osmosis is no longer GC, and is now instead the scheduler)

p2p/conn/connection.go
```diff
@@ -852,29 +859,28 @@ func (ch *Channel) isSendPending() bool {
...
 		atomic.AddInt32(&ch.sendQueueSize, -1) // decrement sendQueueSize
 	} else {
-		packet.Data = ch.sending[:maxSize]
-		packet.EOF = false
+		ch.nextPacketMsg.Data = ch.sending[:maxSize]
+		ch.nextPacketMsg.EOF = false
 		ch.sending = ch.sending[maxSize:]
 	}
-	return packet
+	wrapPacketMsgWithBuffers(ch.nextPacketMsg, ch.nextP2pWrapperPacketMsg, ch.nextPacket)
```
cason, line 876: This methods is only called here. Why not move its implementation here?

p2p/conn/connection.go
```diff
@@ -852,29 +859,28 @@ func (ch *Channel) isSendPending() bool {
 	return true
 }
 
-// Creates a new PacketMsg to send.
+// Updates the nextPacket proto message for us to send.
 // Not goroutine-safe.
-func (ch *Channel) nextPacketMsg() tmp2p.PacketMsg {
-	packet := tmp2p.PacketMsg{ChannelID: int32(ch.desc.ID)}
+func (ch *Channel) updateNextPacket() {
```
cason, line 864: Change the name of the method also in line 843/850.

p2p/conn/connection.go
```diff
@@ -920,32 +926,33 @@ func (ch *Channel) updateStats() {
...
+	msg := &tmp2p.Packet{}
+	mustWrapPacketInto(pb, msg)
+	return msg
+}
 
+func mustWrapPacketInto(pb proto.Message, dst *tmp2p.Packet) {
 	switch pb := pb.(type) {
-	case *tmp2p.Packet: // already a packet
-		msg = *pb
+	case *tmp2p.Packet: // already a packet, make a copy
```
cason, line 936: Is this case possible with the new changes?

p2p/conn/connection.go
```diff
@@ -779,6 +779,10 @@ type Channel struct {
 	sending       []byte
 	recentlySent  int64 // exponential moving average
 
+	nextPacketMsg           *tmp2p.PacketMsg
```
cason, line 782: Can you add a comment above mentioning that we reuse the same packets for sending purposes?

cason: Moreover, why we need pointers here? We allocate these fields only once, right?

cason (commented):
Great work.

Left, again, some implementation-related questions.

p2p/conn/connection.go
ValarDragon: We need the pointer since we need to store the on heap copy. (If we stored the stack object, then if we're not careful we may take a copy to stack and then push that to heap on every pointer call)

We could just be careful but seems easier to use a pointer and not have to think about it

melekes (approved):
Thanks @ValarDragon ❤️

cason:
Maybe a stupid question, but why do we use protobufs for low-level communication (packets, ping, pong)? Of course we have an overhead here, but what is the advantage of that? Those messages are super simple, size-prefixed messages, ordinary network exchanging []byte s.

melekes:
> Maybe a stupid question, but why do we use protobufs for low-level communication (packets, ping, pong)? Of course we have an overhead here, but what is the advantage of that?

There isn't any advantage imho.

PR #3424, opened 2024-07-04
perf(autofile): Regex should be only compiled once + RW Lock

itsdevbear:
Re-compiling the regex was allocating
 188mb per hour.

image


ValarDragon:
Id remove the RWmutex here personally. This is for the most part only called from single threaded contexts, so your paying extra overhead to have an RWMutex over a standard mutex. (I'd be shocked if a blocking profile showed anytime under this mutex)

Only compiling once is great!

ValarDragon (approved):

internal/autofile/group.go
```diff
@@ -422,7 +411,7 @@ func filePathForIndex(headPath string, index int, maxIndex int) string {
 // GroupReader provides an interface for reading from a Group.
 type GroupReader struct {
 	*Group
-	mtx       sync.Mutex
+	mtx       sync.RWMutex
```
melekes, line 414:
```suggestion
```

melekes (approved):
Thanks @itsdevbear ❤️

hvanz (approved):

PR #3425, opened 2024-07-04
perf(p2p): Avoid allocating a new timer on every packet.

itsdevbear:
image


p2p/conn/connection.go
```diff
@@ -802,15 +804,22 @@ func (ch *Channel) SetLogger(l log.Logger) {
...
+	ch.timer.Reset(defaultSendTimeout)
+
 	select {
 	case ch.sendQueue <- bytes:
 		atomic.AddInt32(&ch.sendQueueSize, 1)
 		return true
-	case <-time.After(defaultSendTimeout):
+	case <-ch.timer.C:
+		// Stop the timer to release its resources
+		if !ch.timer.Stop() {
```
melekes, line 820: If we stop the timer, it will never fire again. This means that upon the next send, we will potentially block forever because there will be no timeout.

melekes (changes requested):
Thanks @itsdevbear ❤️  But I think there's an issue with the changes 👇

p2p/conn/connection.go
```diff
@@ -802,15 +804,22 @@ func (ch *Channel) SetLogger(l log.Logger) {
 	ch.Logger = l
 }
 
-// Queues message to send to this channel.
+// sendBytes queues a message to send to this channel.
 // Goroutine-safe
```
cason, line 808: Not goroutine safe anylonger.

p2p/conn/connection.go
itsdevbear: Reset(..) will reset it

itsdevbear:
yup I think this change is fundamentally flawed. Going to look into how we can not block here and then have a way to clear the sendBytes queue in another way

PR #3426, opened 2024-07-05
feat(reactor): remove usage of time.sleep cause it was big ghetto on god fr fr

itsdevbear:
The PR removes the usage of time.Sleep within the consensus reactor.


cason:
Why?

itsdevbear:
time.sleep is a bit messy and there is a gothonic solution to this design using select statements

internal/consensus/reactor.go
```diff
@@ -728,15 +726,10 @@ OUTER_LOOP:
 							BlockID: maj23.ToProto(),
 						},
 					})
-					time.Sleep(conR.conS.config.PeerQueryMaj23SleepDuration)
```
melekes, line 731: to preserve the old behavior we must `Reset` ticker here

internal/consensus/reactor.go
```diff
@@ -635,88 +637,84 @@ OUTER_LOOP:
...
-
-		time.Sleep(conR.conS.config.PeerGossipSleepDuration)
 	}
 }
 
 // NOTE: `queryMaj23Routine` has a simple crude design since it only comes
 // into play for liveness when there's a signature DDoS attack happening.
 func (conR *Reactor) queryMaj23Routine(peer p2p.Peer, ps *PeerState) {
-OUTER_LOOP:
+	ticker := time.NewTicker(conR.conS.config.PeerQueryMaj23SleepDuration)
```
melekes, line 703: I actually like this part ❤️

internal/consensus/reactor.go
```diff
@@ -635,88 +637,84 @@ OUTER_LOOP:
...
+	defer ticker.Stop()
+
 	for {
```
melekes, line 706: Should we tick immediately to preserve the old behavior?

melekes (commented):
I appreciate the effort here, but tickers make this code less predictable. I don't think there are any immediate downsides to ticking more often, but I'd rather keep the old code if there are no significant performance gains.

PR #3430, opened 2024-07-05
perf(mempool): get peer state once on broadcast routine

hvanz:
In each broadcast routine, before sending a transaction, we read the peer's state so we can compare the height of the transaction against the peer's height. So, for each transaction we want to send, we call `peer.Get(types.PeerStateKey).(PeerState)` that reads the (pointer to the) peer's state. This PR makes this call once, before starting the loop through the list of transactions. This is possible because `PeerState` is set once (on the consensus reactor) and never unset.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/mempool/reactor.go
+++ b/mempool/reactor.go
@@ -203,6 +203,22 @@ func (memR *Reactor) broadcastTxRoutine(peer p2p.Peer) {
 		}
 	}
 
+	var peerState PeerState
+	// Wait until the peer's state is ready. We initialize it in the consensus reactor, but when we
+	// add the peer in Switch, the order in which we call reactors#AddPeer is different every time
+	// due to us using a map. Sometimes other reactors will be initialized before the consensus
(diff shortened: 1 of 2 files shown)


mempool/reactor.go
```diff
@@ -203,6 +203,22 @@ func (memR *Reactor) broadcastTxRoutine(peer p2p.Peer) {
...
+	// reactor. We should wait a few milliseconds and retry. We assume the pointer to the state is
+	// set once and never unset.
+	for {
+		if ps, ok := peer.Get(types.PeerStateKey).(PeerState); ok {
+			peerState = ps
+			break
+		}
+		// Peer does not have a state yet.
+		time.Sleep(PeerCatchupSleepIntervalMS * time.Millisecond)
+		continue
```
melekes, line 219:
```suggestion
```

melekes (approved):
👍

PR #3459, opened 2024-07-08
refactor(mempool): Add `Iterator` to replace `TxsFront` and `TxsWaitChan` methods

### LINKED ISSUE #3303: mempool: fix `TxsFront` and `TxsWaitChan` in `CListMempool` that leak implementation details
`CListMempool.TxsFront` and `CListMempool.TxsWaitChan` should not leak implementation details. We need to separate reactor from pool implementation.

hvanz:
Closes #3303 

This PR adds:
- the `Entry` interface, implemented by the existing `mempoolTx` data type, 
- the `Iterator` interface, with one method `WaitNextCh() <-chan Entry`.

On the routine that sends transactions to a peer, we can now use an `Iterator` to select the next mempool entry to send, instead of the methods `TxsFront` and `TxsWaitChan` in `CListMempool`. These two methods leak implementation details about the CList and are now marked as deprecated. Now all the logic needed to wait for and pick the next available entry is implemented in `CListMempool.WaitNextCh`.

---

#### PR checklist

- [X] Tests written/updated

### DIFF
--- a/.changelog/unreleased/deprecations/3303-mempool-deprecate-txsfront-txswaitchan.md
+++ b/.changelog/unreleased/deprecations/3303-mempool-deprecate-txsfront-txswaitchan.md
@@ -1,0 +1,2 @@
+- `[mempool]` Mark methods `TxsFront` and `TxsWaitChan` in `CListMempool` as deprecated. They should
+  be replaced by the new `CListIterator` ([\#3303](https://github.com/cometbft/cometbft/pull/3303)).
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -785,7 +785,9 @@ func (rc *recheck) consideredFull() bool {
 	return rc.recheckFull.Load()
 }
 
-// CListIterator implements an Iterator that traverses the CList sequentially.
+// CListIterator implements an Iterator that traverses the CList sequentially. When the current
+// entry is removed from the mempool, the iterator starts from the beginning of the CList. When it
+// reaches the end, it waits until a new entry is appended.
 type CListIterator struct {
 	txs    *clist.CList    // to wait on and retrieve the first entry
 	cursor *clist.CElement // pointer to the current entry in the list
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -220,7 +220,7 @@ func (mem *CListMempool) Flush() {
 //
 // Safe for concurrent use by multiple goroutines.
 //
-// Deprecated: Use Iterator instead.
+// Deprecated: Use CListIterator instead.
 func (mem *CListMempool) TxsFront() *clist.CElement {
 	return mem.txs.Front()
 }
@@ -231,7 +231,7 @@ func (mem *CListMempool) TxsFront() *clist.CElement {
 //
 // Safe for concurrent use by multiple goroutines.
 //
-// Deprecated: Use Iterator instead.
+// Deprecated: Use CListIterator instead.
 func (mem *CListMempool) TxsWaitChan() <-chan struct{} {
 	return mem.txs.WaitChan()
 }
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -219,6 +219,8 @@ func (mem *CListMempool) Flush() {
 // FIXME: leaking implementation details!
 //
 // Safe for concurrent use by multiple goroutines.
+//
+// Deprecated: Use Iterator instead.
 func (mem *CListMempool) TxsFront() *clist.CElement {
 	return mem.txs.Front()
 }
@@ -228,6 +230,8 @@ func (mem *CListMempool) TxsFront() *clist.CElement {
 // element)
 //
 // Safe for concurrent use by multiple goroutines.
+//
+// Deprecated: Use Iterator instead.
 func (mem *CListMempool) TxsWaitChan() <-chan struct{} {
 	return mem.txs.WaitChan()
 }
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -799,8 +799,8 @@ func (mem *CListMempool) NewIterator() Iterator {
 	}
 }
 
-// WaitNextCh returns a channel to wait for the next available entry. The
-// channel will be closed once the entry is added to the channel.
+// WaitNextCh returns a channel to wait for the next available entry. The channel will be closed
+// once the entry is added to the channel or when reaching the end of the list.
 func (iter *CListIterator) WaitNextCh() <-chan Entry {
 	ch := make(chan Entry)
 	// Spawn goroutine that waits for the next entry, saves it locally, and puts it in the channel.
@@ -815,6 +815,7 @@ func (iter *CListIterator) WaitNextCh() <-chan Entry {
 			<-iter.cursor.NextWaitChan()
 			iter.cursor = iter.cursor.Next()
 		}
+		// Check if we reached the end of the list (CElement.Next returned nil).
 		if iter.cursor != nil {
 			ch <- iter.cursor.Value.(Entry)
 		}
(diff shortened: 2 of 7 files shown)


mempool/clist_mempool.go
```diff
@@ -780,3 +784,41 @@ func (rc *recheck) setRecheckFull() bool {
...
+			iter.cursor = iter.txs.Front()
+		} else {
+			// Wait for the next entry after the current one.
+			<-iter.cursor.NextWaitChan()
+			iter.cursor = iter.cursor.Next()
+		}
+		if iter.cursor != nil {
+			ch <- iter.cursor.Value.(Entry)
+		}
+		close(ch)
```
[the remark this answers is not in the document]
melekes: Or change functions docs

.changelog/unreleased/deprecations/3303-mempool-deprecate-txsfront-txswaitchan.md
```diff
@@ -0,0 +1,2 @@
+- `[mempool]` Mark methods `TxsFront` and `TxsWaitChan` in `CListMempool` as deprecated. They should
```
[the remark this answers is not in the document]
hvanz: I've seen it in the [changelog](https://github.com/cometbft/cometbft/blob/a3ec97e3b15afafd57b44305d4cc40a801d4bc35/CHANGELOG.md?plain=1#L748) before. And unclog will just pick up any folder name to create a section in md file.

mempool/clist_mempool.go
```diff
@@ -780,3 +784,41 @@ func (rc *recheck) setRecheckFull() bool {
...
+			// We are at the beginning of the iteration or the saved entry got removed: wait until
+			// the list becomes not empty and select the first entry.
+			<-iter.txs.WaitChan()
+			iter.cursor = iter.txs.Front()
+		} else {
+			// Wait for the next entry after the current one.
+			<-iter.cursor.NextWaitChan()
+			iter.cursor = iter.cursor.Next()
+		}
+		if iter.cursor != nil {
```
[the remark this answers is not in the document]
hvanz: `iter.cursor.Next()` will be nil at the end of the list. I added a comment to explain.

mempool/reactor.go
```diff
@@ -218,26 +215,24 @@ func (memR *Reactor) broadcastTxRoutine(peer p2p.Peer) {
...
-					continue
-				}
-			case <-peer.Quit():
-				return
-			case <-memR.Quit():
-				return
+		select {
+		case entry = <-iter.WaitNextCh():
+			// If the entry we were looking at got garbage collected (removed), try again.
+			if entry == nil {
```
melekes, line 229: If we reach the end, do we also retry here?

mempool/clist_mempool.go
```diff
@@ -780,3 +784,42 @@ func (rc *recheck) setRecheckFull() bool {
...
+}
+
+func (mem *CListMempool) NewIterator() Iterator {
+	return &CListIterator{
+		txs: mem.txs,
+	}
+}
+
+// WaitNextCh returns a channel to wait for the next available entry. The channel will be closed
+// once the entry is added to the channel or when reaching the end of the list.
```
sergio-mena, line 803: Is it safe to call this function from several goroutines? (I don't think it is). Can you clarify this in this comment?

mempool/clist_mempool.go
```diff
@@ -780,3 +784,42 @@ func (rc *recheck) setRecheckFull() bool {
...
+		} else {
+			// Wait for the next entry after the current one.
+			<-iter.cursor.NextWaitChan()
+			iter.cursor = iter.cursor.Next()
+		}
+		// Check if we reached the end of the list (CElement.Next returned nil).
+		if iter.cursor != nil {
+			ch <- iter.cursor.Value.(Entry)
+		}
+		close(ch)
```
melekes, line 822: Remember that closing a channel also sends a `nil` value 

```
case entry = <-iter.WaitNextCh():
```

will fire with `nil` and we will execute `continue`

melekes (approved):
👍 Not sure about channel closing and retrying semantics, otherwise LGTM

mempool/clist_mempool.go
```diff
@@ -780,3 +784,42 @@ func (rc *recheck) setRecheckFull() bool {
...
+	go func() {
+		if iter.cursor == nil {
+			// We are at the beginning of the iteration or the saved entry got removed: wait until
+			// the list becomes not empty and select the first entry.
+			<-iter.txs.WaitChan()
+			iter.cursor = iter.txs.Front()
+		} else {
+			// Wait for the next entry after the current one.
+			<-iter.cursor.NextWaitChan()
+			iter.cursor = iter.cursor.Next()
```
sergio-mena, line 816: What if the TX that `iter.cursor` points to got removed while we were blocked in line 815?

mempool/mempool.go
```diff
@@ -148,3 +148,25 @@ func PostCheckMaxGas(maxGas int64) PostCheckFunc {
...
+	Tx() types.Tx
+
+	// Height returns the height of the latest block at the moment the entry was created.
+	Height() int64
+
+	// GasWanted returns the amount of gas required by the transaction.
+	GasWanted() int64
+
+	// IsSender returns whether we received the transaction from the given peer ID.
+	IsSender(peerID p2p.ID) bool
```
sergio-mena, line 164: Would it be more future-proof to define a `Sender` function that simply returns the sender?

mempool/clist_mempool.go
sergio-mena: Hmm, I think the answer to this is [here](https://github.com/cometbft/cometbft/pull/3459/files#diff-3e6afafe7994a696d508a5a5b093343bdb5d1ff2ff00bab0318d2c87547e74e5R228)?

mempool/clist_mempool.go
hvanz: It is not and we don't do that. What is safe is to create multiple iterators on the same CList, as we do with one `broadcastTxRoutine` per peer. I'll add a comment to clarify though.

mempool/clist_mempool.go
hvanz: Exactly. It will set cursor to nil and start from the beginning on the next iteration.

mempool/mempool.go
sergio-mena: Good point

mempool/clist_mempool.go
hvanz: The receiver will get `nil` when closing the channel without sending any message on it. This happens when `iter.cursor == nil`. When `iter.cursor != nil`, it first send the message on the channel and then it closes it, but the receiver (line 227) already got the message. (I checked this behaviour on the code.)

sergio-mena (approved):
🎸🎸

mempool/clist_mempool.go
hvanz: To make it more clear I can change it to:
```
if iter.cursor != nil {
    ch <- iter.cursor.Value.(Entry)
} else {
    close(ch)
}
```

mempool/reactor.go
hvanz: In principle, yes, that's what would happen.

Here, `entry` is nil either because `Clist.Front` or `CElement.Next` returned nil. If `Next` returns nil because it's at the end of the list and the list is not empty, we shouldn't set the cursor to nil because we will start iterating from the beginning of the list, sending the same transactions more than once. The other reason I see that both `Front` and `Next` return nil is when the entry got removed just before calling these functions. I've seen these scenarios happen in unit tests, but they are difficult to reproduce.

I think it's better to investigate these scenarios in another issue; this PR should be just a refactoring.

hvanz:
@mergifyio backport v1.x

Issue #3508, opened 2024-07-11
docs: add front matter to ABCI 2.0 tutorial


andynog:
The ABCI 2.0 Forum Application tutorial (#1374) was merged with the application code and the documentation. 

But the front matter for the Markdown files in `/docs/tutorials/forum-application` were missing the front matter on top of the page so they will not render properly in the documentation site.

DoD:

- Add front matter to pages
- Test and ensure the left side navigation on the documentation site display the pages properly in the right order

PR #3509, opened 2024-07-11
docs: Add front matter to ABCI 2.0 tutorial

andynog:
Close: #3508 

Tested locally and left hand navigation displays properly with this fix.

tutorial_navigation

#### PR checklist

### DIFF
--- a/docs/tutorials/forum-application/README.md
+++ b/docs/tutorials/forum-application/README.md
@@ -1,0 +1,16 @@
+---
+order: 1
+parent:
+  title: Forum Application Tutorial
+  order: 3
+---
+
+# ABCI 2.0 Forum Application Tutorial
+
+In this tutorial, we will build a Forum application using [ABCI 2.0](https://docs.cometbft.com/v1.0/spec/abci/).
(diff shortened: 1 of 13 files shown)


docs/tutorials/forum-application/README.md
```diff
@@ -0,0 +1,16 @@
+---
+order: 1
+parent:
+  title: Forum Application Tutorial
+  order: 3
```
melekes, line 5:
```suggestion
  order: 2
```

? because below is 3,4 and 1?

melekes (approved):

melekes:
On the 2nd thought, I think it should be the last item, not the first one, in tutorials.

sergio-mena:
> On the 2nd thought, I think it should be the last item, not the first one, in tutorials.

I agree

sergio-mena (approved):

melekes:
ss

https://github.com/cometbft/cometbft/pull/3509#issuecomment-2224684819 wasn't able to do this. Anyway, this should not block this PR.

cason:
I think the site is build every night or something like that. We probably can do this manually, from https://github.com/cometbft/cometbft-docs.

Issue #3472, opened 2024-07-10
p2p: add `HasChannel` method to `p2p.Peer` interface


cason:
The `p2p.Peer` interface is the way reactors interact (e.g., send messages) to connected peers:

https://github.com/cometbft/cometbft/blob/55493e04ef5977121435f25ae0cb01693b03aba3/p2p/peer.go#L23-L48

A node only connects to a peer if there is an intersection between the channels the node supports and the peer supports. But it is not mandatory that node and peer implement exactly the same set of of channels (although standard upstream nodes will). In this case, there is no point to keep routines attempting to send messages in a channel that a peer does not implement.

In fact, we check if the peer implements a channel before sending a message to it, therefore sending a message to a peer that does not implement a channel will fail (return `false`):

https://github.com/cometbft/cometbft/blob/55493e04ef5977121435f25ae0cb01693b03aba3/p2p/peer.go#L268-L273

But most reactors maintain one (mempool, evidence) or some (consensus, in particular) routines for sending messages to peers. But if the peer does not implement that channel, those channels are just useless, and should not even execute.

## Proposal

Render the already existing method `hasChannel` public and add it to the `p2p.Peer` interface:

https://github.com/cometbft/cometbft/blob/55493e04ef5977121435f25ae0cb01693b03aba3/p2p/peer.go#L306-L315

This will allow not starting useless, and possibly resource intensive (see #3414 ) sending routines in several reactors.

Issue #3414, opened 2024-07-03
Prevent busy-waiting in consensus gossip routines, when their channel's SendQueue is full


ValarDragon:
### Summary

The consensus gossip routines can busy-wait, when their outbound message queues are full. There are several cases where peer.Send fails, and we continue to the outer-loop, and we-will re-enter the queue to send more messages.

We observed what appears to be hitting this busy waiting in CPU profiles in Osmosis mainnet. It was observed when we had sustained 5MB blocks for ~15 minutes. We see tremendously high numbers of calls in gossipDataRoutine the proposal.ToProto, (and gossipDataRoutine in general) taking tremendous amounts of system resources.  (And same in Vote, but on a different hot-path for failing gossip)

Here is what we saw over one hour, with 15 minutes of 5MB blocks:

![image](https://github.com/cometbft/cometbft/assets/6440154/a6f2e355-9fb3-45a7-9cac-571e88d6fe07)

I have never visually noticed proposal.ToProto taking a long time in another prod benchmark before.

Here is what we typically see over one hour (at least relative proportions within here are on par with what I've seen over dozens of profiles):
![image](https://github.com/cometbft/cometbft/assets/6440154/08cf667c-03e6-40c4-a0c9-609a1996c3fd)

So the time (and codepath) looks like were busy waiting to me.

### Proposal

- [ ] Understand and fix however its possible that we filled our SendBuffer
  - Its of size 100, so the only thing I can think of its CPU times in outbound packet clearing (recent PR's I've made help improve) and related to @evanforbes findings on the OS not sending packets due to potential congestion control. (#3346 should help, but won't fix core issue)
- [ ] Prevent this edge case from ever busy waiting. (I'll make a PR)

ValarDragon:
Turns out my BusyWaiting due to `Send` failure doesn't seem right, since I think Send blocks for 10s. I have no idea why we got this pattern then of so many .ToProto calls :thinking: 

I feel like this issue is some sort of BusyWaiting, we can repeatedly do this ToProto call

cason:
> Turns out my BusyWaiting due to Send failure doesn't seem right,

Yes. When `Send` fails we already waited for a while and the peer is probably gone.

cason:
I have in my backlog to check every `Send()` call in the consensus reactor and try to replace it by a `TrySend()` whenever possible.

ValarDragon:
[...]

PR #3510, opened 2024-07-12
feat(p2p): render `HasChannel(chID)` is a public `p2p.Peer` method

cason:
Closes: #3472 

It also prevents reactors from starting routines intended to send messages to a peer that does not implement/support a given channel. Because all `Send()` or `TrySend()` calls from this routine will be useless, always returning `false` and possibly producing some busy-wait behavior (see https://github.com/cometbft/cometbft/issues/3414).

The changes are restricted to: mempool and evidence reactor, because they use a single channel and have a sending routine peer peer, and two of the consensus routines, for Data and Votes.

Block and State sync reactors have smarter ways to deal with unresponsive peers, so probably this check is not needed.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/.changelog/unreleased/features/3472-p2p-has-channel-api.md
+++ b/.changelog/unreleased/features/3472-p2p-has-channel-api.md
@@ -1,0 +1,3 @@
+- `[p2p]` `HasChannel(chID)` method added to the `Peer` interface, used by
+  reactors to check whether a peer implements/supports a given channel.
+  ([#3472](https://github.com/cometbft/cometbft/issues/3472))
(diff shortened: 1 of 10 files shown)


melekes (approved):
👍

melekes:
The changelog entry is missing.

ValarDragon (approved):
Great job!! Will get this into osmosis as soon as approved here

melekes:
@cason should we backport this to v1?

cason:
Not sure, what is your opinion?

PR #3512, opened 2024-07-12
feat(p2p): remove redundant checks on sending methods

### LINKED ISSUE #3511: p2p: remove redundant checks on sending methods
Both `Send()` and `TrySend()` perform this initial check before attempting to send a message to a peer:

https://github.com/cometbft/cometbft/blob/63ef074cc945cc9bd3ca7ce0d8c1ed1ba8cdbf2a/p2p/peer.go#L268-L273

This method, at the end, calls the corresponding method from [`MConnection`](https://github.com/cometbft/cometbft/blob/main/p2p/conn/connection.go), which perform the same checks.

Actually, the existence of the channel is checked in a very similar way, producing an error message that can be relevant for operators. By re-doing this check at the `p2p.peer` implementation, this error is never printed.

The other test is whether the peer is running, as a process. The connection checks whether the connection is running. But a peer is an abstraction that encapsulate a connection: if the connection is stopped/broken, so is the peer.

My suggestion therefore is to remove both checks.

cason:
Closes #3511.

---

#### PR checklist

- [x] Tests written/updated


cason:
Changed the destination branch to prevent lint errors. It should be updated to `main` once https://github.com/cometbft/cometbft/pull/3510 is merged.

melekes (approved):
👍

ValarDragon (approved):
Nice, happy to get this removed from here. I did not understand why we needed either of these either.

(Seemed like they both should be handled at different layers of abstraction, e.g. not gossiping if the peer doesn't have a channel)

cason:
> (Seemed like they both should be handled at different layers of abstraction, e.g. not gossiping if the peer doesn't have a channel)

Yes, after https://github.com/cometbft/cometbft/pull/3512 we have this API method, so the sender can check this condition before using send methods.

PR #3513, opened 2024-07-12
test(e2e): add support for testing different key types, including BLS

sergio-mena:
While working on another crypto to be added to CometBFT, I realised we are not testing other key types in our nightlies, in particular BLS isn't tested.

This PR:
* adds support for BLS in the runner, generator, and `e2e` tests
* adds support for BLS, secp256k1, and ed25519 key types in the `e2e` generator (for the nightlies)
* completes support for changing the key type at testnet level. Some support was existing but it wasn't complete (so it didn't work)

There is a TODO in the BLS crypto side. I added a temporary workaround and contacted the Zondax folks to include a small patch in `github.com/cosmos/crypto`. If they agree to merge the patch, `GenPrivKeyFromSecret` will become trivial.

---

#### PR checklist

- [x] Tests written/updated


crypto/bls12381/key_bls12381.go
```diff
@@ -37,7 +39,20 @@ var _ crypto.PrivKey = &PrivKey{}
...
+// TODO simplify once `github.com/cosmos/crypto/` has `GenPrivKeyFromSeed`
+func GenPrivKeyFromSecret(secret []byte) (PrivKey, error) {
+	seed := sha256.Sum256(secret) // We need 32 bytes
+
+	// Defensive check: have we generated a valid secret key?
+	secretKey := blst.KeyGen(seed[:]).Serialize()
+	if bls12381.IsZero(secretKey) {
+		return nil, errors.New("generated secret key is zero")
+	}
+	return secretKey, nil
```
melekes, line 52:
```suggestion
	return secretKey.Marshal(), nil
```

Do we need to marshal the secret key similar to how we do it in `NewPrivateKeyFromBytes` and `GenPrivKey`?

crypto/bls12381/key_bls12381.go
```diff
@@ -37,7 +39,20 @@ var _ crypto.PrivKey = &PrivKey{}
 
 type PrivKey []byte
 
-// NewPrivateKeyFromBytes build a new key from the given bytes.
+// NewPrivateKeyFromBytes generates a new random key using `secret` for the seed
+// TODO simplify once `github.com/cosmos/crypto/` has `GenPrivKeyFromSeed`
+func GenPrivKeyFromSecret(secret []byte) (PrivKey, error) {
```
melekes, line 44: Do you mind adding a unit test for this method?

sergio-mena: Yes, very good point (waiting for the simplification of the function... depending on `github.com/cosmos/crypto`)

crypto/bls12381/key_bls12381.go
sergio-mena: I think so. Please check the simplified version if you have the chance. It does the `.Marshall()` as you suggest here

crypto/bls12381/key_bls12381.go
sergio-mena: @melekes I added 2 UTs in 2a38ef5. The second panics without the fix I did in prod code. When you are back, please check that I am _fixing_ rather than _introducing_ a bug.

cason (approved):
Legit.

crypto/bls12381/key_bls12381.go
melekes: That doesn't look right. The comment explicitly says: "If msg is larger than MaxMsgLen, SHA256 sum will be signed instead of the raw bytes."

melekes: Fixed in https://github.com/cometbft/cometbft/pull/3513/commits/affc6dfa8d105a590f9d1e7602dbe43c249a8761. See the commit message for the explanation.

melekes (approved):
👍

melekes (approved):
👍

crypto/bls12381/key_bls12381.go
sergio-mena: Cool, thanks for this @melekes !

PR #3523, opened 2024-07-15
fix(tests): fix obvious things in `tests.mk`

sergio-mena:
As per our internal discussion today, here are some changes I think we should have.

Feel free to comment on further changes you may want to have in `tests.mk`

---

#### PR checklist

- ~[ ] Tests written/updated~

### DIFF
--- a/tests.mk
+++ b/tests.mk
@@ -66,7 +66,7 @@ test_deadlock:
 # Implements test splitting and running. This is pulled directly from
 # the github action workflows for better local reproducibility.
 
-GO_TEST_FILES != find $(CURDIR) -name "*_test.go"
+GO_TEST_FILES := $(shell find $(CURDIR) -name "*_test.go")
 
 # default to four splits by default
 NUM_SPLIT ?= 4
--- a/tests.mk
+++ b/tests.mk
@@ -72,6 +72,6 @@ $(BUILDDIR)/packages.txt:$(GO_TEST_FILES) $(BUILDDIR)
 split-test-packages:$(BUILDDIR)/packages.txt
 	split -d -n l/$(NUM_SPLIT) $< $<.
 
-# Used, in special, by the GitHub CI, in order to run tests in parallel 
+# Used by the GitHub CI, in order to run tests in parallel
 test-group-%:split-test-packages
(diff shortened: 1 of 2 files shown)


tests.mk
```diff
@@ -47,22 +47,29 @@ test_release:
...
 test:
 	@echo "--> Running go test"
-	@go test -p 1 $(PACKAGES) -tags deadlock,bls12381
+	@go test -p 1 $(PACKAGES) -tags bls12381
 .PHONY: test
 
 test_race:
 	@echo "--> Running go test --race"
-	@go test -p 1 -v -race $(PACKAGES)
+	@go test -p 1 -race $(PACKAGES)
```
cason, line 58: No bls here?

tests.mk
```diff
@@ -47,22 +47,29 @@ test_release:
...
+
+# The format statement filters out all packages that don't have tests.
+# Note we need to check for both in-package tests (.TestGoFiles) and
+# out-of-package tests (.XTestGoFiles).
+$(BUILDDIR)/packages.txt:$(GO_TEST_FILES) $(BUILDDIR)
+	go list -f "{{ if (or .TestGoFiles .XTestGoFiles) }}{{ .ImportPath }}{{ end }}" ./... | sort > $@
+
+split-test-packages:$(BUILDDIR)/packages.txt
+	split -d -n l/$(NUM_SPLIT) $< $<.
+test-group-%:split-test-packages
```
cason, line 74:
```suggestion

# Used, in special, by the GitHub CI, in order to run tests in parallel 
test-group-%:split-test-packages
```

cason (approved):
+1

tests.mk
sergio-mena: Good catch

tests.mk
```diff
@@ -47,22 +47,29 @@ test_release:
...
-	@go test -p 1 -v -race $(PACKAGES)
+	@go test -p 1 -race $(PACKAGES)
 .PHONY: test_race
 
 test_deadlock:
-	@echo "--> Running go test --deadlock"
-	@go test -p 1 -v  $(PACKAGES) -tags deadlock
-.PHONY: test_race
+	@echo "--> Running go test with deadlock support"
+	@go test -p 1 -v  $(PACKAGES) -tags deadlock,bls12381
```
sergio-mena, line 63:
```suggestion
	@go test -p 1 $(PACKAGES) -tags deadlock,bls12381
```

PR #3524, opened 2024-07-16
chore(abci): update comments that do not start with the name of the exported element

hoanguyenkh:
Refactor code and update comments

---

#### PR checklist

- [ ] Tests written/updated


abci/types/util.go
```diff
@@ -23,7 +23,7 @@ func (v ValidatorUpdates) Len() int {
 	return len(v)
 }
 
-// XXX: doesn't distinguish same validator with different power.
+// Less doesn't distinguish the same validator with different power.
```
cason, line 26: I wouldn't remove the tag here. This should be a FIXME, TODO?

cason (commented):
Thank you for that, left some comments.

abci/types/util.go
hoanguyenkh: i don't know what reason is, let me rollback the comment

melekes (approved):
👍

melekes:
Thanks @hoanguyenkh ❤️

PR #3525, opened 2024-07-16
fix(types): Genesis doc validation of validators Pubkey types.

### LINKED ISSUE #3514: Genesis file validation doesn't check the validators' `PubKey` type.
### Context
The function validating the genesis file:
https://github.com/cometbft/cometbft/blob/63ef074cc945cc9bd3ca7ce0d8c1ed1ba8cdbf2a/types/genesis.go#L69
does not check if each validator uses a `PubKey` type that we accept, that is, it appears in `genesis.ConsensusParams.Validator.PubKeyTypes`.

Consequently, we incorrectly allow a chain to start even if a validator uses an unsupported `PubKey` type.

### The Change
Add a check that, for each validators in the genesis, makes sure that their key type is contained in the list `genesis.ConsensusParams.Validator.PubKeyTypes`.

alesforz:
Closes #3514.

### Changes
- Added unit test demonstrating the problem.
- Added code in `ValidateAndComplete` that, for each validator in the genesis doc, checks if their `PubKey.Type()` is present in the accepted `PubKeyTypes` of the chain.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/types/genesis.go
+++ b/types/genesis.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"slices"
 	"time"
 
 	"github.com/cometbft/cometbft/crypto"
@@ -86,6 +87,7 @@ func (genDoc *GenesisDoc) ValidateAndComplete() error {
 		return err
 	}
 
+	acceptedPubKeyTypes := genDoc.ConsensusParams.Validator.PubKeyTypes
 	for i, v := range genDoc.Validators {
 		if v.Power == 0 {
 			return fmt.Errorf("the genesis file cannot contain validators with no voting power: %v", v)
@@ -95,6 +97,11 @@ func (genDoc *GenesisDoc) ValidateAndComplete() error {
 		}
 		if len(v.Address) == 0 {
 			genDoc.Validators[i].Address = v.PubKey.Address()
+		}
+
+		if !slices.Contains(acceptedPubKeyTypes, v.PubKey.Type()) {
+			formatStr := "validator %v uses an unsupported pubkey type: %q"
+			return fmt.Errorf(formatStr, v, v.PubKey.Type())
 		}
 	}
 
--- a/types/genesis.go
+++ b/types/genesis.go
@@ -113,7 +113,7 @@ func GenesisDocFromJSON(jsonBlob []byte) (*GenesisDoc, error) {
 	genDoc := GenesisDoc{}
 	err := cmtjson.Unmarshal(jsonBlob, &genDoc)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("invalid json for GenesisDoc: %s", err)
 	}
 
 	if err := genDoc.ValidateAndComplete(); err != nil {
--- a/types/genesis_test.go
+++ b/types/genesis_test.go
@@ -68,7 +68,6 @@ func TestGenesisBad(t *testing.T) {
 
 	for i, testCase := range testCases {
 		_, err := GenesisDocFromJSON(testCase)
-		t.Log(err)
 		formatStr := "test case %i: expected error for invalid genesis doc"
 		require.Error(t, err, formatStr, i)
 	}
--- a/types/genesis_test.go
+++ b/types/genesis_test.go
@@ -48,6 +48,28 @@ func TestGenesisBad(t *testing.T) {
 				`},"power":"10","name":""}` +
 				`]}`,
 		),
+		// unsupported validator pubkey type
+		[]byte(
+			`{
+				"genesis_time": "0001-01-01T00:00:00Z",
+				"chain_id": "test-chain-QDKdJr",
+				"initial_height": "1000",
+				"validators": [{
+					"pub_key":{"type":"tendermint/PubKeyEd25519","value":"AT/+aaL1eB0477Mud9JMm8Sh8BIvOYlPGC9KkIUmFaE="},
+					"power":"10",
+					"name":""
+				}],
+				"app_hash":"",
+				"app_state":{"account_owner": "Bob"},
+				"consensus_params": {
+					"synchrony":  {"precision": "1", "message_delay": "10"},
+					"validator": {"pub_key_types":["secp256k1"]},
+					"block": {"max_bytes": "100"},
+					"evidence": {"max_age_num_blocks": "100", "max_age_duration": "10"},
+					"feature": {"vote_extension_enable_height": "0", "pbts_enable_height": "0"}
+				}
+			}`,
+		),
 	}
 
 	for _, testCase := range testCases {
--- a/types/genesis_test.go
+++ b/types/genesis_test.go
@@ -51,6 +51,7 @@ func TestGenesisBad(t *testing.T) {
 		// unsupported validator pubkey type
 		[]byte(
 			`{
+				"chain_id": "test-chain-QDKdJr",
 				"validators": [{
 					"pub_key":{"type":"tendermint/PubKeyEd25519","value":"AT/+aaL1eB0477Mud9JMm8Sh8BIvOYlPGC9KkIUmFaE="},
 					"power":"10",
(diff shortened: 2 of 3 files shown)


types/genesis.go
```diff
@@ -96,6 +98,11 @@ func (genDoc *GenesisDoc) ValidateAndComplete() error {
 		if len(v.Address) == 0 {
 			genDoc.Validators[i].Address = v.PubKey.Address()
 		}
+
+		if !slices.Contains(acceptedPubKeyTypes, v.PubKey.Type()) {
+			formatStr := "validator %v uses an unsupported pubkey type: %q"
+			return fmt.Errorf(formatStr, v, v.PubKey.Type())
```
cason, line 104: I'd merge the two lines, as we do in the remaining of the code.

types/genesis.go
```diff
@@ -113,7 +120,7 @@ func GenesisDocFromJSON(jsonBlob []byte) (*GenesisDoc, error) {
 	genDoc := GenesisDoc{}
 	err := cmtjson.Unmarshal(jsonBlob, &genDoc)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("invalid json for GenesisDoc: %s", err)
```
cason, line 123: Good catch, making errors clear are good. I wonder if a message like "failed to unmarshall" would be better.

types/genesis_test.go
```diff
@@ -48,6 +48,28 @@ func TestGenesisBad(t *testing.T) {
 				`},"power":"10","name":""}` +
 				`]}`,
 		),
+		// unsupported validator pubkey type
+		[]byte(
+			`{
+				"genesis_time": "0001-01-01T00:00:00Z",
```
cason, line 54: This is actually optional, can be removed from the test.

types/genesis_test.go
```diff
@@ -48,6 +48,28 @@ func TestGenesisBad(t *testing.T) {
...
+				"chain_id": "test-chain-QDKdJr",
+				"initial_height": "1000",
```
cason, line 56: Same here.

types/genesis_test.go
```diff
@@ -48,6 +48,28 @@ func TestGenesisBad(t *testing.T) {
...
+				"validators": [{
+					"pub_key":{"type":"tendermint/PubKeyEd25519","value":"AT/+aaL1eB0477Mud9JMm8Sh8BIvOYlPGC9KkIUmFaE="},
+					"power":"10",
+					"name":""
+				}],
+				"app_hash":"",
+				"app_state":{"account_owner": "Bob"},
+				"consensus_params": {
+					"synchrony":  {"precision": "1", "message_delay": "10"},
```
cason, line 65: Same here.

types/genesis_test.go
```diff
@@ -48,6 +48,28 @@ func TestGenesisBad(t *testing.T) {
...
+					"validator": {"pub_key_types":["secp256k1"]},
+					"block": {"max_bytes": "100"},
+					"evidence": {"max_age_num_blocks": "100", "max_age_duration": "10"},
+					"feature": {"vote_extension_enable_height": "0", "pbts_enable_height": "0"}
```
cason, line 69: Same here.

types/genesis_test.go
```diff
@@ -48,6 +48,28 @@ func TestGenesisBad(t *testing.T) {
...
+			`{
+				"genesis_time": "0001-01-01T00:00:00Z",
+				"chain_id": "test-chain-QDKdJr",
+				"initial_height": "1000",
+				"validators": [{
+					"pub_key":{"type":"tendermint/PubKeyEd25519","value":"AT/+aaL1eB0477Mud9JMm8Sh8BIvOYlPGC9KkIUmFaE="},
+					"power":"10",
+					"name":""
+				}],
+				"app_hash":"",
```
cason, line 62: Same here.

types/genesis_test.go
```diff
@@ -48,6 +48,28 @@ func TestGenesisBad(t *testing.T) {
...
+				"app_state":{"account_owner": "Bob"},
```
cason, line 63: Same here.

cason (commented):
Good work.

I'd make the test scenario really minimal. Other policy we have is to add a test unit that fails in one commit, then the code that corrects the test in another commit.

Also, I don't think this is a `feat(crypto)` as it refers to the genesis file (`types` package), could be also `genesis`.

types/genesis.go
alesforz: I write this sort of code in 2 lines when I think it increases readability.
When there are long error messages, the `fmt.Errof(...)` line is so long that it wraps to the next line.

I could update the other error messages to do this as well (in fact, in my editor all the other messages wrap to the next line, making it a bit awkward to read). What do you think?

types/genesis.go
alesforz: I think writing errors in the form "failed to do X" or "error doing X" is a bit redundant. It's an error, so we already know that something failed to do something 😄 

When writing errors, I try as much as possible to follow the advice [here](https://google.github.io/styleguide/go/best-practices#adding-information-to-errors), because I think it leads to better, more useful errors.

But if there is an error style we use in the codebase, I'll follow it. 

Any thoughts?

types/genesis_test.go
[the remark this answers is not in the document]
alesforz: What do you mean with "non-standard" and "standard" key type?

My understanding is that the validators declare their key types using key's _name_, that is:

- [tendermint/PubKeyEd25519](https://github.com/cometbft/cometbft/blob/93fcedc4fc522b980f668a5953190e3979ea116c/crypto/ed25519/ed25519.go#L50)
- [tendermint/PubKeySecp256k1 ](https://github.com/cometbft/cometbft/blob/93fcedc4fc522b980f668a5953190e3979ea116c/crypto/secp256k1/secp256k1.go#L22)
- [cometbft/PubKeyBls12_381](https://github.com/cometbft/cometbft/blob/93fcedc4fc522b980f668a5953190e3979ea116c/crypto/bls12381/const.go#L17)
- [tendermint/PubKeySr25519](https://github.com/cometbft/cometbft/blob/93fcedc4fc522b980f668a5953190e3979ea116c/crypto/sr25519/encoding.go#L7) 

Whereas the consensus parameters `ValidatorParams` lists the accepted key type by using the key's _type_, that is:

- [ed25519](https://github.com/cometbft/cometbft/blob/93fcedc4fc522b980f668a5953190e3979ea116c/crypto/ed25519/ed25519.go#L62C13-L62C20)
- [secp256k1](https://github.com/cometbft/cometbft/blob/93fcedc4fc522b980f668a5953190e3979ea116c/crypto/secp256k1/secp256k1.go#L24C17-L24C26)
- [bls12_381](https://github.com/cometbft/cometbft/blob/93fcedc4fc522b980f668a5953190e3979ea116c/crypto/bls12381/const.go#L11C13-L11C22)
- [sr25519](https://github.com/cometbft/cometbft/blob/93fcedc4fc522b980f668a5953190e3979ea116c/crypto/sr25519/privkey.go#L25C13-L25C20)

And that's what the unit test I wrote does. Am I missing or misunderstanding something?

types/genesis.go
cason: We don't have a strict policy regarding that, for me honestly is the same. I try to follow what I see more often in the remaining of the code, which is not great but at least preserve some "style".

Really optional here.

types/genesis.go
cason: Good point.

types/genesis_test.go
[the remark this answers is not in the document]
cason: Standard is what comes with `types.DefaultValidatorParams()`, namely, what is considered when we don't set this field of the consensus parameter "struct".

Right now, the standard is `ABCIPubKeyTypeEd25519   = ed25519.KeyType`.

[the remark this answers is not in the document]
cason: But I understand that maintaining this code if we change the standards wouldn't be trivial, so I am resolving this.

types/genesis_test.go
```diff
@@ -48,6 +48,21 @@ func TestGenesisBad(t *testing.T) {
...
+				"consensus_params": {
+					"validator": {"pub_key_types":["secp256k1"]},
+					"block": {"max_bytes": "100"},
+					"evidence": {"max_age_num_blocks": "100", "max_age_duration": "10"},
+				}
+			}`,
+		),
 	}
 
 	for _, testCase := range testCases {
```
cason, line 68: Do you mind improving the error message in line 70?

cason (approved):
Good to go.

types/genesis_test.go
alesforz: Just saw your message. I pushed a commit with an error message that better reflects the test's purpose.

PR #3526, opened 2024-07-17
fix(node): Check that the Sha256Checksum in a ChecksummedGenesisDoc isn't nil.

### LINKED ISSUE #2938: Effects of a nil `Sha256Checksum` in a `ChecksummedGenesisDoc`
The `GenesisDocProvider` is a type alias for a func
https://github.com/cometbft/cometbft/blob/ce68e90dadb89c3162b8b549dfc3da3c10ec0aa8/node/setup.go#L55

When the logic was implemented, I believe that the expectation was that this `GenesisDocProvider` was going to be 'provided' using `DefaultGenesisDocProviderFunc`, which calculates the sha256 checksum from the jsonBlob of the genesis file content
https://github.com/cometbft/cometbft/blob/ce68e90dadb89c3162b8b549dfc3da3c10ec0aa8/node/setup.go#L59

The problem is that this provider could come from a custom provider. For example, the Cosmos-SDK has a different genesis signature and it needs to pass it's own provider logic, something like:
```
	appGenesisProvider := func() (node.ChecksummedGenesisDoc, error) {
		appGenesis, err := genutiltypes.AppGenesisFromFile(cmtCfg.GenesisFile())
		if err != nil {
			return node.ChecksummedGenesisDoc{}, err
		}
		gen, err := appGenesis.ToGenesisDoc()
		if err != nil {
			return node.ChecksummedGenesisDoc{}, err
		}
		return node.ChecksummedGenesisDoc{GenesisDoc: gen}, nil
	}
```
which returns a `func() (ChecksummedGenesisDoc, error)` that can be used in a `node.NewNode()` as the `GenesisDocProvider`, e.g.
```
tmNode, err := node.NewNode(
		context.TODO(),
		...
		appGenesisProvider,
		...
	)
```

but this might cause issues in 
https://github.com/cometbft/cometbft/blob/ce68e90dadb89c3162b8b549dfc3da3c10ec0aa8/node/setup.go#L573

because the `Sha256Checksum` field is not validated (look at the example above (`return node.ChecksummedGenesisDoc{GenesisDoc: gen}`) so it will be nil. 

It will fail when trying to save it in the `stateDB` here because a `nil` value cannot be saved on the database
https://github.com/cometbft/cometbft/blob/ce68e90dadb89c3162b8b549dfc3da3c10ec0aa8/node/setup.go#L606

and there is other logic in this function that relies on a `Sha256Checksum`, so in my opinion, we might need to account for that in the code or at least add more tests. I agree that in part this is caused by a integrator not adding the checksum (maybe we need to add some more information to the docs to say that is expected), but at the same time as a defensive measure we might need to ensure we can handle this scenario and that it will not cause other issues in the future.

alesforz:
Closes #2938.

### Changes
- Added unit test for the affected function `LoadStateFromDBOrGenesisDocProviderWithConfig` to showcase the problem

---

#### PR checklist

- [x] Tests written/updated


alesforz:
I'm not sure whether it's worth it to add an additional nil check `ChecksummedGenesisDoc.Sha256Checksum` in the affected function.
Even if it's nil, the function does fail gracefully with an error, i.e., it doesn't cause a crash.

I would rather add docs to indicate that custom implementation of `GenesisDocProvider` must return a _complete_ `ChecksummedGenesisDoc`.

sergio-mena:
@alesforz thanks for this. In the case of `nil` (i.e., what's described in the issue) I agree with you. But what happens if the `genesisDocProvider` functor returns e.g., a `Sha256Checksum` with the wrong length. If you modify the unit test you wrote so that `genDocProvider` returns a bad-length checksum the test should also fail.

node/setup_test.go
cason, line 1: I'd move this to the existing `node_test.go` file.

cason:
I think we must be sure that `NewNode` fails with an error when a bad `genDocProvider` is provided to the constructor. This is probably already happening, as you show with this test unit.

cason:
In other words, to have this merged, we need a test that passes. It should require an error under some circumstances, specifically when the provided  `genDocProvider` is bad.

But at the end, if someone uses this method with a non-standard `genDocProvider`, an error is returned and the node is not created. We have maybe to check if the error message is clear enough.

alesforz:
I did some digging: apart from the unit test that this PR adds, there are only 2 unit tests testing that `NewNode` returns an error [here](https://github.com/cometbft/cometbft/blob/2fa55d440c36e8157261da694c0a65d5d510041a/node/node.go#L295): [`TestNodeNewNodeGenesisHashMismatch`](https://github.com/cometbft/cometbft/blob/2fa55d440c36e8157261da694c0a65d5d510041a/node/node_test.go#L543) and [`TestNodeGenesisHashFlagMismatch`](https://github.com/cometbft/cometbft/blob/2fa55d440c36e8157261da694c0a65d5d510041a/node/node_test.go#L637).  However they only check if there is a genesis hash mismatch [here](https://github.com/cometbft/cometbft/blob/2fa55d440c36e8157261da694c0a65d5d510041a/node/setup.go#L615) and [here](https://github.com/cometbft/cometbft/blob/2fa55d440c36e8157261da694c0a65d5d510041a/node/setup.go#L604) respectively.

@cason to your point:
> In other words, to have this merged, we need a test that passes. It should require an error under some circumstances, specifically when the provided `genDocProvider` is bad.
> 
> But at the end, if someone uses this method with a non-standard `genDocProvider`, an error is returned and the node is not created. We have maybe to check if the error message is clear enough.

it seems to me that `LoadStateFromDBOrGenesisDocProviderWithConfig` *does* return an error if given a bad `genDocProvider`. It calls [`ValidateAndComplete`](https://github.com/cometbft/cometbft/blob/2fa55d440c36e8157261da694c0a65d5d510041a/node/setup.go#L593) on the genesis doc and checks if the SHA256 checksum is nil or a mismatch. In both cases, it will return an error, thus causing the call to `NewNode` to fail. I agree we have to make sure the error messages are clear enough.

Additional notes:
- `LoadStateFromDBOrGenesisDocProviderWithConfig` does not do a sanity check on the checksum itself. I think the unit test that @sergio-mena suggested above will pass, i.e., it will execute [this line](https://github.com/cometbft/cometbft/blob/2fa55d440c36e8157261da694c0a65d5d510041a/node/setup.go#L610) if called with an empty `operatorGenesisHashHex` and if somehow the length of the genesis doc in the DB is 0. In this case, the call to `SetSync` will succeed, as the function only checks if the checksum is nil, not if it is invalid, shorter, etc. Is it possible that these preconditions will be satisfied at runtime?
- We do not unit test the majority of error returns of `LoadStateFromDBOrGenesisDocProviderWithConfig` except those I'm talking about here. Should we try to add them?

alesforz:
> @alesforz thanks for this. In the case of `nil` (i.e., what's described in the issue) I agree with you. But what happens if the `genesisDocProvider` functor returns e.g., a `Sha256Checksum` with the wrong length. If you modify the unit test you wrote so that `genDocProvider` returns a bad-length checksum the test should also fail.

Just added this unit test. The function does not return an error, but the test fails because the test expects one.
So at least we should do some sanity check on the checksum itself.

node/setup_test.go
```diff
@@ -9,40 +9,66 @@ import (
...
-		cfg.Storage.GenesisHash,
-		nil,
-	)
-	if err != nil {
-		t.Errorf("test failed with error: %s", err)
+			return checksummedGenesisDoc, nil
+		}
 	}
+
+	t.Run("NilGenesisChecksum", func(t *testing.T) {
```
sergio-mena, line 47: Excellent refactoring!

sergio-mena:
> but the test fails because the test expects one

Awesome. Way to go! First the failing test, then we fix it. Please go ahead!

When you're done fixing the code you can mark this as "ready for review".

alesforz:
> > but the test fails because the test expects one
> 
> Awesome. Way to go! First the failing test, then we fix it. Please go ahead!
> 
> When you're done fixing the code you can mark this as "ready for review".

Ok, so I'll write a sanity check for the SHA256 checksum.
I can easily write a _syntactic_ check on it (i.e., is it 64 bytes long? is it a hex string?). However, is it sufficient? Do we check anywhere else that the given checksum is the actual checksum of the genesis doc?
Meaning: if a user gives a custom `genDocProvider`, shouldn't we check whether the `ChecksummedGenesisDoc` contains the correct hash rather than a hex string of garbage (or the hash of another thing)? A syntactic check will tell us if `ChecksummedGenesisDoc.Sha256Checksum` is a valid SHA256 hash, not whether it is the SHA256 _we want_. 

As far as I can tell, we don't do such a check. if called with an empty `operatorGenesisHashHex` and if no previous genesis doc in the DB, the function will store in the DB anything it's given as the SHA256 checksum, including an invalid checksum, or a valid random hash.

Doing a proper check entails re-hashing the `ChecksummedGenesisDoc.GenesisDoc` and checking if it matches with the given `ChecksummedGenesisDoc.Sha256Checksum`. However, we have to re-marshal the genesis doc to JSON to do so and, with our current JSON lib, this means loading it all in memory which can be expensive.

Any thoughts? @cason @sergio-mena

cason:
I think that nodes have to agree on block 1 (the first block) on the hash of the genesis (block 0), right? So if a node comes up with a random hash, it will not accept block 1.

cason:
Again, this is hard to check, as we are somehow testing a method that is passed to the constructor and can have any implementation.

sergio-mena:
Adding on to @cason's thoughts.
I think some (basic) fast, syntactic checks should be enough to close #2938. Let's not get down a rabbit hole here: as @cason says, if the hash we are passed is not the hash we want, we'll know it early enough. But basic syntactic checks are cheap, and would help speed up troubleshooting if that's the source of the problem.

alesforz:
> I think that nodes have to agree on block 1 (the first block) on the hash of the genesis (block 0), right? So if a node comes up with a random hash, it will not accept block 1.

Ok, so we could get away with only having a syntactic sanity check on the checksum. 

> Again, this is hard to check, as we are somehow testing a method that is passed to the constructor and can have any implementation.

We cannot check the implementation of the closure of course, but we can double-check what it returns: the genesis doc goes through a call to `ValidateAndComplete`. But the only way to validate the checksum is to re-hash the genesis doc ourselves and verify it is equal with the checksum that the closure returns. 

At the moment, the latter is doable in 2 ways:
1. we could re-marshal to json the genesis doc object, then hash it and check it is equal with the one the closure returns. This will be expensive in terms of memory.
2. if we had access to the genesis _file_ on disk, we could hash it directly without loading the entire thing into memory. However, do we have access to it at this stage?

cason (approved):
Looks good.

But I would still use the existing `node_test.go` instead of introducing a `setup_test.go`.

node/setup.go
```diff
@@ -594,6 +594,12 @@ func LoadStateFromDBOrGenesisDocProviderWithConfig(
 		return sm.State{}, nil, ErrGenesisDoc{Err: err}
 	}
 
+	checkSumStr := hex.EncodeToString(csGenDoc.Sha256Checksum)
+	if err := tmhash.ValidateSHA256(checkSumStr); err != nil {
+		const formatStr = "invalid genesis doc SHA256 checksum: %s"
```
melekes, line 599: What's the benefit of this? If the `const` is not reused anywhere?

alesforz: It shows me at a glance that `formatStr` is immutable, plus there is a performance benefit for using `const`, albeit  _minor_ in this case.
Tbh, it's my code-style preference: if a function call using a `string` is long enough to wrap to a new line in my editor, I create a separate `string` like I did here. It's all for readability.

PR #3527, opened 2024-07-17
chore: update cometbft Go version to 1.22.5

alesforz:
This PR bumps the go version to 1.22.5.
To be more precise, we maintain Go 1.22 as the minimum required version to compile the project, but we bump to Go 1.22.5 our CI suite of tests and Dockerfile(s).

---

#### PR checklist

~- [ ] Tests written/updated~


alesforz:
2 tests fail because they are still pulling Docker images with go 1.22.2. We also need to push new Docker images with go 1.22.5.

cason:
I'd suggest you to check other Go bumping PRs in this repo, as we never know what does it affect.

api/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/cometbft/cometbft/api
 
-go 1.22
+go 1.22.5
```
melekes, line 3: Do we have to be specific here? Can't we simply require `1.22`?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/cometbft/cometbft
 
-go 1.22.2
+go 1.22.5
```
melekes, line 3: same question

README.md
```diff
@@ -67,7 +67,7 @@ Please see [SECURITY.md](./SECURITY.md).
 
 | CometBFT version | Requirement | Version        | Tested with |
 |------------------|-------------|----------------|-------------|
-| main             | Go version  | 1.22 or higher | 1.22        |
+| main             | Go version  | 1.22 or higher | 1.22.5      |
```
melekes, line 70:
```suggestion
| main             | Go version  | 1.22 or higher | 1.22      |
```

That might create the illusion that we only tested this with 1.22.5, but we also tested CometBFT with other minor Go releases.

alesforz: I understood the column `Tested with` as "we tested with Go versions up to 1.22.5". Maybe we could update it to say "up to 1.22.5" to make it clearer?

api/go.mod
alesforz: How did the team handle previous Go version bumps? Only bumps to major releases or also minor revisions?
I thought it was the latter, because cometbft `go.mod` is currently set to the minor revision 1.22.2, thus indicating that's the minimum version we want people to compile with. 

I suppose we can set 1.22 as a minimum version, but leave Dockerfiles and CI-related things to 1.22.5 to be consistent with the `README.md` saying that we tested with 1.22.5. 

What do you think?

go.mod
alesforz: see above :)

api/go.mod
melekes: Agree. Because we don't need to enforce a minor version unless we use some language features present in X or there's a CVE that is only fixed in X.

melekes: Right. Anyway, I don't think it's our job to enforce the Go version with some fixed CVEs.

melekes (approved):
👍

PR #3530, opened 2024-07-22
fix(ci): run `make all` error 'No rule to make target 'check''

Halimao:
Running `make all` would cause an error: `make: *** No rule to make target 'check', needed by 'all'.  Stop.`

The `check` target has already removed, let's delete it


cason:
The `check` entry was removed by https://github.com/tendermint/tendermint/pull/3949.

It was replaced by a `tools.mk`, later removed by https://github.com/tendermint/tendermint/pull/6102.

cason (approved):
Thanks.

sergio-mena (approved):
LGTM

PR #3541, opened 2024-07-22
chore(state): Copy `Iterator` `Key()` and `Value()` APIs return values.

alesforz:
Related to cometbft/cometbft-db#168.

#### Changes
We recently updated cometbft-db `Iterator` type `Key()` and `Value()` APIs to return their values directly instead of a copy.
Therefore, this PR ensures that code in cometbft that modifies or stores references to the return value of those APIs creates a copy before continuing.

---

#### PR checklist

~- [ ] Tests written/updated~

### DIFF
--- a/abci/example/kvstore/kvstore.go
+++ b/abci/example/kvstore/kvstore.go
@@ -494,8 +494,13 @@ func (app *Application) getValidators() (validators []types.ValidatorUpdate) {
 	}
 	for ; itr.Valid(); itr.Next() {
 		if isValidatorTx(itr.Key()) {
-			validator := new(types.ValidatorUpdate)
-			err := types.ReadMessage(bytes.NewBuffer(itr.Value()), validator)
+			var (
+				validator       = new(types.ValidatorUpdate)
+				value           = itr.Value()
+				valueWorkingCpy = make([]byte, len(value))
+			)
+			copy(valueWorkingCpy, value)
+			err := types.ReadMessage(bytes.NewBuffer(valueWorkingCpy), validator)
 			if err != nil {
 				panic(err)
 			}
--- a/abci/example/kvstore/kvstore.go
+++ b/abci/example/kvstore/kvstore.go
@@ -494,13 +494,8 @@ func (app *Application) getValidators() (validators []types.ValidatorUpdate) {
 	}
 	for ; itr.Valid(); itr.Next() {
 		if isValidatorTx(itr.Key()) {
-			var (
-				validator       = new(types.ValidatorUpdate)
-				value           = itr.Value()
-				valueWorkingCpy = make([]byte, len(value))
-			)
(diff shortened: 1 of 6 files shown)


sergio-mena:
If/when this is ready for review, please hit the "ready for review" button

abci/example/kvstore/kvstore.go
```diff
@@ -494,8 +494,13 @@ func (app *Application) getValidators() (validators []types.ValidatorUpdate) {
...
 		if isValidatorTx(itr.Key()) {
-			validator := new(types.ValidatorUpdate)
-			err := types.ReadMessage(bytes.NewBuffer(itr.Value()), validator)
+			var (
+				validator       = new(types.ValidatorUpdate)
+				value           = itr.Value()
+				valueWorkingCpy = make([]byte, len(value))
+			)
+			copy(valueWorkingCpy, value)
+			err := types.ReadMessage(bytes.NewBuffer(valueWorkingCpy), validator)
```
melekes, line 503: Why do we need a copy here? We're unmarshaling value using Protobuf.

alesforz: You are right, it's unnecessary to copy here. I added the copy thinking about the `bytes.NewBuffer` call, whose docs say `The new [Buffer] takes ownership of buf, and the caller should not use buf after this call`. But going down the rabbit hole of calls from `types.ReadMessage` I see we only read the value.
I will revert this commit, thanks for catching this! 😄

melekes (approved):
Thanks @alesforz ❤️ 

This PR should also bump cometbft-db version imho.

alesforz:
> Thanks @alesforz ❤️
> 
> This PR should also bump cometbft-db version imho.

Shouldn't we first bump the Go version of cometBFT to 1.22.5, i.e., the same version that the latest cometbft-db uses?

melekes:
> > Thanks @alesforz ❤️
> > This PR should also bump cometbft-db version imho.
> 
> Shouldn't we first bump the Go version of cometBFT to 1.22.5, i.e., the same version that the latest cometbft-db uses?

yes

alesforz:
> > > Thanks @alesforz ❤️
> > > This PR should also bump cometbft-db version imho.
> > 
> > 
> > Shouldn't we first bump the Go version of cometBFT to 1.22.5, i.e., the same version that the latest cometbft-db uses?
> 
> yes

There is now a [separate PR](https://github.com/cometbft/cometbft/pull/3596) bumping cometbft-db.
I would merge this PR first though, so that the change in the `Iterator` APIs will be in place before the update to cometbft-db.
I'm afraid that if we update cometbft-db version first, we will have  a `HEAD` status that uses cometbft-db v0.13.0 but that does not copy `Key()` and `Value()`, thus causing problems.

PR #3557, opened 2024-07-25
fix(indexer): Ineffective or missing break statements in kv package.

### LINKED ISSUE #3544: fix(indexer): Ineffective or missing `break` statements in `kv` package.
### The Problem
There are a [few](https://github.com/cometbft/cometbft/blob/1bdb7c5604e976a39c266a7a6251f652c83c1787/state/indexer/block/kv/kv.go#L630)  [lines](https://github.com/cometbft/cometbft/blob/1bdb7c5604e976a39c266a7a6251f652c83c1787/state/indexer/block/kv/kv.go#L677) [of](https://github.com/cometbft/cometbft/blob/1bdb7c5604e976a39c266a7a6251f652c83c1787/state/indexer/block/kv/kv.go#L493) [code](https://github.com/cometbft/cometbft/blob/1bdb7c5604e976a39c266a7a6251f652c83c1787/state/indexer/block/kv/kv.go#L524) that have incorrect behavior.

Namely, the `break` statements only exit the enclosing `select`, not the `for` loop. To exit the `for` loop, we should add a label to it.
e.g.,:
```go
loop:
	for ; it.Valid(); it.Next() {
		// do stuff...
		
		select {
		case <-ctx.Done():
			break loop

		default:
		}
	}
```

as the code stands, it will continue the `for` loop even if the `Context.context` expires and fires the `ctx.Done()` channel.
Note that in the same file there is [similar](https://github.com/cometbft/cometbft/blob/1bdb7c5604e976a39c266a7a6251f652c83c1787/state/indexer/block/kv/kv.go#L394) [code](https://github.com/cometbft/cometbft/blob/1bdb7c5604e976a39c266a7a6251f652c83c1787/state/indexer/block/kv/kv.go#L711) that correctly breaks out of the `select` statement using a label.

### Goals
- Add a `for` loop label where it's missing
- Have the faulty `break` statements point to the label

alesforz:
Closes #3544.

### Changes
- Adds missing `for` loop labels in the kv indexer 
- Previously ineffective `break` statements now point to their enclosing `for` loop labels to exit upon reception on the `ctx.Done()` channel.

---

#### PR checklist

~- [ ] Tests written/updated~
~- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments~

### DIFF
--- a/.changelog/unreleased/bug-fixes/3544-indexer-break-statement.md
+++ b/.changelog/unreleased/bug-fixes/3544-indexer-break-statement.md
@@ -1,3 +1,3 @@
-- `[indexer]` Fixed ineffective selec break statements; they now
+- `[indexer]` Fixed ineffective select break statements; they now
   point to their enclosing for loop label to exit
   ([\#3544](https://github.com/cometbft/cometbft/issues/3544))
--- a/.changelog/unreleased/bug-fixes/3544-indexer-break-statement.md
+++ b/.changelog/unreleased/bug-fixes/3544-indexer-break-statement.md
@@ -1,0 +1,3 @@
+- `[indexer]` Fixed ineffective selec break statements; they now
+  point to their enclosing for loop label to exit
+  ([\#3544](https://github.com/cometbft/cometbft/issues/3544))
--- a/state/indexer/block/kv/kv.go
+++ b/state/indexer/block/kv/kv.go
@@ -491,7 +491,7 @@ LOOP:
 
 		select {
 		case <-ctx.Done():
-
+			break LOOP
 		default:
 		}
 	}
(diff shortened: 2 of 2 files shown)


.changelog/unreleased/bug-fixes/3544-indexer-break-statement.md
```diff
@@ -0,0 +1,3 @@
+- `[indexer]` Fixed ineffective selec break statements; they now
```
sergio-mena, line 1:
```suggestion
- `[indexer]` Fixed ineffective select break statements; they now
```

sergio-mena (approved):
The breaks you've _modified_ LGTM. The breaks you've _introduced_ also LGTM, but I have less confidence on my review there.
I'd wait for @jmalicevic to take a quick look at those introduced `break` statements.

melekes (approved):
👍

PR #3595, opened 2024-07-30
refactor(config/node)!: Remove genesisHash from config

### LINKED ISSUE #3547: refactor(config): Remove genesis hash from config
## Genesis hash saved in config file but read-only

#1324 saves the genesis hash in the config file to check it against the has from the DB. However, its sole purpose is to [forward](https://github.com/cometbft/cometbft/blob/fbca5acef2a4410d28c3b11baf364dd4c788190c/node/node.go#L295) this config flag to be checked on node boot-up against the hash of the genesis file if it is the first boot or the hash stored in the DB if the node has already read in a genesis file before. The user is NOT supposed to put the hash into the config file manually.  We decided at the time to not break any APIs and apply this approach.

Upon further discussion, we chose this to be a suboptimal way of doing things and bad UX. The flag should be passed as an argument, even if APIs are being broken. 

Once a PR is open, we should however check whether we are causing major breaking changes for some users and let them know that this update is coming.

jmalicevic:
Closes #3547 .

This PR removes the `GenesisHash` from the config file because this field is not meant to be updated by the operator. It was only introduced to avoid breaking an API and used to pass the genesis hash provided via cli to the node constructor. 

As this feature is going into an unreleased version, we decided to break the API of `DefaultNewNode`. We maintain the signature of `NewNode` as this is probably used by the community more and introduce `NewNodeWithCliParams`.

`NewNodeWithCliParams` allows for passing different parameters from the cli to the node. 

The PR also introduces a new type `CliParams`. At the moment this struct contains just the `GenesisHash` but it was done this way so that we can in the future extend it with new parameters without breaking APIs.


melekes (approved):
👍

.changelog/unreleased/breaking-changes/3595-node-remove-genesishash.md
```diff
@@ -0,0 +1,3 @@
+- `[node]` Go API breaking change to `DefaultNewNode`. The function passes 
+`NonPersistentCliParams` to a node now.
```
cason, line 2:
```suggestion
`CliParams` to a node now.
```

.changelog/unreleased/improvements/3595-config-remove-genesishash.md
```diff
@@ -0,0 +1,2 @@
+- `[config]` Remove unused `GenesisHash` flag
+  ([\#3595](https://github.com/cometbft/cometbft/pull/3595))
```
cason, line 2:
```suggestion
  ([\#3547](https://github.com/cometbft/cometbft/issues/3547)
```

config/config.go
cason, line 1: Please update the documentation at `docs/references/config/config.toml.md` by removing this field from there as well.

node/setup.go
```diff
@@ -74,25 +87,26 @@ func DefaultGenesisDocProviderFunc(config *cfg.Config) GenesisDocProvider {
 }
 
 // Provider takes a config and a logger and returns a ready to go Node.
-type Provider func(*cfg.Config, log.Logger) (*Node, error)
+type Provider func(*cfg.Config, log.Logger, CliParams) (*Node, error)
 
 // DefaultNewNode returns a CometBFT node with default settings for the
 // PrivValidator, ClientCreator, GenesisDoc, and DBProvider.
 // It implements NodeProvider.
```
cason, line 94:
```suggestion
// It implements Provider.
```

node/setup.go
```diff
@@ -74,25 +87,26 @@ func DefaultGenesisDocProviderFunc(config *cfg.Config) GenesisDocProvider {
 }
 
 // Provider takes a config and a logger and returns a ready to go Node.
-type Provider func(*cfg.Config, log.Logger) (*Node, error)
+type Provider func(*cfg.Config, log.Logger, CliParams) (*Node, error)
```
cason, line 90: Changing this type is also a breaking change.

cason (commented):
This PR has two purposes:

1. Receive arbitrary CLI parameters when creating a new node
2. Remove the `genesis_hash` field from the config file, replacing it by a CLI parameter

I am almost inclined for having two PRs, but this is probably an overkill.

node/setup.go
jmalicevic: This was there from before but we can piggyback the change here as its minor.

jmalicevic:
> This PR has two purposes:
> 
> 1. Receive arbitrary CLI parameters when creating a new node
> 2. Remove the `genesis_hash` field from the config file, replacing it by a CLI parameter
> 
> I am almost inclined for having two PRs, but this is probably an overkill.

This is a good point and you are right. But it is probably overkill to do it for this change. We could however introduce a changelog entry under features so that it is clear this is there?

jmalicevic:
> > This PR has two purposes:
> > 
> > 1. Receive arbitrary CLI parameters when creating a new node
> > 2. Remove the `genesis_hash` field from the config file, replacing it by a CLI parameter
> > 
> > I am almost inclined for having two PRs, but this is probably an overkill.
> 
> This is a good point and you are right. But it is probably overkill to do it for this change. We could however introduce a changelog entry under features so that it is clear this is there?

Actually no, this is not a feature visible to the users, it is internal for us as developers or people who fork the repo perhaps so I not sure it falls under the classic feature category.

.changelog/unreleased/improvements/3595-config-remove-genesishash.md
jmalicevic: I will leave the PR number as it was also in our contributing guidelines.

node/setup.go
jmalicevic: Done

config/config.go
jmalicevic: Done!

cason (approved):
Thank you for that.

Lets see how our users will react to this change. Actually, I am not really convinced that we need to provide some many versions of `NewNode`...

node/node.go
```diff
@@ -286,13 +287,46 @@ func NewNode(ctx context.Context,
...
+func NewNodeWithCliParams(ctx context.Context,
+	config *cfg.Config,
+	privValidator types.PrivValidator,
+	nodeKey *p2p.NodeKey,
+	clientCreator proxy.ClientCreator,
+	genesisDocProvider GenesisDocProvider,
+	dbProvider cfg.DBProvider,
+	metricsProvider MetricsProvider,
+	logger log.Logger,
+	cliParams CliParams,
```
melekes, line 317: Can't `CliParams` be an option? Then, we won't need a separate `NewNodeWithCliParams` call and can use `NewNode` as usual.

PR #3603, opened 2024-08-02
chore(crypto/bls12381): remove cosmos/crypto wrapper for BLS curve

melekes:
I don't think it adds much value; hence, we can use the [underlying library](https://github.com/supranational/blst/blob/master/bindings/go) directly. This PR is the preparation for the BLS signature aggregation.

* remove `cosmos/crypto` wrapper for BLS curve

---

#### PR checklist

- [x] Tests written/updated


cason:
This failure on `TestMaxCommitBytes` is not a random stuff, right? Probably the size of the sigs on the block header have changed.

cason:
Isn't this breaking right now? We are changing the size of the block headers, aren't we?

melekes:
> Isn't this breaking right now? We are changing the size of the block headers, aren't we?

I will break this PR into 2.

crypto/bls12381/const.go
```diff
@@ -4,7 +4,7 @@ const (
 	// PrivKeySize defines the length of the PrivKey byte array.
 	PrivKeySize = 32
 	// PubKeySize defines the length of the PubKey byte array.
-	PubKeySize = 48
```
melekes, line 7: It was not 48!

crypto/bls12381/key_bls12381.go
```diff
@@ -18,6 +20,21 @@ const (
...
 )
 
+var (
+	// ErrDeserialization is returned when deserialization fails.
+	ErrDeserialization = errors.New("bls12381: deserialization error")
+
+	dstMinSig = []byte("BLS_SIG_BLS12381G1_XMD:SHA-256_SSWU_RO_NUL_")
+)
+
+// For minimal-pubkey-size operations.
```
melekes, line 30: We need to change this to minimal-signature-size operations because we care about signature size more!

cason:
Can you change the title to "remove cosmos/crypto wrapper for BLS curve"?

crypto/bls12381/key_bls12381.go
cason: Can you open an issue for that?

cason (approved):
Legit.

crypto/bls12381/key_bls12381.go
melekes: https://github.com/cometbft/cometbft/issues/3624

Issue #3624, opened 2024-08-05
crypto/bls12381: change to minimal-signature-size


melekes:
We must change BLS pubkey and signature types to minimal-signature-size operations because we care more about signature size! We don't send pubkeys over TCP, we send signatures.

https://github.com/cometbft/cometbft/pull/3603#discussion_r1704118134

dynst:
> We must change BLS pubkey and signature types to minimal-signature-size operations because we care more about signature size! We don't send pubkeys over TCP, we send signatures.

This got reverted from 48-byte [`minimal-signature-size`](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-05#section-2.1) back to `minimal-pubkey-size` in #3860 so that Berachain can implement the (EVM execution client compatible) EngineAPI using this type.

But does every other CometBFT chain have to be bound by the technical constraints Berachain is under? Can we reopen this to add a BLS key for chains that don't need 96 byte signatures for compatibility reasons?

PR #3627, opened 2024-08-05
feat(crypto/bls12381): change to minimal-signature-size

melekes:
Closes https://github.com/cometbft/cometbft/issues/3624

---

#### PR checklist

- [x] Tests written/updated


types/block.go
```diff
@@ -593,9 +593,14 @@ const (
 const (
 	// Max size of commit without any commitSigs -> 82 for BlockID, 8 for Height, 4 for Round.
 	MaxCommitOverheadBytes int64 = 94
-	// Commit sig size is made up of 96 bytes for the signature, 20 bytes for the address,
-	// 1 byte for the flag and 14 bytes for the timestamp.
-	MaxCommitSigBytes int64 = 131 + 10 // where's the 10 from?
+
+	// 4 bytes for field tags + 1 byte for signature LEN + 1 byte for
+	// validator address LEN + 1 byte for timestamp LEN.
+	maxCommitSigProtoEncOverhead = 4 + 1 + 1 + 1 + 3 // 3 ???
```
melekes, line 599: still don't know where 3 bytes are coming from

melekes:
```
message CommitSig {
  BlockIDFlag               block_id_flag     = 1;
  bytes                     validator_address = 2;
  google.protobuf.Timestamp timestamp         = 3
      [(gogoproto.nullable) = false, (gogoproto.stdtime) = true];
  bytes signature = 4;
}
```

cason: Protobuf overhead?

We really need a way to compute that properly.

types/signable.go
cason, line 1: There is a test `state/tx_filter_test.go` failing also because we are changing this header size... Gosh.

types/block.go
melekes: > Protobuf overhead?

Yes, but it's unclear from where. `varint`s of 3 fields shouldn't take more than 1 byte because they are all very small.

types/signable.go
melekes: Thanks! Fixed

types/signable.go
```diff
@@ -3,13 +3,17 @@ package types
...
+	"github.com/cometbft/cometbft/crypto/sr25519"
 	cmtmath "github.com/cometbft/cometbft/libs/math"
 )
 
 // MaxSignatureSize is a maximum allowed signature size for the Proposal
 // and Vote.
-// XXX: secp256k1 does not have Size nor MaxSize defined.
-var MaxSignatureSize = cmtmath.MaxInt(ed25519.SignatureSize, cmtmath.MaxInt(bls12381.SignatureLength, 64))
+// XXX: secp256k1 does not have max signature size defined.
+var MaxSignatureSize = cmtmath.MaxInt(
```
cason, line 13: The problem of this approach is that blocks produced using different signing algorithms will have different maximum payload size. We need to re-think this way of computing the maximum payload size, as it depends on the signing algorithm we are using...

cason (approved):
Legit.

But we need a better solution for defining the maximum payload size for a block, depending on the signing algorithm adopted.

PR #3634, opened 2024-08-06
feat(mempool): query app for `LanesInfo` and pass to mempool constructor

### LINKED ISSUE #3475: mempool/lanes: Query app for lane info
Implement part of the mempool initialisation to fetch lanes info as described in the [ADR](https://github.com/cometbft/cometbft/blob/main/docs/references/architecture/adr-118-mempool-lanes.md#initialization).

- Add new function to fetch lanes info from the app via `proxy.AppConnQuery`
  - It should validate the received info (see [here](https://github.com/cometbft/cometbft/blob/feature/mempool-qos/docs/references/architecture/adr-113-mempool-lanes.md#lanes-definition))
  - Possibly defined in the `mempool` package
- Call the above function to fetch the lanes info in `NewNode` and pass the response to `NewCListMempool`.
  - Currently the `Handshake` function in `internal/consensus/replay.go` [queries](https://github.com/cometbft/cometbft/blob/55493e04ef5977121435f25ae0cb01693b03aba3/internal/consensus/replay.go#L244) the app for info. We don't want to mix the mempool queries with consensus, so one option is to move `proxyApp.Query().Info(...)` outside of `Handshake`
  - Ideally we want to send an `Info` request to the app only once.

Requires #3502

jmalicevic:
Closes #3475 

---

#### PR checklist

- [x] Tests written/updated


mempool/lanes_info.go
```diff
@@ -13,6 +12,38 @@ type LanesInfo struct {
 	defaultLane types.Lane
 }
 
+type ErrEmptyLaneDefaultPrioSet struct {
```
hvanz: Move all these Err definitions to `mempool/errors.go`.

mempool/lanes_info.go
```diff
@@ -21,29 +52,41 @@ func FetchLanesInfo(lanePriorities []uint32, defLane types.Lane) (*LanesInfo, er
...
 	if err := info.validate(); err != nil {
-		return nil, fmt.Errorf("invalid lane info: %v, info: %v", err, info)
+		return nil, err
 	}
 
 	return &info, nil
 }
 
 func (info *LanesInfo) validate() error {
+	// If no lanes are provided the default priority is 0
```
hvanz, line 29: This is a valid scenario that was missing in the prototype.

mempool/clist_mempool_test.go
```diff
@@ -292,14 +292,33 @@ func TestMempoolUpdate(t *testing.T) {
 	}
 }
 
+func TestMempoolLanesValidation(t *testing.T) {
```
hvanz, line 295: I think a test scenario with empty `lanePriorities` and `defLane` = 0 is missing.

mempool/clist_mempool_test.go
```diff
@@ -292,14 +292,33 @@ func TestMempoolUpdate(t *testing.T) {
...
+	appInfo := abci.InfoResponse{}
+
+	_, err := FetchLanesInfo(appInfo.LanePriorities, types.Lane(appInfo.DefaultLanePriority))
+	require.NoError(t, err)
+
+	_, err = FetchLanesInfo(appInfo.LanePriorities, types.Lane(1))
```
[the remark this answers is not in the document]
jmalicevic: Should it not be invalid if `lanes` is empty? Because at this point, `appInfo.LanePriorities` has no items. I thought that in that case only a default priority of 0 was acceptable, am I wrong?

mempool/clist_mempool_test.go
jmalicevic: That should be the test at L298 because when you initialize `appInfo` the list is empty and the value for `DefaultLanePriority` is 0. 

Note that in line 296 `appInfo` is just an initialization of the `InfoResponse` struct with default values. It is not an actual call to the application.

mempool/clist_mempool_test.go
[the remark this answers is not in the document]
hvanz: The value 0 for lanes (including the default lane) is reserved for when we do CheckTx and there are no lanes, or when the tx is invalid, so the app returns 0 on the lane field of the CheckTx response. That's why the default lane is 1.

internal/consensus/replay.go
```diff
@@ -240,10 +240,17 @@ func (h *Handshaker) NBlocks() int {
 
 // TODO: retry the handshake/replay if it fails ?
 func (h *Handshaker) Handshake(ctx context.Context, proxyApp proxy.AppConns) error {
+	return h.HandshakeWithABCIRes(ctx, nil, proxyApp)
+}
+
+func (h *Handshaker) HandshakeWithABCIRes(ctx context.Context, res *abci.InfoResponse, proxyApp proxy.AppConns) error {
```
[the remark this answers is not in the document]
jmalicevic: I was trying to avoid breaking public go APIs in case someone is calling this outside the repo, but as this is a major version, we can also break it

internal/consensus/replay.go
```diff
@@ -240,10 +240,17 @@ func (h *Handshaker) NBlocks() int {
...
 	// Handshake is done via ABCI Info on the query conn.
-	res, err := proxyApp.Query().Info(ctx, proxy.InfoRequest)
-	if err != nil {
-		return fmt.Errorf("error calling Info: %v", err)
+	if res == nil {
+		var err error
+		res, err = proxyApp.Query().Info(ctx, proxy.InfoRequest)
```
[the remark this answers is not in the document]
jmalicevic: We need `proxyApp` below when calling `ReplayBlocks`. But I agree that perhaps there is no need to make the call again. We could put the extra call in `Handshake(` if we decide to keep it, in case people are using that function and then pass the retrieved arguments?

mempool/clist_mempool_test.go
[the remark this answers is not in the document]
jmalicevic: I see, I misunderstood that from the ADR. But still if the list is empty and the default priority is anything but 0 it should still fail, just maybe with the error that the lane is not in the list?

mempool/clist_mempool_test.go
jmalicevic: Do you think it is necessary to define a default line? What if the application sends you a list of lanes without 1?

hvanz: The ADR mentions the validation rule "the list lane_priorities is empty if and only if default_lane_priority is 0." but I think we can relax this rule a bit by allowing the app to have a non-empty list of lanes without a default lane. Then on a CheckTxResponse that doesn't specify a lane, the mempool will just assign the default lane `types.Lane(1)`, which must exists always on the mempool side (as defined in the ADR).

mempool/clist_mempool_test.go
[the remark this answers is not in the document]
hvanz: Ok, probably that's better.

mempool/clist_mempool_test.go
jmalicevic: I agree, but what if the non empty list of lanes does not contain our predefined default lane (1)? From what I understood the list of priorities does not have to be a contiguous array. 

So a user provides you a list of priorities :{2,4,6} and we set the default priority to {1}. This will then throw an error that the default priority is not in the priority list.

internal/consensus/replay.go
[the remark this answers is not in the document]
hvanz: Ah I didn't see that `proxyApp` is needed below. I would just keep one `Handshake` function, without retrying the call to keep the code as simple as possible. (Unless we need it for example for testing.)

mempool/clist_mempool_test.go
```diff
@@ -267,13 +292,33 @@ func TestMempoolUpdate(t *testing.T) {
 	}
 }
 
+func TestMempoolFetchLanesInfo(t *testing.T) {
+	appInfo := abci.InfoResponse{}
+
+	_, err := FetchLanesInfo([]uint32{}, types.Lane(0))
+	require.NoError(t, err)
+
+	_, err = FetchLanesInfo(appInfo.LanePriorities, types.Lane(1))
```
hvanz, line 301:
```suggestion
	_, err := FetchLanesInfo([]uint32{}, types.Lane(0))
	require.NoError(t, err)

	_, err = FetchLanesInfo([]uint32{}, types.Lane(1))
```

.changelog/unreleased/features/3634-query-app-for-lanes-info.md
```diff
@@ -0,0 +1,3 @@
+- `[mempool/node]` retrieve `LanesInfo` from application and pass on to mempool constructor. 
+Move the ABCI `Info` call from the `Handshake` function to the `NewNodeWithCliParams` function.
+  ([#3634](https://github.com/cometbft/cometbft/pull/3634))
```
sergio-mena, line 3: Question... rather than adding a log line for every incremental change we do on mempool lanes, wouldn't it be better to just add one log at the end (when we merge back to `main`) saying "mempool lanes implemented as per ADR 118".
To me, log lines are different from "git log" in that it is more coarse-grained?
Anyway, this is just a thought... WDYT?

internal/consensus/replay.go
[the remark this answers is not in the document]
sergio-mena: I agree with @hvanz in that we probably don't want to retry the call. Logic is clearer like that IMO

mempool/errors.go
```diff
@@ -83,3 +83,35 @@ func (e ErrFlushAppConn) Error() string {
 func (e ErrFlushAppConn) Unwrap() error {
 	return e.Err
 }
+
+type ErrEmptyLanesDefaultLaneSet struct {
+	Info LanesInfo
+}
+
+func (e ErrEmptyLanesDefaultLaneSet) Error() string {
+	return fmt.Sprintf("invalid lane info:if list of lanes is empty, then defaultLane should be 0, but %v given; info %v", e.Info.defaultLane, e.Info)
```
sergio-mena, line 92:
```suggestion
	return fmt.Sprintf("invalid lane info: if list of lanes is empty, then defaultLane should be 0, but %v given; info %v", e.Info.defaultLane, e.Info)
```

mempool/errors.go
```diff
@@ -83,3 +83,35 @@ func (e ErrFlushAppConn) Error() string {
...
+}
+
+type ErrBadDefaultLaneNonEmptyLaneList struct {
+	Info LanesInfo
+}
+
+func (e ErrBadDefaultLaneNonEmptyLaneList) Error() string {
+	return fmt.Sprintf("invalid lane info:default lane cannot be 0 if list of lanes is non empty; info: %v", e.Info)
```
sergio-mena, line 100:
```suggestion
	return fmt.Sprintf("invalid lane info: default lane cannot be 0 if list of lanes is non empty; info: %v", e.Info)
```

mempool/errors.go
```diff
@@ -83,3 +83,35 @@ func (e ErrFlushAppConn) Error() string {
...
+}
+
+type ErrDefaultLaneNotInList struct {
+	Info LanesInfo
+}
+
+func (e ErrDefaultLaneNotInList) Error() string {
+	return fmt.Sprintf("invalid lane info:list of lanes does not contain default lane; info %v", e.Info)
```
sergio-mena, line 108:
```suggestion
	return fmt.Sprintf("invalid lane info: list of lanes does not contain default lane; info %v", e.Info)
```

node/node.go
```diff
@@ -402,8 +402,13 @@ func NewNodeWithCliParams(ctx context.Context,
 	// Create the handshaker, which calls RequestInfo, sets the AppVersion on the state,
 	// and replays any blocks as necessary to sync CometBFT with the app.
 	consensusLogger := logger.With("module", "consensus")
+
+	appInfoResponse, err := proxyApp.Query().Info(ctx, proxy.InfoRequest)
+	if err != nil {
+		return nil, fmt.Errorf("error calling Info: %v", err)
```
sergio-mena, line 408:
```suggestion
		return nil, fmt.Errorf("error calling ABCI Info method: %v", err)
```

node/setup.go
```diff
@@ -264,10 +265,16 @@ func createMempoolAndMempoolReactor(
...
 	memplMetrics *mempl.Metrics,
 	logger log.Logger,
+	appInfoResponse *abci.InfoResponse,
 ) (mempl.Mempool, mempoolReactor) {
 	switch config.Mempool.Type {
 	// allow empty string for backward compatibility
 	case cfg.MempoolTypeFlood, "":
+		lanesInfo, err := mempl.FetchLanesInfo(appInfoResponse.LanePriorities, types.Lane(appInfoResponse.DefaultLanePriority))
+		if err != nil {
+			panic(fmt.Sprintf("Could not get lanes info from app: %s", err))
```
sergio-mena, line 275:
```suggestion
			panic(fmt.Sprintf("could not get lanes info from app: %s", err))
```

mempool/clist_mempool_test.go
```diff
@@ -267,13 +292,33 @@ func TestMempoolUpdate(t *testing.T) {
...
+}
+
 // Test dropping CheckTx requests when rechecking transactions. It mocks an asynchronous connection
 // to the app.
 func TestMempoolUpdateDoesNotPanicWhenApplicationMissedTx(t *testing.T) {
 	mockClient := new(abciclimocks.Client)
 	mockClient.On("Start").Return(nil)
 	mockClient.On("SetLogger", mock.Anything)
 	mockClient.On("Error").Return(nil).Times(4)
+	mockClient.On("Info", mock.Anything, mock.Anything).Return(&abci.InfoResponse{}, nil)
```
sergio-mena, line 321: Is there a reason why are we mocking "empty" here and "one single lane" in the calls below? (just trying to understand)

mempool/clist_mempool_test.go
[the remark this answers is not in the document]
sergio-mena: My opinion is that someone calling `FetchLanesInfo` with these parameters (empty list and non-zero default) doesn't really know what they are doing. So I'd rather warn them with an error (it might be a silly mistake that is better if we point it out), even if we could theoretically deal with that _weird_ input.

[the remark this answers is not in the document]
hvanz: I agree to make it fail with an empty list and non-zero default lane.

.changelog/unreleased/features/3634-query-app-for-lanes-info.md
hvanz: I agree with @sergio-mena on having just one log entry for Lanes, but this PR affects other parts of the code (node and consensus). In this case, I would leave a log entry like: 
> `[node]` Move ABCI `Info` call from the `Handshake` function in consensus module to `NewNodeWithCliParams` in node.

jmalicevic: I don't mind, but I think we should keep note at least in an external doc on the particular changes like this where we break or change APIs. These functions might get lost once we submit a big PR

mempool/clist_mempool_test.go
hvanz: Actually, all these tests pass even when mocking `Info` with an "empty" response. There's no need that `Info` returns an lane information because it's not needed for these tests. It will be different on the PR where we implement lanes, because the test will need lane information to create the lanes to be tested. So, in summary, in this PR we should mock all `Info` calls with an empty response.

.changelog/unreleased/features/3634-query-app-for-lanes-info.md
hvanz: Just a reminder to add a breaking-change changelog file for `Handshake` if its signature changes.

jmalicevic: @sergio-mena , @hvanz  to be clear, the consensus is we break the API of `Handshake`  ? If not , then I will move 
```
if res == nil {
		var err error
		res, err = proxyApp.Query().Info(ctx, proxy.InfoRequest)
```
into `Handshake` and pass res to ` HandshakeWithABCIRes` instead of `nil`  but remove the repeated query execution from `HandshakeWithABCIRes` as suggested below.

mempool/clist_mempool_test.go
jmalicevic: @hvanz , @sergio-mena I still think that having a non empty list of priorities without explicitly setting a default priority (defaultPriority is 0), should be an error. If you feel strongly against that, I can remove this as an error case.

mempool/clist_mempool_test.go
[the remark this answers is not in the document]
jmalicevic: This is solved then with this test, resolving the conversation.

.changelog/unreleased/features/3634-query-app-for-lanes-info.md
jmalicevic: Yes, I understood that, I just want to make sure we are on the same page with regards to breaking the API. Regardless of that, the repeated call is going out of the new function.

jmalicevic:
@hvanz , @sergio-mena  I changed the signature of Handshake and added a changelog entry. Let me know if it is ok now,

internal/consensus/replay.go
```diff
@@ -238,12 +238,10 @@ func (h *Handshaker) NBlocks() int {
 	return h.nBlocks
 }
 
-// TODO: retry the handshake/replay if it fails ?
-func (h *Handshaker) Handshake(ctx context.Context, proxyApp proxy.AppConns) error {
+func (h *Handshaker) Handshake(ctx context.Context, res *abci.InfoResponse, proxyApp proxy.AppConns) error {
 	// Handshake is done via ABCI Info on the query conn.
```
hvanz, line 242: I think this comment does not apply anymore.

mempool/lanes_info.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/cometbft/cometbft/types"
+)
+
+type LanesInfo struct {
+	lanes       []types.Lane
+	defaultLane types.Lane
+}
+
+// Query app info to return the required information to initialize lanes.
```
hvanz, line 14:
```suggestion
// FetchLanesInfo builds the information required to initialise lanes given the data queried from the app.
```

mempool/lanes_info.go
```diff
@@ -0,0 +1,64 @@
...
+func FetchLanesInfo(laneList []uint32, defLane types.Lane) (*LanesInfo, error) {
```
hvanz, line 15: I would rename this function to `BuildLanesInfo` or something similar. In the prototype, this function was fetching (querying) the info from the app, but that's not the case anymore.

hvanz (approved):
Looks great! Thanks a lot @jmalicevic!

sergio-mena (approved):
🚀

PR #3635, opened 2024-08-06
feat(grpc): Add gRPC Node Service

### LINKED ISSUE #3633: - Provides information about the node providing the gRPC interface.

alesforz:
Closes #3633.

This PR implements the `NodeService` gRPC as defined #2693.

Requirements:
- [x] Added protobuf files for the `NodeService` endpoint
- [x] Implemented service's server code
  - [x] `GetStatus`
  - [x] `GetHealth`
- [x] Implemented client's code
  - [x] `GetStatus`
  - [x] `GetHealth`
- [ ] Added e2e tests
  - [ ] `GetStatus`
  - [ ] `GetHealth`
- [x] Updated the config

---

#### PR checklist

- [ ] Tests written/updated


rpc/grpc/server/services/nodeservice/service.go
```diff
@@ -0,0 +1,201 @@
...
+		return nil, status.Errorf(codes.Canceled, formatStr, err)
+	}
+
+	nodeInfo, err := s.collectNodeInfo()
+	if err != nil {
+		l.Error(err.Error(), "err", err)
+
+		clientErrMSg := "node's basic information unavailable"
+		return nil, status.Error(codes.Internal, clientErrMSg)
+	}
```
alesforz, line 60: Is there a standard we follow to return errors from APIs?
For example, here (and below as well) I log the detailed error, but return to the client a more generic error that masks what really happened internally. In this particular example, the error the function logs is something like `p2p.NodeInfo concrete type != p2p.DefaultNodeInfo: <weird_type_name_here>`

The assumption is that clients don't need to know or wouldn't understand the details of the error. Is it useful information for a client to know that the node's code is using an unexpected concrete type for the interface `p2p.NodeInfo`?

In short, I'm wondering if we make this sort of distinction on error returns, or if we simply return the "unfiltered" error to clients.

alesforz:
The list of option functions [here](https://github.com/cometbft/cometbft/blob/3b80085c35864f851b0c5034b4391aab2e059517/rpc/grpc/client/client.go#L69) doesn't have `WithBlockResultsService(...)`, yet [below](https://github.com/cometbft/cometbft/blob/3b80085c35864f851b0c5034b4391aab2e059517/rpc/grpc/client/client.go#L147) we check if the service is enabled (I suppose it will never be if that function isn't present and called).
Was `WithBlockResultsService(...)` not added on purpose or was it an oversight?

alesforz:
I'm adding e2e tests for the `GetStatus` endpoint.
The testnet uses [this](https://github.com/cometbft/cometbft/blob/6a88835fb252b198e100c596df3473f2a5a047a2/test/e2e/pkg/testnet.go#L114) `Node` type to emulate a CometBFT node.

However, the `GetStatus` [endpoint](https://github.com/cometbft/cometbft/blob/6a88835fb252b198e100c596df3473f2a5a047a2/rpc/grpc/server/services/nodeservice/service.go#L39) relies on the [`Environment`](https://github.com/cometbft/cometbft/blob/e46f3b62b5322cc5d81126ae4b33b1c15886d486/rpc/core/env.go#L72) struct--created when a node starts [here](https://github.com/cometbft/cometbft/blob/e46f3b62b5322cc5d81126ae4b33b1c15886d486/node/node.go#L732)--to access the information about the current status of the node.

Is there a way to get the same information provided by the `Environment` struct from a testnet `Node`?
I want to create an `expectedStatus` variable storing the testnet node info to compare it to whatever the endpoint returns after calling `gRPCClient.GetStatus()` in the test.

rpc/grpc/server/services/nodeservice/service.go
andynog: We should log an error message that the operator can send to implementers for troubleshooting. When it comes to clients, we should provide a custom error message that gives an idea of what caused the error, but doesn't reveal too many internal details.

andynog:
> Does the `GetHealth` endpoint return any information, or is it just a sort of ping? That is, if the endpoint replies to the client, it's all good. The [JSON-RPC endpoint](https://docs.cometbft.com/main/rpc/#/Info/health) seems to be just that; it doesn't return any information other than a status `200 OK`.

We should maintain consistent behavior with the health endpoint by returning an 'OK' response or something indicating that the endpoint is operational.

andynog:
> The list of option functions [here](https://github.com/cometbft/cometbft/blob/3b80085c35864f851b0c5034b4391aab2e059517/rpc/grpc/client/client.go#L69) doesn't have `WithBlockResultsService(...)`, yet [below](https://github.com/cometbft/cometbft/blob/3b80085c35864f851b0c5034b4391aab2e059517/rpc/grpc/client/client.go#L147) we check if the service is enabled (I suppose it will never be if that function isn't present and called). Was `WithBlockResultsService(...)` not added on purpose or was it an oversight?

Yes, good catch, looks like it should have been implemented [here](https://github.com/cometbft/cometbft/pull/1168/files#diff-87e9c0086d39b1f644a3bbc0e7a141264a92c607905f923ee24fd77e8b27ad28)

andynog:
> [added](https://github.com/cometbft/cometbft/blob/7026ab7a19b58c2cee1f65001b8b856ffed3488a/test/e2e/tests/grpc_test.go#L217)

that's strange, when you test in e2e, the nodes you run (depending on the manifest you use) run in docker containers and they should run the gRPC server implementation. The gRPC client should be able to query the `GetStatus` and retrieve a response. Not sure if you tried to test this outside e2e, but you could just run `cometbft` with the gRPC service enabled and test this endpoint `GetStatus` with a gRPC custom client (simple go program) just to ensure you can get a client and server to communicate.

alesforz:
> > [added](https://github.com/cometbft/cometbft/blob/7026ab7a19b58c2cee1f65001b8b856ffed3488a/test/e2e/tests/grpc_test.go#L217)
> 
> that's strange, when you test in e2e, the nodes you run (depending on the manifest you use) run in docker containers and they should run the gRPC server implementation. The gRPC client should be able to query the `GetStatus` and retrieve a response. Not sure if you tried to test this outside e2e, but you could just run `cometbft` with the gRPC service enabled and test this endpoint `GetStatus` with a gRPC custom client (simple go program) just to ensure you can get a client and server to communicate.

To be clear: the error I get is an error that one of the helper functions that the `GetStatus` I implemented [returns](https://github.com/cometbft/cometbft/blob/d7478bb08b7eabfc73a3412e76b4f62e7f62d705/rpc/grpc/server/services/nodeservice/service.go#L185). Therefore, the endpoint is working. 
It also indicates that all the information that I collect [here](https://github.com/cometbft/cometbft/blob/d7478bb08b7eabfc73a3412e76b4f62e7f62d705/rpc/grpc/server/services/nodeservice/service.go#L112) and [here](https://github.com/cometbft/cometbft/blob/d7478bb08b7eabfc73a3412e76b4f62e7f62d705/rpc/grpc/server/services/nodeservice/service.go#L149) is available _somewhere_ in the `e2e.Node` type, otherwise the gRPC call would have returned earlier with a different error. I just can't figure out where that info is.

alesforz:
> > Does the `GetHealth` endpoint return any information, or is it just a sort of ping? That is, if the endpoint replies to the client, it's all good. The [JSON-RPC endpoint](https://docs.cometbft.com/main/rpc/#/Info/health) seems to be just that; it doesn't return any information other than a status `200 OK`.
> 
> We should maintain consistent behavior with the health endpoint by returning an 'OK' response or something indicating that the endpoint is operational.

What about doing what ABCI does?
We have a protobuf message:
```
message GetHealthResponse {
  uint32 code = 1;
}
```
whose code will be 0 if everything is alright (meaning status `OK`).

andynog:
> > > Does the `GetHealth` endpoint return any information, or is it just a sort of ping? That is, if the endpoint replies to the client, it's all good. The [JSON-RPC endpoint](https://docs.cometbft.com/main/rpc/#/Info/health) seems to be just that; it doesn't return any information other than a status `200 OK`.
> > 
> > 
> > We should maintain consistent behavior with the health endpoint by returning an 'OK' response or something indicating that the endpoint is operational.
> 
> What about following ABCI's convention? We return a protobuf message:
> 
> ```proto
> message GetHealthResponse {
>   uint32 code = 1;
> }
> ```
> 
> whose code will be 0 if everything is alright (meaning status `OK`).

Using a `uint32` would be appropriate if we expect to return multiple codes, for example each code would represent an un-healthy state that we would need to create a convention/table and outline somewhere (assuming `0` is `healthy`). 

But I think this might overcomplicate the expected behavior which is trying to see if the service is simply `healthy` or `un-healthy`, in this case a `bool` might be more appropriate.

sergio-mena:
@alesforz should we reopen this?

alesforz:
> @alesforz should we reopen this? Do you need a hand to get it over the finish line?

Apologies for missing your earlier message @sergio-mena 🙏 

Due to a shift in priorities, I had to temporarily set this work aside, but I'm now resuming it. The discussions with @andynog, both here and offline, have raised some important questions that are valuable to tackle now. Addressing these will enable a more efficient implementation of the other gRPC endpoints as well. I'm currently drafting a document with comments, questions, and potential design decisions to align on a unified implementation strategy for the gRPC endpoints. This might involve minor refactoring of the existing gRPC endpoints.

I will reopen this PR once we have completed the above steps.

PR #3637, opened 2024-08-06
test(e2e): fix docker file ENV command to prevent LegacyKeyValueFormat warning

andynog:
when running:

`test/e2e/make docker` 

getting some warnings:

```
 2 warnings found (use docker --debug to expand):
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 15)
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 33)
 ```

this PR fixes the `Dockerfile` and prevent the warnings

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

PR #3638, opened 2024-08-06
feat(mempool): add support for lanes

### LINKED ISSUE #3484: mempool/lanes: Implement lanes in `CListMempool`
Main component of the Lanes feature. It includes splitting the CList struct into multiple lists, mempool iterators for tx dissemination, reaping, and rechecking (implementing the WRR  scheduling algorithm), and basic testing.

Requires #3475 

Complementary tasks:
- #3756
- #3762
- #3833
- #3847
- #3862
- #3966
- #3987
- #3996

jmalicevic:
Closes #3484 

- [X] To be rebased to `feature/mempoool-qos` once #3634  is merged. 

---

#### PR checklist

- [X] Tests written/updated


mempool/clist_mempool.go
```diff
@@ -648,41 +712,49 @@ func (mem *CListMempool) recheckTxs() {
...
-			Tx:   tx,
-			Type: abci.CHECK_TX_TYPE_RECHECK,
-		})
-		if err != nil {
-			panic(fmt.Errorf("(re-)CheckTx request for tx %s failed: %w", log.NewLazySprintf("%v", tx.Hash()), err))
+		// Give some time to finish processing the responses; then finish the rechecking process, even
+		// if not all txs were rechecked.
+		select {
+		case <-time.After(mem.config.RecheckTimeout):
+			mem.recheck.setDone()
```
jmalicevic, line 745: This resets `isRechecked` and sets it to false. It means that for a brief moment, until the next loop iteration begins, the mempool thinks rechecking is done.

mempool/clist_mempool.go
```diff
@@ -193,9 +230,10 @@ func (mem *CListMempool) PreUpdate() {
 	}
 }
 
+// Size returns the total number of transactions in the mempool (that is, all lanes).
 // Safe for concurrent use by multiple goroutines.
 func (mem *CListMempool) Size() int {
-	return mem.txs.Len()
+	return int(mem.numTxs.Load())
```
sergio-mena, line 236:
```suggestion
	numTxs := mem.numTxs.Load()
	if numTxs > math.MaxInt {
		panic(fmt.Sprintf("mempool has too many transactions (corrupted?); %v > %v ", numTxs, math.MaxInt))
	}
	return int(numTxs)
```
Can we add this for type-cast safety?

mempool/clist_mempool.go
```diff
@@ -350,13 +369,19 @@ func (mem *CListMempool) handleCheckTxResponse(tx types.Tx, sender p2p.ID) func(
...
+		}
+
 		// Add tx to mempool and notify that new txs are available.
 		memTx := mempoolTx{
 			height:    mem.height.Load(),
 			gasWanted: res.GasWanted,
 			tx:        tx,
 		}
-		if mem.addTx(&memTx, sender) {
+		if mem.addTx(&memTx, sender, lane) {
```
sergio-mena, line 384: So elegant...

mempool/clist_mempool.go
```diff
@@ -381,29 +406,46 @@ func (mem *CListMempool) addTx(memTx *mempoolTx, sender p2p.ID) bool {
...
 			"height", mem.height.Load(),
 			"total", mem.Size(),
 		)
 		return false
 	}
 
+	// Get lane's clist.
+	txs, ok := mem.lanes[lane]
+	if !ok {
+		mem.logger.Error("Lane does not exist", "tx", log.NewLazySprintf("%v", tx.Hash()), "lane", lane)
```
sergio-mena, line 426:
```suggestion
		mem.logger.Error("Lane does not exist, not adding TX", "tx", log.NewLazySprintf("%v", tx.Hash()), "lane", lane)
```

mempool/clist_mempool.go
```diff
@@ -381,29 +406,46 @@ func (mem *CListMempool) addTx(memTx *mempoolTx, sender p2p.ID) bool {
...
+	// Update metrics.
 	mem.metrics.TxSizeBytes.Observe(float64(len(tx)))
 
 	mem.logger.Debug(
-		"added valid transaction",
+		"Added transaction",
 		"tx", tx.Hash(),
+		"lane", lane,
 		"height", mem.height.Load(),
 		"total", mem.Size(),
```
sergio-mena, line 450: Do we want to also add the size of the lane to the Debug line?

mempool/clist_mempool.go
```diff
@@ -420,12 +462,30 @@ func (mem *CListMempool) RemoveTxByKey(txKey types.TxKey) error {
...
 	mem.txsBytes.Add(int64(-len(tx)))
-	mem.logger.Debug("removed transaction", "tx", tx.Hash(), "height", mem.height.Load(), "total", mem.Size())
+	mem.numTxs.Add(int64(-1))
+
+	mem.logger.Debug(
+		"Removed transaction",
+		"tx", tx.Hash(),
+		"lane", lane,
+		"height", mem.height.Load(),
+		"total", mem.Size(),
```
sergio-mena, line 487: Do we want to also add the size of the lane to the Debug line?

mempool/clist_mempool.go
```diff
@@ -648,41 +712,49 @@ func (mem *CListMempool) recheckTxs() {
 		return
 	}
 
-	mem.recheck.init(mem.txs.Front(), mem.txs.Back())
+	// Recheck all transactions in each lane, sequentially.
+	// TODO: parallelize rechecking on lanes?
```
sergio-mena, line 716: I guess a PR/issue to handle this TODO is on its way. I don't think we should leave this in this state, even for an MVP

mempool/clist_mempool.go
```diff
@@ -700,20 +772,18 @@ type recheck struct {
...
-	}
-}
-
 func (rc *recheck) init(first, last *clist.CElement) {
 	if !rc.done() {
 		panic("Having more than one rechecking process at a time is not possible.")
 	}
 	rc.cursor = first
 	rc.end = last
+	rc.doneCh = make(chan struct{})
```
sergio-mena, line 781: `doneCh` had previously capacity 1. Is this change intended?

mempool/clist_mempool.go
```diff
@@ -700,20 +772,18 @@ type recheck struct {
...
 	rc.numPendingTxs.Store(0)
 	rc.isRechecking.Store(true)
+	rc.recheckFull.Store(false)
```
sergio-mena, line 784: Why is this added here?

mempool/clist_mempool.go
```diff
@@ -729,39 +799,30 @@ func (rc *recheck) setDone() {
...
 		// Reached end of the list without finding a matching tx.
 		rc.setDone()
 	}
 	if rc.done() {
-		// Notify that recheck has finished.
-		select {
-		case rc.doneCh <- struct{}{}:
-		default:
-		}
+		close(rc.doneCh) // notify channel that recheck has finished
```
sergio-mena, line 810: Are these changes because of Lanes, or are they also applicable to `v1.x`?

sergio-mena, line 810: Are these changes because of Lanes, or are they also applicable to `v1.x`?

mempool/clist_mempool.go
sergio-mena: I think I understand: is it because we're now closing the channel, rather than sending `struct{}{}`?
If that is the case, please check [this comment](https://github.com/cometbft/cometbft/pull/3638/files?diff=split&w=1#r1711826972).

mempool/clist_mempool.go
```diff
@@ -818,26 +882,73 @@ func (mem *CListMempool) NewIterator() Iterator {
...
+
+// Next implements the classical Weighted Round Robin (WRR) algorithm.
+//
+// In classical WRR, the iterator cycles over the lanes. When a lane is selected, Next returns an
+// entry from the selected lane. On subsequent calls, Next will return the next entries from the
+// same lane until `lane` entries are accessed or the lane is empty, where `lane` is the priority.
+// The next time, Next will select the successive lane with lower priority.
+//
+// TODO: Note that this code does not block waiting for an available entry on a CList or a CElement, as
+// was the case on the original code. Is this the best way to do it?
```
sergio-mena, line 927: We will need to profile to see if we are spinning too much (or add a Debug log to check if that log shows up too often)

mempool/clist_mempool.go
```diff
@@ -818,26 +882,73 @@ func (mem *CListMempool) NewIterator() Iterator {
...
+		lane := iter.mp.sortedLanes[iter.currentLaneIndex]
+		if iter.mp.lanes[lane].Len() == 0 || iter.counters[lane] >= uint32(lane) {
+			// Reset the counter only when the limit on the lane was reached.
+			if iter.counters[lane] >= uint32(lane) {
+				iter.counters[lane] = 0
+			}
+			iter.currentLaneIndex = (iter.currentLaneIndex + 1) % len(iter.mp.sortedLanes)
+			continue
+		}
+		// TODO: if we detect that a higher-priority lane now has entries, do we preempt access to the current lane?
```
sergio-mena, line 914: IMO, not in the MVP

mempool/clist_mempool.go
```diff
@@ -818,26 +882,73 @@ func (mem *CListMempool) NewIterator() Iterator {
...
+// priority value.
+func (iter *CListIterator) PickLane() types.Lane {
+	// Loop until finding a valid lane.
+	for {
+		// If the current lane is not valid, continue with the next lane with lower priority, in a
+		// round robin fashion.
+		lane := iter.mp.sortedLanes[iter.currentLaneIndex]
+		if iter.mp.lanes[lane].Len() == 0 || iter.counters[lane] >= uint32(lane) {
+			// Reset the counter only when the limit on the lane was reached.
+			if iter.counters[lane] >= uint32(lane) {
```
sergio-mena, line 908: Can we declare  the counters as `uint` (rather than `uint32`) so that this type cast is safer?

mempool/clist_mempool.go
```diff
@@ -818,26 +882,73 @@ func (mem *CListMempool) NewIterator() Iterator {
...
+// PickLane returns a _valid_ lane on which to iterate, according to the WRR algorithm. A lane is
+// valid if it is not empty or the number of accessed entries in the lane has not yet reached its
+// priority value.
+func (iter *CListIterator) PickLane() types.Lane {
+	// Loop until finding a valid lane.
+	for {
+		// If the current lane is not valid, continue with the next lane with lower priority, in a
+		// round robin fashion.
+		lane := iter.mp.sortedLanes[iter.currentLaneIndex]
+		if iter.mp.lanes[lane].Len() == 0 || iter.counters[lane] >= uint32(lane) {
```
sergio-mena, line 906: Suggestion to make this logic easier to follow:

* Separate this two conditions into two "if"s (one after the other... actually the check on Length last)
* include lines 911-912 below in both "if"s

Lines 911-912 appear now twice, but we keep the checks sequential and don't mix them up

mempool/clist_mempool.go
```diff
@@ -818,26 +882,73 @@ func (mem *CListMempool) NewIterator() Iterator {
...
 			close(ch)
 		}
 	}()
 	return ch
 }
+
+// PickLane returns a _valid_ lane on which to iterate, according to the WRR algorithm. A lane is
+// valid if it is not empty or the number of accessed entries in the lane has not yet reached its
+// priority value.
+func (iter *CListIterator) PickLane() types.Lane {
```
sergio-mena, line 900: Is there a task to thoroughly test this function (in particular, that it follows the WRR algorithm)

mempool/clist_mempool.go
sergio-mena: An update on this... we need to improve this logic for sure as, ATM, this is a busy-wait loop (so the channel is useless).

To validate this, I've run an `e2e` test (`simple.toml`) with the following patch (on top of this PR):
```diff
diff --git a/mempool/clist_mempool.go b/mempool/clist_mempool.go
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -900,6 +900,7 @@ func (iter *CListIterator) WaitNextCh() <-chan Entry {
 func (iter *CListIterator) PickLane() types.Lane {
 	// Loop until finding a valid lane.
 	for {
+		iter.mp.logger.Info("XXX PickLane, beginning of loop")
 		// If the current lane is not valid, continue with the next lane with lower priority, in a
 		// round robin fashion.
 		lane := iter.mp.sortedLanes[iter.currentLaneIndex]
diff --git a/mempool/reactor.go b/mempool/reactor.go
--- a/mempool/reactor.go
+++ b/mempool/reactor.go
@@ -240,10 +240,12 @@ func (memR *Reactor) broadcastTxRoutine(peer p2p.Peer) {
 			return
 		}
 
+		memR.Logger.Info("XXX before select")
 		select {
 		case entry = <-iter.WaitNextCh():
 			// If the entry we were looking at got garbage collected (removed), try again.
 			if entry == nil {
+				memR.Logger.Info("XXX entry is nil")
 				continue
 			}
 		case <-peer.Quit():
diff --git a/test/e2e/networks/simple.toml b/test/e2e/networks/simple.toml
--- a/test/e2e/networks/simple.toml
+++ b/test/e2e/networks/simple.toml
@@ -1,4 +1,2 @@
 [node.validator00]
 [node.validator01]
-[node.validator02]
-[node.validator03]
```

When I ran the testnet, I could see the follwing line being logged endlessly:

```log
validator01  | I[2024-08-09|19:57:12.532] XXX PickLane, beginning of loop              module=mempool
```

sergio-mena: I have an idea how we could do it. We can talk about it on Mon (to see if you agree on the approach)

mempool/clist_mempool.go
jmalicevic: I think this was done because we now wait for every lane then close a channel and open a new one. So it is only because of lanes, not 100% sure that it will stay like this atm.

mempool/clist_mempool.go
hvanz: I'd say yes.

mempool/clist_mempool.go
hvanz: It was easier to recheck lane by lane sequentially. Doing it in parallel will require some more work. I agree to not leave the TODO here.

mempool/clist_mempool.go
hvanz: We are calling recheck.init on multiple instances. It's just to be sure that all variables are initialized in the correct state. We should also clean up the variables when a recheck instance finishes.

mempool/clist_mempool.go
hvanz: No. It is definitively needed.

jmalicevic: I am working on it, I pushed some tests yesterday, working on cleaning them up. They caught errors that happened in e2e (where we keep returning the same transactions all the time).

mempool/clist_mempool.go
hvanz: I think this applies also to the `main` branch. I will create a separate PR with `main` as target.

hvanz:
@melekes suggested [here](https://github.com/cometbft/cometbft/pull/3700/files#r1716821642) to use the standard `list.List` instead of `clist.CList`. If we finally reimplement the mechanism to wait for available entries in `CListMempool` (and lock every access to the entries with the new mutex), then we can simplify the code by using `list.List`.

sergio-mena:
@hvanz oh yes! I support that! (but will need to measure performance, to make sure the lock won't become a contention point... which is why I think they went for clists at the first place)

mempool/clist_mempool.go
```diff
@@ -96,6 +132,7 @@ func NewCListMempool(
 		option(mp)
 	}
 
+	mp.logger.Info("CListMempool created", "defaultLane", mp.defaultLane, "lanes", mp.sortedLanes)
```
hvanz, line 135: There's no proper `logger` yet. This message won't show anywhere.

mempool/clist_mempool.go
hvanz: Fixed with #3862.

mempool/clist_mempool.go
hvanz: Done.

mempool/clist_mempool.go
```diff
@@ -537,14 +607,17 @@ func (mem *CListMempool) ReapMaxBytesMaxGas(maxBytes, maxGas int64) types.Txs {
...
 	// size per tx, and set the initial capacity based off of that.
-	// txs := make([]types.Tx, 0, cmtmath.MinInt(mem.txs.Len(), max/mem.avgTxSize))
-	txs := make([]types.Tx, 0, mem.txs.Len())
-	for e := mem.txs.Front(); e != nil; e = e.Next() {
-		memTx := e.Value.(*mempoolTx)
-
-		txs = append(txs, memTx.tx)
+	// txs := make([]types.Tx, 0, cmtmath.MinInt(mem.Size(), max/mem.avgTxSize))
+	txs := make([]types.Tx, 0, mem.Size())
+	mem.reapIter.Reset(mem.lanes)
```
hvanz, line 612: There's a risk of race conditions here, as shown in the [failed test](https://github.com/cometbft/cometbft/actions/runs/10686803248/job/29622932068?pr=3638) `TestBroadcastTxSync`. We assumed that Reap methods are only called by BlockExecutor, but this is not true. In particular, `ReapMaxTxs` is called by the RPC endpoint `UnconfirmedTxs`. A solution is to just create a new iterator on every Reap call, instead of reusing only one.

hvanz: Fixed by #3996

abci/example/kvstore/kvstore.go
```diff
@@ -139,7 +139,7 @@ func (app *Application) Info(context.Context, *types.InfoRequest) (*types.InfoRe
 	}
 
 	var defaultLanePriority uint32
-	if len(app.lanes) > 0 {
```
sergio-mena, line 142: Why this change?

jmalicevic: I pushed for it because we explicitly pass `nil` as lanes in some places but I realize that it won't throw any error in that case, the condition will just be false. We can revert the change.

mempool/iterators.go
```diff
@@ -0,0 +1,208 @@
...
+// Non-blocking version of the WRR iterator to be used for reaping and
+// rechecking transactions.
+//
+// Lock must be held on the mempool when iterating: the mempool cannot be
+// modified while iterating.
+type NonBlockingWRRIterator struct {
+	WRRIterator
+}
+
+func NewWRRIterator(mem *CListMempool) *NonBlockingWRRIterator {
```
sergio-mena, line 33: Name looks confusing to me. Can we rename to `NewNonBlockingWRRIterator`? This way there's no room for ambiguities.

sergio-mena: Maybe to shorten all names, we could do:
* `NewNonBlockingIterator`
* `NewBlockingIterator`
(just a thought)

mempool/iterators.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+}
+
+// Next returns the next element according to the WRR algorithm.
+func (iter *NonBlockingWRRIterator) Next() Entry {
+	lane := iter.sortedLanes[iter.laneIndex]
+	numEmptyLanes := 0
+	for {
+		// Skip empty lane or if cursor is at end of lane.
+		if iter.cursors[lane] == nil {
```
sergio-mena, line 64: In `PickLane` (part of the blocking iterator) the condition analogous to this one is more complex, because the fact that the cursor is `nil` doesn't necessarily mean that we are at the end of the lane. Why is it different here?

mempool/iterators.go
```diff
@@ -0,0 +1,208 @@
...
+		panic(fmt.Errorf("Iterator picked a nil entry on lane %d", lane))
+	}
+	iter.cursors[lane] = iter.cursors[lane].Next()
+	iter.counters[lane]++
+	return elem.Value.(*mempoolTx)
+}
+
+// BlockingWRRIterator implements a blocking version of the WRR iterator,
+// meaning that when no transaction is available, it will wait until a new one
+// is added to the mempool.
```
sergio-mena, line 92:
```suggestion
// is added to the mempool.
// Unlike `NonBlockingWRRIterator`, this iterator is expected to work with an evolving mempool.
```

mempool/iterators.go
sergio-mena: Is it because we don't support changes in the mempool once the iterator has been created?

jmalicevic: I will try to rethink this again, but the blocking iterator deals with lanes that can be edited while iterating (via recheck etc), here the assumption was the lanes are iterated over when reaping or rechecking , thus locked.

jmalicevic: But now that you mention it, we should either write down somewhere that this is how it has to be used or change the condition to be the same as in the blocking one. Because this is now a public interface, it can be used in ways we can't foresee now. @hvanz

hvanz: Two reasons for this condition:
- This non-blocking iterator should be used with `updateMtx` locked, so that entries are not added or removed while iterating. There's a comment above the `NonBlockingWRRIterator` struct, maybe we should make it more clear.
- When the iterator is reset (reset is also called when the iterator is created), the cursors are put at the beginning of each lane. So if a lane is empty, its cursor will remain nil. The condition `iter.cursors[lane] == nil` works both for when the lane is empty, or when the cursor reaches the end of the list.

abci/example/kvstore/kvstore.go
sergio-mena: I think it's more "goish" to use `len` unless there's a good reason not to use it

mempool/iterators.go
sergio-mena: I get it. Let me add a suggestion to clarify this

mempool/iterators.go
```diff
@@ -0,0 +1,208 @@
...
+func (iter *WRRIterator) nextLane() types.Lane {
+	iter.laneIndex = (iter.laneIndex + 1) % len(iter.sortedLanes)
+	return iter.sortedLanes[iter.laneIndex]
+}
+
+// Non-blocking version of the WRR iterator to be used for reaping and
+// rechecking transactions.
+//
+// Lock must be held on the mempool when iterating: the mempool cannot be
+// modified while iterating.
```
sergio-mena, line 28:
```suggestion
// This iterator does not support changes on the underlying mempool once initialised (or `Reset`),
// therefore the lock must be held on the mempool when iterating.
```

mempool/iterators.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+	iter := &NonBlockingWRRIterator{
+		WRRIterator: baseIter,
+	}
+	iter.Reset(mem.lanes)
+	return iter
+}
+
+// Reset must be called before every use of the iterator.
+func (iter *NonBlockingWRRIterator) Reset(lanes map[types.Lane]*clist.CList) {
```
sergio-mena, line 47: I see that after fixing the possible race condition, this is only called from `NewWRRIterator` above, and ReCheck.

Would it make sense to make this method private, and have ReCheck call `NewWRRIterator` rather than `Reset`? This way we reduce the possibility of future misuse.

mempool/iterators.go
```diff
@@ -0,0 +1,208 @@
...
+	go func() {
+		// Add the next entry to the channel if not nil.
+		lane := iter.PickLane()
+		if entry := iter.Next(lane); entry != nil {
+			ch <- entry.Value.(Entry)
+			close(ch)
+		} else {
+			// Unblock the receiver (it will receive nil).
+			close(ch)
+		}
```
sergio-mena, line 126:
```suggestion
		}
		// Unblock the receiver (it may receive nil).
		close(ch)
```

mempool/clist_mempool.go
```diff
@@ -47,11 +48,21 @@ type CListMempool struct {
 	// Keeps track of the rechecking process.
 	recheck *recheck
 
-	// Data in `txs` and `txsMap` must to be kept in sync and updated atomically.
+	// Data in the following variables must to be kept in sync and updated atomically.
 	txsMtx   cmtsync.RWMutex
-	txs      *clist.CList                    // concurrent linked-list of valid txs
-	txsMap   map[types.TxKey]*clist.CElement // for quick access to txs
+	lanes    map[types.Lane]*clist.CList     // each lane is a linked-list of (valid) txs
```
sergio-mena, line 53: Do we still need `CList` (i.e., "concurrent" list)?
If the answer if no, we probably need to change it in a further PR (shouldn't block this one).

mempool/clist_mempool.go
```diff
@@ -47,11 +48,21 @@ type CListMempool struct {
...
+	txsMap   map[types.TxKey]*clist.CElement // for quick access to the mempool entry of a given tx
 	txsBytes int64                           // total size of mempool, in bytes
+	numTxs   int64                           // total number of txs in the mempool
+
+	addTxChMtx    cmtsync.RWMutex // Protects the fields below
```
sergio-mena, line 58: Do we really need two mutexes here? (this one and the one in line 52). It the fact of having two mutexes for performance reasons? If yes, are we sure there's no risk of deadlock when using them?

mempool/clist_mempool.go
```diff
@@ -47,11 +48,21 @@ type CListMempool struct {
...
+	addTxCh       chan struct{}   // Blocks until the next TX is added
+	addTxSeq      int64
```
sergio-mena, line 60:
```suggestion
	addTxSeq      int64 // helps detect is new TXs have been added to a given lane
```

mempool/clist_mempool.go
jmalicevic: @hvanz  has a draft PR that should remove them. However, we discussed to try and merge this first and then have a new PR to address this potential issue. Especially since that code is based of a slightly older version of this branch so it takes some time to redesign it atm.

mempool/clist_mempool.go
```diff
@@ -47,11 +48,21 @@ type CListMempool struct {
...
+	addTxLaneSeqs map[types.Lane]int64
```
sergio-mena, line 61:
```suggestion
	addTxLaneSeqs map[types.Lane]int64  // Sequence of the last TX added to a given lane
```

mempool/clist_mempool.go
```diff
@@ -572,13 +642,17 @@ func (mem *CListMempool) ReapMaxTxs(max int) types.Txs {
...
+		max = mem.Size()
 	}
 
-	txs := make([]types.Tx, 0, cmtmath.MinInt(mem.txs.Len(), max))
-	for e := mem.txs.Front(); e != nil && len(txs) <= max; e = e.Next() {
-		memTx := e.Value.(*mempoolTx)
-		txs = append(txs, memTx.tx)
+	txs := make([]types.Tx, 0, cmtmath.MinInt(mem.Size(), max))
+	iter := NewWRRIterator(mem)
+	for len(txs) <= max {
```
sergio-mena, line 650: I know this isn't introduced in this PR, but this doesn't respect the `max` limit, it may return one more transaction than those that fit in `max`

melekes (commented):
nice work 👍

mempool/clist_mempool.go
```diff
@@ -667,23 +741,28 @@ func (mem *CListMempool) recheckTxs() {
...
 
 		// Send CheckTx request to the app to re-validate transaction.
 		resReq, err := mem.proxyAppConn.CheckTxAsync(context.TODO(), &abci.CheckTxRequest{
-			Tx:   tx,
+			Tx:   memTx.Tx(),
 			Type: abci.CHECK_TX_TYPE_RECHECK,
 		})
 		if err != nil {
-			panic(fmt.Errorf("(re-)CheckTx request for tx %s failed: %w", log.NewLazySprintf("%X", tx.Hash()), err))
+			panic(fmt.Errorf("(re-)CheckTx request for tx %s failed: %w", log.NewLazySprintf("%v", memTx.Tx().Hash()), err))
```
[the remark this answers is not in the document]
hvanz: This was fixed in #3837 in `main`. We need to merge `main` to the feature branch.

mempool/clist_mempool.go
hvanz: yeah, I would like to replace `clist.CList` by `list.List` in a future PR. There's still one thing that we need from CList, which is that `Len()` is guarded by an internal mutex in CList but not in List, so that creates race conditions.

abci/example/kvstore/kvstore.go
jmalicevic: yes, my bad, lots of c/c++ ptsd

mempool/iterators.go
hvanz: In #4013 I use the names `BlockingIterator` and `NonBlockingIterator` for two new interfaces, and I keep `BlockingWRRIterator` and `NonBlockingWRRIterator` for the implementations.

mempool/iterators.go
hvanz: I implemented this change you propose in #4013. I call `NewNonBlockingIterator` rather than `Reset` when rechecking, and then remove `Reset`.

mempool/clist_mempool.go
```diff
@@ -373,53 +395,79 @@ func (mem *CListMempool) handleCheckTxResponse(tx types.Tx, sender p2p.ID) func(
...
 	_ = memTx.addSender(sender)
-	e := mem.txs.PushBack(memTx)
+	e := txs.PushBack(memTx)
+
+	// Update auxiliary variables.
 	mem.txsMap[tx.Key()] = e
 	mem.txsBytes += int64(len(tx))
+	mem.numTxs++
 
+	// Notify iterator there's a new transaction.
```
sergio-mena, line 455:
```suggestion
	// Notify iterators there's a new transaction.
```

mempool/clist_mempool.go
```diff
@@ -701,38 +774,45 @@ func (mem *CListMempool) recheckTxs() {
...
-		doneCh: make(chan struct{}, 1),
-	}
+func newRecheck(iter *NonBlockingIterator) *recheck {
+	r := recheck{}
+	r.iter = iter
+	return &r
 }
 
-func (rc *recheck) init(first, last *clist.CElement) {
+func (rc *recheck) init(mp *CListMempool) {
```
sergio-mena, line 802: Two minor comments:
* Shouldn't it be more _compact_ if it was `newRecheck` the one that received `mp`, rather than `init`?
  * At the end of the day, the recheck is designed to always work with the same mempool, isn't it?
* As a follow-up, if `newRecheck` is now receiving `mp`, then it doesn't need to get `iter`, it can create it just as `init` does

mempool/clist_mempool.go
```diff
@@ -701,38 +774,45 @@ func (mem *CListMempool) recheckTxs() {
...
-	rc.end = last
 	rc.numPendingTxs.Store(0)
+	rc.iter = NewNonBlockingIterator(mp)
+
+	rc.cursor = rc.iter.Next()
+	if rc.cursor == nil {
+		return
+	}
+	rc.doneCh = make(chan struct{})
 	rc.isRechecking.Store(true)
```
sergio-mena, line 814: But then, if the mempool is empty

* we won't setup the channel... but later in the code we receive from than channel, will that work?
* about `isRechecking` I'm less sure, but just in case, shouldn't we make sure we set it to false here?

mempool/clist_mempool.go
jmalicevic: I added the parameter for `newRecheck` but I do not see a way around moving the argument from `init` without having a `mp` field within the `recheck`struct?

mempool/clist_mempool.go
jmalicevic: `isDone` checks whether `isRechecking` is false. 
About the channel you are right, we might want to not return here early because even with setting the other stuff, if the mempool is empty it will be set downstream.

mempool/clist_mempool.go
sergio-mena: Yes, that's what I was thinking... rather than providing the mempool at every call, just provide it a creation time (`newRecheck`) and then reuse for any further call. For that, you need to keep a reference to the mempool in the struct itself. Do you see a problem with that?

mempool/clist_mempool.go
sergio-mena: Yes, for the channel, we are calling `doneRechecking()` inside `recheckTxs`, whether the mempool is empty or not. So, if the mempool is empty, we probably need to create it, and then close it right away. Otherwise, we are using the channel from the previous "incarnation" of rechecktx. Maybe we can trace the code and convince ourselves that the channel exists and is closed anyway, but what about the "first incarnation"?

sergio-mena: Hmm, actually we don't close the channel if reCheckTx times out 🤦

sergio-mena: Let me propose something... bear with me

sergio-mena: How about this?
```
diff --git a/mempool/clist_mempool.go b/mempool/clist_mempool.go
index 7b97ce52f1..64b551d6a4 100644
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -807,10 +807,11 @@ func (rc *recheck) init(mp *CListMempool) {
 	rc.iter = NewNonBlockingIterator(mp)
 
 	rc.cursor = rc.iter.Next()
+	rc.doneCh = make(chan struct{})
 	if rc.cursor == nil {
+		rc.setDone()
 		return
 	}
-	rc.doneCh = make(chan struct{})
 	rc.isRechecking.Store(true)
 	rc.recheckFull.Store(false)
 }
@@ -826,6 +827,7 @@ func (rc *recheck) setDone() {
 	rc.cursor = nil
 	rc.recheckFull.Store(false)
 	rc.isRechecking.Store(false)
+	close(rc.doneCh) // notify channel that recheck has finished
 }
 
 // findNextEntryMatching searches for the next transaction matching the given transaction, which
@@ -849,7 +851,6 @@ func (rc *recheck) findNextEntryMatching(tx *types.Tx) (found bool) {
 
 	if rc.cursor == nil { // reached end of list
 		rc.setDone()
-		close(rc.doneCh) // notify channel that recheck has finished
 	}
 	return found
 }
```

hvanz: I think this last proposal works well.

hvanz: Still, `recheck` has many atomic variables when actually they should all be guarded by one mutex. We haven't hit a race condition yet.

jmalicevic: Added the code Sergio wrote.

mempool/clist_mempool.go
jmalicevic: No just double checking

jmalicevic: Done @sergio-mena

sergio-mena: 🙏

mempool/clist_mempool.go
sergio-mena: @jmalicevic thanks!

@hvanz I agree. Maybe we can get rid of all the atomic vars, and protect them with one single mutex. And also, see if we can have one mutex instead of two. But all that should be done in a further PR

mempool/clist_mempool.go
```diff
@@ -701,38 +774,48 @@ func (mem *CListMempool) recheckTxs() {
...
-	doneCh        chan struct{}   // to signal that rechecking has finished successfully (for async app connections)
-	numPendingTxs atomic.Int32    // number of transactions still pending to recheck
-	isRechecking  atomic.Bool     // true iff the rechecking process has begun and is not yet finished
-	recheckFull   atomic.Bool     // whether rechecking TXs cannot be completed before a new block is decided
+	iter          *NonBlockingIterator
+	cursor        Entry         // next expected recheck response
+	doneCh        chan struct{} // to signal that rechecking has finished successfully (for async app connections)
+	numPendingTxs atomic.Int32  // number of transactions still pending to recheck
+	isRechecking  atomic.Bool   // true iff the rechecking process has begun and is not yet finished
+	recheckFull   atomic.Bool   // whether rechecking TXs cannot be completed before a new block is decided\
```
sergio-mena, line 793:
```suggestion
	recheckFull   atomic.Bool   // whether rechecking TXs cannot be completed before a new block is decided
```

PR #3644, opened 2024-08-07
fix(types): add and ABCIPubKeyType for sr25519 key

### LINKED ISSUE #3643: types: add and `ABCIPubKeyType` for `sr25519` key
This key type was introduced by https://github.com/tendermint/tendermint/pull/4190, several years ago.

But we don't have an associated key type in the consensus parameters:

https://github.com/cometbft/cometbft/blob/06669cc51a2b54edbd5f21c24a4504894b67a217/types/params.go#L28-L30

As as result, when fixing https://github.com/cometbft/cometbft/issues/3514 we start having problems on nightly e2e tests because this key type can be used but it is not registered. It can be used because it is declared here:

https://github.com/cometbft/cometbft/blob/06669cc51a2b54edbd5f21c24a4504894b67a217/test/e2e/generator/generate.go#L75

cason:
Closes #3643.

Unfortunately, we don't have an automatic way to include every crypto key added to here, or at least to check that all are included.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/types/params.go
+++ b/types/params.go
@@ -12,6 +12,7 @@ import (
 	"github.com/cometbft/cometbft/crypto/bls12381"
 	"github.com/cometbft/cometbft/crypto/ed25519"
 	"github.com/cometbft/cometbft/crypto/secp256k1"
+	"github.com/cometbft/cometbft/crypto/sr25519"
 	"github.com/cometbft/cometbft/crypto/tmhash"
 )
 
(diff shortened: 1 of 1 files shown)


andynog (changes requested):
I'm doing some local testing on e2e cherry-picking this change and some preliminary tests I still get the e2e errors. Let me test a bit more before merging this.

andynog (approved):
These changes looks good to me and I've tested it locally with e2e and they fix the error. Please see comment https://github.com/cometbft/cometbft/issues/3643#issuecomment-2274150781

PR #3646, opened 2024-08-08
feat(crypto)!: remove sr25519 curve

melekes:
looks like no one is actually using it

---

#### PR checklist

- [x] Tests written/updated


cason (approved):
Legit.

But I think we should deprecate this key type in `v1.x`.

melekes:
> But I think we should deprecate this key type in v1.x.

Why? To give a chance to migrate? If they are using it w/ 0.38 line, then they would be able to continue using it for at least 1 year. Not sure we want to support sr25519 (in v1) for 2 years.

cason:
> Not sure we want to support sr25519 (in v1) for 2 years.

Ok, you got a point.  But if some asks for it, we need to re-introduce.

PR #3647, opened 2024-08-08
fix(mempool): tx is skipped when sending to a lagging peer or sending fails

hvanz:
Fixes a bug introduced in #3459 where instead of retrying the same tx when the peer is lagging, it was skipping it and moving on to the next entry. The same used to happen when sending the transaction failed: instead of retrying to send, it skipped it and proceeded to the next tx. This PR also reverts #3430, which was not working as expected.

Note that the first commit introduces a new test `TestMempoolReactorSendLaggingPeer` that initially fails and then it's fixed: https://github.com/cometbft/cometbft/actions/runs/10300181146/job/28509056466?pr=3647 

---

#### PR checklist

- [X] Tests written/updated
- ~~[ ] Updated relevant documentation (`docs/` or `spec/`) and code comments~~


melekes (approved):
👍

PR #3654, opened 2024-08-09
feat(mempool): add error ErrRecheckFull

hvanz:
Complements #3314

This PR adds a new error with a more clear error message when checking if the mempool is full while rechecking is in progress after a new block was committed.

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

PR #3659, opened 2024-08-09
feat(mempool)!: Add new `Contains` method to `Mempool` interface

hvanz:
Required for #3658 

---

#### PR checklist

- [ ] Tests written/updated
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments

PR #3660, opened 2024-08-09
feat(mempool): Do not send removed txs

### LINKED ISSUE #3658: mempool: Transaction may be sent to peer even if it was removed
### Problem Definition

The loop in the routine that sends txs to a peer first picks from the mempool the next entry to send, it makes some checks on the tx and the peer, and lastly it sends the message with the transaction. In the meantime, the entry can be removed from the mempool by Update or during rechecking, so it does not make sense anymore to disseminate this transaction.

### Proposal

Just before sending the message with the transaction, check if the transactions is still in the mempool.

hvanz:
Fixes #3658 

Requires #3659 

---

#### PR checklist

- [x] Tests written/updated


melekes (approved):
👍

PR #3693, opened 2024-08-13
bug(grpc): implemented missing optional function for BlocksResultsService in gRPC client

### LINKED ISSUE #3692: fix(grpc): Add enabling functional options for `BlocksResultsService`
Reposting from [this](https://github.com/cometbft/cometbft/issues/3635#issuecomment-2276280362_) discussion.

### The Problem
The list of option functions [here](https://github.com/cometbft/cometbft/blob/3b80085c35864f851b0c5034b4391aab2e059517/rpc/grpc/client/client.go#L69) doesn't have `WithBlockResultsService(...)`, yet [below](https://github.com/cometbft/cometbft/blob/3b80085c35864f851b0c5034b4391aab2e059517/rpc/grpc/client/client.go#L147) we check if the service is enabled (I suppose it will never be if that function isn't present and called).

We need to implement `WithBlockResultsService(...)`.

alesforz:
Closes #3692.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/rpc/grpc/client/client.go
+++ b/rpc/grpc/client/client.go
@@ -94,6 +94,17 @@ func WithBlockServiceEnabled(enabled bool) Option {
 	}
 }
 
+// WithBlockResultsServiceEnabled allows control of whether or not to create a
+// client for interacting with the block results service of a CometBFT node.
+//
+// If disabled and the client attempts to access the block results service API, the
+// client will panic.
+func WithBlockResultsServiceEnabled(enabled bool) Option {
+	return func(b *clientBuilder) {
+		b.blockResultsServiceEnabled = enabled
+	}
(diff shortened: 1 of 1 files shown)


andynog (approved):
Looks good to me, thanks for fixing that.

one thing I've realized is that these `With[Service]Enabled` methods are not used anywhere, probably it would be a good idea to at least implement some unit tests for them, but that can be a future PR.

andynog:
We should also look into back porting this to the `v0.38.x-experimental` branch since the gRPC services were ported to that branch too. 

https://github.com/cometbft/cometbft/blob/v0.38.x-experimental/rpc/grpc/client/client.go

PR #3700, opened 2024-08-14
fix(mempool): Update variables in `CListMempool` atomically

### LINKED ISSUE #3694: Race condition when creating new mempool entry
## Bug Report

`CListMempool` has a few variables for storing the state of the mempool entries, such as `txs: CList`, `txsMap: sync.Map`, and `txsBytes`, which is used for checking if the mempool is full. All these functions should be updated atomically, but it's currently not the case.

A recent improvement to `broadcastTxRoutine` checks if the transaction chosen to be disseminated next is still in the mempool just before sending it (#3658). This change has surfaced a race condition in the function `addTx` that creates and adds new entries to the mempool, in particular [this code](https://github.com/cometbft/cometbft/blob/9466bdaf6dacbb473a9c9a5e9c49d628b5dd6d00/mempool/clist_mempool.go#L398-L400):
```
398:    e := mem.txs.PushBack(memTx)
399:    mem.txsMap.Store(txKey, e)
400:    mem.txsBytes.Add(int64(len(tx)))
```

This is the execution trace that makes some test to fail:
1. When adding a new tx, line 398 will effectively create a new entry in `txs`. 
2. Now the context switches to `broadcastTxRoutine`, which [picks the new entry](https://github.com/cometbft/cometbft/blob/9466bdaf6dacbb473a9c9a5e9c49d628b5dd6d00/mempool/reactor.go#L229) to send to a peer. 
3. Before sending the transaction, `broadcastTxRoutine` [checks if the entry is still in the mempool](https://github.com/cometbft/cometbft/blob/9466bdaf6dacbb473a9c9a5e9c49d628b5dd6d00/mempool/reactor.go#L277) using `txsMap`, but line 399 in `addTx` was not executed yet, so the transaction is not sent.

Note that the function `addTx` is guarded by `updateMtx` only when using a local ABCI client (via `handleCheckTxResponse` and `CheckTx`). But even this mutex does not prevent `broadcastTxRoutine` to access the mempool entries concurrently.

hvanz:
Fixes #3694 

This PR replaces:
```
type CListMempool struct {
	txsBytes atomic.Int64 // total size of mempool, in bytes
        ...
	txs    *clist.CList
	txsMap sync.Map
        ...
}
```
by
```
type CListMempool struct {
        ...
	txsMtx   cmtsync.RWMutex
	txs      *clist.CList                    // concurrent linked-list of valid txs
	txsMap   map[types.TxKey]*clist.CElement // for quick access to txs
	txsBytes int64                           // total size of mempool, in bytes
        ...
}
```
The new `txsMtx` mutex is used every time the variables `txs`, `txsMap`, and `txsBytes` are accessed.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -584,8 +584,8 @@ func (mem *CListMempool) ReapMaxTxs(max int) types.Txs {
 
 // GetTxByHash returns the types.Tx with the given hash if found in the mempool, otherwise returns nil.
 func (mem *CListMempool) GetTxByHash(hash []byte) types.Tx {
-	mem.txsMtx.Lock()
-	defer mem.txsMtx.Unlock()
+	mem.txsMtx.RLock()
+	defer mem.txsMtx.RUnlock()
 
 	if elem, ok := mem.txsMap[types.TxKey(hash)]; ok {
 		return elem.Value.(*mempoolTx).tx
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -145,7 +145,7 @@ func (mem *CListMempool) addSender(txKey types.TxKey, sender p2p.ID) error {
 	memTx := elem.Value.(*mempoolTx)
 	if found := memTx.addSender(sender); found {
 		// It should not be possible to receive twice a tx from the same sender.
-		return ErrTxAlreadySeen
+		return ErrTxAlreadyReceivedFromSender
 	}
 	return nil
 }
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -27,8 +27,7 @@ const noSender = p2p.ID("")
 // mempool uses a concurrent list structure for storing transactions that can
 // be efficiently accessed by multiple concurrent readers.
 type CListMempool struct {
-	height   atomic.Int64 // the last block Update()'d to
-	txsBytes atomic.Int64 // total size of mempool, in bytes
+	height atomic.Int64 // the last block Update()'d to
 
 	// notify listeners (ie. consensus) when txs are available
 	notifiedTxsAvailable atomic.Bool
@@ -49,9 +48,10 @@ type CListMempool struct {
 	recheck *recheck
 
 	// Data in `txs` and `txsMap` must to be kept in sync and updated atomically.
-	txsMtx cmtsync.RWMutex
-	txs    *clist.CList                    // concurrent linked-list of valid txs
-	txsMap map[types.TxKey]*clist.CElement // for quick access to txs
+	txsMtx   cmtsync.RWMutex
+	txs      *clist.CList                    // concurrent linked-list of valid txs
+	txsMap   map[types.TxKey]*clist.CElement // for quick access to txs
+	txsBytes int64                           // total size of mempool, in bytes
 
 	// Keep a cache of already-seen txs.
 	// This reduces the pressure on the proxyApp.
@@ -124,7 +124,7 @@ func (mem *CListMempool) removeAllTxs() {
 		e.DetachPrev()
 	}
 	mem.txsMap = make(map[types.TxKey]*clist.CElement)
-	mem.txsBytes.Store(0)
+	mem.txsBytes = 0
 }
 
 // addSender adds a peer ID to the list of senders on the entry corresponding to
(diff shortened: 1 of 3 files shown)


mempool/clist_mempool.go
```diff
@@ -569,7 +584,10 @@ func (mem *CListMempool) ReapMaxTxs(max int) types.Txs {
 
 // GetTxByHash returns the types.Tx with the given hash if found in the mempool, otherwise returns nil.
 func (mem *CListMempool) GetTxByHash(hash []byte) types.Tx {
-	if elem, ok := mem.getCElement(types.TxKey(hash)); ok {
+	mem.txsMtx.Lock()
```
melekes, line 587: RLock?

hvanz: yes! I changed it to RLock.

mempool/clist_mempool.go
```diff
@@ -49,11 +47,11 @@ type CListMempool struct {
...
 	recheck *recheck
 
-	// Concurrent linked-list of valid txs.
-	// `txsMap`: txKey -> CElement is for quick access to txs.
-	// Transactions in both `txs` and `txsMap` must to be kept in sync.
-	txs    *clist.CList
-	txsMap sync.Map
+	// Data in `txs` and `txsMap` must to be kept in sync and updated atomically.
+	txsMtx   cmtsync.RWMutex
+	txs      *clist.CList                    // concurrent linked-list of valid txs
```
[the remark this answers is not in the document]
melekes: Both from the comment and the actual implementation (clist becomes https://pkg.go.dev/container/list).

melekes (approved):
👍

mempool/clist_mempool.go
```diff
@@ -349,55 +362,55 @@ func (mem *CListMempool) handleCheckTxResponse(tx types.Tx, sender p2p.ID) func(
 			return
 		}
 
+		// Check that tx is not already in the mempool. This can happen when the
+		// cache overflows. See https://github.com/cometbft/cometbft/pull/890.
+		txKey := tx.Key()
+		if mem.Contains(txKey) {
+			if err := mem.addSender(txKey, sender); err != nil {
```
andynog, line 369: the `memTx.addSender` might return `false` when a `peerID == 0` but  not sure if this is possible and what would be the implications here since in the `mem.addSender` you are just looking for a `found` and returning an error in this case. Not sure if the `!found` would be also a error case and would be possible when `peerID == 0`. 

Probably outside the scope of this PR but just commenting here since I was understanding this logic a bit better.

andynog (approved):
lgtm, I believe the mutex locks/unlocks were properly added in all the places needed with this refactoring.

hvanz:
Some benchmark results.

On the branch of this PR (commit `6e7e88b1e`):
```
BenchmarkReap-8                      153           8161910 ns/op          485791 B/op      10001 allocs/op
BenchmarkCheckTx-8                 13608            100176 ns/op            2772 B/op         46 allocs/op
BenchmarkParallelCheckTx-8         10000            117761 ns/op            2791 B/op         51 allocs/op
BenchmarkUpdateRemoteClient-8       3664            569765 ns/op            9888 B/op        211 allocs/op
BenchmarkCheckDuplicateTx-8       337864              3503 ns/op               0 B/op          0 allocs/op
```

On the commit this PR is targeting (`9466bdaf6`):
```
BenchmarkReap-8                       145           8299426 ns/op          485795 B/op      10001 allocs/op
BenchmarkCheckTx-8                  12240            114495 ns/op            2679 B/op         50 allocs/op
BenchmarkParallelCheckTx-8          13131             81848 ns/op            2808 B/op         55 allocs/op
BenchmarkUpdateRemoteClient-8        3640            608122 ns/op           10261 B/op        219 allocs/op
BenchmarkCheckDuplicateTx-8        538094              2127 ns/op               0 B/op          0 allocs/op
```

mempool/clist_mempool.go
hvanz: `peerID == 0` handles the case when there's no sender, that is, the tx came via an RPC endpoint. It's not an error.

mempool/clist_mempool.go
```diff
@@ -212,14 +230,15 @@ func (mem *CListMempool) Flush() {
 	mem.updateMtx.RLock()
 	defer mem.updateMtx.RUnlock()
 
-	mem.txsBytes.Store(0)
```
jmalicevic, line 215: Is this replaced somewhere with a write of 0?

hvanz: Yes, in `removeAllTxs`, which is guarded by `txsMtx`.

PR #3711, opened 2024-08-15
fix(abci/tutorials): Refactored test to appease linter

jmalicevic:
This version of linter was failing for the test in the forum demo app. 

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/abci/tutorials/abci-v2-forum-app/test/find_user_test.go
+++ b/abci/tutorials/abci-v2-forum-app/test/find_user_test.go
@@ -46,16 +46,10 @@ func TestFindUserByName(t *testing.T) {
 		t.Fatalf("Failed to find user by name: %v", err1)
 	}
 
-	if foundUser1 == nil {
-		t.Fatalf("User not found")
-	}
-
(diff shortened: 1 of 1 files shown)


abci/tutorials/abci-v2-forum-app/test/find_user_test.go
```diff
@@ -46,16 +46,10 @@ func TestFindUserByName(t *testing.T) {
...
-		t.Fatalf("User not found")
-	}
-
-	if foundUser1.Name != "user2" {
-		// t.Fatalf("Expected user2, but got %s", foundUser.Name)
-		println("Expected user2, but got %s", foundUser1.Name)
-	}
-
-	if foundUser1.Name == "user2" {
+	if foundUser1 != nil {
```
jmalicevic, line 49: I know we prefer early returns but for some reason if I place the check for nil first and call fatalF it does not recognize it as termination, but still reports the 'Possible nil pointer dereference' error

melekes (approved):

PR #3719, opened 2024-08-16
test(mempool): Allow custom log levels on reactor tests

hvanz:
Also disable the P2P log messages on the switches, which are not relevant for mempool.

In particular, this PR changes the log level of `TestReactor_MaxTxBytes`, which is very verbose and it currently could take 12s with the default "debug" level. With "info" level it takes 300ms.

For reviewers, the PR can be reviewed commit by commit.


melekes (approved):
👍

Issue #3720, opened 2024-08-16
BLS still has some deterministic failures in the `e2e` nighlies


sergio-mena:
In order to prevent that acting as noise, let us:

- [x] Remove BLS from the possible key types in the generator
  -  #3721
- [x] Fix all deterministic failures offline
  - #3726
- [x] Re-introduce BLS as a key type in the generator
  - #3726

PR #3721, opened 2024-08-16
fix(e2e): remove bls from e2e generator

sergio-mena:
Part of #3720

With this change, the `e2e` generator no longer generates manifests with bls as key type, so bls is no longer tested in the nightlies.

This is temporary, to be reverted also as part of #3720 in a future PR

Manual nightly run: [here](https://github.com/cometbft/cometbft/actions/runs/10418554752/job/28854915428)

---

#### PR checklist

- ~[ ] Tests written/updated~


andynog (approved):
looks good to me

PR #3724, opened 2024-08-16
chore(mempool/lanes): merge feature/mempool-qos

hvanz:


mempool/clist_mempool.go
```diff
@@ -409,30 +432,12 @@ func (mem *CListMempool) handleCheckTxResponse(tx types.Tx, sender p2p.ID) func(
 // Called from:
 //   - handleCheckTxResponse (lock not held) if tx is valid
 func (mem *CListMempool) addTx(memTx *mempoolTx, sender p2p.ID, lane types.Lane) bool {
+	mem.txsMtx.Lock()
+	defer mem.txsMtx.Unlock()
+
 	tx := memTx.tx
 	txKey := tx.Key()
 
-	// Check if the transaction is already in the mempool.
```
jmalicevic, line 415: Are we checking this now somewhere else?

hvanz: Yes, it's being checked in `handleCheckTxResponse`, outside of `addTx`, because it doesn't need to be guarded by `txsMtx`. This change comes from `main`.

jmalicevic (approved):

PR #3725, opened 2024-08-16
fix(mempool): Fix mutex in `Flush` method

### LINKED ISSUE #2443: mempool: CListMempool.Flush invokes a read-lock meanwhile it performs write/mutating operations; inadequately tested because there will be data races too

hvanz:
Fixes #2443

Note that `Flush` is only called from an (unsafe) RPC endpoint, so it can be called at any time.

---

#### PR checklist

- [ ] Tests written/updated


andynog (approved):
Looks good to me

PR #3726, opened 2024-08-16
fix(e2e): fix BLS-related failures

sergio-mena:
Part of #3720

`e2e` tests _require_ a method that generates a private key deterministically (from a secret). Without such method, is it not possible for the "tests" step of the `e2e` runner to recreate the node structure as it exists within the runner itself.

Manual nightly runs:
- [1st try](https://github.com/cometbft/cometbft/actions/runs/10424046666)
- [2nd try](https://github.com/cometbft/cometbft/actions/runs/10424939531), after 53321d4

---

#### PR checklist

- ~[ ] Tests written/updated~

### DIFF
--- a/crypto/bls12381/key.go
+++ b/crypto/bls12381/key.go
@@ -30,6 +30,11 @@ var _ crypto.PrivKey = &PrivKey{}
 // PrivKey represents a BLS private key noop when blst is not set as a build flag and cgo is disabled.
 type PrivKey []byte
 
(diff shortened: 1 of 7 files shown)


andynog (approved):
changes look good to me, had to fix one unit test that was failing because of a change in the return of `PubKey()`, other than that lgtm

andynog:
I've reviewed and merged this. Had to fix a couple of tests in the crypto and encoding. @sergio-mena and @melekes please take a look at my commits and see if they made sense.

PR #3728, opened 2024-08-18
build(crypto/secp256k1): Update `btcec/v2` to v2.3.4

sigv:
Running a `go get` should consistently build this project as per the README. However, the latest patch release of `btcec/v2` is not a proper Semantic Versioning patch. It removes an error return value from `ecdsa.SignCompact` function (btcsuite/btcd@e5d15fd). This is functionally a no-op, as the underlying function was hardcoded to always return `nil` as error. However, this is still a breaking change (compiler/build error) for all consumers of this function, as they historically expected two return values, but now only get one.

Ref: btcsuite/btcd#2211
Ref: #3531
Ref: #3536

---

#### PR checklist

- [ ] Tests written/updated


sigv:
As described in [Contributing > Dependencies](https://github.com/cometbft/cometbft/blob/905355f7577ef1705a9dfb7c3ab12375d2822123/CONTRIBUTING.md#dependencies):

> We use [Go modules](https://github.com/golang/go/wiki/Modules) to manage dependencies.
>
> That said, the `main` branch of every CometBFT repository should just build with `go get`, which means they should be kept up-to-date with their dependencies so we can get away with telling people they can just `go get` our software.

This PR proposes the minimal change scope for ensuring latest `github.com/btcsuite/btcd/btcec/v2` can be used in such manner. The return signature of `secp256k1.Sign` is retained as `([]byte, error)`, so that backporting is clear.

Further clean-up (removing the `error` from return values) would impact the project widely, and should be considered a breaking refactor. It should be done in a different PR (yet to be opened) and would presumably not get backported due to the risks of conflicts.

crypto/secp256k1/secp256k1.go
```diff
@@ -118,10 +118,7 @@ func (privKey PrivKey) Sign(msg []byte) ([]byte, error) {
...
 
 	sum := sha256.Sum256(msg)
-	sig, err := ecdsa.SignCompact(priv, sum[:], false)
-	if err != nil {
-		return nil, err
-	}
+	sig := ecdsa.SignCompact(priv, sum[:], false)
 
 	// remove the first byte which is compactSigRecoveryCode
 	return sig[1:], nil
```
melekes, line 124: This will panic if `sig` is `nil`. Should we check for `nil` value?

sigv: The call to [`ecdsa.SignCompact`](https://github.com/btcsuite/btcd/blob/btcec/v2.3.4/btcec/ecdsa/signature.go#L238) is then handed over to `github.com/decred/dcrd/dcrec/secp256k1/v4` module's [`ecdsa.SignCompact`](https://github.com/decred/dcrd/blob/dcrec/secp256k1/v4.0.1/dcrec/secp256k1/ecdsa/signature.go#L712). The function there always seems to return a value. I am not seeing where a `nil` could come from.

The change that the `btcec/v2` project made (btcsuite/btcd@e5d15fddb9c486dfe382bc95dc38d2fd247813bf) was replace:
```go
return secp_ecdsa.SignCompact(key, hash, isCompressedKey), nil
```
with
```go
return secp_ecdsa.SignCompact(key, hash, isCompressedKey)
```

Previously `nil` was not being checked in CometBFT, so I think there is no need to check for `nil` now. 🤔

sigv:
Pushed a new commit as df846546ac2f751b793c26e80bb872729926d678 updated `btcutil` dependency on the next line (causing merge conflict).

melekes (approved):
Thanks @sigv ❤️

melekes:
@mergify backport v0.38.x

melekes:
@mergify backport v0.37.x

sigv:
@melekes, do you think it makes sense to remove `error` from return signature of [`crypto.PrivKey.Sign(msg []byte)`](https://github.com/cometbft/cometbft/blob/17abb80063cd7d33baa5590bb679a66f19d1de69/crypto/crypto.go#L31) function? All the implementations are now hard-coded to return `nil` as `error` with exception [`bls12381`](https://github.com/cometbft/cometbft/blob/17abb80063cd7d33baa5590bb679a66f19d1de69/crypto/bls12381/key.go#L65) which outright `panic`s. It would be a breaking change, but I am not seeing a scenario where an `error` is reasonable to return.

Should I open a PR?

PR #3756, opened 2024-08-19
feat(mempool/lanes): Store lane in `mempoolTx` and remove `txsLanes`

hvanz:
We don't need to store the lane in a separate map. We can just extend `mempoolTx` with a `lane` field.

---

#### PR checklist

- [ ] Tests written/updated


jmalicevic (approved):
LGTM.

PR #3762, opened 2024-08-20
feat(lanes): Change lane selection algorithm in kvstore app

hvanz:
Depends on #3833

Currently it computes a hash to assign a lane, which makes it impossible to easily reproduce in tests. With this change it will use the key in a "key=value" transaction.

### DIFF
--- a/abci/example/kvstore/kvstore.go
+++ b/abci/example/kvstore/kvstore.go
@@ -195,6 +195,7 @@ func (app *Application) assignLane(tx []byte) uint32 {
 		return app.lanes[defaultLane]
 	}
 
+// If the tx is of the form 2=2 we will assign it a lane. For any other type of transaction sent to the kvstore, it will go to the default lane.  
 	keyInt, err := strconv.Atoi(key)
 	if err != nil {
 		return app.lanes[defaultLane]
--- a/abci/example/kvstore/kvstore.go
+++ b/abci/example/kvstore/kvstore.go
@@ -195,7 +195,9 @@ func (app *Application) assignLane(tx []byte) uint32 {
 		return app.lanes[defaultLane]
 	}
 
-// If the tx is of the form 2=2 we will assign it a lane. For any other type of transaction sent to the kvstore, it will go to the default lane.  
+	// If the transaction key is an integer (for example, a transaction of the
+	// form 2=2), we will assign a lane. Any other type of transaction will go
+	// to the default lane.
 	keyInt, err := strconv.Atoi(key)
 	if err != nil {
 		return app.lanes[defaultLane]
--- a/abci/example/kvstore/kvstore.go
+++ b/abci/example/kvstore/kvstore.go
@@ -167,26 +167,21 @@ func (app *Application) InitChain(_ context.Context, req *types.InitChainRequest
 // - `=` is not the first or last byte.
 // - if key is `val` that the validator update transaction is also valid.
 func (app *Application) CheckTx(_ context.Context, req *types.CheckTxRequest) (*types.CheckTxResponse, error) {
-	code := CodeTypeOK
 	// If it is a validator update transaction, check that it is correctly formatted
 	if isValidatorTx(req.Tx) {
 		if _, _, _, err := parseValidatorTx(req.Tx); err != nil {
-			code = CodeTypeInvalidTxFormat
+			return &types.CheckTxResponse{Code: CodeTypeInvalidTxFormat}, nil //nolint:nilerr // error is not nil but it returns nil
 		}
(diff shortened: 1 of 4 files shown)


abci/example/kvstore/kvstore.go
```diff
@@ -166,34 +167,64 @@ func (app *Application) InitChain(_ context.Context, req *types.InitChainRequest
 // - `=` is not the first or last byte.
 // - if key is `val` that the validator update transaction is also valid.
 func (app *Application) CheckTx(_ context.Context, req *types.CheckTxRequest) (*types.CheckTxResponse, error) {
+	code := CodeTypeOK
 	// If it is a validator update transaction, check that it is correctly formatted
 	if isValidatorTx(req.Tx) {
 		if _, _, _, err := parseValidatorTx(req.Tx); err != nil {
-			if app.useLanes {
-				return &types.CheckTxResponse{Code: CodeTypeInvalidTxFormat, Lane: app.lanes["val"]}, nil
```
jmalicevic, line 173: You could leave this for validator txs to always have a high priority, these txs should anyways not be of the form '2=2' , right?

abci/example/kvstore/kvstore.go
```diff
@@ -166,34 +167,64 @@ func (app *Application) InitChain(_ context.Context, req *types.InitChainRequest
...
-	var lane uint32
-	txHash := tmhash.Sum(req.Tx)
+
+	return &types.CheckTxResponse{Code: code, GasWanted: 1, Lane: lane}, nil
+}
+
+// assignLane deterministically computes a lane for the given tx.
+func (app *Application) assignLane(tx []byte) uint32 {
+	if isValidatorTx(tx) {
+		return app.lanes["val"] // priority 9
```
jmalicevic, line 195: Maybe leave it above as a form of early return in this case.

abci/example/kvstore/kvstore.go
```diff
@@ -166,34 +167,64 @@ func (app *Application) InitChain(_ context.Context, req *types.InitChainRequest
...
 		if _, _, _, err := parseValidatorTx(req.Tx); err != nil {
-			if app.useLanes {
-				return &types.CheckTxResponse{Code: CodeTypeInvalidTxFormat, Lane: app.lanes["val"]}, nil
-			}
-			return &types.CheckTxResponse{Code: CodeTypeInvalidTxFormat}, nil
+			code = CodeTypeInvalidTxFormat
 		}
 	} else if !isValidTx(req.Tx) {
-		return &types.CheckTxResponse{Code: CodeTypeInvalidTxFormat}, nil
+		code = CodeTypeInvalidTxFormat
```
jmalicevic, line 177: If it is invalid should we not return here immediately 
```suggestion
		return &types.CheckTxResponse{Code: CodeTypeInvalidTxFormat, Lane: uint32(0) }, nil
```

abci/example/kvstore/kvstore.go
```diff
@@ -166,34 +167,64 @@ func (app *Application) InitChain(_ context.Context, req *types.InitChainRequest
...
+	if isValidatorTx(tx) {
+		return app.lanes["val"] // priority 9
+	}
+
+	key, _, err := parseTx(tx)
+	if err != nil {
+		return app.lanes[defaultLane]
+	}
+
+	keyInt, err := strconv.Atoi(key)
```
jmalicevic, line 203:
```suggestion
// If the tx is of the form 2=2 we will assign it a lane. For any other type of transaction sent to the kvstore, it will go to the default lane.  
	keyInt, err := strconv.Atoi(key)
```

abci/example/kvstore/kvstore.go
hvanz: These txs have the form `val=...`.

abci/example/kvstore/kvstore.go
hvanz: Yes, probably that's better. I will change it.

abci/example/kvstore/kvstore.go
hvanz: Refactored CheckTx, so I think this is not needed.

abci/example/kvstore/kvstore.go
jmalicevic: Exactly so for those we could already return here, no need to do the other stuff.

jmalicevic (approved):

abci/example/kvstore/kvstore.go
hvanz: Below there's a check for `!app.useLanes`. If it holds, we don't have to return a lane.

PR #3763, opened 2024-08-20
refactor(lanes): Simplify concurrency of `BlockingIterator`

hvanz:
- Remove `addTxChMtx` and just use `txsMtx` instead to guard `addTxCh` and `addTxSeq`. These two variables are updated when adding a transaction, so they are already guarded by `txsMtx`.
- `addTxLaneSeqs` needs to be safely read concurrently by multiple iterators so I made it a `sync.Map`. With this change we don't need to lock `txsMtx` in `PickLane`.
- `PickLane` now is not blocking. Now, when it doesn't find a lane, it returns false, and blocking happens in `WaitNextCh`.
- A new variable `addTxLane` holds the lane where the latest transaction was added. Now each iterator that waits for a new tx can read this value to pick directly this lane.

### DIFF
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -489,11 +489,11 @@ func (mem *CListMempool) addTx(tx types.Tx, gasWanted int64, sender p2p.ID, lane
 	return true
 }
 
-// newTxCh returns a channel to wait until a new transaction is added to the
+// addedTxCh returns a channel to wait until a new transaction is added to the
 // mempool.
 //
 // Safe for concurrent use by multiple iterators.
-func (mem *CListMempool) newTxCh() <-chan struct{} {
+func (mem *CListMempool) addedTxCh() <-chan struct{} {
 	mem.txsMtx.RLock()
 	ch := mem.addTxCh
 	mem.txsMtx.RUnlock()
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -60,7 +60,6 @@ type CListMempool struct {
 	numTxs    int64                           // total number of txs in the mempool
 	addTxCh   chan struct{}                   // blocks until the next tx is added
 	addTxSeq  int64                           // helps detect if new txs have been added to a given lane
-	addTxLane types.Lane                      // lane where the most recent transaction was added
 
 	addTxLaneSeqs sync.Map // sequence of the last tx added to a given lane (types.Lane -> int64)
 
@@ -471,7 +470,6 @@ func (mem *CListMempool) addTx(tx types.Tx, gasWanted int64, sender p2p.ID, lane
 	mem.laneBytes[lane] += int64(len(tx))
 
 	// Notify iterators there's a new transaction.
-	mem.addTxLane = lane
 	close(mem.addTxCh)
 	mem.addTxCh = make(chan struct{})
 
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -472,10 +472,10 @@ func (mem *CListMempool) addTx(tx types.Tx, gasWanted int64, sender p2p.ID, lane
 	// Notify iterators there's a new transaction.
 	select {
 	case mem.addTxCh <- lane:
+		close(mem.addTxCh)
+		mem.addTxCh = make(chan types.Lane)
 	default:
 	}
-	close(mem.addTxCh)
-	mem.addTxCh = make(chan types.Lane)
 
 	// Update metrics.
 	mem.metrics.TxSizeBytes.Observe(float64(len(tx)))
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -58,8 +58,9 @@ type CListMempool struct {
 	laneBytes map[types.Lane]int64            // number of bytes per lane (for metrics)
 	txsBytes  int64                           // total size of mempool, in bytes
 	numTxs    int64                           // total number of txs in the mempool
-	addTxCh   chan types.Lane                 // blocks until the next tx is added
+	addTxCh   chan struct{}                   // blocks until the next tx is added
 	addTxSeq  int64                           // helps detect if new txs have been added to a given lane
+	addTxLane types.Lane                      // lane where the most recent transaction was added
 
 	addTxLaneSeqs sync.Map // sequence of the last tx added to a given lane (types.Lane -> int64)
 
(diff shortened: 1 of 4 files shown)


mempool/clist_mempool.go
```diff
@@ -490,6 +489,39 @@ func (mem *CListMempool) addTx(tx types.Tx, gasWanted int64, sender p2p.ID, lane
 	return true
 }
 
+// newTxCh returns a channel to wait until a new transaction is added to the
+// mempool.
+//
+// Safe for concurrent use by multiple iterators.
+func (mem *CListMempool) newTxCh() <-chan struct{} {
```
sergio-mena, line 496: `getTxCh`?

mempool/iterators.go
```diff
@@ -116,45 +116,48 @@ func NewBlockingIterator(mem *CListMempool) Iterator {
...
-// algorithm. A lane is valid if it is not empty or it is not over-consumed,
-// meaning that the number of accessed entries in the lane has not yet reached
-// its priority value in the current WRR iteration. It will block until a
-// transaction is available in any lane.
-func (iter *BlockingIterator) PickLane() types.Lane {
+// pickLane returns a _valid_ lane containing the next transaction to access
+// according to the WRR algorithm. A lane is valid if it is not empty or it is
+// not over-consumed, meaning that the number of accessed entries in the lane
+// has not yet reached its priority value in the current WRR iteration. It will
+// block until a transaction is available in any lane. It returns false if all
```
sergio-mena, line 143: > It will block until a transaction is available in any lane

Is that so?

mempool/iterators.go
```diff
@@ -116,45 +116,48 @@ func NewBlockingIterator(mem *CListMempool) Iterator {
 func (iter *BlockingIterator) WaitNextCh() <-chan Entry {
 	ch := make(chan Entry)
 	go func() {
+		// Pick a lane containing the next entry to access.
+		lane, ok := iter.pickLane()
+		if !ok {
+			// There are no transactions to take from any lane. Wait until a new
+			// transaction is added to the mempool.
+			<-iter.mp.newTxCh()
+			lane = iter.mp.latestTxLane()
```
sergio-mena, line 125: I have a small concern about this potentially breaking the WRR algorithm
Between lines 124 and 125, there's no lock held, so many things could happen between the two. For instance 3 transactions could be added to three different lanes. If that's the case, we'll pick the last of those three transactions (rather than following WRR).

Since anyway the callsite `WaitNextCh` accepts `nil`, I would just (1) wait on `newTxCh`, return `nil`, and let the callsite (the reactor) call us again, so we can apply WRR (and we know there's something as we unblocked, so no busy-wait).

Let me try to express it in code changes... bear with me

mempool/clist_mempool.go
hvanz: I renamed it `addedTxCh`.

mempool/iterators.go
hvanz: I forgot to update this comment. Fixed.

mempool/iterators.go
sergio-mena: Something like this:
```diff
diff --git a/mempool/iterators.go b/mempool/iterators.go
--- a/mempool/iterators.go
+++ b/mempool/iterators.go
@@ -122,7 +122,8 @@ func (iter *BlockingIterator) WaitNextCh() <-chan Entry {
 			// There are no transactions to take from any lane. Wait until a new
 			// transaction is added to the mempool.
 			<-iter.mp.newTxCh()
-			lane = iter.mp.latestTxLane()
+			close(ch)
+			return
 		}
 
 		// Add the next entry to the channel if not nil.
```

hvanz: Yeah, it makes sense. That would break WRR. How about something like the following, so we do everything inside the same function, without closing and opening the channel every time:
```diff
diff --git a/mempool/iterators.go b/mempool/iterators.go
index ead951044..db2123fc7 100644
--- a/mempool/iterators.go
+++ b/mempool/iterators.go
@@ -118,11 +118,11 @@ func (iter *BlockingIterator) WaitNextCh() <-chan Entry {
        go func() {
                // Pick a lane containing the next entry to access.
                lane, ok := iter.pickLane()
-               if !ok {
+               for !ok {
                        // There are no transactions to take from any lane. Wait until a new
                        // transaction is added to the mempool.
                        <-iter.mp.addedTxCh()
-                       lane = iter.mp.latestTxLane()
+                       lane, ok = iter.pickLane()
                }
 
                // Add the next entry to the channel if not nil.
```

mempool/iterators.go
```diff
@@ -116,45 +116,48 @@ func NewBlockingIterator(mem *CListMempool) Iterator {
...
+// has not yet reached its priority value in the current WRR iteration. It
+// returns false if all lanes are empty or don't have transactions that have not
+// yet been accessed.
+func (iter *BlockingIterator) pickLane() (types.Lane, bool) {
 	// Start from the last accessed lane.
 	lane := iter.sortedLanes[iter.laneIndex]
 
-	iter.mp.addTxChMtx.RLock()
-	defer iter.mp.addTxChMtx.RUnlock()
-
```
sergio-mena, line 141: I see `pickLane` no longer takes a lock on the mempool. I think that's why you turned `lastTxSeq` into a `sync.Map`.

I know this is the "most performant", but I'd still consider taking a read lock on the mempool (i.e., `txsMtx`). Why? Because not holding the lock means we're searching the mempool while it is potentially being modified. For instance (and we have considered this example a few weeks ago):
* mempool empty
* we call `pickLane()`
* while the `for` loop in line 145 is about to return in line 160
  * a new tx arrives
  * we add the tx to a lane the `for` loop has already visited
  * we close `addTxCh`, and create a new one
* `pickLane()` returns in line 160
* `WaitNextCh` blocks on the newly created `addTxCh`
* We end up in a situation where a tx is in the mempool, available for this iterator, but we are blocked on `addTxCh`

The example can be made worse replacing "a tx" by "a number of tx".

I know this is not likely to cause big problems in practice, but those quirks are never fully understood until they bite hard (days of troubleshooting).

So, if taking a read lock here is not impacting performance, I'd prefer that way, as the whole concurrency model here is much easier to understand.

mempool/iterators.go
sergio-mena: Yeah, like this one more

hvanz: Done.

sergio-mena: 🙏

mempool/iterators.go
sergio-mena: 100% agree. Since we'll soon be running a "dry-run" QA test with 200 nodes, we can be seeing there if taking `txsMtx` (read lock) is impacting performance w.r.t. `v1.x`. If we indeed see performance suffers, we can have a patch ready with the version where `pickLane` doesn't take the lock to "swap in". WDYT?

hvanz:
Closing in favor of #4106. The code optimization will be left for another time.

PR #3765, opened 2024-08-20
feat(p2p/metrics): add `MessageSendDelay` metrics [WIP]

cason:
This is an EXPERIMENT in order to assess some undesired behavior of the p2p communication layer. The idea is to run testnets/e2e tests with this additional metric in order to retrieve some result.

The solution is not optimized and SHOULD NOT BE MERGED at the moment.

---

This metric computes the average send delay for messages, per destination peer (ID) and channel (ID).

The metric is computed properly for the `peer.Send()` method because it is blocking. Since `peer.TrySend()` method is not blocking, we artificially assume a 10s duration when if fails (because the send queue is full).

---

#### PR checklist

- [ ] Tests written/updated


melekes:
What triggered this change? Is there an issue?

p2p/metrics.go
```diff
@@ -34,11 +35,15 @@ type Metrics struct {
...
 	MessageSendBytesTotal metrics.Counter `metrics_labels:"message_type"`
+	// Average delay for sending messages to a peer in a channel.
+	MessageSendDelaySeconds metrics.Gauge `metrics_labels:"peer_id, channel_id"`
 }
 
 type peerPendingMetricsCache struct {
 	mtx             sync.Mutex
 	perMessageCache map[reflect.Type]*peerPendingMetricsCacheEntry
+
+	perChannelChache map[byte]*peerPendingDelayMetrics
```
melekes, line 46:
```suggestion
	perChannelCache map[byte]*peerPendingDelayMetrics
```

cason:
> What triggered this change? Is there an issue?

Sorry for not explaining. This is a experiment that me and @sergio-mena are running.

p2p/metrics.go
```diff
@@ -79,6 +94,21 @@ func (c *peerPendingMetricsCache) AddPendingRecvBytes(msgType reflect.Type, addB
...
 
+func (c *peerPendingMetricsCache) AddPendingMessageSendDelay(chID byte, delay time.Duration) {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+	if entry, ok := c.perChannelChache[chID]; ok {
+		entry.count++
+		entry.sendDelay += delay
+	} else {
+		c.perChannelChache[chID] = &peerPendingDelayMetrics{
+			chID:      chID,
```
sergio-mena, line 105: Why do we need this field here (we have it as the map's key)?

p2p/peer.go
```diff
@@ -256,15 +256,39 @@ func (p *peer) Status() cmtconn.ConnectionStatus {
 //
 // thread safe.
 func (p *peer) Send(e Envelope) bool {
-	return p.send(e.ChannelID, e.Message, p.mconn.Send)
+	start := time.Now()
```
sergio-mena, line 259: `cmttime`? I know it doesn't make any difference, as it's not used as wallclock time, but to calculate duration, but with `cmttime` the code is more homogeneous.

p2p/metrics.go
cason: True, ehe.

p2p/peer.go
cason: We don't use `cmttime` in this file, but this is a good point. We lose the non-monotonic part, but this shouldn't be a problem.

p2p/peer.go
```diff
@@ -388,6 +405,12 @@ func (p *peer) metricsReporter() {
 						entry.pendingRecvBytes = 0
 					}
 				}
+				for chID, entry := range p.pendingMetrics.perChannelCache {
+					p.metrics.MessageAverageSendDelay.
+						With("peer_id", string(p.ID())).
+						With("channel_id", string(chID)).
+						Set(entry.sendDelay.Seconds() / float64(entry.count))
+				}
```
sergio-mena, line 413:
```suggestion
				}
				p.pendingMetrics.perChannelCache = make(map[byte]*peerPendingDelayMetrics)
```
As we want a gauge, we need to reset every time we report

sergio-mena:
I think we can close this PR, @cason ?

PR #3825, opened 2024-08-21
feat(lanes/e2e): let loader define the lane of each tx

hvanz:
In e2e, transactions generated with `runner load` have a constant prefix `"a="`, which does not work with how lanes are currently assigned. This PR lets `runner load` to set the lane of each tx that it generates, and it includes it in the tx's `Payload`. Then  on `CheckTx`, the app parses the payload to obtain the lane. 

Note that e2e's `TestApp_Tx` generates transaction values that are not `Payload`s, so for these cases, `CheckTx` should  not fail if it cannot parse the tx, it must just assign the 0 lane.

The loader assigns lanes picked randomly from the list of lanes defined in the app. This PR adds to the e2e manifest an option `load_lane_weights`, which is used by the loader to select the lane proportionally to its weight.

---

#### PR checklist

- [ ] Tests written/updated


test/loadtime/payload/payload.proto
```diff
@@ -14,5 +14,6 @@ message Payload {
   uint64                    size        = 3;
   google.protobuf.Timestamp time        = 4;
   bytes                     id          = 5;
-  bytes                     padding     = 6;
+  uint32                    lane        = 6;
```
hvanz, line 17: This is a breaking change that only affects e2e tests and QA. I added `lane` in position 6, and moving `padding` to 7, because I think `padding` needs to go last.

sergio-mena: I think we don't need padding to go last, it'll work either way

test/e2e/app/app.go
```diff
@@ -315,21 +323,25 @@ func (app *Application) CheckTx(_ context.Context, req *abci.CheckTxRequest) (*a
...
+	lane := extractLane(value)
 
 	return &abci.CheckTxResponse{Code: kvstore.CodeTypeOK, GasWanted: 1, Lane: lane}, nil
 }
 
+// extractLane returns the lane field if value is a Payload, otherwise returns 0.
+func extractLane(value string) uint32 {
+	valueBytes, err := hex.DecodeString(value)
+	if err != nil {
+		return 0
```
sergio-mena, line 335: Can we panic here? This should never happen, and could hide a problem hard to troubleshoot

test/e2e/pkg/testnet.go
```diff
@@ -108,6 +111,8 @@ type Testnet struct {
 	DefaultZone                                          string
 	PbtsEnableHeight                                     int64
 	PbtsUpdateHeight                                     int64
+	LanePriorities                                       []uint32
```
sergio-mena, line 114: What is the difference between the new fields `LoadLaneWeights`, and `LanePriorities`... I'd say we only need one of them?

test/e2e/pkg/testnet.go
```diff
@@ -108,6 +111,8 @@ type Testnet struct {
...
+	LanePrioritiesIndex                                  int
```
sergio-mena, line 115: What is this parameter for? I don't see this used anywhere

test/e2e/pkg/testnet.go
```diff
@@ -648,6 +673,32 @@ func (t Testnet) HasPerturbations() bool {
...
+	totalWeight := uint(0)
+	for _, w := range weights {
+		totalWeight += w
+	}
+	r := cmtrand.Int31n(int32(totalWeight))
+
+	// Return i when the random number falls in the i'th bucket.
+	cursor := uint(0)
+	for i := 0; i < len(weights); i++ {
+		cursor += weights[i]
```
sergio-mena, line 688:
```suggestion
	i, w := range weights {
		cursor += w
```

test/e2e/pkg/testnet.go
hvanz: Thanks, I'll remove it. It's dead code for a previous implementation.

test/e2e/app/app.go
hvanz: Done.

test/e2e/pkg/testnet.go
hvanz: `LanePriorities` stores the lanes as defined by the app when the testnet is initialised, so we don't fetch them every time we generate a tx.

test/loadtime/payload/payload.proto
hvanz: Ok, then I'll leave them in the original order.

test/e2e/pkg/manifest.go
```diff
@@ -92,6 +92,12 @@ type Manifest struct {
 	// assign lanes to generated transactions proportionally to their weights.
 	LoadLaneWeights []uint `toml:"load_lane_weights"`
 
+  // LogLevel specifies the log level to be set on all nodes.
```
sergio-mena, line 95:
```suggestion
	// LogLevel specifies the log level to be set on all nodes.
```

test/e2e/pkg/testnet.go
sergio-mena: Ah OK, I think I finally got it:

* `LanePriorities` get the value from `app.LaneDefinitions()` which is hard-coded (so always the same). Always three lanes: 9, 4, 1
* `LoadLaneWeights` is configurable via the manifest and defines, for each of the three lanes, how often the load generator is gonna generate TXs for that lane

If this is not accurate, please let me know.

hvanz: That's correct. I added a few comments to clarify.

PR #3827, opened 2024-08-21
feat(lanes): Add metrics `LaneSize`, `LaneBytes` and `TxDuration`

### LINKED ISSUE #3506: mempool/lanes: Add new metrics for size of each lane on number of txs and bytes
Currently there are two metrics for the size of the mempool `Size` and `SizeBytes`. We need similar metrics with a `lane` label. Then the two original metrics will become redundant because they could be computed from the new ones.

hvanz:
Closes #3506 

---

#### PR checklist

- [ ] Tests written/updated


mempool/clist_mempool.go
```diff
@@ -720,12 +744,23 @@ func (mem *CListMempool) Update(
...
 }
 
+// updateSizeMetrics updates the size-related metrics of a given lane.
+func (mem *CListMempool) updateSizeMetrics(lane types.Lane) {
+	label := strconv.FormatUint(uint64(lane), 10)
+	mem.metrics.LaneSize.With("lane", label).Set(float64(mem.lanes[lane].Len()))
+	mem.metrics.LaneBytes.With("lane", label).Set(float64(mem.LaneBytes(lane)))
+	// TODO: do we want to keep the following redundant metrics? The total sizes can be computed from the other two.
+	mem.metrics.Size.Set(float64(mem.Size()))
+	mem.metrics.SizeBytes.Set(float64(mem.SizeBytes()))
```
hvanz, line 761: Do we want to keep these two redundant metrics? The total sizes can be computed from the lane sizes.

jmalicevic: Can you easily do it in prometheus or grafana? If yes, then we could mark it as deprecated because this metric is heavily used, I would not just remove it.

jmalicevic (approved):

mempool/clist_mempool.go
hvanz: Yes, in Prometheus for example `cometbft_mempool_size` is the same as `sum(cometbft_mempool_lane_size) by (job)`. I will mark them as deprecated.

jmalicevic: Cool, then I think we can mark it as deprecated?

Issue #3828, opened 2024-08-21
privval: duplicate `SignerListener: Listening for new connection module=privval` during chain upgrades


tony-iqlusion:
### Setup

- **CometBFT version**: v0.38.9
- **Have you tried the latest version**: no
- **ABCI app**: gaiad
- **Environment**:
  - **OS**: Debian 12
  - **Install tools**:
  - **Others**:

### Config

```toml
priv_validator_laddr = "X.X.X.X:Y"
````

### What happened?

*During a chain upgrade*, CometBFT seemingly attempts to accept two `privval` connections. TMKMS only opens one such connection, which otherwise works/remains stable. But in the logs (see below), the node continues to print the following in a loop, which is confusing to users during a chain upgrade and spammy in logs during a time when people are trying to keep an eye on them to figure out what happened.

### What did you expect to happen?

These lines do not appear in the logs, though they seem to be a symptom of a deeper problem.

### How to reproduce it

This is the tough part: we only witness this during a chain upgrade. It has impacted several recent chain upgrades however, including today's gaia v19 upgrade as well as recent Neutron upgrades.

It seems to be happening with every recent chain upgrade, however, when TMKMS is being used as a remote signer.

The problem persists after the chain has started, but can be fixed after a chain upgrade has completed by restarting the node.

### Logs

```
2:36PM INF SignerListener: Listening for new connection module=privval
2:36PM ERR SignerListener: Error accepting connection err="accept tcp [::]:<port>: i/o timeout" module=privval
2:36PM INF SignerListener: Listening for new connection module=privval
... [etc] ...
```

The TMKMS side shows a stable connection without problems:

```
INFO tmkms::session: [cosmoshub-4@tcp://x.x.x.x:y] connected to validator successfully
WARN tmkms::session: [cosmoshub-4@tcp://x.x.x.x:y]: unverified validator peer ID! (ac9e6866a2b448a4444514b808f564bffe9d2b5d)
INFO tmkms::session: [cosmoshub-4@tcp://[10.11.8.11:5007](x.x.x.x:y)] signed Prevote:<nil> at h/r/s 21835201/0/1 (101 ms)
```

### `dump_consensus_state` output

N/A

### Anything else we need to know

This doesn't actually *break* anything, though it is very confusing during chain upgrades, and introduces a red herring when people are otherwise trying to debug TMKMS issues (e.g. people interpreted this error as potentially being a problem with enabling extension vote signing, even though everything was working)

[...]

PR #3830, opened 2024-08-21
fix(privval): CV ignore duplicate privval listen when connected

chillyvee:
Resolve https://github.com/cometbft/cometbft/issues/3828

On cometbft startup, duplicate privval request can enter the channel between the time listen begins and a send is attempted.  

The duplicate listen request will timeout since there is not another signer connecting, filling logs with errors.  A duplicate listen request is undesired.

---

#### PR checklist

- [x] Tests written/updated


chillyvee:
The automated test failures are interesting.  In seda mainnet, the daemon works well with the updates.  

Is there any chance the automated tests rely on the availability of a privval listener even when the remote signer is already connected?

chillyvee:
Also there is some concern since GetAvailableConnection by consuming connectionAvailableCh

```
func (se *signerEndpoint) GetAvailableConnection(connectionAvailableCh chan net.Conn) bool {
        select {
        case se.conn = <-connectionAvailableCh:
                return true
...
}        
```

Which is only "refilled"  by service loop.  Would this introduce a blocking condition service loop does not continuously fail and repopulate the connectionAvailableCh channel?

```
func (sl *SignerListenerEndpoint) serviceLoop() {
                        // We have a good connection, wait for someone that needs one otherwise cancellation
                        sl.Logger.Info("SignerListener: Connected")
                        select {
                        case sl.connectionAvailableCh <- conn:
                        case <-sl.Quit():
                                return
                        }
```

chillyvee:
Deadloack appears to be fixed by limiting mutex lock to variable assignment only

privval/signer_listener_endpoint.go
```diff
@@ -181,6 +181,14 @@ func (sl *SignerListenerEndpoint) serviceLoop() {
 	for {
 		select {
 		case <-sl.connectRequestCh:
+			// On start, listen timeouts can queue a duplicate connect request to queue
+			// while the first request connects.  Drop duplicate request.
+			if sl.IsConnected() {
+				sl.Logger.Error("SignerListener: Connected. Drop Listen Request")
```
melekes, line 187:
```suggestion
				sl.Logger.Debug("SignerListener: Connected. Drop Listen Request")
```

privval/signer_endpoint.go
```diff
@@ -53,12 +53,14 @@ func (se *signerEndpoint) GetAvailableConnection(connectionAvailableCh chan net.
...
-
+	// serviceLoop will deadlock if IsConnected() is called while WaitConnection() holds the mutex
 	select {
-	case se.conn = <-connectionAvailableCh:
-	case <-time.After(maxWait):
+	case conn := <-connectionAvailableCh:
+		// Lock only to modify se.conn
+		se.connMtx.Lock()
+		defer se.connMtx.Unlock()
+		se.conn = conn
```
melekes, line 62:
```suggestion
		se.conn = conn
		se.connMtx.Unlock()
```

privval/signer_endpoint.go
```diff
@@ -53,12 +53,14 @@ func (se *signerEndpoint) GetAvailableConnection(connectionAvailableCh chan net.
...
-	defer se.connMtx.Unlock()
-
+	// serviceLoop will deadlock if IsConnected() is called while WaitConnection() holds the mutex
 	select {
-	case se.conn = <-connectionAvailableCh:
-	case <-time.After(maxWait):
+	case conn := <-connectionAvailableCh:
+		// Lock only to modify se.conn
+		se.connMtx.Lock()
+		defer se.connMtx.Unlock()
```
melekes, line 61:
```suggestion
```

melekes (commented):
Thanks @chillyvee ❤️

melekes (approved):

chillyvee:
Thank you for your code suggestions @melekes

privval/signer_endpoint.go
```diff
@@ -53,12 +53,14 @@ func (se *signerEndpoint) GetAvailableConnection(connectionAvailableCh chan net.
...
 func (se *signerEndpoint) WaitConnection(connectionAvailableCh chan net.Conn, maxWait time.Duration) error {
-	se.connMtx.Lock()
-	defer se.connMtx.Unlock()
-
+	// serviceLoop will deadlock if IsConnected() is called while WaitConnection() holds the mutex
 	select {
-	case se.conn = <-connectionAvailableCh:
-	case <-time.After(maxWait):
+	case conn := <-connectionAvailableCh:
+		// Lock only to modify se.conn
```
cason, line 59: This does the same as `SetConnection()`. We should use it instead.

cason (commented):
I had really bad time trying to understand this code. Why do we keep locks for all calls, including calls with waiting involved (like the one fixed), with block read and writes involved?

Anyway, I would really appreciate a test unit producing the reported situation and a fix in the code that specifically addresses the test unit.

chillyvee:
> Why do we keep locks for all calls, including calls with waiting involved (like the one fixed), with block read and writes involved?

@cason Appreciate your review.

The locks prevent go from reporting race conditions on asynchronous reads and writes to the same variable in different threads.

We'll try to contribute a unit test and may need to ask for help in creating it.

chillyvee:
Test case added and checks pass

privval/signer_listener_endpoint.go
```diff
@@ -34,11 +34,13 @@ type SignerListenerEndpoint struct {
 	connectRequestCh      chan struct{}
 	connectionAvailableCh chan net.Conn
 
-	timeoutAccept time.Duration
-	pingTimer     *time.Ticker
-	pingInterval  time.Duration
+	timeoutAccept   time.Duration
+	acceptFailCount int
```
cason, line 38:
```suggestion

```

privval/signer_listener_endpoint.go
```diff
@@ -34,11 +34,13 @@ type SignerListenerEndpoint struct {
...
+	pingTimer       *time.Ticker
+	pingInterval    time.Duration
 
 	instanceMtx cmtsync.Mutex // Ensures instance public methods access, i.e. SendRequest
+	ivarMtx     cmtsync.Mutex // Ensures instance variable access, i.e. acceptFailCount
```
cason, line 43:
```suggestion
        acceptFailCount int
	ivarMtx                cmtsync.Mutex // Ensures instance variable access, i.e. acceptFailCount
```

cason: It is probably enough to have an atomic integer here, but works the same way with that.

privval/signer_listener_endpoint.go
```diff
@@ -181,9 +195,17 @@ func (sl *SignerListenerEndpoint) serviceLoop() {
...
+			if sl.IsConnected() {
+				sl.Logger.Debug("SignerListener: Connected. Drop Listen Request")
+				continue
+			}
+
+			// Listen for remote signer
 			conn, err := sl.acceptNewConnection()
 			if err != nil {
-				sl.Logger.Error("SignerListener: Error accepting connection", "err", err)
+				sl.Logger.Error("SignerListener: Error accepting connection", "err", err, "acceptFailCount", sl.getAcceptFailCount())
```
cason, line 208:
```suggestion
				sl.Logger.Error("SignerListener: Error accepting connection", "err", err, "failures", sl.getAcceptFailCount())
```

cason (approved):
Thank you for adding the test unit, identifying and fixing the behavior :)

privval/signer_listener_endpoint.go
melekes: I'm in favor of atomic int as well

chillyvee: changed to atomic.Uint32

privval/signer_listener_endpoint.go
chillyvee: updated

chillyvee:
Checks pass with atomic uint32

cason:
I think we should merge this once tests are ok (they should, the problems are due to concurrently changes on `main`). What do you think, @melekes ?

chillyvee:
All checks passing except for this:

https://github.com/cometbft/cometbft/pull/3830/checks?check_run_id=29365796524
```
Mergify / Rule: Make sure PR are up to date before merging (update)
failed 3 hours ago in 0s
Mergify doesn't have permission to update
For security reasons, Mergify can't update this pull request. Try updating locally.
GitHub response: refusing to allow a GitHub App to create or update workflow .github/workflows/e2e-manual-debug.yml without workflows permission
```

Is there anything we can do to assist?  Not familiar with this exact check, but doesn't look like there are any conflicts.

PR #3833, opened 2024-08-22
fix(lanes): revert changes to recheck mechanism

hvanz:
There's a problem when rechecking reaches the end of a lane. It finishes the recheck process, and all subsequent recheck responses are discarded.


jmalicevic (approved):
Temporary reverting until we come up with a better wait to call `init` only once. 

This current version has the bug @hvanz pointed out, it is also timing out after every lane. 
The cursor values in `findNextEntryMatching` also have to be updated and maintained differently to support lanes and this is not doing it.

PR #3837, opened 2024-08-22
fix(mempool): Log tx hashes as hex instead of bytes

hvanz:
Also, capitalise all first letters of logging messages.

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

andynog (approved):
lgtm

PR #3847, opened 2024-08-23
feat(lanes): Add non-blocking iterator

hvanz:
For reaping and rechecking. Implements WRR scheduling.

Also moved all iterator tests to `clist_iterator_test.go`.

---

#### PR checklist

- [x] Tests written/updated


mempool/errors.go
```diff
@@ -11,9 +11,6 @@ var ErrTxNotFound = errors.New("transaction not found in mempool")
 // ErrTxInCache is returned to the client if we saw tx earlier.
 var ErrTxInCache = errors.New("tx already exists in cache")
 
-// ErrLaneNotFound is returned to the client when a lane is not found.
-var ErrLaneNotFound = errors.New("lane not found in mempool")
```
jmalicevic, line 15: Why are we removing this?

hvanz: I think these are changes in the target branch.

mempool/clist_iterator_test.go
```diff
@@ -0,0 +1,404 @@
...
+	const numLanes = 3
+	const numTxs = 100
+
+	var wg sync.WaitGroup
+	wg.Add(2)
+
+	var counter atomic.Int64
+	go func() {
+		waitForNumTxsInMempool(numTxs, mp)
+		fmt.Println("mempool height ", mp.height.Load())
```
jmalicevic, line 137:
```suggestion
```

mempool/clist_mempool.go
```diff
@@ -959,12 +1049,11 @@ func (iter *CListIterator) PickLane() types.Lane {
 			continue
 		}
 
-		if iter.counters[lane] >= uint32(lane) {
+		if iter.counters[lane] >= uint(lane) {
```
sergio-mena, line 1052: Isn't this logic overlapping with what is now implemented in `NonBlockingWRRIterator) Next()` ? Could we further simplify if here?

hvanz: Probably it's overlapping. But PickLane is not in its final state. Once that code is stable we can see if we can refactor.

PR #3849, opened 2024-08-23
feat(crypto): modular crypto library with 100% backward compatibility

greg-szabo:
> This is an example implementation to initiate a discussion.

This PR makes the crypto library modular and allows the end-user to plug in a custom crypto library for signing without forking the CometBFT codebase.

The solution is 100% backward compatible, so all existing codebases work without any changes.

This solution is a hack. The main goal was 100% compatibility, which restricts us from using best practices. Below, you will find a few examples of where we could do better, either by making it more inconvenient to use (which improves robustness) or by breaking compatibility to some degree (which improves readability and future-proofing).

### The problem
Our current `crypto/ed25519` and `crypto/secp256k1` libraries are not enough for our end users. Adding more libraries would require more maintenance from the core team, and adding some kind of modular implementation would break the existing codebase.

### The proposed solution
Repurpose the `crypto/ed25519` implementation and add non-breaking functionality to override the library through behavior injection.

First, the publicly shared variables were collected into a `CustomOptions` struct so that they can be overridden when the new crypto library is registered at startup. Then, the structs with method implementations were collected into interfaces, like `CustomPrivKey`, `CustomPubKey`, or `CustomBatchVerifier`. When a custom library is registered, the ed25519 methods are overridden by these custom implementations.

If an end-user registers a custom crypto library implementation, it overrides the built-in ed25519 implementation. All calls to it will be redirected to the custom crypto library implementation, and the most important publicly accessible variables (`PubKeyName`, etc.) are also rewritten during the registration.

### Why override `ed25519`?
The crypto libraries are entangled with other parts of the codebase. `types/params.go` uses some constants directly, `priv_val/socker_listeners.go` directly uses the PrivKey struct, and the list goes on. If we implemented a `crypto/custom` module instead, the user would not be able to override these points and change the behavior of the codebase unless we change how those points are implemented.

> Limit: The user must decide to either use the built-in ed25519 implementation or override it with their own crypto library. Both cannot be used.

One use case involves a user forking CometBFT to change the underlying cryptographic library. We will likely find more cases like this than cases where the user wants multiple cryptographic libraries in parallel. Multiple libraries will also require breaking changes, so this seemed a better first step.

### TL;DR
The file `crypto/ed25519/ed25519.go` was modified to accept custom crypto libraries. An example of implementing a custom crypto library is in `crypto/custom/mysecp_test.go`. (It's a secp256k1 implementation different from the built-in one.) The custom implementation is initialized and used in the `crypto/custom/custom_test.go` file. `TestMain` shows how to register the custom library, and the tests below show how it is used. The test `TestProveCustomSecpCorrect` has detailed steps to reproduce the signing process using `openssl` to confirm that the same algorithm is implemented. (The built-in secp256k1 implementation is incompatible with the OpenSSL implementation because of compression.)

### More details
The first problem that makes overriding the crypto libraries hard is that they register their private and public key types for amino-encoding in the module's `init` function. This means that by the time the user has the chance to rename their private key or public key names, they are already registered. I did not find a standard way to delete entries from this registry.

Golang modules call the `init` functions in a particular order. For 100% backward compatibility, an earlier module's `init` function can make recommendations to the `init` function of the `ed25519` module. We are using this to signal to the `ed25519` module if it should register the default names (for the default behavior) or not register any PrivKey/PubKey types because the user will do that later. This is achieved by importing the `crypto/custom` module, an empty module (the user must use `import _`) that runs earlier than the `crypto/ed25519` module. It sets a flag that tells the `ed25519` module that a custom implementation will be registered and there is no need to do amino-encoding type registration in the `init` function. You can see this in action in the example at `crypto/custom/custom_test.go`.

The second (or better yet, "zeroeth") problem was managing that flag between the `custom` and `ed25519` modules. Both look at the `crypto/custom/indicator` module to avoid circular dependencies, which store the flag in a global variable that can only be set once. To summarize, if the user imports the `custom` module, it sets a flag in the `indicator` module, which is read by the `ed25519` module and tells it to wait for a custom crypto library implementation. If the `custom` module was not imported, the indicator module is untouched, and the `ed25519` module will assume the default implementation will be used.

> Limit: for 100% compatibility, weird imports have to be used. A small but breaking change could make this initialization much simpler.

The next issue to resolve was how to implement the behavior injection. The proper way would be to create custom functions for each method implemented in ed25519. For example, Sign would have a global variable with a function signature, something like this:
```
var CustomSign func(privKey PrivKey, msg []byte) ([]byte, error)
```
We should have around 10 globally reprogrammable function-variables so the `ed25519` module can call the user-defined custom code.

To simplify this example, I expanded on the interfaces defining most methods in the crypto library. Most functionality is described by `crypto.PrivKey`, `crypto.PubKey`, and other interfaces. So, the `CustomPrivKey`, `CustomPubKey`, and similar interfaces were born and added to the `crypto/ed25519` library. This has the benefit that I don't have to redefine a bunch of function signatures (`Sign(msg []byte)` is already defined in the interface). The drawback is that instantiating these interfaces with a specific implementation is done through global variables. The internal state has to be carefully managed (see the BatchVerification example). We might have race conditions when multiple signatures are signed in parallel. This was not observed, but if the assumption is correct, we must fall back to defining every custom method from these interfaces with the added `privKey PrivKey` as seen above. It's alright; it's just a bit more work. For the sake of discussion, I think this version is fine.

> Limit: Defining the custom crypto for `ed25519` through global variables may lead to race conditions. Instead, fall back on proper behavior injection with custom function signatures.

As the codebase evolves, we might break the current status quo. In that case, I'd like to highlight generics introduced in Go 1.18, which might make for a better custom implementation for us. It would be nice to define precisely what these crypto libraries have to show and use a standardized interface (or struct) to configure and use them.

This PR's goal was to highlight the compatibility concerns surrounding a modular crypto library system. To stay compatible, we must accept earlier decisions as the status quo. Disregard better design patterns or newer tools that make life simpler. At some point, we will have to start adding breaking changes and decide how much of the codebase we want to overthrow. I hope the discussions around this example and the topic in general will give us a better perspective.

Note: The PR was opened on `main`, but because this codebase doesn't change much, I expect it to be portable to older releases.

---

#### PR checklist

- [ ] Tests written/updated


melekes (commented):
Love this PoC 👍 

I agree with @sergio-mena above comment. We should also keep in mind Zaki's comment here: https://github.com/cometbft/cometbft/pull/3249#issuecomment-2173330817

melekes:
It looks like this work is stale. Plus, I don't see a path forward here. The concept is great, but implementing this in a flexible way (meaning something that will work with any curve) will require more work.

PR #3860, opened 2024-08-26
revert(crypto): revert BLS sig size downgrade

melekes:
because it's not Ethereum compatible and breaks Berachain

Reverts https://github.com/cometbft/cometbft/pull/3627

---

#### PR checklist

- [x] Tests written/updated


sergio-mena (approved):
LGTM

PR #3867, opened 2024-08-26
feat(p2p): export metrics regarding the rate limiter sleep time 

### LINKED ISSUE #3865: flowrate: export metrics regarding the rate limiter sleep time
As for #3864 we realized that send and receive rate limiters have impact on the overall performance of connections between nodes, and therefore of the nodes themselves (consensus duration, in particular).

We need some more data to understand the behavior of the flow rate limiter implementation. It is actually based on this package https://pkg.go.dev/github.com/mxk/go-flowrate/flowrate?utm_source=godoc that has not been recently updated, the last release is from 2014.

The package itself produce some data using the `Monitor.Status()` method. This method is invoked by the Multiples connection to provide its own `Status()`, which is consumed by `p2p.Peer` instances to produce some metrics. But at the moment, none of the data produced by the rate limiter is considered or reported as a metric.

In addition, the `Status` produced by the rate limiter does not include the data we find most relevant in our investigations, that is the time that each limiter spend on `time.Sleep` calls, as part of the [waitNextSample](https://github.com/cometbft/cometbft/blob/e67223f8f6eb5d687723d7cd62879d14d58ca5fb/internal/flowrate/flowrate.go#L261) method.

We should collect the total time spent on those `time.Sleep` calls and possibly the maximum amount of time spent on a request to the limiter during a given time interval. This information could be used to produce metrics in the `p2p` package and understand the impact of those idle periods/sleeps in the overall performance of the system.

cason:
Closes #3865.

The introduced metrics `p2p_recv_rate_limiter_delay` and `p2p_send_rate_limiter_delay` are counters, that keep track of the seconds spent by the respective rate limiters on wait/sleep calls.

You probably should use `rate(comet_p2p_recv_rate_limiter_delay[5s])` on Prometheus to get the amount of time per second that those rate limiters are sleeping. When those values get close to 1, the channel is stalled and consensus has a very unstable behavior, if it does not stall at all too.

---

#### PR checklist

- [x] Tests written/updated


internal/flowrate/flowrate.go
```diff
@@ -29,9 +29,9 @@ type Monitor struct {
 	sBytes int64         // Number of bytes transferred since sLast
 	sLast  time.Duration // Most recent sample time (stop time when inactive)
 	sRate  time.Duration // Sampling rate
-
-	tBytes int64         // Number of bytes expected in the current transfer
```
sergio-mena, line 33: I understand we're removing this because we don't use it

sergio-mena (approved):

internal/flowrate/flowrate.go
cason: I removed the `SetTransferSize()` method and all associated fields.

melekes (approved):
👍

internal/flowrate/flowrate.go
cason: I have undone this, as it does not behave to this PR.

sergio-mena: (y)

Issue #3868, opened 2024-08-26
e2e: Cannot run targeting a e2e node with version `v1.x`


andynog:
If we use a manifest such as:

```
[node.validator]
    version = "cometbft/e2e-node:v1.x"
```

When running the e2e test, it errors out with:

```
I[2024-08-26|18:00:47.891] Running all tests in ./tests/...             
--- FAIL: TestGRPC_Version (0.00s)
    --- FAIL: TestGRPC_Version/validator (0.00s)
        e2e_test.go:95: 
                Error Trace:    /Users/andynogueira/Development/Go/src/github.com/cometbft/cometbft/test/e2e/tests/grpc_test.go:33
                                                        /Users/andynogueira/Development/Go/src/github.com/cometbft/cometbft/test/e2e/tests/e2e_test.go:95
                Error:          Not equal: 
                                expected: "1.0.0-dev"
                                actual  : "1.0.0-rc1"
                                
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -1.0.0-dev
                                +1.0.0-rc1
                Test:           TestGRPC_Version/validator
FAIL
FAIL    github.com/cometbft/cometbft/test/e2e/tests     4.114s
FAIL
```

The root cause is the difference between the `CMTSemVer` in `version.go` between the `main` and `v1.x` branch

On `main`:

```
package version

const (
	// CMTSemVer is used as the fallback version of CometBFT
	// when not using git describe. It uses semantic versioning format.
	CMTSemVer = "1.0.0-dev"
```

On `v1.x`

```
package version

const (
	// CMTSemVer is the used as the fallback version of CometBFT
	// when not using git describe. It is formatted with semantic versioning.
	CMTSemVer = "1.0.0-rc1"
```

The `version.go` files needs to be updated on `main` to match the one in `v1.x`.

andynog:
@sergio-mena, I believe this change will not cause any side effect with the `e2e` infra or releases. If you have any concerns about it please let me know. Thanks!

sergio-mena:
I took a quick look at `TestGRPC_Version` and the logic of the test is incorrect. When you are testing en e2e node, its version is not necessarily the same version that was used to build the tests.

So the logic of the test should 
* either find a way to figure out the node's version (this information should be made available in the corresponding `Node` struct of the `Testnet`
* or be relaxed to just make sure that the version makes sense syntactically

andynog:
[...]

PR #3869, opened 2024-08-26
e2e: update version to match v1.x (#3868)

andynog:
close: 3868

update version to run e2e `v1.x` images

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -1,9 +1,9 @@
 package version
 
 const (
-	// CMTSemVer is used as the fallback version of CometBFT
-	// when not using git describe. It uses semantic versioning format.
-	CMTSemVer = "1.0.0-dev"
+	// CMTSemVer is the used as the fallback version of CometBFT
+	// when not using git describe. It is formatted with semantic versioning.
+	CMTSemVer = "1.0.0-rc1"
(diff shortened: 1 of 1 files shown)


melekes (approved):

version/version.go
```diff
@@ -1,9 +1,9 @@
 package version
 
 const (
-	// CMTSemVer is used as the fallback version of CometBFT
-	// when not using git describe. It uses semantic versioning format.
-	CMTSemVer = "1.0.0-dev"
+	// CMTSemVer is the used as the fallback version of CometBFT
```
hvanz, line 4:
```suggestion
	// CMTSemVer is used as the fallback version of CometBFT
```

version/version.go
```diff
@@ -1,9 +1,9 @@
 package version
 
 const (
-	// CMTSemVer is used as the fallback version of CometBFT
-	// when not using git describe. It uses semantic versioning format.
-	CMTSemVer = "1.0.0-dev"
```
sergio-mena, line 6: Sorry, but I don't agree with this change. See my comment [here](https://github.com/cometbft/cometbft/issues/3868#issuecomment-2313201043).

sergio-mena (changes requested):

PR #3873, opened 2024-08-27
test(mempool):  Add two `Update` benchmarks

hvanz:
Add new `BenchmarkUpdate` and `BenchmarkUpdateAndRecheck`.
Also fix `BenchmarkCheckDuplicateTx`.

---

#### PR checklist

- [ ] Tests written/updated


types/tx.go
```diff
@@ -91,12 +91,6 @@ func (txs Txs) hashList() [][]byte {
 // Txs is a slice of transactions. Sorting a Txs value orders the transactions
 // lexicographically.
 
-func (txs Txs) Len() int      { return len(txs) }
```
melekes, line 94: Go API break => we can't backport to v0.38.x

melekes (approved):

types/tx.go
jmalicevic: Do we need this field only for this benchmark? If so, can we extract it somehow differently to not break APIs jsut for the sake of testing. If we are using it somewhere else in the mempool lanes work then this function should go together in a PR with where it is useful in production, not here.

melekes: Or we can just leave it there for v0.38.x

hvanz: We can also mark these functions as deprecated (even in v0.38.x).

jmalicevic: Then we can leave them in 0.38.x, mark them as deprecated in v1.x and remove them later on.

jmalicevic (approved):

hvanz:
@Mergifyio backport v0.38.x

Issue #3415, opened 2024-07-03
blocksync: define the exact conditions for a node to attempt block syncing


cason:
[Block sync](https://github.com/cometbft/cometbft/blob/main/docs/explanation/core/block-sync.md) is a protocol that allows a node joining an existing network to catch-up _faster_ to the head (last block) of the blockchain. The same result, but at a substantially slower speed can be achieved by joining the consensus protocol and receiving from peers the blocks decided at each height of consensus, accompanied by a set of `Precommit` votes forming a commit for that block.

The conditions for a node to leave the Block Sync protocol and join the consensus protocol are encoded in the [`IsCaughtUp() method`](https://github.com/cometbft/cometbft/blob/4a077a6888c94aacb3c19212a3e2aafea9e8c000/internal/blocksync/pool.go#L182). There are somewhat complex but the rationale is to compare the local height, the latest block in the local copy of the blockchain, with the latest (highest) height reported by the node's peers as part of the Block Sync protocol. If the reported peers' heights is not more than 1 unit higher than the local height, the node is ready to join the consensus protocol.

There are some corner cases to consider, though. Essentially, the node should not block forever on Block Sync if it fails to retrieve block from its peers (after some reasonable amount of time). But the code at the moment requires the node to have peers that respond to the node's requests in a timely manner, which gives rise to other corner scenarios to be considered.

Of particular interest is the case in which the node running Block Sync is an active validator and it holds a substantial portion of the total validator voting power. In this scenario, most likely to be observed in testnets, the node might not need to receive any block from peers (i.e., to catch-up) because the network cannot commit blocks in the absence of the node, and its voting power as a validator. This particular scenario has been reported by users and is addressed by https://github.com/cometbft/cometbft/pull/3406.

The goal of this PR is to better formalize the conditions under which a node should remain attempting to retrieve blocks from its peers via Block Sync protocol, as we are likely to find out further corner cases not covered by the current implementation.

Notice that this issue is particularly relevant for networks from release v0.38.x, where block sync has became mandatory (see https://github.com/tendermint/tendermint/pull/8433 for additional context). Up to release v0.37.x, node operators could just disable Block Sync in the configuration file if they realize that a node was getting stuck in this phase of the node's bootstrap/recovery procedure.

### Definition of Done

- [ ] Define, validate and formalize the conditions for a node to stay in the Block Sync protocol
   - [ ] Stretch goal: write a specification and update/extend the documentation for this protocol
[...]

PR #3885, opened 2024-08-28
refactor(node): simplify 'node is blocking chain' logic

sergio-mena:
Contributes to #3415

This is mainly refactoring to simplify `onlyValidatorIsUs` and `localNodeBlocksTheChain` (since the latter implies the former).
It is a follow-up of #3406 (this is the part of #3406 that doesn't need to be backported)

---

#### PR checklist

- ~[ ] Tests written/updated~


types/validator_set.go
```diff
@@ -360,6 +360,14 @@ func (vals *ValidatorSet) findProposer() *Validator {
 	return proposer
 }
 
+func (vals *ValidatorSet) ValidatorBlocksTheChain(address []byte) bool {
+	_, val := vals.GetByAddress(address)
+	if val == nil {
+		return false
+	}
+	return val.VotingPower > vals.TotalVotingPower()/3
```
melekes, line 368:
```suggestion
	return val.VotingPower >= vals.TotalVotingPower()/3
```

?

melekes (approved):

types/validator_set.go
cason: We probably need a test unit for this :)

node/setup.go
```diff
@@ -309,7 +300,7 @@ func createMempoolAndMempoolReactor(
 		reactor := mempl.NewReactor(
 			config.Mempool,
 			mp,
-			waitSync,
+			true,
```
cason, line 303: I assume this is always true because block sync is mandatory?

cason (approved):
Legit.

cason:
There is a test error that might be related to the changes:
`
 2024/09/02 07:58:15 failed to unmarshal response: error in json rpc client, with http response metadata: (Status: 200 OK, Protocol HTTP/1.1) : RPC error -32603 - Internal error: endpoint is closed while node is catching up`

node/setup.go
sergio-mena: Actually it was so mindless that it was wrong 🤦.
Lucky, a couple of unit tests failed, which allowed me to spot the problem and validate the fix.

sergio-mena:
This was the first time in a long time that unit tests, rather than e2e, saved me from screwing up 😌

sergio-mena:
Unit tests really saved me here (rather than the usual e2e)

Issue #3864, opened 2024-08-26
p2p/conn: assymetry between send and receive limiters provokes instability under high load


cason:
A [multiplex connection](https://github.com/cometbft/cometbft/blob/main/p2p/conn/connection.go) instance, use to connect peers at p2p level, is equipped with two rate limiters ([`flow.Monitor`](https://github.com/cometbft/cometbft/blob/main/internal/flowrate/flowrate.go)), intended to enforce a maximum send and receive rates between two nodes. The maximum send and receive rates are configured by the `p2p.send_rate` and `p2p.recv_rate` parameters, with the same default value of `5120000`, i.e., 5000 KB/s.

During regular operation and under operational load, the rate limiters shouldn't affect the operation of the connections. They essentially compute the instant send and receive rates, attest that they are under the configured limits and immediately return. But once the load applied on the sender side, namely the number and size of messages enqueued to be sent to the peer, increases _and_ provided that the underlying connection has available bandwidth to support that load, the rate limiters start to impact the operation. They do that essentially by sleeping for some amount of time, so that, by not sending any bytes during that period, the instant rates drop onto levels below the configured maximum rates. Whenever one of these rate limiters sleep, they block the sending or receiving routines, meaning that no data is exchanged by the peers. The undesirable result of which is that all messages that are being sent will be artificially delayed by the amount of time that the rate limiters sleep.

The issue that me and mostly @sergio-mena have identified is that there is some asymmetry on the use of the send and receive rate limiters. This means that given the same (or maybe even similar) observed instant load they are likely to allow a different amount of traffic (bytes) to be written to the underlying channel, in the case of the send limiter, and be read from the underlying channel, in the case of the receive limiter. In short, as this needs to be measure in more details, the sending side can send a batch of messages once allowed by the rate limiter, while the receiving side is more stricter, receiving a single packet per call to the rate limiter. Being a little more precise, the send routine asks to the rate limiter to send `X` bytes and may end up sending `Y > X` bytes to the channel (potentially, in the worst case, `Y = 10X`), while the receive routine asks to the rate limiter to receive `X` bytes and reads _at most_ `X` bytes from the underlying channel. Notice that in both cases the exact amount of bytes read or written from or to the channel is a posteri provided to the rate limiters so that they can update their measured instant rates.

[...]

PR #3891, opened 2024-08-29
perf: add metrics in mempool, block executor; and remove `BlockProcessingTime`

sergio-mena:
Contributes to #3864

These new metrics are needed to monitor some of the scenarios described in #3864, where the bottleneck is not the sender's p2p rate limiter.

Removing `BlockProcessingTime`, as the metric is confusing (it's in ms, in contrast to other metrics expressing duration). Besides, the metric `MethodTimingSeconds` (for label `FinalizeBlock`) is measuring the same.

---

#### PR checklist

- ~[ ] Tests written/updated~


hvanz (approved):
Only updating the changelog is missing, otherwise LGTM.

andynog (approved):
The changes lgtm, I will approve, but I believe the docs would need to be updated too
https://docs.cometbft.com/v1.0/explanation/core/metrics

state/metrics.go
```diff
@@ -66,4 +63,7 @@ type Metrics struct {
 	// The duration of accesses to the state store labeled by which method
 	// was called on the store.
 	StoreAccessDurationSeconds metrics.Histogram `metrics_bucketsizes:"0.0002, 10, 5" metrics_buckettype:"exp" metrics_labels:"method"`
+
+	// The duration of event firing related to a new block
+	BlockEventsDurationSeconds metrics.Gauge
```
cason, line 68: Can we come with a better name for it?

Like `FireBlockEventsDelay` or something like that?

Moreover, all metrics should be in seconds by default. See the metric you have removed. The time unit should be part of the description, not the metric name.

cason (approved):
I don't like much the name of the new metric. But I see its relevancy.

state/metrics.go
sergio-mena: I added the `Seconds` suffix because it seems to be the norm in most of our existing metrics.

The one I've removed seems to be an exception in many ways (not having a units suffix, using `UnixNano, and being measured in milliseconds. Indeed I wasted several hours because of that mess. So, next time I work with a metric I want do know in which units it is measured.

Having the units in the description is OK, but as stated, most of our metrics have a suffix with the units, so we should be consistent.

sergio-mena: Following your proposal, I'll rename it `BlockEventsDurationSeconds` to `FireBlockEventsDelaySeconds`

PR #3904, opened 2024-08-30
feat(flowrate): new implementation [DRAFT]

cason:
DRAFT

---

#### PR checklist

- [ ] Tests written/updated


internal/flowrate/monitor.go
```diff
@@ -0,0 +1,64 @@
...
+	Status() Status
+
+	// Hack to set the current rEMA.
+	// Used by block sync reactor.
+	SetREMA(rEMA float64)
+}
+
+type Monitor struct {
+	FlowRate
+}
```
sergio-mena, line 37: What's the point of this `struct`? You can directly use the interface everywhere you use this, and makes the code simpler. If you prefer, you could rename the interface itself `Monitor`

internal/flowrate/xrate/monitor.go
```diff
@@ -0,0 +1,98 @@
+package xrate
+
+import (
+	"time"
+
+	"golang.org/x/time/rate"
+
+	legacyimpl "github.com/cometbft/cometbft/internal/flowrate/legacy"
```
sergio-mena, line 8:
```suggestion
	"github.com/cometbft/cometbft/internal/flowrate"
```

internal/flowrate/xrate/monitor.go
```diff
@@ -0,0 +1,98 @@
...
+
+	// rate.Limiter configuration
+	burst int
+	rate  int64
+
+	// Tokens reserved but not yet used
+	reserved int
+
+	statusMtx cmtsync.Mutex
+	status    legacyimpl.Status
```
sergio-mena, line 23:
```suggestion
	status    flowrate.Status
```

internal/flowrate/xrate/monitor.go
```diff
@@ -0,0 +1,98 @@
...
+	panic("SetREMA unimplemented by flowrate/xrate package")
+}
+
+func (m *Monitor) addSleepTime(duration time.Duration) {
+	m.statusMtx.Lock()
+	m.status.SleepTime += duration
+	m.statusMtx.Unlock()
+}
+
+func (m *Monitor) Status() legacyimpl.Status {
```
sergio-mena, line 92:
```suggestion
func (m *Monitor) Status() flowrate.Status {
```

internal/flowrate/xrate/monitor.go
cason: This does not work. The `legacy` contains a package with `flowrate` name.

internal/flowrate/monitor.go
cason: To not change any code on the `p2p` side.

internal/flowrate/monitor.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"time"
+
+	legacyimpl "github.com/cometbft/cometbft/internal/flowrate/legacy"
+	"github.com/cometbft/cometbft/internal/flowrate/xrate"
+)
+
+var DefaultImplementation = "xrate"
+
+type Status = legacyimpl.Status
```
sergio-mena, line 12: Actually, in order for the dependencies to be cleaner, I'd move the Status definition (which is common to both implementations) here, and then would import it  in `legacy`

cason: You would create a loop :)

internal/flowrate/xrate/monitor.go
```diff
@@ -0,0 +1,98 @@
...
+	// Update rate limiter parameters, if needed
+	// XXX: this should never change after the first call
+	if maxRate != m.rate {
+		m.limiter.SetLimit(rate.Limit(maxRate))
+		m.rate = maxRate
+	}
+	if want > m.burst {
+		m.limiter.SetBurst(want)
+		m.burst = want
+	}
```
sergio-mena, line 42: This is awkward. I think you're doing it to conform to the existing interface. If that's the case, please add a comment saying this will go away when we get rid of the legacy monitor.

internal/flowrate/xrate/monitor.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func (m *Monitor) Limit(want int, maxRate int64, block bool) (n int) {
+	// Update rate limiter parameters, if needed
+	// XXX: this should never change after the first call
+	if maxRate != m.rate {
+		m.limiter.SetLimit(rate.Limit(maxRate))
+		m.rate = maxRate
+	}
+	if want > m.burst {
```
sergio-mena, line 39: Are you sure `want` is a good approximation of a burst config?

internal/flowrate/xrate/monitor.go
```diff
@@ -0,0 +1,98 @@
...
+			panic("ReserveN returned false")
+		}
+
+		// Requested tokens are available after res.Delay()
+		delay := res.Delay()
+		if delay > 0 {
+			start := time.Now()
+			time.Sleep(delay)
+			sleepTime += time.Since(start)
+		}
```
sergio-mena, line 63: Wouldn't it be simpler (and more precise) if we accumulated all the `delay` values that we're getting from the limiter and did just one sleep just after the loop?

internal/flowrate/xrate/monitor.go
```diff
@@ -0,0 +1,98 @@
...
+	status    legacyimpl.Status
+}
+
+func New(_, _ time.Duration) *Monitor {
+	return &Monitor{
+		limiter: rate.NewLimiter(rate.Inf, 0),
+	}
+}
+
+func (m *Monitor) Limit(want int, maxRate int64, block bool) (n int) {
```
sergio-mena, line 32: I struggle to see what this method would be useful for if `block` is set to false...

cason: We never used it with `false` actually :)

internal/flowrate/xrate/monitor.go
cason: We always ask the same `want` in the p2p connection.

So this is a late initialization call at the end.

Looks terrible, but I think it is legit to have this check.

internal/flowrate/xrate/monitor.go
cason: Yes, it would be simpler. But not, it would be less precise.

I've tried both.

PR #3962, opened 2024-09-02
test(e2e): Log multiple failed sent txs in one line

hvanz:
Currently `runner load` will log an error for every tx that it failed to send, which can hundreds of lines. With this change, every distinct error will be logged once every ~1 second, with the number of times it failed during that time.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/test/e2e/runner/load.go
+++ b/test/e2e/runner/load.go
@@ -24,7 +24,7 @@ func Load(ctx context.Context, testnet *e2e.Testnet) error {
 	initialTimeout := 1 * time.Minute
 	stallTimeout := 30 * time.Second
 	chSuccess := make(chan struct{})
-	chFailed := make(chan struct{})
+	chFailed := make(chan error)
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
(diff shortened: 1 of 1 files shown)


test/e2e/runner/load.go
```diff
@@ -55,8 +56,9 @@ func Load(ctx context.Context, testnet *e2e.Testnet) error {
 		case <-chSuccess:
 			success++
 			timeout = stallTimeout
-		case <-chFailed:
+		case err := <-chFailed:
 			failed++
+			errorCounter[err.Error()]++
```
melekes, line 61: Shouldn't we use the type here?

melekes (approved):

test/e2e/runner/load.go
hvanz: You mean to use `error` as map key? That's not possible because `error` is not comparable.

melekes: https://go.dev/play/p/RbGVvqwAAap ?

melekes: anyway, it's not super important.

hvanz: yeah, I've tried that before, but for some reason, it didn't work because each error will be considered different, even when the corresponding strings are the same.

PR #3964, opened 2024-09-02
feat(e2e): simplify the relation between manifest and testnet

cason:
Refactors the `e2e` testbed to remove some unnecessary duplication between `Manifest` and `Testnet`, and `ManifestNode` and `Node`. Fields that are identical in the two types are removed from the latter, which includes (extends) the former. Fields that had the same name but different types or semantics in the two versions, were renamed in the manifest.

With this change, to add a new field to the configuration of a testnet or a node, it is enough to update the `Manifest` or `ManifestNode`.  The new field must also be added to `Testnet` or `Node` if it needs to be somehow parsed (see the examples of the `Seeds` and `PersistentPeers` fields).

---

#### PR checklist

- [ ] Tests written/updated


test/e2e/pkg/manifest.go
```diff
@@ -145,10 +145,10 @@ type Manifest struct {
 
 // ManifestNode represents a node in a testnet manifest.
 type ManifestNode struct {
-	// Mode specifies the type of node: "validator", "full", "light" or "seed".
+	// ModeStr specifies the type of node: "validator", "full", "light" or "seed".
 	// Defaults to "validator". Full nodes do not get a signing key (a dummy key
 	// is generated), and seed nodes run in seed mode with the PEX reactor enabled.
-	Mode string `toml:"mode"`
+	ModeStr string `toml:"mode"`
```
melekes, line 151: What is the benefit of putting type into a variable name?

test/e2e/pkg/testnet.go
```diff
@@ -179,46 +141,19 @@ func NewTestnetFromManifest(manifest Manifest, file string, ifd InfrastructureDa
...
-		VoteExtensionSize:                manifest.VoteExtensionSize,
-		PeerGossipIntraloopSleepDuration: manifest.PeerGossipIntraloopSleepDuration,
-		ExperimentalMaxGossipConnectionsToPersistentPeers:    manifest.ExperimentalMaxGossipConnectionsToPersistentPeers,
-		ExperimentalMaxGossipConnectionsToNonPersistentPeers: manifest.ExperimentalMaxGossipConnectionsToNonPersistentPeers,
-		ABCITestsEnabled: manifest.ABCITestsEnabled,
-		DefaultZone:      manifest.DefaultZone,
-		PbtsEnableHeight: manifest.PbtsEnableHeight,
-		PbtsUpdateHeight: manifest.PbtsUpdateHeight,
-	}
-	if manifest.InitialHeight > 0 {
```
melekes, line 220: Is this no longer needed?

test/e2e/pkg/manifest.go
cason: We have `Testnet.Mode` and had `Manifest.Node`. They are from different types.

Ok, it is true that `type Mode = String` but they are different variables.

I renamed all variables in `Manifest` that collide with names in `Testnet`.

test/e2e/pkg/testnet.go
cason: See line 155 of the new version.

In this old version we set to 1 on the constructor, the update to `manifest.InitialHeight` if it is set.

In the new version, `testnet.InitialHeight == manifest.InitialHeight` by construction. And if it is unset, we set it to 1 in line 156.

test/e2e/pkg/manifest.go
andynog: I'm also trying to understand the motivation for these renames here, struct fields are scoped to the struct, so you can't experience 'collisions' in a sense that you can't access fields of one struct while referencing another, even if fields have the same names.

Could you please point in the code where this collision could happen ? If it's just a matter of providing context via the field name, I'd think we should just apply KISS principle and leave the way it is. But I might have missed the point with the rename.

andynog (commented):
I've reviewed and made some comments and had some questions.

melekes (approved):

andynog (approved):
lgtm, overall it improves the code

test/e2e/pkg/manifest.go
andynog: All good

Issue #3878, opened 2024-08-27
node cannot sync if network has no snapshots


charithabandi:
Currently, If node has statesync enabled and if there are no snapshots in the network, the node just hangs there in the discovery phase. 

Can the behavior be changed to switch to blocksync mode if snapshots are not discovered after an extended period of time(configurable)?

cason:
> Can the behavior be changed to switch to blocksync mode if snapshots are not discovered after an extended period of time(configurable)?

Yes, this is a good to have. But notice that this will only work if there is a valid initial state.

melekes:
@charithabandi, what would be a good default value for `an extended period of time(configurable)`?

PR #3965, opened 2024-09-02
feat(statesync)!: replace discovery_time with max_discovery_time

melekes:
if the node can't discover snapshots for 2 min
(`statesync.max_discovery_time`), switch to blocksync. Remove
`statesync.discovery_time` from the configuration. If
`statesync.max_discovery_time` is zero, the node will be retrying
indefinitely.

Closes https://github.com/cometbft/cometbft/issues/3878

### DIFF
--- a/config/config.toml.tpl
+++ b/config/config.toml.tpl
@@ -405,8 +405,9 @@ trust_height = {{ .StateSync.TrustHeight }}
 trust_hash = "{{ .StateSync.TrustHash }}"
 trust_period = "{{ .StateSync.TrustPeriod }}"
 
-# Time to spend discovering snapshots before initiating a restore.
-discovery_time = "{{ .StateSync.DiscoveryTime }}"
+# Time to spend discovering snapshots before switching to blocksync. If set to
+# 0, state sync will be trying indefinitely.
+max_discovery_time = "{{ .StateSync.MaxDiscoveryTime }}"
 
 # Temporary directory for state sync snapshot chunks, defaults to the OS tempdir (typically /tmp).
 # Will create a new, randomly named directory within, and remove it when done.
--- a/statesync/reactor.go
+++ b/statesync/reactor.go
@@ -264,7 +264,7 @@ func (r *Reactor) recentSnapshots(n uint32) ([]*snapshot, error) {
 
 // Sync runs a state sync, returning the new state and last commit at the snapshot height.
 // The caller must store the state and commit in the state database and block store.
-func (r *Reactor) Sync(stateProvider StateProvider, discoveryTime time.Duration) (sm.State, *types.Commit, error) {
+func (r *Reactor) Sync(stateProvider StateProvider, maxDiscoveryTime time.Duration) (sm.State, *types.Commit, error) {
 	r.mtx.Lock()
 	if r.syncer != nil {
 		r.mtx.Unlock()
@@ -286,7 +286,8 @@ func (r *Reactor) Sync(stateProvider StateProvider, discoveryTime time.Duration)
 
 	hook()
 
-	state, commit, err := r.syncer.SyncAny(discoveryTime, hook)
+	const discoveryTime = 5 * time.Second
+	state, commit, err := r.syncer.SyncAny(discoveryTime, maxDiscoveryTime, hook)
 
 	r.mtx.Lock()
 	r.syncer = nil
--- a/statesync/syncer.go
+++ b/statesync/syncer.go
@@ -23,10 +23,6 @@ import (
 const (
 	// chunkTimeout is the timeout while waiting for the next chunk from the chunk queue.
 	chunkTimeout = 2 * time.Minute
-
-	// minimumDiscoveryTime is the lowest allowable time for a
-	// SyncAny discovery time.
-	minimumDiscoveryTime = 5 * time.Second
 )
 
 var (
@@ -139,18 +135,18 @@ func (s *syncer) RemovePeer(peer p2p.Peer) {
 	s.snapshots.RemovePeer(peer.ID())
 }
 
-// SyncAny tries to sync any of the snapshots in the snapshot pool, waiting to discover further
-// snapshots if none were found and discoveryTime > 0. It returns the latest state and block commit
-// which the caller must use to bootstrap the node.
-func (s *syncer) SyncAny(discoveryTime time.Duration, retryHook func()) (sm.State, *types.Commit, error) {
-	if discoveryTime != 0 && discoveryTime < minimumDiscoveryTime {
-		discoveryTime = 5 * minimumDiscoveryTime
-	}
+// SyncAny tries to sync any of the snapshots in the snapshot pool, waiting to
+// discover further snapshots if none were found within discoveryTime. It
+// returns the latest state and block commit which the caller must use to
+// bootstrap the node.
+//
+// If none snapshots are found after maxDiscoveryTime, errNoSnapshots is
+// returned.
+func (s *syncer) SyncAny(discoveryTime, maxDiscoveryTime time.Duration, retryHook func()) (sm.State, *types.Commit, error) {
+	timeStart := time.Now()
 
-	if discoveryTime > 0 {
-		s.logger.Info("Discovering snapshots", "discoverTime", discoveryTime)
-		time.Sleep(discoveryTime)
-	}
+	s.logger.Info(fmt.Sprintf("Discovering snapshots for %v", discoveryTime))
+	time.Sleep(discoveryTime)
 
 	// The app may ask us to retry a snapshot restoration, in which case we need to reuse
 	// the snapshot and chunk queue from the previous loop iteration.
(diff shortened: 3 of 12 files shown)


melekes:
I don't think `statesync.discovery_time` is necessary. The default 5 sec for retrying is good for most chains. Most of the timeouts like these are internal to CometBFT.

config/config.toml.tpl
```diff
@@ -409,8 +409,9 @@ trust_height = {{ .StateSync.TrustHeight }}
 trust_hash = "{{ .StateSync.TrustHash }}"
 trust_period = "{{ .StateSync.TrustPeriod }}"
 
-# Time to spend discovering snapshots before initiating a restore.
-discovery_time = "{{ .StateSync.DiscoveryTime }}"
+# Time to spend discovering snapshots before switching to blocksync. If set to
```
jmalicevic, line 412: Should we indicate here what the default is as 2min is not a short amount of time so people might want to lower this.

melekes: We can change the default to 30 sec or 1 min. Why would people want to go lower?

cason:
I think that the semantics of the configuration flags are different.

The original `discovery_time` is the minimum amount of time we wait for a good snapshot to be retrieved. Notice that we sleep for that time, except if set to `0`. The point here is that we want to wait for long enough to select the best possible snapshot.

The new `max_discovery_time` is the maximum amount of time we wait for _any_ snapshot, from what I understood. And we are still sleep for that amount of time, which does not make much sense to me. This means that in any circumstance, we wait for  `max_discovery_time` before attempting to install a snapshot.

Which is not a solution for the reported problem.

config/config.toml.tpl
jmalicevic: I dont think we need to change it, but I understood that the initial complaint was that they get stuck too long and want to move on.. So we leave 2 mins but just tell them. No need to change it ourselves.

cason:
To prove my point, in the `main` branch:

```
cometbft/statesync $ go test                                                                                                                                                                 
PASS
ok  	github.com/cometbft/cometbft/statesync	3.470s
```

While in this branch:

```
cometbft/statesync $ git switch 3878-statesync-max-discovery-tme
branch '3878-statesync-max-discovery-tme' set up to track 'origin/3878-statesync-max-discovery-tme'.
Switched to a new branch '3878-statesync-max-discovery-tme'
$ go test

```

I am still waiting for the output.

cason:
> I don't think `statesync.discovery_time` is necessary. The default 5 sec for retrying is good for most chains. Most of the timeouts like these are internal to CometBFT.

You can set it to zero, which means take the first snapshot you find (no sleep).

You can be more conservative and tell state sync to wait for more than `5s` in order to find a best snapshot.

cason:
In any case, if we create a `max_discovery_time` configuration, which is a good idea, we shouldn't sleep by that amount of time. We should only check if we still don't have a good snapshot after that duration.

melekes:
> I think that the semantics of the configuration flags are different.
> 
> The original `discovery_time` is the minimum amount of time we wait for a good snapshot to be retrieved. Notice that we sleep for that time, except if set to `0`. The point here is that we want to wait for long enough to select the best possible snapshot.
> 
> The new `max_discovery_time` is the maximum amount of time we wait for _any_ snapshot, from what I understood. And we are still sleep for that amount of time, which does not make much sense to me. This means that in any circumstance, we wait for `max_discovery_time` before attempting to install a snapshot.
> 
> Which is not a solution for the reported problem.

I don't think that's accurate.

We are still waiting for `discovery_time` to select the best possible snapshot. I merely removed it from the config.

The `max_discovery_time`, which should probably be named `max_discovery_time_before_switching_to_blocksync`, but it's a bit long, is the maximum time to discover any snapshot before switching to block sync. Note that before, we'd stay in limbo if no snapshots were discovered. Now, after `max_discovery_time`, we will switch to block sync.

statesync/syncer.go
```diff
@@ -165,11 +167,11 @@ func (s *syncer) SyncAny(discoveryTime time.Duration, retryHook func()) (sm.Stat
...
 		}
 		if snapshot == nil {
-			if discoveryTime == 0 {
+			if maxDiscoveryTime > 0 && time.Since(timeStart) >= maxDiscoveryTime {
 				return sm.State{}, nil, errNoSnapshots
 			}
 			retryHook()
-			s.logger.Info("sync any", "msg", log.NewLazySprintf("Discovering snapshots for %v", discoveryTime))
+			s.logger.Info("sync any", "msg", fmt.Sprintf("Discovering snapshots for %v", discoveryTime))
 			time.Sleep(discoveryTime)
```
melekes, line 175: here we sleep for 5sec as before

statesync/syncer.go
```diff
@@ -139,17 +139,19 @@ func (s *syncer) RemovePeer(peer p2p.Peer) {
...
 	}
 
-	if discoveryTime > 0 {
-		s.logger.Info("Discovering snapshots", "discoverTime", discoveryTime)
-		time.Sleep(discoveryTime)
-	}
+	timeStart := time.Now()
+
+	s.logger.Info(fmt.Sprintf("Discovering snapshots for %v", discoveryTime))
+	time.Sleep(discoveryTime)
```
cason, line 154: We always sleep here. Before we didn't if `discoveryTime == 0`.

cason:
I assume that this test should take 2 minutes, the maximum configured time:

```
$ go test -v -run TestSyncer_SyncAny_noSnapshots
=== RUN   TestSyncer_SyncAny_noSnapshots
^Csignal: interrupt
FAIL	github.com/cometbft/cometbft/statesync	236.409s
```

I think that this is not what is happening here.

statesync/syncer.go
melekes: > discoveryTime == 0

True. But the default value was 5sec. So, to skip time here, a user needed to change the variable's value in the config.

melekes:
```
r; echo -e 'go test ./statesync/...'; go test ./statesync/...
go test ./statesync/...
?       github.com/cometbft/cometbft/statesync/mocks    [no test files]
ok      github.com/cometbft/cometbft/statesync  3.664s
```

config/config.toml.tpl
melekes: > I understood that the initial complaint was that they get stuck too long and want to move on..

yeah, because we **never** switched to blocksync before. Not after 2min, never.

melekes:
I overlooked the duration of the tests. @cason thanks for pointing that out 🙏 🥇

andynog (approved):
looks good to me, made a few suggestions and comments.

andynog:
I've also did a quick test on a local setup with a `max_discovery_time` of `30 seconds` and it switches to `blocksync`

```
I[2024-10-25|19:59:43.213] Starting state sync                          module=statesync 
I[2024-10-25|19:59:43.213] Downloading trusted light block using options module=light 
I[2024-10-25|19:59:43.215] service start                                module=p2p peer=539ffc12a0ac78970dab31ae8cdcfdd4285b2162@10.186.73.3:26656 msg="Starting Peer service" impl="Peer{MConn{10.186.73.3:26656} 539ffc12a0ac78970dab31ae8cdcfdd4285b2162 out}"
I[2024-10-25|19:59:43.215] service start                                module=p2p peer=539ffc12a0ac78970dab31ae8cdcfdd4285b2162@10.186.73.3:26656 msg="Starting MConnection service" impl=MConn{10.186.73.3:26656}
I[2024-10-25|19:59:43.227] service start                                module=state msg="Starting Pruner service" impl=Pruner
I[2024-10-25|19:59:43.227] Started node                                 module=main nodeInfo="{ProtocolVersion:{P2P:9 Block:11 App:0} DefaultNodeID:0a9c99096b50a6d72a24d5aa5286d3f7022b3555 ListenAddr:tcp://10.186.73.4:26656 Network:gen-group00-0002 Version:1.0.0-dev Channels:40202122233038606100 Moniker:validator03 Other:{TxIndex:on RPCAddress:tcp://0.0.0.0:26657}}"
I[2024-10-25|19:59:43.227] Discovering snapshots for 5s                 module=statesync 
I[2024-10-25|19:59:43.227] Started pruning blocks                       module=state interval=10s
I[2024-10-25|19:59:48.231] sync any                                     module=statesync msg="Discovering snapshots for 5s"
I[2024-10-25|19:59:48.323] service start                                module=p2p peer=4fe295e4cfad69f1247ad85975c6fd87757195db@10.186.73.2:45776 msg="Starting Peer service" impl="Peer{MConn{10.186.73.2:45776} 4fe295e4cfad69f1247ad85975c6fd87757195db in}"
I[2024-10-25|19:59:48.323] service start                                module=p2p peer=4fe295e4cfad69f1247ad85975c6fd87757195db@10.186.73.2:45776 msg="Starting MConnection service" impl=MConn{10.186.73.2:45776}
I[2024-10-25|19:59:53.232] sync any                                     module=statesync msg="Discovering snapshots for 5s"
I[2024-10-25|19:59:58.233] sync any                                     module=statesync msg="Discovering snapshots for 5s"
I[2024-10-25|20:00:03.235] sync any                                     module=statesync msg="Discovering snapshots for 5s"
I[2024-10-25|20:00:08.236] sync any                                     module=statesync msg="Discovering snapshots for 5s"
I[2024-10-25|20:00:13.214] Ensure peers                                 module=pex numOutPeers=1 numInPeers=1 numDialing=0 numToDial=9
I[2024-10-25|20:00:13.215] We need more addresses. Sending pexRequest to random peer module=pex peer="Peer{MConn{10.186.73.3:26656} 539ffc12a0ac78970dab31ae8cdcfdd4285b2162 out}"
I[2024-10-25|20:00:13.215] No addresses to dial. Falling back to seeds  module=pex 
E[2024-10-25|20:00:13.236] State sync failed                            module=statesync err="no suitable snapshots found"
I[2024-10-25|20:00:13.236] service start                                module=blocksync msg="Starting BlockPool service" impl=BlockPool```

File: statesync/reactor.go [hunk 4]
```diff
@@ -286,7 +286,9 @@ func (r *Reactor) Sync(stateProvider StateProvider, discoveryTime time.Duration)
 
 	hook()
 
-	state, commit, err := r.syncer.SyncAny(discoveryTime, hook)
+	// The time to spend discovering snapshots before initiating a restore.
```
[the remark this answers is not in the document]
melekes: True, I will update

PR #3967, opened 2024-09-02
feat(e2e): `config` flag for arbitrary CometBFT configurations

### LINKED ISSUE #3832: e2e: allow customize Comet's config parameters for all or specific nodes
The `e2e` infrastructure creates a number of nodes that are executed as a network for testing purposes.

The nodes are created using the _default_ configuration values (from `config/config.go`), some of which are changed based on parameters provided in the manifest. But if one wants to run an e2e execution using _non-standard_ parameters, there are two possibilities:

1. Generate the network/testnet from the manifest, then edit the multiple `config/config.toml` files (manually)
2. Change the defaults values of the same parameters in `config/config.go`, which requires re-compiling Comet and producing a new image (i.e., at least 60s)

We should be able to configure in the manifest file for an execution _any_ CometBFT configuration parameter, so that it is applied to the configuration file of all nodes (in the case this setup happens in the general section of the manifest) or specific nodes (in the case this setup happens in the section referring to that specific node).

I can see different ways to implement this feature, but I am open to additional suggestions:

1. The `runner`, that reads the manifest, should be able to interpret any config parameter used the CometBFT's configuration file. For instance, if `p2p.send_rate = 10000` is declared in the general section of the manifest (or in the section of a given node), all nodes (resp., that given node) will have this setup on their configuration file, or at least this setup will override the default parameters in the configuration file. This would be an ideal solution, but I am afraid that it is not simple to implement.
2. The manifest could contain a flag with a filename, say `comet_config_file` pointing to a file that has the same format of traditional config files. The settings in this file would then replace the standards, like in the ordinary execution. This would be the same as creating a _default_ `config.Config` then loading configurations from the provided file, finally dumping the resulting configuration file to the produced home directories used by e2e. This solutions is less practical then the previous, but still much better than what we have now.
3. This is an option that somehow combines the previous two. We could create a configuration flag, say `comet_config`, which can be a field that can contain multiple lines (so, like a file), or a section whose content is interpreted as it was part of Comet configuration. This is pretty similar to 2., the difference is that instead of loading a file, we load a multi-line string parameter or a section, but like option 1. would allow we to keep everything that was customized inside the same manifest file.

Any other ideas are welcome, but I think that having this feature will speed up significantly debugging problems using the e2e infra.

cason:
Addresses #3832.

Introduces a flag `config` to the manifest. It is a list of strings that should have the format `key = value`. Keys are any of the configuration parameters included in a CometBFT configuration file.

The flag can be global, meaning that the configurations will be applied for every node in the network, or specific for a given node, when inserted in the section referring to that node. Local/specific configurations are applied after the global ones, therefore overriding them.

Example:

```
config = [
   "p2p.send_rate = 51200",
   "filter_peers = true",
]
```

The configurations are applying using `viper`, as I could not find another way to load configuration parameters. They are always loaded as a string, and this appears to work correctly.

Note: this PR is based atop #3964 for simplicity, but it does not need to.

---

#### PR checklist

- [ ] Tests written/updated


test/e2e/pkg/manifest.go
```diff
@@ -141,6 +141,11 @@ type Manifest struct {
 	// -1 denotes it is set at genesis.
 	// 0 denotes it is set at InitChain.
 	PbtsUpdateHeight int64 `toml:"pbts_update_height"`
+
+	// Config is a set of key-value config entries to write to CometBFT's
+	// configuration files for all nodes. The format is "key = value".
+	// Example: "p2p.send_rate = 512000.
```
melekes, line 147:
```suggestion
	// Example: "p2p.send_rate = 512000".
```

test/e2e/pkg/manifest.go
```diff
@@ -242,6 +247,11 @@ type ManifestNode struct {
 
 	// Simulated clock skew for this node
 	ClockSkew time.Duration `toml:"clock_skew"`
+
+	// Config is a set of key-value config entries to write to CometBFT's
+	// configuration files for this node. The format is "key = value".
+	// Example: "p2p.send_rate = 512000.
```
melekes, line 253:
```suggestion
	// Example: "p2p.send_rate = 512000".
```

melekes (approved):
👍

hvanz:
Wouldn't it be better to use in the manifest the `Config` struct of the `config` package instead of strings of the form`"key = value"`? Then in the manifest one could write
```
[node.validator01]
config.P2P.send_rate = 512000
```
That way we don't need to care when the configuration changes, and we can reuse the existing validation functions.

cason:
> Wouldn't it be better to use in the manifest the `Config` struct of the `config` package instead of strings of the form`"key = value"`? Then in the manifest one could write
> 
> ```
> [node.validator01]
> config.P2P.send_rate = 512000
> ```
> 
> That way we don't need to care when the configuration changes, and we can reuse the existing validation functions.

That was my original idea, but it does not work.

The `Config` object use the Viper setup, which `mapstructure` tags. Which are not recognized by the `toml` package we use to decode the manifest.

cason:
> That way we don't need to care when the configuration changes, and we can reuse the existing validation functions.

We still check the config parameters, as we call the validate method.

hvanz:
> The Config object use the Viper setup, with mapstructure tags. Which are not recognized by the toml package we use to decode the manifest.

yeah, I thought about that. We would need to use the `mapstructure` tags in the manifest too instead of the `toml` tags. But I guess the implementation would be quite different from this PR.

cason:
> yeah, I thought about that. We would need to use the mapstructure tags in the manifest too instead of the toml tags. 

Initially I was thinking on the opposite direction, namely, getting riding of this dependency on Viper. But after playing a little with it, we probably should use viper on e2e as well.

Notice that we don't need necessarily to use the `mapstructure` tag, see the companion PR for the genesis file. We can change the tag Viper uses, but we need to be consistent.

andynog (changes requested):
I think this logic is not working properly. Seems that the way is implemented might produce unnecessary behaviors. 

Quick test.

Had a simple manifest
```
config = [
    "p2p.send_rate = 51200",
    "filter_peers = true",
]

[node.validator]
```

This emitted 
```
(e2e) > ./build/runner -f networks/single.toml setup
I[2024-09-03|14:25:08.905] setup                                        msg="Generating testnet files in `networks/single`"
D[2024-09-03|14:25:08.906] Applying Comet config                        node=validator p2p.send_rate=51200
D[2024-09-03|14:25:08.906] Applying Comet config                        node=validator filter_peers=true
```

Checked the config and value for `filter_peers` was set
```
# If true, query the ABCI app on connecting to a new peer
# so the app can decide if we should keep the connection or not
filter_peers = true
```

But let's say I've misspelled the field name as `filter_peer` (singular not plural)
```
config = [
    "p2p.send_rate = 51200",
    "filter_peer = true",
]

[node.validator]
```

Then when I ran the runner `./build/runner -f networks/single.toml setup`, the runner accepted that field as if it was valid
```
(e2e) > ./build/runner -f networks/single.toml setup
I[2024-09-03|14:24:12.991] setup                                        msg="Generating testnet files in `networks/single`"
D[2024-09-03|14:24:12.992] Applying Comet config                        node=validator p2p.send_rate=51200
D[2024-09-03|14:24:12.992] Applying Comet config                        node=validator filter_peer=true
```
But in the generated `config.toml` the `filter_peers` is `false`
```
filter_peers = false
```
So if we are accepting these configuration as a "key = value" specified as a string, I think we need to at least validate that the `key` matches a real field in the `config` to avoid problems with renames or false expectations when running tests.

cason:
> So if we are accepting these configuration as a "key = value" specified as a string, I think we need to at least validate that the key matches a real field in the config to avoid problems with renames or false expectations when running tests.

I agree, but how to do that?

andynog (approved):
I will just unblock this, my previous comment is still valid but the implementation to solve the validation problem is not easy. Might need a future PR if we can get it to work.

PR #3969, opened 2024-09-02
feat(e2e): allow customizing genesis parameters on the manifest

### LINKED ISSUE #3961: e2e: allow customizing genesis parameters on the manifest
Similarly to #3832, we should be able to set all genesis parameters using the manifest.

At the moment we can customize some of them, using variables like `BlockMaxBytes`, `VoteExtensionsUpdateHeight`, etc. (see this [method](https://github.com/cometbft/cometbft/blob/c3cb1950ce8337b540cad6c2122c09ae45dd0e4e/test/e2e/runner/setup.go#L130)).

Ideally, we should be able to set any value. Which should not be particularly complex since the genesis is a JSON file and it should be easy to parse _any_ genesis field from the manifest, which is also a JSON file.

cason:
Closes #3961.

Similarly to #3967, introduces a new `genesis` field for manifest files. The format is `key = value` and multiple entries can be provided. For example,

```
genesis = [ 
    "consensus_params.evidence.max_bytes = 1000", 
    "chain_id = pr3969" 
]
```

After the genesis document is produced by the `setup` implementation of the runner, the provided genesis customized configurations are applied to the generated genesis file. The solution also uses `viper`, as it allows setting only some specific keys, leaving the pre-produced content untouched.

Note: this PR is based atop https://github.com/cometbft/cometbft/pull/3964 for simplicity, but it does not need to.

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

cason:
There is a lint error I don't have idea what it means:
```
 golangci-lint-full.......................................................Failed
- hook id: golangci-lint-full
- exit code: 1

WARN The linter 'exportloopref' is deprecated (since v1.60.2) due to: Since Go1.22 (loopvar) this linter is no longer relevant. Replaced by copyloopvar. 
test/e2e/runner/setup.go:18:2: import 'github.com/mitchellh/mapstructure' is not allowed from list 'main' (depguard)
	"github.com/mitchellh/mapstructure"
	^
```

andynog (changes requested):
Same comment made in #3967 

We need to validate the `key` with a valid field in `genesis` otherwise we might get undesired behaviors


```
(e2e) > ./build/runner -f networks/single.toml setup
I[2024-09-03|14:34:21.615] setup                                        msg="Generating testnet files in `networks/single`"
D[2024-09-03|14:34:21.616] Applying Genesis config                      consensus_params.evidence.max_bytes=1000
D[2024-09-03|14:34:21.616] Applying Genesis config                      chain_id=pr3969
```

here `chain_id` was set in the `genesis.json` file, but mispelling the field name to `chainid` 

```
(e2e) > ./build/runner -f networks/single.toml setup
I[2024-09-03|14:34:32.827] setup                                        msg="Generating testnet files in `networks/single`"
D[2024-09-03|14:34:32.827] Applying Genesis config                      consensus_params.evidence.max_bytes=1000
D[2024-09-03|14:34:32.827] Applying Genesis config                      chainid=pr3969
```

validation accepted invalid field name, so of course it was not set
```
{
  "genesis_time": "2024-09-03T14:34:32.827736Z",
  "chain_id": "single",
```

andynog (approved):
I will just unblock this, my previous comment is still valid but the implementation to solve the validation problem is not easy. Might need a future PR if we can get it to work.

test/e2e/pkg/testnet.go
```diff
@@ -399,10 +399,16 @@ func (t Testnet) Validate() error {
...
+	for _, entry := range t.Genesis {
+		tokens := strings.Split(entry, " = ")
+		if len(tokens) != 2 {
+			return fmt.Errorf("invalid genesis entry: \"%s\", "+
+
 	for _, entry := range t.Config {
 		tokens := strings.Split(entry, " = ")
 		if len(tokens) != 2 {
 			return fmt.Errorf("invalid config entry: \"%s\", "+
+
```
cason, line 411:
```suggestion
```

PR #3970, opened 2024-09-03
docs(crypto/secp256k1): fix Address docs

### LINKED ISSUE #3727: doc/spec/core/encoding: secp256k1 encoding is inconsistent with the documentation
We can see that the address of secp256k1 pubkey is the same with bitcoin and cosmos sdk.

It's derived from sha256 and then ripemd160

https://github.com/cometbft/cometbft/blob/fe696a4757c908415dd210ae241298b86dea5a3e/crypto/secp256k1/secp256k1.go#L146-L158

But the documentation 

> The address is the first 20-bytes of the SHA256 hash of the raw 32-byte public key

melekes:
Also:
* assert ripemd160.Size == crypto.AddressSize
* write tests for PrivKeySize and PubKeySize

Closes #3727

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/crypto/secp256k1/secp256k1.go
+++ b/crypto/secp256k1/secp256k1.go
@@ -33,13 +33,15 @@ var _ crypto.PrivKey = PrivKey{}
 // PrivKey implements PrivKey.
 type PrivKey []byte
 
-// Bytes marshalls the private key using amino encoding.
+// Bytes returns the privkey as bytes.
 func (privKey PrivKey) Bytes() []byte {
 	return []byte(privKey)
 }
 
 // PubKey performs the point-scalar multiplication from the privKey on the
 // generator point to get the pubkey.
+//
+// See secp256k1.PrivKeyFromBytes.
 func (privKey PrivKey) PubKey() crypto.PubKey {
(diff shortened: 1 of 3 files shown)


cason (approved):
Legit.

crypto/secp256k1/secp256k1.go
```diff
@@ -139,22 +143,33 @@ const PubKeySize = 33
...
 	hasherSHA256 := sha256.New()
-	_, _ = hasherSHA256.Write(pubKey) // does not error
+	_, err := hasherSHA256.Write(pubKey)
+	if err != nil {
+		panic(err)
+	}
 	sha := hasherSHA256.Sum(nil)
 
+	// Check if the size of the hash is what we expect.
+	if ripemd160.Size != crypto.AddressSize {
```
islishude, line 159: I think we can move this to init()

islishude:
the haher.Write() returns error is because it implements io.Writer interface

fyi 

https://pkg.go.dev/crypto/sha256#Sum256

https://cs.opensource.google/go/x/crypto/+/refs/tags/v0.26.0:ripemd160/ripemd160.go;l=64

they don't actually have an error

PR #3973, opened 2024-09-03
feat(config): disallow flood mempool w/ 0 capacity

### LINKED ISSUE #2824: mempool size should not allow a value equals 0 in the config
I think we should change the mempool logic to force the size config parameter to be greater than 0 (don't allow equal to 0). That can cause some weird behavior with the isFull logic and CheckTx

```
func (mem *CListMempool) isFull(txSize int) error {
	var (
		memSize  = mem.Size()
		txsBytes = mem.SizeBytes()
	)

	if memSize >= mem.config.Size || int64(txSize)+txsBytes > mem.config.MaxTxsBytes {
		return ErrMempoolIsFull{
			NumTxs:      memSize,
			MaxTxs:      mem.config.Size,
			TxsBytes:    txsBytes,
			MaxTxsBytes: mem.config.MaxTxsBytes,
		}
	}

	return nil
}
```

If the `memSize` is zero (mempool empty) and the value for `mem.config.Size` is also `0` then isFull will return an error (`0 == 0`) , not that is the mempool is full but it is actually has no capacity to store txs. 

And in `CheckTx` the `isFull` call is the first thing that is checked so it will always return an error

```
func (mem *CListMempool) CheckTx(tx types.Tx) (*abcicli.ReqRes, error) {
	mem.updateMtx.RLock()
	// use defer to unlock mutex because application (*local client*) might panic
	defer mem.updateMtx.RUnlock()

	txSize := len(tx)

	if err := mem.isFull(txSize); err != nil {
		return nil, err
	}
...
```
and in the config for the mempool we allow the Size to be 0, only check for negative numbers
```
func (cfg *MempoolConfig) ValidateBasic() error {
	if cfg.Size < 0 {
		return cmterrors.ErrNegativeField{Field: "size"}
	}
...
```
I know that it would be rare for someone to set the mempool size to `0` in the config. This is not a bug per se but just a weird behavior in my opinion. 

We should think about if this should be allowed and change the logic (in validate, check `if cfg.Size <= 0`) and update the config doc for mempool.size to say only values greater than `0` would be valid . 

Ideally the mempool size should set to at least a factor of the `max_tx_bytes` and `max_txs_bytes`, like if 1MB and 5MB respectively then the size should be set to at least 5 (txs)
https://github.com/cometbft/cometbft/blob/main/docs/references/config/config.toml.md#mempoolsize

melekes:
Closes #2824

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -966,7 +966,7 @@ type MempoolConfig struct {
 	// transactions and a 5MB maximum mempool byte size, the mempool will
 	// only accept five transactions.
 	MaxTxsBytes int64 `mapstructure:"max_txs_bytes"`
-	// Size of the cache (used to filter transactions we saw earlier) in transactions
+	// Size of the cache (used to filter transactions we saw earlier) in transactions.
 	CacheSize int `mapstructure:"cache_size"`
 	// Do not remove invalid transactions from the cache (default: false)
 	// Set to true if it's not possible for any invalid transaction to become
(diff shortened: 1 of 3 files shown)


andynog (approved):
lgtm, thanks @melekes

PR #3978, opened 2024-09-03
test(mempool/lanes): Prepare consensus tests for lanes

hvanz:
In the consensus tests related to mempool, the mempool is initialised without lanes info. This PR queries the app and pass the lanes info when creating the mempool.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/internal/consensus/common_test.go
+++ b/internal/consensus/common_test.go
@@ -430,11 +430,24 @@ func subscribeToVoterBuffered(cs *State, addr []byte) <-chan cmtpubsub.Message {
 }
 
 // -------------------------------------------------------------------------------
+// application
+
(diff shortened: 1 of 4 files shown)


internal/consensus/common_test.go
```diff
@@ -429,22 +429,36 @@ func subscribeToVoterBuffered(cs *State, addr []byte) <-chan cmtpubsub.Message {
 	return ch
 }
 
+// -------------------------------------------------------------------------------
+// application
+
+func newAppWithInfo(t *testing.T) (*kvstore.Application, []byte, *mempl.LanesInfo) {
```
jmalicevic, line 435: I would return the whole response here so the tests can reuse what they need. We might have a case in the future where we need more fields.

hvanz: Done.

jmalicevic (approved):

PR #3980, opened 2024-09-03
test(mempool/lanes): Add initialisation functions to kvstore that don't use lanes

hvanz:
This is in preparation for testing applications that use and don't use lanes.

### DIFF
--- a/abci/example/kvstore/kvstore.go
+++ b/abci/example/kvstore/kvstore.go
@@ -72,22 +72,36 @@ func NewApplication(db dbm.DB, lanes map[string]uint32) *Application {
 	}
 }
 
-// NewPersistentApplication creates a new application using the goleveldb database engine.
-func NewPersistentApplication(dbDir string) *Application {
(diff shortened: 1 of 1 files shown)


abci/example/kvstore/kvstore.go
```diff
@@ -88,13 +79,28 @@ func NewPersistentApplication(dbDir string) *Application {
 	if err != nil {
 		panic(fmt.Errorf("failed to create persistent app at %s: %w", dbDir, err))
 	}
-	return NewApplication(db)
+	return NewApplication(db, DefaultLanes())
 }
 
 // NewInMemoryApplication creates a new application from an in memory database.
 // Nothing will be persisted.
 func NewInMemoryApplication() *Application {
```
jmalicevic, line 87: You need a new persistent application with and w/o lanes

hvanz: Done.

jmalicevic (approved):

PR #3984, opened 2024-09-03
fix(light): cross-check proposer priorities in retrieved validator sets

cason:
The light client implementation compares the `ValidatorSet` instances received from different sources (the primary against one or more witnesses). In particular, the state of the proposer selection algorithm, represented by the  `ProposerPriority` field of each `Validator` present in the `ValidatorSet`, should be identical.

In order to compare those fields, we add  a`ProposerPriorityHash()` to the `ValidatorSet` type. This is needed because the  `Hash()` method of the same type does not include the `ProposerPriority` fields in its computation.

When the light client detects distinct states of the proposer selection algorithm, the `VerifyLightBlockAtHeight` method of the `light.Client` type, used for instance by the state sync protocol, returns an error.

---

#### PR checklist

- [x] Tests written/updated


sergio-mena (approved):

cason:
https://github.com/cometbft/cometbft/security/advisories/GHSA-g5xx-c4hv-9ccc

PR #3987, opened 2024-09-03
test(mempool/lanes): Use app with lanes in tests + do not test ordering

hvanz:
Depends on #3978 and #3980 (then we should merge feature branch into target branch). 

This PR removes from kvstore the option `useLanes`, so it's not possible to disable lanes once the app is created. All tests use kvstore with lanes as default, except those that initialise the app with `New...ApplicationWithoutLanes`.

Also, some reactor tests were failing when checking that txs were received in order. We only need to check that the txs are in the mempool; order is tested elsewhere.


abci/example/kvstore/kvstore.go
```diff
@@ -125,23 +138,19 @@ func (app *Application) Info(context.Context, *types.InfoRequest) (*types.InfoRe
...
-			Data:                fmt.Sprintf("{\"size\":%v}", app.state.Size),
-			Version:             version.ABCIVersion,
-			AppVersion:          AppVersion,
-			LastBlockHeight:     app.state.Height,
-			LastBlockAppHash:    app.state.Hash(),
-			LanePriorities:      app.lanePriorities,
-			DefaultLanePriority: app.lanes[defaultLane],
-		}, nil
+	var defaultLanePriority uint32
+	if len(app.lanes) > 0 {
```
jmalicevic, line 142: What happens if `app.lanes` is `nil`? Should this be 

```
if app.lanes != nil {

   defaultLanePriority = app.lanes[defaultLane] 
} 
```

mempool/reactor_test.go
```diff
@@ -614,7 +614,7 @@ func checkTxsInOrder(t *testing.T, txs types.Txs, reactor *Reactor, reactorIndex
 	// Check that all transactions in the mempool are in the same order as txs.
 	reapedTxs := reactor.mempool.ReapMaxTxs(len(txs))
 	require.Equal(t, len(txs), len(reapedTxs))
-	for i, tx := range txs {
+	for i, tx := range reapedTxs {
```
jmalicevic, line 617: I think this cannot stay this way because the check is reundant (compares `tx` which is assigned the value `reapedTx[i]` with `reapedTxs[i]`

hvanz: That's right. We actually don't need to check that txs are in order, just that they are in the mempool. I changed them.

abci/example/kvstore/kvstore.go
hvanz: It's the same, but I'm changing it to make it more clear.

mempool/reactor_test.go
```diff
@@ -434,8 +434,8 @@ func TestMempoolReactorMaxActiveOutboundConnectionsNoDuplicate(t *testing.T) {
 
 	// Now the third reactor should start receiving transactions from the first reactor and
 	// the fourth reactor from the second
-	checkTxsInOrder(t, txs, reactors[2], 0)
-	checkTxsInOrder(t, txs, reactors[3], 0)
+	checkTxsInMempool(t, txs, reactors[2], 0)
```
jmalicevic, line 437: The tests do not need to confirm order as we already have separate unit tests to confirm that works.

jmalicevic (approved):

PR #3988, opened 2024-09-03
chore(deps): replace gofrs/uuid with google/uuid

melekes:
to reduce the dependencies surface.

---

#### PR checklist

- [ ] Tests written/updated


andynog (approved):
lgtm, only suggestion is to remove the old one from https://github.com/cometbft/cometbft/blob/c7c2f97a99f6dd385b514923cdc3fdcad3bf84b9/.golangci.yml#L101

also, I haven't generated a report with `load` results but that would be a good idea to check if everything works as expected.

melekes:
> also, I haven't generated a report with load results but that would be a good idea to check if everything works as expected.

I've checked it ✅

PR #3989, opened 2024-09-03
feat(internal/metrics): copy metrics package from go-kit

melekes:
We're only using a handful of lines, so it makes sense to copy to reduce the dependency surface. The last commit in go-kit was 6 months ago, and it seems abandoned [go-kit/kit](https://github.com/go-kit/kit).

---

#### PR checklist

- [x] Tests written/updated


go.mod
```diff
@@ -1,14 +1,16 @@
 module github.com/cometbft/cometbft
 
-go 1.22
+go 1.22.5
+
+toolchain go1.23.0
```
andynog, line 5: any particular reason we need to force the `1.23` toolchain for building the project ?

andynog (approved):
at first I tried to understand why we are copying the files instead of using a fork, but then realized that the original package include logic for many providers other than prometheus which is the only one we use, so I agree it makes sense here to just copy what we need to reduce the surface area.

Overall the changes look good to me. I've also ran an `e2e` test and I could get the data on my grafana dashboard so I assume everything is good. 

Thanks!

cason:
I am not sure if I agree with this proposal. This means that we are maintaining this package now? How often the upstream package is updated, has bugs to be fixed, etc.?

Moreover, why an internal package? If someone wants to extend Comet and add new metrics to their implementation?

Anyway, I think we should move to a modular solution and adding internal packages is definitely not the way to get there.

melekes:
> This means that we are maintaining this package now? How often the upstream package is updated, has bugs to be fixed, etc.?

**Six months ago**. No bug fixes. It's unmaintained.

> Moreover, why an internal package? If someone wants to extend Comet and add new metrics to their implementation?

There is no preference here. I can make it public.

go.mod
melekes: Not really. `go mod tidy` adds it automatically

cason:
Ok, the fact that it is unmaintained is an important information, can you add it to the PR description?

I would render it public, adding it to `libs/` for instance.

melekes:
> Ok, the fact that it is unmaintained is an important information, can you add it to the PR description?
> 
> I would render it public, adding it to `libs/` for instance.

done ✅

cason:
The failing test is related to the package being removed:

https://github.com/cometbft/cometbft/actions/runs/10714492203/job/29708265321?pr=3989

melekes:
> The failing test is related to the package being removed:
> 
> [cometbft/cometbft/actions/runs/10714492203/job/29708265321?pr=3989](https://github.com/cometbft/cometbft/actions/runs/10714492203/job/29708265321?pr=3989)

It looks like a non-deterministic failure. I will take a look, thanks

cason:
Ok, the test units on this copied library are producing super random errors:

https://github.com/cometbft/cometbft/actions/runs/10944499019/job/30386514771?pr=4129

https://github.com/cometbft/cometbft/actions/runs/10944445079/job/30386330921?pr=4104

PR #3996, opened 2024-09-04
fix(mempool/lanes): Do not reuse iterator when reaping

hvanz:
There's a risk of race conditions on the iterator for reaping, as shown in the [failed test](https://github.com/cometbft/cometbft/actions/runs/10686803248/job/29622932068?pr=3638) TestBroadcastTxSync. We assumed that Reap methods are only called by `BlockExecutor`, but this is not true. In particular, `ReapMaxTxs` is called by the RPC endpoint `UnconfirmedTxs`. 

This PR creates a new iterator on every Reap call, instead of reusing only one.

Benchmarks with one reusable iterator and with an iterator created on each Reap call (this PR).
```
BenchmarkReap-8             1272            894477 ns/op          485765 B/op      10001 allocs/op
BenchmarkReap-8             1269            928269 ns/op          486134 B/op      10006 allocs/op
```

### DIFF
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -64,8 +64,6 @@ type CListMempool struct {
 	defaultLane types.Lane
 	sortedLanes []types.Lane // lanes sorted by priority, in descending order
 
-	reapIter *NonBlockingWRRIterator
-
 	// Keep a cache of already-seen txs.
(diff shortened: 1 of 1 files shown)


sergio-mena (approved):
LGTM

PR #4006, opened 2024-09-05
feat(consensus): print ErrAddingVote with a reason

### LINKED ISSUE #2395: error adding vote / failed to process message error logs
## Bug Report

Screenshot 2024-02-20 at 10 21 57 PM

This occurs in the following err check https://github.com/osmosis-labs/cometbft/blob/490dd752eb8a1a934542f70fdc8960594e06973b/consensus/state.go#L890-L899

Wanted to report here that we get this error fairly often, and wanted to ensure the comet team was tracking, and if there are plans to fix in a future upgrade.

This is on cometbft v0.37.4 on current Osmosis mainnet.

melekes:
in "Failed to process message" error log instead of a separate Info log.

Closes #2395

Before:

```
failed to process message err=error adding vote
```

After:

```
Failed to process message err=error adding vote: invalid signature
```

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/internal/consensus/state.go
+++ b/internal/consensus/state.go
@@ -2247,7 +2247,7 @@ func (cs *State) tryAddVote(vote *types.Vote, peerID p2p.ID) (bool, error) {
 		// If the vote height is off, we'll just ignore it,
 		// But if it's a conflicting sig, add it to the cs.evpool.
 		// If it's otherwise invalid, punish peer.
-		//nolint: gocritic
+
 		if voteErr, ok := err.(*types.ErrVoteConflictingVotes); ok {
 			if cs.privValidatorPubKey == nil {
 				return false, ErrPubKeyIsNotSet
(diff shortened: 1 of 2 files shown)


internal/consensus/state.go
```diff
@@ -2273,19 +2273,14 @@ func (cs *State) tryAddVote(vote *types.Vote, peerID p2p.ID) (bool, error) {
 			)
 
 			return added, err
-		} else if errors.Is(err, types.ErrVoteNonDeterministicSignature) {
-			cs.Logger.Info("Vote has non-deterministic signature", "err", err)
```
cason, line 2277: This will become an Error level message, see line 965.

internal/consensus/state.go
```diff
@@ -2273,19 +2273,14 @@ func (cs *State) tryAddVote(vote *types.Vote, peerID p2p.ID) (bool, error) {
...
-		} else if errors.Is(err, types.ErrInvalidVoteExtension) {
-			cs.Logger.Info("Vote has invalid extension")
```
cason, line 2279: This will become an Error level message, see line 965.

cason: This line, btw, is not covered by any test unit.

internal/consensus/state.go
cason: This line is covered by `TestByzantineConflictingProposalsWithPartition` test unit.

cason (approved):
Legit.

We are adding more clutter to the logs, in particular for users running with log level `error` for the consensus module. But this should not harm.

PR #4007, opened 2024-09-05
ci: remove archived technote-space/get-diff-action

melekes:
This repository has been archived by the owner on Nov 24, 2023. It is now read-only. See https://github.com/technote-space/get-diff-action

---

#### PR checklist

- [ ] Tests written/updated


.github/workflows/build.yml
```diff
@@ -32,18 +44,8 @@ jobs:
...
-          PATTERNS: |
-            **/*.go
-            "!test/"
-            go.mod
-            go.sum
-            Makefile
-
       - name: install
         run: GOOS=${{ matrix.goos }} GOARCH=${{ matrix.goarch }} make build
-        if: "env.GIT_DIFF != ''"
```
andynog, line 46: I'm just trying to understand the implication of removing this. Does it mean that the install will run all the time now even if the diff is null ? If so, isn't this will incur additional costs ? 

But at the same time looks like this is triggered only when changes are made (pull_request or push) so I think these `env.GIT_DIFF` were not very helpful.

andynog (approved):
lgtm, just added one question.

.github/workflows/build.yml
melekes: `env.GIT_DIFF` comes from https://github.com/technote-space/get-diff-action#usage. It was how the action used to work - computing `GIT_DIFF` and requiring this `if` condition to skip the given action.

The install will run only when the changes are made in `paths` (pull_request or push).

melekes:
NOTE: This requires removing the "Required" status from some jobs OR making it optional.

.github/workflows/build.yml
```diff
@@ -4,10 +4,22 @@ name: Build
...
+    paths:
+      - "**.go"
+      - go.mod
+      - go.sum
+      - Makefile
+      - "!test/**"
   merge_group:
   push:
     branches:
       - main
```
sergio-mena, line 16: I know this is not part of this PR but, shouldn't we be running this on every release branch as well?

sergio-mena (approved):
Thanks for this!

.github/workflows/build.yml
melekes: We do run the `build` workflow for every release branch (e.g., [v0.38.x](https://github.com/cometbft/cometbft/blob/v0.38.x/.github/workflows/build.yml#L9)

Issue #3977, opened 2024-09-03
mempool/lanes: Write lane-specific unit tests


hvanz:
Depends on #3978 and #3980.

- Allow kvstore to initialise with lanes (lanes = nil)
- Unit test scenarios with nil as lanes (in `clist_mempool_test.go`)

jmalicevic:
Closed with #3484 and #4008

PR #4008, opened 2024-09-05
feat(mempool/test): Add test to check order when lanes are disabled

jmalicevic:
Addresses  #3977 

---

#### PR checklist

- [x] Tests written/updated


hvanz:
This is the same test as `TestReapOrderMatchesGossipOrder` but with a different app. I think it should be possible to just have all under one test with multiple scenarios with app as parameter.

hvanz (approved):
👍

PR #4009, opened 2024-09-05
feat(consensus): proposer locks part set until all parts are added

cason:
Addresses #1742.

When a validator is the round proposer, it produces a `Proposal` and a number of block `Part`s that it then delivers to itself ([code](https://github.com/cometbft/cometbft/blob/ec2f3e2bc7eebe1dba69ef2a681c20fd973f332f/internal/consensus/state.go#L1258-L1271)). When a `Proposal` is accepted, the validator initializes a `types.PartSet` instance from the proposed `BlockID` ([code](https://github.com/cometbft/cometbft/blob/4952be833d1b1254f79f0e03b862f3adfba4e64f/internal/consensus/state.go#L2091-L2093)). Then, when each of the block `Part`s are delivered, they are added to this part set. Once a `Part` is added to the `cs.ProposalBlockParts` part set, it becomes available to the consensus reactor, that sends them to peers.

The issue reported in #1742 is that we add the block parts in order to `cs.ProposalBlockParts`, and one by one, even when the we are the proposer. When we are the proposer we know all the block parts and could produce `cs.ProposalBlockParts` immediately, instead of processing all `Part`s, one by one, as they were received from the network. This would have as a consequence that the consensus reactor will only start sending block `Part`s to peers once all of them are available. Since we randomly choose the next part to send to a peer, we are less likely to send to all peers the same part. That is what currently happens, as reported in #1742.

The solution proposed here is relatively simple and non-intrusive. To the `types.PartSet` type we add a boolean `locked` value. Once the proposer receives its own `Proposal`, it creates and _locks_ its `cs.ProposalBlockParts`, which is unlocked when all the expected parts are received and processed.

The consensus reactor, in turn, only send blocks `Part`s to peers when `cs.ProposalBlockParts` is unlocked. When the node is not the proposed of the round, it is always unlocked. When the node is the current round proposer, it is unlocked once all the parts it should contain are added, likely fixing the issue reported in  #1742.

---

#### PR checklist

- [x] Tests written/updated


cason:
Some my internal comments:

1. Why we need to protect the added field with a lock? I am not sure, some methods use a lock, others don't. For safety, I adopted locks.
2. Since it is a single boolean, why not use an atomic boolean? It should probably work as well, but probably requires some initialization.

melekes (approved):
👍

internal/consensus/state.go
```diff
@@ -2090,6 +2090,13 @@ func (cs *State) defaultSetProposal(proposal *types.Proposal, recvTime time.Time
 	// TODO: We can check if Proposal is for a different block as this is a sign of misbehavior!
 	if cs.ProposalBlockParts == nil {
 		cs.ProposalBlockParts = types.NewPartSetFromHeader(proposal.BlockID.PartSetHeader)
+
+		// If we are the proposer, lock the PartSet until we load all
+		// the block parts, that should come just after this Proposal.
+		// Addresses: https://github.com/cometbft/cometbft/issues/1742
```
sergio-mena, line 2096:
```suggestion
```
This info will be available in git blame anyway

internal/consensus/state.go
```diff
@@ -2090,6 +2090,13 @@ func (cs *State) defaultSetProposal(proposal *types.Proposal, recvTime time.Time
...
+		if bytes.Equal(pubKey.Bytes(), cs.privValidatorPubKey.Bytes()) {
```
sergio-mena, line 2097: I know you didn't introduce it, but, could you rename variable `pubKey` to `proposerPubKey` to make the code less confusing?

types/part_set.go
```diff
@@ -329,6 +334,24 @@ func (ps *PartSet) GetReader() io.Reader {
 	return NewPartSetReader(ps.parts)
 }
 
+func (ps *PartSet) Locked() bool {
```
sergio-mena, line 337:
```suggestion
func (ps *PartSet) IsLocked() bool {
```

types/part_set_test.go
```diff
@@ -55,13 +56,22 @@ func TestBasicPartSet(t *testing.T) {
...
 	// Reconstruct data, assert that they are equal.
 	data2Reader := partSet2.GetReader()
 	data2, err := io.ReadAll(data2Reader)
 	require.NoError(t, err)
 
 	assert.Equal(t, data, data2)
+
+	// Test locking
+	partSet2.Lock()
+	assert.True(t, partSet2.Locked())
```
sergio-mena, line 70:
```suggestion
	partSet2.Lock()
	assert.True(t, partSet2.Locked())
	partSet2.Lock()
	assert.True(t, partSet2.Locked())
```

sergio-mena (approved):
LGTM, thanks for this

internal/consensus/state.go
```diff
@@ -2090,9 +2090,15 @@ func (cs *State) defaultSetProposal(proposal *types.Proposal, recvTime time.Time
 	// TODO: We can check if Proposal is for a different block as this is a sign of misbehavior!
 	if cs.ProposalBlockParts == nil {
 		cs.ProposalBlockParts = types.NewPartSetFromHeader(proposal.BlockID.PartSetHeader)
+
+		// If we are the proposer, lock the PartSet until we load all
+		// the block parts, that should come just after this Proposal.
+		if bytes.Equal(proposerPubKey.Bytes(), cs.privValidatorPubKey.Bytes()) {
```
sergio-mena, line 2096:
```suggestion
		if cs.isProposer(cs.privValidatorPubKey.Address()) {
```

cason: I have accepted it then reverted it.

We have to test if the `Proposal` has been signed by us.

PR #4016, opened 2024-09-06
fix(e2e): viper errors with invalid `config` and `genesis` manifest keys

cason:
Addresses comments https://github.com/cometbft/cometbft/pull/3969#pullrequestreview-2277655582 and https://github.com/cometbft/cometbft/pull/3967#pullrequestreview-2277635023 on previous PRs.

The solution is in fact simple, as it is possible to tell viper to return an error if one of the configured entries (key/value pairs) were not applied to the destination structure, because they don't exist or are invalid.

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

test/e2e/pkg/testnet.go
```diff
@@ -826,3 +821,12 @@ func parseCsv(csvString string) ([][]string, error) {
...
 	return records, nil
 }
+
+func ParseKeyValueField(name string, field string) (key string, value string, err error) {
+	tokens := strings.Split(field, " = ")
+	if len(tokens) != 2 {
+		return key, value, fmt.Errorf("invalid '%s' field: \"%s\", "+
+			"expected \"key = value\"", name, field)
+	}
+	return tokens[0], tokens[1], nil
```
andynog, line 831:
```suggestion
	tokens := strings.Split(field, "=")
	if len(tokens) != 2 {
		return key, value, fmt.Errorf("invalid '%s' field: \"%s\", "+
			"expected \"key = value\"", name, field)
	}
	return strings.TrimSpace(tokens[0]), strings.TrimSpace(tokens[1]), nil
```

We should allow the `key=value` entries (without enforcing space between the '='), these would be valid and written as `key = value` in the `config.toml` file

Should also change in line 539
```
	for _, entry := range n.Config {
		_, _, err := ParseKeyValueField("config", entry)
		if err != nil {
			return err
		}
	}
```

andynog (approved):
lgtm, I've done some testing and other than 1 suggestion I think it's all good.

One thing I've noticed is that if there's an error generated because of the parsing errors in the config section of the manifest (e.g. invalid key), the testnet files are still generated (except the ones in the config folder). But that's a different scope and could be something to improve later.

Thanks !

PR #4040, opened 2024-09-09
fix(mempool): Propagate `CheckTx` errors to caller

hvanz:
Solves #1516

This PR changes `ReqRes`'s callback on the ABCI client to return an error. This error is then returned by the functions `handleCheckTxResponse` and `handleRecheckTxResponse`.  In case the tx is being added via an RPC endpoint, the error is propagated from `handleCheckTxResponse` up to the caller.

---

#### PR checklist

- [x] Tests written/updated


abci/client/client.go
```diff
@@ -115,18 +116,25 @@ func (r *ReqRes) InvokeCallback() {
...
+
 // GetCallback returns the configured callback of the ReqRes object which may be
 // nil. Note, it is not safe to concurrently call this in cases where it is
 // marked done and SetCallback is called before calling GetCallback as that
 // will invoke the callback twice and create a potential race condition.
 //
 // ref: https://github.com/tendermint/tendermint/issues/5439
-func (r *ReqRes) GetCallback() func(*types.Response) {
+//
+// Deprecated: not used.
```
melekes, line 136: Why not remove it?

melekes (approved):
👍

abci/client/client.go
hvanz: Wouldn't that be API breaking? If yes, I'd remove it in the next major release.

melekes: It is, but it's okay to break API in `v1` and `main`

hvanz: Ok, I removed it.

hvanz: Actually, since this is the only breaking part in this PR, I restored the method and created another PR to remove it #4083.

PR #4042, opened 2024-09-09
ci: add new workflow to notify users

melekes:
about breaking changes to v1.x

---

#### PR checklist

- [ ] Tests written/updated


.github/workflows/notify-about-breaking-changes.yml
```diff
@@ -0,0 +1,71 @@
...
+          echo "$CONTENT"
+          echo "::set-output name=file_content::$CONTENT"
+
+      - name: Post to a Slack channel
+        id: slack
+        if: steps.read-content.outputs.file_content != ''
+        uses: slackapi/slack-github-action@v1.27.0
+        with:
+          channel-id: 'C03Q5J9SXS8' # cometbft-engineering
+          # channel-id: 'shared-sdk-comet'
```
andynog, line 56: I am assuming that the sdk channel is commented out to test the engineering channel first and if it works we add to the sdk channel so all good

andynog (approved):
lgtm, nice job @melekes

Issue #4039, opened 2024-09-09
GO-2024-3107


melekes:
https://pkg.go.dev/vuln/GO-2024-3107

Vulnerable versions: before go1.22.7, from go1.23.0-0 before go1.23.1

```
Vulnerability #1: GO-2024-3107
    Stack exhaustion in Parse in go/build/constraint
  More info: https://pkg.go.dev/vuln/GO-2024-3107
  Standard library
    Found in: go/build/constraint@go1.22.5
    Fixed in: go/build/constraint@go1.22.7
    Example traces found:
Error:       #1: scripts/metricsgen/metricsgen.go:158:27: metricsgen.ParseMetricsDir calls parser.ParseDir, which eventually calls constraint.Parse

Vulnerability #2: GO-[20](https://github.com/cometbft/cometbft/actions/runs/10766490607/job/29852325945?pr=4025#step:6:21)24-3106
    Stack exhaustion in Decoder.Decode in encoding/gob
  More info: https://pkg.go.dev/vuln/GO-2024-3106
  Standard library
    Found in: encoding/gob@go1.[22](https://github.com/cometbft/cometbft/actions/runs/10766490607/job/29852325945?pr=4025#step:6:23).5
    Fixed in: encoding/gob@go1.22.7
    Example traces found:
Error:       #1: rpc/jsonrpc/server/http_server.go:262:15: server.defaultHandler.ServeHTTP calls http.ServeMux.ServeHTTP, which eventually calls gob.Decoder.Decode

Vulnerability #3: GO-20[24](https://github.com/cometbft/cometbft/actions/runs/10766490607/job/29852325945?pr=4025#step:6:25)-3105
    Stack exhaustion in all Parse functions in go/parser
  More info: https://pkg.go.dev/vuln/GO-2024-3105
  Standard library
    Found in: go/parser@go1.22.5
    Fixed in: go/parser@go1.22.7
    Example traces found:
Error:       #1: scripts/metricsgen/metricsgen.go:158:[27](https://github.com/cometbft/cometbft/actions/runs/10766490607/job/29852325945?pr=4025#step:6:28): metricsgen.ParseMetricsDir calls parser.ParseDir
Error:       #2: scripts/metricsgen/metricsgen.go:204:24: metricsgen.GenerateMetricsFile calls format.Source, which eventually calls parser.ParseFile
```

andynog:
https://github.com/cometbft/cometbft/pull/4046 should unblock at least the ci tests

PR #4046, opened 2024-09-09
build(deps): Bump the Go version to 1.22.7

cason:
Should temporarily solve #4039.

---

#### PR checklist

- [ ] Tests written/updated


go.mod
```diff
@@ -1,6 +1,9 @@
 module github.com/cometbft/cometbft
 
-go 1.22
+go 1.22.5
```
melekes, line 3:
```suggestion
go 1.22.7
```

melekes (approved):

cason:
We need a new version of `cometbft-db` to this to work, but in this case we are using v1.23 indirectly....

andynog:
If we cut a cometbft-db `v0.14.1` bumping the version I think it will unblock this. I've pushed a PR on `cometbft-db` for that

melekes:
@mergify backport v1.x

PR #4062, opened 2024-09-10
fix(mempool): encode tx.Hash() as %X in logs

altergui:
in vocdoni CI we check for log lines with "invalid chars", and we spotted some at cometbft logs.
```
comet: tx already received from peer sender=cb6a9612b39147cfc40e7514b334139004b7bf17 tx="W�:��2�{ڋUn-:�ы��ؖZWS��E]iE-"
```

you likely intended that tx= to be a hex string, so here's a proposed fix


andynog (approved):
lgtm thanks @altergui

PR #4081, opened 2024-09-11
chore: rename whitelist->allowlist, blacklist->denylist

altergui:
in statesync, adopt "rejectlist" instead
to match the wording used in that package

### DIFF
--- a/spec/abci/abci++_app_requirements.md
+++ b/spec/abci/abci++_app_requirements.md
@@ -1052,7 +1052,7 @@ can be spoofed by adversaries.
 Apps may also want to consider state sync denial-of-service vectors, where adversaries provide
 invalid or harmful snapshots to prevent nodes from joining the network. The application can
 counteract this by asking CometBFT to ban peers. As a last resort, node operators can use
-P2P configuration options to whitelist a set of trusted peers that can provide valid snapshots.
+P2P configuration options to list an exclusive set of trusted peers that can provide valid snapshots.
 
 ##### Transition to Consensus
 
--- a/p2p/pex/addrbook.go
+++ b/p2p/pex/addrbook.go
@@ -94,7 +94,7 @@ type addrBook struct {
 	ourAddrs   map[string]struct{}
 	privateIDs map[p2p.ID]struct{}
 	addrLookup map[p2p.ID]*knownAddress // new & old
-	badPeers   map[p2p.ID]*knownAddress // blacklisted peers
+	badPeers   map[p2p.ID]*knownAddress // banned peers
 	bucketsOld []map[string]*knownAddress
 	bucketsNew []map[string]*knownAddress
 	nOld       int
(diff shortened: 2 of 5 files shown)


andynog:
Hi @altergui ! Thank you for your PR and the effort you’ve put into improving the codebase. I noticed that the primary changes in this submission are focused on replacing terms like "whitelist" and "blacklist" with "allowlist" and "denylist."

Could you please provide some context or reasoning behind these semantic adjustments? I’d like to better understand the motivation, whether it’s related to technical considerations, or another goal.

docs/references/architecture/tendermint-core/adr-053-state-sync-prototype.md
```diff
@@ -211,15 +211,15 @@ Snapshots must also be garbage collected after some configurable time, e.g. by k
 
 * Do we need incremental chunk verification for first version?
 
-    > No, we'll start simple. Can add chunk verification via a new snapshot format without any breaking changes in Tendermint. For adversarial conditions, maybe consider support for whitelisting peers to download chunks from.
+    > No, we'll start simple. Can add chunk verification via a new snapshot format without any breaking changes in Tendermint. For adversarial conditions, maybe consider support for allowlisting peers to download chunks from.
```
hvanz, line 214: Tendermint Core docs are here for historical purposes. I think they should not be changed.

spec/abci/abci++_app_requirements.md
```diff
@@ -1052,7 +1052,7 @@ can be spoofed by adversaries.
 Apps may also want to consider state sync denial-of-service vectors, where adversaries provide
 invalid or harmful snapshots to prevent nodes from joining the network. The application can
 counteract this by asking CometBFT to ban peers. As a last resort, node operators can use
-P2P configuration options to whitelist a set of trusted peers that can provide valid snapshots.
+P2P configuration options to allowlist a set of trusted peers that can provide valid snapshots.
```
hvanz, line 1055: I don't think we have a styling guide for this case but, for reference, Google recommends to not use allowlist and denylist as verbs  https://developers.google.com/style/word-list#allowlist

altergui:
> Could you please provide some context or reasoning behind these semantic adjustments? I’d like to better understand the motivation, whether it’s related to technical considerations, or another goal.

sure, cf https://www.splunk.com/en_us/blog/learn/blacklist-whitelist-inclusivity.html

> In today’s rapidly changing world, continuing to make an effort to promote diversity and inclusion cannot be emphasized enough. [The language we use plays a vital role](https://www.splunk.com/en_us/blog/learn/inclusive-language-workplace.html) in shaping our attitudes towards various issues. In the world of technology and cybersecurity, terms such as “blacklist” and “whitelist” have been customary for decades, but it is essential that we recognize that these terms may mistakenly continue exclusionary connotations.

spec/abci/abci++_app_requirements.md
altergui: thanks a lot! i'll amend the commit

altergui: replaced by `list an exclusive`

andynog:
I understand the context much better now. Thanks for bringing that up.

It seems like it's perfectly fine to make changes to comments, specifications, and private variables without causing any breaking changes. With regards to the ADRs, as @hvanz pointed out, these are historical documents, and we usually keep them unchanged unless there's a specific reason to update them, such as a change in specifications or design.

PR #4083, opened 2024-09-11
fix(abci/client)!: Remove unused `ReqRes` method `GetCallback`

hvanz:
#### PR checklist

- [ ] Tests written/updated


hvanz:
Fixed in #4040

PR #4092, opened 2024-09-13
feat(mempool/lanes): Allow lanes with equal priorities

### LINKED ISSUE #4014: mempool/lanes: Allow lanes to have the same priority
Currently a lane is identified by its priority, so different lanes cannot have the same priority. This limitation was only made to simplify the prototype.

jmalicevic:
Closes #4014 .

The PR changes the proto files in `InfoResponse` to return a map of `LaneID(string) : LanePrio(uint32)` instead of an array of priorities and the default lane is identified by its ID (a string). 

CheckTx returns a `LaneID` (a string).

Before marking it ready for review, we need to make sure that the corner case of default lane and priorities are properly addressed now with these changes. 

---

#### PR checklist

- [x] Tests written/updated


proto/cometbft/abci/v1/types.proto
```diff
@@ -307,7 +316,7 @@ message CheckTxResponse {
   reserved 9 to 11;
   reserved "sender", "priority", "mempool_error";
 
-  uint32 lane = 12;
+  Lane lane = 12;
```
hvanz, line 319: Why return the priority? The mempool already has all priorities from the handshake.

mempool/clist_mempool.go
```diff
@@ -85,37 +84,39 @@ type CListMempoolOption func(*CListMempool)
 func NewCListMempool(
 	cfg *config.MempoolConfig,
 	proxyAppConn proxy.AppConnMempool,
-	lanesInfo *LanesInfo,
+	lanesInfo *LaneData,
```
hvanz, line 87: Also rename variable to `laneData`.

mempool/lanes_info.go
```diff
@@ -1,64 +1,62 @@
...
 // lanes given the data queried from the app.
-func BuildLanesInfo(laneList []uint32, defLane types.Lane) (*LanesInfo, error) {
-	lanes := make([]types.Lane, len(laneList))
-	for i, l := range laneList {
-		lanes[i] = types.Lane(l)
-	}
-	info := LanesInfo{lanes: lanes, defaultLane: defLane}
-	if err := info.validate(); err != nil {
+func BuildLanesInfo(laneList map[string]uint32, defLane string) (*LaneData, error) {
+	// lanes := make([]types.LaneID, len(laneList))
```
hvanz, line 15: remove comments?

proto/cometbft/abci/v1/types.proto
```diff
@@ -261,8 +261,9 @@ message InfoResponse {
   int64 last_block_height   = 4;
   bytes last_block_app_hash = 5;
 
-  repeated uint32 lane_priorities = 6;
-  uint32 default_lane_priority = 7;
+  map<string, uint32> lane_info = 6;
```
hvanz, line 264: I would keep the name `lane_priorities` to make clear what the map is about. Otherwise I would add a comment to clarify this line.

mempool/lanes_info.go
```diff
@@ -1,64 +1,62 @@
...
 import (
-	"slices"
-
 	"github.com/cometbft/cometbft/types"
 )
 
-type LanesInfo struct {
-	lanes       []types.Lane
-	defaultLane types.Lane
+type LaneData struct {
```
hvanz, line 7: With this renaming there are variable names, file names and comments that need to be updated because they are inconsistent. 

I think it makes sense that on the application and the protos we have lane "data" that is converted to lane "info" when imported into Comet, but the names (and all the rest) have to be consistent.

mempool/clist_mempool.go
```diff
@@ -52,19 +51,19 @@ type CListMempool struct {
...
-	addTxLaneSeqs map[types.Lane]int64 // Sequence of the last TX added to a given lane
+	addTxChMtx    cmtsync.RWMutex        // Protects the fields below
+	addTxCh       chan struct{}          // Blocks until the next TX is added
+	addTxSeq      int64                  // Helps detect is new TXs have been added to a given lane
+	addTxLaneSeqs map[types.LaneID]int64 // Sequence of the last TX added to a given lane
 
 	// Immutable fields, only set during initialization.
-	defaultLane types.Lane
+	defaultLane string
 	sortedLanes []types.Lane // lanes sorted by priority, in descending order
```
hvanz, line 67: Just realised that `sortedLanes` is used only by iterators so we could move it inside the iterators and create it when iterators are created. In any case, I would split this array in two: one with sorted lane ids and a map from lane id to priority, so then we can get rid of the `types.Lane` struct.

types/tx.go
```diff
@@ -197,4 +197,12 @@ func ComputeProtoSizeForTxs(txs []Tx) int64 {
 }
 
 // A lane is defined by its priority.
-type Lane uint32
+type LaneID string
+
+type Lane struct {
```
hvanz, line 202: I would try to get rid of this struct. I see it's only used by iterators.

hvanz:
Do we have any test scenario with multiple lanes having the same priority?

mempool/clist_mempool.go
jmalicevic: I liked this suggestion but when I was looking at removing it and having the sort happen every time we make an iterator, I am not sure it is the best way to go. The iterators are created a lot (on every reap etc.) and we'd have to sort every time. 

I also looked at removing the struct, I can do it but in many places we have to create two arrays that we have to iterator over in parallel instead of one. I personally find this cleaner, but if you feel strongly about it I can remove it?

test/loadtime/payload/payload.proto
```diff
@@ -15,5 +20,5 @@ message Payload {
   google.protobuf.Timestamp time        = 4;
   bytes                     id          = 5;
   bytes                     padding     = 6;
-  uint32                    lane        = 7;
+  Lane                    lane        = 7;
```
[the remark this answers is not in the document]
jmalicevic: We can do that but then we'd have to create a `struct Lane` in e2e or oterate over two arrays (similarly to what we would have to do for iterators). I don't like implicit correlation and order over two arrays but I dont mind moving the struct out of the proto file and keeping it only in the `.go` files.

hvanz (commented):
I made a second pass. I know it's a draft PR so you probably thought about making later some of the changes I'm suggesting.

test/e2e/app/app.go
```diff
@@ -60,8 +61,7 @@ type Application struct {
 	// It's OK not to persist this, as it is not part of the state machine
 	seenTxs sync.Map // cmttypes.TxKey -> uint64
 
-	lanes          map[string]uint32
-	lanePriorities []uint32
+	lanesInfo map[string]uint32
```
hvanz, line 64: Rename to `lanePriorities`.

test/e2e/app/app.go
```diff
@@ -161,7 +161,7 @@ func DefaultConfig(dir string) *Config {
 }
 
 // LaneDefinitions returns the (constant) list of lanes and their priorities.
-func LaneDefinitions(lanes map[string]uint32) (map[string]uint32, []uint32) {
+func LaneDefinitions(lanes map[string]uint32) (map[string]uint32, []payload.Lane) {
```
hvanz, line 164: This function could be split in two because its two results are used separately.

test/e2e/app/app.go
```diff
@@ -356,21 +363,21 @@ func (app *Application) CheckTx(_ context.Context, req *abci.CheckTxRequest) (*a
 		return &abci.CheckTxResponse{Code: kvstore.CodeTypeOK, GasWanted: 1}, nil
 	}
 	lane := extractLane(value)
-	return &abci.CheckTxResponse{Code: kvstore.CodeTypeOK, GasWanted: 1, Lane: lane}, nil
+	return &abci.CheckTxResponse{Code: kvstore.CodeTypeOK, GasWanted: 1, LaneId: lane.GetId()}, nil
 }
 
 // extractLane returns the lane field if value is a Payload, otherwise returns 0.
-func extractLane(value string) uint32 {
+func extractLane(value string) payload.Lane {
```
hvanz, line 370: Return lane id and update doc-comment.

hvanz (commented):
Third pass.

test/e2e/app/app.go
jmalicevic: I replaced it as as string because it is returned as such in checkTx and we'd have to convert it right back.

test/e2e/app/app.go
jmalicevic: I would leave them as is atm for the MVP and we can see down the line if we want to change this.

mempool/iterators_test.go
```diff
@@ -294,64 +296,119 @@ func TestBlockingIteratorsConsumeAllTxs(t *testing.T) {
 	}
 }
 
-// TODO automate the lane numbers so we can change the number of lanes
-// and increase the number of transactions.
+// Confirms that the transactions are returned in the same order
+// Note that for the cases with equal priorities the actual order
+// will depend on the way we iterate over the map of lanes
```
hvanz, line 301:
```suggestion
// Confirms that the transactions are returned in the same order.
// Note that for the cases with equal priorities the actual order
// will depend on the way we iterate over the map of lanes.
```

hvanz (approved):
🚀

abci/example/kvstore/kvstore.go
```diff
@@ -138,19 +138,18 @@ func (app *Application) Info(context.Context, *types.InfoRequest) (*types.InfoRe
 		}
 	}
 
-	var defaultLanePriority uint32
-	if len(app.lanes) > 0 {
-		defaultLanePriority = app.lanes[defaultLane]
+	defLane := ""
+	if len(app.lanes) != 0 {
+		defLane = "default"
```
melekes, line 143: Should this be a `const`?

mempool/clist_mempool.go
```diff
@@ -80,6 +79,14 @@ var _ Mempool = &CListMempool{}
 // CListMempoolOption sets an optional parameter on the mempool.
 type CListMempoolOption func(*CListMempool)
 
+// Lane corresponds to a transaction class as defined by the application.
+// A lane is identified by a string name and has priority level.
+// Different lanes can have the same priority.
+type Lane struct {
+	id       types.LaneID
```
melekes, line 86: Why is `Lane` defined in the `mempool` package and the following two types in the `types` package?

mempool/clist_mempool.go
```diff
@@ -80,6 +79,14 @@ var _ Mempool = &CListMempool{}
 // CListMempoolOption sets an optional parameter on the mempool.
 type CListMempoolOption func(*CListMempool)
 
+// Lane corresponds to a transaction class as defined by the application.
+// A lane is identified by a string name and has priority level.
+// Different lanes can have the same priority.
+type Lane struct {
```
melekes, line 85: Does `Lane` need to be public?

node/setup.go
```diff
@@ -280,7 +280,7 @@ func createMempoolAndMempoolReactor(
 	switch config.Mempool.Type {
 	// allow empty string for backward compatibility
 	case cfg.MempoolTypeFlood, "":
-		lanesInfo, err := mempl.BuildLanesInfo(appInfoResponse.LanePriorities, types.Lane(appInfoResponse.DefaultLanePriority))
+		lanesInfo, err := mempl.BuildLanesInfo(appInfoResponse.LanePriorities, types.LaneID(appInfoResponse.DefaultLane))
```
hvanz, line 283: I think `types.LaneID` should not be here. The second parameter to `BuildLanesInfo` should be changed to `string`.

abci/example/kvstore/kvstore.go
hvanz: Agree, it should be a const. It appears in many places.

mempool/clist_mempool.go
hvanz: Agree to move LaneID and LanePriority to the mempool package, probably to clist_mempool.go.

mempool/clist_mempool.go
hvanz: No. A lane is an implementation detail of CListMempool and it's not used outside of it.

PhelpsC43 (approved):

PR #4104, opened 2024-09-16
fix(bls12381): NewPublicKeyFromBytes: add missing infinity check

melekes:
Fixes a regression introduced in https://github.com/cometbft/cometbft/pull/3603


cason:
Any idea of why this happens?

```
cometbft/crypto/bls12381 $ go test                                                                                                                                                                                    
?   	github.com/cometbft/cometbft/crypto/bls12381	[no test files]
```

cason (commented):
Left some comments, but I couldn't run the tests locally.

melekes:
> Any idea of why this happens?
> 
> ```
> cometbft/crypto/bls12381 $ go test                                                                                                                                                                                    
> ?   	github.com/cometbft/cometbft/crypto/bls12381	[no test files]
> ```

you need to add `bls12381` build flag: `go test -tags=bls12381`

crypto/bls12381/key_test.go
```diff
@@ -167,3 +170,37 @@ func TestPubKey_MarshalJSON(t *testing.T) {
 	err = pubKey2.UnmarshalJSON(jsonBytes)
 	require.NoError(t, err)
 }
+
+func TestPubKey_NewPublicKeyFromBytes(t *testing.T) {
```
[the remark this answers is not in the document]
melekes: Mention how?

crypto/bls12381/key_test.go
```diff
@@ -172,6 +172,8 @@ func TestPubKey_MarshalJSON(t *testing.T) {
 }
 
 func TestPubKey_NewPublicKeyFromBytes(t *testing.T) {
```
cason, line 174:
```suggestion
func TestPubKey_NewPublicKeyFromInvalidBytes(t *testing.T) {
```

crypto/bls12381/key_test.go
```diff
@@ -167,3 +170,44 @@ func TestPubKey_MarshalJSON(t *testing.T) {
...
+		pkc := pk.Serialize()
+		if pkc == nil {
+			return nil, errors.New("could not serialize pubkey")
+		}
+		return pkc, nil
+	}
+
+	invalidPubKeys := map[string]string{
+		"NotInG1":   "8123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef",
+		"InfFalseB": "800000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
```
cason, line 197: This is a valid key, right? At least, the test passes with it even without the introduced changes.

crypto/bls12381/key_test.go
```diff
@@ -167,3 +170,44 @@ func TestPubKey_MarshalJSON(t *testing.T) {
...
+		"InfTrueB":  "c01000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
```
cason, line 198: The same as previous comment.

crypto/bls12381/key_test.go
cason: But now I am not sure. Are some of those keys valid?

cason (commented):
Could run the tests, thank you.

crypto/bls12381/key_test.go
melekes: I copied it from the report. I think it means "infinite point A". The other one is "infinite point B". A and B are two parts. 

> One of the initially non-obvious things about BLS12-381 is that we're really dealing with two curves, not one.
https://hackmd.io/@benjaminion/bls12-381

cason: But shouldn't it fail or not?

melekes: I still think it's helpful to keep them even though they don't yield an error on `main`.

cason (approved):
Legit.

cason:
But are we sure that the CI is running the added test? How does it know that it needs to add a `-tag` parameter?

cason:
> But are we sure that the CI is running the added test? How does it know that it needs to add a -tag parameter?

Ok, it appears that it does not fail in the CI, see the last commit.

melekes:
> > But are we sure that the CI is running the added test? How does it know that it needs to add a -tag parameter?
> 
> Ok, it appears that it does not fail in the CI, see the last commit.

Thanks for spotting it 👍 fixed in [`d527708` (#4104)](https://github.com/cometbft/cometbft/pull/4104/commits/d5277082185c3e25d003dc80317bddc8959f0466)

PR #4106, opened 2024-09-16
refactor(lanes): make `pickLane` non blocking

hvanz:
By making the `pickLane` function non blocking, the main goal of this refactoring is to remove the unlock/lock pattern of `addTxChMtx` in `pickLane`. This prepares the code to add a cancellable context to the blocking iterator (see #4107), which now we can add to `WaitNextCh` (around the line `<-addTxCh`), outside `pickLane`.

### DIFF
--- a/mempool/iterators.go
+++ b/mempool/iterators.go
@@ -123,12 +123,21 @@ func NewBlockingIterator(mem *CListMempool) Iterator {
 func (iter *BlockingIterator) WaitNextCh() <-chan Entry {
 	ch := make(chan Entry)
 	go func() {
-		// Add the next entry to the channel if not nil.
-		lane := iter.pickLane()
-		if entry := iter.next(lane); entry != nil {
-			ch <- entry.Value.(Entry)
+		var lane types.Lane
+		for {
+			l, addTxCh := iter.pickLane()
+			if addTxCh == nil {
+				lane = l
+				break
+			}
+			// There are no transactions to take from any lane. Wait until at
+			// least one is added to the mempool and try again.
+			<-addTxCh
 		}
-		// Unblock the receiver (it may receive nil).
+		if elem := iter.next(lane); elem != nil {
+			ch <- elem.Value.(Entry)
+		}
+		// Unblock receiver in case no entry was sent (it will receive nil).
 		close(ch)
 	}()
 	return ch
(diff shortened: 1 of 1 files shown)


mempool/iterators.go
```diff
@@ -156,12 +166,9 @@ func (iter *BlockingIterator) PickLane() types.Lane {
...
 			if numEmptyLanes >= len(iter.sortedLanes) {
-				// There are no lanes with non-accessed entries. Wait until a new tx is added.
-				ch := iter.mp.addTxCh
-				iter.mp.addTxChMtx.RUnlock()
-				<-ch
-				iter.mp.addTxChMtx.RLock()
-				numEmptyLanes = 0
+				// There are no lanes with non-accessed entries. Wait until a
+				// new tx is added.
+				return 0, iter.mp.addTxCh
```
hvanz, line 171: Now it returns without advancing the indexes. `pickLane` will be called again starting with the same indexes as when it returned here the previous time.

mempool/iterators.go
```diff
@@ -123,29 +123,39 @@ func NewBlockingIterator(mem *CListMempool) Iterator {
...
+			}
+			// There are no transactions to take from any lane. Wait until at
+			// least one is added to the mempool and try again.
+			<-addTxCh
+		}
+		if elem := iter.next(lane); elem != nil {
+			ch <- elem.Value.(Entry)
 		}
-		// Unblock the receiver (it may receive nil).
+		// Unblock receiver in case no entry was sent (it will receive nil).
```
sergio-mena, line 140:
```suggestion
		// Unblock receiver (in case no entry was sent, it will receive nil).
```

mempool/iterators.go
```diff
@@ -123,29 +123,39 @@ func NewBlockingIterator(mem *CListMempool) Iterator {
...
 		close(ch)
 	}()
 	return ch
 }
 
-// PickLane returns a _valid_ lane on which to iterate, according to the WRR
+// pickLane returns a _valid_ lane on which to iterate, according to the WRR
 // algorithm. A lane is valid if it is not empty or it is not over-consumed,
```
sergio-mena, line 147:
```suggestion
// algorithm. A lane is valid if it is not empty and it is not over-consumed,
```

mempool/iterators.go
```diff
@@ -175,17 +182,12 @@ func (iter *BlockingIterator) PickLane() types.Lane {
...
 }
 
-// Next returns the next element according to the IWRR algorithm.
-//
-// In classical WRR, the iterator cycles over the lanes. When a lane is selected, Next returns an
-// entry from the selected lane. On subsequent calls, Next will return the next entries from the
-// same lane until `lane` entries are accessed or the lane is empty, where `lane` is the priority.
-// The next time, Next will select the successive lane with lower priority.
-func (iter *BlockingIterator) Next(lane types.Lane) *clist.CElement {
+// next returns the next entry from the given lane and updates to WRR variables.
```
sergio-mena, line 189:
```suggestion
// next returns the next entry from the given lane and updates WRR variables.
```

sergio-mena (approved):
💜 this. Simple and perfect!

sergio-mena:
One question (just came to my mind after approving). Do we want to remove one of the mutexes and do everything with the other? (if you agree, probably for a different PR)

sergio-mena:
Also, would it be useful to include `TestBlockingIteratorsConsumeAll` (added in #3763) here?

hvanz:
> Do we want to remove one of the mutexes and do everything with the other? (if you agree, probably for a different PR)

If we use `txsMtx` also in the iterators, we would be adding more contention on that mutex. Since the current code with `addTxMtx` works as expected, and it's more performant, I'd leave it as it is.

hvanz:
> Also, would it be useful to include TestBlockingIteratorsConsumeAll (added in https://github.com/cometbft/cometbft/pull/3763) here?

This test was already merged into the feature branch in #4103.

PR #4058, opened 2024-09-10
fix(mempool): Pass cancellable context to iterator

hvanz:
Fixes #4057

---

#### PR checklist

- [ ] Tests written/updated
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments

PR #4107, opened 2024-09-16
fix(lanes): Pass cancellable context to blocking iterator

hvanz:
Depends on #4106; merge that one first.

This is equivalent to #4058 on the `main` branch. It was made on a different PR because there are too many differences in the code to be ported to the feature branch.

Besides the context, the iterator now also has a name which is used for debugging with concurrent iterators.

### DIFF
--- a/mempool/iterators.go
+++ b/mempool/iterators.go
@@ -7,37 +7,46 @@ import (
 	"github.com/cometbft/cometbft/types"
 )
 
-// WRRIterator is the base struct for implementing iterators that traverse lanes with
-// the classical Weighted Round Robin (WRR) algorithm.
-type WRRIterator struct {
+// IWRRIterator is the base struct for implementing iterators that traverse lanes with
+// the Interleaved Weighted Round Robin (WRR) algorithm.
+// https://en.wikipedia.org/wiki/Weighted_round_robin
+type IWRRIterator struct {
 	sortedLanes []types.Lane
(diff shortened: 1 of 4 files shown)


mempool/iterators.go
```diff
@@ -100,18 +101,22 @@ func (iter *NonBlockingIterator) Next() Entry {
...
 	iter := IWRRIterator{
 		sortedLanes: mem.sortedLanes,
 		cursors:     make(map[types.Lane]*clist.CElement, len(mem.sortedLanes)),
 		round:       1,
 	}
 	return &BlockingIterator{
 		IWRRIterator: iter,
+		ctx:          ctx,
 		mp:           mem,
+		name:         name,
```
jmalicevic, line 119: If the non blocking iterator also has a name, I would prefix this with "B" and the other one with "NB". But not mandatory.

jmalicevic (approved):

mempool/iterators.go
hvanz: I don't think we need a name for the non-blocking iterator. Usually there's only one active non-blocking iterator in the mempool (when reaping). When rechecking there's one for sending the CheckTx request and another for processing the responses. In all those cases, the mempool doesn't change. It's different for blocking iterators because there could be many acting concurrently and with a dynamic mempool (for example, the name is used in `TestBlockingIteratorsConsumeAllTxs`).

PR #4116, opened 2024-09-18
fix(bls12381): remove superflous hashing

melekes:
The blst library already does it. Pre-hashing introduces an additional attack vector.


sergio-mena (approved):
🙏

PR #4125, opened 2024-09-19
test(blocksync): increase capacity of errorsCh

melekes:
similar to how we set capacity to 1000 in production to avoid locking.

```go
const capacity = 1000                      // must be bigger than peers count
errorsCh := make(chan peerError, capacity) // so we don't block in #Receive#pool.AddBlock
```
https://github.com/cometbft/cometbft/blob/main/internal/blocksync/reactor.go#L99C1-L100C91


andynog (approved):
lgtm

melekes:
```
--- FAIL: TestBlockPoolMaliciousNode (11.60s)
    pool_test.go:360: error with peer bad: requested block #20 from [], not bad
    pool_test.go:71: 
        	Error Trace:	/home/runner/work/cometbft/cometbft/internal/blocksync/pool_test.go:71
        	            				/home/runner/work/cometbft/cometbft/internal/blocksync/pool_test.go:45
        	            				/opt/hostedtoolcache/go/1.23.1/x64/src/runtime/asm_amd64.s:[17](https://github.com/cometbft/cometbft/actions/runs/10954972836/job/30417932742#step:6:18)00
        	Error:      	Received unexpected error:
        	            	requested block #20 from [], not bad
        	Test:       	TestBlockPoolMaliciousNode
    pool_test.go:370: Pool caught up, malicious peer was banned at least once, start consensus.
```

PR #4127, opened 2024-09-19
test(mempool): enhanced `TestReactorNoBroadcastToSender`

cason:
The test should be able now to catch the error introduced by https://github.com/cometbft/cometbft/pull/3657.

----

To see how it would break, please checkout the branch `cason/test-nobroadcastsender`, then:

```
$ cd mempool
$ go test -v -run TestReactorNoBroadcastToSender
```

And wait it to fail, after 2 minutes. You are going to see the busy-loop around, logs entries starting with `Skipping tx`, forever for the same transaction.

Important: the test unit only works in the traditional mempool design, without lanes. We should consider adding a variant of it that should also pass when lanes are considered.

---

#### PR checklist

- [x] Tests written/updated


mempool/reactor.go
```diff
@@ -272,6 +272,8 @@ func (memR *Reactor) broadcastTxRoutine(peer p2p.Peer) {
 
 		// Do not send this transaction if we receive it from peer.
 		if entry.IsSender(peer.ID()) {
+			memR.Logger.Debug("Skipping transaction, peer is sender",
```
melekes, line 275: Do we need all these debug lines?

melekes: Keep in mind that every such line introduces an additional performance penalty.

cason: This is a good point, but currently we don't have any debug lines on this method, which is pretty important.

cason: What we can do here is to cache the costly part of this call, the `entry.Tx()` and re-use the value in all debug entries.

What do you think?

melekes: Use `log.NewLazySprintf`? Good idea 👍

melekes (approved):
👍

mempool/reactor.go
cason: Solved in https://github.com/cometbft/cometbft/pull/4127/commits/d3961f9d035e4d2c75535daf368d7e421fcf1260.

cason: However, we are still paying the cost of `tx.Hash()` even without debug.

This is the same as in the rest of this code. Namely, we aren't saving anything here.

hvanz:
Now that the mempool has lanes, the generated transactions will be assigned to different lanes, therefore they will arrive to the second node in a different order than FIFO. So one way to fix the test is to change the last line to only check that txs have arrived to the second reactor's mempool (with `checkTxsInMempool`) instead of checking that they have arrived and they have arrived in order (`checkTxsInOrder`).

If you still want to check that they have arrived in order, don't use lanes in this test. That will require more changes in the auxiliary functions because you will need to create the apps without lanes, with `app := kvstore.NewInMemoryApplicationWithoutLanes()` instead of `app := kvstore.NewInMemoryApplication()`.

hvanz (approved):
Thanks for improving the test! As mentioned elsewhere, it would be better to test also the case with more than one lane, but the test in this PR is already better than the original one.

PR #4131, opened 2024-09-19
feat(ci): codecov workflow

greg-szabo:
This is a PR to test codecov.

---

#### PR checklist

- [ ] Tests written/updated


greg-szabo:
I'm going to close this PR with "needs more work". Adding CodeCov to the CI is easy but our unit tests fail left-and-right. (Sporadically.)

We should only do this when we have enough confidence in our unit tests.

@sergio-mena @andynog @hvanz Tagging you here so the information doesn't get lost.

Issue #4133, opened 2024-09-20
libs/metrics: test failures


melekes:
Ok, the test units on this copied library are producing super random errors:

https://github.com/cometbft/cometbft/actions/runs/10944499019/job/30386514771?pr=4129

https://github.com/cometbft/cometbft/actions/runs/10944445079/job/30386330921?pr=4104

_Originally posted by @cason in https://github.com/cometbft/cometbft/issues/3989#issuecomment-2361405133_

andynog:
For this test https://github.com/cometbft/cometbft/actions/runs/10944499019/job/30386514771?pr=4129

Looks like the test has to spin up a prometheus server and use the response to execute the tests, I'm assuming that for some reason if the prometheus cannot be started the test should fail.
```
	s := httptest.NewServer(promhttp.HandlerFor(stdprometheus.DefaultGatherer, promhttp.HandlerOpts{}))
	defer s.Close()

	scrape := func() string {
		resp, _ := http.Get(s.URL)
		buf, _ := io.ReadAll(resp.Body)
		resp.Body.Close()
		return string(buf)
	}
```

I'll improve the test

andynog:
reopening, since I believe #4140 also contributes to fixing these problems

PR #4140, opened 2024-09-20
test(metrics/prometheus): refactor scrape in prometheus test

andynog:
close: #4133

### DIFF
--- a/libs/metrics/prometheus/prometheus_test.go
+++ b/libs/metrics/prometheus/prometheus_test.go
@@ -36,6 +36,7 @@ import (
 
 	stdprometheus "github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/stretchr/testify/assert"
 
 	"github.com/cometbft/cometbft/libs/metrics/teststat"
 )
@@ -43,13 +44,6 @@ import (
 func TestCounter(t *testing.T) {
 	s := httptest.NewServer(promhttp.HandlerFor(stdprometheus.DefaultGatherer, promhttp.HandlerOpts{}))
 	defer s.Close()
-
-	scrape := func() string {
-		resp, _ := http.Get(s.URL)
-		buf, _ := io.ReadAll(resp.Body)
-		resp.Body.Close()
-		return string(buf)
-	}
 
 	namespace, subsystem, name := "ns", "ss", "foo"
 	re := regexp.MustCompile(namespace + `_` + subsystem + `_` + name + `{alpha="alpha-value",beta="beta-value"} ([0-9\.]+)`)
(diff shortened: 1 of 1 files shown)


libs/metrics/prometheus/prometheus_test.go
```diff
@@ -248,3 +226,22 @@ func TestInconsistentLabelCardinality(t *testing.T) {
 		"a", "1", "b", "2", "c", "KABOOM!",
 	).Add(123)
 }
+
+func spinUpPrometheusServer() *httptest.Server {
+	s := httptest.NewServer(promhttp.HandlerFor(stdprometheus.DefaultGatherer, promhttp.HandlerOpts{}))
+
+	// minimal delay to allow the prometheus server come up with results and avoid errors during test
```
melekes, line 233: What results? At this point, we have not registered any metrics.

andynog: I think you are asking only about the comment right ? I meant just for the server to be up, not results, I can remove that from the comment

libs/metrics/prometheus/prometheus_test.go
```diff
@@ -37,21 +37,16 @@ import (
...
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/stretchr/testify/require"
 
 	"github.com/cometbft/cometbft/libs/metrics/teststat"
 )
 
 func TestCounter(t *testing.T) {
-	s := httptest.NewServer(promhttp.HandlerFor(stdprometheus.DefaultGatherer, promhttp.HandlerOpts{}))
+	s := spinUpPrometheusServer()
+	require.NotNil(t, s)
```
melekes, line 47: I don't think this is necessary.

libs/metrics/prometheus/prometheus_test.go
```diff
@@ -82,16 +77,10 @@ func TestCounter(t *testing.T) {
 }
 
 func TestGauge(t *testing.T) {
-	s := httptest.NewServer(promhttp.HandlerFor(stdprometheus.DefaultGatherer, promhttp.HandlerOpts{}))
+	s := spinUpPrometheusServer()
+	require.NotNil(t, s)
```
melekes, line 81: same

libs/metrics/prometheus/prometheus_test.go
```diff
@@ -159,16 +143,10 @@ func TestHistogram(t *testing.T) {
 	// limit. That is, the count monotonically increases over the buckets. This
 	// requires a different strategy to test.
 
-	s := httptest.NewServer(promhttp.HandlerFor(stdprometheus.DefaultGatherer, promhttp.HandlerOpts{}))
+	s := spinUpPrometheusServer()
+	require.NotNil(t, s)
```
melekes, line 147: same

melekes (approved):

Issue #2458, opened 2024-02-27
e2e must be able to test rollback command


melekes:
Follow-up to https://github.com/cometbft/cometbft/pull/2136

Sergio: "I still feel the need to test this via blackbox techniques, as the UTs are mocking a lot of things.
I'm approving this PR now, and will try to drive https://github.com/cometbft/cometbft/pull/2283 to completion so that we can test the rollback command in CI (e2e) with constant changes in validator sets and consensus params."

PR #2283, opened 2024-02-09
feat(e2e): introduce oscillations for validators and consensus params

sergio-mena:
Partially addresses #2458

This PR adds a new config to the `e2e` manifest that causes the `e2e` app to flip, at every height:

* The consensus param, by flipping between to valid lists of key types
* The validator set, by picking one validator, and flipping its power between 0 and 1 in alternate heights

This allows us to simulate chains with changes to these data structures at every height.

IMPORTANT: Do _not_ review commit by commit, as I radically changed the approach to flipping validators (to another way more surgical)

---

#### PR checklist

- [x] Tests written/updated
- ~[ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)~
- [x] Updated relevant documentation (`docs/` or `spec/`) and code comments

PR #4141, opened 2024-09-20
refactor(e2e): change `*Node` to string in keys of some maps

sergio-mena:
Partially addresses #2458

Change the `*Node` as keys of some maps in `e2e` and replace them by the node's name.

Having pointers as keys of a map brings about subtleties (how comparison is done) that we don't need. Code is much clearer with node names, even if it takes a bit longer to execute some loops (we need to look up the nodes). But this is test code, so clarity and robustness comes before performance of the testing code.

I decided to tackle this refactor in an different PR, rather than piggybacking it to #2283

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/test/e2e/pkg/testnet.go
+++ b/test/e2e/pkg/testnet.go
@@ -74,14 +74,13 @@ const (
 // Testnet represents a single testnet.
 // It includes all fields from the associated Manifest instance.
 type Testnet struct {
-	Manifest
+	*Manifest
 
 	Name string
 	File string
 	Dir  string
 
 	IP               *net.IPNet
-	Validators       map[string]int64
 	ValidatorUpdates map[int64]map[string]int64
 	Nodes            []*Node
 }
@@ -141,14 +140,13 @@ func NewTestnetFromManifest(manifest Manifest, file string, ifd InfrastructureDa
 	}
 
 	testnet := &Testnet{
-		Manifest: manifest,
+		Manifest: &manifest,
 
 		Name: filepath.Base(dir),
 		File: file,
 		Dir:  dir,
 
 		IP:               ipNet,
-		Validators:       map[string]int64{},
 		ValidatorUpdates: map[int64]map[string]int64{},
 		Nodes:            []*Node{},
 	}
@@ -283,20 +281,14 @@ func NewTestnetFromManifest(manifest Manifest, file string, ifd InfrastructureDa
 	}
 
 	// Set up genesis validators. If not specified explicitly, use all validator nodes.
-	if manifest.ValidatorsMap != nil {
-		for validatorName, power := range *manifest.ValidatorsMap {
-			validator := testnet.LookupNode(validatorName)
-			if validator == nil {
-				return nil, fmt.Errorf("unknown validator %q", validatorName)
-			}
-			testnet.Validators[validator.Name] = power
-		}
-	} else {
+	if manifest.ValidatorsMap == nil {
+		validatorsMap := make(map[string]int64)
 		for _, node := range testnet.Nodes {
 			if node.Mode == ModeValidator {
-				testnet.Validators[node.Name] = 100
+				validatorsMap[node.Name] = 100
 			}
 		}
+		manifest.ValidatorsMap = &validatorsMap
 	}
 
 	// Set up validator updates.
--- a/test/e2e/pkg/testnet.go
+++ b/test/e2e/pkg/testnet.go
@@ -81,8 +81,8 @@ type Testnet struct {
 	Dir  string
 
 	IP               *net.IPNet
-	Validators       map[*Node]int64
-	ValidatorUpdates map[int64]map[*Node]int64
+	Validators       map[string]int64
+	ValidatorUpdates map[int64]map[string]int64
 	Nodes            []*Node
 }
 
@@ -148,8 +148,8 @@ func NewTestnetFromManifest(manifest Manifest, file string, ifd InfrastructureDa
 		Dir:  dir,
 
 		IP:               ipNet,
-		Validators:       map[*Node]int64{},
-		ValidatorUpdates: map[int64]map[*Node]int64{},
+		Validators:       map[string]int64{},
+		ValidatorUpdates: map[int64]map[string]int64{},
 		Nodes:            []*Node{},
 	}
 	if testnet.InitialHeight == 0 {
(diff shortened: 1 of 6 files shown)


test/e2e/tests/validator_test.go
```diff
@@ -178,10 +178,11 @@ func (s *validatorSchedule) Increment(heights int64) {
...
 }
 
-func makeVals(valMap map[*e2e.Node]int64) []*types.Validator {
+func makeVals(testnet *e2e.Testnet, valMap map[string]int64) []*types.Validator {
 	vals := make([]*types.Validator, 0, len(valMap))
-	for node, power := range valMap {
-		vals = append(vals, types.NewValidator(node.PrivvalKey.PubKey(), power))
+	for valName, power := range valMap {
+		validator := testnet.LookupNode(valName)
+		vals = append(vals, types.NewValidator(validator.PrivvalKey.PubKey(), power))
```
andynog, line 185: the `LookupNode` can return `nil`, should test for that and only append in that case ? not sure if there are corner cases that can make it return `nil` but just highlighting the fact here, but it would panic if `validator` is `nil` too

andynog (approved):
lgtm, had one comment but I think it is irrelevant to the current logic so feel free to disregard.

melekes (approved):

test/e2e/tests/validator_test.go
alesforz: This applies to other parts of the refactorings above as well.

test/e2e/pkg/testnet.go
```diff
@@ -81,8 +81,8 @@ type Testnet struct {
 	Dir  string
 
 	IP               *net.IPNet
-	Validators       map[*Node]int64
-	ValidatorUpdates map[int64]map[*Node]int64
+	Validators       map[string]int64
```
cason, line 84: Since the type of `Validators` here is now identical to in `pkg/manifest.go` we don't even need to keep it here.

Namely, you could only rename `ValidatorsMap` to `Validators` in `Manifest` and remove this field from here.

test/e2e/pkg/testnet.go
```diff
@@ -81,8 +81,8 @@ type Testnet struct {
...
+	ValidatorUpdates map[int64]map[string]int64
```
cason, line 85: Same comment as above.

test/e2e/pkg/testnet.go
cason: See https://github.com/cometbft/cometbft/pull/3964 for more details.

cason (commented):
Following https://github.com/cometbft/cometbft/pull/3964, we should remove the two refactored fields from `Testnet` type, since they match the version on `Manifest`.

test/e2e/tests/validator_test.go
sergio-mena: Good point folks. I'll revisit the PR with this in mind

test/e2e/pkg/testnet.go
sergio-mena: `manifest.ValidatorUpdatesMap` is of type `map[string]map[string]int64`, whereas `testnet.ValidatorUpdates` is of type`map[int64]map[string]int64`.
I'm sure there's a way to turn `manifest.ValidatorUpdatesMap` into the same type as `testnet.ValidatorUpdates`, and thus get rid of the latter. But as this is not trivial, I'd rather leave it out of this PR.

test/e2e/tests/validator_test.go
sergio-mena: Addressed

test/e2e/pkg/testnet.go
sergio-mena: Addressed

cason: We can rename it back to `Validators`, see https://github.com/cometbft/cometbft/pull/4166

test/e2e/pkg/testnet.go
```diff
@@ -74,15 +74,14 @@ const (
 // Testnet represents a single testnet.
 // It includes all fields from the associated Manifest instance.
 type Testnet struct {
-	Manifest
+	*Manifest
```
cason, line 77: Are you sure this doesn't break anything?

test/e2e/runner/setup.go
```diff
@@ -150,9 +150,13 @@ func MakeGenesis(testnet *e2e.Testnet) (types.GenesisDoc, error) {
 	if testnet.PbtsUpdateHeight == -1 {
 		genesis.ConsensusParams.Feature.PbtsEnableHeight = testnet.PbtsEnableHeight
 	}
-	for validator, power := range testnet.Validators {
+	for valName, power := range *testnet.Manifest.ValidatorsMap {
+		validator := testnet.LookupNode(valName)
+		if validator == nil {
+			return types.GenesisDoc{}, fmt.Errorf("unknown validator %q for genesis doc", valName)
```
cason, line 156: Should never happen, right?

test/e2e/runner/setup.go
```diff
@@ -425,8 +432,12 @@ func MakeAppConfig(node *e2e.Node) ([]byte, error) {
 		validatorUpdates := map[string]map[string]int64{}
 		for height, validators := range node.Testnet.ValidatorUpdates {
 			updateVals := map[string]int64{}
-			for node, power := range validators {
-				updateVals[base64.StdEncoding.EncodeToString(node.PrivvalKey.PubKey().Bytes())] = power
+			for valName, power := range validators {
+				validator := node.Testnet.LookupNode(valName)
+				if validator == nil {
+					return nil, fmt.Errorf("unknown validator %q for validator updates in testnet, height %d", valName, height)
```
cason, line 438: Should never happen too, should we panic?

cason (approved):
Legit.

cason:
Consider merging https://github.com/cometbft/cometbft/pull/4166 into this.

test/e2e/pkg/testnet.go
sergio-mena: It was breaking when not used as a pointer, because we would copy some fields (non-pointer fields) but not others. A total mess.
As we are now modifying the Manifest down below in this function, those modifications didn't propagate (hence I needed to turn it into a pointer)

test/e2e/runner/setup.go
sergio-mena: Yes, but as suggested by @alesforz and @andynog , it's cleaner code if we check for `nil`.

test/e2e/runner/setup.go
sergio-mena: I'd rather return the error and have the callsite(s) decide whether they want to panic or not

PR #4147, opened 2024-09-21
docs(docker): fix links

orionpulse6:
These two external docs links seems to be broken due to the architecture change.


orionpulse6:
@melekes @andynog  PLZ have a check

orionpulse6:
GM! Is there anything I need to fix here? @andynog

DOCKER/README.md
cason, line 1: We probably should point to the `v1.x` docs here.

melekes (approved):
👍

PR #4164, opened 2024-09-23
revert(goleveldb): remove `goleveldb` flag

melekes:
https://github.com/cometbft/cometbft-db/releases/tag/v1.0.1


UPGRADING.md
```diff
@@ -227,14 +227,6 @@ definitions:
 
 - `consensus.skip_timeout_commit` has been removed in favor of `consensus.timeout_commit=0s`.
 
-### Database
-
-- The default database has been changed from `goleveldb` to
```
jmalicevic, line 232: If golevelDB still the default?

melekes: No. You think we should still mention this in UPGRADING?

andynog (approved):
lgtm, I've done some e2e tests with different DBs and seems fine. The only suggestion is to include something in the upgrading doc

Issue #4115, opened 2024-09-18
e2e: `run-multiple.sh` script should save the dump of failed tests


cason:
This is a helper script to run multiple executions, one per manifest.

If an execution fails, it dumps some data to the standard output and then cleans the directory with the actual execution data:

https://github.com/cometbft/cometbft/blob/276996ad958475b69727be2c57d4d0d818849a55/test/e2e/run-multiple.sh#L24-L35

It should instead create a temporary directory with the information of the failed execution, containing:
   - The `network/${manifest%.toml}` execution directory
   - The log of the current execution (`log` action)
   - It would be great to have the logs of every node that was part of the execution (as in #4113)

andynog:
I've already implemented this logic, it's the concept of `runner` executions and I was using it to better test the `e2e` nightly failures.

https://github.com/cometbft/cometbft/tree/andy/e2e-preserve-logs

you can run with "executions" enabled

```
./build/runner -f networks/nightly/gen-group01-0004.toml --execution
I[2024-09-20|13:53:42.069] Starting initial network nodes...            
I[2024-09-20|13:53:42.070] load                                         msg="Starting transaction load (16 workers)..."
....
I[2024-09-20|13:55:17.123] wait until                                   msg="Waiting for all nodes to reach height 51..."
I[2024-09-20|13:55:26.540] Running all tests in ./tests/...             
ok      github.com/cometbft/cometbft/test/e2e/tests     46.244s
I[2024-09-20|13:56:15.085] saving execution                             msg="saving e2e network execution information"

```

and it will save all the logs and manifest used in each run (could add logic to save the node data/config too), the nice thing you can run the same manifest multiple times, maybe modifying a parameter, and still keep the old results because it saves each run within a time date folder

```
./networks_executions/gen-group01-0004
└── 20240920_095615
    ├── manifest.toml
    └── nodes
        ├── validator01
        │   ├── docker-errors.log
        │   └── docker.log
        ├── validator02
        │   ├── docker-errors.log
        │   └── docker.log
        ├── validator03
        │   ├── docker-errors.log
        │   └── docker.log
        └── validator04
            ├── docker-errors.log
            └── docker.log
```

This logic can be expanded on many fronts, like some ideas parse the results and generate automated reports in order to improve the time it currently takes to comb through the results or eventually do diffs of results based on tweaks between runs 😉

cason:
https://github.com/cometbft/cometbft/pull/4165#issuecomment-2371801702

[...]

PR #4165, opened 2024-09-23
test(e2e): option to save runner executions

andynog:
close: #4115 

Add an option to the runner to save executions. Specify `--execution` or `-e` after the runner command to save the manifest, logs, config, data, etc.

```
./build/runner -f networks/simple.toml --execution
```

or if using `run-multiple.sh` specify a `SAVE_EXECUTION` variable:

```
SAVE_EXECUTION=1 ./run-multiple.sh networks/nightly/gen-group00-000*.toml 
```

This is a sample result, files will be in the `networks_executions` folder:

```
./networks_executions
├── gen-group00-0000
│   └── 20240923_180722
│       ├── manifest.toml
│       ├── nodes
│       │   └── validator01
│       │       ├── config
│       │       ├── data
│       │       ├── docker-errors.log
│       │       └── docker.log
│       ├── prometheus.yaml
│       └── zones.csv
├── gen-group00-0001
│   └── 20240923_180830
│       ├── manifest.toml
│       ├── nodes
│       │   └── validator01
│       │       ├── config
│       │       ├── data
│       │       ├── docker-errors.log
│       │       └── docker.log
│       ├── prometheus.yaml
│       └── zones.csv
└── gen-group00-0002
    └── 20240923_181033
        ├── manifest.toml
        ├── nodes
        │   ├── validator01
        │   │   ├── config
        │   │   ├── data
        │   │   ├── docker-errors.log
        │   │   └── docker.log
        │   ├── validator02
        │   │   ├── config
        │   │   ├── data
        │   │   ├── docker-errors.log
        │   │   └── docker.log
        │   ├── validator03
        │   │   ├── config
        │   │   ├── data
        │   │   ├── docker-errors.log
        │   │   └── docker.log
        │   └── validator04
        │       ├── config
        │       ├── data
        │       ├── docker-errors.log
        │       └── docker.log
        ├── prometheus.yaml
        └── zones.csv
```

---

#### PR checklist

- [ ] Tests written/updated


test/e2e/run-multiple.sh
```diff
@@ -17,11 +17,19 @@ fi
...
 	START=$SECONDS
 	echo "==> Running testnet: $MANIFEST"
 
-	if ! ./build/runner -f "$MANIFEST"; then
+	if [ -n "${SAVE_EXECUTION+set}" ]; then
+		EXECUTION_FLAG="--execution"
+		echo "==> Execution flag detected, saving files after run"
+	fi
+
+	if ! ./build/runner -f "$MANIFEST" "$EXECUTION_FLAG"; then
```
cason, line 32: Notice that we still dumps the manifest and logs to stdout. But I think we need that for the CI, right?

Maybe if `SAVE_EXECUTION` is set, we could skip the dumping?

test/e2e/runner/execution.go
```diff
@@ -0,0 +1,202 @@
...
+				strippedLine := strings.TrimSpace(parts[1])
+
+				// Write the stripped line to the docker log file
+				_, err := writer.WriteString(strippedLine + "\n")
+				if err != nil {
+					logger.Error("error saving execution", "msg", "error writing to log file", "file", logFile, "err", err.Error())
+					return err
+				}
+
+				// If the line is for an error, write the stripped line to the docker log error file
```
cason, line 90: So Comet's error messages are written to both files?

cason (approved):
Thank you for that. For some reason, I thought the logic would be simpler, but apparently you are covering all the corner cases :)

cason:
Ok, I will just test it locally and report what I find here. Code-wise, looks great.

test/e2e/runner/execution.go
```diff
@@ -0,0 +1,202 @@
...
+
+		// Get the logs from the Docker container
+		data, err := docker.ExecComposeOutput(context.Background(), testnet.Dir, "logs", node.Name)
+		if err != nil {
+			logger.Error("error saving execution", "msg", "error getting logs from container", "node", node.Name, "err", err.Error())
+			return err
+		}
+
+		// Create a file to write the processed lines
+		nodeFolder := filepath.Join(executionFolder, "nodes", node.Name)
```
cason, line 44: Do we need this `nodes` sub-directory?

test/e2e/README.md
```diff
@@ -71,6 +71,12 @@ Multiple testnets can be run with the `run-multiple.sh` script:
 ./run-multiple.sh networks/generated/gen-group3-*.toml
 ```
 
+To save the execution files specify the `SAVE_EXECUTION` variable:
+
+```sh
+ SAVE_EXECUTION=1 ./run-multiple.sh networks/generated/gen-group3-*.toml
+ ```
+
```
cason, line 79: I would add here a section to the introduced section below, which explains what is saved, how, and where.

test/e2e/README.md
```diff
@@ -174,6 +180,17 @@ func init() {
...
+If you need to save the files used in a runner execution, specify the `--execution` or `-e` flag, which will save the
+node files, manifest, and any other asset used during the test. It will also save the docker log files for each node.
+Each execution is saved under a `network_executions` folder, and at each run, a subfolder with a time and date timestamp
+will be created as the name of the folder. This way, you can test changing parameters in the manifest and later check
+the logs to see how that affected the run behavior and tests.
+
+```
+./build/runner -f networks/simple --execution
+```
```
cason, line 193: I would add here a comment mentioning that the saved executions can get pretty large in some cases, so to use this with care.

cason: This is specially relevant to the `data` dirs of each CometBFT home directory, but it can also be the case of logs, in particular when `debug` level is used.

test/e2e/run-multiple.sh
andynog: I left it there because on CI we need the dumps so we can download the logs from the Github Actions. This variable was introduced to run this locally. I'm not sure if we can get the assets from the Github Action node runners so specifying this variable would not make sense, this is more for local testing for now.

andynog:
> I wonder whether we shouldn't consider to have an additional command to the runner to dump all the data. I am considering here, in particular, when we use the several steps/actions of or infra, instead of running a full execution (i.e., when we pass a command to `build/runner -f <manifest> <command>`).

you mean a `runner -f <manifest> execute` ? instead of passing a flag `--execution` ?

test/e2e/runner/execution.go
andynog: maybe not, I was wondering that, maybe if I remove it there might be a possibility you can start `docker-compose` from the execution run directory and since the config and data folders are in there, maybe there's a way to "re-start" the chain for inspection `cometbft inspection` so you could have access to the RPC for retrieving data from the chain (e.g. `/block`)

test/e2e/README.md
andynog: OK, I'll add that

cason:
> you mean a `runner -f <manifest> execute` ? instead of passing a flag `--execution` ?

Yes, although `execution` is not a great name, I'd say `dump` or `save` or something?

Moreover, I'd maintain the `--execution` flag to when we run the "full" test. 

Recalling that the "full" tests means: `setup`, `start`, `load`, `perturb`, `wait`, `test`, `stop`, and `cleanup`. To save data we should call this new method between `stop` and `cleanup`.

test/e2e/run-multiple.sh
cason: My suggestion is that the code inside this `if` is kind of irrelevant when `-e` is passed. Because it will print the manifest, dump all logs, then cleanup. The former two are already performed by the new code, storing everything inside the `networks_executions/` dir.

andynog:
> Recalling that the "full" tests means: `setup`, `start`, `load`, `perturb`, `wait`, `test`, `stop`, and `cleanup`. To save data we should call this new method between `stop` and `cleanup`.

I has to be before stop, because the docker logs are dumped because the `stop` command does a `docker-compose down` and all the instances are gone after the stop. The way is currently is done, is that the containers are paused and a dump of the logs for each instance is taken, this works like a snapshot of the logs. Then the `stop` command can run

andynog:
I'm going to switch this to a draft PR for now until we have some decision and agreement on implementation details and I can tackle them.

alesforz:
> > Recalling that the "full" tests means: `setup`, `start`, `load`, `perturb`, `wait`, `test`, `stop`, and `cleanup`. To save data we should call this new method between `stop` and `cleanup`.
> 
> I has to be before `stop`, because the docker logs are dumped because the `stop` command does a `docker-compose down` and all the instances are gone after the stop. The way is currently is done, is that the containers are paused and a dump of the logs for each instance is taken, this works like a snapshot of the logs. Then the `stop` command can run

I agree with adding a `save` command to be optionally executed before `stop`. It would make the troubleshooting in e2e much less painful imho 😄

If we add it, we probably want to execute it also when we run the full "automated" test with:
```bash
./build/runner -f <manifest>
```

andynog:
Blocked until it is refactored as per comments and suggestions above.

andynog:
There are some design considerations that needs to be addressed before introducing this logic. This PR approach to solve the problem might not be the optimal way or it would need considerable refactoring. So closing this PR and once the design considerations are addressed in #4115 a new PR can be opened with the refactored/new solution.

PR #4169, opened 2024-09-24
fix(lanes): Panic when lane is not found

hvanz:
There is always at least one lane, so it should not happen that a lane is not found. This has cost me some debugging time.


mempool/clist_mempool.go
```diff
@@ -291,7 +291,7 @@ func (mem *CListMempool) LaneSizes(lane LaneID) (numTxs int, bytes int64) {
 
 	txs, ok := mem.lanes[lane]
 	if !ok {
-		return 0, bytes
+		panic(fmt.Errorf("lane %s not found while fetching lane size", lane))
```
jmalicevic, line 294: Can it happen that we give it the wrong lane?

hvanz: That would be a bug. And if we pass a wrong but existing lane, it won't panic.

jmalicevic (approved):

PR #4175, opened 2024-09-24
feat(e2e): Add `load_max_seconds` to manifest + optional flags to `load` command

hvanz:
A set of improvements to E2E for sending load from the command line. This is needed in DigitalOcean settings to send load with a lane field in the `Payload`, which is not supported by `test/loadtime/cmd/load`. In the future we may able to replace this tool by `runner load`.

Main commits:
- [Add load_max_seconds to manifest](https://github.com/cometbft/cometbft/commit/f1f317c8bad60a691bfb41bbed96d0fdd0c0d553)
- [Add rate, size, conn, time flags to load command](https://github.com/cometbft/cometbft/commit/e76476e21b4d1ee78ecd3518e9da2537de21f1ad)
- [Add nodes flag to load command, for a list of target node names to send the load](https://github.com/cometbft/cometbft/commit/ea4d60966b4deda40892a36ac68026c45bc099a8)


melekes (approved):
👍

PR #4177, opened 2024-09-25
feat(e2e): `Manifest.Validators` is not a pointer to a map, but a map

cason:
As per title.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/test/e2e/pkg/testnet.go
+++ b/test/e2e/pkg/testnet.go
@@ -281,14 +281,12 @@ func NewTestnetFromManifest(manifest Manifest, file string, ifd InfrastructureDa
 	}
 
 	// Set up genesis validators. If not specified explicitly, use all validator nodes.
-	if testnet.Validators == nil || len(*testnet.Validators) == 0 {
-		validatorsMap := make(map[string]int64)
+	if len(testnet.Validators) == 0 {
 		for _, node := range testnet.Nodes {
 			if node.Mode == ModeValidator {
-				validatorsMap[node.Name] = 100
+				testnet.Validators[node.Name] = 100
 			}
 		}
-		testnet.Validators = &validatorsMap
 	}
 
 	// Set up validator updates.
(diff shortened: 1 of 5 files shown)


sergio-mena (approved):
🎸

test/e2e/pkg/testnet.go
```diff
@@ -281,14 +281,12 @@ func NewTestnetFromManifest(manifest Manifest, file string, ifd InfrastructureDa
 	}
 
 	// Set up genesis validators. If not specified explicitly, use all validator nodes.
-	if testnet.Validators == nil || len(*testnet.Validators) == 0 {
-		validatorsMap := make(map[string]int64)
```
alesforz, line 285: given that the `validatorsMap[node.Name] = 100` causes an "assignment to nil map" panic, I think we must keep the  nil check + initialization if needed.

test/e2e/runner/setup.go
```diff
@@ -150,7 +150,7 @@ func MakeGenesis(testnet *e2e.Testnet) (types.GenesisDoc, error) {
 	if testnet.PbtsUpdateHeight == -1 {
 		genesis.ConsensusParams.Feature.PbtsEnableHeight = testnet.PbtsEnableHeight
 	}
-	for valName, power := range *testnet.Manifest.Validators {
+	for valName, power := range testnet.Validators {
```
alesforz, line 153: I it correct here to drop one "level" of dereference? From `testnet.Manifest.Validators` to `testnet.Validators`.

test/e2e/tests/validator_test.go
```diff
@@ -153,7 +153,7 @@ type validatorSchedule struct {
 
 func newValidatorSchedule(t *testing.T, testnet *e2e.Testnet) *validatorSchedule {
 	t.Helper()
-	valMap := *testnet.Manifest.Validators        // genesis validators
+	valMap := testnet.Validators                  // genesis validators
```
alesforz, line 156: Same as [above](https://github.com/cometbft/cometbft/pull/4177/files#r1774903314)

test/e2e/runner/setup.go
cason: I think the code is clear. `Testnet` "extends `Manifest` so it inherits its fields.

We are using it like that in the remaining of this method.

test/e2e/tests/validator_test.go
cason: Same reply as above.

test/e2e/runner/setup.go
alesforz: You are right 👍

alesforz (approved):

PR #4181, opened 2024-09-26
test(mempool): Add benchmarks for blocking iterator

hvanz:
As per the title.

---

#### PR checklist

- [ ] Tests written/updated


mempool/bench_test.go
```diff
@@ -148,3 +155,159 @@ func BenchmarkUpdateRemoteClient(b *testing.B) {
...
+		mp.Flush()
+		wg := sync.WaitGroup{}
+		wg.Add(numIterators)
+		b.StartTimer()
+
+		// Concurrent iterators.
+		for j := 0; j < numIterators; j++ {
+			go func() {
+				defer wg.Done()
+				b.StopTimer()
```
melekes, line 212: This looks like a data race to me. I would rewrite it to avoid any potential problems.

```go
                 wg := sync.WaitGroup{}
		wg.Add(numIterators)
		
		// Concurrent iterators.
		iters := make([]*BlockingIterator, numIterators)
		for j := 0; j < numIterators; j++ {
			iters[j] = NewBlockingIterator(context.TODO(), mp, strconv.Itoa(j))
		}
		
		b.StartTimer()
		
			go func(iter *BlockingIterator) {
				defer wg.Done()
```

mempool/bench_test.go
```diff
@@ -148,3 +155,159 @@ func BenchmarkUpdateRemoteClient(b *testing.B) {
...
+
+		wg := sync.WaitGroup{}
+		wg.Add(numIterators)
+		b.StartTimer()
+
+		// Concurrent iterators.
+		for j := 0; j < numIterators; j++ {
+			go func() {
+				defer wg.Done()
+				b.StopTimer()
```
melekes, line 262: same

mempool/bench_test.go
hvanz: Thanks for pointing out @melekes. I didn't notice any data race but I see the risk of happening. 

Anyway, I decided to remove these two last test benchmarks that add and remove txs concurrently because they were too slow and difficult to benchmark correctly. For the moment, I added a simple benchmark with multiple concurrent iterators accessing all txs in the mempool, without any extra routine modifying the mempool concurrently.

melekes (approved):
👍

PR #4198, opened 2024-09-30
chore(storage): Remove in-memory copy of the genesis from `Node` 

### LINKED ISSUE #1290: Do not keep the entire genesis file in memory during node operation
### Context
Based on the findings in #1286, it might be possible to remove the logic that keeps the entire genesis file in memory during node operation. Or at least, load the genesis information on demand only when needed and free it up from memory when not needed. We need to ensure this is not a breaking change and that the current logic that needs to validate/use the genesis information doesn't change any behavior or normal operation. This should be a seamless change that would not require any operational changes. If implemented, it would also be beneficial to provide benchmark reports on the a few metrics before and after this change to demonstrate that are advantages such as less memory footprint during node operation.

### Proposal
See discussion below.

### ToDo
- [x] #4234
- [x] #4248
- [x] Benchmark the difference in a `Node` memory usage between before and after the changes.

alesforz:
Closes #1290 

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -52,7 +52,8 @@ type Node struct {
 
 	// config
 	config        *cfg.Config
-	genesisDoc    *types.GenesisDoc   // initial validator set
+	genesisDoc    *types.GenesisDoc // initial validator set
+	genesisTime   time.Time
 	privValidator types.PrivValidator // local node's validator key
 
 	// network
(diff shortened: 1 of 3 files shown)


alesforz:
Closed because we are going for a different strategy to solve the issue. See #1290 top comment.

PR #4229, opened 2024-10-01
fix(test): Initialise all mempools with lane info from app

hvanz:
There are reports of some test failing because of non-existing lanes (eg. https://github.com/cometbft/cometbft/actions/runs/11121278617/job/30900042983)

This PR always initialises mempools in tests with lane information fetched from the app.


internal/consensus/common_test.go
```diff
@@ -432,12 +432,9 @@ func subscribeToVoterBuffered(cs *State, addr []byte) <-chan cmtpubsub.Message {
...
 
-func fetchAppInfo(t *testing.T, app abci.Application) (*abci.InfoResponse, *mempl.LanesInfo) {
-	t.Helper()
-	resp, err := app.Info(context.Background(), proxy.InfoRequest)
-	require.NoError(t, err)
-	lanesInfo, err := mempl.BuildLanesInfo(resp.LanePriorities, resp.DefaultLane)
-	require.NoError(t, err)
+func fetchAppInfo(app abci.Application) (*abci.InfoResponse, *mempl.LanesInfo) {
+	resp, _ := app.Info(context.Background(), proxy.InfoRequest)
+	lanesInfo, _ := mempl.BuildLanesInfo(resp.LanePriorities, resp.DefaultLane)
```
alesforz, line 437: why skip the error checking here?

hvanz: Because functions `app.Info` and `BuildLanesInfo` are tested elsewhere. `fetchAppInfo` is an auxiliary function for consensus tests, and it would require a lot of changes just to test here that those functions do not return an error.

jmalicevic (approved):

PR #4235, opened 2024-10-02
chore(RPC): `Environment` stores only one in-memory copy of the Genesis

### LINKED ISSUE #4234: `Environment` stores either a pointer to a `GenesisDoc` or the genesis' chunks, but not both.

alesforz:
Closes #4234.

### Changes
The [`Environment`](https://github.com/cometbft/cometbft/blob/8273634cdefb2fb22765a14f9acb9fca608e9982/rpc/core/env.go#L72) type will:
- Store a pointer to a `GenesisDoc` if the genesis file size is below a threshold currently set to 16 MB.
- If the genesis file exceeds this threshold, it will split it into 16 MB chunks and store the hashes of each chunk in memory in a slice. In this case, the pointer to the `GenesisDoc` will be `nil`. 

Because the `InitGenesisChunks` method of `Environment` is responsible for creating the chunks, most of the changes of this PR focus on refactoring this method and adding unit tests for it.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -758,7 +758,8 @@ func (n *Node) ConfigureRPC() (*rpccore.Environment, error) {
 		Config: *n.config.RPC,
 	}
 	if err := rpcCoreEnv.InitGenesisChunks(); err != nil {
-		return nil, err
+		errMsg := "could not create the genesis file chunks and cache them: %s"
+		return nil, fmt.Errorf(errMsg, err)
 	}
 	return &rpcCoreEnv, nil
 }
--- a/rpc/core/env.go
+++ b/rpc/core/env.go
@@ -101,41 +101,6 @@ type Environment struct {
 	genChunks []string
 }
 
-func validatePage(pagePtr *int, perPage, totalCount int) (int, error) {
-	if perPage < 1 {
-		panic(fmt.Sprintf("zero or negative perPage: %d", perPage))
-	}
-
-	if pagePtr == nil { // no page parameter
-		return 1, nil
-	}
-
-	pages := ((totalCount - 1) / perPage) + 1
-	if pages == 0 {
-		pages = 1 // one page (even if it's empty)
-	}
-	page := *pagePtr
-	if page <= 0 || page > pages {
-		return 1, fmt.Errorf("page should be within [1, %d] range, given %d", pages, page)
-	}
-
-	return page, nil
-}
-
-func (*Environment) validatePerPage(perPagePtr *int) int {
-	if perPagePtr == nil { // no per_page parameter
-		return defaultPerPage
-	}
-
-	perPage := *perPagePtr
-	if perPage < 1 {
-		return defaultPerPage
-	} else if perPage > maxPerPage {
-		return maxPerPage
-	}
-	return perPage
-}
-
 // InitGenesisChunks checks whether it makes sense to create a cache of chunked
 // genesis data. It is called once on Node startup.
 // Rules of chunking:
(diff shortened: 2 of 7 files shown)


alesforz:
### Additional Context
The changes we want to introduce in this PR would break the behavior of the `/genesis_chunked` endpoint. 

First a reminder of how our RPC APIs that return the genesis work:
- `/genesis`: if the genesis size is smaller than 16 MB, it returns the entire genesis encoded in JSON. If it is larger, it returns an error suggesting the user to query the `/genesis_chunked` endpoint instead.
- `/genesis_chunked`: returns the genesis file one 16 MB chunk at a time. The caller is responsible for making multiple calls to `/genesis_chunked` to rebuild the entire genesis from the chunks.

In the current implementation, the `InitGenesisChunks` method (IMHO incorrectly) creates a slice of chunks even if the genesis file is less than 16 MB. This slice contains only one element: the entire genesis encoded in base64.

Theoretically, when the genesis file is smaller than 16 MB, it should be retrieved using the `/genesis` endpoint since there are no chunks at all. However, because the current implementation of `InitGenesisChunks` creates a slice of chunks regardless of the genesis size, it is possible to retrieve a small (less than 16 MB) genesis using the `/genesis_chunked` endpoint by passing 0 as the chunk ID. The caller will receive the entire genesis encoded in base64.

In contrast, our PR's changes prevent the creation of a slice of chunks if the genesis file is smaller than 16 MB (see #4234 for reasons why). This would break the `/genesis_chunked` endpoint for small genesis files. Because we don't want to break public APIs that users may depend upon, we have introduced code to maintain the existing functionality of `/genesis_chunked` for small genesis files while implementing our changes.

The changes are here.

node/node.go
```diff
@@ -758,7 +758,8 @@ func (n *Node) ConfigureRPC() (*rpccore.Environment, error) {
 		Config: *n.config.RPC,
 	}
 	if err := rpcCoreEnv.InitGenesisChunks(); err != nil {
-		return nil, err
+		errMsg := "could not create the genesis file chunks and cache them: %s"
```
melekes, line 761: What's the benefit of having a separate variable?

rpc/core/env.go
```diff
@@ -100,6 +101,65 @@ type Environment struct {
...
+
+	data, err := cmtjson.Marshal(env.GenDoc)
+	if err != nil {
+		return err
+	}
+
+	// If genesis is less than 16MB, then no chunking.
+	// Keep a pointer to a GenesisDoc in env.GenDoc.
+	if len(data) <= genesisChunkSize {
+		env.genChunks = nil
```
melekes, line 134: Does that mean we will be hitting this code path again and again because `if len(env.genChunks) > 0 {` only checks that N of chunks is > 0, but not when we don't need them?

node/node.go
alesforz: It's just for readability. There's no performance benefit if that's what you are asking.

rpc/core/env.go
alesforz: I'm not sure I understand the question 🤔 
- We don't create chunks if the genesis size is less than 16 MB. Therefore, to understand whether we need to chunk or not, we have to compute the size of the genesis.
- The implementation that we have in production computes the size of the genesis from the `[]byte` slice obtained by calling `json.Marshal`, rather than the `[]byte` slice obtained just after reading the file contents. To keep the existing behavior, the changes in this PR also compute the size from the serialized genesis.
- `InitGenesisChunks` is called only once from `Node.ConfigureRPC()` at startup, so the check `if len(env.genChunks) > 0` is performed only once.
- We initialize and modify `env.genChunks` only in `InitGenesisChunks`. Since this field is unexported, there's no risk of accidental initialization elsewhere.

We can discuss whether computing the genesis's size from the serialized `[]byte` blob rather than directly reading the file is the best approach. Serialization removes extra characters (e.g., spaces, tabs), thus reducing the input size. For example, in tests, I used a ~69 MB genesis file that reduced to ~34 MB when serialized. That's a big difference: using the file size would create 5 chunks, but using the serialized size results in 3 chunks. Since we use the JSON `[]byte` blob in production, I preferred not to change this behavior.

Does this make sense? Or is your question about something different?

node/node.go
melekes: The opposite approach (direct embedding) is as readable. But it's up to you!

melekes (approved):

PR #4249, opened 2024-10-04
feat(types): increase default `SynchronyParams` consensus parameters values

### LINKED ISSUE #4246: pbts: increase the default synchrony values to higher ones
We should be more conservative on default values, refer to https://github.com/tendermint/tendermint/pull/7788 for more details.

cason:
Closes #4246

- `Precision` was set to `505ms`, because it should be greater than 0.5s, as in https://github.com/tendermint/tendermint/issues/7724
- `MessageDelay` was increased to `15s`, following the measurements reported in https://github.com/tendermint/tendermint/pull/7788, and considering some outputs of `e2e` tests

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/types/params.go
+++ b/types/params.go
@@ -194,10 +194,10 @@ func DefaultFeatureParams() FeatureParams {
 
 func DefaultSynchronyParams() SynchronyParams {
 	// Default values determined based on experimental results and on
-	// https://github.com/tendermint/tendermint/issues/7202
+	// https://github.com/cometbft/cometbft/issues/4246
 	return SynchronyParams{
-		Precision:    500 * time.Millisecond,
-		MessageDelay: 2 * time.Second,
+		Precision:    505 * time.Millisecond,
+		MessageDelay: 15 * time.Second,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


hvanz (approved):
LGTM

types/params.go
```diff
@@ -194,10 +194,10 @@ func DefaultFeatureParams() FeatureParams {
 
 func DefaultSynchronyParams() SynchronyParams {
 	// Default values determined based on experimental results and on
-	// https://github.com/tendermint/tendermint/issues/7202
+	// https://github.com/cometbft/cometbft/issues/4246
 	return SynchronyParams{
-		Precision:    500 * time.Millisecond,
-		MessageDelay: 2 * time.Second,
+		Precision:    505 * time.Millisecond,
```
sergio-mena, line 199: Just curious, why 505 ms? (I mean, why the extra 5 ms)

sergio-mena (approved):
Thanks a lot for this 🙏

types/params.go
cason: This is the number that Willian has adopted in the original implementation.

PR #4250, opened 2024-10-04
chore(Node): Node does not keep a pointer to a GenesisDoc after initialization

### LINKED ISSUE #4248: `Node` does not keep a pointer to a `GenesisDoc` after initialization.
Work to be done as part of #1290.

### Context
#1286 [answers](https://github.com/cometbft/cometbft/issues/1286#issuecomment-2353309745) the question "_Does a `Node` need an in-memory copy of the genesis for its entire lifecycle?_" with a No.

After initialization, a `Node` only requires the [`GenesisTime`](https://github.com/cometbft/cometbft/blob/07e225bfae36b5cbac1dd4347ea7393d1f5fdaee/types/genesis.go#L40) field from the genesis. Once a `Node` has completed its startup, which includes initializing an [`Environment`](https://github.com/cometbft/cometbft/blob/07e225bfae36b5cbac1dd4347ea7393d1f5fdaee/rpc/core/env.go#L72) struct serving the RPC APIs is initialized, it no longer needs the genesis at all.

### ToDo

- [x] Change the implementation of `Node` so that it no longer keeps a pointer to the `GenesisDoc` after startup.
- [x] `Node` stores the `GenesisTime` as a field to use it after startup.

alesforz:
Closes #4248.

### Changes
The `Node` type will:
 - create a singleton instance of `Environment` when setting up the RPC APIs.
 - set the pointer to the `GenesisDoc` to `nil` once startup is complete so that the GC can collect the in-memory `GenesisDoc` object.
 - store a `GenesisTime` field.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -23,6 +23,7 @@ import (
 	bc "github.com/cometbft/cometbft/internal/blocksync"
 	cs "github.com/cometbft/cometbft/internal/consensus"
 	"github.com/cometbft/cometbft/internal/evidence"
+	cmtjson "github.com/cometbft/cometbft/libs/json"
 	"github.com/cometbft/cometbft/libs/log"
 	cmtpubsub "github.com/cometbft/cometbft/libs/pubsub"
 	"github.com/cometbft/cometbft/libs/service"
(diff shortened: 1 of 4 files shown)


jmalicevic (approved):
Needs a changelog ,entry, great work otherwise !

PR #4281, opened 2024-10-09
fix(config): Dynamic mempool type when writing config

corverroos:
When writing `config.toml` via `config.WriteConfigFile` the `mempool.type` field is hardcoded to `flood`. This fixes it to populate it from provided config struct.


melekes (approved):
👍

PR #4299, opened 2024-10-17
Chore(RPC): Reduce Genesis Chunks Size

### LINKED ISSUE #4298: Chore(RPC): Reduce Genesis Chunks Size
### Context

The JSON-RPC APIs expose the `/genesis_chunked` endpoint to allow users to download the genesis file one chunk at a time. It is the user's responsibility to manage multiple calls to `/genesis_chunked` until the genesis is fully reassembled.

In the [code](https://github.com/cometbft/cometbft/blob/4eb7c058dd99c092b932c79d7cae33010f57cad9/rpc/core/env.go#L33), we define a chunk to be 16 MB in size. Testing with a node running locally and a genesis file large enough to create chunks, I queried the endpoint with:

```bash
curl -X GET "http://localhost:26657/v1/genesis_chunked?chunk=0" -H "accept: application/json"
```
and received part of the chunk plus the following message:
```
(18) transfer closed with outstanding read data
```
Since I was running the node locally, I'd say we can discard network issues as a cause. This message can also occur if the server or client has a timeout set for the request: if there is too much data to transfer before the timeout expires, the connection may be closed.
 
 I tried running the query again after reducing the chunk size to 4 MB, and the query completed successfully without any extra messages. This suggests that it might be a timeout issue.

Users could customize CometBFT's [server configuration](https://github.com/cometbft/cometbft/blob/4eb7c058dd99c092b932c79d7cae33010f57cad9/rpc/jsonrpc/server/http_server.go#L45) to increase the timeouts. However, I argue that we should provide a "just works" setup so users don't have to change the server configuration (I didn't change it during my tests). Also, users could circumvent the problem by using the WebSocket connection that the API offers. However, using WebSockets is more cumbersome compared to a "traditional" HTTP request.

It seems that the simplest change to make is reducing the chunk size. As far as I know, HTTP doesn't impose a size limit on payloads, therefore having 16 MB chunks isn't wrong per se. But reducing the chunk size to 2–4 MB would allow the RPC APIs to work well with the default server configuration.

### ToDo
- [x] Reduce the genesis chunks size.

alesforz:
Closes #4298.

**Note**: merge #4247 before this.
---

#### PR checklist

- [] Tests written/updated

### DIFF
--- a/rpc/core/env.go
+++ b/rpc/core/env.go
@@ -30,7 +30,7 @@ const (
 
 	// genesisChunkSize is the maximum size, in bytes, of each
 	// chunk in the genesis structure for the chunked API.
-	genesisChunkSize = 16 * 1024 * 1024 // 16
+	genesisChunkSize = 2 * 1024 * 1024 // 2 MB
 )
 
 // These interfaces are used by RPC and must be thread safe


andynog (approved):
Seems OK, more chunks will be created but I'm assuming the client side will handle that. 

These values are very subjective to network conditions but probably going with a lower value might be beneficial to constrained environments

rpc/core/env.go
```diff
@@ -30,7 +30,7 @@ const (
 
 	// genesisChunkSize is the maximum size, in bytes, of each
 	// chunk in the genesis structure for the chunked API.
-	genesisChunkSize = 16 * 1024 * 1024 // 16
+	genesisChunkSize = 2 * 1024 * 1024 // 2 MB
```
jmalicevic, line 33: Based on the fact that too frequent requests might be considered as DOS attacks, not sure if it is the best idea to lower this this much.

andynog: that's a good point, we don't have a rate limiter to the RPC requests (only in P2P) in my understanding is that the operators implement their own rate limiter via proxy servers or load balancers. I think this parameter should be part of the config since it should be dynamic (respect or be in sync with external rate-limiters)

alesforz: @jmalicevic that's a valid concern, but I agree with @andynog that it's the operator's responsibility to have rate limiting (e.g., via reverse-proxies or load balancers) in place.

Sure, a higher chunk size means less RPC requests to get the whole genesis, but I think that nothing prevents a malicious client from trying to DDoS a node by spamming requests to `/genesis_chunked` anyway. They can do so regardless of the chunk size.

jmalicevic: Ok, let's reduce it but please add a changelog entry as this might break people having scripts to read the chunk sizes in place. So both as a fix and under breaking changes.

melekes (approved):

PR #4306, opened 2024-10-18
chore(rpc): optimize genesis file chunking

### LINKED ISSUE #1289: Optimize genesis file chunking
As part of #1037, we need to optimize how the genesis file is chunked.

we need to implement logic to:

- Determine the number of chunks (based on the chunk size and file size). This logic already exists, might need to adapt for optimal chunk size.
- Store the chunks in the file system (TBD in the `tmp` or `data` folder inside home)
- Return chunks loaded from the file system (not from storage) for the `genesis_chunked` endpoint.
- [OPTIONAL] Validate chunks (re-create the file from chunks) and hash it to ensure it matches hash in storage or hash of genesis in the file system (`config` folder). Maybe this could be validated when the node starts to ensure nothing has been tampered with.

#### DOD 
- [x] Do not store chunked genesis files in memory. 
- [x] Store the chunks on storage or file system and fetch them on demand. 
- [x] Rethink the default size of the genesis chunk and use an appropriate value.

This completes the work started in Phase II of #1037, thus finalizing the implementation of the chunking strategy discussed [here](https://github.com/cometbft/cometbft/issues/1289#issuecomment-2422499013).

alesforz:
Closes #1289.

#### `Environment` type
The `Environment` type will:
- store the genesis file's path on disk
- store a `map[int]string` where the keys are the chunks' IDs, and the values are the chunk's path on disk.
- not store a pointer to a `types.GenesisDoc` anymore
- not store a slice of genesis file chunks anymore. Instead the method `InitGenesisChunks` will create chunks and store them on disk, 1 file per chunk.

#### `Node` type
The `Node` type will:
- not store a pointer to a `GenesisDoc` anymore. 
- pass the genesis file's path on disk to a new `Environment object when creating it in the `ConfigureRPC()` method.

#### RPC `/genesis_chunked`
This endpoint will implement the chunking strategy described [here](https://github.com/cometbft/cometbft/issues/1290#issuecomment-2377326927).

---

#### PR checklist

- [ ] Tests written/updated


alesforz:
Work moved to new branch.

Issue #4302, opened 2024-10-17
p2p: QUIC transport [tracking issue]


melekes:
#### Steps

- [ ] #4148

milez-git13:
[template_qwen_image_illustration_lora.json](https://github.com/user-attachments/files/31258042/template_qwen_image_illustration_lora.json)

Issue #3250, opened 2024-06-12
p2p: connection multiplexing does not work as expected on the receiving side


cason:
This issue tries to summarize the root cause of a problem already detected by several users and in several issues. The point is not to replace the original issues but to discuss a solution, which apparently would require an architectural change in the operation of the p2p communication layer.

## Context

Two nodes in a CometBFT network are connected using a single, authenticated, encrypted connection. The baseline connection is a raw TCP connection, which is encapsulated into a [secret connection](https://github.com/cometbft/cometbft/blob/main/p2p/conn/secret_connection.go), that is responsible for authenticating the end-points and encrypting the traffic.

The (secret) connection is managed by a multiplex connection ([`MConnection`](https://github.com/cometbft/cometbft/blob/main/p2p/conn/connection.go)) that implement the abstraction of multiple _channels_ between the two nodes. Ideally, channels should work as independent connections between peers, not interfering with each other.

### Sender side

On the sender side of a multiplex connection, although some issues have been raised, we have a behavior the matches the abstraction. Each channel has a sending queue, with a configurable length, and a priority. The multiplex connection logic selects from the existing channels, based on their priorities, the next packet to be sent through the (real) connection. Messages are split into packets, with a maximum size, so that priorities apply to packets (data length), not number of messages (there is some criticism regarding this choice, e.g., https://github.com/cometbft/cometbft/issues/2954). But in any case, the interference between channels is reduced, as in a channel with multiple outstanding massages to send does not prevent messages of different channels from being sent.

### Receiver side

On the receiver side, however, the behavior is different. There is a single receive buffer per channel, to which are added received packets belonging to that channel. When a receive buffer contains a full message, the multiplex connection delivers the received message using the `onReceive` callback. The call to `onReceive` is synchronous, meaning that no other raw message (packet or control message) is received or processed while the callback function does not return. This would not be a problem if the `onReceive` implementation was non-blocking and fast. This assumption, however, has turned out to not be observed with the evolution and added complexity of the code base, which turns out to be a problem, first described in https://github.com/cometbft/cometbft/issues/2685.

### Receiving callback

[...]

PR #4307, opened 2024-10-19
refactor(p2p)!: extract `MConnection` from `Peer`

melekes:
Contributes to https://github.com/cometbft/cometbft/issues/4302

**It's now the job of a `Transport` implementation to emit a multiplexed connection.**

Previously, `MConnection` was created inside `Peer`, and `Transport` was responsible for returning a raw `net.Conn`. With the upcoming QUIC transport, this is no longer acceptable. Now, TCP transport returns a multiplexed connection (`MConnection`). QUIC transport, likewise, will be emitting multiplexed connection too. The difference is, of course, in case of TCP it's out-of-protocol multiplexing.

Note I've made an internal "queue" in `MConnection`, which stores packets and the associated "channelID" (streamID). So, `MConnection` has a `readLoop`, which pushes incoming packets onto this queue (`map[streamID]chan []byte` actually). The packets then get read in `Peer#streamReadLoop` (1 per stream), which calls `Stream#Read`. So even though there's a single TCP connection, there's an abstraction of streams of data.

p2p drawio

## New interfaces

Transport:

```go
// Transport connects the local node to the rest of the network.
type Transport interface {
	// NetAddr returns the network address of the local node.
	NetAddr() na.NetAddr

	// Accept waits for and returns the next connection to the local node.
	Accept() (Connection, *na.NetAddr, error)

	// Dial dials the given address and returns a connection.
	Dial(addr na.NetAddr) (Connection, error)
}

// StreamDescriptor describes a data stream. This could be a substream within a
// multiplexed TCP connection, QUIC stream, etc.
type StreamDescriptor interface {
	// StreamID returns the ID of the stream.
	StreamID() byte
	// MessageType returns the type of the message sent/received on this stream.
	MessageType() proto.Message
}
```

Connection:

```go
// Connection is a multiplexed connection that can send and receive data
// on multiple streams.
type Connection interface {
	// OpenStream opens a new stream on the connection with an optional
	// description.
	OpenStream(streamID byte, desc any) (Stream, error)

	// LocalAddr returns the local network address, if known.
	LocalAddr() net.Addr

	// RemoteAddr returns the remote network address, if known.
	RemoteAddr() net.Addr

	// Close closes the connection.
	// If the protocol supports it, a reason will be sent to the remote.
	// Any blocked Read operations will be unblocked and return errors.
	Close(reason string) error

	// FlushAndClose flushes all the pending bytes and closes the connection.
	// If the protocol supports it, a reason will be sent to the remote.
	// Any blocked Read operations will be unblocked and return errors.
	FlushAndClose(reason string) error

	// ConnectionState returns basic details about the connection.
	// Warning: This API should not be considered stable and might change soon.
	ConnectionState() any

	// ErrorCh returns a channel that will receive errors from the connection.
	ErrorCh() <-chan error
}

// Stream is the interface implemented by QUIC streams or multiplexed TCP connection.
type Stream interface {
	ReceiveStream
	SendStream
	// SetDeadline sets the read and write deadlines associated with the connection. It is equivalent to calling both
	// SetReadDeadline and SetWriteDeadline.
	SetDeadline(t time.Time) error
}

// A ReceiveStream is a unidirectional Receive Stream.
type ReceiveStream interface {
	// Read reads data from the stream.
	// Read can be made to time out and return a net.Error with Timeout() == true
	// after a fixed time limit; see SetDeadline and SetReadDeadline.
	// If the stream was canceled by the peer, the error is a StreamError and
	// Remote == true.
	// If the connection was closed due to a timeout, the error satisfies
	// the net.Error interface, and Timeout() will be true.
	io.Reader
	// SetReadDeadline sets the deadline for future Read calls and
	// any currently-blocked Read call.
	// A zero value for t means Read will not time out.
	SetReadDeadline(t time.Time) error
}

// A SendStream is a unidirectional Send Stream.
type SendStream interface {
	// Write writes data to the stream.
	// Write can be made to time out and return a net.Error with Timeout() == true
	// after a fixed time limit; see SetDeadline and SetWriteDeadline.
	// If the stream was canceled by the peer, the error is a StreamError and
	// Remote == true.
	// If the connection was closed due to a timeout, the error satisfies
	// the net.Error interface, and Timeout() will be true.
	io.Writer
	// Close closes the write-direction of the stream.
	// Future calls to Write are not permitted after calling Close.
	// It must not be called concurrently with Write.
	// It must not be called after calling CancelWrite.
	io.Closer
	// SetWriteDeadline sets the deadline for future Write calls
	// and any currently-blocked Write call.
	// Even if write times out, it may return n > 0, indicating that
	// some data was successfully written.
	// A zero value for t means Write will not time out.
	SetWriteDeadline(t time.Time) error
}
```

<details>
<summary>Copilot Summary</summary>

This pull request introduces several changes to the `cometbft` codebase, focusing on refactoring the P2P (peer-to-peer) communication layer to use a new `abstract` package. The changes include updating import statements, modifying function signatures, and adding new types to the `abstract` package.

### Refactoring to use `abstract` package:
* [`internal/blocksync/reactor.go`](diffhunk://#diff-899a514f0eaa9d7eadf01ac7b51f0c86eca12992ced5ce21293003532ed4edd2L177-R180): Updated `StreamDescriptors` function to return `[]abstract.StreamDescriptor` instead of `[]p2p.StreamDescriptor`.
* [`internal/consensus/reactor.go`](diffhunk://#diff-c53cca1a09efa0da0a0427e03e09580577ab49680121b98b8e256403c7638be3L156-R167): Modified `StreamDescriptors` function to use `abstract.StreamDescriptor`. [[1]](diffhunk://#diff-c53cca1a09efa0da0a0427e03e09580577ab49680121b98b8e256403c7638be3L156-R167) [[2]](diffhunk://#diff-c53cca1a09efa0da0a0427e03e09580577ab49680121b98b8e256403c7638be3L175-R184)
* [`internal/evidence/reactor.go`](diffhunk://#diff-cd5b5850d0c8ff4327cc6fdc260f6ae89a7529e76d817defc3c740a6e2140df1L55-R58): Changed `StreamDescriptors` function to return `[]abstract.StreamDescriptor`.
* [`mempool/reactor.go`](diffhunk://#diff-3e6afafe7994a696d508a5a5b093343bdb5d1ff2ff00bab0318d2c87547e74e5L76-R86): Updated `StreamDescriptors` function to return `[]abstract.StreamDescriptor`.
* [`node/setup.go`](diffhunk://#diff-bbaa90ad1c80edb1f0084525de9b5e22da605e22d5e0b374d04c1e6a363d6f9eR421-R429): Refactored `createTransport` function to use `abstract.Transport` and `tcpconn.DefaultMConnConfig`. [[1]](diffhunk://#diff-bbaa90ad1c80edb1f0084525de9b5e22da605e22d5e0b374d04c1e6a363d6f9eR421-R429) [[2]](diffhunk://#diff-bbaa90ad1c80edb1f0084525de9b5e22da605e22d5e0b374d04c1e6a363d6f9eL480-R488)

### Adding new types to `abstract` package:
* [`p2p/abstract/connection.go`](diffhunk://#diff-3290d4e5c2150f292b559d4dd0bf96b1baec901709720af0e960fc81b6887688R1-R86): Added new `Connection` and `Stream` interfaces.
* [`p2p/abstract/transport.go`](diffhunk://#diff-d46e138cfe0ae7fef85308f237552ea48fb82a55d979489c18e253404339eb4cR1-R28): Introduced `Transport` and `StreamDescriptor` interfaces.

### Code improvements and fixes:
* [`internal/blocksync/pool.go`](diffhunk://#diff-758b3b1171ec0de0566f658ca2de4179ebebc2a7c01dd76dfe4984e923e97898L432-R432): Fixed logging format in `banPeer` function.
* [`internal/blocksync/reactor_test.go`](diffhunk://#diff-ea0c63e7ea95f30c15eb20a420c9932f748dff723423fa947678db9ebac592e2R290-R291): Added timeout mechanism to `TestBadBlockStopsPeer` to prevent infinite loops. [[1]](diffhunk://#diff-ea0c63e7ea95f30c15eb20a420c9932f748dff723423fa947678db9ebac592e2R290-R291) [[2]](diffhunk://#diff-ea0c63e7ea95f30c15eb20a420c9932f748dff723423fa947678db9ebac592e2R303-R306) [[3]](diffhunk://#diff-ea0c63e7ea95f30c15eb20a420c9932f748dff723423fa947678db9ebac592e2R328-R339)
* [`node/node.go`](diffhunk://#diff-375d57e386f20eaa5f09f02bb9d28bfc48ac3dca18d0325f59492208219e5618R509-R510): Set logger for transport in `NewNodeWithCliParams` function.

These changes aim to improve the modularity and maintainability of the P2P communication layer by abstracting common functionalities and enhancing the overall code structure.
</details>


internal/blocksync/reactor.go
```diff
@@ -174,9 +175,9 @@ func (bcR *Reactor) OnStop() {
 }
 
 // StreamDescriptors implements Reactor.
-func (*Reactor) StreamDescriptors() []p2p.StreamDescriptor {
-	return []p2p.StreamDescriptor{
-		&tcpconn.ChannelDescriptor{
+func (*Reactor) StreamDescriptors() []abstract.StreamDescriptor {
```
melekes, line 178: Can you think of a better name than `abstract`? `common`?

p2p/abstract/transport.go
```diff
@@ -0,0 +1,39 @@
...
+	// Cleanup any resources associated with the given connection.
+	//
+	// Must be run when the peer is dropped for any reason.
+	Cleanup(conn Connection) error
+
+	// UpdateStreamDescriptors updates the stream descriptors when a new reactor
+	// is added or an old one gets removed.
+	//
+	// See StreamDescriptor
+	UpdateStreamDescriptors(descs []StreamDescriptor)
```
melekes, line 29: There may be a way to remove this.

p2p/handshake.go
```diff
@@ -27,22 +30,14 @@ type ErrRejected struct {
...
-}
-
 func (e ErrRejected) Error() string {
 	if e.isAuthFailure {
 		return fmt.Sprintf("auth failure: %s", e.err)
 	}
 
 	if e.isDuplicate {
 		if e.conn != nil {
-			return fmt.Sprintf(
```
melekes, line 42: Check that we're not making errors that are less informative upstream.

p2p/mock/peer.go
```diff
@@ -77,9 +79,47 @@ func (mp *Peer) Get(key string) any {
...
-func (*Peer) SetRemovalFailed()          {}
-func (*Peer) GetRemovalFailed() bool     { return false }
+func (mp *Peer) RemoteIP() net.IP          { return mp.ip }
+func (mp *Peer) SocketAddr() *na.NetAddr   { return mp.addr }
+func (mp *Peer) RemoteAddr() net.Addr      { return &net.TCPAddr{IP: mp.ip, Port: 8800} }
+func (mp *Peer) Conn() abstract.Connection { return mockConnection{mp.server} }
+func (*Peer) SetRemovalFailed()            {}
+func (*Peer) GetRemovalFailed() bool       { return false }
+
+type mockStream struct {
```
melekes, line 89: reuse p2p/mockStream?

p2p/peer.go
```diff
@@ -24,6 +26,23 @@ import (
 // granularity.
 const metricsTickerDuration = 1 * time.Second
 
+// peerConfig is used to bundle data we need to fully setup a Peer with an
+// MConn, provided by the caller of Accept and Dial (currently the Switch). This
+// a temporary measure until reactor setup is less dynamic and we introduce the
+// concept of PeerBehaviour to communicate about significant Peer lifecycle
+// events.
+// TODO(xla): Refactor out with more static Reactor setup and PeerBehaviour.
```
melekes, line 34:
```suggestion
```

p2p/peer.go
```diff
@@ -189,20 +265,24 @@ func (p *peer) String() string {
 // SetLogger implements BaseService.
 func (p *peer) SetLogger(l log.Logger) {
 	p.Logger = l
-	p.mconn.SetLogger(l)
+	// p.mconn.SetLogger(l)
```
melekes, line 268: uncomment or remove

p2p/peer.go
```diff
@@ -253,46 +335,67 @@ func (p *peer) SocketAddr() *na.NetAddr {
...
+	streamID := e.ChannelID
+	stream, ok := p.streams[streamID]
+	if !ok {
+		panic(fmt.Sprintf("stream %d not found", streamID))
+	}
+	_ = stream.SetWriteDeadline(time.Now().Add(10 * time.Second))
+	return p.send(e.ChannelID, e.Message, stream.Write)
 }
 
 // TrySend msg bytes to the channel identified by chID byte. Immediately returns
```
melekes, line 356: not accurate anymore

p2p/peer.go
```diff
@@ -364,18 +467,19 @@ func (p *peer) metricsReporter() {
...
-				sendQueueSize += float64(chStatus.SendQueueSize)
-			}
-
-			p.metrics.RecvRateLimiterDelay.With("peer_id", string(p.ID())).
-				Add(status.RecvMonitor.SleepTime.Seconds())
-			p.metrics.SendRateLimiterDelay.With("peer_id", string(p.ID())).
-				Add(status.SendMonitor.SleepTime.Seconds())
-
-			p.metrics.PeerPendingSendBytes.With("peer_id", string(p.ID())).Set(sendQueueSize)
+			// TODO: uncomment
```
melekes, line 470: TODO

p2p/transport/tcp/conn/connection.go
```diff
@@ -40,17 +41,15 @@ const (
...
 
-type (
-	receiveCbFunc func(chID byte, msgBytes []byte)
-	errorCbFunc   func(any)
+	// Capacity of the receive channel for each stream.
+	maxRecvChanCap = 100
 )
 
 /*
+TODO: rewrite
```
melekes, line 52: TODO

p2p/transport/tcp/conn/connection.go
```diff
@@ -237,6 +217,10 @@ func (c *MConnection) OnStart() error {
 	return nil
 }
 
+func (c *MConnection) Conn() net.Conn {
```
melekes, line 220: Is this still needed?

p2p/transport/tcp/conn/connection.go
```diff
@@ -321,6 +310,45 @@ func (c *MConnection) OnStop() {
...
+func (c *MConnection) RemoteAddr() net.Addr {
+	return c.conn.RemoteAddr()
+}
+
+func (c *MConnection) OpenStream(streamID byte) (abstract.Stream, error) {
+	return &MConnectionStream{conn: c, streadID: streamID}, nil
+}
+
+func (c *MConnection) FlushAndClose(string) error {
+	c.FlushStop()
```
melekes, line 330: inline

p2p/transport/tcp/conn/connection.go
```diff
@@ -926,3 +905,103 @@ func mustWrapPacketInto(pb proto.Message, dst *tmp2p.Packet) {
...
+
+	// No messages to read.
+	return 0, nil
+}
+
+func (s MConnectionStream) Write(b []byte) (n int, err error) {
+	if err := s.conn.sendBytes(s.streadID, b, s.writeTimeout()); err != nil {
+		return 0, err
+	}
+	return len(b), nil
```
melekes, line 954: mention it's optimistic

p2p/transport/tcp/conn/errors.go
```diff
@@ -9,6 +9,9 @@ var (
 	ErrInvalidSecretConnKeySend = errors.New("send invalid secret connection key")
 	ErrInvalidSecretConnKeyRecv = errors.New("invalid receive SecretConnection Key")
 	ErrChallengeVerification    = errors.New("challenge verification failed")
+
+	ErrTimeout    = errors.New("read/write timeout")
```
melekes, line 13: docs

p2p/transport/tcp/tcp.go
```diff
@@ -118,26 +119,24 @@ type MultiplexTransport struct {
 	// TODO(xla): This config is still needed as we parameterise peerConn and
 	// peer currently. All relevant configuration should be refactored into options
 	// with sane defaults.
-	mConfig conn.MConnConfig
+	mConfig *conn.MConnConfig
 }
 
 // Test multiplexTransport for interface completeness.
 var (
 	_ transportLifecycle = (*MultiplexTransport)(nil)
```
melekes, line 127: is this still needed?

p2p/transport/tcp/conn/connection.go
melekes: yes

p2p/abstract/transport.go
melekes: Probably better to leave this for a separate PR.

p2p/peer.go
```diff
@@ -174,13 +186,68 @@ func newPeer(
 	return p
 }
 
+func (p *peer) readLoop(streamID byte, stream abstract.Stream) {
```
melekes, line 189:
```suggestion
func (p *peer) streamReadLopp(streamID byte, stream abstract.Stream) {
```

p2p/peer.go
```diff
@@ -174,13 +186,68 @@ func newPeer(
...
+			p.onPeerError(p, r)
+		}
+	}()
+
+	for {
+		if !p.IsRunning() {
+			return
+		}
+
+		buf := make([]byte, 1024) // TODO max msg size for this stream
```
melekes, line 202: TODO

p2p/peer.go
```diff
@@ -338,16 +450,16 @@ func (p *peer) GetRemovalFailed() bool {
...
 }
 
 // CanSend returns true if the send queue is not full, false otherwise.
-func (p *peer) CanSend(chID byte) bool {
-	if !p.IsRunning() {
-		return false
-	}
-	return p.mconn.CanSend(chID)
-}
+// func (p *peer) CanSend(chID byte) bool {
```
melekes, line 457: remove?

p2p/peer_test.go
```diff
@@ -18,157 +17,150 @@ import (
...
+	require.NoError(t, err)
 	t.Cleanup(func() {
 		if err := p.Stop(); err != nil {
 			t.Error(err)
 		}
 	})
 
-	assert.True(p.CanSend(testCh))
-	assert.True(p.Send(Envelope{ChannelID: testCh, Message: &p2p.Message{}}))
+	// TODO: uncomment or remove
```
melekes, line 74: remove?

p2p/switch.go
```diff
@@ -771,7 +734,14 @@ func (sw *Switch) addOutboundPeerWithConfig(
 		return err
 	}
 
-	nodeInfo, err := handshake(sw.nodeInfo, conn, sw.config.HandshakeTimeout)
+	stream, err := conn.OpenStream(HandshakeStreamID, nil)
+	if err != nil {
+		sw.Logger.Error("Failed to open handshake stream", "err", err)
```
melekes, line 739:
```suggestion
		sw.Logger.Error("Failed to open handshake stream", "peer", addr, "err", err)
```

p2p/switch.go
```diff
@@ -675,21 +638,27 @@ func (sw *Switch) acceptRoutine() {
 			break
 		}
 
-		nodeInfo, err := handshake(sw.nodeInfo, conn, sw.config.HandshakeTimeout)
+		stream, err := conn.OpenStream(HandshakeStreamID, nil)
+		if err != nil {
+			sw.Logger.Error("Failed to open handshake stream", "err", err)
```
melekes, line 643:
```suggestion
			sw.Logger.Error("Failed to open handshake stream", "peer", addr, "err", err)
```

p2p/test_util.go
```diff
@@ -258,9 +305,11 @@ func testPeerConn(
 	conn := rawConn
 
 	// Fuzz connection
+	// TODO: uncomment
```
melekes, line 308: TODO

p2p/transport/tcp/conn/connection.go
```diff
@@ -269,16 +235,80 @@ func (c *MConnection) stopServices() (alreadyStopped bool) {
...
+	default:
+	}
+
+	return c.Stop()
+}
+
+// FlushAndClose replicates the logic of OnStop. It additionally ensures that
+// all successful writes will get flushed before closing the connection.
+func (c *MConnection) FlushAndClose(reason string) error {
+	// inform the error channel that we are shutting down.
```
melekes, line 301: TODO: make sure c.Quit() is closed

melekes:
opening up for some preliminary reviews 😅

melekes:
ready for review ✅

p2p/transport/tcp/conn/connection.go
```diff
@@ -116,8 +98,16 @@ type MConnection struct {
 	created time.Time // time of creation
 
 	_maxPacketMsgSize int
+
+	mtx sync.RWMutex
+	// streamID -> list of incoming messages
+	recvMsgsByStreamID map[byte]chan []byte
```
melekes, line 104: error prone + "unwanted" queue => this is why I don't think it shouldn't be merged into `main`. The only way is to remove it and bring back old API like `onReceive(...)`

internal/blocksync/reactor.go
cason: Is there a reason for these interfaces not be on the base `p2p` package? Is an import loop issue?

cason: Could them be place in `p2p/transport` as well?

p2p/transport/tcp/tcp.go
```diff
@@ -186,15 +181,17 @@ func (mt *MultiplexTransport) Dial(
...
+	mconn, _, err := mt.upgrade(c, &addr)
 	if err != nil {
 		return nil, err
 	}
+	mconn.SetLogger(mt.logger.With("remote", addr))
+
+	go mt.cleanupConn(c.RemoteAddr(), mconn.Quit())
 
-	return secretConn, nil
+	return mconn, mconn.Start()
```
cason, line 192: Maybe we should not start the `mconn` here, but when the peer starts.

In this way we can call `OpenStream` before the multiplex connection is started, avoiding locks around the streams table.

p2p/peer.go
```diff
@@ -253,46 +359,69 @@ func (p *peer) SocketAddr() *na.NetAddr {
...
+	} else if !p.HasChannel(streamID) {
 		return false
 	}
+
 	msgType := getMsgType(msg)
 	if w, ok := msg.(types.Wrapper); ok {
 		msg = w.Wrap()
 	}
+
+	sLogger := p.Logger.With("streamID", streamID, "msgType", msgType)
```
cason, line 406: Are we sure we want to allocate a logger for each message we send?

p2p/peer.go
```diff
@@ -253,46 +359,69 @@ func (p *peer) SocketAddr() *na.NetAddr {
...
+	stream, ok := p.streams[streamID]
+	if !ok {
+		panic(fmt.Sprintf("stream %d not found", streamID))
+	}
+	_ = stream.SetWriteDeadline(time.Now().Add(100 * time.Millisecond))
+	return p.send(e.ChannelID, e.Message, stream.Write)
 }
 
-func (p *peer) send(chID byte, msg proto.Message, sendFunc func(byte, []byte) bool) bool {
+func (p *peer) send(streamID byte, msg proto.Message, sendFunc func([]byte) (int, error)) bool {
```
cason, line 394: You could pass the timeout, for computing the deadline, here as a parameter, so to avoid duplicated code in `Send()` and `TrySend()`.

p2p/peer.go
```diff
@@ -338,15 +462,7 @@ func (p *peer) GetRemovalFailed() bool {
 
 // RemoteAddr returns peer's remote network address.
 func (p *peer) RemoteAddr() net.Addr {
-	return p.peerConn.conn.RemoteAddr()
-}
-
-// CanSend returns true if the send queue is not full, false otherwise.
-func (p *peer) CanSend(chID byte) bool {
```
cason, line 345: Again, for my internal documentation, this was not used in release branches, it is legacy code.

p2p/peer.go
```diff
@@ -364,7 +480,8 @@ func (p *peer) metricsReporter() {
 	for {
 		select {
 		case <-metricsTicker.C:
-			status := p.mconn.Status()
+			// TODO: this is a bit of a hack, we should have a better way to get the status.
```
cason, line 483: This is indeed terrible. The best way I can see is to create a `p2p/status` package and declare the common fields there.

p2p/peer.go
```diff
@@ -174,13 +191,73 @@ func newPeer(
 	return p
 }
 
+func (p *peer) streamReadLoop(streamID byte, stream transport.Stream) {
```
cason, line 194: I would be against, at this point, implementing parallel processing of messages. This is a complex issue, discussed in https://github.com/cometbft/cometbft/issues/3250.

But from the current design, it appears that we need this architecture...

p2p/peer.go
```diff
@@ -189,20 +266,47 @@ func (p *peer) String() string {
...
-	if err := p.mconn.Start(); err != nil {
-		return err
+	// TODO: establish priority for reading from streams (consensus -> evidence -> mempool).
+	for streamID, stream := range p.streams {
+		go p.streamReadLoop(streamID, stream)
 	}
 
 	go p.metricsReporter()
+
+	// Handle connection errors.
```
cason, line 298: This is another go routine we add to each peer, ehehe.

p2p/mock_transport.go
cason, line 1: Shouldn't this be inside `mock` or `mocks`?

p2p/switch.go
```diff
@@ -771,8 +740,17 @@ func (sw *Switch) addOutboundPeerWithConfig(
 		return err
 	}
 
-	nodeInfo, err := handshake(sw.nodeInfo, conn, sw.config.HandshakeTimeout)
+	stream, err := conn.OpenStream(HandshakeStreamID, nil)
```
cason, line 743: I wonder if this logic can be moved into the `handshake` method, since it is the same here and on the accept routine.

p2p/switch_test.go
```diff
@@ -173,20 +169,15 @@ func TestSwitches(t *testing.T) {
 	}
 	// Test broadcast and TryBroadcast on different channels in parallel.
 	// We have no channel capacity concerns, as each broadcast is on a distinct channel
-	s1.Broadcast(Envelope{ChannelID: byte(0x00), Message: ch0Msg})
```
cason, line 176: The issue here is that this channel is the handshake one?

p2p/switch.go
```diff
@@ -675,21 +644,27 @@ func (sw *Switch) acceptRoutine() {
 			break
 		}
 
-		nodeInfo, err := handshake(sw.nodeInfo, conn, sw.config.HandshakeTimeout)
+		stream, err := conn.OpenStream(HandshakeStreamID, nil)
```
cason, line 647: `HandshakeStreamID = 0` right now. I think we should use the maximum possible ID for it, to really avoid mess with channel/stream IDs used by reactors.

cason (approved):
Second round of reviewing.

Left other comments. There is a very odd e2e error, apparently a full node is getting a wrong validator set for some reason. Probably not related to the p2p refactoring.

I really think that this PR should be split in multiple ones, since we are addressing a lot of stuff here. Some for which we don't have much clarity (e.g., removing the `onReceive` callback versus multiple consuming routines + dropping messages).

I will approve to unblock the work. But we need to fully review the changes when the feature branch is ready to go.

p2p/handshake.go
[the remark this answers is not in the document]
melekes: > By the way, QUIC does not implement the net.Conn interface?

No, it doesn't.

p2p/peer.go
```diff
@@ -36,11 +53,8 @@ type Peer interface {
 	IsOutbound() bool   // did we dial the peer
 	IsPersistent() bool // do we redial this peer when we disconnect
 
-	// Conn returns the underlying connection.
-	Conn() net.Conn
```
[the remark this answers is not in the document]
melekes: > We used to have a CloseConn() method here instead, that I assume is not needed any longer?

`CloseConn` was a hack. I've found out that we can use `Close` method instead.

```
	err = p.(*peer).peerConn.Close("simulate failure")
	require.NoError(t, err)
```

p2p/switch_test.go
melekes: > The issue here is that this channel is the handshake one?

The handshake stream is closed once the handshake is done. The same ID can be reused later on.

melekes: > The issue here is that this channel is the handshake one?

I don't remember exactly. Let me try to revert it.

melekes: done ✅

p2p/mock_transport.go
melekes: Yes, but it can't be due to **circular import**.

p2p/peer.go
melekes: merged with metrics reporting ✅

p2p/peer.go
melekes: done ✅

p2p/peer.go
```diff
@@ -174,13 +191,73 @@ func newPeer(
...
+		msgType = p.streamInfoByStreamID[streamID].msgType
+		logger  = p.Logger.With("stream", streamID)
+	)
+
+	for {
+		if !p.IsRunning() {
+			return
+		}
+
+		n, err := stream.Read(buf)
```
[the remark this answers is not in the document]
melekes:
```
// Read reads whole messages from the internal map, which is populated by the
// global read loop.
// thread-safe.
func (s *MConnectionStream) Read(b []byte) (n int, err error) {
```

already documented

p2p/peer.go
```diff
@@ -109,21 +123,24 @@ func (pc peerConn) RemoteIP() net.IP {
...
-	// raw peerConn and the multiplex connection
 	peerConn
-	mconn *tcpconn.MConnection
 
 	// peer's node info and the channel it knows about
 	// channels = nodeInfo.Channels
 	// cached to avoid copying nodeInfo in HasChannel
 	nodeInfo ni.NodeInfo
 	channels []byte
+	streams  map[byte]transport.Stream
```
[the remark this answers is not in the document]
melekes: Because `streamInfoByStreamID ` is passed from the outside? I believe `streamInfoByStreamID ` is also reused across all peers, ie. it's not deep copy

p2p/transport/tcp/conn/stream_descriptor.go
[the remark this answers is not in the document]
melekes: Yes. It will look similar but will not contain fields like `SendQueueCapacity` or `Priority`.

p2p/peer.go
```diff
@@ -253,46 +359,69 @@ func (p *peer) SocketAddr() *na.NetAddr {
...
-func (p *peer) Status() tcpconn.ConnectionStatus {
-	return p.mconn.Status()
+func (p *peer) Status() any {
+	return p.ConnectionState()
 }
 
-// Send msg bytes to the channel identified by chID byte. Returns false if the
-// send queue is full after timeout, specified by MConnection.
+// Send sends the given envelope via the stream identified by e.ChannelID.
+// Returns false if the send queue is full after 10s.
```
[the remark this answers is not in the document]
melekes: done ✅

p2p/peer.go
```diff
@@ -253,46 +359,69 @@ func (p *peer) SocketAddr() *na.NetAddr {
...
+	return p.send(e.ChannelID, e.Message, stream.Write)
 }
 
-// TrySend msg bytes to the channel identified by chID byte. Immediately returns
-// false if the send queue is full.
+// TrySend tries to send the given envelope the stream identified by e.ChannelID.
+// Returns false if the send queue is full after 100ms.
 //
 // thread safe.
 func (p *peer) TrySend(e Envelope) bool {
```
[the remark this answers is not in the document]
melekes: reverted ✅

p2p/transport/tcp/conn/connection.go
```diff
@@ -780,51 +810,51 @@ func newChannel(conn *MConnection, desc ChannelDescriptor) *Channel {
...
+		case <-ch.conn.Quit():
+			return ErrNotRunning
+		}
 	}
-}
 
-// Queues message to send to this channel.
-// Nonblocking, returns true if successful.
-// Goroutine-safe.
-func (ch *Channel) trySendBytes(bytes []byte) bool {
```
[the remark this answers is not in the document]
melekes: agree! reverted ✅

melekes:
@cason thanks for the review 👏

p2p/transport/tcp/conn/connection.go
```diff
@@ -269,153 +228,177 @@ func (c *MConnection) stopServices() (alreadyStopped bool) {
...
+
+func (c *MConnection) LocalAddr() net.Addr {
+	return c.conn.LocalAddr()
+}
+
+func (c *MConnection) RemoteAddr() net.Addr {
+	return c.conn.RemoteAddr()
+}
+
+func (c *MConnection) OpenStream(streamID byte, desc any) (abstract.Stream, error) {
```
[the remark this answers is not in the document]
melekes: done ✅

p2p/transport/tcp/conn/connection.go
```diff
@@ -269,153 +227,177 @@ func (c *MConnection) stopServices() (alreadyStopped bool) {
...
 	}
 
-	c.Logger.Debug("Send", "channel", chID, "conn", c, "msgBytes", log.NewLazySprintf("%X", msgBytes))
+	// c.Logger.Debug("Send",
+	// 	"streamID", chID,
+	// 	"msgBytes", log.NewLazySprintf("%X", msgBytes),
+	// 	"timeout", timeout)
 
-	// Send message to channel.
+	c.mtx.RLock()
```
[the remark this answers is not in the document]
melekes: > Do not start the connection until all needed streams are configured (where the writes occur)

I went with this.

p2p/transport/tcp/conn/connection.go
```diff
@@ -687,6 +728,29 @@ FOR_LOOP:
...
+	// Make a copy to avoid a DATA RACE.
+	msgCopy := make([]byte, len(msgBytes))
+	copy(msgCopy, msgBytes)
+
+	// Push the message.
+	select {
+	case ch <- msgCopy:
+		return nil
+	default: // Drop the message if the buffer is full.
+		return fmt.Errorf("receive buffer is full for stream %X", streamID)
```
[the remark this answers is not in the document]
melekes: done ✅

PR #4312, opened 2024-10-20
feat(e2e): allow customizing genesis parameters on the manifest (backport #3969)

### LINKED ISSUE #3961: e2e: allow customizing genesis parameters on the manifest

cason:
Closes #3961.

Similarly to #3967, introduces a new `genesis` field for manifest files. The format is `key = value` and multiple entries can be provided. For example,

```
genesis = [ 
    "consensus_params.evidence.max_bytes = 1000", 
    "chain_id = pr3969" 
]
```

After the genesis document is produced by the `setup` implementation of the runner, the provided genesis customized configurations are applied to the generated genesis file. The solution also uses `viper`, as it allows setting only some specific keys, leaving the pre-produced content untouched.

Note: this PR is based atop
https://github.com/cometbft/cometbft/pull/3964 for simplicity, but it does not need to.

---

This is an manual backport of pull request https://github.com/cometbft/cometbft/pull/3969.

#### PR checklist

- [ ] Tests written/updated

---------

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -109,6 +109,7 @@ linters-settings:
           - github.com/spf13
           - github.com/stretchr/testify/require
           - github.com/syndtr/goleveldb
+          - github.com/mitchellh/mapstructure
       test:
         files:
           - "$test"
--- a/test/e2e/pkg/manifest.go
+++ b/test/e2e/pkg/manifest.go
@@ -142,6 +142,11 @@ type Manifest struct {
 	// 0 denotes it is set at InitChain.
 	PbtsUpdateHeight int64 `toml:"pbts_update_height"`
 
+	// Genesis is a set of key-value config entries to write to the
+	// produced genesis file. The format is "key = value".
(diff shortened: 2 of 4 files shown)


cason (commented):
Diffs are identical. But I can't approve this PR :(

melekes (approved):

PR #4313, opened 2024-10-20
fix(e2e): viper errors with invalid `config` and `genesis` manifest keys (backport #4016)

cason:
Addresses comments https://github.com/cometbft/cometbft/pull/3969#pullrequestreview-2277655582 and https://github.com/cometbft/cometbft/pull/3967#pullrequestreview-2277635023 on previous PRs.

The solution is in fact simple, as it is possible to tell viper to return an error if one of the configured entries (key/value pairs) were not applied to the destination structure, because they don't exist or are invalid.

---

This is an manual backport of pull request https://github.com/cometbft/cometbft/pull/4016.

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -109,6 +109,7 @@ linters-settings:
           - github.com/spf13
           - github.com/stretchr/testify/require
           - github.com/syndtr/goleveldb
+          - github.com/mitchellh/mapstructure
       test:
         files:
           - "$test"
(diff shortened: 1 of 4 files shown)


cason (commented):
Diffs match. But I can't approve this manual backport.

melekes (approved):

PR #4321, opened 2024-10-23
fix(state/indexer): copy value

### LINKED ISSUE #4295: incompatiblity: tx_search is reset to a single entry when using goleveldb with cometbft v0.38.x
Goleveldb (not pebble) is broken when using tx_search with cometbft v0.38.12 and cometbft-db v0.13.0+

I don't know if this happens in `main` or not.

If users should not use cometbft-db > 0.12.x with cometbft v0.38.x, we should probably say that.  

I think it is more likely that this breakage was unintentional, related to the contract changes mentioned in the release nodes for cometbft-db:

* https://github.com/cometbft/cometbft-db/blob/v1.0.1/CHANGELOG.md#v0130

Note that users of cometbft v0.38.x may wish to use:

* the latest pebble
* the latest rocksdb
* goleveldb

....all in one place, and that the only way to do that is to either fork cometbft-db or just use the latest version.

This PR:

* https://github.com/cometbft/cometbft/pull/4297

fixes the dependency versioning only.  I think that there are additional issues here in cometbft that prevent goleveldb from working correctly when used with cometbft v0.38.x and cometbft-db v0.13.0+

melekes:
Because, in cometbft-db v0.13+, the iterator is being reused so we need to always copy key and value if we're storing them in a map or other in-memory structure.

Closes #4295

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/cometbft/cometbft
 
 go 1.22.7
 
-toolchain go1.22.7
+toolchain go1.23.1
 
 require (
 	github.com/BurntSushi/toml v1.4.0
--- a/go.mod
+++ b/go.mod
@@ -2,8 +2,6 @@ module github.com/cometbft/cometbft
 
 go 1.22.7
 
-toolchain go1.23.1
-
 require (
 	github.com/BurntSushi/toml v1.4.0
 	github.com/adlio/schema v1.3.6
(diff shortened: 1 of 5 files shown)


andynog (approved):
The changes look good to me. I've performed some local testing with the kvstore and before I was only getting one event from tx_search using `goleveldb` backend, but now I get multiple evens as expected. So seems to be fixed.

But I there is a go vulnerability check that is failing in CI, I don't get it running `make vulncheck` on my local machine though.

melekes:
This patch was also verified by the Noble team ✅

melekes:
@mergify backport v0.37.x

PR #4322, opened 2024-10-23
perf(light): Use cache when verifying signatures

### LINKED ISSUE #2365: [perf]: Remove duplicated signature checks in light.VerifyNonAdjacent
### Summary

Currently light.VerifyNonAdjacent runs two signature checks:
- [VerifyCommitLightTrusting](https://github.com/cometbft/cometbft/blob/main/light/verifier.go#L55-L56) - Check that 1/3rd of the trusted header's weight has signed the untrusted header.
- [VerifyCommitLight](https://github.com/cometbft/cometbft/blob/main/light/verifier.go#L71) - Check that 2/3rds of the untrusted header's weight has signed the untrusted header.

Notice however, that for the signatures in VerifyCommitLight that were also checked in VerifyCommitLightTrusting, we have checked them twice.

### Problem Definition

We see light.VerifyNonAdjacent taking lots of CPU time on Osmosis, coming from IBC. (Arguably IBC should have a copy of this logic there though)

VerifyCommitLightTrusting from light.VerifyNonAdjacent is currently 1.3% of the sync speed, and VerifyCommitLight from light.VerifyNonAdjacent is 4.1% of the sync speed. Which leads me to believe that fixing this signature duplication will lower the net sync speed by 1.3%. (Note: While this seems small right now, the other parts of the system, in particular the DB, is getting notably sped up. I expect once the next major release rolls around, that these IBC components are 4x more expensive within  sync speed) 

When Osmosis switches to a TM release that supports batch verification, these impacts will lower by a factor of 2, but I think are still significant enough performance for us to improve.

### Proposal

There seems to me to be two ways to alleviate this:
- Restructure the logic to make these signatures act as counted, but not need to be included in what we verify
- Make a small cache for signature checks here, make variants of the methods that allow passing in a cache, and using a cache as an argument.

My intuition is that a small cache would be:
1) Very simple to reason about correctness
2) Improve the speed quite effectively!

wllmshao:
Closes #2365

This PR adds two new methods to `light` package: `VerifyCommitLightTrustingWithCache` and `VerifyCommitLightWithCache` so that we don't have to verify the same signatures twice.

Pending some perf tests/benchmarks to measure performance improvement.


types/validation.go
```diff
@@ -273,11 +278,16 @@ func verifyCommitBatch(
 		// Validate signature.
 		voteSignBytes := commit.VoteSignBytes(chainID, int32(idx))
 
-		// add the key, sig and message to the verifier
-		if err := bv.Add(val.PubKey, voteSignBytes, commitSig.Signature); err != nil {
-			return err
+		cacheKey := string(val.Address) + string(voteSignBytes)
+		_, ok := verifiedSignatureCache[cacheKey]
```
ValarDragon, line 282: Nice! This has a malleabiltiy vector though if addresses weren't fixed size. Its also not zero-allocation, which is good to have at the sig verification level.

Perhaps structure the map using vote sign bytes as the key, and then the val address as the value? Not sure we can avoid the string casts for immutability reasons, so maybe we just live with the allocs. Should see how much it slows it down in the normal case.

Also can we allow passing in a nil map (skipping the check if nil)

wllmshao: Updated to make these changes.
- Using `string(voteSignBytes)` as the key and val address as the value
- Check `verifiedSignatureCache != nil` before any usage. Pass `nil` where ever the caching does not need to be used

types/validation.go
```diff
@@ -273,11 +279,19 @@ func verifyCommitBatch(
...
-		// add the key, sig and message to the verifier
-		if err := bv.Add(val.PubKey, voteSignBytes, commitSig.Signature); err != nil {
-			return err
+		cacheKey, cacheValAddress, sigIsInCache := "", []byte{}, false
+		if verifiedSignatureCache != nil {
+			cacheKey = string(voteSignBytes)
+			cacheValAddress, sigIsInCache = verifiedSignatureCache[cacheKey]
+		}
+
+		if !sigIsInCache || !bytes.Equal(cacheValAddress, commitSig.ValidatorAddress) {
```
ValarDragon, line 288: Can we make a single boolean for "cacheHit"

ValarDragon (commented):
LGTM, nice work! 

But we should add a tests around non-nil caches being provided, and tally logic for counting to 2/3rds still working!

types/validation.go
wllmshao: Good call, done.

Will work on adding tests.

wllmshao:
@ValarDragon thinking about this more, aren't the VoteSignBytes actually likely to be the same across signatures? I see a comment that says they differ only in the timestamp, the rest of the fields are identical. Would it not be better to use the validator address or the commit signature as the cache key?

ValarDragon:
So we need to be sure that all 3 arguments for `Verify(sign_bytes, signature, public_key) ` are the same, ideally with 0 allocations.

The key being the signature makes sense to me, since your right sign_bytes can get re-used. And then the value of the map being struct{sign_bytes, public_key} perhaps, where we check those two fields

types/validation.go
```diff
@@ -273,11 +284,19 @@ func verifyCommitBatch(
 		// Validate signature.
 		voteSignBytes := commit.VoteSignBytes(chainID, int32(idx))
 
-		// add the key, sig and message to the verifier
-		if err := bv.Add(val.PubKey, voteSignBytes, commitSig.Signature); err != nil {
-			return err
+		cacheHit := false
+		if verifiedSignatureCache != nil {
+			cacheVal, sigIsInCache := verifiedSignatureCache[string(commitSig.Signature)]
+			cacheHit = sigIsInCache && bytes.Equal(cacheVal.ValidatorAddress, val.PubKey.Address()) && bytes.Equal(cacheVal.VoteSignBytes, voteSignBytes)
```
wllmshao, line 290: this now assumes that the `val.PubKey` is uniquely identified by the `val.PubKey.Address()`, is that a safe assumption?
if it is not true, we will get false cache hits and will skip doing signature verification when we actually need to

wllmshao:
I added unit tests and changed the cache schema. Opening this for review.

.changelog/unreleased/improvements/2365-deduplicate-light-signature-checks.md
```diff
@@ -0,0 +1 @@
+- `[light]` Remove duplicated signature checks in light.VerifyNonAdjacent ([\#2365](https://github.com/cometbft/cometbft/issues/2365))
```
melekes, line 1: no need for the duplicate changelog

types/validation.go
```diff
@@ -18,6 +19,11 @@ func shouldBatchVerify(vals *ValidatorSet, commit *Commit) bool {
 		vals.AllKeysHaveSameType()
 }
 
+type SignatureCacheValue struct {
```
melekes, line 22: go docs are missing

types/validation.go
melekes: Can we store the validator's pubkey instead of the address to avoid making that assumption?

types/validator_set.go
```diff
@@ -731,9 +731,9 @@ func (vals *ValidatorSet) VerifyCommit(chainID string, blockID BlockID,
 // VerifyCommitLight verifies +2/3 of the set had signed the given commit.
 // It does NOT count all signatures.
 func (vals *ValidatorSet) VerifyCommitLight(chainID string, blockID BlockID,
```
melekes, line 733: Maybe it makes sense to add a new method `VerifyCommitLightWithCache` to avoid breaking the API

melekes (commented):
Thanks @wllmshao ❤️ Would love to see this merged

types/validation.go
wllmshao: added

.changelog/unreleased/improvements/2365-deduplicate-light-signature-checks.md
wllmshao: removed the one in breaking-changes since I have made this non-breaking

types/validation.go
wllmshao: Since validator pubkey is type `crypto.PubKey`, it seems we can't in general do an equality check with it, so we can't determine if two pubkeys match when checking for a cache hit. For example, `ed25519.PubKey` is uncomparable.
I've switched this to use the pubkey bytes, which might be better than address? but we can discuss further.

types/validation.go
```diff
@@ -144,13 +194,16 @@ func VerifyCommitLightTrusting(
 // This method DOES check all the signatures.
 //
 // CONTRACT: must run ValidateBasic() on commit before verifying.
+// The cache provided will be used to skip signature verification for entries where the
```
melekes, line 197: There's no cache here afaik

melekes:
```
--- FAIL: TestValidatorSet_VerifyCommitLightTrustingWithCache_UpdatesCache (0.00s)
    validation_test.go:317: 
        	Error Trace:	/home/runner/work/cometbft/cometbft/types/validation_test.go:317
        	Error:      	Not equal: 
        	            	expected: []byte{0x7e, 0x52, 0x35, 0xcf, 0x12, 0xf8, 0x77, 0xc2, 0xe, 0xe7, 0x29, 0xfb, 0x35, 0x6c, 0x74, 0x52, 0x9, 0x35, 0xeb, 0x40, 0x9f, 0x9a, 0xb7, 0x14, 0xee, 0x88, 0x71, 0x6, 0xf1, 0x31, 0xd7, 0xd}
        	            	actual  : []byte{0xb4, 0x7, 0xd, 0xa5, 0x70, 0x78, 0xc1, 0x33, 0x97, 0xeb, 0x35, 0x11, 0x67, 0x30, 0x8d, 0x7b, 0x76, 0x58, 0xba, 0xca, 0x1f, 0x27, 0x9, 0x84, 0x5c, 0x8f, 0xa5, 0x3f, 0xa1, 0x90, 0x34, 0x[19](https://github.com/cometbft/cometbft/actions/runs/11583831147/job/32259956806?pr=4322#step:6:20)}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,4 +1,4 @@
        	            	 ([]uint8) (len=32) {
        	            	- 00000000  7e 52 35 cf 12 f8 77 c2  0e e7 29 fb 35 6c 74 52  |~R5...w...).5ltR|
        	            	- 00000010  09 35 eb 40 9f 9a b7 14  ee 88 71 06 f1 31 d7 0d  |.5.@......q..1..|
        	            	+ 00000000  b4 07 0d a5 70 78 c1 33  97 eb 35 11 67 30 8d 7b  |....px.3..5.g0.{|
        	            	+ 00000010  76 58 ba ca 1f [27](https://github.com/cometbft/cometbft/actions/runs/11583831147/job/32259956806?pr=4322#step:6:28) 09 84  5c 8f a5 3f a1 90 34 19  |vX...'..\..?..4.|
        	            	 }
        	Test:       	TestValidatorSet_VerifyCommitLightTrustingWithCache_UpdatesCache
FAIL
```

types/validation.go
wllmshao: sorry, I put this in the wrong place, fixed.

wllmshao:
> ```
> --- FAIL: TestValidatorSet_VerifyCommitLightTrustingWithCache_UpdatesCache (0.00s)
>     validation_test.go:317: 
>         	Error Trace:	/home/runner/work/cometbft/cometbft/types/validation_test.go:317
>         	Error:      	Not equal: 
>         	            	expected: []byte{0x7e, 0x52, 0x35, 0xcf, 0x12, 0xf8, 0x77, 0xc2, 0xe, 0xe7, 0x29, 0xfb, 0x35, 0x6c, 0x74, 0x52, 0x9, 0x35, 0xeb, 0x40, 0x9f, 0x9a, 0xb7, 0x14, 0xee, 0x88, 0x71, 0x6, 0xf1, 0x31, 0xd7, 0xd}
>         	            	actual  : []byte{0xb4, 0x7, 0xd, 0xa5, 0x70, 0x78, 0xc1, 0x33, 0x97, 0xeb, 0x35, 0x11, 0x67, 0x30, 0x8d, 0x7b, 0x76, 0x58, 0xba, 0xca, 0x1f, 0x27, 0x9, 0x84, 0x5c, 0x8f, 0xa5, 0x3f, 0xa1, 0x90, 0x34, 0x[19](https://github.com/cometbft/cometbft/actions/runs/11583831147/job/32259956806?pr=4322#step:6:20)}
>         	            	
>         	            	Diff:
>         	            	--- Expected
>         	            	+++ Actual
>         	            	@@ -1,4 +1,4 @@
>         	            	 ([]uint8) (len=32) {
>         	            	- 00000000  7e 52 35 cf 12 f8 77 c2  0e e7 29 fb 35 6c 74 52  |~R5...w...).5ltR|
>         	            	- 00000010  09 35 eb 40 9f 9a b7 14  ee 88 71 06 f1 31 d7 0d  |.5.@......q..1..|
>         	            	+ 00000000  b4 07 0d a5 70 78 c1 33  97 eb 35 11 67 30 8d 7b  |....px.3..5.g0.{|
>         	            	+ 00000010  76 58 ba ca 1f [27](https://github.com/cometbft/cometbft/actions/runs/11583831147/job/32259956806?pr=4322#step:6:28) 09 84  5c 8f a5 3f a1 90 34 19  |vX...'..\..?..4.|
>         	            	 }
>         	Test:       	TestValidatorSet_VerifyCommitLightTrustingWithCache_UpdatesCache
> FAIL
> ```

There seemed to have been some nondeterminism with combining the valsets in this unit tests. Should be fixed now.

melekes (approved):

melekes:
Please don't force-push. It makes reviewing impossible

cason (commented):
I wonder whether the signatures cache should I type by itself, instead of using a `map[string]SignatureCacheValue`. This will probably render it simple to implement and test.

Moreover, I wonder whether the tests can indeed confirm that the cache is being used. I think that a mock signature verifier/key should be used in this case, so we are really sure that the verification methods are not invoked.

cason:
> This is breaking for both types and light, should my changelog reflect that? For now I've only mentioned light in the changelog.

If you are only adding methods, it is not API breaking. Can you fix the description?

cason:
> This PR assumes that pubkeys are uniquely identified by the Address. Is that safe?

Yes. The consensus logic uses it, for instance `proposerAddr := cs.privValidatorPubKey.Address()`. And the `Vote` type include the same `Address()` that should match the address in the validator set.

wllmshao:
> I wonder whether the signatures cache should I type by itself, instead of using a `map[string]SignatureCacheValue`. This will probably render it simple to implement and test.
> 
> Moreover, I wonder whether the tests can indeed confirm that the cache is being used. I think that a mock signature verifier/key should be used in this case, so we are really sure that the verification methods are not invoked.

Created a new type for `SignatureCache` to replace the map usage.
Added unit tests that use mocks to ensure that the cache is being used and signature verifications are not happening.

types/signature_cache.go
```diff
@@ -0,0 +1,36 @@
...
+	Add(key string, value SignatureCacheValue)
+	Get(key string) (SignatureCacheValue, bool)
+	Len() int
+}
+
+type signatureCache struct {
+	cache map[string]SignatureCacheValue
+}
+
+func NewSignatureCache() SignatureCache {
```
melekes, line 19: I don't see how this is better than the original `map[string]SignatureCacheValue`. @cason is this what you had in mind?

types/signature_cache.go
```diff
@@ -0,0 +1,36 @@
+package types
+
+// The value type for the verified signature cache.
+type SignatureCacheValue struct {
+	ValidatorPubKeyBytes []byte
```
melekes, line 5: We can use validator's address here

melekes (commented):
Thanks 🙏

types/signature_cache.go
wllmshao: updated

types/signature_cache.go
cason: Yes, so we can implement this abstraction in the way we prefer.

types/validation.go
```diff
@@ -298,22 +356,44 @@ func verifyCommitBatch(
...
 	ok, validSigs := bv.Verify()
 	if ok {
 		// success
+		if verifiedSignatureCache != nil {
+			for i := range validSigs {
+				idx := batchSigIdxs[i]
+				sig := commit.Signatures[idx]
+				verifiedSignatureCache.Add(string(sig.Signature), SignatureCacheValue{
+					ValidatorAddress: sig.ValidatorAddress,
+					VoteSignBytes:    commit.VoteSignBytes(chainID, int32(idx)),
```
cason, line 374: This call requires proto encoding, namely is kind of expensive. Don't we have these bytes already computed?

types/signature_cache.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+type SignatureCache interface {
+	Add(key string, value SignatureCacheValue)
+	Get(key string) (SignatureCacheValue, bool)
+	Len() int
+}
+
+type signatureCache struct {
+	cache map[string]SignatureCacheValue
```
cason, line 16: This cache grows over time, should we somehow restrict its size?

.changelog/unreleased/improvements/2365-deduplicate-light-signature-checks.md
cason, line 1: Should we add a new changelog entry for the creation of the `SignatureCache` type?

cason (approved):
Thank you a lot for this effort.

There are some details that we should take care of, but I'd say that we can take it over from here.

Issue #2848, opened 2024-04-18
Find strategy to remove blockID marshalling from most addVote calls


ValarDragon:
### Summary

We see that addVote calls lead to lots of heap allocations, due to calls to blockID.Key(). However this is something that only needs to get computed once per block, and right now we are massively overpaying! I think we should find an elegant way to cache this, I am not sure what is the best way, just that it would lower the RAM allocations (and presumably net time)

![image](https://github.com/cometbft/cometbft/assets/6440154/15b41c2a-5978-44d1-95f9-b8d881ed3f23)

### Problem Definition

This is adding performance cost to every addVote routine

### Proposal

Some potential ways to fix it:
- Make a global BlockId -> BlockIdKey cache? (Seems silly)
- Make every BlockId struct just contain the key, so it keeps getting copied?

cason:
Not easy... Although I don't really know why we use the block hash and the proto-encoded version of the block part set to produce the key... As the former would be enough.

cason:
> Make every BlockId struct just contain the key, so it keeps getting copied?

It is a public type, included in votes, proposals, etc., etc. Not easy to add fields to it. In any case, whenever we receive a message with a BlockID, a new instance of it is created, we don't re-use them when they match.

ValarDragon:
> How would you uniquely identify a BlockID? By the Hash (which end-ups being a []byte)?

That seems ideal to me, as long as the block hash also commits the part set header. (Which I think it does? But admittedly I haven't poked around much) 

Otherwise we could also make the key `hash || partsetheader.hash`.

Eric-Warehime:
Opened a PR which reduces the allocations slightly. I've toyed around with some versions which go further and actually cache the Key value, but there does seem to be some assumption in the repo right now that all of the fields of the BlockID are equivalent, and I'm not 100% confident that ignoring the part set header is non-breaking.

Happy to take another look at doing that if that's the case though.

cason:
I think that caching the `BlockID` ID (ehe) would be the best way to go, but I don't see an obvious way to implement this.

PR #4324, opened 2024-10-24
perf(types/vote_set): Remove BlockID.Key usage for equality in AddVote

Eric-Warehime:
Partially addresses #2848 by removing the calls to `Key` in `AddVote` when doing equality checks.

Since `Equals checks compare all of the fields in the BlockID and nested BlockPartSet, this is functionally the same and avoids the allocations that Key incurs.

This includes a benchmark that shows the reduction in memory allocations.


melekes (commented):
no need to pass `blockID` if we can get it from `vote`

melekes (approved):
Thanks @Eric-Warehime ❤️

ValarDragon:
Nice, ty!

PR #4340, opened 2024-10-25
perf(log)!: `LazyHash` for producing txs and blocks debug hashes

### LINKED ISSUE #4167: mempool: `log.NewLazySprintf("%X", entry.Tx().Hash())` does not improve performance
The idea of using `log.NewLazySprintf` is to only parse the arguments if the debug message is produced, depending of the debug level. It works as long as the entries have a `String()` implementation that is only invoked when the log message is produced.

However, in this case, we are calling `tx.Hash()`, which expensive, even when debug level > debug.

              However, we are still paying the cost of `tx.Hash()` even without debug.

This is the same as in the rest of this code. Namely, we aren't saving anything here.

_Originally posted by @cason in https://github.com/cometbft/cometbft/pull/4127#discussion_r1772861533_

wesl-ee:
Closes #4167 

Generalizes LazyBlockHash as LazyHash so now Txs and Blocks are hashed only on evaluation of the debug printf

#go-api-breaking

### DIFF
--- a/.changelog/unreleased/breaking-changes/4340-lazy-hash.md
+++ b/.changelog/unreleased/breaking-changes/4340-lazy-hash.md
@@ -1,4 +1,6 @@
 - `[log]` LazyBlockHash -> LazyHash
-  * LazyHash lazily stringifies things with a `Hash` method
-  * tx `Hash` ret type changed to HexBytes to fit interface
+  * LazyBlockHash replaced with more generic LazyHash which lazily evaluates
+    a tx or block hash when the stringer interface is invoked. Good for use
+    with debug statements so the item is only hashed when print is invoked
+  * tx `Hash` ret type changed to HexBytes to fit this interface
   [\#4340](https://github.com/cometbft/cometbft/pull/4340)
--- a/.changelog/unreleased/breaking-changes/4340-lazy-hash.md
+++ b/.changelog/unreleased/breaking-changes/4340-lazy-hash.md
@@ -1,0 +1,4 @@
+- `[log]` LazyBlockHash -> LazyHash
+  * LazyHash lazily stringifies things with a `Hash` method
+  * tx `Hash` ret type changed to HexBytes to fit interface
+  [\#4340](https://github.com/cometbft/cometbft/pull/4340)
--- a/libs/log/lazy.go
+++ b/libs/log/lazy.go
@@ -30,8 +30,8 @@ type hashable interface {
 	Hash() cmtbytes.HexBytes
 }
 
-// NewLazyHash defers Hash until the Stringer interface is invoked. This is
-// particularly useful for avoiding calling Sprintf when debugging is not
+// NewLazyHash defers calling `Hash()` until the Stringer interface is invoked.
+// This is particularly useful for avoiding calling Sprintf when debugging is not
 // active.
 func NewLazyHash(inner hashable) *LazyHash {
 	return &LazyHash{inner}
--- a/libs/log/lazy.go
+++ b/libs/log/lazy.go
@@ -22,21 +22,21 @@ func (l *LazySprintf) String() string {
 	return fmt.Sprintf(l.format, l.args...)
 }
 
-type LazyBlockHash struct {
-	block hashable
+type lazyHash struct {
+	inner hashable
 }
 
 type hashable interface {
 	Hash() cmtbytes.HexBytes
 }
 
-// NewLazyBlockHash defers block Hash until the Stringer interface is invoked.
-// This is particularly useful for avoiding calling Sprintf when debugging is not
+// NewLazyHash defers Hash until the Stringer interface is invoked. This is
+// particularly useful for avoiding calling Sprintf when debugging is not
 // active.
-func NewLazyBlockHash(block hashable) *LazyBlockHash {
-	return &LazyBlockHash{block}
+func NewLazyHash(block hashable) *lazyHash {
+	return &lazyHash{block}
 }
 
-func (l *LazyBlockHash) String() string {
-	return l.block.Hash().String()
+func (l *lazyHash) String() string {
+	return l.inner.Hash().String()
 }
--- a/libs/log/lazy.go
+++ b/libs/log/lazy.go
@@ -22,7 +22,7 @@ func (l *LazySprintf) String() string {
 	return fmt.Sprintf(l.format, l.args...)
 }
 
-type lazyHash struct {
+type LazyHash struct {
 	inner hashable
 }
 
@@ -33,10 +33,10 @@ type hashable interface {
 // NewLazyHash defers Hash until the Stringer interface is invoked. This is
 // particularly useful for avoiding calling Sprintf when debugging is not
 // active.
-func NewLazyHash(inner hashable) *lazyHash {
-	return &lazyHash{inner}
+func NewLazyHash(inner hashable) *LazyHash {
+	return &LazyHash{inner}
 }
 
-func (l *lazyHash) String() string {
+func (l *LazyHash) String() string {
 	return l.inner.Hash().String()
 }
--- a/libs/log/lazy.go
+++ b/libs/log/lazy.go
@@ -22,6 +22,10 @@ func (l *LazySprintf) String() string {
 	return fmt.Sprintf(l.format, l.args...)
 }
 
+// LazyHash is a wrapper around a hashable object that defers the Hash call
+// until the Stringer interface is invoked.
+// This is particularly useful for avoiding calling Sprintf when debugging is
+// not active.
 type LazyHash struct {
 	inner hashable
 }
--- a/libs/log/lazy.go
+++ b/libs/log/lazy.go
@@ -33,8 +33,8 @@ type hashable interface {
 // NewLazyHash defers Hash until the Stringer interface is invoked. This is
 // particularly useful for avoiding calling Sprintf when debugging is not
 // active.
-func NewLazyHash(block hashable) *lazyHash {
-	return &lazyHash{block}
+func NewLazyHash(inner hashable) *lazyHash {
+	return &lazyHash{inner}
 }
 
 func (l *lazyHash) String() string {
--- a/libs/log/lazy_test.go
+++ b/libs/log/lazy_test.go
@@ -1,0 +1,25 @@
+package log_test
+
+import (
+	"testing"
+
+	"github.com/cometbft/cometbft/internal/test"
+	"github.com/cometbft/cometbft/libs/log"
+)
+
+func TestLazyHash_Txs(t *testing.T) {
+	const height = 2
+	const numTxs = 5
+	txs := test.MakeNTxs(height, numTxs)
+
+	for i := 0; i < numTxs; i++ {
+		hash := txs[i].Hash()
+		lazyHash := log.NewLazyHash(txs[i])
+		if lazyHash.String() != hash.String() {
+			t.Fatalf("expected %s, got %s", hash.String(), lazyHash.String())
+		}
+		if len(hash) <= 0 {
+			t.Fatalf("expected non-empty hash, got empty hash")
+		}
+	}
+}
--- a/libs/log/lazy_test.go
+++ b/libs/log/lazy_test.go
@@ -13,12 +13,11 @@ func TestLazyHash_Txs(t *testing.T) {
 	txs := test.MakeNTxs(height, numTxs)
 
 	for i := 0; i < numTxs; i++ {
-		hash := txs[i].Hash()
 		lazyHash := log.NewLazyHash(txs[i])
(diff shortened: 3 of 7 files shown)


mempool/clist_mempool.go
```diff
@@ -790,7 +790,7 @@ func (mem *CListMempool) Update(
 		// https://github.com/tendermint/tendermint/issues/3322.
 		if err := mem.RemoveTxByKey(tx.Key()); err != nil {
 			mem.logger.Debug("Committed transaction not in local mempool (not an error)",
-				"tx", log.NewLazySprintf("%X", tx.Hash()),
+				"tx", fmt.Sprintf("%X", log.NewLazyHash(tx)),
```
melekes, line 793:
```suggestion
				"tx", log.NewLazyHash(tx),
```

mempool/clist_mempool.go
```diff
@@ -629,7 +629,7 @@ func (mem *CListMempool) handleRecheckTxResponse(tx types.Tx) func(res *abci.Res
 		// If tx is invalid, remove it from the mempool and the cache.
 		if (res.Code != abci.CodeTypeOK) || postCheckErr != nil {
 			// Tx became invalidated due to newly committed block.
-			mem.logger.Debug("Tx is no longer valid", "tx", log.NewLazySprintf("%X", tx.Hash()), "res", res, "postCheckErr", postCheckErr)
+			mem.logger.Debug("Tx is no longer valid", "tx", fmt.Sprintf("%X", log.NewLazyHash(tx)), "res", res, "postCheckErr", postCheckErr)
```
melekes, line 632:
```suggestion
			mem.logger.Debug("Tx is no longer valid", "tx", log.NewLazyHash(tx), "res", res, "postCheckErr", postCheckErr)
```

mempool/clist_mempool.go
```diff
@@ -610,7 +610,7 @@ func (mem *CListMempool) handleRecheckTxResponse(tx types.Tx) func(res *abci.Res
 
 		// Check whether the rechecking process has finished.
 		if mem.recheck.done() {
-			mem.logger.Error("Failed to recheck tx", "tx", log.NewLazySprintf("%X", tx.Hash()), "err", ErrLateRecheckResponse)
+			mem.logger.Error("Failed to recheck tx", "tx", fmt.Sprintf("%X", log.NewLazyHash(tx)), "err", ErrLateRecheckResponse)
```
melekes, line 613:
```suggestion
			mem.logger.Error("Failed to recheck tx", "tx", log.NewLazyHash(tx), "err", ErrLateRecheckResponse)
```

mempool/clist_mempool.go
```diff
@@ -548,7 +548,7 @@ func (mem *CListMempool) RemoveTxByKey(txKey types.TxKey) error {
 
 	mem.logger.Debug(
 		"Removed transaction",
-		"tx", log.NewLazySprintf("%X", memTx.tx.Hash()),
+		"tx", fmt.Sprintf("%X", log.NewLazyHash(memTx.tx)),
```
melekes, line 551:
```suggestion
		"tx", log.NewLazyHash(memTx.tx),
```

mempool/clist_mempool.go
```diff
@@ -511,7 +511,7 @@ func (mem *CListMempool) addTx(tx types.Tx, gasWanted int64, sender nodekey.ID,
 
 	mem.logger.Debug(
 		"Added transaction",
-		"tx", log.NewLazySprintf("%X", tx.Hash()),
+		"tx", fmt.Sprintf("%X", log.NewLazyHash(tx)),
```
melekes, line 514:
```suggestion
		"tx", log.NewLazyHash(tx),
```

mempool/clist_mempool.go
```diff
@@ -449,7 +449,7 @@ func (mem *CListMempool) handleCheckTxResponse(tx types.Tx, sender nodekey.ID) f
 			if err := mem.addSender(txKey, sender); err != nil {
 				mem.logger.Error("Could not add sender to tx", "tx", tx.Hash(), "sender", sender, "err", err)
 			}
-			mem.logger.Debug("Reject tx", "tx", log.NewLazySprintf("%X", tx.Hash()), "height", mem.height.Load(), "err", ErrTxInMempool)
+			mem.logger.Debug("Reject tx", "tx", fmt.Sprintf("%X", log.NewLazyHash(tx)), "height", mem.height.Load(), "err", ErrTxInMempool)
```
melekes, line 452: same

mempool/clist_mempool.go
```diff
@@ -412,7 +412,7 @@ func (mem *CListMempool) handleCheckTxResponse(tx types.Tx, sender nodekey.ID) f
 			mem.tryRemoveFromCache(tx)
 			mem.logger.Debug(
 				"Rejected invalid transaction",
-				"tx", log.NewLazySprintf("%X", tx.Hash()),
+				"tx", fmt.Sprintf("%X", log.NewLazyHash(tx)),
```
melekes, line 415: same

mempool/clist_mempool.go
```diff
@@ -399,7 +399,7 @@ func (mem *CListMempool) handleCheckTxResponse(tx types.Tx, sender nodekey.ID) f
 
 		// Check that rechecking txs is not in process.
 		if !mem.recheck.done() {
-			panic(log.NewLazySprintf("rechecking has not finished; cannot check new tx %X", tx.Hash()))
+			panic(fmt.Sprintf("rechecking has not finished; cannot check new tx %X", log.NewLazyHash(tx)))
```
melekes, line 402: same

mempool/clist_mempool.go
```diff
@@ -380,7 +380,7 @@ func (mem *CListMempool) CheckTx(tx types.Tx, sender nodekey.ID) (*abcicli.ReqRe
 		Type: abci.CHECK_TX_TYPE_CHECK,
 	})
 	if err != nil {
-		panic(fmt.Errorf("CheckTx request for tx %s failed: %w", log.NewLazySprintf("%X", tx.Hash()), err))
+		panic(fmt.Errorf("CheckTx request for tx %s failed: %w", fmt.Sprintf("%X", log.NewLazyHash(tx)), err))
```
melekes, line 383: same

mempool/clist_mempool.go
```diff
@@ -367,7 +367,7 @@ func (mem *CListMempool) CheckTx(tx types.Tx, sender nodekey.ID) (*abcicli.ReqRe
 		// (eg. after committing a block, txs are removed from mempool but not cache),
 		// so we only record the sender for txs still in the mempool.
 		if err := mem.addSender(tx.Key(), sender); err != nil {
-			mem.logger.Error("Could not add sender to tx", "tx", log.NewLazySprintf("%X", tx.Hash()), "sender", sender, "err", err)
+			mem.logger.Error("Could not add sender to tx", "tx", fmt.Sprintf("%X", log.NewLazyHash(tx)), "sender", sender, "err", err)
```
melekes, line 370: same

mempool/clist_mempool.go
```diff
@@ -855,7 +855,7 @@ func (mem *CListMempool) recheckTxs() {
 			Type: abci.CHECK_TX_TYPE_RECHECK,
 		})
 		if err != nil {
-			panic(fmt.Errorf("(re-)CheckTx request for tx %s failed: %w", log.NewLazySprintf("%X", memTx.Tx().Hash()), err))
+			panic(fmt.Errorf("(re-)CheckTx request for tx %s failed: %w", fmt.Sprintf("%X", log.NewLazyHash(memTx.Tx())), err))
```
melekes, line 858: same

melekes (commented):
Thanks @wesl-ee ❤️ 

I think we can now remove fmt.Sprintf because it's `HexBytes`

melekes:
```

libs/log/lazy.go:36:34: unexported-return: exported func NewLazyHash returns unexported type *log.lazyHash, which can be annoying to use (revive)
func NewLazyHash(inner hashable) *lazyHash {
                                 ^
test/e2e/tests/app_test.go:96:47: unnecessary conversion (unconvert)
		require.Equal(t, res.Hash, cmtbytes.HexBytes(hash))
```

andynog (changes requested):
Thanks @wesl-ee. Overall looks good to me. Could you please check @melekes suggestions ? Also, since this seems to be breaking change, could you also please add a changelog entry.

mempool/clist_mempool.go
```diff
@@ -399,7 +399,7 @@ func (mem *CListMempool) handleCheckTxResponse(tx types.Tx, sender nodekey.ID) f
 
 		// Check that rechecking txs is not in process.
 		if !mem.recheck.done() {
-			panic(fmt.Sprintf("rechecking has not finished; cannot check new tx %X", log.NewLazyHash(tx)))
+			panic(fmt.Sprint("rechecking has not finished; cannot check new tx", log.NewLazyHash(tx)))
```
melekes, line 402:
```suggestion
			panic(fmt.Sprint("rechecking has not finished; cannot check new tx ", log.NewLazyHash(tx)))
```

libs/log/lazy.go
```diff
@@ -22,7 +22,7 @@ func (l *LazySprintf) String() string {
 	return fmt.Sprintf(l.format, l.args...)
 }
 
-type lazyHash struct {
+type LazyHash struct {
```
melekes, line 25: would be great to add Go docs here

.changelog/unreleased/breaking-changes/4340-lazy-hash.md
cason, line 1: Can you improve the description here?

cason: You are making a more generic type `LazyHash` from the type `LazyBlockHash` for all types that are "hashable"

mempool/clist_mempool.go
cason: We don't need to call a lazy log when panicing...

types/tx.go
```diff
@@ -26,7 +26,7 @@ type (
 )
 
 // Hash computes the TMHASH hash of the wire encoded transaction.
-func (tx Tx) Hash() []byte {
```
cason, line 29: This is a breaking change, isn't it?

libs/log/lazy.go
```diff
@@ -22,21 +22,21 @@ func (l *LazySprintf) String() string {
...
+	inner hashable
 }
 
 type hashable interface {
 	Hash() cmtbytes.HexBytes
 }
 
-// NewLazyBlockHash defers block Hash until the Stringer interface is invoked.
-// This is particularly useful for avoiding calling Sprintf when debugging is not
+// NewLazyHash defers Hash until the Stringer interface is invoked. This is
```
cason, line 33:
```suggestion
// NewLazyHash defers calling `Hash()` until the Stringer interface is invoked.
```

libs/log/lazy.go
```diff
@@ -22,21 +22,21 @@ func (l *LazySprintf) String() string {
...
+// particularly useful for avoiding calling Sprintf when debugging is not
```
cason, line 34:
```suggestion
// This is particularly useful for avoiding calling Sprintf when debugging is not
```

libs/log/lazy_test.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/cometbft/cometbft/libs/log"
+)
+
+func TestLazyHash_Txs(t *testing.T) {
+	const height = 2
+	const numTxs = 5
+	txs := test.MakeNTxs(height, numTxs)
+
+	for i := 0; i < numTxs; i++ {
+		hash := txs[i].Hash()
```
cason, line 16: Can you change the order here?

We want to make sure that line 17 does not call `Hash()`.

cason (commented):
Thank you for this, the solution looks very elegant.

Left some comments.

libs/log/lazy_test.go
cason: Actually, we should not call `Hash()` at all here, you can do that in the `if`.

types/tx.go
wesl-ee: Actually places that use it didn't seem to break for me (see app_test.go change) so it's not a breaking change in this sense. I think I didn't need to make any changes elsewhere since the type HexBytes is a `[]byte` type with methods on it defined.

cason: But you have added a `!` to the title, meaning that it is breaking.

I am ok with this change, provided that nothing breaks for outside users.

cason: Leave as-is.

andynog (commented):
thanks for adding the changelog entry

andynog (approved):

melekes (approved):
👍

PR #4348, opened 2024-10-28
feat(e2e): Support latency emulation on 'fast' Docker image

hvanz:
On #4408 we removed the dependency on Python to run the latency emulation script, by replacing it for a shell script with just `tc` (traffic control) commands. Now we are able to run the script in the slim Docker image generated with `make fast`.

Note that the base image of `Dockerfile.fast` was changed from `distroless/static-debian11` to `alpine`, because it is easier to install `tc` and run scripts from `alpine`. The difference in the image size is not really significant: 48.5MB in distroless vs 56.7MB in alpine.

---

#### PR checklist

- [ ] Tests written/updated


andynog:
Happy to review this @hvanz when it's ready for review

cason:
This PR does much more than its title makes us to expect, ehe.

We should discuss at some point whether it makes sense to keep the `fast` version of the Docker.

hvanz:
I split the original PR in two. The refactoring of removing the Python script is in #4408. This PR builds on top of that one.

melekes (approved):
👍

cason (approved):
Legit.

PR #4349, opened 2024-10-28
chore(rpc): optimize genesis file chunking

### LINKED ISSUE #1289: Optimize genesis file chunking

alesforz:
Closes #1289.

#### `Environment` type
The `Environment` type will:
- store the genesis file's path on disk
- store a `map[int]string` where the keys are the chunks' IDs, and the values are the chunk's path on disk.
- not store a pointer to a `types.GenesisDoc` anymore
- not store a slice of genesis file chunks anymore. Instead the method `InitGenesisChunks` will create chunks and store them on disk, 1 file per chunk.

#### `Node` type
The `Node` type will:
- not store a pointer to a `GenesisDoc` anymore. 
- pass the genesis file's path on disk to a new `Environment` object when creating it in the `ConfigureRPC()` method.

#### RPC `/genesis_chunked`
This endpoint will implement the chunking strategy described [here](https://github.com/cometbft/cometbft/issues/1290#issuecomment-2377326927).

---

#### PR checklist

- [x] Tests written/updated


rpc/core/env.go
```diff
@@ -105,70 +109,82 @@ type Environment struct {
...
 		// This should not happen.
-		return errors.New("could not create the genesis file chunks and cache them because the genesis doc is unavailable")
+		return errors.New("missing genesis file path on disk")
 	}
 
-	data, err := cmtjson.Marshal(env.GenDoc)
+	gFileSize, err := fileSize(gFilePath)
 	if err != nil {
-		return fmt.Errorf("encoding genesis doc to JSON: %w", err)
+		return fmt.Errorf("gauging genesis file size: %s", err)
```
melekes, line 149:
```suggestion
		return fmt.Errorf("gauging genesis file size: %w", err)
```

rpc/core/env.go
```diff
@@ -105,70 +109,82 @@ type Environment struct {
...
-		var (
-			start = i * genesisChunkSize
-			end   = start + genesisChunkSize
-		)
-		if end > len(data) {
-			end = len(data)
-		}
+	gChunksDir, err := mkChunksDir(gFilePath, _chunksDir)
+	if err != nil {
+		return fmt.Errorf("preparing chunks directory: %s", err)
```
melekes, line 159:
```suggestion
		return fmt.Errorf("preparing chunks directory: %w", err)
```

rpc/core/env.go
```diff
@@ -105,70 +109,82 @@ type Environment struct {
...
-		// we make a copy here so that the original data isn't retained in memory.
-		// The GC will collect the data slice after exiting the function.
-		// Without the copy, it would keep it in memory.
-		chunk := make([]byte, end-start)
-		copy(chunk, data[start:end])
-		chunks[i] = base64.StdEncoding.EncodeToString(chunk)
+	// chunking required
+	chunkIDToPath, err := writeChunks(gFilePath, gChunksDir, genesisChunkSize)
+	if err != nil {
+		return fmt.Errorf("chunking large genesis file: %s", err)
```
melekes, line 165:
```suggestion
		return fmt.Errorf("chunking large genesis file: %w", err)
```

rpc/core/env.go
```diff
@@ -105,70 +109,82 @@ type Environment struct {
...
+// We call the function:
+//   - before creating new genesis file chunks, to make sure we start with a clean
+//     directory.
+//   - when a Node shuts down, to clean up the file system.
+func (env *Environment) Cleanup() error {
+	gFileDir := filepath.Dir(env.GenesisFilePath)
+	chunksDir := filepath.Join(gFileDir, _chunksDir)
+
+	if err := os.RemoveAll(chunksDir); err != nil {
+		return fmt.Errorf("deleting genesis file chunks' folder: %s", err)
```
melekes, line 186:
```suggestion
		return fmt.Errorf("deleting genesis file chunks' folder: %w", err)
```

rpc/core/env.go
```diff
@@ -245,3 +261,109 @@ func (env *Environment) latestUncommittedHeight() int64 {
...
+// fileSize returns the size of the file at the given path.
+func fileSize(fPath string) (int, error) {
+	// we use os.Stat here instead of os.ReadFile, because we don't want to load
+	// the entire file into memory just to compute its size from the resulting
+	// []byte slice.
+	fInfo, err := os.Stat(fPath)
+	if errors.Is(err, fs.ErrNotExist) {
+		return 0, fmt.Errorf("the file is unavailable at %s", fPath)
+	} else if err != nil {
+		return 0, fmt.Errorf("accessing file: %s", err)
```
melekes, line 274:
```suggestion
		return 0, fmt.Errorf("accessing file: %w", err)
```

rpc/core/env.go
```diff
@@ -245,3 +261,109 @@ func (env *Environment) latestUncommittedHeight() int64 {
...
+func mkChunksDir(gFilePath string, dirName string) (string, error) {
+	var (
+		gFileDir = filepath.Dir(gFilePath)
+		dirPath  = filepath.Join(gFileDir, dirName)
+	)
+	if _, err := os.Stat(dirPath); err == nil {
+		// directory already exists; this might happen it the node crashed and
+		// could not do cleanup. Delete it to start from scratch.
+		if err := os.RemoveAll(dirPath); err != nil {
+			return "", fmt.Errorf("deleting existing chunks directory: %s", err)
```
melekes, line 295:
```suggestion
			return "", fmt.Errorf("deleting existing chunks directory: %w", err)
```

rpc/core/env.go
```diff
@@ -245,3 +261,109 @@ func (env *Environment) latestUncommittedHeight() int64 {
...
+		}
+	} else if !os.IsNotExist(err) {
+		return "", fmt.Errorf("accessing directory: %s", err)
```
melekes, line 298:
```suggestion
		return "", fmt.Errorf("accessing directory: %w", err)
```

rpc/core/env.go
```diff
@@ -245,3 +261,109 @@ func (env *Environment) latestUncommittedHeight() int64 {
...
+	}
+
+	if err := os.Mkdir(dirPath, 0o755); err != nil {
```
melekes, line 301: why 755?

rpc/core/env.go
```diff
@@ -245,3 +261,109 @@ func (env *Environment) latestUncommittedHeight() int64 {
...
+
+// writeChunk writes a chunk of the genesis file to disk, saving it to dir.
+// Each chunk file name's format will be: chunk_[chunkID].part, e.g., chunk_42.part.
+func writeChunk(chunk []byte, dir string, chunkID int) (string, error) {
+	var (
+		chunkName = "chunk_" + strconv.Itoa(chunkID) + ".part"
+		chunkPath = filepath.Join(dir, chunkName)
+	)
+	if err := os.WriteFile(chunkPath, chunk, 0o600); err != nil {
+		return "", fmt.Errorf("writing chunk to disk: %s", err)
```
melekes, line 316:
```suggestion
		return "", fmt.Errorf("writing chunk to disk: %w", err)
```

rpc/core/env.go
```diff
@@ -245,3 +261,109 @@ func (env *Environment) latestUncommittedHeight() int64 {
...
+				break
+			}
+
+			formatStr := "chunk %d: reading genesis file: %s"
+			return nil, fmt.Errorf(formatStr, chunkID, err)
+		}
+
+		chunkPath, err := writeChunk(buf[:n], gChunksDir, chunkID)
+		if err != nil {
+			return nil, fmt.Errorf("chunk %d: %s", chunkID, err)
```
melekes, line 362:
```suggestion
			return nil, fmt.Errorf("chunk %d: %w", chunkID, err)
```

rpc/core/env.go
```diff
@@ -245,3 +261,109 @@ func (env *Environment) latestUncommittedHeight() int64 {
...
+		chunkIDToPath = make(map[int]string)
+	)
+	for chunkID := 0; ; chunkID++ {
+		n, err := gFile.Read(buf)
+		if err != nil {
+			if errors.Is(err, io.EOF) {
+				break
+			}
+
+			formatStr := "chunk %d: reading genesis file: %s"
```
melekes, line 356:
```suggestion
			formatStr := "chunk %d: reading genesis file: %w"
```

rpc/core/errors.go
```diff
@@ -17,7 +17,7 @@ var (
 	ErrTimedOutWaitingForTx    = errors.New("timed out waiting for tx to be included in a block")
 	ErrGenesisRespSize         = errors.New("genesis response is too large, please use the genesis_chunked API instead")
 	ErrChunkNotInitialized     = errors.New("genesis chunks are not initialized")
-	ErrNoChunks                = errors.New("no chunks")
+	ErrNoChunks                = errors.New("genesis file is small, therefore there are no chunks to serve. Please use the /genesis API instead")
```
melekes, line 20: Can we make it so that if the genesis file is too small, we just return it when the first chunk is assessed? If so, we could deprecate `/genesis` completely. 

From the app developer's point of view, it's cumbersome to have to switch from one endpoint to another.

rpc/core/net.go
```diff
@@ -109,28 +111,40 @@ func (env *Environment) UnsafeDialPeers(
...
 		// See discussion in the following PR for why we still serve chunk 0 even
 		// if env.genChunks is nil:
 		// https://github.com/cometbft/cometbft/pull/4235#issuecomment-2389109521
-		if chunk == 0 {
-			genesisJSON, err := cmtjson.Marshal(env.GenDoc)
+		if chunkID == 0 {
+			fGenesis, err := os.ReadFile(env.GenesisFilePath)
 			if err != nil {
-				return nil, fmt.Errorf("retrieving requested chunk (id=0): %s", err)
+				return nil, fmt.Errorf("retrieving genesis file from disk: %s", err)
```
melekes, line 143:
```suggestion
				return nil, fmt.Errorf("retrieving genesis file from disk: %w", err)
```

rpc/core/net.go
```diff
@@ -140,23 +154,27 @@ func (env *Environment) GenesisChunked(
...
+	if id > len(env.genesisChunks)-1 {
+		return nil, ErrInvalidChunkID{id, len(env.genesisChunks) - 1}
+	}
 
-	if id > len(env.genChunks)-1 {
-		return nil, ErrInvalidChunkID{id, len(env.genChunks) - 1}
+	chunkPath := env.genesisChunks[id]
+	chunk, err := os.ReadFile(chunkPath)
+	if err != nil {
+		return nil, fmt.Errorf("retrieving chunk %d from disk: %s", id, err)
```
melekes, line 169:
```suggestion
		return nil, fmt.Errorf("retrieving chunk %d from disk: %w", id, err)
```

melekes (commented):
Thanks @alesforz ❤️ 

main q: https://github.com/cometbft/cometbft/pull/4349#discussion_r1820229574

rpc/core/errors.go
alesforz: It already behaves like that 😄 when the genesis file is small, querying the `/genesis_chunked` endpoint with `chunkID=0` will return the entire genesis file encoded in `base64`. See [here](https://github.com/cometbft/cometbft/blob/6927e7f967b919a0084edf08db9802aade59cb63/rpc/core/net.go#L136) .

The API returns `ErrNoChunks` when the genesis file is small and you query `genesis_chunked` with `chunkID >= 1`. The previous error msg was just `"no chunks"` which I thought wasn't very informative.

rpc/core/env.go
alesforz: 755 should be read, write, and execute permission to the owner, but only read and execute for others. 
The owner does need all three because we create and write the files to store in the directory at runtime, and later we want to read them.

Do you think it should be 700? i.e., no permissions for others?

melekes: I think it's fine, but yeah probably others don't need to see that.

melekes (approved):

rpc/core/env.go
alesforz: We can make the change 👍

.changelog/unreleased/improvements/4235-4234-environment-stores-one-genesis.md
```diff
@@ -1,3 +1 @@
-- `[rpc]` Store either a pointer to a GenesisDoc or the genesis' chunks, but not both in Environment (RPC API)
-  a GenesisDoc or the genesis' chunks, but not both.
-  ([\#4235](https://github.com/cometbft/cometbft/pull/4235))
\ No newline at end of file
+- `[rpc]` Store either a pointer to a GenesisDoc or the genesis' chunks, but not both in Environment (RPC API) ([\#4235](https://github.com/cometbft/cometbft/pull/4235))
```
cason, line 1:
```suggestion
- `[rpc]` Store either a pointer to a `GenesisDoc` or the genesis' chunks, but not both in Environment (RPC API) ([\#4235](https://github.com/cometbft/cometbft/pull/4235))
```

.github/workflows/gobenchdata-base.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         with:
           go-version: ${{ env.GO_VERSION }}
 
-      - run: go install github.com/bobheadxi/gobenchdata@v1.3.1
+      - run: go install go.bobheadxi.dev/gobenchdata@v1.3.1
```
cason, line 24: Why?

rpc/jsonrpc/server/http_server.go
cason, line 1: Here you probably should make the old methods to use use the new one with an empty `shutdownTasks` list.

rpc/core/env.go
```diff
@@ -105,70 +109,82 @@ type Environment struct {
...
-	genChunks []string
+	GenesisFilePath string // the genesis file's full path on disk
+
+	// genesisChunk is a map of chunk ID to its full path on disk.
+	// If the genesis file is smaller than genesisChunkSize, then this map will be
+	// nil, because there will be no chunks on disk.
+	// This map is convenient for the `/genesis_chunked` API to quickly find a chunk
+	// by its ID, instead of having to reconstruct its path each time, which would
+	// involve multiple string operations.
+	genesisChunks map[int]string
```
cason, line 120: `genesisChunksFiles` or something, for better readability?

rpc/core/env.go
```diff
@@ -105,70 +109,82 @@ type Environment struct {
...
-//     An `Environment` object will store a pointer to the genesis in its GenDoc
-//     field. Its genChunks field will be set to nil. `/genesis` RPC API will return
-//     the GenesisDoc itself.
-//   - if the genesis file's size is > genesisChunkSize, then use chunking. An
-//     `Environment` object will store a slice of base64-encoded chunks in its
-//     genChunks field. Its GenDoc field will be set to nil. `/genesis` RPC API will
-//     redirect users to use the `/genesis_chunked` API.
+//   - if the genesis file's size is <= genesisChunkSize, this function returns
+//     without doing anything. The `/genesis` RPC API endpoint will fetch the genesis
+//     from disk to serve requests.
```
cason, line 129:
```suggestion
//     file from disk to serve requests.
```

rpc/core/env.go
```diff
@@ -245,3 +261,109 @@ func (env *Environment) latestUncommittedHeight() int64 {
...
+
+// writeChunks reads the genesis file in chunks of size chunkSize, and writes them
+// to disk.
+// gFilePath is the genesis file's full path on disk.
+// gChunksDir is the directory where the chunks will be stored on disk.
+// chunkSize is the size of a chunk, that is, writeChunks will read the genesis file
+// in chunks of size chunkSize.
+// It returns a map where the keys are the chunk IDs, and the values are the chunks'
+// path on disk. E.g.,:
+// map[0] = /users/user/.cometbft/config/genesis-chunks/chunk_0.part
```
cason, line 330:
```suggestion
// map[0] = $HOME/.cometbft/config/genesis-chunks/chunk_0.part
```

rpc/core/env.go
```diff
@@ -245,3 +261,109 @@ func (env *Environment) latestUncommittedHeight() int64 {
...
+// map[1] = /users/user/.cometbft/config/genesis-chunks/chunk_1.part
```
cason, line 331:
```suggestion
// map[1] = $HOME/.cometbft/config/genesis-chunks/chunk_1.part
```

cason (commented):
Left some generic comments, looks like a good feature, but this PR includes some stuff not directly associated to it.

alesforz:
> Left some generic comments, looks like a good feature, but this PR includes some stuff not directly associated to it.

🤔 What parts do you think aren't associated with the feature?

rpc/jsonrpc/server/http_server.go
alesforz: Are you referring to other parts of the codebase that currently call `Serve` or `ServeTLS`? I didn't modify them because `Serve` and `ServeTLS` are exported methods and can be called from anywhere.

With the changes in this PR, the `rpc/jsonrpc/server` package provides `Serve`, `ServeWithShutdown`, `ServeTLS`, and `ServeTLSWithShutdown` as APIs. I believe it makes sense for callers to choose the method that best suits their needs. For example, if they need to execute additional tasks at shutdown, they should use `ServeWithShutdown`; if they don't need any extra work, they can use `Serve`.

In the case of `Node`, it requires some cleanup, so it calls the `...WithShutdown` versions of the methods.

Does it make sense or did you mean something else?

cason (approved):
The change was more comple

rpc/jsonrpc/server/http_server.go
cason: No, no, the new method and old method are identical except for the last lines.

You can refactor the old methods to call the new ones with an empty `shutdownTasks` list.

It is the same code, at the end, but with less duplication.

PR #4351, opened 2024-10-28
test(e2e): give the runner extra time to wait for a killed node

andynog:
One reason some e2e tests are failing is that the kill perturbation has a short timeout value of 20 seconds, causing waitForNode to time out. Doubling this timeout improves test stability and helps ensure the tests pass

[test failure](https://github.com/cometbft/cometbft/actions/runs/11546210646/job/32134226559)

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/test/e2e/runner/perturb.go
+++ b/test/e2e/runner/perturb.go
@@ -64,6 +64,10 @@ func PerturbNode(ctx context.Context, node *e2e.Node, perturbation e2e.Perturbat
 		}
 		if node.PersistInterval == 0 {
 			timeout *= 5
+		} else {
+			// still need to give some extra time to the runner
+			// to wait for the node to restart when killing
+			timeout *= 2
 		}
 
 	case e2e.PerturbationPause:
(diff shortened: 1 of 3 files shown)


melekes (approved):
Thanks @andynog ❤️

test/e2e/runner/perturb.go
```diff
@@ -64,6 +64,10 @@ func PerturbNode(ctx context.Context, node *e2e.Node, perturbation e2e.Perturbat
 		}
 		if node.PersistInterval == 0 {
 			timeout *= 5
+		} else {
```
cason, line 67: What is the rationale of this if?

I assume that when `PersistInterval == 0`, the application has no state so it should restart from genesis, is that right?

test/e2e/runner/rpc.go
```diff
@@ -92,9 +92,11 @@ func waitForNode(ctx context.Context, node *e2e.Node, height int64, timeout time
 			return nil, ctx.Err()
 		case <-timer.C:
 			status, err := client.Status(ctx)
+			sinceLastChanged := time.Since(lastChanged)
```
cason, line 95: You could just call `time.Since` again in the log message, but ok.

cason (approved):
We could maybe increase the standard timeout to 30s.

test/e2e/runner/perturb.go
andynog: to be honest, just looking at the PersistInterval field, I only see it used in validation and perturb logic but in `if` clauses, so I think this field might not have logic that drives the state persistence. We can revisit it in the future, my PR just adds more time to the timeout in this case.

Issue #3118, opened 2024-05-26
Speedup WAL encoding


ValarDragon:
### Summary

If we look at WAL encoding times, its significantly larger than just one disk write! Here is a CPU profile of it on a live-synced full node running for 1 hour on Osmosis:
![image](https://github.com/cometbft/cometbft/assets/6440154/f8435c4f-8154-4329-b3e5-97209ef42e53)

- If we look at the proto.Marshal time, its basically all spent in serializing time objects (Or finding the size of a time object).
- Consensus.WalToProto is paying lots of time to wrap every internal consensus message with pointers, pushing data to heap
- The makeslice call is done because we can't just directly do three writes to consensus WAL that are guaranteed to be in sequence. We could fix this with a simple `multiWrite` function on the autofile. 

```[tasklist]
### Tasks
- [ ] Add a draft title or issue reference here
```

### Problem Definition

This helps improve the speed of consensus, as many things go through the WAL. 

### Proposal

Work towards fixes towards aforementioned overheads (in addition to making fewer things write to WAL)

melekes:
see https://github.com/cometbft/cometbft/pull/4353#issuecomment-2449701437

PR #4353, opened 2024-10-28
perf: Remove External Votes From Wal

Eric-Warehime:
Don't write votes coming from external nodes to the WAL.

This will improve both CPU time spent in terms of blocking syscalls and churn in terms of memory allocation for vote buffers being flushed to disk. You can see the effects in these images:
Screenshot 2024-10-25 at 5 07 02 PM

Screenshot 2024-10-25 at 4 55 27 PM

Also adds a new e2e test to better exercise crash recovery. This introduces network-wide perturbations which allow us to kill or restart every node on the network at once. 

Follow ups for this PR:
* I want to introduce some additional configuration options to allow us to specify sleep times for concurrent perturbations.  
The kill and restart in the included test file ends up recovering from each one on the same round without advancing in between them since there is no delay to allow the nodes to produce another round before running the next perturbation.

---

#### PR checklist

- [X] Tests written/updated

### DIFF
--- a/internal/consensus/byzantine_test.go
+++ b/internal/consensus/byzantine_test.go
@@ -231,10 +231,10 @@ func TestByzantinePrevoteEquivocation(t *testing.T) {
 			proposal.Signature = p.Signature
 
 			// send proposal and block parts on internal msg queue
-			lazyProposer.sendInternalMessage(msgInfo{&ProposalMessage{proposal}, "", cmttime.Now()})
+			lazyProposer.sendInternalMessage(msgInfo{&ProposalMessage{proposal}, "", cmttime.Now(), true})
 			for i := 0; i < int(blockParts.Total()); i++ {
 				part := blockParts.GetPart(i)
-				lazyProposer.sendInternalMessage(msgInfo{&BlockPartMessage{lazyProposer.Height, lazyProposer.Round, part}, "", time.Time{}})
+				lazyProposer.sendInternalMessage(msgInfo{&BlockPartMessage{lazyProposer.Height, lazyProposer.Round, part}, "", time.Time{}, true})
 			}
 			lazyProposer.Logger.Info("Signed proposal", "height", height, "round", round, "proposal", proposal)
 			lazyProposer.Logger.Debug(fmt.Sprintf("Signed proposal block: %v", block))
--- a/internal/consensus/reactor.go
+++ b/internal/consensus/reactor.go
@@ -334,13 +334,13 @@ func (conR *Reactor) Receive(e p2p.Envelope) {
 		switch msg := msg.(type) {
 		case *ProposalMessage:
 			ps.SetHasProposal(msg.Proposal)
-			conR.conS.peerMsgQueue <- msgInfo{msg, e.Src.ID(), cmttime.Now()}
+			conR.conS.peerMsgQueue <- msgInfo{msg, e.Src.ID(), cmttime.Now(), true}
 		case *ProposalPOLMessage:
 			ps.ApplyProposalPOLMessage(msg)
 		case *BlockPartMessage:
 			ps.SetHasProposalBlockPart(msg.Height, msg.Round, int(msg.Part.Index))
 			conR.Metrics.BlockParts.With("peer_id", string(e.Src.ID())).Add(1)
-			conR.conS.peerMsgQueue <- msgInfo{msg, e.Src.ID(), time.Time{}}
+			conR.conS.peerMsgQueue <- msgInfo{msg, e.Src.ID(), time.Time{}, true}
 		default:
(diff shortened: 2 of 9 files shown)


tac0turtle:
ref https://github.com/cometbft/cometbft/issues/3118

cason:
We can't do that.

The consensus algorithm assumes that if an output or action is produced (e.g., we lock a `BlockID` and sign a `Precommit` for it), all the information required for producing this action or output is persisted.

Otherwise, when restarting the node without this information, the node can produce different outputs or actions (e.g., since we haven't locked that `BlockID`, we can sign a `Prevote` for a different value), which is an equivocation.

cason:
By the way, received messages are asynchronously written to the WAL (at almost zero cost).

We only flush the WAL when outputs or actions are produced (e.g., when we sign a `Precommit` in the same example). When this happens, we pay the cost of synchronously writing the output and all inputs that have lead the validator to produce this output (at least the ones that were not synced to disk by the operating system meanwhile).

Eric-Warehime:
>By the way, received messages are asynchronously written to the WAL (at almost zero cost).
> We only flush the WAL when outputs or actions are produced (e.g., when we sign a Precommit in the same example). When this happens, we pay the cost of synchronously writing the output and all inputs that have lead the validator to produce this output (at least the ones that were not synced to disk by the operating system meanwhile).

This is true, there's no fsync for these, but we do call `wal.Write` which is in the critical path for receiving any message in the consensus state. And that call does buffer allocation, protobuf encoding, CRC checksums, etc., which still incurs a cost from both a time and memory perspective, so if we can agree that a change like this is safe, the data suggests that it would have meaningful benefits to Comet's performance.

> Otherwise, when restarting the node without this information, the node can produce different outputs or actions (e.g., since we haven't locked that BlockID, we can sign a Prevote for a different value), which is an equivocation.

I'm not sure that equivocation is the problem here--the PrivVal prevents us from double signing any vote.  We haven't changed that, and we're also still synchronously writing all internal votes to the WAL so we'll always know what's been committed to locally. Also, the `EventDataRoundState` is still being written so our Handshake should at least start from the correct place. 

I think the risk present in this PR is that we're at a round state for which we can never recover the corresponding consensus state since we didn't write down the votes that got us to that round state. 

My current belief is that if each node still has its own votes then the p2p network will just end up replaying all of the votes on restart and we'll end up just fine. I couldn't get any of the tests in `replay_test.go` to fail using this change, and my additional e2e tests stop and start the whole network fine. I'm going to look into extending the WAL recovery testing to prove this change actually is safe to make.

cason:
> I'm not sure that equivocation is the problem here--the PrivVal prevents us from double signing any vote.

It is not only equivocation, it can be for instance, amnesia. 

Recalling the example I've provided, if we see a Polka for `v` in a round `r`, we lock `v` and `r` and reject anything `!= v` in rounds `r' > r`.

If we only rely on the network for receiving messages (either old or new), we might receive a proposal for round `r' > r` while in a state where we haven't seen yet the Polka for `v` in round `r`. So we are accepting (prevoting) the new value, while we should be locked on `v` and reject anything different. This is a misbehavior that can lead to a scenario where different values are committed in different rounds.

(Actually, it is even more complex, because we only lock the value while we are in the `prevote` step of round `r`, which might not be the case when recovering, as the system has moved on to larger rounds).

cason:
> Also, the EventDataRoundState is still being written so our Handshake should at least start from the correct place.

Can you point me out where this happens? From what I am aware, the only non-message event we log is `EndHeight`.

ValarDragon:
Huh? How does removing votes received from gossip, from being written to WAL make sense?

Original post gives no explanation of this.

WAL replay is precisely so you can rederive your exact consensus state on crash recovery...

ValarDragon (changes requested):
Huh? How does removing votes received from gossip, from being written to WAL make sense?

Original post and replies give no explanation of this.

WAL replay is precisely so you can rederive your exact consensus state on crash recovery...

ValarDragon:
One thing you _could_ do is buffer these in RAM for one big WAL write before prevote / precommit.

ValarDragon:
> This is true, there's no fsync for these, but we do call wal.Write which is in the critical path for receiving any message in the consensus state. And that call does buffer allocation, protobuf encoding, CRC checksums, etc., which still incurs a cost from both a time and memory perspective, so if we can agree that a change like this is safe, the data suggests that it would have meaningful benefits to Comet's performance.

The answer to me seems like its simply:
- make wal.write not expensive, none of thats fundamental. We can just change the encoding, and keep it in memory via a buffered writer until the wal.flush call

melekes:
I really appreciate the valuable discussion on the topic! Let's continue chatting in https://github.com/cometbft/cometbft/issues/3118

Thank you all for your contributions ❤️

PR #4358, opened 2024-10-29
fix(ci): [gobenchdata] fix path to binary

melekes:


andynog (approved):
lgtm, even though it is a custom domain, the address is what is suggested in https://github.com/marketplace/actions/continuous-benchmarking-for-go

PR #4359, opened 2024-10-29
ci: remove Markdown commented code from PR template

cason:
As it often ends up in the Git history, since we forget to delete it.

---

#### PR checklist

- [ ] Tests written/updated


cason:
An alternative solution to consider: https://github.com/marketplace/actions/remove-markdown-comments-in-pr

melekes (approved):

jmalicevic:
I agree its super annoying to have the comments always be part of the commit message, is tehre a way to have a separate template for commits and PRs? I think we should still somehow point people who just open PRs to some guidelines. So, if there is no template, maybe replace the big comment with  a shorter one pointing to the [contributing guidelines](https://github.com/cometbft/cometbft/blob/main/CONTRIBUTING.md#ease-of-reviewing) section on PRs. Just an opinion thought.

PR #4364, opened 2024-10-29
revert(state/indexer): copy value (#4326)

gartnera:
Revert https://github.com/cometbft/cometbft/pull/4326 as it did not need to be backported to v0.37.x as https://github.com/cometbft/cometbft/issues/4295 did not effect v0.37.x

This cometbft-db upgrade is a breaking change which caused database corruption (or at least consensus failures) on our cosmos sdk v0.47.x application when upgrading: https://github.com/zeta-chain/node/pull/3045 (revert with details: https://github.com/zeta-chain/node/pull/3065)

I'm unsure of how to write this up in the changelog.

---

#### PR checklist

- [ ] Tests written/updated


melekes:
😭 😭 😭  that change is starting to become a nightmare

melekes:
Would it be possible to update cosmos SDK? I can work on that.

melekes:
> which caused database corruption (or at least consensus failures)

can you share details?

gartnera:
> > which caused database corruption (or at least consensus failures)
> 
> can you share details?

Updated description. Our revert PR has logs: https://github.com/zeta-chain/node/pull/3065

melekes:
We don't need to revert the whole PR—just the part that updates the `cometbft-db` version. I will do it.

gartnera:
> We don't need to revert the whole PR—just the part that updates the cometbft-db version. I will do it.

Yeah that should work. But with https://github.com/cometbft/cometbft/pull/4369 there's a double allocation/copy since `it.Value()` returns a copy and then is copied again.

melekes:
> Yeah that should work. But with https://github.com/cometbft/cometbft/pull/4369 there's a double allocation/copy since it.Value() returns a copy and then is copied again.

Correct. But it won't break for users who've decided to upgrade to `cometbft-db >= v0.13` and patch Cosmos SDK.

melekes:
Sorry about this 😥 We should've handled this better.

PR #4367, opened 2024-10-30
fix(store): `BenchmarkLoadValidators` no longer panics

alesforz:
### Context
The `BenchmarkLoadValidators` benchmark in the `state` package panics because it can't find the validators' information for the specified heights. The issue is caused by a missing `DBKeyLayout` in the test database configuration, which leads to the `LoadValidators` call failing.

### This Changes
- the benchmark now creates a new state `Store` by explicitly specifying the `DBKeyLayout` in the `StoreOptions`. 
- `store` constructor checks if `DBKeyLayout` is set; if not, it sets it to a default value.

---

#### PR checklist

- [x] Tests written/updated
- ~[] Updated relevant documentation (`docs/` or `spec/`) and code comments~

### DIFF
--- a/state/store.go
+++ b/state/store.go
@@ -237,10 +237,18 @@ func NewStore(db dbm.DB, options StoreOptions) Store {
 		StoreOptions: options,
 	}
 
+	if options.DBKeyLayout == "" {
+		options.DBKeyLayout = "v1"
+	}
+
 	dbKeyLayoutVersion := setDBKeyLayout(&store, options.DBKeyLayout)
 
 	if options.Logger != nil {
-		options.Logger.Info("State store key layout version ", "version", "v"+dbKeyLayoutVersion)
+		options.Logger.Info(
+			"State store key layout version ",
+			"version",
+			"v"+dbKeyLayoutVersion,
+		)
 	}
 
 	return store
--- a/state/store_test.go
+++ b/state/store_test.go
@@ -66,12 +66,17 @@ func BenchmarkLoadValidators(b *testing.B) {
 
 	config := test.ResetTestRoot("state_")
 	defer os.RemoveAll(config.RootDir)
+
 	dbType := dbm.BackendType(config.DBBackend)
+
 	stateDB, err := dbm.NewDB("state", dbType, config.DBDir())
 	require.NoError(b, err)
+
 	stateStore := sm.NewStore(stateDB, sm.StoreOptions{
 		DiscardABCIResponses: false,
+		DBKeyLayout:          "v2",
 	})
+
 	state, err := stateStore.LoadFromDBOrGenesisFile(config.GenesisFile())
 	if err != nil {
 		b.Fatal(err)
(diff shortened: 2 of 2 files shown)


.changelog/unreleased/bug-fixes/4367-fix-bench-load-validators.md
```diff
@@ -0,0 +1,2 @@
+- BenchmarkLoadValidators no longer panics
```
melekes, line 1: don't think we need a changelog for this

melekes (approved):

.changelog/unreleased/bug-fixes/4367-fix-bench-load-validators.md
alesforz: should I delete it then?

melekes: I'd consider this changelog user-facing (for devs using Comet and ops). We don't need it because Git stores all the history for us.

melekes: > should I delete it?

yes please

jmalicevic:
Great catch Alessandro. But I think the change should actually be in the constructor of the state store. This was an oversight on my part but the constructor should check whether the DBKeyLayout is nil and set it to the default. 

Like in the block store here: https://github.com/cometbft/cometbft/blob/fad150950f1b6095b89e140e1286de027fe9778f/store/store.go#L141

alesforz:
> Great catch Alessandro. But I think the change should actually be in the constructor of the state store. This was an oversight on my part but the constructor should check whether the DBKeyLayout is nil and set it to the default.
> 

Ah, thank you! 😄 I will add this check in the constructor.
I will keep the change to explicitly set `DBKeyLayout` to `"v2"` in the test, because a [few lines below it calls](https://github.com/cometbft/cometbft/blob/3298a3578569e812f07c01e4b5a6f06d0f2509af/state/store_test.go#L97) `SaveValidatorInfo` with `"v2"`, which I suppose isn't the default value.

andynog (approved):
I can confirm that running on `main` the benchmark test fails and with this PR it works. Lgtm.

andynog:
the go benchmark was still failing even after with this, I think there might be some resource constraints when running the tests and it might be timing out, so I just pushed some small changes to the Workflow to increase the timeout value and the CPU to see if it passes.

andynog:
So, in my previous commit, I've increased the timeout and instead of failing after 15min, it took [almost 1 hour and 10 minutes](https://github.com/cometbft/cometbft/actions/runs/11599584795/job/32298049761) to finish the tests but at least they completed. But the workflow still failed because could not find the `base.json`. 

I've made some additional modifications to the workflow to download the `base.json` (assuming it will be uploaded by the `gobenchdata-base.yml` workflow that I've triggered manually). I've also added some information about output `ulimit` and change the one related to files to see if it helps the test to run faster.

melekes (approved):

.github/workflows/gobenchdata-base.yml
```diff
@@ -23,7 +23,7 @@ jobs:
 
       - run: go install go.bobheadxi.dev/gobenchdata@v1.3.1
 
-      - run: go test -bench . -benchmem ./... | gobenchdata --json base.json
+      - run: go test -timeout 90m -cpu $(nproc) -bench . -benchmem ./... | gobenchdata --json base.json
```
jmalicevic, line 26: I don't think this should be part of this PR though.

Issue #3519, opened 2024-07-15
p2p: intervals when re-dialing persistent peers are not as expected


cason:
The method used for dialing addresses configured as persistent peers is the following:

https://github.com/cometbft/cometbft/blob/b47d18ea417ad0873c2adf78b491d785d1012dc2/p2p/switch.go#L391-L438

According with the comments in the code, and some of our documentation, the total duration of re-connection attempts should be around 1 day:

https://github.com/cometbft/cometbft/blob/b47d18ea417ad0873c2adf78b491d785d1012dc2/p2p/switch.go#L23-L31

But, from this snap of the code, derived from the previous one, the total interval between the configured attempts (30) is actually around `8h14m`: https://go.dev/play/p/JWfU6lerps5.

Namely, the expected behavior does not match the implemented behavior.

PR #4377, opened 2024-10-31
fix(p2p): adjust backoff seconds to increase reconnect retries close to 24 hours

andynog:
close: #3519 

Adjust `reconnectBackOffBaseSeconds` to increase reconnect retries to up 1 day (~24 hours).

The new value can be validated here: https://go.dev/play/p/k8F5rS-i24p, which will show that the total time is increased to almost 24 hours. 

Initial reconnecting time: 2m8.493s
Total reconnecting time. : 23h55m56.249s

The `reconnectBackOffBaseSeconds` is increased by a bit over 10% (from 3.0 to 3.4 seconds) so this would not affect reconnection retries too much.

#### PR checklist


melekes (approved):
👍

cason:
Hum, I think the problem is not on the base of the backoff, but the fact that the second loop starts from `0`:

https://go.dev/play/p/mm3Hb13m5Cs

p2p/switch.go
```diff
@@ -387,7 +387,8 @@ func (sw *Switch) stopAndRemovePeer(p Peer, reason any) {
 }
 
 // reconnectToPeer tries to reconnect to the addr, first repeatedly
-// with a fixed interval, then with exponential backoff.
+// with a fixed interval (approximately 2 minutes), then with
+// exponential backoff (approximately close to 24 hours).
```
cason, line 391: The problem here for me is the second four (lines 424/426) starts from 0. Anything with power 0 == 1.

cason (changes requested):
Thank you for addressing this, but I think the solution goes in another direction (see my comments).

andynog:
Thanks @cason, I've made the suggested changes.

p2p/switch.go
```diff
@@ -423,7 +425,7 @@ func (sw *Switch) reconnectToPeer(addr *na.NetAddr) {
 
 	sw.Logger.Error("Failed to reconnect to peer. Beginning exponential backoff",
 		"addr", addr, "elapsed", time.Since(start))
-	for i := 0; i < reconnectBackOffAttempts; i++ {
+	for i := 1; i < reconnectBackOffAttempts; i++ {
```
cason, line 428:
```suggestion
	for i := 1; i <= reconnectBackOffAttempts; i++ {
```

cason: Otherwise it takes 8h.

cason (approved):
Thank for this. There is a `<` that must become `<=`.

andynog:
Thanks @cason, should we also back port this to `v0.38.x` and `v0.37.x` seems the logic is the same.

cason:
> should we also back port this to v0.38.x and v0.37.x seems the logic is the same.

I think we should.

PR #4339, opened 2024-10-25
fix(config): do not populate deprecated timeout_commit with non-zero value

wesl-ee:
Closes #4337

PR #4401, opened 2024-11-01
chore(abci): set NextBlockDelay to 1s (default)

melekes:
follow-up to https://github.com/cometbft/cometbft/pull/4339

otherwise, CometBFT spits out blocks like crazy 😸


cason:
`test/app/test.sh: line 15: abci-cli: command not found`

cason:
Can you also change that on the `e2e` application? Or is it already fixed?

melekes:
> Can you also change that on the `e2e` application? Or is it already fixed?

already fixed

andynog (approved):
lgtm, e2e would also be good

By the way, could not find any information about it on https://docs.cometbft.com/v1.0/spec/abci/abci++_app_requirements

andynog:
> > Can you also change that on the `e2e` application? Or is it already fixed?
> 
> already fixed

Yes here https://github.com/cometbft/cometbft/blob/732e2dfc23e30b6275c3f97bd7735751c6806eb9/test/e2e/app/app.go#L480

andynog:
@melekes for some reason the there are a couple of tests that are failing. I've re-run then a few times and they continue to fail so it's worth investigate.

wesl-ee (approved):
makes sense. in the absence of a defined NextBlockDelay the fallback is what's defined in the config, which defaults to 0 even when unset as this field is deprecated

melekes:
> @melekes for some reason the there are a couple of tests that are failing. I've re-run then a few times and they continue to fail so it's worth investigate.

Maybe we're using `kvstore` app in those, and `NextBlockDelay` expected to be `0`.

abci/example/kvstore/kvstore.go
```diff
@@ -115,6 +119,13 @@ func (app *Application) SetGenBlockEvents() {
 	app.genBlockEvents = true
 }
 
+// SetNextBlockDelay sets the delay for the next finalized block. Default is 0
+// here because kvstore is mostly used for testing. In production, the default
+// is 1s.
```
cason, line 124:
```suggestion
// is 1s, mimicing the default for the deprecated `timeout_commit` parameter.
```

proxy/client.go
```diff
@@ -263,20 +265,36 @@ func (r *remoteClientCreator) newABCIClient() (abcicli.Client, error) {
 // "_connsync" variant (i.e. "kvstore_connsync", etc.), which attempts to
 // replicate the same concurrency model as the remote client.
 func DefaultClientCreator(addr, transport, dbDir string) ClientCreator {
+	// Default is zero for kvstore and persistent_kvstore.
+	// Set to 1s to mimic the real world app.
```
cason, line 269:
```suggestion
	// Replaces deprecated `timeout_commit` parameter.
	// Set to 1s to mimic the real world app.
```

cason (approved):
+1

Issue #2666, opened 2024-03-21
e2e: Reimplement latency emulation Python script in Go


hvanz:
### Summary
The latency emulation script is executed on each node on which we want to emulate latency on outgoing messages. It does so by setting traffic control with the `tc` command. The current implementation in Python was adapted from an existing Python script. 

### Problem
The problems with this approach are that we need to copy the Python file to each node to be executed and that Python needs to be installed. This last point also prevents the use of this script in the `Dockerfile.fast` image, which is a very slim that only runs a go binary.

### Proposal
The solution is to reimplement the script in go as part of the e2e framework.

PR #4408, opened 2024-11-01
refactor(e2e): Replace latency emulation Python script by generating a shell script in e2e's setup

hvanz:
Solves #2666 

Currently latency emulation is set up on each e2e node by running a Python script that requires two csv files, one with information about all the nodes' IP addresses and another with latency values between different data-center zones.

This PR generates, for each node in the testnet, a single shell script with the same `tc` (traffic control) commands as the Python scripts. The shell script is then called on the entry point of the Docker image.

To reviewers: follow the commits.

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

test/e2e/runner/setup.go
```diff
@@ -500,3 +516,78 @@ func UpdateConfigStateSync(node *e2e.Node, height int64, hash []byte) error {
 	bz = regexp.MustCompile(`(?m)^trust_hash =.*`).ReplaceAll(bz, []byte(fmt.Sprintf(`trust_hash = "%X"`, hash)))
 	return os.WriteFile(cfgPath, bz, 0o644) //nolint:gosec
 }
+
+// tcCommands generates a list of tc (traffic control) commands to emulate
```
cason, line 520: So this guy creates the `emulate-latency.sh` script?

test/e2e/docker/entrypoint
cason, line 1: Shouldn't we change this on other entrypoints? E.g., the `builtin` one?

test/e2e/runner/setup.go
cason: Do you mind moving this function to a different file, like `latency.go` or `latency_emulation.go`?

cason (approved):
Thank for that, very helpful removing `python` requirement and script from the image :)

test/e2e/runner/setup.go
hvanz: `tcCommands` generates the content of `emulate-latency.sh`. I moved this function to to a new file `latency_emulation.go`.

PR #4420, opened 2024-11-04
docs(metrics): Fix typo

gibson042:
See https://docs.cometbft.com/v0.38/core/metrics :
> Listen address can be changed in the config file (see `instrumentation.prometheus\_listen\_addr`).


melekes (approved):
Thanks @gibson042 ❤️

PR #4427, opened 2024-11-04
fix: modify changelog entry to reflect the right message for PR #4377

andynog:
just updating a changelog entry for PR #4377 to reflect the real change, it was not the backoff seconds that was adjusted in the end so better message needed in changelog entry.

Fixing only in `main` since it will be fixed on the back ports individually.


melekes (approved):

cason (approved):
+1

PR #4429, opened 2024-11-04
test(e2e): always output manifest

andynog:
This PR changes the logic in `run-multiple` to always output the manifest content, not only when it fails. 

This will be helpful to troubleshoot errors where the infrastructure might be the root cause of the error (test is killed by CI because it was using too many resources). 

Always showing the manifest content in the log will allow local testing to try to reproduce the problem.


melekes (approved):
👍

cason (approved):
+1

andynog:
Thanks @cason and @melekes

Issue #3007, opened 2024-05-06
Change gossip routines sleep setup


ValarDragon:
### Summary

Currently every gossip routine attempts to gossip a vote or block part. If it has no more block parts to gossip, it sleeps (by default 100ms). This has three issues:

- We get an unnecessary delay for first part gossip. When we are ready to gossip new block parts, we will be unnecessarily sleeping for another rand[0,100]ms.
- If we only have one block part, we gossip it to a peer, then we have nothing else to gossip, we will sleep for 100ms. If another block part comes in, we have to wait that [0,100]ms.
- Wasted CPU when we are not in the relevant gossip phase

### Proposal

Make each gossip thread use a channel that "wakes it up". 

For block part gossip:
- When we receive a new block part, go to the next block, or detect our peer is on a new block, restart our loop. 

For vote gossip:
- When we receive a new vote, go to the next block, or detect our peer is on a new block, restart our loop. 

Otherwise keep the current structure for a loop that is "awake".

cason:
In summary, we need a way to signal to the gossip data routines that new block parts are, or might be, available to be sent. This is a limitation of the current design, where consensus logic and reactor "communicate" via shared memory.

A possibility that we have discussed is to leverage the existing synchronous signal `cs.evsw.FireEvent(types.EventProposalBlockPart, msg)` between the consensus logic and reactor.

cason:
For votes, there is also a synchronous event produced by the consensus logic: `cs.evsw.FireEvent(types.EventVote, vote)`.

cason:
>  For votes, I don't understand the purpose of those sleeps 😢

We only sleep when we don't find a vote to send to the peer. We sleep because we retrieve votes from the consensus logic via shared memory. The option we have been discussing is to explicitly signalize the reactor when there are new stuff (e.g., votes) available in the shared state to be sent.

While we find votes suitable to be sent, we skip the sleeps in the loops.

cason:
This can be in conjunction with #3211.

PR #4436, opened 2024-11-05
perf(consensus)!: eliminate busy loop allocation of heap memory for round state 

wesl-ee:
This goroutine runs continuously with a minimal, negligible timer (so the Go scheduler will not starve). For practical purposes the timer does not exist because it is so short (would fire 10,000x per second).

https://github.com/cometbft/cometbft/blob/48db7ac0037db4e16eab7536a05966f928a4b70b/internal/consensus/reactor.go#L560-L561

This routine acquires locks to both (1) the consensus state struct and subsequently (2) the round state struct. It allocates new memory for a copy of consensus state anytime it can acquire the lock, which is often, as the lock is only held when consensus state is being mutated or read. The size of this allocation is approximately 216 bytes on a 64-bit arch. The struct is about half pointers and half fixed-size fields.

https://github.com/cometbft/cometbft/blob/48db7ac0037db4e16eab7536a05966f928a4b70b/internal/consensus/state.go#L259-L264

Proposal
---------

This fix removes the hot loop update of the reactor's view of round state by copying the round state from the event bus. In the case this is not passed, consensus state is accessed the same as it was before in `updateRoundStateNoCsLock()`, i.e. without acquiring the lock.

Another change worth noting is passing and storing the struct itself instead of a pointer to the struct to more clearly indicate this is an owned clone of the round state.

These two changes should reduce garbage collector thrash.

---

#### PR checklist

- [ ] Tests written/updated


melekes (commented):
awesome ❤️ would be great to see the difference between heap profiles. @wesl-ee what difference did you observe?

cason:
I think this follows the lines of https://github.com/cometbft/cometbft/issues/3007.

internal/consensus/state.go
```diff
@@ -558,7 +557,7 @@ func (cs *State) updateRoundStep(round int32, step cstypes.RoundStepType) {
 }
 
 // enterNewRound(height, 0) at cs.StartTime.
-func (cs *State) scheduleRound0(rs *cstypes.RoundState) {
+func (cs *State) scheduleRound0(rs cstypes.RoundState) {
```
cason, line 560: So now this is a copy of it?

internal/consensus/state.go
```diff
@@ -249,7 +249,7 @@ func (cs *State) GetLastHeight() int64 {
 
 // GetRoundState returns a shallow copy of the internal consensus state.
 // This function is thread-safe.
-func (cs *State) GetRoundState() *cstypes.RoundState {
+func (cs *State) GetRoundState() cstypes.RoundState {
```
cason, line 252: This is a breaking change. But does it need to be one? We can just change the private method instead.

internal/consensus/reactor.go
```diff
@@ -61,10 +62,11 @@ func NewReactor(consensusState *State, waitSync bool, options ...ReactorOption)
 	conR := &Reactor{
 		conS:          consensusState,
 		waitSync:      atomic.Bool{},
-		rs:            consensusState.GetRoundState(),
-		initialHeight: consensusState.state.InitialHeight,
+		rs:            consensusState.RoundState,
```
cason, line 65: Not sure if I understood the change here.

internal/consensus/reactor.go
```diff
@@ -421,31 +416,58 @@ func (conR *Reactor) subscribeToBroadcastEvents() {
...
 		}); err != nil {
 		conR.Logger.Error("Error adding listener for events (Vote)", "err", err)
 	}
 
 	if err := conR.conS.evsw.AddListenerForEvent(subscriber, types.EventProposalBlockPart,
 		func(data cmtevents.EventData) {
 			conR.broadcastHasProposalBlockPartMessage(data.(*BlockPartMessage))
-			conR.updateRoundStateNoCsLock()
+
+			// update reactor's view of round state
```
cason, line 464: This is a good question, not sure if I trust the concurrency model of the synchronous bus.

internal/consensus/reactor.go
```diff
@@ -47,9 +47,10 @@ type Reactor struct {
 	waitSync atomic.Bool
 	eventBus *types.EventBus
 
-	rsMtx         cmtsync.RWMutex
-	rs            *cstypes.RoundState
-	initialHeight int64 // under rsMtx
+	initialHeight atomic.Int64
+
+	rsMtx cmtsync.RWMutex
+	rs    cstypes.RoundState
```
cason, line 53: Not sure about the benefits of this change.

cason (commented):
Thank you for that.

This is something that we need to find a way to address, as described in https://github.com/cometbft/cometbft/issues/3007.

But we need some performance evaluation, or at least tracing to attest the benefits of this approach.

cason:
I'd like to have the opinion of @ValarDragon regarding this change, as the author of the referred issue.

internal/consensus/state.go
melekes: Nothing wrong with breaking change here. We are not going to backport this, don't we?

cason: We are probably not, but I am not really convinced about the performance improvement that it can provide.

So, if we are to break, lets do that while providing a measurable benefit.

internal/consensus/state.go
alesforz: Not a deep copy. Some of `RoundState` fields are pointers. Passing a `RoundState` object as value will copy _the pointers_, not whatever those pointers point to. For example:
```go
// just an example, suppose this is properly inizialized.
rs := &cstypes.RoundState{} 

cs.scheduleRound0(rs)
```
The pointer fields of the original `rs` and the `RoundState` inside the `scheduleRound0` function will point to the same data in memory.

wesl-ee: ^

internal/consensus/state.go
wesl-ee: I can collect profiling on this change today and follow up w a comment. Removal of the hot loop `updateRoundStateRoutine()` was enough reason for me to tackle this change without a profile, as removing a Go routine that always copies is almost definitely better, but I can provide numbers to back up.

internal/consensus/reactor.go
wesl-ee: 🙃 swapped order of lines in the course of doing this change and made it into the diff. I removed this superfluous change

internal/consensus/reactor.go
wesl-ee: Reverted this change, I removed the getter at one point but brought it back and this diff still persisted

internal/consensus/state.go
cason: It is a shallow copy, as commented in line 259. 

But for internal transitions like that, I am not sure whether this makes much sense.

Moreover, this method is so minor - it schedules a timeout - that this shallow copy is pretty much unnecessary.

internal/consensus/reactor.go
```diff
@@ -48,8 +48,8 @@ type Reactor struct {
 	eventBus *types.EventBus
 
 	rsMtx         cmtsync.RWMutex
-	rs            *cstypes.RoundState
-	initialHeight int64 // under rsMtx
+	rs            cstypes.RoundState
```
cason, line 51:
```suggestion
	rs            cstypes.RoundState // copy of consensus state
```

internal/consensus/reactor.go
```diff
@@ -48,8 +48,8 @@ type Reactor struct {
...
+	initialHeight atomic.Int64
```
cason, line 52: Not regarding this PR, but the reason for which we need to keep this is so.... annoying and minor. I really would find a different solution for that.

internal/consensus/reactor.go
```diff
@@ -286,9 +284,8 @@ func (conR *Reactor) Receive(e p2p.Envelope) {
 		case *HasProposalBlockPartMessage:
 			ps.ApplyHasProposalBlockPartMessage(msg)
 		case *VoteSetMaj23Message:
-			conR.rsMtx.RLock()
-			height, votes := conR.rs.Height, conR.rs.Votes
-			conR.rsMtx.RUnlock()
+			rs := conR.getRoundState()
```
cason, line 287: We are operating in a slightly possibly outdated version of the round state here.

At the same time, this message is not very common (it is protected by a 2s sleep, see `queryMaj23Routine` method).

We probably here could get a copy of the consensus state instead of relying on the local copy we keep in the reactor.

internal/consensus/reactor.go
```diff
@@ -373,9 +368,8 @@ func (conR *Reactor) Receive(e p2p.Envelope) {
 		}
 		switch msg := msg.(type) {
 		case *VoteSetBitsMessage:
-			conR.rsMtx.RLock()
-			height, votes := conR.rs.Height, conR.rs.Votes
-			conR.rsMtx.RUnlock()
+			rs := conR.getRoundState()
```
cason, line 371: The same rationale of the line 287 comment applies here.

We probably should get a copy of the actual consensus state.

internal/consensus/state.go
```diff
@@ -1947,7 +1946,7 @@ func (cs *State) finalizeCommit(height int64) {
 
 	// cs.StartTime is already set.
 	// Schedule Round0 to start soon.
-	cs.scheduleRound0(&cs.RoundState)
+	cs.scheduleRound0(cs.RoundState)
```
cason, line 1949: No need to create a shallow copy here.

internal/consensus/reactor.go
```diff
@@ -421,31 +415,58 @@ func (conR *Reactor) subscribeToBroadcastEvents() {
 	const subscriber = "consensus-reactor"
 	if err := conR.conS.evsw.AddListenerForEvent(subscriber, types.EventNewRoundStep,
 		func(data cmtevents.EventData) {
-			conR.broadcastNewRoundStepMessage(data.(*cstypes.RoundState))
-			conR.updateRoundStateNoCsLock()
+			rs := data.(cstypes.RoundState)
+
+			// update reactor's view of round state
```
cason, line 420: A suggestion here is to create a method to update the local copy of the round state.

The changes on these four methods are the same:

1. Lock the local copy
2. Update it (the source is different in the former two than in the latter two)
3. Unlock the local copy

I am not sure if changing the order here has an impact. Notice that in the original code we broadcast the message, then update the local copy. Here the order is inverted.

cason (approved):
Thank you for that, I think the code becomes easier to read and I've never liked that almost busy loop to update the consensus state.

The right direction to go is to transfer the state from the consensus implementation to the reactor via "message passing", as we are already doing using the synchronous bus (although, a that synchronous bus is not an ideal setup as well).

What we really miss, however, is an assessment of the impact of these changes in performance, even if restricted to heap allocations and garbage collection costs.

I will approve to unblock, but I recommend considering my comments.

wesl-ee:
Rebased and profiled this PR directly against main using the command `go tool pprof -diff_base ./heap-theirs -http=: ./heap-mine` for the heap profile.  Over a 10s run on an e2e network I can see that this PR did the following:

- Reduced the magnitude of heap space allocated by `GetRoundState` by 3.14MiB (% change -4.23%)
- Reduced the number of objects allocated by `GetRoundState` by 15,825 (% change 3.90%)

Screenshot 2024-11-13 at 3 57 38 PM


Over a 10m run the difference was more clear:

- Reduced the magnitude of heap space allocated by `GetRoundState` by 133.39MiB (% change 23.47%)
- Reduced the number of objects allocated by `GetRoundState` by 672,438 (% change 12.79%)

Screenshot 2024-11-13 at 3 56 10 PM

internal/consensus/reactor.go
cason: I am pretty sure that the profiling results would not change if we use here the latest version of the consensus state, instead of the (possibly stale) local copy.

internal/consensus/state.go
wesl-ee: Done, no more shallow copy here

internal/consensus/reactor.go
wesl-ee: Done

internal/consensus/state.go
wesl-ee: Done

internal/consensus/reactor.go
```diff
@@ -286,9 +284,11 @@ func (conR *Reactor) Receive(e p2p.Envelope) {
 		case *HasProposalBlockPartMessage:
 			ps.ApplyHasProposalBlockPartMessage(msg)
 		case *VoteSetMaj23Message:
-			conR.rsMtx.RLock()
-			height, votes := conR.rs.Height, conR.rs.Votes
-			conR.rsMtx.RUnlock()
+			// Get the updated round state as our view may be stale
+			conR.conS.mtx.Lock()
```
cason, line 288:
```suggestion
```

internal/consensus/reactor.go
```diff
@@ -286,9 +284,11 @@ func (conR *Reactor) Receive(e p2p.Envelope) {
...
+			rs := conR.conS.getRoundState()
```
cason, line 289:
```suggestion
			rs := conR.conS.GetRoundState()
```

internal/consensus/reactor.go
```diff
@@ -286,9 +284,11 @@ func (conR *Reactor) Receive(e p2p.Envelope) {
...
+			conR.conS.mtx.Unlock()
```
cason, line 290:
```suggestion
```

internal/consensus/reactor.go
```diff
@@ -373,9 +371,12 @@ func (conR *Reactor) Receive(e p2p.Envelope) {
 		}
 		switch msg := msg.(type) {
 		case *VoteSetBitsMessage:
-			conR.rsMtx.RLock()
-			height, votes := conR.rs.Height, conR.rs.Votes
-			conR.rsMtx.RUnlock()
+			// Get the updated round state as our view may be stale
+			conR.conS.mtx.Lock()
```
cason, line 375:
```suggestion
```

internal/consensus/reactor.go
```diff
@@ -373,9 +371,12 @@ func (conR *Reactor) Receive(e p2p.Envelope) {
...
+			rs := conR.conS.getRoundState()
```
cason, line 376:
```suggestion
			rs := conR.conS.GetRoundState()
```

internal/consensus/reactor.go
```diff
@@ -373,9 +371,12 @@ func (conR *Reactor) Receive(e p2p.Envelope) {
...
+			conR.conS.mtx.Unlock()
```
cason, line 377:
```suggestion
```

cason (commented):
Thank for the updates.

Notice that we can call `rs := conR.conS.GetRoundState()` instead of taking the lock, using the private method, then releasing the lock. Which is exactly what the public method does.

internal/consensus/reactor.go
cason: https://github.com/cometbft/cometbft/issues/4491

wesl-ee:
this seems transient and unrelated. I see this test pass locally but cannot rerun CI myself

cason:
Thank you very much for this @wesl-ee. Sorry for being picky, this part of the code is super sensitive.

melekes (commented):
🚀

PR #4443, opened 2024-11-06
ci: remove backporting to v0.38.x-experimental

melekes:
in favor of manual merge once in a while. Manually solving conflicts upon each merge takes too much time and not worth it.


cason (approved):
Legit.

PR #4454, opened 2024-11-07
chore: updates in preparation for v1.0.0-rc2

andynog:
Updates  in preparation for `v1.0.0-rc2`:

- `summary.md`
- Tutorial instructions

#### PR checklist

- [ ] Tests written/updated


.changelog/unreleased/summary.md
```diff
@@ -38,11 +38,16 @@ team, including:
...
 
 None of these changes are state machine-breaking for CometBFT-based networks,
 but could be breaking for some users who depend on the Protobuf definitions type
-URLs. See the [upgrading guidelines](./UPGRADING.md) and specific changes below
-for more details.
+URLs.
+
+See the [upgrading guidelines](./UPGRADING.md) and the specific changes below for more details. In this release,
+we are also introducing a migration guide, please refer to the
+[Upgrading from CometBFT v0.38.x to v1.0](./docs/guides/upgrades/v0.38-to-v1.0.md) document
```
melekes, line 50:
```suggestion
[Upgrading from CometBFT v0.38.x to v1.0](./docs/guides/upgrades/v0.38-to-v1.0.md) document.
```

.changelog/unreleased/summary.md
```diff
@@ -1,4 +1,4 @@
-*July 1, 2024*
+* November 12, 2024*
```
cason, line 1:
```suggestion
*November 12, 2024*
```

cason: Otherwise this becomes a bullet list

.changelog/unreleased/summary.md
cason, line 1: Links from lines 19 and 22 are broken.

cason: Also, can we discuss the order of the numbered list?

I think PBTS is the major feature of this version. But it is down below the list.

.changelog/unreleased/summary.md
```diff
@@ -38,11 +38,16 @@ team, including:
     algorithm used by CometBFT for computing block times.
     When activated on a chain, it replaces the pre-existing BFT-time algorithm.
     See [spec](./spec/consensus/proposer-based-timestamp) doc for PBTS.
+8. Added support for [BLS12-381 curve](https://github.com/cometbft/cometbft/pull/2765) that can
```
cason, line 41: This sounds to me an experimental support, since we need a build flag.

Not sure if this is part of the major features.

cason: And if it is, we should add a link to what this "build tag" means.

Again, the release binaries would not contain it, so.... not a major feature.

CHANGELOG.md
cason, line 1: The changelog entry for ([#2786](https://github.com/cometbft/cometbft/pull/2786)) is pretty bad. Are we removing the support for these DBs? It is not a config change, IMO.

cason: The changelog entry ([#2397](https://github.com/cometbft/cometbft/pull/2397)) refers to `internal/state`. We don't have this package (it is just `state`).

cason: [kvstore] Function NewApplication now has an extra lanes map[string]uint32 parameter ([#3980](https://github.com/cometbft/cometbft/pull/3980))

We don't have lanes in v1.x. Why that hell this entry is here?

cason: Ah, ok, these are all unreleased changes... the problem here is that they appear to be part of the RC.

cason: There is a change in the `PrepareProposal` ABCI method signature/proto.

I couldn't find an entry referring to it.

cason (commented):
Thanks for this.

We have to pay attention to what goes to the CHANGELOG of this new release. It is kind of confusing because we are considering all **unreleased** changes, which include stuff that is only in `main`.

Nonetheless, some entries are poorly written and the description of some important changes is missing.

andynog:
> We have to pay attention to what goes to the CHANGELOG of this new release. It is kind of confusing because we are considering all **unreleased** changes, which include stuff that is only in `main`.

@cason, yes, the `CHANGELOG.md` was added here but just because we tend to update it when we do a release. But since we are back porting this PR to `v1.x` we should not have the CHANGELOG.md in this PR. I'll exclude it. It is generated with `unclog` so we should not need to modify the file directly. 

Then I will open another PR with the generated `CHANGELOG.md` in `v1.x` when we back port this and we agree on the content of `summary` and the `.changelog` entries.

Thanks !

melekes (approved):
👍

PR #4473, opened 2024-11-12
fix(e2e): do not oscilate validator set when it contains a single validator

cason:
This issue was introduced by https://github.com/cometbft/cometbft/pull/2283.

I've sneaked in a minor change in the script, so that we can grep for "Testnet failed" and go directly to the right manifests.

---

#### PR checklist

- [ ] Tests written/updated


alesforz (approved):

andynog (approved):
lgtm, that's exactly what I had in[ my branch](https://github.com/cometbft/cometbft/commit/535791cbcbbf261ea8563e124a481132b6da6215) but I didn't push a PR since I was waiting on validation. This fixes the problem.

Issue #4318, opened 2024-10-22
mempool/dog: Write spec


hvanz:
Write a specification in English of the DOG protocol, based on the description in #3263.

The documention should be located in `/spec/mempool/dog-gossip.md` (or similar).

Since the new protocol is an extension of the current one (called Flood), we can split the spec in two:
- [x] #4476
- [ ] #4477

PR #4476, opened 2024-11-12
spec(mempool/dog): Add spec of Flood gossip protocol

hvanz:
Contributes to #4318

The DOG spec is based on Flood's spec (#4477)

[README.md](https://github.com/cometbft/cometbft/tree/hvanz/flood-spec-4318/spec/mempool/gossip/)


spec/mempool/gossip/p2p.qnt
```diff
@@ -0,0 +1,141 @@
...
+    def Peers(node) = peers.get(node)
+
+    //--------------------------------------------------------------------------
+    // A node is in the network if it has peers.
+    val nodesInNetwork = NodeIDs.filter(node => node.Peers().nonEmpty())
+    val nodesNotInNetwork = NodeIDs.exclude(nodesInNetwork)
+
+    // A node sends a message to the given set of target nodes. It updates the
+    // list of incoming messages `_msgs`.
+    pure def send(node, _msgs, targetNodes, msg) =
```
cason, line 57: Should we call it broadcast, multicast, multi-send or something like that?

I read `send` and I think about unicast.

spec/mempool/gossip/p2p.qnt
```diff
@@ -0,0 +1,141 @@
...
+    // node has.
+    const InitialPeers: NodeID -> Set[NodeID]
+
+    //--------------------------------------------------------------------------
+    // State
+    //--------------------------------------------------------------------------
+    // To model network communication, each node has a queue (a list) of
+    // incoming messages. Node A sends a message to a node B by appending the
+    // message to B's queue. We use queues to model that messages arrive in
+    // order. Messages have a sender (a node ID).
```
cason, line 31:
```suggestion
    // FIFO order. Messages have a sender (a node ID).
```

cason: Do we really need this assumption?

spec/mempool/gossip/p2p.qnt
```diff
@@ -0,0 +1,141 @@
...
+    //
+    // In the actual implementation, transaction messages are transamitted on
+    // the `Mempool` data channel of the P2P layer. Control messages are usually
+    // transmitted on other channels with different priorities. Here we model a
+    // single, reliable channel.
+    //
+    // The type variable `msg` can be instantiated on the message types of
+    // different protocols.
+    var msgs: NodeID -> List[(NodeID, msg)]
```
cason, line 40: You probably want to make names small, but how about`incoming_msgs`?

spec/mempool/gossip/p2p.qnt
```diff
@@ -0,0 +1,141 @@
...
+    // A node receives one of the incoming messages from a peer and handles it
+    // according to its type.
+    action receiveFromPeer(node, handleMessage) = all {
+        require(length(node.Msgs()) > 0),
+        val someMsg = node.Msgs().head()
+        val sender = someMsg._1
+        val msg = someMsg._2
+        // We model receiving of a message as taking the head of the list of
+        // incoming messages.
+        val updatedMsgs = msgs.update(node, tail)
```
cason, line 75: Is `tail` a Quint function?

spec/mempool/gossip/flood.qnt
```diff
@@ -0,0 +1,143 @@
...
+    type Message =
+        | TxMsg(TX)
+
+    //--------------------------------------------------------------------------
+    // Actions
+    //--------------------------------------------------------------------------
+
+    // Upon receiving a transaction message from a peer, the node attempts to
+    // add the transaction to its mempool.
+    action handleMessage(node, _msgs, sender, msg) =
```
cason, line 27: So `_msgs` here is of the same time of `var msgs` in `p2p.qnt`?

spec/mempool/gossip/flood.qnt
```diff
@@ -0,0 +1,143 @@
...
+        node.Peers().exclude(node.sendersOf(tx))
+
+    //--------------------------------------------------------------------------
+    // All possible state transitions in the protocol.
+    action nextState = any {
+        // Transaction dissemination: node sends transaction to subset of peers.
+        nondet node = oneOf(nodesInNetwork)
+        node.disseminateNextTx(mkTargetNodes, TxMsg),
+
+        // User-initiated transactions: node receives a transaction from a user.
```
cason, line 93:
```suggestion
        // User-submitted transactions: node receives a transaction from a user.
```

cason (commented):
Great work.

Sorry my Quint-wise ignorance on some comments.

I wonder, or suggest:

1. How about the contents of `spec/mempool/quint`? Can we re-use, update, do something with it?
2. I think Quint specs are great, but several readers would also like to have an English, even if simplistic and pointing to Quint files, version of it.

spec/mempool/gossip/p2p.qnt
hvanz: Yes, we assume that the transport layer guarantees a message ordering. That's why the type of the incoming messages is a List instead of a Set. I'll clarify in the text.

spec/mempool/gossip/p2p.qnt
hvanz: Yes, it's one of the predefined functions of lists.

spec/mempool/gossip/mempool.qnt
```diff
@@ -0,0 +1,204 @@
...
+
+    // Transactions are validated by an external entity. The validation function
+    // must be deterministic. In the actual implementation, the mempool makes a
+    // CheckTx ABCI call to the application, which validates the transaction. 
+    //
+    // In this simplified specification we model all transactions as valid. To
+    // model invalid transactions, `valid` should be declared as a model
+    // parameter (a `const`) and instantiated with a deterministic function of
+    // type `(TX) => bool`.
+    pure def valid(tx) = true
```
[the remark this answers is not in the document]
hvanz: That's correct. When defining a model of this spec, we would set `const Txs` to be, for example, "Set("tx1", "tx2", "tx3", "invalid"). But it's not needed on this spec.

spec/mempool/gossip/flood.qnt
hvanz: `_msgs` has the same type as `msgs` in the `p2p` module. This parameter is to pass updated versions of `msgs`, so first `receiveFromPeer` removes the first message in the queue and pass the tail to `handleMessage`, later in dog.qnt, `_msgs` may updated with a `HaveTx` message. I clarified in the text.

hvanz:
> 1. How about the contents of spec/mempool/quint? Can we re-use, update, do something with it?

That spec is at a different abstraction level. For instance, the mempool is modelled simply as a set of transactions. There are no details on gossip, cache, senders, etc. It's more about the relation of mempool with consensus, which is modelled as two sets of proposed and decided transactions. So one of the main goals of this spec is to declare properties on at that level of abstraction; it's not really useful for specifying the gossip protocols.

By the way, I now see that those Quint files are broken, the parser complains with `'oneOf' must be the outermost expression in a nondet definition`.

hvanz:
> 2. I think Quint specs are great, but several readers would also like to have an English, even if simplistic and pointing to Quint files, version of it.

I started writing the spec in English. It was not my intention to write Quint at this point. As I was writing pseudo-code, I needed to add definitions, clarifications, and more text on the meaning of the pseudo-code. I realized I was writing something very similar to Quint, so I decided to just switch the language of the spec. So I put all text in Quint comments and wrote the code as simply as possible so that it can be read as pseudo-code, abstracting all complex parts to lower level modules. My goal is that the top modules self-describe the protocols and can be read without knowing Quint or the low level modules, unless there's a need to understand the semantics of some part of the code.

I think it's not a good idea to have both the formal spec and the English spec that say more or less the same but in different files and different languages. It duplicates the work. I know that reading markdown is more pleasing than comments, so ideally, and what I was going to suggest to the Quint team, is to have the VSCode Quint plugin with support for a viewing mode that displays comments as markdown, the same as when one previews markdown files, which is natively supported by VSCode. cc: @bugarela

If forgot that the Quint docs suggests to use [literate specifications](https://quint-lang.org/docs/literate). I haven't tried it but I guess the writing process would be different because I would have to transpile markdown to Quint every time I want to change the code. And also I could not rely so easily on type checking.

cason:
> By the way, I now see that those Quint files are broken, the parser complains with 'oneOf' must be the outermost expression in a nondet definition.

So can you add some note on that directory and/or in the common ancestor of this one, rendering clear that it is outdated or experimental?

cason:
When I mentioned the English spec is not to provide the same abstractions or the same logic, which is typically clearer in Quint than in English. It is more to define the entry points of the Quint spec, the types or methods to put more attention, besides how to run then, what to verify and how.

spec/mempool/gossip/mempool.qnt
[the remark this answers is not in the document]
cason: This is something that could go in the English spec, for instance.

cason (approved):

spec/mempool/gossip/p2p.qnt
hvanz: I renamed it `incomingMsgs` (which was the original name 😄 )

bugarela:
Regarding Quint <-> markdown, we have currently two options and I'm happy to work on improving them or setting up something new that would be useful for this kind of thing.
1. Literate specifications, like a recent one here: https://github.com/left-curve/left-curve/blob/c440d02c2b7e15c92ca9c4fed9a589bf1f8dc3fc/grug/jellyfish-merkle/spec/invariants.md (check the raw file and the rendered one. A complete checkable Quint spec is produced out of this file, so it is a live document that we can check on the CI for example)
- This looks great and enables all the freedom of markdown
- It's a bit annoying to make changes to the Quint model with this. What I do is to change the acutal `.qnt` file and, once I'm done, reflect the changes on the markdown and make sure the tangled (generated from md) file is the same the `.qnt` one
- It might be possible to write tooling that typechecks quint snippets inside markdown, and then we would be able to edit markdown directly. This is what noteeboks (likye jupyter notebook) do.

2. Generating markdown out of Quint comments. This is currently possible with the `quint docs` cli command
- The current format is tuned for generating builtin operator documentation, and it's probably not a good general format. Idk if we could come up with a general format that would be useful for multiple different projects, as each project might want to organize the markdown file in a specific way. 
- This requires that all content from markdown is embedded in Quint comments, which I think would either degrade the markdown quality or the quint file quality.
- This doesn't have the problem of having to update something when you update the model, as the markdown file can be easily re-generated

hvanz:
Thanks @bugarela! I like your literate spec example and I will give it a try. I couldn't find the qnt files it imports (hashes, tree, etc.). Where are those? And, do you keep the generated quint file somewhere in the repo?

I tried `quint docs` and it outputs only the declarations with signatures but it ignores all the comments. Does it expect a special format? I couldn't find documentation in the website.

bugarela:
> Thanks @bugarela! I like your literate spec example and I will give it a try. I couldn't find the qnt files it imports (hashes, tree, etc.). Where are those? And, do you keep the generated quint file somewhere in the repo?

This is all under a private repo for now, we only sent that `md` file to the public repository at this point.

> I tried quint docs and it outputs only the declarations with signatures but it ignores all the comments. Does it expect a special format? I couldn't find documentation in the website.

Yes, it only considers docstrings, which in Quint are triple-slashed comments before top level definitions/modules (copied from rust)
```bluespec
/// foo is bla bla
val foo = 1

/// my def is cool
pure def f(x) = {
  // This comment cannot be triple slash as it does not come before a top-level definition
  x + 1
}
```

hvanz:
I changed the Quint files for Markdown files with Quint code snippets. I think this format is much better.

@bugarela check out how I structured all the Quint code at the end of the md files using macros. In this way I don't have to insert line breaks in html comments and the md file is much cleaner.

bugarela:
I considered using that! I like how it solves indentation, but I don't like not having a line between each definition. If I were to use this format, I'd still keep my line break comments :sweat_smile:

cason:
>  but I don't like not having a line between each definition.

This renders the code super hard to read indeed.

hvanz:
Obviously line breaks between definitions is better, and even some comments would help when reading the code, but the quint code in this spec is not to read, now that we have the markdown files. I hope by the time we need to expand the specs with tests, models, etc. (not planned for the moment) there will be a `quint fmt` tool ready 😉 Otherwise we'll need to resort to put line breaks in comments.

cason:
> . I hope by the time we need to expand the specs with tests, models, etc. (not planned for the moment) there will be

Ok, I would say we merge this PR as it is, but can you add this follow-up work to the existing issue or a new one?

melekes (commented):
❤️

PR #4477, opened 2024-11-12
spec(mempool/dog): Add spec of DOG gossip protocol

hvanz:
Solves #4318

Based on #4476 

[README.md](https://github.com/cometbft/cometbft/tree/hvanz/dog-spec-4318/spec/mempool/gossip)


melekes (approved):
🏆

spec/mempool/gossip/dog.md
```diff
@@ -0,0 +1,606 @@
...
+* `adjustInterval`: the time (in milliseconds) that the controller waits between successive
+  calls of `adjustRedundancy`.
+    ```bluespec "params" +=
+    const adjustInterval: int
+    ```
+	This interval should allow sufficient time for control messages (`HaveTx` and `Reset`) to
+	propagate through the network and take effect.
+	
+    A minimum value for `adjustInterval` depends on the network’s round-trip time (RTT). Assuming that
+    global latency typically stays below 500ms (maybe an excessive number, need references here), the
```
cason, line 293: 500ms is not exaggerated. We should not think about network-level latencies, which are rarely higher than that, but on end-to-end latency. This includes all the processing, encoding/decoding, multiplexing, and protocol overhead. Plus the fact that a node does not only process transactions, it receives a bunch of other messages.

Being conservative here is helpful.

spec/mempool/gossip/dog.md
```diff
@@ -0,0 +1,606 @@
...
+action tryAddTx(node, _incomingMsgs, optionalSender, tx) = 
+    if (not(hash(tx).in(node.Cache())))
+        node.tryAddFirstTimeTx(_incomingMsgs, optionalSender, tx)
+    else
+        node.processDuplicateTx(_incomingMsgs, optionalSender, tx)
+```
+
+* **Adding first-time transactions**
+
+    `tryAddFirstTimeTx` attempts to add a received first-time transaction `tx` to the mempool by
```
cason, line 422:
```suggestion
    `tryAddFirstTimeTx` attempts to add a first-time-received transaction `tx` to the mempool by
```

spec/mempool/gossip/dog.md
```diff
@@ -0,0 +1,606 @@
...
+    where `replyHaveTx` will send a `HaveTx` message if `tx` comes from a peer and `HaveTx` messages are
+    not blocked:
+    ```bluespec "actions" +=
+    pure def replyHaveTx(node, _incomingMsgs, _rc, optionalSender, tx) =
+        if (optionalSender.isSome() and not(_rc.isHaveTxBlocked))
+            val targets = optionalSender.optionToSet()
+            (node.multiSend(_incomingMsgs, targets, HaveTxMsg(hash(tx))), _rc.blockHaveTx())
+        else (_incomingMsgs, _rc)
+    ```
+    Immediately after sending the `HaveTx` message the controller will block sending a new one.
```
cason, line 463: This is an information that is only introduced here. I realized that before by investigating the Quint code.

This is not a detail and should be clear from the beginning: at each cycle of the redundancy control mechanism, the node can send at most one `HaveTx` message.

spec/mempool/gossip/dog.md
```diff
@@ -0,0 +1,606 @@
...
+        peers' = peers,
+        mempool' = mempool,
+        senders' = senders,
+        rc' = rc,
+    }
+    ```
+    The list of `tx`’s senders contains the node IDs from which `node` received the transaction,
+    ordered by the arrival time of the corresponding messages. To avoid disabling the routes from
+    all those senders at once, the protocol picks the first sender in the list, which is the first
+    peer from which `node` received `tx`. Subsequent entries in the list are nodes whose transaction
```
cason, line 496:
```suggestion
    peer from which `node` received `tx` for the first time. Subsequent entries in the list are nodes whose transaction
```

spec/mempool/gossip/dog.md
```diff
@@ -0,0 +1,606 @@
...
+    action handleResetMessage(node, _incomingMsgs, sender) = all {
+        dr' = dr.update(node, drs => drs.enableRoute(sender)),
+        incomingMsgs' = _incomingMsgs,
+        peers' = peers,
+        mempool' = mempool,
+        senders' = senders,
+        rc' = rc,
+    }
+    ```
+    This will allow traffic to flow again to the sender, and other nodes will dynamically adapt to the
```
cason, line 513: To and from the sender, right?

cason (approved):
Very good formalization work.

I think my most relevant comment is the decision when to send `HaveTx` messages. It is not clear, until we go deep into details, that we send it at most once every cycle of the redundancy manager loop.

Regarding the generated Quint models, I don't think we can actually run any of them, right? Maybe as a follow-up task, create a model that allows playing with DOG, for instance, but setting up the required constants, defining the default `init` and `step` actions.

spec/mempool/gossip/dog.md
hvanz: Do you think then 1000ms is not enough for an end-to-end RTT? That is, should we increase the minimal value `adjustInterval` to 1500ms, 2000ms, or more?

spec/mempool/gossip/dog.md
hvanz: Only to the sender. I fixed the spec and updated the text.

spec/mempool/gossip/dog.md
hvanz: Thanks for pointing out. I made it more clear in a couple of places in the Redundancy Control section.

spec/mempool/gossip/dog.md
```diff
@@ -0,0 +1,606 @@
...
+    ```
+
+2. A node handles a message received from a peer.
+    ```bluespec "steps" +=
+    // Peer message handling
+    nondet node = oneOf(nodesInNetwork)
+    node.receiveFromPeer(handleMessage),
+    ```
+
+3. A node disseminates a transaction currently in its mempool to a subset of peers.
```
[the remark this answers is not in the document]
hvanz: That's right. It was only mentioned in the mempool module. I added a new paragraph explaining here.

spec/mempool/gossip/dog.md
```diff
@@ -0,0 +1,606 @@
...
+
+```bluespec "routing" +=
+def DisabledRoutes(node) = dr.get(node)
+pure def isSourceOrTargetIn(node, route) = node == route._1 or node == route._2
+```
+</details>
+
+## Redundancy Control
+
+The Redundancy Controller (RC) implements a closed-loop mechanism that dynamically adjusts the level
```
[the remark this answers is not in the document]
hvanz: [Closed-loop](https://en.wikipedia.org/wiki/Closed-loop_controller) as the term used in control systems. I expanded the text to make it more clear.

spec/mempool/gossip/README.md
```diff
@@ -17,6 +17,18 @@ transactions in the network.
...
+- [Dynamic Optimal Graph (DOG)](dog.md). Building on Flood, DOG introduces improvements that
+  significantly reduces redundant messages, enhancing bandwidth efficiency while maintaining similar
+  latency and resilience properties.
+  - Pros:
+    + Latency: DOG does not introduce delays or extra messages that add latency as in "pull"
+      mechanisms. Instead, it selectively filters transactions messages before forwarding them to
+      peers.
+    + Byzantine Fault Tolerance (BFT): DOG keeps a minimum level of transaction redundancy for
+      preserving the resilience needed to mitigate the impact of malicious attacks.
+    + Bandwidth: DOG implements a routing mechanism that reduces bandwidth usage by orders of
```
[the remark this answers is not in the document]
hvanz: Changed to "significantly"

spec/mempool/gossip/README.md
```diff
@@ -17,6 +17,18 @@ transactions in the network.
...
 
+- [Dynamic Optimal Graph (DOG)](dog.md). Building on Flood, DOG introduces improvements that
+  significantly reduces redundant messages, enhancing bandwidth efficiency while maintaining similar
+  latency and resilience properties.
+  - Pros:
+    + Latency: DOG does not introduce delays or extra messages that add latency as in "pull"
+      mechanisms. Instead, it selectively filters transactions messages before forwarding them to
+      peers.
+    + Byzantine Fault Tolerance (BFT): DOG keeps a minimum level of transaction redundancy for
+      preserving the resilience needed to mitigate the impact of malicious attacks.
```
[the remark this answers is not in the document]
hvanz: Fixed.

spec/mempool/gossip/README.md
```diff
@@ -17,6 +17,18 @@ transactions in the network.
...
+    + Bandwidth: DOG implements a routing mechanism that reduces bandwidth usage by orders of
+      magnitude when compared to Flood.
+
```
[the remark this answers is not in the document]
hvanz: That makes sense. I added two.

spec/mempool/gossip/dog.md
hvanz: Below it writes 1000ms. I was thinking of using this as default value. Or maybe 1500ms to be sure.

spec/mempool/gossip/dog.md
```diff
@@ -0,0 +1,606 @@
...
+	This interval should allow sufficient time for control messages (`HaveTx` and `Reset`) to
+	propagate through the network and take effect.
+	
+    A minimum value for `adjustInterval` depends on the network’s round-trip time (RTT). Assuming that
+    global latency typically stays below 500ms (maybe an excessive number, need references here), the
+    interval should be set to at least 1000ms to ensure stability and avoid over-adjustment.
+
+    Optimal values depend on empirical measurements of network latency. However, in practice, a value of
+    1000ms or above is recommended to allow message processing and delivery times in diverse network
+    environments.
```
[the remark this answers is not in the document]
hvanz: added

PR #4478, opened 2024-11-13
e2e:  fix oscilate validator set logic when the manifest contains validator updates

cason:
WIP.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/test/e2e/pkg/testnet.go
+++ b/test/e2e/pkg/testnet.go
@@ -371,6 +371,14 @@ func NewTestnetFromManifest(manifest Manifest, file string, ifd InfrastructureDa
 
 		const flipSpan = 3000
 		for i := max(1, manifest.InitialHeight); i < manifest.InitialHeight+flipSpan; i++ {
+			// FIXME: we do not flip the validator when there is
+			// **any** scheduled validator update for that height.
+			// We may have a validator update that affects a
+			// **different** validator here and the height can be
+			// odd. We add our validator back in odd heights, so we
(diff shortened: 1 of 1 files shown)


andynog:
@cason please let me know if you need help with this. I was looking at the initial implementation and it seems you are going into the direction of adding validator updates to the manifest.toml file (?). This will probably need some changes to the generator too.

cason:
> @cason please let me know if you need help with this. I was looking at the initial implementation and it seems you are going into the direction of adding validator updates to the manifest.toml file (?). This will probably need some changes to the generator too.

Yes, I need. The `generator` has nothing to do here. It indeed produce validator updates at random heights, but this logic is ok.

The problem is that in the new logic, the `setup` method of the runner produce oscillations, validator updates at every height. Those are written to the application's config file.

The problem is that the updates produced by the generator (which are more relevant changes) might conflict with the validator updates produced by the oscillation logic (which just turn on and off a validator, with voting power 1 and 0). The latter has to be fixed.

cason:
By the way, if no one has bandwidth to proceed on this work, we should open an issue.

Until then, as workaround, we should probably disable the `ConstantFlip` flag in the produced manifests for nightly executions. Otherwise, the nightly e2e tests will continue fail for this reason, which is a lot of clutter.

jmalicevic (approved):

cason:
@andynog, feel free to merge it if you find it legit. So we don't have that much failures from this night's tests.

andynog (approved):
lgtm

PR #4484, opened 2024-11-13
chore(state): `v1LegacyLayout` keys no longer use `fmt.Sprintf()`.

alesforz:
The `v1LegacyLayout` type uses `fmt.Sprintf()` to create the DB keys. This PR substitutes `fmt.Sprintf` with string concatenation, which is 3-4 times faster for such a simple string.

In general, we shouldn't use `fmt.Sprintf()` to concatenate strings, because it's inefficient. 
Exceptions could be constructing error messages (but we should use `fmt.Errorf()` for that) or tests.

---

#### PR checklist

- ~[ ] Tests written/updated~

### DIFF
--- a/.changelog/unreleased/improvements/4484-v1keylayout-sprintf.md
+++ b/.changelog/unreleased/improvements/4484-v1keylayout-sprintf.md
@@ -1,0 +1,1 @@
+- v1LegacyLayout keys no longer use fmt.Sprintf ([\#4484](https://github.com/cometbft/cometbft/pull/4484))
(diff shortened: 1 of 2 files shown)


melekes (approved):
👍

jmalicevic:
Maybe you can do the same for the blockStore? https://github.com/cometbft/cometbft/blob/alesforz/v1keylayout-sprintf/store/db_key_layout.go

alesforz:
> Maybe you can do the same for the blockStore? https://github.com/cometbft/cometbft/blob/alesforz/v1keylayout-sprintf/store/db_key_layout.go

Will do this in a separate PR 👍

alesforz:
**Note**: the changes in this PR are superseded by #4490.

PR #4489, opened 2024-11-13
chore(store): `store.v1LegacyLayout` no longer uses `fmt.Sprintf()` to create DB keys.

alesforz:
The `v1LegacyLayout` type of pkg `store` uses `fmt.Sprintf()` to create the DB keys. 
This PR refactors functions using `fmt.Sprintf` to improve their speed and memory usage.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/store/db_key_layout.go
+++ b/store/db_key_layout.go
@@ -1,7 +1,7 @@
 package store
 
 import (
-	"fmt"
+	"strconv"
 
 	"github.com/google/orderedcode"
 )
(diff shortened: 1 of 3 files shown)


store/db_key_layout.go
```diff
@@ -24,32 +24,37 @@ type v1LegacyLayout struct{}
...
 // CalcBlockPartKey implements BlockKeyLayout.
 func (*v1LegacyLayout) CalcBlockPartKey(height int64, partIndex int) []byte {
-	return []byte(fmt.Sprintf("P:%v:%v", height, partIndex))
+	var (
+		keyPrefix = "P:"
+		keySuffix = strconv.FormatInt(height, 10) + ":" + strconv.Itoa(partIndex)
+		keyStr    = keyPrefix + keySuffix
+	)
+	return []byte(keyStr)
 }
```
alesforz, line 49: `CalcBlockPartKey` can also be implemented using a `strings.Builder`:
```go
func (*v1LegacyLayout) CalcBlockPartKey(height int64, partIndex int) []byte {
    prefix = "P:"
      
    var sb strings.Builder
    sb.WriteString(prefix)
    sb.WriteString(strconv.FormatInt(x, 10))
    sb.WriteByte(':')
    sb.WriteString(strconv.Itoa(y))
      
    return []byte(sb.String())
}
```
Benchmarks on my laptop show using a `Builder` is ~27% faster then raw string concatenation.
However, the `Builder` allocates 8 bytes per call, whereas raw string concatenation allocates none.

Given that we call this function whenever we load/store block parts (i.e., a lot), I decided to implement raw string concatenation to avoid creating (cumulatively) quite some garbage for the GC to collect.

What do you think? If we prefer speed over memory allocs, I can change this to use a `Builder`.

store/db_key_layout.go
```diff
@@ -24,32 +25,40 @@ type v1LegacyLayout struct{}
...
 
 // CalcBlockHashKey implements BlockKeyLayout.
 func (*v1LegacyLayout) CalcBlockHashKey(hash []byte) []byte {
-	return []byte(fmt.Sprintf("BH:%x", hash))
+	hexHash := make([]byte, hex.EncodedLen(len(hash)))
+	hex.Encode(hexHash, hash)
+
+	return append([]byte{'B', 'H', ':'}, hexHash...)
 }
```
alesforz, line 37: Yes, this is faster than using `fmt.Sprintf` 😛 

On another note: why do we encode to hex the hash before using it as the key's suffix?

alesforz:
TODO: Thought about another refactoring to make this faster and with less memory usage. Will work on it tomorrow.

store/db_key_layout.go
```diff
@@ -23,33 +24,72 @@ type BlockKeyLayout interface {
...
-	return []byte(fmt.Sprintf("C:%v", height))
+// It builds a key in the format "C:height".
+func (v *v1LegacyLayout) CalcBlockCommitKey(height int64) []byte {
+	return v.buildKey([]byte{'C', ':'}, height)
 }
 
 // CalcBlockHashKey implements BlockKeyLayout.
+// It builds a key in the format "BH:hash".
 func (*v1LegacyLayout) CalcBlockHashKey(hash []byte) []byte {
-	return []byte(fmt.Sprintf("BH:%x", hash))
```
alesforz, line 32: Why do we encode to hex the hash before using it as the key's suffix?

alesforz:
Some benchmarks follow.

### `CalcBlockHashKey()`
```
BenchmarkKeyHex/Sprintf-11          14093776    85.42 ns/op    104 B/op    2 allocs/op
BenchmarkKeyHex/Preallocate-11      35725405    34.61 ns/op     80 B/op    1 allocs/op
BenchmarkKeyHex/NoHex-11            97159107    12.39 ns/op     48 B/op    1 allocs/op
```
`Sprintf` is what's in `main` now. `Preallocate`  is the refactoring in this PR. `NoHex` is if we don't hex the hash `[]byte`.

### `CalcBlockPartKey()`
```
BenchmarkMultiPartStr/Sprintf-11         20937796    57.44 ns/op    8 B/op    1 allocs/op
BenchmarkMultiPartStr/Preallocate-11    174204424     6.872 ns/op   0 B/op    0 allocs/op
````
`Sprintf` is what's in `main` now. `Preallocate`  is the refactoring in this PR.

### Other Methods
They all compute keys of the form `<some_prefix>:<height>` so benchmarks are the same.
```
BenchmarkKey/Sprintf-11              33424916    35.72 ns/op    5 B/op    1 allocs/op
BenchmarkKey/Preallocate-11         475301016     2.503 ns/op   0 B/op    0 allocs/op
```
`Sprintf` is what's in `main` now. `Preallocate`  is the refactoring in this PR.

jmalicevic (approved):

PR #4493, opened 2024-11-15
feat(docs): ADR-119 : Mempool DOG

### LINKED ISSUE #4319: mempool/dog: Write ADR
Write an ADR for the protocol implementation given the spec written in #4318.

jmalicevic:
Closes #4319 

[adr-119-dog-mempool-gossip.md](https://github.com/cometbft/cometbft/blob/jasmina/dog-adr/docs/references/architecture/adr-119-dog-mempool-gossip.md)

---

#### PR checklist

- [ ] Tests written/updated


docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,239 @@
...
+type Reset struct {
+
+}
+```
+
+#### Monitoring
+
+This ADR introduces a set of metrics which can be used to observe the parameters of the protocol. 
+
+- `HaveTxMsgsReceived`
```
hvanz, line 137: I don't think we need all these metrics in the final version. I would like to see the redundancy level, but instead of `Redundancy` I would replace it with two new metrics `FirstTimeTxs` and `DuplicateTxs` from where we can compute redundancy. In fact we already have the number of duplicates in `AlreadyReceivedTxs`.

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,239 @@
...
+- `DisabledRoutes` - Number of disabled routes.
+
+- `Redundancy` -  The current level of redundancy computed as the ratio between duplicate and unique transactions. 
+
+
+
+### Configuration parameters
+
+
+- `config.EnableDOGProtocol: bool`: Enabling or disabling the DOG protocol
```
hvanz, line 164: Ideally, on the final version we would leave just one configuration: `EnableDOGProtocol`. For the other parameters described in the spec (`TargetRedundancy`, `TargetRedundancyDeltaPercent`, and (now) `AdjustInverval`) I would try to define values based on the experiments, and then make them constants  in the code. Currently, those numbers are `TargetRedundancy = 0.5`, `TargetRedundancyDeltaPercent = 20`, and `AdjustInverval = 1000`.

hvanz: We need to decide if `EnableDOGProtocol` will be `true` by default.

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,239 @@
...
+    peer.send(Reset)
+  if redundancy > redundancyUpperBound
+    enableRouteDisabling() 
+```
+
+Redundancy is adjusted whenever a new transaction is added into the mempool.
+(TODO - revisit whether this should also be done on duplicates.)
+
+
+### Impacted areas of code
```
hvanz, line 73: The `Mempool` interface will have a new method `GetSenders(txKey types.TxKey) ([]nodekey.ID, error)`, and the `Entry` interface, `Senders() []nodekey.ID`.

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,239 @@
...
+	mtx          cmtsync.RWMutex
+	firstTimeTxs int64 // number of transactions received for the first time
+	duplicates   int64 // number of duplicate transactions
+
+	// If true, do not send HaveTx messages.
+	blockHaveTx atomic.Bool
+}
+
+```
+
```
hvanz, line 114: In addition to the new data structures, I would include the changes to the current code. 
- Handling of new messages in `Receive` function, in the new channel.
- Extend `TryAddTx` to count first-time and duplicate txs; reply with a `HaveTx` message if tx is in cache. 
- Filter disabled routes in `broadcastTxRoutine` (only if DOG is enabled, otherwise keep filtering senders as in the current code).
- New logic on `RemovePeer`.

hvanz: It's important to emphasize that the new code does not change the current behaviour.

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,239 @@
...
+## Consequences
+
+> This section describes the consequences, after applying the decision. All
+> consequences should be summarized here, not just the "positive" ones.
+
+### Positive
+
+### Negative
+
```
hvanz, line 231:
```suggestion
- The new protocol is not compatible with the existing experimental feature that limits disseminating transactions up to a specified number of peers. They cannot be both enabled simultaneously. This needs to be clear in the documentation.
```

docs/references/architecture/adr-119-dog-mempool-gossip.md
jmalicevic: I removed the HaveTx and Reset msgs from the metrics but I left redundancy. I think it's still beneficial to have it pre-computed because we reset these values (`FirstTimeTxs` and `DuplicateTxs`) periodically and the operator might not adjust the computation correctly. Not sure if it would be a problem but it's safer if we just compute it.

hvanz: I forgot that `DuplicateTxs` is not the same as `AlreadyReceivedTxs` because the first one is reset on every adjustment. Anyway, we would still need to consider `FirstTimeTxs` and `DuplicateTxs` instead of `Redudancy`.
In many testnets I saw that dividing the `DuplicateTxs` metric by the `FirstTimeTxs` metric is more precise than the `Redundancy` metric.

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+is due to transaction gossiping. 
+
+DOG is a protocol that aims to reduce the number of duplicate transactions sent
+while maintaining the resilience to attacks. 
+
+## Alternative Approaches
+
+The existing alternative approaches for transaction gossiping are:
+- FLOOD - the current transaction gossiping protocol in CometBFT
+- CAT Mempool - the transaction gossiping proposed by Celestia
```
cason, line 30: Link?

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+- Limiting number of peers to send a transaction to - Experimental protocol in CometBFT
```
cason, line 31: Links?

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+- Etherium's transaction gossiping protocol.
```
cason, line 32: How about a link?

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+
+
+### FLOOD - the current transaction gossiping protocol in CometBFT
+
+Currently, CometBFT gossips each transaction to all the connected peers, except the sender
+of the transaction. Thus, for every transaction, CometBFT keeps a list of senders to make sure
+it is not sent again to a node that has sent it.
```
cason, line 39:
```suggestion
it is not sent again to a node from which the node has received it.
```

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+- Etherium's transaction gossiping protocol.
+
+
+### FLOOD - the current transaction gossiping protocol in CometBFT
+
+Currently, CometBFT gossips each transaction to all the connected peers, except the sender
+of the transaction. Thus, for every transaction, CometBFT keeps a list of senders to make sure
+it is not sent again to a node that has sent it. 
+This does not prevent a node to send the transaction to a node that has sent it to the 
+sender, thus creating a lot of duplicates.
```
cason, line 41: This definition of sender is complex...

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+
+We start the design section with a general description of how the algorithm works. 
+
+CometBFT caches received transactions and stores the IDs of all the peers that have sent it in a `senders` list. When a transaction enters the mempool,
+if it was already received it can be retrieved from the cache. When this happens a certain number of times, the receiving node notifies the sender to stop forwarding any transaction from the peer 
+that had originally sent the transaction. The exact point when the notification is sent is defined below as a redundancy threshold. (TODO LiNK)
+
+Namely, let's assume that transaction `tx_1` was sent by node `B` to node `C` which forwarded it to `A`. `A` already has the transaction and notifies node `C` about this. Node `C` will look up the senders of transaction `tx_1` and in the future disable transaction forwarding from node `B` to node `A`.
+
+In reality, if node `C` ha received `tx_1` from multiple senders (`X`,`Y`, `B`), it will pick the first one that has sent it and disable that route.
```
cason, line 63: The first one, except the first one? I mean, the transaction must be received first, then duplicates are.

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+### Impacted areas of code
+
+The bulk of the changes is constrained to the mempool reactor. 
+
+
+#### Breaking changes and new additions
+
+- The `Mempool` interface is expanded with a method : 
+```go
+// GetSenders returns the list of node IDs from which we receive the given transaction.
```
cason, line 105:
```suggestion
// GetSenders returns the list of node IDs from which we have received a transaction.
```

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+GetSenders(txKey types.TxKey) ([]nodekey.ID, error)
```
cason, line 106:
```suggestion
GetSenders(txKey types.TxKey) ([]p2p.ID, error)
```

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+``` 
+- The `Entry` interface in the mempool package is expanded with a method:
+```go
+// Senders returns the list of registered peers that sent us the transaction.
+	Senders() []nodekey.ID
```
cason, line 111:
```suggestion
Senders() []p2p.ID
```

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+```
+- We introduce a new communication channel, called `Mempool Control Channel`. The channel
+is used to transmit the messages needed for the implementation of the protocol. The channel ID
+is `31` and it has a priority of `10` .
```
cason, line 115:
```suggestion
is `31` and it has priority `10`.
```

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+
+### Redundancy control
+
+In an ideal setting, receiving a transaction more than once is not needed. But in a Byzantine setting, allowing for only one route for a transaction can introduce an attack vector. 
+
+Operators can thus configure a desired redundancy and the gossip protocol will adjust route disabling based on this setting. Redundancy is defined as the ratio between duplicate and unique transactions. 
+
+<!-- Operators can also define how much they tolerate the redundancy to deviate from the desired level. -->
+
+By default, the redundancy is set to `1` and impacts the gossiping of transactions as follows:
```
cason, line 77: Here I have more a question than anything.

I see a factor as something that starts from 1. There is not 0 factor. So when I read that the target redundancy is 1, I think that there is no redundancy.

We should understand which way to measure it is more clear to the users.

Notice that in practice this does not change anything, essentially, as now `redundancy := allReceivedTxs / firstTimeTx`, where `allReceivedTxs := firstTimeTx + duplicateTxs`.

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+
+`Receive`
+
+The function now handles messages from two channels. The `MempoolControlChannel`
+transmits `HaveTx` and `Reset` messages that keep track of routes between peers. 
+
+`TryAddTx` 
+
+When DOG is enabled we count the number of unique and duplicate transactions, and send
+`HaveTx` messages if the redundancy is too high.
```
cason, line 165: I thought this action was periodical, not triggered by the reception of a message.

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+The function now handles messages from two channels. The `MempoolControlChannel`
+transmits `HaveTx` and `Reset` messages that keep track of routes between peers. 
+
+`TryAddTx` 
+
+When DOG is enabled we count the number of unique and duplicate transactions, and send
+`HaveTx` messages if the redundancy is too high. 
+
+Once a `HaveTx` message is sent,
+sending further messages is blocked. It is unblocked by the redundancy control mechanism
```
cason, line 168: Sending what is blocked?

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+
+Once a `HaveTx` message is sent,
+sending further messages is blocked. It is unblocked by the redundancy control mechanism 
+once it is triggered again. This will be discussed below, but overall the frequency
+of `HaveTx` messages should be aligned with the time it takes to send a message and 
+adjust routing between peers. 
+
+`broadcastTxRoutine`
+
+Before sending a transaction, we filter the peers to send to based on the `disabeldRoutes`. Again,
```
cason, line 175:
```suggestion
Before sending a transaction, we filter the peers to send to based on the `disabledRoutes`. Again,
```

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+only if DOG is enabled. 
+
+`RemovePeer`
+
+When a peer is removed, we remove any existing entries related to this peer from the `DisabledRoutes`
```
cason, line 180:
```suggestion
When a peer is removed, we remove any existing entries related to this peer from the `disabledRoutes` 
```

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+This lead to routes beeing disabled too many times, and the process of stabilizing redundancy would be re-triggered
+too frequently. The adapted logic relies in the redundancy controller to decide whether a `Reset` 
+message should be sent. 
+
+
+#### New p2p messages
+
+The protocol introduces two new p2p messages whose protobuf definition is given below:
+
+```proto
```
cason, line 194: Does not look like proto.

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+}
+```
+
+#### Monitoring
+
+The impact of the protocol on operations can also be observed by looking at the following metrics that already exist:
+
+- `AlreadyReceivedTx` - the number of redundant transactions. When DOG is enabled these values should drop. 
+
+- `BytesReceived` - This metric shows the number of bytes received per message type. Without DOG, the transactions dominate the number of bytes, while, when enabled, the block parts dominate.
```
cason, line 212:
```suggestion
- `BytesReceived` - This metric shows the number of bytes received per message type. Without DOG, the transactions tend to dominate the number of bytes, while, when enabled, the block parts should dominate it.
```

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+
+This ADR introduces a set of metrics which can be used to observe the parameters of the protocol:
```
cason, line 214: I assume these are metrics of the `mempool` module.

cason (commented):
I know that this is a draft, but I reviewed this PR in order to learn more about DOG.

Left a couple of comments and suggestions.

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+- 2024-11-25: Applied comments from @hvanz and added information on decided metrics (@jmalicevic)
+
+## Status
+Accepted: [#3297][tracking-issue].
+
+
+## Context
+
+The current transaction dissemination protocol of the mempool sends each received 
+transaction to all the peers a node is connected to, except for the sender.
```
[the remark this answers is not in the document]
jmalicevic: Based on [this](https://github.com/cometbft/cometbft/blob/541c4667db44a88914f8b8984ca3013e2e2b3ce8/mempool/mempoolTx.go#L38) I would say it does not send it to any recorded sender.

docs/references/architecture/adr-119-dog-mempool-gossip.md
jmalicevic: At the moment it just gets `senders[0]` and disabls the route between `senders[0]->A`

docs/references/architecture/adr-119-dog-mempool-gossip.md
jmalicevic: Ok noted, we are keeping it in line with main until it gets reverted but note to change it.

docs/references/architecture/adr-119-dog-mempool-gossip.md
jmalicevic: 5

docs/references/architecture/adr-119-dog-mempool-gossip.md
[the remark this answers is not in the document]
cason: We can receive txs from RPC. There is no sender in this case.

docs/references/architecture/adr-119-dog-mempool-gossip.md
jmalicevic: The count is done on reception, and HaveTx will be sent only then. Whether it is sent is decided periodically. 

I will clarify this but basically:
- OnNewMessage: 
   incDuplicate
       if(DogEnabled)
          if(HaveTxUnblocked)
             SendHaveTx

another process 

Every 1s {
  redundancy = computeRedundancy()
  if redundancy > threshold
      unblockHaveTx
  if redundancy < threshold
     resetRoutes

}

docs/references/architecture/adr-119-dog-mempool-gossip.md
jmalicevic: `HaveTx`. Fixed that.

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+
+Currently, CometBFT gossips each transaction to all the connected peers, except the sender
+of the transaction. Thus, for every transaction, CometBFT keeps a list of senders to make sure
+it is not sent again to a node that has sent it. 
+This does not prevent a node to send the transaction to a node that has sent it to the 
+sender, thus creating a lot of duplicates. 
+
+This is the exact problem DOG was designed to tackle.
+
+
```
hvanz, line 45:
```suggestion
### CAT protocol

Content-Addressable Transaction (CAT) is a "push-pull" gossip protocol originally implemented by Celestia on top of CometBFT's priority-mempool (aka v1, previously also in Tendermint Core), which existed in CometBFT until v0.37. 

With CAT, nodes forward ("push") transactions received from users via RPC endpoints to their peers. When a node `A` receives a transaction from another node, it will notify its peers that it has the transaction with a `SeenTx(hash(tx))` message. A node `B`, upon receiving the `SeenTx` message from `A`, will check that if the transaction is in its mempool. If not, it will "pull" the transaction from `A` by sending a `WantTx(hash(tx))` message, to which `A` responds with a `Tx` message carrying the full transaction. While `SeenTx` and `HaveTx` messages are much smaller than `Tx` messages, these additional communication steps introduce latency when disseminating transactions. The CAT protocol was especially designed for networks with low throughput and large transaction sizes. 

Efforts to port the CAT mempool to CometBFT were documented in #2027. Experimental results on a small testnet from #1472 showed that CAT effectively reduces bandwidth usage. However, its impact on latency was not evaluated in those tests. Porting CAT was finally deprioritised in favor of DOG.

```

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+- Entire network should use DOG. Otherwise the impact will be minimal. 
+
+
+> The protocol implicitly favors routes with low latency, by cutting routes to peers that send the duplicate transaction at a later time.
+
+- Is this true? We don't cut ties with the sender of the transaction, rather we are telling it to cut ties with some other node that has sent us the tx before that. 
+So sender could still forward us transactions but not the ones received from this particular node. 
+
+- If the frequency of `HaveTx` messages is too high, nodes will have too many routes cut.
+- Except for enabling DOG on all nodes, users do not have to change any other aspects of their application.
```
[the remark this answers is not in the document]
jmalicevic: > So we don't support a mixed network?

We actually do. If there are nodes with DOG disabled, those will not get the protocol messages.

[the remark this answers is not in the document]
hvanz: The new `MempoolControlChannel` will not be opened during the p2p handshake. DOG nodes will just log an error message when trying to send `HaveTx` or `Reset` messages to non-DOG peers.

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,351 @@
...
+
+## Context
+
+The current transaction dissemination protocol of the mempool sends each received 
+transaction to all the peers a node is connected to, except for the sender.
+In case transactions were received via RPC, they are broadcasted to all
+connected peers. 
+While resilient to Byzantine attacks, this type of transaction gossiping is 
+causing a lot of network traffic and nodes receiving duplicate transactions
+very frequently.
```
hvanz, line 21:
```suggestion
While this approach provides resilience to Byzantine attacks and optimal transaction latency, it generates significant network traffic, with nodes receiving an exponential number of duplicate transactions.
```

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,351 @@
...
+connected peers. 
+While resilient to Byzantine attacks, this type of transaction gossiping is 
+causing a lot of network traffic and nodes receiving duplicate transactions
+very frequently. 
+
+Benchmarks have also confirmed a large portion of the sent and received bytes 
+by the network layer of a node is due to transaction gossiping. 
+
+DOG is a protocol that aims to reduce the number of duplicate transactions received
+while maintaining the resilience to attacks.
```
hvanz, line 27:
```suggestion
while maintaining the resilience to attacks and low transaction latency. 
```

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,351 @@
...
+
+Efforts to port the CAT mempool to CometBFT were documented in [\#2027]. Experimental results on a small testnet from [\#1472] showed that CAT effectively reduces bandwidth usage. However, its impact on latency was not evaluated in those tests. Porting CAT was finally deprioritised in favor of DOG.
+
+### Limiting the number of peers a transaction is forwarded to
+
+CometBFT allows operators to configure `p2p.experimental_max_gossip_connections_to_non_persistent_peers` and `p2p.experimental_max_gossip_connections_to_persistent_peers`  as a maximum number of peers to send  transactions to.
```
hvanz, line 60:
```suggestion
CometBFT allows operators to configure `p2p.experimental_max_gossip_connections_to_non_persistent_peers` and `p2p.experimental_max_gossip_connections_to_persistent_peers`  as a maximum number of peers to send  transactions to ([#1558](https://github.com/cometbft/cometbft/pull/1558) and [#1584](https://github.com/cometbft/cometbft/pull/1584)). 
```

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,351 @@
...
+ can be found in Issue [\#4320].
+
+
+
+## User recommendation
+
+- The Entire network should use DOG. Otherwise the impact will be minimal. 
+
+
+- The protocol implicitly favors faster routes, by cutting routes to peers that send the duplicate transaction at a later time.
```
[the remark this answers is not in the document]
jmalicevic: Actually should this not be rephrased to 

The protocol implicitly favors faster routes, by cutting routes **through** peers that send the duplicate transaction at a later time.

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,363 @@
...
+Overall DOG reduces the bandwidth used by CometBFT, freeing up the network for consensus related messages. 
+For DOG to work as expected, the entire network should be using this, although it is not required for the network
+to behave correctly. It should however not be used in combination with the parameters `p2p.experimental_max_gossip_connections_to_non_persistent_peers` and `p2p.experimental_max_gossip_connections_to_persistent_peers` set to `true`.
+
+### Positive
+
+- Except for enabling DOG on all nodes, users do not have to change any other aspects of their application. 
+- Gossiping using DOG leads to a significant reduction in network traffic caused by transactions. 
+- Routes with lower latencies are implicitly favoured by cutting routes to peers that send the duplicate 
+- By cutting routes that go through a peer that sent the duplicates later in time, the protocol implicitly favors faster routes.
```
hvanz, line 341: These two items are the same.

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,363 @@
...
+## Consequences
+
+Overall DOG reduces the bandwidth used by CometBFT, freeing up the network for consensus related messages. 
+For DOG to work as expected, the entire network should be using this, although it is not required for the network
+to behave correctly. It should however not be used in combination with the parameters `p2p.experimental_max_gossip_connections_to_non_persistent_peers` and `p2p.experimental_max_gossip_connections_to_persistent_peers` set to `true`.
+
+### Positive
+
+- Except for enabling DOG on all nodes, users do not have to change any other aspects of their application. 
+- Gossiping using DOG leads to a significant reduction in network traffic caused by transactions.
```
hvanz, line 339:
```suggestion
- Gossiping using DOG leads to a significant reduction in network traffic caused by transactions. 
  Moreover, in our experiments most metrics have improved, with no observed degradation in any area. The 
  reduction of redundant messages has a system-wide positive impact, contributing to faster consensus, lower 
  transaction latency, and more efficient resource utilization (#4606).
```

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,363 @@
...
+### Positive
+
+- Except for enabling DOG on all nodes, users do not have to change any other aspects of their application. 
+- Gossiping using DOG leads to a significant reduction in network traffic caused by transactions. 
+- Routes with lower latencies are implicitly favoured by cutting routes to peers that send the duplicate 
+- By cutting routes that go through a peer that sent the duplicates later in time, the protocol implicitly favors faster routes.
+
+### Negative
+
+- The new protocol is not compatible with the existing experimental feature that limits disseminating transactions up to a specified number of peers. If both are enabled simultaneously, the mempool will work but not as expected.
```
hvanz, line 345: I would move this under Neutral.

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,363 @@
...
+to behave correctly. It should however not be used in combination with the parameters `p2p.experimental_max_gossip_connections_to_non_persistent_peers` and `p2p.experimental_max_gossip_connections_to_persistent_peers` set to `true`.
+
+### Positive
+
+- Except for enabling DOG on all nodes, users do not have to change any other aspects of their application. 
+- Gossiping using DOG leads to a significant reduction in network traffic caused by transactions. 
+- Routes with lower latencies are implicitly favoured by cutting routes to peers that send the duplicate 
+- By cutting routes that go through a peer that sent the duplicates later in time, the protocol implicitly favors faster routes.
+
+### Negative
```
hvanz, line 343: We need to put something in Negative. I would add something like:
- It takes around 5 to 15 minutes for nodes to build a stable routing table, depending on network conditions and node configuration. This shouldn't pose a problem as it is expected for nodes to run for a long time.

docs/references/architecture/adr-119-dog-mempool-gossip.md
hvanz: In the end we decided to leave only `Redundancy`. `FirstTimeTxs` and `DuplicateTxs` were useful for the initial experiments but will not be needed for nodes in production.

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,363 @@
...
+
+The third parameter defines the bounds of acceptable redundancy levels: 
+- `TargetRedundancyDeltaPercent: float`: Set to `0.1` (10%). It defines the bounds
+of acceptable redundancy levels. The actual redundancy will be: `redundancy +- redundancy*delta`. A lower delta 
+did not lead to a visible reduction in redundancy, while slightly increasing the number of control messages sent. 
+We have therefore opted to not reduce this value further. It does not impact the protocol as much as the target redundancy or the redundancy adjustment interval. 
+
+Part of the work on the protocol was extensive testing of its performance and impact on the network, as well 
+as the impact of the network configuration and load on the protocol itself. 
+Issue [\#4320] covers the experiments performed on DOG.
```
hvanz, line 300:
```suggestion
Issue [\#4543] covers the experiments performed on DOG.
```

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,363 @@
...
+- The Entire network should use DOG. Otherwise the impact will be minimal. 
+
+
+- The protocol implicitly favors faster routes, by cutting routes through peers that send the duplicate transaction at a later time.
+
+
+- If the frequency of `HaveTx` messages is too high, nodes will have too many routes cut.
+
+-  While we did not observe any single node missing out on transactions when the lowering the target 
+reduncancy event to `0.1`, we do not advise users to do this without a strong justification and testing.
```
hvanz, line 326: We have not tested a 0.1 target with perturbations and sending each tx to multiple nodes. So no, it's not recommended.

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,367 @@
...
+### Limiting the number of peers a transaction is forwarded to
+
+CometBFT allows operators to configure `p2p.experimental_max_gossip_connections_to_non_persistent_peers` and `p2p.experimental_max_gossip_connections_to_persistent_peers`  as a maximum number of peers to send  transactions to ([#1558](https://github.com/cometbft/cometbft/pull/1558) and [#1584](https://github.com/cometbft/cometbft/pull/1584)). 
+This reduces bandwitdth compared to when they are disabled but there is no rule to determine which peers transactions
+are not forwarded to. DOG can be looked at as an enahanced, more informed version of this protocol.
+
+A node can use either this or DOG, but not both at the same time.
+
+
+# libp2p's gossipSub gossiping protocol
```
hvanz, line 66:
```suggestion
### libp2p's gossipSub protocol
```

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,367 @@
...
+(`flood` ) dissemination protocol, but make it optional. The protocol can be activated/deactivated
+via a new config flag `mempool.dog_protocol_enabled`. More details on this in the sections below. 
+
+## Detailed Design
+
+We start the design section with a general description of how the algorithm works. The protocol is explained in detail in the accompanying [specification](https://github.com/cometbft/cometbft/blob/main/spec/mempool/gossip/dog.md). 
+
+CometBFT caches received transactions and stores the IDs of all the peers that have sent it in a `senders` list. When a transaction enters the mempool,
+if it was already received it can be retrieved from the cache. When this happens a certain number of times, the receiving node notifies the sender to stop forwarding any transaction from the peer 
+that had originally sent the transaction. The exact point when the notification is sent is defined below as a redundancy threshold (see `Redundancy control`).
```
hvanz, line 88:
```suggestion
CometBFT nodes cache received transactions and store the IDs of all the peers that have sent it in a `senders` list. 
When an incoming transaction is in the cache, the receiving node notifies the sender that it already has that transaction. The sender will stop forwarding any transaction from the peer that had originally sent the transaction.
```

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,367 @@
...
+```go
+// GetSenders returns the list of node IDs from which we have received a transaction.
+GetSenders(txKey types.TxKey) ([]p2p.ID, error)
+``` 
+- The `Entry` interface in the mempool package is expanded with a method:
+```go
+// Senders returns the list of registered peers that sent us the transaction.
+	Senders() []p2p.ID
+```
+- We introduce a new communication channel, called `Mempool Control Channel`. The channel
```
hvanz, line 143:
```suggestion
- We introduce a new communication channel, called `MempoolControlChannel`. The channel
```

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,367 @@
...
+map and explicitly trigger redundancy re-adjustment. The later is not strictly needed, but 
+can lead to quicker propagation of this information through the network.
+
+Note that we had considered a scenario where, upon removing a peer, a node sends `ResetRoute` messages to all its peers. 
+This lead to routes being disabled too many times, and the process of stabilizing redundancy would be re-triggered
+too frequently. The adapted logic relies in the redundancy controller to decide whether a `ResetRoute` 
+message should be sent. 
+
+
+#### New p2p messages
```
hvanz, line 220: I would move this section that describes the new messages at the beginning of the Detailed Design section.

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,367 @@
...
+The CometBFT team has decided to implement the protocol on top of the existing mempool
+(`flood` ) dissemination protocol, but make it optional. The protocol can be activated/deactivated
+via a new config flag `mempool.dog_protocol_enabled`. More details on this in the sections below. 
+
+## Detailed Design
+
+We start the design section with a general description of how the algorithm works. The protocol is explained in detail in the accompanying [specification](https://github.com/cometbft/cometbft/blob/main/spec/mempool/gossip/dog.md). 
+
+CometBFT nodes cache received transactions and store the IDs of all the peers that have sent it in a `senders` list. 
+When an incoming transaction is in the cache, the receiving node notifies the sender that it already has that transaction. The sender will stop forwarding any transaction from the peer that had originally sent the transaction.
```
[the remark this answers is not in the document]
jmalicevic: I will clarify this, you are right. It will basically disable forwarding from the sender to the guy who said that he has the transaction.

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,367 @@
...
+GetSenders(txKey types.TxKey) ([]p2p.ID, error)
+``` 
+- The `Entry` interface in the mempool package is expanded with a method:
+```go
+// Senders returns the list of registered peers that sent us the transaction.
+	Senders() []p2p.ID
+```
+- We introduce a new communication channel, called `MempoolControlChannel`. The channel
+is used to transmit the messages needed for the implementation of the protocol. The channel ID
+is `31` with priority `10` (for context, the mempool channel sending transaction has a priority of `5`).
```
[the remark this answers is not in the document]
jmalicevic: Yes

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,367 @@
...
+
+Efforts to port the CAT mempool to CometBFT were documented in [\#2027]. Experimental results on a small testnet from [\#1472] showed that CAT effectively reduces bandwidth usage. However, its impact on latency was not evaluated in those tests. Porting CAT was finally deprioritised in favor of DOG.
+
+### Limiting the number of peers a transaction is forwarded to
+
+CometBFT allows operators to configure `p2p.experimental_max_gossip_connections_to_non_persistent_peers` and `p2p.experimental_max_gossip_connections_to_persistent_peers`  as a maximum number of peers to send  transactions to ([#1558](https://github.com/cometbft/cometbft/pull/1558) and [#1584](https://github.com/cometbft/cometbft/pull/1584)). 
+This reduces bandwitdth compared to when they are disabled but there is no rule to determine which peers transactions
+are not forwarded to. DOG can be looked at as an enahanced, more informed version of this protocol.
```
melekes, line 61:
```suggestion
are not forwarded to. DOG can be considered an enhanced, more informed version of this protocol.
```

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,367 @@
...
+
+A node can use either this or DOG, but not both at the same time.
```
melekes, line 63: Having both is confusing for users imho, and we should remove the former once DOG is out.

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,367 @@
...
+
+ Peers form meshes, where nodes within a mesh are subscribed to the same topic and the size of the mesh is defined with a configuration parameter. 
+
+ `gossipSub` allows peers to indicate they already have a particular transaction so that other peers do not forward it to them, by sending a `IHAVE` message. That, however this is done for each individual message. DOG on the other hand, instructs peers to stop forwarding anything from a given peer with one message. 
+
+ There is [ongoing work](https://github.com/libp2p/specs/pull/413/files) within `gossipSub` that might achieve something similar. A peer can send a message instructing another peer not to forward anything to it. DOG on the other hand blocks only one route, allowing traffic to continue through the peer.
+
+## Decision
+
+The CometBFT team has decided to implement the protocol on top of the existing mempool
```
melekes, line 78: Is there a plan to make it default eventually?

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,367 @@
...
+
+The received and sent bytes metrics are reported by message type which enables operators to observe the number of `HaveTx` and `ResetRoute` messages sent.
+
+
+This ADR introduces a set of metrics into the `mempool` module. They can be used to observe the parameters of the protocol: 
+
+
+- `mempool.disabled_routes` - Number of disabled routes.
+
+- `mempool.redundancy` -  The current level of redundancy computed as the ratio between duplicate and unique transactions.
```
melekes, line 269: why do some config params have a `dog` prefix while others do not? Shouldn't we either use `dog` prefix everywhere or not use it at all?

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,367 @@
...
+
+- `mempool.dog_protocol_enabled: bool`: Enabling or disabling the DOG protocol. `true` by default.
+The only reason for this is the incompatibility of DOG with the existing experimental feature
+that disables sending transactions to all peers. 
+
+
+The [specification](https://github.com/cometbft/cometbft/spec/mempool/gossip/dog.md) of the protocol introduces 3 additional variables. 
+Out of the tree, we have made the following configuration parameters:
+ 
+- `mempool.dog_adjust_interval: time.Duration`: Set to `1s` by default. Indicates how often the redundancy controller readjusts
```
melekes, line 283: What does it depend on? Does it have to be a configuration parameter?

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,367 @@
...
+```go
+  redundancy := duplicateTxs / firstTimeTxs
+  if redundancy < r.redundancyLowerBound:
+    peer.send(Reset)
+  if redundancy > redundancyUpperBound
+    enableRouteDisabling() 
+```
+
+The number of unique and duplicate transactions is updated as transactions come in and,
+periodically (every `1s`), DOG recomputes the redundnacy of the system.
```
melekes, line 118:
```suggestion
periodically (every `1s`), DOG recomputes the system's redundancy. 
```

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,367 @@
...
+```
+
+The number of unique and duplicate transactions is updated as transactions come in and,
+periodically (every `1s`), DOG recomputes the redundnacy of the system. 
+
+A redundancy of `1` implies that we allow one duplicate transaction for every unique transaction.
+
+#### Redundancy adjustment triggering
+
+As explained, redundancy is adjusted periodicially. It is also explicitly triggered when a peer is disconnected.
```
melekes, line 124:
```suggestion
As explained, redundancy is adjusted periodically. It is also explicitly triggered when a peer is disconnected. 
```

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,367 @@
...
+The number of unique and duplicate transactions is updated as transactions come in and,
+periodically (every `1s`), DOG recomputes the redundnacy of the system. 
+
+A redundancy of `1` implies that we allow one duplicate transaction for every unique transaction.
+
+#### Redundancy adjustment triggering
+
+As explained, redundancy is adjusted periodicially. It is also explicitly triggered when a peer is disconnected. 
+While not required to adjusted immediately on a disconnect, we trigger the re-evaluation of existing redundancy 
+as an optimization to speed up the time between changes in the network and the state of the syetem.
```
melekes, line 126:
```suggestion
as an optimization to speed up the time between changes in the network and the state of the system. 
```

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,367 @@
...
+
+Efforts to port the CAT mempool to CometBFT were documented in [\#2027]. Experimental results on a small testnet from [\#1472] showed that CAT effectively reduces bandwidth usage. However, its impact on latency was not evaluated in those tests. Porting CAT was finally deprioritised in favor of DOG.
+
+### Limiting the number of peers a transaction is forwarded to
+
+CometBFT allows operators to configure `p2p.experimental_max_gossip_connections_to_non_persistent_peers` and `p2p.experimental_max_gossip_connections_to_persistent_peers`  as a maximum number of peers to send  transactions to ([#1558](https://github.com/cometbft/cometbft/pull/1558) and [#1584](https://github.com/cometbft/cometbft/pull/1584)). 
+This reduces bandwitdth compared to when they are disabled but there is no rule to determine which peers transactions
```
melekes, line 60:
```suggestion
This reduces bandwidth compared to when they are disabled, but there is no rule to determine which peers transactions
```

melekes (commented):
great work @jmalicevic and @hvanz 👍

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+  to byzantine attacks.
+   <!-- TODO should we remove this too? If yes, remove the above sentence 
+  regarding operator configuration -->
+
+In the first version of the protocol, we were inclined to expose more fine tuning
+parameters to operators. But after extensive experimentation we have settled on values for 
+each of them. 
+
+
+- `TargetRedundancyDeltaPercent: float`: Value in the range `[0, 1)` that defines the bounds
```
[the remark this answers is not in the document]
hvanz: This delta parameter is a percentage, so 0.2 means 20% of the target value. 

Then the bounds are `target +- target * delta_percent`. With target = 0, the bounds are [0,0].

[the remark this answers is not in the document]
hvanz: In the code this parameter is an integer to make it more simple. We don't so much granularity.

docs/references/architecture/adr-119-dog-mempool-gossip.md
hvanz: There's some confusion about the definition of redundancy level, which we don't define as a factor. Maybe this needs to fixed or clarified in the text.

The definition is `redundancy = duplicateTxs / firstTimeTxs`, and the spec defines special values for corner cases. That's why 0 is a valid value, even if we don't allow a target of 0 redundancy. The default target redundancy is 1.

docs/references/architecture/adr-119-dog-mempool-gossip.md
```diff
@@ -0,0 +1,287 @@
...
+
+
+### Configuration parameters
+
+
+- `config.EnableDOGProtocol: bool`: Enabling or disabling the DOG protocol. `true` by default.
+<!-- TODO verify this. But not a single benchmark
+showed a reason to not enable it. --->
+- `config.TargetRedundancy: float`: The redundancy level that the gossip protocol should aim to
+  maintain. It is `1` by default. It cannot be `0` as this would open the node
```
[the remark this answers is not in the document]
hvanz: We defined `redundancy = duplicateTxs / firstTimeTxs`; you're thinking `redundancy = (firstTimeTxs+duplicateTxs) / firstTimeTxs`. You'd get the same if set to `1.0000000001`.

docs/references/architecture/adr-119-dog-mempool-gossip.md
hvanz: I agree to remove the experimental feature.

docs/references/architecture/adr-119-dog-mempool-gossip.md
hvanz: Yes, it will be enabled by default. It only improves the performance; we haven't seen any disadvantage in having it enabled.

jmalicevic: Should we say so here @hvanz

docs/references/architecture/adr-119-dog-mempool-gossip.md
jmalicevic: We should, I did not update this section apparently, it was written before we renamed them with dog. Good catch, thanks Anton!

docs/references/architecture/adr-119-dog-mempool-gossip.md
jmalicevic: We ran experiments varying number of peers, tx load and target redundancy and did not see a strong indicator to change this to a particularly lower or higher value. But we left it as a config because it might be that for some users this can be too frequent.

melekes (approved):

docs/references/architecture/adr-119-dog-mempool-gossip.md
hvanz: Yes, probably it's better to make it explicit in the ADR.

hvanz (approved):
🚀 🚀

Issue #4486, opened 2024-11-13
CometBFT uses a single DB backend


alesforz:
Partially addresses #1039.

### Context
The work in this issue will make CometBFT use only pebbleDB as database, thus deprecating [cometbft-db](https://github.com/cometbft/cometbft-db). 
As CometBFT will use the same [`DB` interface](https://github.com/cometbft/cometbft-db/blob/8bd4531b5122530f39a90d385cb0b0523dbe881e/types.go#L21) that cometbft-db uses, users can provide their own implementation of the interface that uses a different database under the hood. They will, therefore, have to change the [`config.DefaultDBProvider` function](https://github.com/cometbft/cometbft/blob/b8df798fd85c68e289c595df0b233f37df5db616/config/db.go#L30) to return their implementation instead of our default.

However, we will support and maintain only the pebbleDB implementation of the `DB` interface that CometBFT will use. 

### ToDo

- [x] Create `internal/storage` pkg to port the `DB` interface and pebbleDB's implementation of it that are currently in cometbft-db. 
  - merged with https://github.com/cometbft/cometbft/pull/4511
- [x] Add tests for pebble functionalities.
  - [x] `DB` inteface. Merged with https://github.com/cometbft/cometbft/pull/4515 
  - [x] `Batch` interface. Merged with https://github.com/cometbft/cometbft/pull/4522
  - [x] `Iterator` interface. Merged: #4530 
- [x] Import `PrefixDB` implementation. Merged: #4562  
- [x] Remove cometbft-db as a dependency. WIP: #4601 
- [x] Remove `DBBackend` field of `config.BaseConfig` and from config options in toml. https://github.com/cometbft/cometbft/pull/4915
- [x] Remove cometbft-db from cometbft build process. Merged with #4650

andynog:
We should also at a task to replace the BadgerDB with PebbleDB in the tutorial

faddat:
> The primary purpose of using a simple database like PebbleDB is to make testing and development straightforward. However, this simplicity should not restrict users from switching to another database if PebbleDB does not meet their specific needs. 

Sounds very good

> They should be able to develop and integrate their own database solution seamlessly.

....using examples from earlier implementations like goleveldb

alesforz:
> We should also at a task to replace the BadgerDB with PebbleDB in the tutorial

I think this is included in the "Documentation updated" task of #1039 🙂

PR #4511, opened 2024-11-18
feat(storage): Add package `internal/storage`

alesforz:
Partially addresses #4486.

### Context

### Changes
This PR adds package `storage` to the `internal` packages: 
- it exports the `DB`, `Batch`, and `Iterator` interfaces that database backends must implement to work with CometBFT. Previously, they were offered by cometbft-db.
- it exports the pebbledb implementation of the above interfaces.

Users can provide their own implementation of these interfaces using a different database under the hood. However, we will support and maintain only the pebble implementation that CometBFT will use.

**Note**: this PR ports the code currently provided by cometbft-db. It does not add new functionalities to either the interfaces or their implementation using pebble. What it does:
- improve docs
- clearer error messages
- minor refactoring of nested `if` statements for readability

---

#### PR checklist

- ~[] Tests written/updated~


melekes (approved):
👍

PR #4513, opened 2024-11-18
feat(e2e): Support sending the same tx to  multiple nodes

### LINKED ISSUE #4430: e2e: Support sending tx load to multiple nodes simultaneously
For [the DOG protocol](https://github.com/cometbft/cometbft/issues/3297) we need test scenarios that send the same transaction to the RPC endpoints of multiple nodes. This may occur, for example, when a user wants her transaction to be committed to the chain as fast as possible. The problem is that this could destabilise the gossip protocol. 

Currently the e2e's `runner load` command, when sending transactions, it picks nodes in the testnet, one by one, in a round-robin fashion. Note that the `load` command has a flag `--nodes` that is used to list the target nodes to send the transaction load.

One way to add support to e2e for sending transactions to multiple nodes simultaneously is to add to the `load` command a new flag `--send-all-nodes` (or a better name), or a new option `--strategy=...`, where the current strategy is `"round-robin"` and the new option would be `"all-target-nodes"`.

jmalicevic:
Closes #4430 

This PR adds a new flag `--num-nodes-per-tx N ` where N is the number of nodes to which the loader is sending the same transactions in an iteration. 

If omitted the transactions are sent in round robin. This was added to facilitate testing the efficiency of the DOG protocol.

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

test/e2e/runner/main.go
```diff
@@ -290,6 +300,9 @@ func NewCLI() *CLI {
 		"Comma-separated list of node names to send load to. Manifest option send_no_load will be ignored.")
 	loadCmd.PersistentFlags().BoolP("internal-ip", "i", false,
 		"Use nodes' internal IP addresses when sending transaction load. For running from inside a DO private network.")
+	loadCmd.PersistentFlags().IntP("duplicate-num-nodes", "", 0,
+		"Number of nodes that will receive the same transactions")
```
hvanz, line 304:
```suggestion
		"Number of nodes that will receive each generated transaction.")
```

test/e2e/runner/load.go
```diff
@@ -174,6 +188,43 @@ FOR_LOOP:
 	wg.Wait()
 }
 
+// loadProcess processes transactions by sending transactions received on the txCh
+// to the client.
+func loadProcessMultiple(ctx context.Context, txCh <-chan types.Tx, chSuccess chan<- struct{}, chFailed chan<- error, nodes []*e2e.Node, useInternalIP bool) {
```
hvanz, line 193: This new function could be easily be merged with `loadProcess`. That is, `loadProcess` is (almost) an instance of `loadProcessMultiple` when `nodes` has one element.

test/e2e/runner/main.go
```diff
@@ -290,6 +300,9 @@ func NewCLI() *CLI {
 		"Comma-separated list of node names to send load to. Manifest option send_no_load will be ignored.")
 	loadCmd.PersistentFlags().BoolP("internal-ip", "i", false,
 		"Use nodes' internal IP addresses when sending transaction load. For running from inside a DO private network.")
+	loadCmd.PersistentFlags().IntP("duplicate-num-nodes", "", 0,
```
hvanz, line 303: The name of the flag is not very clear to me but I don't have an alternative name to propose.

test/e2e/runner/load.go
```diff
@@ -174,6 +188,43 @@ FOR_LOOP:
...
+				logger.Info("non-fatal error creating node client", "error", err)
+				return
+			}
+		}
+	}
+	for tx := range txCh {
+		var wg sync.WaitGroup
+		wg.Add(len(nodes))
+		for _, n := range nodes {
+			go func(client *rpchttp.HTTP) {
```
hvanz, line 215: This will generate a lot of goroutines but we've seen that it works in practice.

hvanz (approved):

test/e2e/runner/main.go
hvanz: How about `num_copies`?  No need to change it if you don't like it.

cason (commented):
I have a number of concerns regarding the implementation of the `load` command and also regarding the proposed changes.

First, the assumption is that the current load generator sends transactions to nodes in a round-robin fashion (as stated in #4430) is likely false. In fact, we instantiate `load_tx_connections` per destination node routines, and all those routines compete for the same input put in a single `txCh` channels containing produced transactions. This means that the same routine, or routines targeting the same node can consume multiple transactions in a row. The round-robin fashion here is a wishful thinking.

The solution proposes the creation of a new load method that, instead of having a routine per connection per destination node, has a routine per connection receiving `N` destination nodes. I am not sure if the way we pick this set of nodes (instead of a single node) is really fair, to start with. If I understood correctly, the set of destination nodes are just the first `N` valid destination nodes in `testnet.Node`. So the new method send all transactions to the same set of nodes, right?

Finally, how I think this could be implemented? I am not sure. Probably we should have another channel containing a `Tx` and a set of nodes to which it was already submitted, then all threads consume this channel and if they are not part of the "destinations" of that transaction and we still want to send the same transaction to multiple destinations, it sends it again and push the updated entry to the same channel. This is not great as well.

Another option is to provide to the main node method a set of "secondary destinations", which can be empty. Once a transaction is sent to the "primary destination", it is also sent to the "secondary destinations", probably we do not need to this in parallel.

In summary, the design of the `load` tool is not great and we realize that when we want to produce any new behavior.

test/e2e/runner/load.go
```diff
@@ -37,6 +37,8 @@ func Load(ctx context.Context, testnet *e2e.Testnet, useInternalIP bool) error {
 
 	txCh := make(chan types.Tx)
```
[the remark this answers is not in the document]
jmalicevic: I agree, but probably not something we should change here. We could at least make it buffered for the number of nodes receiving the tx at the same time.

test/e2e/runner/load.go
jmalicevic: Resolved with 14d0555f5

test/e2e/runner/load.go
jmalicevic: I am not sure how to do it differently so that the txs are sent in parallel.. We could use BroadcastTxAsync or have one go routine per X nodes (in case we have this situation as with 200 nodes where we send 100 duplicates).

test/e2e/runner/main.go
jmalicevic: Renamed to `num-nodes-per-tx` . Although that is also not accurate. Implies that **each** tx gets duplicated, while that is not true.

hvanz:
I think this fixes it: https://github.com/cometbft/cometbft/pull/4561

.changelog/unreleased/features/4513-e2e-same-load.md
```diff
@@ -0,0 +1,2 @@
+- `[e2e]` Add `duplicate-num-nodes N` to send same transactions to N nodes.
```
hvanz, line 1:
```suggestion
- `[e2e]` Add option `--load_num_nodes_per_tx N` to send each transaction to N nodes.
```

PR #4515, opened 2024-11-19
feat(storage): Add unit tests for pebble implementation of `DB` interface.

alesforz:
Partially addresses #4486.

### Context

### Changes
This PR adds unit tests to the pebble implementation of the `DB` interface that CometBFT will use once we remove cometbft-db.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/internal/storage/pebbledb.go
+++ b/internal/storage/pebbledb.go
@@ -259,6 +259,7 @@ func (pDB *PebbleDB) Close() error {
 }
 
 // Print prints all the key/value pairs in the database for debugging purposes.
+//
 // It implements the [DB] interface for type PebbleDB.
 func (pDB *PebbleDB) Print() error {
 	itr, err := pDB.Iterator(nil, nil)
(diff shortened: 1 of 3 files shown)


internal/storage/pebbledb.go
```diff
@@ -19,7 +19,6 @@ var _ DB = (*PebbleDB)(nil)
 // NewPebbleDB returns a new PebbleDB instance using the default options.
 func NewPebbleDB(name, dir string) (*PebbleDB, error) {
 	opts := &pebble.Options{}
-	opts.EnsureDefaults()
```
melekes, line 22: not needed?

melekes (approved):
👍

internal/storage/pebbledb.go
alesforz: `NewPebbleDB` always calls `NewPebbleDBWithOpts` which already calls `opts.EnsureDefaults()` 🙂

Issue #4519, opened 2024-11-20
GitHub Action Slack notification workflow not working.


andynog:
Looks like [upgrading](https://github.com/cometbft/cometbft/pull/4505) to `slack-github-action@v2.0.0` [broke our Slack notifications](https://github.com/cometbft/cometbft/actions/runs/11936697259/job/33271279972). 

Need to ensure the notifications are working again on all workflows that had the slack GitHub Action bumped to `2.0.0`.

Tasks:
- [ ] Add workflow that can be triggered manually to test the notifications. This is helpful for testing the right code and template for the notifications. #4520 
- [ ] Fix upgraded workflows (#4505) with the right code and templates

PR #4520, opened 2024-11-20
fix: add a new manual slack notification workflow for testing

andynog:
Partially close: #4519 

#### PR checklist

- [ ] Tests written/updated


andynog:
@melekes the Slack templates were completely overhauled many breaking changes to update so to avoid many refactoring it is better to test with this, and then when this works fine fix the others, also the other workflows it is hard to trigger a notification unless something fails or gets released

melekes (approved):

PR #4522, opened 2024-11-21
feat(storage): Add unit tests for pebble implementation of `Batch` interface.

alesforz:
Partially addresses #4486.

### Context

### Changes
This PR adds unit tests to the pebble implementation of the `Batch` interface that CometBFT will use once we remove cometbft-db.

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

PR #4531, opened 2024-11-22
perf(p2p): don't remarshal within broadcast

wesl-ee:
upstream from osmosis/cometbft https://github.com/osmosis-labs/cometbft/pull/141

---

#### PR checklist

- [ ] Tests written/updated
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments

PR #4546, opened 2024-11-25
perf(p2p): don't remarshal within broadcast

wesl-ee:
upstream from osmosis/cometbft https://github.com/osmosis-labs/cometbft/pull/141

moved from #4531 

---

#### PR checklist

- [ ] Tests written/updated


p2p/switch.go
```diff
@@ -286,10 +287,22 @@ func (sw *Switch) Broadcast(e Envelope) {
...
-		}(p)
+	sw.Logger.Debug("TryBroadcast", "channel", e.ChannelID)
+
+	marshalMsg := e.Message
+	if wrapper, ok := e.Message.(types.Wrapper); ok {
+		marshalMsg = wrapper.Wrap()
+	}
+	marshalledMsg, err := proto.Marshal(marshalMsg)
+	if err != nil {
+		panic(fmt.Sprintf("failed to marshal proto msg: err %v", err))
```
melekes, line 298:
```suggestion
		panic(fmt.Sprintf("failed to marshal proto msg %v: %v", marshalMsg, err))
```

melekes (approved):
👍

p2p/switch.go
```diff
@@ -286,10 +287,22 @@ func (sw *Switch) Broadcast(e Envelope) {
...
-		}(p)
+	sw.Logger.Debug("TryBroadcast", "channel", e.ChannelID)
+
+	marshalMsg := e.Message
+	if wrapper, ok := e.Message.(types.Wrapper); ok {
+		marshalMsg = wrapper.Wrap()
+	}
+	marshalledMsg, err := proto.Marshal(marshalMsg)
+	if err != nil {
+		panic(fmt.Sprintf("failed to marshal proto msg %v: %v", marshalMsg, err))
```
cason, line 298: Shouldn't panic.

p2p/switch.go
```diff
@@ -286,10 +287,22 @@ func (sw *Switch) Broadcast(e Envelope) {
...
+	}
+	marshalledEnvelope := MarshalledEnvelope{
```
cason, line 300: In my honest opinion, the best solution here is to include in the existing `p2p.Envelope` a private field with the marshalled version of the message. Everything would be simpler.

p2p/peer.go
```diff
@@ -46,6 +46,7 @@ type Peer interface {
 	HasChannel(chID byte) bool // Does the peer implement this channel?
 	Send(e Envelope) bool      // Send a message to the peer, blocking version
 	TrySend(e Envelope) bool   // Send a message to the peer, non-blocking version
+	TrySendMarshalled(e MarshalledEnvelope) bool
```
cason, line 49: Is anyone outside the p2p package using this method?

If not, there is no point in adding it to the public interface.

p2p/switch.go
cason: Notice that `func (p *peer) send` does not panic.

cason (commented):
Honestly, I don't like this solution.

Sending a message includes 3 steps:

1. Marshaling the `proto.Message` into bytes
2. Calling the underlying connection method passing the bytes
3. Registering metrics and returning the output (`true` or `false`)

The `Switch.Broadcast` method is a helper that takes the set of connected peers and send the same messages to all of them. It does not work well because step 1. is unnecessarily repeated for every peer. So what we want is to avoid 1. to be performed multiple times for the same message.

As suggested in my comments, probably the best way of achieving so is to add a private field to `p2p.Envelope` that stores the marshalled version of the `Message` field. In this way, the existing sending methods can save step 1. without huge changes in interfaces, implementations, etc.

wesl-ee:
since this is not an upstream it seems we want, closing this

PR #4547, opened 2024-11-26
refactor(mempool): give reason why tx is invalid

julienrbrt:
Gives back more info to the client when a tx is invalid. (due to changes in https://github.com/cometbft/cometbft/pull/4040)
Right now it is only debug logged on the node, which is not so useful for the client to understand what's up.

Additionally, I am not even sure if returning an error here instead of the response is great.
This is a client breaking changes. Before the tx would go through but the ABCI response would be non-zero (15 - no signature supplied).

```yaml
code: 15
codespace: sdk
data: ""
events: []
gas_used: "0"
gas_wanted: "0"
height: "0"
info: ""
logs: []
raw_log: 'empty signatures: no signatures supplied'
timestamp: ""
tx: null
txhash: 7F704013E8B09D4FF382EE8B67D37B5B3CD777E610A765437E07F309EACA5738
```

Now it fails earlier with with an rpc error:

```bash
error in json rpc client, with http response metadata: (Status: 200 OK, Protocol HTTP/1.1). RPC error -32603 - Internal error: broadcast error on transaction validation: tx is invalid
```

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -422,7 +422,7 @@ func (mem *CListMempool) handleCheckTxResponse(tx types.Tx, sender nodekey.ID) f
 				return postCheckErr
 			}
 
-			return fmt.Errorf("%s: %w", res.Log, ErrInvalidTx)
+			return fmt.Errorf("%w: %s", ErrInvalidTx, res.Log)
 		}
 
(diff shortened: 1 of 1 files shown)


julienrbrt:
Closing, as probably shouldn't be returning an error actually. I will mention this on Slack.

PR #4550, opened 2024-11-26
fix(mempool): Extend `ErrInvalidTx` with `CheckTxResponse` info

hvanz:
When calling the RPC endpoints `broadcast_tx_*`, the mempool returns a `reqRes` object with a potential error. Before #4040, `reqRes` didn't have an error. An error on the ABCI CheckTx call would be returned encoded in the fields Code, Log, etc, of `ResultBroadcastTx`. Any other internal error of the mempool would be discarded.

The problem is that now when the transaction is invalid, `broadcast_tx_*` return `ErrTxBroadcast{Source: ErrCheckTxFailed, ErrReason: err}` where `err`  is simply `ErrInvalidTx`, without any other information. This PR adds all the fields in `ResultBroadcastTx` (Code, Data, Log, Codespace, and Hash) to `ErrInvalidTx`, so that this information is available gain to the `broadcast_tx_*` caller.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/.changelog/unreleased/breaking-changes/4550-mempool-err-invalid-tx.md
+++ b/.changelog/unreleased/breaking-changes/4550-mempool-err-invalid-tx.md
@@ -1,0 +1,2 @@
+- `[mempool]` Extend `ErrInvalidTx` with new fields taken from `CheckTxResponse`
+  ([\#4550](https://github.com/cometbft/cometbft/pull/4550)).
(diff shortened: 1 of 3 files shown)


julienrbrt (approved):
Great, thanks!

cason (commented):
There is also an `Info` field in the `CheckTxResponse`, no idea what is used for.

I wonder why we don't just return the full `res`, ehe.

cason:
Finally, regarding the "update docs/spec" in the check list, we might inform which information is returned by the caller (tx submitter) under https://docs.cometbft.com/v1.0/spec/abci/abci++_methods#checktx.

melekes (approved):
👍

Issue #4452, opened 2024-11-07
logs: on-demand or configuration colorized outputs


cason:
With the introduction of slog (https://github.com/cometbft/cometbft/pull/4374), logs have colors indicating the logging level. This is a nice feature, although already present in the legacy logger implementation (used up to `v1.x`).

But at the same time, when the output of CometBFT is redirected to a file, the special terminal tags formatting the text are also written to the output file. This renders parsing and navigating the log files a bad operator experience.

I think we have three possible solutions here:

1. Disable colors at all in Comet's logs
2. Add a configuration switch to disable coloring (to be use, for instance, in test units and e2e infra)
3. Have a log setup that is smart enough to detect whether the output is a terminal, with color support, or not

The legacy solution adopts alternative 3., which is also my preferred one.

cason:
#4475 is an implementation of solution 2.

This is not my preferred way, but could be a decent workaround.

Probably combining 2. with 3. would be the ideal. This would be similar as the behavior of the `ls` command, for instance, with options `auto`, `always`, `never`.

PR #4551, opened 2024-11-26
fix(e2e): application logger does not use colors

cason:
Follow-up of #4475.

The e2e application by default produces a colorized output. This should not be the case for the same reasons explained in https://github.com/cometbft/cometbft/issues/4452.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/test/e2e/app/app.go
+++ b/test/e2e/app/app.go
@@ -191,7 +191,7 @@ func NewApplication(cfg *Config) (*Application, error) {
 	for keyType := range cmttypes.ABCIPubKeyTypesToNames {
 		allKeyTypes = append(allKeyTypes, keyType)
 	}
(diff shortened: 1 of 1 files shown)


test/e2e/app/app.go
```diff
@@ -156,6 +156,9 @@ type Config struct {
 	// * An existing validator will be chosen, and its power will alternate 0 and 1
 	// * `ConsensusParams` will be flipping on and off key types not set at genesis
 	ConstantFlip bool `toml:"constant_flip"`
+
+	// Colored log output.
```
cason, line 160: This is currently a noop, since we don't load the configuration from any file.

melekes (approved):
👍

PR #4556, opened 2024-11-29
chore(p2p)!: reintroduce `p2p.ID` type and internalize `nodeinfo`

### LINKED ISSUE #4549: p2p: restore the type `p2p.ID`
This change was introduced by https://github.com/cometbft/cometbft/pull/4148, where `p2p.ID` has become `p2p/nodekey.ID`.

Sorry, but this does not make sense.

melekes:
Closes #4549

It can be reviewed commit by commit.


p2p/test_util.go
```diff
@@ -319,13 +319,17 @@ func (mockNodeInfo) Handshake(net.Conn, time.Duration) (ni.NodeInfo, error) { re
 
 func testNodeInfo(id nodekey.ID, name string) ni.Default {
```
cason, line 320: Still `nodekey` here, don't think it changes anything.

cason (approved):
Thank very much for that.

p2p/test_util.go
melekes: But this is `p2p` package...

PR #4557, opened 2024-11-29
chore(p2p)!: internalize `nodekey`

melekes:
Refs #4549


mempool/mocks/mempool.go
```diff
@@ -8,7 +8,7 @@ import (
 
 	mock "github.com/stretchr/testify/mock"
 
-	nodekey "github.com/cometbft/cometbft/p2p/nodekey"
+	nodekey "github.com/cometbft/cometbft/p2p/internal/nodekey"
```
melekes, line 11: Maybe you know why `mockery` does not use `p2p.ID` (an alias for `nodekey.ID`; `type ID = nodekey.ID`) here?

In the interface, I have `p2p.ID` (see mempool.go file), so I would expect `mockery` to use that instead of resolving the type...

hvanz (approved):
👍

PR #4561, opened 2024-11-29
feat(e2e): for loading with num-nodes-per-tx, use one channel per node

hvanz:
Solves #4430

by fixing #4513

This solution creates multiple transaction channels, one per node in the testnet. On each transaction generated, it will send the transaction to  `LoadNumNodesPerTx` random channels.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/test/e2e/runner/load.go
+++ b/test/e2e/runner/load.go
@@ -36,13 +36,9 @@ func Load(ctx context.Context, testnet *e2e.Testnet, useInternalIP bool) error {
 	started := time.Now()
 	u := [16]byte(uuid.New()) // generate run ID on startup
 
-	txChs := make([](chan types.Tx), len(testnet.Nodes))
-	for i := range testnet.Nodes {
-		txChs[i] = make(chan types.Tx)
-	}
-	go loadGenerate(ctx, txChs, testnet, u[:])
-
-	for i, n := range testnet.Nodes {
+	// Nodes that will receive load.
+	targetNodes := make([]*e2e.Node, 0)
+	for _, n := range testnet.Nodes {
 		if len(testnet.LoadTargetNodes) == 0 {
 			if n.SendNoLoad {
 				continue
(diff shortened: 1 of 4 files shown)


test/e2e/runner/load.go
```diff
@@ -172,10 +160,21 @@ func createTxBatch(ctx context.Context, txCh chan<- types.Tx, testnet *e2e.Testn
...
 				}
 
-				select {
-				case txCh <- tx:
-				case <-ctx.Done():
-					return
+				var nodeIndices []int
+				if testnet.LoadNumNodesPerTx <= 1 {
+					// Pick one random node to send the transaction.
+					nodeIndices = []int{cmtrand.Intn(len(testnet.Nodes))}
```
jmalicevic, line 166: Given the random generator, are we sure it is not going to pick the same node every time? Or at least very often? Should we in this case not use an increasing list of idxs perhaps?

test/e2e/runner/load.go
```diff
@@ -172,10 +160,21 @@ func createTxBatch(ctx context.Context, txCh chan<- types.Tx, testnet *e2e.Testn
...
+				} else {
+					// Pick LoadNumNodesPerTx random nodes (channel indices) to
+					// send the transaction.
+					nodeIndices = cmtrand.Perm(len(testnet.Nodes))[:testnet.LoadNumNodesPerTx]
+				}
+				for _, i := range nodeIndices {
+					select {
+					case txChs[i] <- tx:
```
jmalicevic, line 174: How do we know that we have not picked an index of a node we are skipping ,thus we will not use this channel in the `Load` function? What we can do is keep a list of valid indexes in `Load` passing it to `LoadGenerate` that will forward it here.

test/e2e/runner/load.go
hvanz: Isn't it a very similar behaviour to what we have now? Currently we have many readers (one per node and per connection) on the same channel waiting for a transaction. 

From the golang [spec](https://go.dev/ref/spec#Select_statements):
> 2. If one or more of the communications can proceed, a single one that can proceed is chosen via a uniform pseudo-random selection. Otherwise, if there is a default case, that case is chosen. If there is no default case, the "select" statement blocks until at least one of the communications can proceed.

jmalicevic: I guess you are right. The only other concern is what I wrote below, if one of those nodes is supposed to be skipped  (because it has `send_no_load` to true.

test/e2e/runner/load.go
hvanz: Good catch. I fixed it.

test/e2e/runner/load.go
```diff
@@ -172,10 +167,21 @@ func createTxBatch(ctx context.Context, txCh chan<- types.Tx, testnet *e2e.Testn
...
-				case <-ctx.Done():
-					return
+				var nodeIndices []int
+				if testnet.LoadNumNodesPerTx <= 1 {
+					// Pick one random node to send the transaction.
+					nodeIndices = []int{cmtrand.Intn(len(targetNodes))}
+				} else {
+					// Pick LoadNumNodesPerTx random nodes (channel indices) to
+					// send the transaction.
+					nodeIndices = cmtrand.Perm(len(targetNodes))[:testnet.LoadNumNodesPerTx]
```
hvanz, line 177: TODO: If LoadNumNodesPerTx is too small compared to len(targetNodes), it may be worth to pick LoadNumNodesPerTx random elements one by one from targetNodes instead of doing a permutation.

PR #4562, opened 2024-11-29
feat(storage): Add `PrefixDB` to pkg `internal/storage`

alesforz:
Partially addresses #4486.

### Context

### Changes
This PR adds type `PrefixDB` to `internal/storage` from cometbft-db. 
`PrefixDB` is a database wrapper that implements  the`DB`, `Batch`, and `Iterator` interfaces.

**Note**: this PR makes some changes to the implementation of `PrefixDB`. Namely:
- `PrefixDB` no longer uses a mutex, because we think that it is sufficient to rely on the wrapped database's concurrency control mechanism. That is, concurrent calls to `PrefixDB`'s methods will already "lock" on the calls to the wrapped DB methods.
- `NewPrefixDB` now returns an error if the given prefix is empty. We think allowing the creation of a `PrefixDB` with no prefix was an oversight in the design.
- `prefixDBBatch` methods now want a pointer receiver. 
- `PrefixDB.Compact` now prepends the prefix to `start` and `end` keys defining the iterator's range. The previous implementation did not do so, and this was likely an oversight. We believed this behavior contradicted the definition of a `PrefixDB` as a logical database that prepends the prefix to all keys before operating on the underlying database.
 
---

#### PR checklist

- [x] Tests written/updated


internal/storage/prefixdb.go
```diff
@@ -0,0 +1,632 @@
...
+		// "correct" key of the form prefix+key, because
+		// prefix < prefix+key, e.g., "foo" < "fooa".
+		// Therefore, we only need to skip it to position the iterator at the first
+		// "correct" key.
+		source.Next()
+	}
+
+	if !source.Valid() || !bytes.HasPrefix(source.Key(), prefix) {
+		return itInvalid, nil
+	}
```
alesforz, line 467: Why do we create the `itInvalid` object? Since it's an invalid iterator, any method call on it will cause a panic, thus making it useless. Why not return an error here instead?

internal/storage/prefixdb.go
```diff
@@ -0,0 +1,632 @@
...
+
+// compile-time check: does *prefixDBIterator satisfy the Iterator interface?
+var _ Iterator = (*prefixDBIterator)(nil)
+
+// newPrefixDBIterator returns a new prefixDBIterator to iterate over a range of
+// database key/value pairs that begin with the given prefix.
+func newPrefixDBIterator(
+	prefix, start, end []byte,
+	source Iterator,
+) (*prefixDBIterator, error) { //nolint:unparam
```
alesforz, line 445: This function does not return an `error`. Why do we have it in its signature?

internal/storage/prefixdb.go
```diff
@@ -0,0 +1,632 @@
...
+// or to call Close concurrently with any other DB method. It is not valid
+// to call any of a DB's methods after the DB has been closed.
+//
+// It implements the [DB] interface for type PrefixDB.
+func (pDB *PrefixDB) Close() error {
+	pDB.mu.Lock()
+	defer pDB.mu.Unlock()
+
+	return pDB.db.Close()
+}
```
alesforz, line 272: The docs of `PrefixDB` in cometbft-db [say](https://github.com/cometbft/cometbft-db/blob/c588fb01e0cfdaba5d14df06d6b623d9cecd096d/prefixdb.go#L17) that it "lets you namespace multiple DBs within a single DB."

However, doesn't this implementation of the `Close` method break this promise? It closes the underlying database. If we have, say, 2 `PrefixDB` objects wrapping the same `DB`, then I think that calling `Close` on of the `PrefixDB`s will close the underlying `DB` for both.

internal/storage/prefixdb.go
```diff
@@ -0,0 +1,632 @@
...
+
+// Print prints all the key/value pairs in the database for debugging purposes.
+//
+// It implements the [DB] interface for type PrefixDB.
+func (pDB *PrefixDB) Print() error {
```
alesforz, line 277: Shouldn't we change this to accept an `io.Writer` for greater flexibility of where to output the prints? This would be a breaking change though.

internal/storage/prefixdb.go
```diff
@@ -0,0 +1,632 @@
...
+// be retrieved via the [Error] method.
+//
+// It implements the [Iterator] interface.
+type prefixDBIterator struct {
+	source Iterator
+	prefix []byte
+	start  []byte
+	end    []byte
+	valid  bool
+	err    error
```
alesforz, line 434: The `err` field is set only once in the `Valid()` method and is never returned elsewhere. All methods of `prefixDBIterator`, except `Domain()` and `Close()`,
will panic if `assertIsValid()` fails. While the `Error()` method does return `err`, is
it actually used if we panic most of the time?

What I'm asking is: Does it make sense to have this `err` field? Why doesn't `Valid()` also panic in that single case? It sets the iterator to invalid anyway, which will cause a panic on the next method call.

internal/storage/prefixdb.go
```diff
@@ -0,0 +1,632 @@
...
+	if bytes.Equal(it.source.Key(), it.prefix) {
+		// Empty keys are not allowed, so if a key exists in the database that
+		// exactly matches the prefix we need to skip it.
+		// The key is going to be lexicograpically smaller than the first "correct"
+		// key of the form prefix+key, because
+		// prefix < prefix+key, e.g., "foo" < "fooa".
+		// Therefore, we only need to skip it to position the iterator at the first
+		// "correct" key.
+		it.Next()
+	}
```
alesforz, line 551: Is the purpose of these `if` statements to double-check that the current key of the wrapped `Iterator` in `prefixDBIterator` is correct?

We already call `Iterator.Valid` on the wrapped iterator to check the validity of the key. Assuming a `prefixDBIterator` is correctly initialized with the key range [`prefix|start_key` : `prefix|end_key`], in what scenarios are these additional checks needed?

For example, consider our implementation of the Pebble iterator [`[pebbleDBIterator](https://github.com/cometbft/cometbft/blob/3be84b67b126fa3fa2ff911acda9029d6c3586f4/internal/storage/pebbledb.go#L479)`](https://github.com/cometbft/cometbft/blob/3be84b67b126fa3fa2ff911acda9029d6c3586f4/internal/storage/pebbledb.go#L479). Its [`Valid()` method](https://github.com/cometbft/cometbft/blob/3be84b67b126fa3fa2ff911acda9029d6c3586f4/internal/storage/pebbledb.go#L543) ensures that the iterator's current key is between `start` and `end`. Now, suppose we create a `prefixDBIterator` wrapping a `pebbleDBIterator` with prefix `foo`, start key `a`, and end key `f`. The key range of this iterator will be [`fooa` : `fooe`] (since the end is exclusive, `foof` won't be included). If I understand correctly, a call to `pebbleDBIterator.Valid()` should be sufficient to verify that the current key is valid, as any key outside [`fooa` : `fooe`] will be considered invalid.

Therefore, it seems that checking the prefix again in the `prefixDBIterator` methods might be unnecessary.

Returning to my initial question: Are we performing these additional checks to guard against third-party `Iterator` implementations that might not strictly enforce key validity and could return keys we should consider invalid? If not, what am I missing?

internal/storage/prefixdb.go
```diff
@@ -0,0 +1,632 @@
...
+		key            = it.source.Key()
+		keyTooShort    = len(key) < len(it.prefix)
+		prefixMismatch = !bytes.Equal(key[:len(it.prefix)], it.prefix)
+	)
+	if keyTooShort || prefixMismatch {
+		const format = "received invalid key from backend: %x (expected prefix %x)"
+		it.err = fmt.Errorf(format, key, it.prefix)
+
+		return false
+	}
```
alesforz, line 574: See my comment about the `Next` method above. The same considerations apply here.

internal/storage/prefixdb.go
```diff
@@ -0,0 +1,632 @@
...
+// It is not safe to close a DB until all outstanding iterators are closed
+// or to call Close concurrently with any other DB method. It is not valid
+// to call any of a DB's methods after the DB has been closed.
+//
+// It implements the [DB] interface for type PrefixDB.
+func (pDB *PrefixDB) Close() error {
+	pDB.mu.Lock()
+	defer pDB.mu.Unlock()
+
+	return pDB.db.Close()
```
alesforz, line 271: The docs of `PrefixDB` in cometbft-db [say](https://github.com/cometbft/cometbft-db/blob/c588fb01e0cfdaba5d14df06d6b623d9cecd096d/prefixdb.go#L17) that it "lets you namespace multiple DBs within a single DB."

However, doesn't this implementation of the `Close` method break this promise? It closes the underlying database. If we have, say, 2 `PrefixDB` objects wrapping the same `DB`, then I think that calling `Close` on of the `PrefixDB`s will close the underlying `DB` for both.

internal/storage/prefixdb.go
```diff
@@ -0,0 +1,632 @@
...
+	var (
+		srcItInvalid  = !it.source.Valid()
+		missingPrefix = !bytes.HasPrefix(it.source.Key(), it.prefix)
+	)
+	if srcItInvalid || missingPrefix {
+		it.valid = false
+		return
+	}
+
+	if bytes.Equal(it.source.Key(), it.prefix) {
```
alesforz, line 542: Is the purpose of these `if` statements to double-check that the current key of the wrapped `Iterator` in `prefixDBIterator` is correct?

We already call `Iterator.Valid` on the wrapped iterator to check the validity of the key. Assuming a `prefixDBIterator` is correctly initialized with the key range [`prefix|start_key` : `prefix|end_key`], in what scenarios are these additional checks needed?

For example, consider our implementation of the Pebble iterator [`pebbleDBIterator`](https://github.com/cometbft/cometbft/blob/3be84b67b126fa3fa2ff911acda9029d6c3586f4/internal/storage/pebbledb.go#L479). Its [`Valid()` method](https://github.com/cometbft/cometbft/blob/3be84b67b126fa3fa2ff911acda9029d6c3586f4/internal/storage/pebbledb.go#L543) ensures that the iterator's current key is between `start` and `end`. Now, suppose we create a `prefixDBIterator` wrapping a `pebbleDBIterator` with prefix `foo`, start key `a`, and end key `f`. The key range of this iterator will be [`fooa` : `fooe`] (since the end is exclusive, `foof` won't be included). If I understand correctly, a call to `pebbleDBIterator.Valid()` should be sufficient to verify that the current key is valid, as any key outside [`fooa` : `fooe`] will be considered invalid.

Therefore, it seems that checking the prefix again in the `prefixDBIterator` methods might be unnecessary.

Returning to my initial question: Are we performing these additional checks to guard against third-party `Iterator` implementations that might not strictly enforce key validity and could return keys we should consider invalid? If not, what am I missing?

internal/storage/prefixdb.go
```diff
@@ -0,0 +1,632 @@
...
+		it.valid = false
+		return false
+	}
+
+	var (
+		key            = it.source.Key()
+		keyTooShort    = len(key) < len(it.prefix)
+		prefixMismatch = !bytes.Equal(key[:len(it.prefix)], it.prefix)
+	)
+	if keyTooShort || prefixMismatch {
```
alesforz, line 569: See my comment about the `Next` method above. The same considerations apply here.

internal/storage/prefixdb.go
```diff
@@ -0,0 +1,632 @@
...
+
+	return newPrefixDBBatch(pDB.prefix, pDB.db.NewBatch())
+}
+
+// Compact compacts the specified range of keys in the database.
+//
+// It implements the [DB] interface for type PrefixDB.
+func (pDB *PrefixDB) Compact(start, end []byte) error {
+	if err := pDB.db.Compact(start, end); err != nil {
+		return fmt.Errorf("prefixed DB namespace compaction: %w", err)
```
alesforz, line 256: `Compact` does not prepend the prefix to `start` and `end` before calling `Compact` on the underlying database. Is this intended or an oversight?

Given the description of `PrefixDB`, I would expect that **all** read and write operations prepend the prefix to the keys. While this means it would only compact keys with the same prefix, which might not be particularly useful, I think not prepending the prefix goes against the `PrefixDB` type documentation.

internal/storage/prefixdb.go
```diff
@@ -0,0 +1,632 @@
...
+}
+
+// compile-time check: does *PrefixDB satisfy the DB interface?
+var _ DB = (*PrefixDB)(nil)
+
+// NewPrefixDB returns a new PrefixDB wrapping the given database and scoping it to
+// the given prefix.
+func NewPrefixDB(db DB, prefix []byte) *PrefixDB {
+	return &PrefixDB{
+		prefix: prefix,
```
alesforz, line 45: Should we check whether the given prefix is empty, or is it a valid case? I don't think there is much point to a `PrefixDB` without a prefix 😅

alesforz:
There might be a subtle bug in the validity check of `prefixDBIterator`. The construct we suggest to use the iterator:
```go
for ; itr.Valid(); itr.Next() {
	key := itr.Key()
	value := itr.Value()
        // do something...
	}
```
panics unexpectedly. 
I'm investigating, but I think it's because when the iterator reaches the last key, [here](https://github.com/cometbft/cometbft/blob/0a8476ad67f7d1ee053c0c6235a5995a7bb93f7d/internal/storage/prefixdb.go#L530) we check the iterator's validity, and it returns `true` (note that this function calls the wrapped iterator's `Valid` method under the hood). Then, [on the next line](https://github.com/cometbft/cometbft/blob/0a8476ad67f7d1ee053c0c6235a5995a7bb93f7d/internal/storage/prefixdb.go#L531), we call `Next` on the wrapped iterator. `Next`, in turn, calls `Valid` again, but this time it returns `false` and sets the wrapped iterator's `isInvalid` field to `true`. Now, when we call `Key` on what we believe is a valid iterator a [few lines below](https://github.com/cometbft/cometbft/blob/0a8476ad67f7d1ee053c0c6235a5995a7bb93f7d/internal/storage/prefixdb.go#L535), it panics because the iterator is now invalid. Note that all of this happens within the _wrapped_ iterator! The `prefixDBIterator` is unaware that the wrapped iterator has become invalid at this point.

Therefore, I think we need to make sure that `prefixDBIterator` marks itself as invalid as soon as the wrapped iterator becomes invalid. Otherwise, as in this bug, `prefixDBIterator` will continue iterating without realizing that the wrapped iterator is now invalid, leading to a panic.

In general, I think we should revisit the implementation of `prefixDBIterator` because its methods repeatedly call the `Valid` method of the wrapped iterator (I counted _four times_ for every key the `prefixDBIterator` processes), which I think is unnecessary.

alesforz:
> There might be a subtle bug in the validity check of `prefixDBIterator`. The construct we suggest to use the iterator panics unexpectedly...

Found and solved the problem. It was caused by a refactoring of the `if` statements. 
Still, I think this point:
> In general, I think we should revisit the implementation of `prefixDBIterator` because its methods repeatedly call the `Valid` method of the wrapped iterator (_four times_ for every call to `Next`, i.e., for every key the `prefixDBIterator` processes), which I think is unnecessary.

is something we should consider.

internal/storage/pebbledb.go
```diff
@@ -14,6 +14,7 @@ type PebbleDB struct {
 	db *pebble.DB
 }
 
+// compile-time check: does *PebbleDB satisfy the DB interface?
```
melekes, line 17: Don't see a strict need for this comment, as most devs are aware of such constructs

internal/storage/prefixdb.go
```diff
@@ -0,0 +1,632 @@
...
+//	err := db.Set([]byte("key"), []byte("value"))
+//	if err != nil {
+//		// handle error
+//	}
+//
+// In this example, the key "key" will be stored in 'baseDB' with the actual key
+// being "namespace:key".
+type PrefixDB struct {
+	mu     sync.Mutex
+	db     DB
```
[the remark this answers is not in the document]
melekes: agree. let's drop it

internal/storage/prefixdb.go
melekes: We definitely should

internal/storage/prefixdb.go
```diff
@@ -0,0 +1,650 @@
...
+}
+
+// Get fetches the value of the given key, or nil if it does not exist.
+// It is safe to modify the contents of key and of the returned slice after Get
+// returns.
+//
+// It implements the [DB] interface for type PrefixDB.
+func (pDB *PrefixDB) Get(key []byte) ([]byte, error) {
+	if len(key) == 0 {
+		return nil, errKeyEmpty
```
melekes, line 57: We should export this error imho

internal/storage/prefixdb.go
```diff
@@ -0,0 +1,650 @@
...
+// cached in memory and synced to disk later during a background flush or
+// compaction. Use [SetSync] to flush the write to disk immediately.
+//
+// It implements the [DB] interface for type PrefixDB.
+func (pDB *PrefixDB) Set(key []byte, value []byte) error {
+	if len(key) == 0 {
+		return errKeyEmpty
+	}
+	if value == nil {
+		return errValueNil
```
melekes, line 103: same

internal/storage/prefixdb.go
melekes: Good catch! This was probably an oversight. @jmalicevic would know more

internal/storage/prefixdb.go
melekes: Agree. It should probably be a noop.

internal/storage/prefixdb.go
melekes: I think the reason we're panicking is to enforce strict rules around `Iterator` usage.

Like you should always check if `Iterator` is valid before proceeding.

```go
...

it.Next()

if !it.Valid() {
  return it.Error()
}
```

internal/storage/prefixdb.go
melekes: I think we should return an error

internal/storage/prefixdb.go
melekes: Agree that it's not needed.

melekes (commented):
great work @alesforz 👍

internal/storage/prefixdb.go
jmalicevic: Oh yes, definitely an oversight. Although the way we use it now it did not cause problems because we pass `nil,nil` as the argument to indicate we want everything compacted. Although tbh I am not sure how this is handled then, if the keys are `nil,nil` we should most likely ignore the prefix. This will just try to compact all the files, it does not delete or alter them in any way. Albeit it might slow things down.

internal/storage/prefixdb.go
jmalicevic: We have to have an option to close the DB but we need to make it clear for users what will happen. In any case I am not really for this idea of wrapping DBs , not sure what the benefit here is. But that is a discussion for another time.

internal/storage/prefixdb.go
jmalicevic: Not at the moment but note it down as potential improvmenet.

internal/storage/prefixdb.go
```diff
@@ -0,0 +1,650 @@
...
+	}
+	pDB.mu.Lock()
+	defer pDB.mu.Unlock()
+
+	prefixedKey := prependPrefix(pDB.prefix, key)
+	value, err := pDB.db.Get(prefixedKey)
+	if err != nil {
+		return nil, fmt.Errorf("prefixed DB namespace get: %w", err)
+	}
+	return value, nil
```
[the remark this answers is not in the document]
alesforz: We are already ignoring that case. If pebble returns a `pebble.ErrNotFound` error, we ignore it and return an empty slice. CometBFT's code always checks if a `len(returnedSlice) == 0` to see if the key was found.

melekes (approved):

internal/storage/prefixdb.go
alesforz: After some thinking, we have decided to leave these checks in place. It is very likely that the reason they are there is to guard against third-party `Iterator` implementations that might not strictly enforce key validity and could return keys we should consider invalid. Changing this requires some in-depth testing to make sure we aren't screwing up validation checks on the keys. We, unfortunately, don't have enough time to do it at the moment.

alesforz:
> Are there any particular reasons why [for `prefixDBBatch`](https://github.com/cometbft/cometbft/blob/84f99f352549b65b0e6a0ae416672228a5792ae8/internal/storage/prefixdb.go#L338) we use a value receiver instead of a pointer, as we do for `pebbleDBBatch`?

Apparently an oversight, changed the receiver to a pointer.

PR #4563, opened 2024-11-29
fix(test): TestBlockPoolMaliciousNode shutdown threads at exit

greg-szabo:
This is a drive-by fix of a test that doesn't shut its threads down until the whole `go test` execution finishes. I think we have a bunch of these, but I came across this one during an unrelated troubleshooting.

Is it worth fixing this? It's not really causing any issues, it's just sloppy coding.

The only way to see any difference is to run the `go test` until it reaches its time limit and panics. In that case, the trace will contain references to the threads.

For example:
```
go test github.com/cometbft/cometbft/blocksync -v -run TestBlockPoolMaliciousNode -count 100 -failfast -race -timeout 30s
```

After 30 seconds the test didn't run 100 times yet, hence `go test` panics. Because the test has been run multiple times already, multiple sets of threads will be reported in the panic. With the fix, only one set is reported.

### DIFF
--- a/blocksync/pool_test.go
+++ b/blocksync/pool_test.go
@@ -1,6 +1,7 @@
 package blocksync
 
 import (
+	"context"
 	"fmt"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


blocksync/pool_test.go
```diff
@@ -314,38 +315,52 @@ func TestBlockPoolMaliciousNode(t *testing.T) {
...
 		for {
-			time.Sleep(1 * time.Second) // Speed of new block creation
-			for _, peer := range peers {
-				peer.height += 1                                   // Network height increases on all peers
-				pool.SetPeerRange(peer.id, peer.base, peer.height) // Tell the pool that a new height is available
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				time.Sleep(1 * time.Second) // Speed of new block creation
```
melekes, line 330: Can we set up a `time.Ticker` here?

blocksync/pool_test.go
```diff
@@ -314,38 +315,52 @@ func TestBlockPoolMaliciousNode(t *testing.T) {
...
-			}
-			first, second, _ := pool.PeekTwoBlocks()
-			if first != nil && second != nil {
-				if second.LastCommit == nil {
-					// Second block is fake
-					pool.RemovePeerAndRedoAllPeerRequests(second.Height)
-				} else {
-					pool.PopRequest()
+			default:
+				time.Sleep(500 * time.Millisecond) // Speed of block verification
```
melekes, line 348: Same question. Let's use a ticker here too.

melekes (commented):
Thanks @greg-szabo ❤️

melekes:
Any reason why the target branch is not `main`?

PR #4565, opened 2024-11-29
feat(e2e): Add perturbation interval to manifest and as command option

### LINKED ISSUE #4481: e2e: add flag and/or manifest option to control the time interval between each perturbation
Currently, when calling the `perturb` command, `runner` will read all perturbations for all nodes declared in the manifest and apply them one by one, with a fixed interval of 3 seconds.

For testing DOG, we need to perturb all nodes during the whole execution of a test instance. For example, while all the nodes in the testnet are running, and transaction load is being injected, we want to run in parallel `runner perturb --interval 30` to trigger one perturbation to one of the nodes every 30 seconds.

hvanz:
Closes #4481 

---

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

PR #4599, opened 2024-12-02
chore(mempool): DOG unit tests 

### LINKED ISSUE #4544: mempool/dog: Write unit tests for routing mechanism

jmalicevic:
Closes #4544, #4545

Add unit tests for DOG. 

---

#### PR checklist

- [ ] Tests written/updated


mempool/reactor_test.go
```diff
@@ -685,3 +685,162 @@ func TestMempoolVectors(t *testing.T) {
...
+
+	firstNodeID := reactors[0].Switch.NodeInfo().ID()
+
+	secondNodeID := reactors[1].Switch.NodeInfo().ID()
+	secondNode := reactors[0].Switch.Peers().Get(secondNodeID)
+
+	thirdNodeID := reactors[2].Switch.NodeInfo().ID()
+
+	reactors[0].router.disableRoute(secondNodeID, firstNodeID)
+	reactors[0].router.disableRoute(thirdNodeID, firstNodeID)
```
hvanz, line 834: In a route, source and target are peers of a node. Here a node disables a route to itself (reactors[0] = firstNodeID).

mempool/reactor_test.go
```diff
@@ -685,3 +685,162 @@ func TestMempoolVectors(t *testing.T) {
...
+	// Add transactions to node 2 from node 3
+	for _, tx := range txUnique {
+		_, err := reactors[1].TryAddTx(tx, thirdNodeFromSecond)
+		require.NoError(t, err)
+	}
+
+	// Add transactions to node 3 from node 2
+	// node3.senders[tx] = node2
+	for _, tx := range txUnique {
+		_, err := reactors[2].TryAddTx(tx, secondNodeFromThird)
```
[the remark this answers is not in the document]
jmalicevic: True, but this is kind of bypassing the gossip system to make sure Node 3 has senders. Although I think now tha tfor these purposes we do not have to send anything to Node 2. All we care about is that node 3 has node 2 in its senders list.

mempool/reactor_test.go
jmalicevic: Yes that's true.. I wanted to test whether if we have more routes they all get deleted. I can create a 4th node to have more peers and do not add the first node. That way users will not be inclined to think this is acceptable behaviour.

mempool/reactor_test.go
```diff
@@ -685,3 +685,182 @@ func TestMempoolVectors(t *testing.T) {
...
+	// Wait for the redundancy adjustment to kick in
+	time.Sleep(1 * time.Second)
+
+	// Make sure that Node 3 has at least one disabled route
+	require.Greater(t, len(reactors[2].router.disabledRoutes), 0)
+
+	require.True(t, reactors[2].router.isRouteDisabled(secondNodeFromThird.ID(), firstNodeFromThird.ID()))
+
+	// This will force Node 3 to delete all disabled routes
+	reactors[2].Switch.StopPeerGracefully(secondNode)
```
jmalicevic, line 806: This is also in fact testing the same thing as the next test but the next one is moreelaborate including more than one route.

mempool/reactor_test.go
```diff
@@ -685,3 +685,184 @@ func TestMempoolVectors(t *testing.T) {
...
+	// Put the interval to a higher value to make sure the values don't get reset
+	config.Mempool.DOGAdjustInterval = 15 * time.Second
+	reactors, _ := makeAndConnectReactors(config, 2, nil)
+
+	// create random transactions
+	txs := NewRandomTxs(numTxs, 20)
+	secondNodeID := reactors[1].Switch.NodeInfo().ID()
+	secondNode := reactors[0].Switch.Peers().Get(secondNodeID)
+
+	txUnique := make(map[string][]byte, len(txs))
```
hvanz, line 705: Is `txUnique` really needed? Can `txs` have duplicate elements?

mempool/reactor.go
```diff
@@ -641,6 +649,6 @@ func (rc *redundancyControl) blockHaveTx() {
 }
 
 func (rc *redundancyControl) triggerAdjustment() {
-	<-rc.adjustTicker.C
+	rc.triggerAdjustmentCh <- struct{}{}
```
hvanz, line 652: This new channel is not needed. It can be replaced by calling `rc.adjustRedundancy(memR)` directly. What I'm wondering is if it's also required to do `rc.adjustTicker.Reset(rc.adjustInterval)` afterwards. Because if two `adjustRedundancy` are triggered at the same time, the second one that executes will have zero received transactions.

mempool/reactor_test.go
```diff
@@ -685,3 +685,184 @@ func TestMempoolVectors(t *testing.T) {
...
+	require.Equal(t, int64(len(txUnique)), reactors[0].redundancyControl.firstTimeTxs)
+	for _, tx := range txUnique {
+		_, err := reactors[0].TryAddTx(tx, secondNode)
+		// The transaction is in cache, hence the Error
+		require.Error(t, err)
+	}
+	require.Equal(t, int64(len(txUnique)), reactors[0].redundancyControl.duplicateTxs)
+
+	reactors[0].redundancyControl.triggerAdjustment()
+	time.Sleep(1 * time.Second)
```
hvanz, line 724: This can be replaced by `reactors[0].redundancyControl.adjustRedundancy(reactors[0])`.

mempool/reactor_test.go
```diff
@@ -685,3 +685,184 @@ func TestMempoolVectors(t *testing.T) {
...
+	// a node that has sent this tx to node 3(node 2) and node1
+	for _, tx := range txUnique {
+		_, err := reactors[0].TryAddTx(tx, thirdNodeFromFirst)
+		// The transaction is in cache, hence the Error
+		require.ErrorIs(t, err, ErrTxInCache)
+	}
+
+	reactors[0].redundancyControl.triggerAdjustment()
+	// Wait for the redundancy adjustment to kick in
+	time.Sleep(1 * time.Second)
```
hvanz, line 798: This can be replaced by `reactors[0].redundancyControl.adjustRedundancy(reactors[0])`.

mempool/reactor.go
jmalicevic: YOu are right, I will replace this.. The reset would be needed in my opinion because you want to be sure the interval is respected afterwards.

mempool/reactor_test.go
jmalicevic: I would leave `triggerAdjustments` as is because the function should then call `adjustRedundancy` followed by a `Reset`.

mempool/reactor_test.go
jmalicevic: there is a function that can be used for this, I edited the code 6ea2193f2

mempool/reactor_test.go
hvanz: Ok. Sounds good. I guess the 1 second delay is to wait for reactors[0] to send a message and for reactors[2] to disable a route. Is 100 ms enough for that? Waiting one second here and there slow downs a lot the tests.

mempool/reactor_test.go
hvanz: Is the following correct? The next one (TestDOGRemoveDisabledRoutesOnDisconnect) tests mainly disabled routes, while this one also tests sending HaveTx messages. Then both are needed.

mempool/reactor.go
```diff
@@ -554,40 +554,43 @@ func newRedundancyControl(config *cfg.MempoolConfig) *redundancyControl {
...
-	for {
-		select {
-		case <-rc.adjustTicker.C:
-			// Compute current redundancy level and reset transaction counters.
-			redundancy := rc.currentRedundancy()
-			if redundancy < 0 {
-				// There were no transactions during the last iteration. Do not adjust.
-				continue
-			}
+func (rc *redundancyControl) adjustRedundancy(memR *Reactor) {
```
hvanz, line 557: This new function is now a refactoring. It doesn't change the logic. Up to you if you want to keep it.

hvanz (approved):
🚀

mempool/reactor.go
jmalicevic: since we call it from `triggerAdjustment` I will leave it. Thanks for all the help on this!

mempool/reactor_test.go
jmalicevic: I cancheck,maybe it will be enough

mempool/reactor.go
hvanz: ah right, that's needed then 👍

mempool/reactor_test.go
jmalicevic: Lowered the value to 100ms.

PR #4601, opened 2024-12-03
build(deps): CometBFT no longer imports `cometbft-db`

alesforz:
Partially addresses #4486.

### Context

### Changes
This PR removes cometbft-db as a dependency and changes all the code to use the new `internal/storage` pkg instead.

### TODO

- [x] code uses `db` package in cometbft; does not import cometbft-db anymore. Done in this branch.
- [x] export `DB`, `Batch`, and `Iterator` interfaces. Merged to this branch: #4690 
- [x] change top level commands in `cmd` folder to create a new database instance using `config.DefaultDBProvider()`. WIP: #4697 

---

#### PR checklist

- [x] Tests written/updated


abci/example/kvstore/kvstore.go
```diff
@@ -60,7 +60,7 @@ type Application struct {
 
 // NewApplication creates an instance of the kvstore from the provided database,
 // with the given lanes and priorities.
-func NewApplication(db dbm.DB, lanePriorities map[string]uint32) *Application {
+func NewApplication(db storage.DB, lanePriorities map[string]uint32) *Application {
```
melekes, line 63: This is a public API. We should not use private types here - `storage.DB`.

light/example_test.go
```diff
@@ -42,7 +42,7 @@ func ExampleClient_Update() {
 		stdlog.Fatal(err)
 	}
 
-	db, err := dbm.NewPebbleDB("light-client-db", dbDir)
+	db, err := storage.NewDB("light-client-db", dbDir)
```
melekes, line 45: Again, we're using private fn inside a public example.

light/store/db/db.go
```diff
@@ -65,11 +65,11 @@ func setDBKeyLayout(db dbm.DB, lightStore *dbs, dbKeyLayoutVersion string) {
 
 // New returns a Store that wraps any DB (with an optional prefix in case you
 // want to use one DB with many light clients).
-func New(db dbm.DB, prefix string) store.Store {
+func New(db storage.DB, prefix string) store.Store {
```
melekes, line 68: I am not convinced that `DB` interface must be private. This is a public API, but we use `internal` interface here. Say you are an outside developer and want to implement your own database. How are you supposed to know what to pass here? I think the correct way is to expose the `DB` interface and use it in all public API so that developers can pass any DB instance as long as it satisfies the `DB` interface.

melekes (commented):
great work @alesforz 👍

light/store/db/db.go
alesforz: See #4690

melekes (approved):
👍

zrbecker:
@melekes I noticed this was marked with wip, but had an approval.

Looks like this PR has been stale about a month. Should this work still be done?

melekes:
> Should this work still be done?

It would be beneficial for CometBFT.

PR #4605, opened 2024-12-04
build(deps): Mock generation no longer uses the `latest` tag on `mockery`.

alesforz:
### Context
Using `latest` for `mockery` causes changes in the mocks with almost every new release, which in turn makes our CI fail. By fixing `mockery`'s version, we can prevent this issue. Additionally, `mockery`'s documentation suggests not to use `latest` anyway, so we were doing it incorrectly from the start.

### Changes
This PR pins the `mockery` version to `v2.49.2` (the latest release). The `main` branch already uses mocks generated with this version, so using any other version of `mockery` would alter the mocks again. At least we get the benefit from the latest bug fixes.

---

#### PR checklist

- ~[ ] Tests written/updated~


jmalicevic (approved):
Can you please follow up with the backport PRs? I think you will have to checkout the branches and regenerate the mocks before merging the change, otherwise the CI will fail on those branches. 

Thanks for this!

PR #4633, opened 2024-12-09
fix(test): `TestBlockPoolMaliciousNode` shutdown threads at exit

melekes:
This is a drive-by fix of a test that doesn't shut its threads down until the whole `go test` execution finishes. I think we have a bunch of these, but I came across this one during an unrelated troubleshooting.

Is it worth fixing this? It's not really causing any issues, it's just sloppy coding.

The only way to see any difference is to run the `go test` until it reaches its time limit and panics. In that case, the trace will contain references to the threads.

For example:
```
go test github.com/cometbft/cometbft/blocksync -v -run TestBlockPoolMaliciousNode -count 100 -failfast -race -timeout 30s
```

After 30 seconds the test didn't run 100 times yet, hence `go test` panics. Because the test has been run multiple times already, multiple sets of threads will be reported in the panic. With the fix, only one set is reported.

Author: @greg-szabo


melekes:
```
--- FAIL: TestMempoolTxsBytes (0.00s)
    clist_mempool_test.go:718: 
        	Error Trace:	/home/runner/work/cometbft/cometbft/mempool/clist_mempool_test.go:718
        	Error:      	Should be in error chain:
        	            	expected: "lane  is full: number of txs 0 (max: 0), total bytes 0 (max: 0)"
        	            	in chain: 
        	Test:       	TestMempoolTxsBytes
```

seems unrelated

alesforz (approved):

PR #4639, opened 2024-12-10
chore(node): `Node` no longer uses a singleton to create an `rpc.Environment`

alesforz:
### Context
Because `rpc.Environment` does not store a `GenesisDoc` in memory anymore, (see #1290), we don't need to create it as a singleton. The risk of storing multiple copies of the genesis in memory isn't there anymore, because we now load it from disk.

### This Change
This PR removes the `sync.Once` construct that we put in place.

### Additional Note
The change also ensures that [`TestProvider`](https://github.com/cometbft/cometbft/blob/b16c6fc2c8b2fc5a468fead32a8fe9057d6cce2f/light/provider/http/http_test.go#L36) in the `light/provider/http` package behaves as expected. In `main`, this test passes because it's set up using a [`MemDB`](https://github.com/cometbft/cometbft-db/blob/4cf60c715fe8daccb9dce3b24295575bd461d5d8/memdb.go#L52), which is a dummy in-memory store rather than a real database. Thus, database `Get` operations always succeed.

However, in the context of the [work to remove cometbft-db](https://github.com/cometbft/cometbft/pull/4601), we now use a "real" database, i.e., one that a `Node` closes when it shuts down. Since the `Environment` object was treated as a singleton, each iteration of `TestProvider` created a new `Node` using the same underlying database. This database would be closed at the end of the first iteration when that iteration's `Node` shut down. Subsequent iterations then attempted to call `Get` on a closed database, causing a panic.

This change fixes that issue.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/.changelog/unreleased/improvements/4639-remove-sync-once.md
+++ b/.changelog/unreleased/improvements/4639-remove-sync-once.md
@@ -1,0 +1,2 @@
+- `[node]` `Node` no longer uses a singleton to create an `rpccore.Environment`
+  ([\#4639](https://github.com/cometbft/cometbft/pull/4639))
--- a/node/node.go
+++ b/node/node.go
@@ -9,7 +9,6 @@ import (
 	"net"
 	"net/http"
 	"os"
-	"sync"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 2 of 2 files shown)


melekes (approved):
👍

PR #4642, opened 2024-12-10
fix(store): `TxIndexer` no longer leaks its database's goroutines.

alesforz:
---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/internal/inspect/inspect.go
+++ b/internal/inspect/inspect.go
@@ -3,8 +3,10 @@ package inspect
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net"
 	"os"
+	"strings"
 
 	"golang.org/x/sync/errgroup"
 
@@ -38,8 +40,9 @@ type Inspector struct {
 
 	// References to the state store and block store are maintained to enable
 	// the Inspector to safely close them on shutdown.
-	ss state.Store
-	bs state.BlockStore
+	ss    state.Store
+	bs    state.BlockStore
+	txIdx txindex.TxIndexer
 }
 
 // New returns an Inspector that serves RPC on the specified BlockStore and StateStore.
@@ -64,6 +67,7 @@ func New(
 		logger: logger,
 		ss:     ss,
 		bs:     bs,
+		txIdx:  txidx,
 	}
 }
 
@@ -93,16 +97,43 @@ func NewFromConfig(cfg *config.Config) (*Inspector, error) {
 // Run starts the Inspector servers and blocks until the servers shut down. The passed
 // in context is used to control the lifecycle of the servers.
 func (ins *Inspector) Run(ctx context.Context) error {
-	defer ins.bs.Close()
-	defer ins.ss.Close()
+	defer ins.Close()
 
 	return startRPCServers(ctx, ins.config, ins.logger, ins.routes)
 }
 
-func startRPCServers(ctx context.Context, cfg *config.RPCConfig, logger log.Logger, routes rpccore.RoutesMap) error {
-	g, tctx := errgroup.WithContext(ctx)
-	listenAddrs := cmtstrings.SplitAndTrimEmpty(cfg.ListenAddress, ",", " ")
-	rh := rpc.Handler(cfg, routes, logger)
+// Close closes all of the databases that the Inspector uses.
+func (ins *Inspector) Close() error {
+	var errs []string
+
+	if err := ins.txIdx.Close(); err != nil {
+		errs = append(errs, "txIdx: "+err.Error())
+	}
+	if err := ins.ss.Close(); err != nil {
+		errs = append(errs, "ss: "+err.Error())
+	}
+	if err := ins.bs.Close(); err != nil {
+		errs = append(errs, "bs: "+err.Error())
+	}
+
+	if len(errs) == 0 {
+		return nil
+	}
+
+	return fmt.Errorf("closing inspector's databases: %s", strings.Join(errs, "; "))
+}
+
+func startRPCServers(
+	ctx context.Context,
+	cfg *config.RPCConfig,
+	logger log.Logger,
+	routes rpccore.RoutesMap,
+) error {
+	var (
+		g, tctx     = errgroup.WithContext(ctx)
+		listenAddrs = cmtstrings.SplitAndTrimEmpty(cfg.ListenAddress, ",", " ")
+		rh          = rpc.Handler(cfg, routes, logger)
+	)
 	for _, listenerAddr := range listenAddrs {
 		server := rpc.Server{
 			Logger:  logger,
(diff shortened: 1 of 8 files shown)


.changelog/unreleased/breaking-changes/4642-close-txindexer.md
```diff
@@ -0,0 +1,2 @@
+- `[state]` `TxIndexer` no longer leaks its database's goroutines.
```
alesforz, line 1: The change also adds a `Close` method to the `TxIndexer` interface, though this is the main benefit of this PR. Can I just add that point to the same changelog here?

jmalicevic: This should be state here in the message.

internal/inspect/inspect.go
```diff
@@ -111,27 +142,38 @@ func startRPCServers(ctx context.Context, cfg *config.RPCConfig, logger log.Logg
...
+					"address", listenerAddr,
+					"certfile", certFile,
+					"keyfile", keyFile,
+				)
+
 				err := server.ListenAndServeTLS(tctx, certFile, keyFile)
 				if !errors.Is(err, net.ErrClosed) {
 					return err
 				}
+
```
jmalicevic, line 162:
```suggestion
```

internal/inspect/inspect.go
```diff
@@ -111,27 +142,38 @@ func startRPCServers(ctx context.Context, cfg *config.RPCConfig, logger log.Logg
...
 				logger.Info("RPC HTTPS server stopped", "address", listenerAddr)
 				return nil
 			})
 		} else {
 			listenerAddr := listenerAddr
+
```
jmalicevic, line 168:
```suggestion
```

internal/inspect/inspect.go
```diff
@@ -111,27 +142,38 @@ func startRPCServers(ctx context.Context, cfg *config.RPCConfig, logger log.Logg
...
 			g.Go(func() error {
 				logger.Info("RPC HTTP server starting", "address", listenerAddr)
+
```
jmalicevic, line 171:
```suggestion
```

internal/inspect/inspect.go
```diff
@@ -111,27 +142,38 @@ func startRPCServers(ctx context.Context, cfg *config.RPCConfig, logger log.Logg
...
 				err := server.ListenAndServe(tctx)
 				if !errors.Is(err, net.ErrClosed) {
 					return err
 				}
+
```
jmalicevic, line 176:
```suggestion
```

internal/inspect/inspect_test.go
```diff
@@ -44,7 +47,9 @@ func TestInspectRun(t *testing.T) {
 	t.Run("from config", func(t *testing.T) {
 		d, err := inspect.NewFromConfig(cfg)
 		require.NoError(t, err)
+
```
jmalicevic, line 50:
```suggestion
```

internal/inspect/inspect_test.go
```diff
@@ -44,7 +47,9 @@ func TestInspectRun(t *testing.T) {
...
 		ctx, cancel := context.WithCancel(context.Background())
+
```
jmalicevic, line 52:
```suggestion
```

state/txindex/indexer_service.go
```diff
@@ -128,4 +128,6 @@ func (is *IndexerService) OnStop() {
 	if is.eventBus.IsRunning() {
 		_ = is.eventBus.UnsubscribeAll(context.Background(), subscriber)
 	}
+
+	is.txIdxr.Close()
```
jmalicevic, line 132: Check for the error?

state/txindex/kv/kv.go
```diff
@@ -61,6 +61,15 @@ func WithCompaction(compact bool, compactionInterval int64) IndexerOption {
 	}
 }
 
+// Close closes the indexer's underlying database. The caller is responsible for
+// calling Close when done with the indexer.
+func (txi *TxIndex) Close() error {
+	if err := txi.store.Close(); err != nil {
```
jmalicevic, line 67: This is closing the prefixDB? Anything special we need to do here?

internal/inspect/inspect.go
jmalicevic: Btw the new lines are to avoid having too many line changes but if it is refactoring done automatically then ignore them.

state/txindex/indexer.go
```diff
@@ -36,6 +36,11 @@ type TxIndexer interface {
 	GetRetainHeight() (int64, error)
 
 	SetRetainHeight(retainHeight int64) error
+
+	// Close closes the underlying database that a TxIndexer uses. It is the
```
melekes, line 40:
```suggestion
	// Close frees any resources associated with the indexer. Note that calls to other methods after calling Close may lead to panic.
```

`TxIndexer` had no notion of a database up until this point. The message should somehow be amended to introduce this gradually.

state/txindex/kv/kv.go
melekes, line 67:
```suggestion
	return txi.store.Close()
```

melekes (commented):
good catch @alesforz 👍 

My only suggestion is to amend the documentation slightly to reflect the fact that we're assuming there's an underlying database used, and `Close` will close it + no calls are allowed after.

state/txindex/kv/kv.go
alesforz: No, this closes the actual database. I did not add anything to close the `PrefixDB` because 1) it's not a "real" database, so nothing to close; 2) che changed the `Close` method of `PrefixDB` to be a no-op; 3) by closing the database, we also make `PrefixDB` unusable (as expected).

internal/inspect/inspect.go
alesforz: I add the empty lines manually because I think it makes the code a bit more readable. Sometimes I feel it is too packed for comfortable parsing, but maybe it's just me 😅

state/txindex/kv/kv.go
alesforz: I prefer to check errors and add context to the returned error message rather than returning a function call’s error directly. This approach creates a “stack trace” in the error message, making it easy to trace the error’s origin. Without these intermediate checks, it’s harder to identify which function call led to the error, and simply returning the function call’s error provides no direct hint about where in the codebase it came from. To be clear:
```go
func A() error { 
    if err := B(); err != nil {
        return fmt.Errorf("A(): %s" err)
    }
}

func B() error { 
   return C()
}

func C() error { 
   // do stuff...
   return errors.New("C(): some error")
}
```
In this example, in case of an error, we won’t see from the final error message returned by `A` that `B` was ever called. CometBFT’s codebase frequently chains calls like this, and I’ve often had to spend extra time determining which functions were involved because the error messages didn’t make it clear.

Does it make sense? What do you think? 😄

melekes: In principle, I agree, but it does make the code more hairy 😄

melekes (approved):

state/txindex/kv/kv.go
jmalicevic: Ok yes, store is the actual DB not the wrapper. GOod.

jmalicevic (approved):

PR #4643, opened 2024-12-10
refactor(p2p)!: merge #4307 back into `main`

### LINKED ISSUE #4553: p2p: merge #4307 into `main`
It would be great to merge #4307 into `main` BUT revert https://github.com/cometbft/cometbft/pull/4307#discussion_r1858978040 part so that the code is optimized for the current TCP transport. See https://github.com/cometbft/cometbft/pull/4307#discussion_r1850397779.

melekes:
Closes #4553 

Refs #4307


p2p/transport/tcp/conn/connection.go
```diff
@@ -269,147 +233,180 @@ func (c *MConnection) stopServices() (alreadyStopped bool) {
...
+		d = desc
+	}
+	c.channelsIdx[streamID] = newChannel(c, d)
+	c.channelsIdx[streamID].SetLogger(c.Logger.With("streamID", streamID))
+	c.msgsByStreamIDMap[streamID] = make(chan []byte, maxRecvChanCap)
+
+	return &MConnectionStream{conn: c, streamID: streamID}, nil
+}
+
+func (c *MConnection) HandshakeStream() transport.HandshakeStream {
```
melekes, line 279: add docs

p2p/transport/tcp/tcp.go
```diff
@@ -186,15 +181,17 @@ func (mt *MultiplexTransport) Dial(
...
 	}
 
-	secretConn, err := mt.upgrade(c, &addr)
+	mconn, _, err := mt.upgrade(c, &addr)
 	if err != nil {
 		return nil, err
 	}
+	mconn.SetLogger(mt.logger.With("remote", addr))
+
+	go mt.cleanupConn(c.RemoteAddr(), mconn.Quit())
```
melekes, line 190: this should probably be one goroutine for all conns

p2p/transport/tcp/conn/connection.go
```diff
@@ -98,10 +102,13 @@ type MConnection struct {
 
 	_maxPacketMsgSize int
 
-	// streamID -> list of incoming messages
-	recvMsgsByStreamID map[byte]chan []byte
```
jmalicevic: I guess this addresses the concern from this [comment](https://github.com/cometbft/cometbft/pull/4307/files#r1850397779) ?

jmalicevic (commented):
Changelog, especially if there are breaking changes (API or otherwise).

p2p/peer.go
```diff
@@ -252,47 +337,74 @@ func (p *peer) SocketAddr() *na.NetAddr {
...
-// Send msg bytes to the channel identified by chID byte. Returns false if the
-// send queue is full after timeout, specified by MConnection.
+// Send sends the given envelope via the stream identified by e.ChannelID. It
+// blocks until the message is sent.
 //
-// thread safe.
-func (p *peer) Send(e Envelope) bool {
-	return p.send(e.ChannelID, e.Message, p.mconn.Send)
+// thread-safe.
+func (p *peer) Send(e Envelope) error {
```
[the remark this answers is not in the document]
jmalicevic: I was about to say it is fine until I saw a use case where when the peer is not running the messages is not sent and we'd send a `nil` error. I would define a specific error for non erroneous cases when a message cannot be sent. Because people should know when a message was not sent.  Or leave it with a bool. Or return (bool,error)?

p2p/transport/tcp/conn/connection.go
melekes: yes

p2p/peer.go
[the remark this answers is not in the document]
alesforz: I'd say `(bool, error)` is the idiomatic solution in Go for such cases. If it is a recoverable situation (and `Send` on a peer not running sounds like it), I also think we should define an error case for it and return `(false, err)` rather than `panic`ing.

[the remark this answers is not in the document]
melekes: > until I saw a use case where when the peer is not running the messages is not sent and we'd send a nil error.

This was an oversight 👍 fixed

melekes:
> Changelog, especially if there are breaking changes (API or otherwise).

done ✅

p2p/transport/tcp/conn/connection.go
```diff
@@ -269,147 +233,181 @@ func (c *MConnection) stopServices() (alreadyStopped bool) {
...
 	if !c.IsRunning() {
-		return false
+		return nil
 	}
 
-	c.Logger.Debug("Send", "channel", chID, "conn", c, "msgBytes", log.NewLazySprintf("%X", msgBytes))
+	// c.Logger.Debug("Send",
+	// 	"streamID", chID,
+	// 	"msgBytes", log.NewLazySprintf("%X", msgBytes),
+	// 	"timeout", timeout)
```
jmalicevic, line 388:
```suggestion
	// c.Logger.Debug("Send",
	// 	"streamID", chID,
	// 	"msgBytes", log.NewLazySprintf("%X", msgBytes),
	// 	"timeout", timeout)
```
Do we want to keep this debug message?

p2p/transport/tcp/conn/connection.go
```diff
@@ -269,147 +233,181 @@ func (c *MConnection) stopServices() (alreadyStopped bool) {
...
-	success := channel.sendBytes(msgBytes)
-	if success {
-		// Wake up sendRoutine if necessary
-		select {
-		case c.send <- struct{}{}:
-		default:
-		}
-	} else {
-		c.Logger.Debug("Send failed", "channel", chID, "conn", c, "msgBytes", log.NewLazySprintf("%X", msgBytes))
+		panic(fmt.Sprintf("Unknown channel %X. Forgot to register?", chID))
```
jmalicevic, line 392: Are we sure we want to panic and crash the node?  Can we have a situation where the channel was for some reason not yet registered but there is something being sent to it ?

p2p/transport/tcp/conn/connection.go
```diff
@@ -269,147 +233,181 @@ func (c *MConnection) stopServices() (alreadyStopped bool) {
...
-
-	c.Logger.Debug("TrySend", "channel", chID, "conn", c, "msgBytes", log.NewLazySprintf("%X", msgBytes))
-
-	// Send message to channel.
-	channel, ok := c.channelsIdx[chID]
-	if !ok {
-		c.Logger.Error(fmt.Sprintf("Cannot send bytes, unknown channel %X", chID))
-		return false
+	if err := channel.sendBytes(msgBytes, blocking); err != nil {
+		// c.Logger.Error("Send failed", "err", err)
```
jmalicevic, line 395:
```suggestion
		c.Logger.Error("Send failed", "err", err)
```
or delete the comment.

p2p/transport/tcp/conn/connection.go
```diff
@@ -655,30 +659,33 @@ FOR_LOOP:
 		case *tmp2p.Packet_PacketMsg:
 			channelID := byte(pkt.PacketMsg.ChannelID)
 			channel, ok := c.channelsIdx[channelID]
-			if pkt.PacketMsg.ChannelID < 0 || pkt.PacketMsg.ChannelID > math.MaxUint8 || !ok || channel == nil {
+			if !ok || pkt.PacketMsg.ChannelID < 0 || pkt.PacketMsg.ChannelID > math.MaxUint8 {
 				err := fmt.Errorf("unknown channel %X", pkt.PacketMsg.ChannelID)
-				c.Logger.Debug("Connection failed @ recvRoutine", "conn", c, "err", err)
-				c.stopForError(err)
+				c.Logger.Error("Connection failed @ recvRoutine", "err", err)
```
jmalicevic, line 664: Are you sure about making this an Error vs Debug?

p2p/transport/tcp/conn/connection.go
```diff
@@ -655,30 +659,33 @@ FOR_LOOP:
...
 				break FOR_LOOP
 			}
 
 			msgBytes, err := channel.recvPacketMsg(*pkt.PacketMsg)
 			if err != nil {
-				if c.IsRunning() {
-					c.Logger.Debug("Connection failed @ recvRoutine", "conn", c, "err", err)
-					c.stopForError(err)
-				}
+				c.Logger.Error("Connection failed @ recvRoutine", "err", err)
```
jmalicevic, line 671: Do you know under which scenarios was it ok not to stop the peer here ? It was stopped only when the inner if was tru ?

p2p/transport/tcp/conn/connection.go
```diff
@@ -655,30 +659,33 @@ FOR_LOOP:
...
+				c.Close(err.Error())
 				break FOR_LOOP
 			}
 			if msgBytes != nil {
-				c.Logger.Debug("Received bytes", "chID", channelID, "msgBytes", msgBytes)
-				// NOTE: This means the reactor.Receive runs in the same thread as the p2p recv routine
-				c.onReceive(channelID, msgBytes)
+				// c.Logger.Debug("Received", "streamID", channelID, "msgBytes", log.NewLazySprintf("%X", msgBytes))
```
jmalicevic, line 676:
```suggestion
				 c.Logger.Debug("Received", "streamID", channelID, "msgBytes", log.NewLazySprintf("%X", msgBytes))
```
Restrore msg ?

p2p/peer.go
```diff
@@ -189,35 +257,53 @@ func (p *peer) String() string {
...
+		p.Logger.Error("Close", "err", err)
+	}
 }
 
 // OnStop implements BaseService.
 func (p *peer) OnStop() {
-	if err := p.mconn.Stop(); err != nil { // stop everything and close the conn
-		p.Logger.Debug("Error while stopping peer", "err", err)
+	if err := p.Close("stopping peer"); err != nil {
+		p.Logger.Error("Close", "err", err)
```
jmalicevic, line 306: We are increasing log level here. Afraid it might create too much noise.

p2p/transport/tcp/conn/connection.go
jmalicevic: Also the log level was debug before.

p2p/transport/tcp/conn/connection.go
melekes: > Are we sure we want to panic and crash the node? Can we have a situation where the channel was for some reason not yet registered but there is something being sent to it ?

Yes. We have discussed this with Daniel in [the https://github.com/cometbft/cometbft/pull/4307](https://github.com/cometbft/cometbft/pull/4307#discussion_r1858988094). We (devs) should make sure all streams are registered before starting `MConnection`.

p2p/peer.go
melekes: But it's only executed if closing underlying `MConnection` fails

p2p/transport/tcp/conn/connection.go
melekes: It's only useful if you're a CometBFT dev debugging the P2P stack. Otherwise, the message we've received is logged in peer.go.

p2p/transport/tcp/conn/connection.go
melekes: good call 👍

p2p/transport/tcp/conn/connection.go
melekes: same as https://github.com/cometbft/cometbft/pull/4643#discussion_r1881446458

jmalicevic (approved):

Issue #4620, opened 2024-12-07
p2p: sent next PEX request too soon


0xHansLee:
### Setup

**CometBFT version** (use `cometbft version` or `git rev-parse --verify HEAD` if installed from source):

v0.38.12

**Have you tried the latest version**: yes/no

yes (v0.38.15)

**ABCI app** (name for built-in, URL for self-written if it's publicly available):

story (https://github.com/piplabs/story)

**Environment**:
- **OS** (e.g. from /etc/os-release): macOS Somona 14.4
- **Go version**: 1.22.7

### Config

```
[p2p]

# Address to listen for incoming connections
laddr = "tcp://0.0.0.0:26656"

# Address to advertise to peers for them to dial. If empty, will use the same
# port as the laddr, and will introspect on the listener to figure out the
# address. IP and port are required. Example: 159.89.10.97:26656
external_address = ""

# Comma separated list of seed nodes to connect to
seeds = "<seed>"

# Comma separated list of nodes to keep persistent connections to
persistent_peers = ""

# Path to address book
addr_book_file = "config/addrbook.json"

# Set true for strict address routability rules
# Set false for private or local networks
addr_book_strict = false

# Maximum number of inbound peers
max_num_inbound_peers = 40

# Maximum number of outbound peers to connect to, excluding persistent peers
max_num_outbound_peers = 10

# List of node IDs, to which a connection will be (re)established ignoring any existing limits
unconditional_peer_ids = ""

# Maximum pause when redialing a persistent peer (if zero, exponential backoff is used)
persistent_peers_max_dial_period = "0s"

# Time to wait before flushing messages out on the connection
flush_throttle_timeout = "100ms"

# Maximum size of a message packet payload, in bytes
max_packet_msg_payload_size = 1024

# Rate at which packets can be sent, in bytes/second
send_rate = 5120000

# Rate at which packets can be received, in bytes/second
recv_rate = 5120000

# Set true to enable the peer-exchange reactor
pex = true

# Seed mode, in which node constantly crawls the network and looks for
# peers. If another node asks it for addresses, it responds and disconnects.
#
# Does not work if the peer-exchange reactor is disabled.
seed_mode = false

# Comma separated list of peer IDs to keep private (will not be gossiped to other peers)
private_peer_ids = ""

# Toggle to disable guard against peers connecting from the same ip.
allow_duplicate_ip = false

# Peer connection configuration.
handshake_timeout = "20s"
dial_timeout = "3s"
```

[...]

PR #4644, opened 2024-12-11
fix(p2p/pex): do not send PEX request in fast dial mode

cason:
Solves #4620.

Fixes an issue introduced by https://github.com/cometbft/cometbft/pull/3360.

In short, when receiving addresses from a configured seed node, the peer immediately dials the received addresses, without waiting for the `defaultEnsurePeersPeriod` (30s). This is a desired behavior, the "fast dial mode" in the title.

However, for preventing abuse, a node only accepts PEX requests from a peer every `minReceiveRequestInterval()` time, set to `defaultEnsurePeersPeriod/3` (10s). When running this "fast dial mode", however, a PEX request can be send, in some unlucky setup, to the same peer without waiting for the full defaultEnsurePeersPeriod` (30s).

The problem is that at the receive side, a node keeps track of the latest PEX request received from each peer. If two requests are received with an interval lower than `minReceiveRequestInterval()`, the peer is considered abusive: it is disconnected with an `ErrReceivedPEXRequestTooSoon` error and banned from the address book.

This PR proposes a workaround to prevent the above mentioned scenario.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/p2p/pex/pex_reactor.go
+++ b/p2p/pex/pex_reactor.go
@@ -410,7 +410,7 @@ func (r *Reactor) ensurePeersRoutine() {
 
 	// fire once immediately.
 	// ensures we dial the seeds right away if the book is empty
-	r.ensurePeers()
+	r.ensurePeers(true)
 
 	// fire periodically
 	ticker := time.NewTicker(r.ensurePeersPeriod)
@@ -418,9 +418,9 @@ func (r *Reactor) ensurePeersRoutine() {
 	for {
 		select {
 		case <-ticker.C:
-			r.ensurePeers()
+			r.ensurePeers(true)
 		case <-r.ensurePeersCh:
-			r.ensurePeers()
+			r.ensurePeers(false)
 		case <-r.book.Quit():
 			return
 		case <-r.Quit():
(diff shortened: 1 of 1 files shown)


p2p/pex/pex_reactor.go
```diff
@@ -504,7 +504,7 @@ func (r *Reactor) ensurePeers() {
 	if r.book.NeedMoreAddrs() {
 		// 1) Pick a random peer and ask for more.
 		peer := r.Switch.Peers().Random()
-		if peer != nil {
+		if peer != nil && ensurePeersPeriodElapsed {
```
jmalicevic, line 507: Two things ,are we not still dialing peers in L487? If that is also not dialed when `ensurePeersPeriodElapsed` is `false`, are we not effectively back to the same behaviour before the fix in #3360 ? 

I might be missing something , but let's say the value of `ensurePeersPeriod`  is 30s. We got peer addresses at 

`t := 10s` . Should we not only wait for 
`tWait := ensurePeersPeriod - t` ?  

We can't have it both ways, dial immediately and make sure the period is respected. Maybe I am missing something here.

jmalicevic: Ok for 1. I realized that that one is not a `PexRequest` so it won't cause any problems. 

And for 2 I am also sure now that it is fine because we actually want the action above in 487 to happen but we do not want to issue a `RequestPeers` request.

melekes: I also thought about 2, but came to the conclusion that it's still useful because we would be trying to pick peers from the address book and dial them; we just won't be sending `PexRequests` to them if we need more addresses.

PR #4652, opened 2024-12-12
perf(mempool): free memory immediately after calculating max

melekes:
I discovered this while profiling my P2P refactoring.


jmalicevic (commented):
Great catch Ant

jmalicevic (approved):

Issue #1546, opened 2023-10-27
Explicitly set a `MaxConcurrentStreams` for GRPC servers


rootulp:
### Summary

Context: https://github.com/celestiaorg/celestia-core/issues/1127

Explicitly set a `MaxConcurrentStreams` for GRPC servers.

### Problem Definition

https://github.com/grpc/grpc-go/security/advisories/GHSA-m425-mq94-257g suggests

> Along with applying the patch, users should also ensure they are using the grpc.MaxConcurrentStreams server option to apply a limit to the server's resources used for any single connection.

CometBFT has already bumped to https://github.com/cometbft/cometbft/blob/22af40641c58f565b6cc694c9e9871a381627309/go.mod#L36

but an explicit `MaxConcurrentStreams` hasn't been defined https://github.com/cometbft/cometbft/blob/22af40641c58f565b6cc694c9e9871a381627309/abci/server/grpc_server.go#L46

### Proposal

```diff
-	s.server = grpc.NewServer() 
+	// Explicitly set MaxConcurrentStreams to apply a limit to the server's resources used for any single connection.
+	// https://github.com/grpc/grpc-go/security/advisories/GHSA-m425-mq94-257g
+	s.server = grpc.NewServer(grpc.MaxConcurrentStreams(100))
```

We can replace `100` with whichever value you deem appropriate. `100` is the default client max concurrent streams. `math.MaxUint32` is the default server max concurrent streams.

Ref:
- https://github.com/grpc/grpc-go/blob/c76d75f4f95626e9e6c0008e6a6120d43747b519/internal/transport/defaults.go#L34
- https://github.com/grpc/grpc-go/blob/c76d75f4f95626e9e6c0008e6a6120d43747b519/internal/transport/http2_client.go#L343
- https://github.com/grpc/grpc-go/blob/c76d75f4f95626e9e6c0008e6a6120d43747b519/server.go#L177

andynog:
Hi @rootulp, thanks for submitting the issue. We are evaluating it. Is Celestia using the ABCI gRPC server in different ways ? As far as I know you use the in-process server right ? 

Also, the a gRPC server (or RPC) should never be exposed publicly (only through a reverse proxy or Cloudflare that in this particular case should be patched and properly configured too).

rootulp:
> As far as I know you use the in-process server right ?

Correct. The only two instances I found in celestia-core are:
[...]

PR #4653, opened 2024-12-12
chore(gRPC): set a value for grpc.MaxConcurrentStreams

andynog:
close: #1546 

In gRPC, concurrent streams per connection refers to the number of active gRPC calls (streams) that can exist simultaneously over a single HTTP/2 connection between a client and a server.  

This PR adds logic to sett a value of `100` for `grpc.MaxConcurrentStreams` which seems to be a reasonable value based on some documentation and use cases related to gRPC. 

This value is not configurable, the goal is to add a max number to prevent the [default value of Max UInt32](https://github.com/grpc/grpc-go/blob/d6a777f952c77822f0190dff71b1fe8fe250538c/server.go#L417) being used.

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
👍

PR #4667, opened 2024-12-14
perf(p2p): do not remarshal envelope message within broadcast

### LINKED ISSUE #4615: perf(p2p): don't remarshal within broadcast
Sending a message includes 3 steps:

1. Marshaling the `proto.Message` into bytes
2. Calling the underlying connection method passing the bytes
3. Registering metrics and returning the output (`true` or `false`)

The `Switch.Broadcast` method is a helper that takes the set of connected peers and send the same messages to all of them. It does not work well because step 1. is unnecessarily repeated for every peer. So what we want is to avoid 1. to be performed multiple times for the same message.

As suggested in my comments, probably the best way of achieving so is to add a private field to `p2p.Envelope` that stores the marshalled version of the `Message` field. In this way, the existing sending methods can save step 1. without huge changes in interfaces, implementations, etc.

_Originally posted by @cason in https://github.com/cometbft/cometbft/pull/4546#pullrequestreview-2460629216_

Original PR: https://github.com/cometbft/cometbft/pull/4546

najeal:
Closes #4615 

This avoids multiple marshals of the `Envelope` message when broadcasting.

### DIFF
--- a/p2p/peer.go
+++ b/p2p/peer.go
@@ -348,7 +348,11 @@ func (p *peer) Send(e Envelope) error {
 		// This should never happen.
 		return fmt.Errorf("stream %d not found", e.ChannelID)
 	}
-	err := p.send(e.Message, stream.Write /* blocking */)
+	msgBytes, err := e.marshalMessage()
+	if err != nil {
+		return err
+	}
+	err = p.send(e.Message, msgBytes, stream.Write /* blocking */)
 	if err != nil {
 		p.Logger.Error("Send", "err", err)
 		return err
@@ -367,8 +371,11 @@ func (p *peer) TrySend(e Envelope) error {
 		// This should never happen.
 		return fmt.Errorf("stream %d not found", e.ChannelID)
 	}
-
-	err := p.send(e.Message, stream.TryWrite /* non-blocking */)
+	msgBytes, err := e.marshalMessage()
+	if err != nil {
+		return err
+	}
+	err = p.send(e.Message, msgBytes, stream.TryWrite /* non-blocking */)
 	if err != nil {
 		if e, ok := err.(transport.WriteError); ok && e.Full() {
 			p.Logger.Debug("Send", "err", err)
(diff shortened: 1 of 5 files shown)


p2p/switch.go
```diff
@@ -247,6 +247,7 @@ func (sw *Switch) OnStop() {
 //
 // NOTE: Broadcast uses goroutines, so order of broadcast may not be preserved.
 func (sw *Switch) Broadcast(e Envelope) {
+	_, _ = e.marshalMessage()
```
melekes, line 250: Is this a noop? If yes, what's the point of having it?

najeal: Calling `marshalMessage()` initializes the private field `messageBytes` in the `Envelope`. Then calls to `marshalMessage()` into subsequent functions will return the `messageBytes` without remarshaling it.

p2p/peer.go
```diff
@@ -348,7 +348,11 @@ func (p *peer) Send(e Envelope) error {
 		// This should never happen.
 		return fmt.Errorf("stream %d not found", e.ChannelID)
 	}
-	err := p.send(e.Message, stream.Write /* blocking */)
+	msgBytes, err := e.marshalMessage()
```
melekes, line 351: Why do we need to marshal msg here? Can't it be done in `send`?

najeal: `send` function does not have the `Envelope` struct, it only knows `Message`. If we marshal the message in `send` we will marshal X time the same message.
It is why `marshalMessage()` is called here, either it has already been call/done in higher context (Broadcast) and it will return value from the private field `messageBytes` or it will marshal the value.
Envelope struct is required to avoid remarshaling.

melekes: Can't we refactor `send` to accept `Envelope`?

najeal: Changes are done 👍

melekes (approved):
👍

melekes:
Could you add a changelog entry? `.changelog/unreleased/improvements/...`?

najeal:
@melekes changelog entry has been added. One test is failing now but it may come from a flaky test as nothing has changed except the new entry.

PR #4675, opened 2024-12-16
perf(internal/bits): Additional speedup to `bitArray.PickRandom`

### LINKED ISSUE #2849: libs/bits: store the number of non-zero bits in a private field
In #2841, part of the changes associate to find out how many bits, in the bit array/set, are set to zero.

Currently, this is achieved by the `getTrueIndices() []int`, or the faster replacement `getNumTrueIndices() int`.

We can have the number of bits set (==1) as a private field, updated as follows:

-  `SetIndex` adds or subtracts 1
-  `copy()` can be adapted easily
- `IsEmpty()` and `IsFull()` become trivial
- `Update()`, I don't even know the point, same as `copy()`
- `UnmarshalJSON()` uses `SetIndex`, so nothing new here
- `FromProto()` can have this field initialized with a private method (the one added here)

_Originally posted by @cason in https://github.com/cometbft/cometbft/issues/2841#issuecomment-2063477201_

TropicalDog17:
Closes #2849

before & after, we observed roughly 30% speed up with this refactor
```goos: linux
goarch: amd64
pkg: github.com/cometbft/cometbft/internal/bits
cpu: 13th Gen Intel(R) Core(TM) i5-13400
BenchmarkPickRandomBitArray-16          15392144                65.31 ns/op            0 B/op          0 allocs/op
PASS
```

```goos: linux
goarch: amd64
pkg: github.com/cometbft/cometbft/internal/bits
cpu: 13th Gen Intel(R) Core(TM) i5-13400
BenchmarkPickRandomBitArray-16          24890334                44.16 ns/op            0 B/op          0 allocs/op
PASS
```


TropicalDog17:
No idea why a test is failing, I think its flaky and unrelated to this changes, but not really sure

internal/bits/bit_array.go
```diff
@@ -15,9 +15,10 @@ import (
 
 // BitArray is a thread-safe implementation of a bit array.
 type BitArray struct {
-	mtx   sync.Mutex
-	Bits  int      `json:"bits"`  // NOTE: persisted via reflect, must be exported
-	Elems []uint64 `json:"elems"` // NOTE: persisted via reflect, must be exported
+	mtx          sync.Mutex
+	TrueBitCount int      // Number of bits set to true
```
melekes, line 19:
```suggestion
	TrueBitCount int `json:"true_bits_count"`     // Number of bits set to true
```

internal/bits/bit_array.go
```diff
@@ -193,6 +216,15 @@ func (bA *BitArray) not() *BitArray {
 	for i := 0; i < len(c.Elems); i++ {
 		c.Elems[i] = ^c.Elems[i]
 	}
+
+	// Flip count is simply total bits minus current true bits
+	c.TrueBitCount = c.Bits - c.TrueBitCount
+
+	// Clear any excess bits in the last element if necessary
```
melekes, line 223: Where's this coming from?

melekes (commented):
Thanks @TropicalDog17 ❤️

internal/bits/bit_array.go
melekes: Sorry, I meant that we were not clearing any excess bits previously.

TropicalDog17: Yeah maybe it isn't really necessary, I have removed

melekes:
@TropicalDog17 could you run `make pre-commit && pre-commit run --all-files`?

TropicalDog17:
@melekes please review this sir

melekes:
@TropicalDog17 could you please add a changelog? in .changelog (we use [unclog](https://github.com/informalsystems/unclog?rgh-link-date=2024-12-16T14%3A10%3A15Z) to manage our changelog)

melekes (approved):
lgtm 👍

TropicalDog17:
> @TropicalDog17 could you please add a changelog? in .changelog (we use [unclog](https://github.com/informalsystems/unclog?rgh-link-date=2024-12-16T14%3A10%3A15Z) to manage our changelog)

Hey, I've added the changelog @melekes

PR #4679, opened 2024-12-17
docs(types): fix function names

linchizhen:


melekes (approved):
Thanks @linchizhen ❤️

PR #4680, opened 2024-12-17
feat(crypto): update `bls12381` pkg

alesforz:
### Changes
- added func `NewPublicKeyFromCompressedBytes` to create a new BLS12-381 public key from a slice of bytes representing a compressed 48-bytes BLS12-381 key created elsewhere.
- added method `Compress` to type `PubKey` to return compress the BLS12-381 public key from 96 to 48 bytes.
- added a new exported error `ErrPubKeyDecompression`.


melekes (approved):
👍

PR #4681, opened 2024-12-17
fix(e2e): skip missing snapshots

melekes:
when pruning. Suggested by GH copilot while nightly test failure debugging.


jmalicevic (approved):
Should we add a changelog here?

melekes:
> Should we add a changelog here?

I don't see a need. Git is the perfect changelog for developers.

PR #4690, opened 2024-12-18
feat(storage): Export `DB` interface

alesforz:
### Context
In reviewing #4601, we discussed about exporting the `DB`, `Batch`, and `Iterator` interfaces instead of leaving them in an `internal` package. After some thinking we decided to export them in a new non-internal package `cmtdb`. However, our implementation of those interfaces remain unexported.

### Changes
This PR adds pkg `cmtdb` exporting the `DB`, `Batch`, and `Iterator` interfaces. Additionally, it changes the code to use this new package rather than the now obsolete `internal/storage` (which this PR removes).

---

#### PR checklist

- [x] Tests written/updated
- ~[] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)~


abci/example/kvstore/kvstore.go
```diff
@@ -13,9 +13,9 @@ import (
 	"time"
 
 	"github.com/cometbft/cometbft/abci/types"
+	"github.com/cometbft/cometbft/cmtdb"
```
melekes, line 16: Why do we need the `cmt` prefix? If it's located within `cometbft/cometbft`

melekes (commented):
👍

abci/example/kvstore/kvstore.go
alesforz: I thought it would make it consistent with other parts of the code in which we import packages by adding the `cmt` prefix.

melekes: Yeah, but you can always alias `cmtdb github.com/cometbft/cometbft/db` if you want.

alesforz: I'll make the changes

melekes (approved):

PR #4698, opened 2024-12-20
chore(changelog): Post release changelog sync

jmalicevic:
---

#### PR checklist

- [ ] Tests written/updated


melekes:
Why are we removing entries from `.changelog/v0.37.0`?

jmalicevic:
> Why are we removing entries from `.changelog/v0.37.0`?

good catch, it also removed the lanes stuff from unreleased.. I used unclogs find duplicates and then deleted everything on the left side via script. Not sure where the 0.37 stuff came from. I'll restore this manually.

jmalicevic:
@melekes  , should be good now.. the problem was that unclog duplicates found the entries we added twice (under breaking changes and under features), and then one of those would be deleted.

PR #4688, opened 2024-12-18
feat(vote extensions)!: support non replay-protected vote extensions

bermuell:
---
Closes: #4682 

Add support for non replay-protected vote extensions.  Non replay-protected vote extensions are not wrapped with additional meta information such as `chain-id`, `height` and `round` (see CanonicalVoteExtension) used for replay-protection. 
These non replay-protected vote extensions are optional.

Note:
- New v2 for proto files of abci and cometbft/types added. v2 is identical to v1 + changes for vote extensions as seen in commit [53ea5be26](https://github.com/cometbft/cometbft/pull/4688/commits/53ea5be26e170e06a0e140d9ab029902e57a2982)
- Bumps ABCI version to v2.3.0

#### PR checklist

- [x] Tests written/updated
- [x] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [x] Updated relevant documentation (`docs/` or `spec/`) and code comments

PR #4721, opened 2024-12-23
fix(e2e): Reduce size of generated vote extensions and don't convert to Hex

jmalicevic:
PR #4688 increased the size of vote extensions used in the e2e app which resulted in e2e tests failing because the messages were too  big. 

This PR does two things:
- Does not convert the Non replay-protected vote extensions to a Hex String (as this doubles their size). 
- Uses a lower value for the generated VEs because this value will in practice be doubled (if the configuration size is 256, we will have `VoteExtension` of size 256 and `NonRpExtension` of size 256). 

---

#### PR checklist

- [ ] Tests written/updated


bermuell (approved):

melekes (approved):
Thanks @jmalicevic ❤️

PR #4727, opened 2024-12-24
fix(types): update `MaxSignatureSize` to include secp256k1eth

melekes:
Follow-up to https://github.com/cometbft/cometbft/pull/2534


jmalicevic (approved):
Thanks!

PR #4729, opened 2024-12-26
feat(p2p): add QUIC transport

melekes:
Refs https://github.com/cometbft/cometbft/issues/4302

TODO:
- [ ] implement [`quic.Conn`](https://github.com/cometbft/cometbft/blob/17c8345bce997eefe6ff13441996f3b6e927b7c2/p2p/transport/quic/conn.go)
- [ ] add tests for the transport and connection


faddat:
basically took the same approach but sadly, got the same result

PR #4732, opened 2024-12-27

vtjl10:
1. `validators set` should likely be `validator set` (missing singular form).
2. `State gets updated` could be `The state gets updated` for clarity and consistency.
3. The phrase `ValidatorUpdates are used to add and remove validators to the current set` could be clearer as `ValidatorUpdates are used to add and remove validators from the current set`


melekes:

PR #4762, opened 2025-01-02
chore(e2e): remove `key_type` from ci manifest

melekes:
key_type entry has no actual effect. The reason is that we never implemented the key_type at the node level in the e2e manifest, but rather at the testnet level.


sergio-mena (approved):
Thanks for this!

Issue #3455, opened 2024-07-08
BLS signature aggregation


melekes:
Since PBTS is ready now, we can aggregate the signatures when using BLS (or [ed25519](https://github.com/cometbft/cometbft/issues/1305)), which should both reduce the chain size and speed up verification.

NOTE: This feature will likely require soft upgrades (otherwise, it would require a hard fork).

ADR: https://github.com/cometbft/cometbft/blob/main/docs/references/rfc/tendermint-core/rfc-018-bls-agg-exploration.md

cason:
The proposed use of signature aggregation is to store block commits (`2/3+` Precommit sets for a value), or even during the vote dissemination process?

For the former we should definitely expect some gains, for the second it is harder to understand how can it be valuable.

melekes:
> The proposed use of signature aggregation is to store block commits (2/3+ Precommit sets for a value), or even during the vote dissemination process?

The former.

simonzg:
Hi, @melekes I'm excited about enabling BLS in cometBFT, is there any timeline for this feature?

melekes:
started some work here: https://github.com/cometbft/cometbft/pull/3632

PR #4763, opened 2025-01-02
feat(crypto/bls12381): signature aggregation

melekes:
Refs https://github.com/cometbft/cometbft/issues/3455

TODO:
- [x] API
- [x] Aggregate the signatures for the block ID and the signatures `nil` value
- [x] ~~Rogue key attacks https://hackmd.io/@benjaminion/bls12-381#Rogue-key-attacks~~
  This only applies when accumulating votes on the fly, afaik.

### DIFF
--- a/crypto/bls12381/aggregation.go
+++ b/crypto/bls12381/aggregation.go
@@ -9,16 +9,20 @@ import (
 )
 
 // ErrAggregation is returned when aggregation fails.
-var ErrorAggregation = errors.New("bls12381: failed to aggregate signatures")
+var ErrAggregation = errors.New("bls12381: failed to aggregate signatures")
 
-// For minimal-signature-size operations.
+// For minimal-pubkey-size operations.
+//
+// Changing this to 'minimal-signature-size' would render CometBFT not Ethereum
+// compatible.
 type (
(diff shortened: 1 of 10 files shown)


crypto/bls12381/aggregation.go
```diff
@@ -0,0 +1,47 @@
...
+//
+// Changing this to 'minimal-signature-size' would render CometBFT not Ethereum
+// compatible.
+type (
+	blstAggregateSignature = blst.P2Aggregate
+)
+
+// AggregateSignatures aggregates the given compressed signatures.
+//
+// Does not group-check the signatures.
```
melekes, line 24: Do we need to group-check them?

crypto/bls12381/aggregation.go
```diff
@@ -0,0 +1,47 @@
...
+	if !agProj.AggregateCompressed(sigsToAgg, false) {
+		return nil, ErrAggregation
+	}
+	agSig := agProj.ToAffine()
+	return agSig.Compress(), nil
+}
+
+// VerifyAggregateSignature verifies the given compressed aggregate signature.
+//
+// Group-checks the signature.
```
melekes, line 36: When receiving from the network, we do need to group-check, correct?

PR #4766, opened 2025-01-03
feat(mempool/dog): optimize sending of ResetRoute

### LINKED ISSUE #4617: feat(mempool/dog): Optimize sending of `ResetRoute`
## DOG Protocol Change Proposal

When transaction redundancy in a node is too low (too few duplicates), a node is potentially open for attacks. When the redundancy controller of the protocol notices this, it triggers sending a `ResetRoute` message to a random peer of the node. This peer will then remove the node from any of its `DisabledRoute` lists and allow future traffic to flow unrestricted to the node.  Routes are disabled when a node sends a `HaveTx` message to this peer. 

Currently the choice of the peer to whom a `ResetRoute` message is sent is random. That means that a node can, in theory, send a `ResetRoute` message to a peer that has not previously received a `HaveTx` message, because the protocol does not keep information about to whom a `HaveTx` message has been sent. 

At the moment this did not have any negative effects but it could mean a node takes more time to catch up on transactions. 

We propose an optimization to the protocol, whereby a node remembers to whom `HaveTx` messages have been sent and only send a `ResetRoute` message to those peers. 

cc @hvanz

najeal:
Closes #4617 

#### PR checklist

- [X] Tests written/updated

### DIFF
--- a/mempool/reactor.go
+++ b/mempool/reactor.go
@@ -282,6 +282,7 @@ func (memR *Reactor) TryAddTx(tx types.Tx, sender p2p.Peer) (*abcicli.ReqRes, er
 					memR.Logger.Error("Failed to send HaveTx message", "peer", senderID, "txKey", txKey, "err", err)
 				} else {
 					memR.Logger.Debug("Sent HaveTx message", "tx", txKey.Hash(), "peer", senderID)
+					memR.router.insertRequestedPartialGossip(sender.ID())
 					// Block HaveTx and restart timer, during which time, sending HaveTx is not allowed.
 					memR.redundancyControl.blockHaveTx()
 				}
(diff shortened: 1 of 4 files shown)


najeal:
I suppose @jmalicevic is concerned as she opened the issue and @hvanz as he has been pinged in.

BastienFaivre:
Hi @najeal, I am implementing the DOG algorithm in Rust and came across the same issue. I had the same idea as you for the solution. However, when implementing it, I realized an interesting point. Suppose the following topology:

```
A --- C
 \   / \
  ---   E
 /   \ /
B --- D
```

I'm sorry for my poor drawing skills, but here are the edges: (A, C), (A, D), (B, C), (B, D), (C, E), (D, E).

In this situation, we can imagine node E receiving duplicates for transactions originating from A and B. Now, suppose node E sends two have_tx messages (for both A's and B's transactions) to the same node: either C or D. If we follow the HashSet solution (`map[p2p.ID]struct{}`) you propose to store the peer to which we've sent a have_tx message, **we lose the information that we've sent two have_tx messages to the same peer!**.
Consequently, if we send a reset route message to this peer, we lose its entry in the HashSet, and the local node will think that all routes are enabled. However, this is not the case. Therefore, if we imagine a situation where the redundancy is still too small, the local node cannot take appropriate action to increase it as it thinks all routes are enabled.

So, I suggest keeping track of all the have_tx sent, not just the peers. Something like `map[p2p.ID]uint`, for example, where would be aware of the number of disabled routes on a given peer.

Let me know what you think about this approach!

najeal:
Hi @BastienFaivre ! Sorry for the delay !
I don't see where we loose information. Supposing we have sent to `C` an `HaveTx` message concerning both `A` & `B`. When `E` is picking `C` from the `HashSet` to send him  a `ResetRoute` Message, `C` will reset all the restricting routes pointing to `E` (both from `A` and `B`). There is no reason for `E` to send again a `ResetRoute` to `C`. Hence `C` will be deleted from the `HashSet`.

BastienFaivre:
Thank you for your response, @najeal! I now see better why we reasoned differently.

I think the meaning of an `have_tx` message is misunderstood. From the [documentation](https://github.com/cometbft/cometbft/blob/main/spec/mempool/gossip/dog.md#messages), `have_tx` messages are _origin's specific_, which means that upon receiving an `have_tx` message from a node X, a given node will **ONLY** stop forwarding to X transactions whose origin is the same as the transaction specified in the `have_tx` message, and not block all traffic to X! There is no such a single `have_tx` message concerning multiple nodes (as you mention in your reasoning). Moreover, a `reset_route` message does not reset all the restricting routes to a given node, [but only one at random](https://github.com/cometbft/cometbft/blob/main/spec/mempool/gossip/dog.md#handling-incoming-messages).

Therefore, `have_tx` messages are independent, forcing us to track them all.

Let me know if I missed one of your points!

najeal:
@BastienFaivre in CometBFT, the SenderNode will also continue to delivers some messages when specific route has not been restricted. `HaveTx` concerns one `SourceNode` whereas a `ResetRoute` does not concern a specific Node 👍
You are right the Picking of node to deliver the `ResetRoute` is random 👍

BastienFaivre:
Hi again, @najeal; sorry for the delay.
Indeed, I've made a mistake in the functionality of the `have_tx` message: it is not origin-specific but first-sender-specific, as mentioned in the [doc](https://github.com/cometbft/cometbft/blob/main/spec/mempool/gossip/dog.md#handling-incoming-messages).
However, this does not change the fact that multiple `have_tx` can be sent to a given node, leading us back to the individual tracking again.
I think you misunderstood my last statement about `reset_route`. Picking the node to deliver is not random; we take one from the `map[p2p].ID]uint` you've implemented. When receiving the `reset_route` message, the chosen node will randomly reset one of the blocked routes.

PR #4771, opened 2025-01-04
fix: broken docs links

bocalhky:


light/doc.go
```diff
@@ -118,7 +118,7 @@ as a wrapper, which verifies all the headers, using a light client connected to
 some other node.
 
 See
-https://docs.cometbft.com/main/core/light-client.html
+https://docs.cometbft.com/v1.0/explanation/core/light-client
```
melekes, line 121:
```suggestion
https://docs.cometbft.com/main/explanation/core/light-client
```

melekes (changes requested):
`main` docs should keep pointing to `main`

melekes (approved):

PR #4772, opened 2025-01-04
fix(crypto/bls12381): fix marshaling FilePVKey

melekes:
BEFORE: running `unsafe-reset-all` results in `priv_key` value in `priv_validator_key.json` being cleared. This is due to changes introduced in https://github.com/cometbft/cometbft/pull/3603/ where we changed the struct definition of PrivKey from type PrivKey `[]byte`  to type `PrivKey struct {sk *blst.SecretKey}`. With the new definition, the combat JSON encoder treats this field as hidden since it starts in lowercase. Even though `PrivKey` implements `json.Marshaller` it doesn't pass the check https://github.com/cometbft/cometbft/blob/main/libs/json/encoder.go#L84 because it's the pointer.

AFTER: running `unsafe-reset-all` results in `priv_key` value in `priv_validator_key.json` being the same.


fridrik01 (approved):
lgtm

zrbecker (approved):

PR #4780, opened 2025-01-06
docs: fix typo in abci/client

meetrick:
Changes included:
1.	Typo fix in ReqRes struct comment:
	•	Fixed a typo in the callbackInvoked variable comment in client.go.

---

#### PR checklist

- [ ] Tests written/updated


abci/client/grpc_client.go
```diff
@@ -155,8 +155,8 @@ func (cli *grpcClient) Error() error {
 // NOTE: The callback may receive internally generated flush responses.
 func (cli *grpcClient) SetResponseCallback(resCb Callback) {
 	cli.mtx.Lock()
+	defer cli.mtx.Unlock()
```
melekes, line 158: sorry, but this doesn't improve anything.

melekes (changes requested):

abci/client/grpc_client.go
meetrick: Oh, i got it. I will change this parts

meetrick: @melekes Of course, in this situation, it’s fine not to use defer for unlocking the mutex. However, I believe it’s recommended to use defer whenever possible when unlocking a mutex.

Anyway, thanks for the review. I’ve removed the defer code from the PR.

melekes:

PR #4795, opened 2025-01-12
fix: Fix incorrect error handling Update metricsgen_test.go

ursulabauer:
I noticed that the variable `err` was being checked before it was updated by the `os.Create` call. This could lead to false error detection if a previous function call returned an unhandled error.  

The corrected code now ensures the error check happens right after `os.Create`, accurately reflecting any issues during file creation. Here's the fix:  

```go
outFile := path.Join(dirName, "out.go")
of, err := os.Create(outFile)
if err != nil {
    t.Fatalf("unable to open file %s: %v", outFile, err)
}
```  

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/scripts/metricsgen/metricsgen_test.go
+++ b/scripts/metricsgen/metricsgen_test.go
@@ -54,8 +54,9 @@ func TestFromData(t *testing.T) {
 				t.Fatalf("unable to parse from dir %q: %v", dir, err)
 			}
 			outFile := path.Join(dirName, "out.go")
+			of, err := os.Create(outFile)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


scripts/metricsgen/metricsgen_test.go
```diff
@@ -54,8 +54,9 @@ func TestFromData(t *testing.T) {
 				t.Fatalf("unable to parse from dir %q: %v", dir, err)
 			}
 			outFile := path.Join(dirName, "out.go")
+			of, err := os.Create(outFile)
```
melekes, line 57: So we're creating the file twice now?

PR #4812, opened 2025-01-13
chore: decommission v0.34 and v0.37 release lines

melekes:
which both reached EOL (end of life) https://github.com/cometbft/cometbft/discussions/590


faddat:
Let me throw in a little counterpoint here okay, most of them are security updates.

melekes:
Good point. We should probably leave them for v0.38, but v0.34 and v0.37 have reached EOL https://github.com/cometbft/cometbft/discussions/590

zrbecker (approved):

PR #4813, opened 2025-01-13
fix(crypto/bls12381): modify `Sign`, `Verify`

### LINKED ISSUE #4783: crypto/bls12381: Sign and Verify should use dstMinPk
See https://github.com/berachain/cometbft/pull/11#discussion_r1905146990

In https://github.com/cometbft/cometbft/pull/3860 we forgot to update the `dst` for signing and verifying operations. It should be `dstMinPk,` not `dstMinSig` because, currently, we're optimizing for the minimal public key.

melekes:
to use `dstMinPk`

Closes #4783


melekes:
Even though it's a breaking change, it was decided to backport this into v1.x because no one (to our knowledge) is using the BLS12-381 curve in production yet => therefore it's safe to fix.

zrbecker (approved):

PR #4814, opened 2025-01-13
chore(deps): tag api to the latest commit, not v1.0.0

melekes:


zrbecker:
The way the SDK handles this is the do a go mod replace in their go.mod, and remove it in release versions.

Example: https://github.com/cosmos/cosmos-sdk/blob/main/go.mod#L170

This way local changes will get picked up on main, and they don't need to constantly update pre-release versions.

They also have a go.work.example file instead of committing the go.work for people who need to use the go.work for modules newly under development.

I think we should use a similar strategy here until we have a new cometbft/api release to update to.

melekes:
@zrbecker sounds good to me.

PR #4816, opened 2025-01-13
fix(pbts): hardening tests for overflows in `SynchronyParams`

### LINKED ISSUE #4815: PBTS: consider a cap on the value for the `MessageDelay` synchronous parameter
The Proposed-Based TimesTamps (PBTS) algorithm includes a mechanism to deal with very strict (low) time bounds for proposals, as discussed in https://github.com/cometbft/cometbft/issues/2184. The decision taken after the referred discussion was to increment the value of `MessageDelay`, one of the synchronous parameters adopted by PBTS, by 10% every new round. This would lead to doubling the configured `MessageDelay` after at most 10 rounds, for instance. The value of the `Precision` parameter remains unchanged though.

One limitation of this approach is that the exponential increasing on `MessageDelay` produces, after an important number of rounds, values that are unrealistically high, and some point they don't increase any longer due to integer overflow. For instance, when considering the default value for `SynchronyParams`, with `MessageDelay` set to `15s` (for round `0`), we have:
- At round `10`, `MessageDelay = 38.906136901s`
- At round `58`, `MessageDelay = 1h2m54.565779439s` (1 hour)
- At round `91`,  `MessageDelay = 24h21m4.873095497s` (1 day)
- From round `213`,  `MessageDelay = 2562047h47m16.854775807s` (more than 292 years), since we overflow `int64`

The overflow itself is not a problem, as it is internally handled by Go, in the way it was implemented. The problem is having as a cap the limits of the `time.Duration = int64` type itself. 

We might consider instead a hard-coded maximum value for `MessageDelay`.

cason:
Closes #4815.

The added test units allowed us to catch overflow scenarios in some architectures, in particular `linux/amd64`. The same is not observed in the `arm64` architecture. Sanity checks were added to prevent this from happening.

Further more, `MessageDelay` is now capped at 24hrs, `Precision` - 30 sec.


cason:
So, the overflow tests that work locally fail in the CI...

cason:
I could reproduce the errors running a Docker image in the `linux/amd64` architecture.

melekes:
great work @cason 💯

internal/consensus/state.go
```diff
@@ -1373,7 +1374,10 @@ func (cs *State) timelyProposalMargins() (time.Duration, time.Duration) {
 
 	// cs.ProposalReceiveTime - cs.Proposal.Timestamp >= -1 * Precision
 	// cs.ProposalReceiveTime - cs.Proposal.Timestamp <= MessageDelay + Precision
-	return -sp.Precision, sp.MessageDelay + sp.Precision
+	if sp.MessageDelay+sp.Precision < 0 { // overflow check: need for some architectures
+		return -sp.Precision, time.Duration(math.MaxInt64)
+	}
+	return -sp.Precision, time.Duration(float64(sp.MessageDelay) + float64(sp.Precision))
```
zrbecker, line 1380: Why are we converting to float64 here and then back to a duration?

It seems like this should be sufficient.

```go
	if sp.MessageDelay < math.MaxInt64-sp.Precision {
		return -sp.Precision, time.Duration(math.MaxInt64)
	}
	return -sp.Precision, sp.MessageDelay + sp.Precision
```

types/params.go
```diff
@@ -142,9 +142,13 @@ type SynchronyParams struct {
 // parameters are set or become insufficient to preserve liveness. Refer to
 // https://github.com/cometbft/cometbft/issues/2184 for more details.
 func (sp SynchronyParams) InRound(round int32) SynchronyParams {
+	messageDelay := time.Duration(math.Pow(1.1, float64(round)) * float64(sp.MessageDelay))
+	if messageDelay < 0 { // overflow check: needed for some architectures
```
zrbecker, line 146: Can we guarantee that we don't overflow into a positive range here?

Checking overflow with MaxInt64 seems safer.

```go
func (sp SynchronyParams) InRound(round int32) SynchronyParams {
	scalingFactor := math.Pow(1.1, float64(round))
	var messageDelay time.Duration
	if scalingFactor*float64(sp.MessageDelay) <= float64(math.MaxInt64) {
		messageDelay = time.Duration(scalingFactor * float64(sp.MessageDelay))
	} else {
		messageDelay = time.Duration(math.MaxInt64)
	}
	return SynchronyParams{
		Precision:    sp.Precision,
		MessageDelay: messageDelay,
	}
}
```

zrbecker (changes requested):

internal/consensus/state.go
melekes: https://go.dev/play/p/D36e4asgpur ?

melekes:
```
sum := adaptedSP.MessageDelay + adaptedSP.Precision // overflows
sum := time.Duration(adaptedSP.MessageDelay + adaptedSP.Precision) // overflows
sum := time.Duration(float64(adaptedSP.MessageDelay) + float64(adaptedSP.Precision)) // does not overflow
```

I agree that we need an additional wrapping in the presence of `if` condition. BUT it doesn't hurt to have it, just in case...

types/params.go
melekes: rounds are always inc by 1. but I guess if someone sets message delay to some really big number than it's possible.

internal/consensus/state.go
zrbecker: Sorry, I have the sense wrong on the inequality.

Should be,
```go
	if sp.MessageDelay > math.MaxInt64-sp.Precision {
		return -sp.Precision, time.Duration(math.MaxInt64)
	}
	return -sp.Precision, sp.MessageDelay + sp.Precision
```

types/params.go
zrbecker: The undefined behavior is coming from casting a too large float64 to a int64 which is undefined in Go. The goal here is basically to avoid the overflow entirely, and make explicit the behavior we want.

zrbecker: Here is an example of the issue we are running into. It doesn't have anything to do with time.Duration. It's just an issue of converting between float64 and int64 when `>= float64(math.MaxInt64)`.

Go Example: https://ideone.com/x7VPF1
C Example: https://ideone.com/0aQJhZ

internal/consensus/state.go
cason: So, this is a workaround, not a solution.

The real solution is to add some "safe add duration" method, say, to `types/time` package (typically imported as `cmttime`?

cason: Regarding the conversion to `float64`, the point is that in the `arm64` architecture, this does the trick. Namely, overflows are detected and handled nicely.

The problem is that this is architecture-dependent. In the case of the `linux/amd64` this does not work, for some unknown reason. This is something to report to Golang.

types/params.go
cason: I will repeat my previous comment here:

So, this is a workaround, not a solution.

The real solution is to add some "safe add duration" method, say, to `types/time` package (typically imported as `cmttime`?

cason:
Thank you for the comments, but notice that probably the best solution, which is **not architecture-dependent**, is what I suggest in https://github.com/cometbft/cometbft/issues/4815.

Namely, we should have a maximum value for `MessageDelay` which is hard-coded. It can be, for the sake of simplicity, `math.MaxInt64`. Although this is an absurdly high value that really does not make sense.

Recall that the timestamp of a block `H` should be higher than the timestamp of the previous block `H-1`, as block times are monotonically increasing.

`math.MaxInt64`, when converted into `time.Duration` is of the order of hundreds of years.

If we establish a maximum value for `MessageDelay`, we can just return it when an overflow can happen.

internal/consensus/state.go
cason: > `if sp.MessageDelay > math.MaxInt64-sp.Precision`

This is more precise, indeed, but what in the case that `MessageDelay` has overflowed? It would be negative, this if of course will not be triggered, and we are returning a negative value.

cason:
Another way to go, is to create a PR that points to this one, with destination branch `cason/pbts-params-overflow`, that:

1. Adds a test unit that fails with the current code
2. Change the code so that the test unit passes
3. We merge your solution to this one, then eventually this one to main

In this way, we have commits from everyone involved and a clear history for future archeological work.

types/params.go
```diff
@@ -142,13 +142,29 @@ type SynchronyParams struct {
...
+		return sp
 	}
+
+	spWithMaxMessageDelay := SynchronyParams{
+		Precision:    sp.Precision,
+		MessageDelay: time.Duration(math.MaxInt64),
+	}
+
+	// overflow check: needed for some architectures
+	if round > 0 && int64(sp.MessageDelay) > math.MaxInt64/int64(round) {
```
cason, line 155: `round` can never be zero here.

types/params.go
```diff
@@ -142,13 +142,29 @@ type SynchronyParams struct {
 // parameters are set or become insufficient to preserve liveness. Refer to
 // https://github.com/cometbft/cometbft/issues/2184 for more details.
 func (sp SynchronyParams) InRound(round int32) SynchronyParams {
-	messageDelay := time.Duration(math.Pow(1.1, float64(round)) * float64(sp.MessageDelay))
-	if messageDelay < 0 { // overflow check: needed for some architectures
-		messageDelay = time.Duration(math.MaxInt64)
+	if round == 0 {
```
cason, line 145: We can be really safe here and test `round <= 0`.

types/params.go
```diff
@@ -142,13 +142,29 @@ type SynchronyParams struct {
...
+	}
+
+	// overflow check: needed for some architectures
+	if round > 0 && int64(sp.MessageDelay) > math.MaxInt64/int64(round) {
+		return spWithMaxMessageDelay
+	}
+
+	// overflow check 2: check exp overflow
+	f := math.Pow(1.1, float64(round)) * float64(sp.MessageDelay)
+	if math.IsInf(f, 1) || math.IsNaN(f) || f > math.MaxInt64 {
```
cason, line 161: How can it be `NaN`?

types/validator_set_test.go
```diff
@@ -1717,7 +1716,7 @@ func TestValidatorSet_AllKeysHaveSameType(t *testing.T) {
 			sameType: false,
 		},
 		{
-			vals:     NewValidatorSet([]*Validator{NewValidator(secp256k1eth.GenPrivKey().PubKey(), 200), NewValidator(secp256k1.GenPrivKey().PubKey(), 200)}),
+			vals:     NewValidatorSet([]*Validator{NewValidator(ed25519.GenPrivKey().PubKey(), 200), NewValidator(secp256k1.GenPrivKey().PubKey(), 200)}),
```
cason, line 1719: Thank for this :)

We needeed to pass a build flag to test this package....

cason (commented):
Thank you for the changes, the code looks better now.

I don't see how some conditions that are covered now can be observed. Namely, test cases where they are triggered.

But possibly is better to be safe than sorry.

cason:
In addition, the log message produced by the consensus protocol does not contain negative values for  the `upper_bound`, such as:

```
2025-01-13T20:14:08.070 INF Prevote step: Proposal is not timely; prevoting nil module=consensus height=2 round=0 timestamp=2025-01-13T19:14:08.159023Z receive_time=2025-01-13T19:14:08.070043Z timestamp_difference=-88.98ms lower_bound=-1ms upper_bound=-2562047h47m16.85377536s
```

That now are corrected to display the maximum value that `time.Duration` can get:

```
2025-01-13T20:14:21.562 INF Prevote step: Proposal is not timely; prevoting nil module=consensus height=2 round=0 timestamp=2025-01-13T19:14:21.651686Z receive_time=2025-01-13T19:14:21.562641Z timestamp_difference=-89.045ms lower_bound=-1ms upper_bound=2562047h47m16.854775807s
```

With the addition of the `TestPBTSTooFarInTheFutureProposalOverflow` test unit.

cason:
I removed the verbose comment regarding logs from the PR description, that was updated to inform about the fixes on the overflow conditions.

cason:
Should we add some overflow test on `ValidateBasic()`?

https://github.com/cometbft/cometbft/blob/7ef4e4397ede2779afc152fa2fb67c4f86af1bff/types/params.go#L298-L305

cason:
By adding the following test unit:

```
$ git diff                                                                                                                                                                                                                                          [cason/pbts-params-overflow] 
diff --git a/types/params_test.go b/types/params_test.go
index e1fb9f506..009d03c01 100644
--- a/types/params_test.go
+++ b/types/params_test.go
@@ -305,6 +305,17 @@ func TestConsensusParamsValidation(t *testing.T) {
                        }),
                        valid: false,
                },
+               {
+                       name: "overflow synchrony",
+                       params: makeParams(makeParamsArgs{
+                               blockBytes:   1,
+                               evidenceAge:  2,
+                               precision:    time.Duration(math.MaxUint64),
+                               messageDelay: time.Duration(math.MaxUint64),
+                               pbtsHeight:   1,
+                       }),
+                       valid: false,
+               },
                // pbts enable height
                {
                        name: "pbts height -1",
```

I observe:

```
$ go test -v -run TestConsensusParamsValidation                                                                                                                                                                                                     [cason/pbts-params-overflow] 
# github.com/cometbft/cometbft/types [github.com/cometbft/cometbft/types.test]
./params_test.go:313:33: constant 18446744073709551615 overflows "time".Duration
./params_test.go:314:33: constant 18446744073709551615 overflows "time".Duration
FAIL	github.com/cometbft/cometbft/types [build failed]
```

Interesting.

cason:
I wonder if with:

```
$ git diff                                                                                                                                                                                                                                          [cason/pbts-params-overflow] 
diff --git a/types/params_test.go b/types/params_test.go
index e1fb9f506..009d03c01 100644
--- a/types/params_test.go
+++ b/types/params_test.go
@@ -305,6 +305,17 @@ func TestConsensusParamsValidation(t *testing.T) {
                        }),
                        valid: false,
                },
+               {
+                       name: "overflow synchrony",
+                       params: makeParams(makeParamsArgs{
+                               blockBytes:   1,
+                               evidenceAge:  2,
+                               precision:    time.Duration(math.MaxUint64),
+                               messageDelay: time.Duration(math.MaxUint64),
+                               pbtsHeight:   1,
+                       }),
+                       valid: false,
+               },
                // pbts enable height
                {
                        name: "pbts height -1",
```

We should accept or not the produced consensus params:

```
$ go test -v -run TestConsensusParamsValidation                                                                                                                                                                                                     [cason/pbts-params-overflow] 
=== RUN   TestConsensusParamsValidation
    params_test.go:377: 
        	Error Trace:	/Users/cason/Workspace/cometbft/types/params_test.go:377
        	Error:      	An error is expected but got nil.
        	Test:       	TestConsensusParamsValidation
        	Messages:   	expected error for non valid params, test: 'overflow synchrony'
--- FAIL: TestConsensusParamsValidation (0.00s)
```

zrbecker:
Since I posted this on some outdated threads going to reiterate here.

There are two different types of overflows happening, and we should avoid them entirely. We should not allow the overflows to happen, and then try to correct afterwords.

The first type of overflow is an integer overflow `a + b` overflows if `a > MathInt64 - b` for b > 0.

The second type of overflow is float64 to int64 conversion overflow `int64(f)` overflows if `f >= MaxInt64`. (`time.Duration(f)` is the same as `int64(f)`. The behavior of this overflow is different between architectures, ARM64 returns MaxInt64, and AMD64 returns MinInt64 (https://www.felixcloutier.com/x86/cvttsd2si).

The two solutions I recommend, neither are workarounds.

1) Always check arithmetic and conversions for overflows.
2) Bound the value of MessageDelay to be much lower to avoid the issue of overflows.

(1) is essentially what we are trying to do right now, but we seem to be checking a bunch of auxiliary stuff that is unnecessary. And it's not entirely clear to me what other arithmetic is being done that can cause addition overflow issues.

(2) I think if we bound MessageDelay to something small like 1 day, we will avoid the overflow question entirely. `time.Duration(MaxInt64)` is nearly 300 years.

@melekes I think solution (2) is what we should go with to avoid overflow issues entirely. We can then add to `ValidateBasic` a check that MessageDelay is below that 1 day limit.

melekes:
> (2) I think if we bound MessageDelay to something small like 1 day, we will avoid the overflow question entirely. time.Duration(MaxInt64) is nearly 300 years. This is a more typical and less complex solution.

This sounds reasonable to me 👍

melekes:
It should be good now @zrbecker

types/params.go
```diff
@@ -141,10 +154,27 @@ type SynchronyParams struct {
...
 // parameters are set or become insufficient to preserve liveness. Refer to
 // https://github.com/cometbft/cometbft/issues/2184 for more details.
+//
+// There's a cap (MaxMessageDelay) on the MessageDelay to prevent overflow.
 func (sp SynchronyParams) InRound(round int32) SynchronyParams {
+	if round <= 0 {
+		return sp
+	}
+
+	d := time.Duration(math.Pow(1.1, float64(round)) * float64(sp.MessageDelay))
```
zrbecker, line 164: This here still can cause an overflow if `round` is large enough. You need to check the duration before you cast it to an int64, i.e., time.Duration.

Again, we should not overflow, and then try and fix it. We should just prevent the overflows.

If you do this it will be safe and not overflow, you can also remove the if statement,

```
d := time.Duration(math.Min(
  float64(MaxMessageDelay),
  math.Pow(1.1, float64(round)) * float64(sp.MessageDelay),
))
```

zrbecker (commented):
Everything looks good here except one comment. Whenever you cast from flaot64 to int64 if you think it can overflow, you should just prevent the overflow, here I recommend doing it by clamping the value with `math.Min`.

types/params.go
```diff
@@ -142,9 +142,29 @@ type SynchronyParams struct {
...
+	}
+
+	// overflow check: needed for some architectures
+	if int64(sp.MessageDelay) > math.MaxInt64/int64(round) {
+		return spWithMaxMessageDelay
+	}
+
+	// overflow check 2: check exp overflow
+	f := math.Pow(1.1, float64(round)) * float64(sp.MessageDelay)
+	if math.IsInf(f, 1) || f > math.MaxInt64 {
```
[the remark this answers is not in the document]
melekes: I think you're reviewing old version of the code.

types/params.go
```diff
@@ -142,9 +142,29 @@ type SynchronyParams struct {
...
+		return sp
+	}
+
+	spWithMaxMessageDelay := SynchronyParams{
+		Precision:    sp.Precision,
+		MessageDelay: time.Duration(math.MaxInt64) - sp.Precision,
+	}
+
+	// overflow check: needed for some architectures
+	if int64(sp.MessageDelay) > math.MaxInt64/int64(round) {
```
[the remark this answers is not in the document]
melekes: Again, the check was removed earlier.

types/params.go
melekes: [`9183157` (#4816)](https://github.com/cometbft/cometbft/pull/4816/commits/918315707ed9185b372a3cab3375b31935bd50d1)

types/params.go
```diff
@@ -283,6 +303,9 @@ func (params ConsensusParams) ValidateBasic() error {
 			return fmt.Errorf("synchrony.Precision must be greater than 0. Got: %d",
 				params.Synchrony.Precision)
 		}
+		if params.Synchrony.MessageDelay > math.MaxInt64-params.Synchrony.Precision {
```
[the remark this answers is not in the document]
melekes: Again, the check was removed earlier.

zrbecker (approved):

types/params.go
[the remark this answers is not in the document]
zrbecker: This comment was from yesterday, was address with other things it looks like.

types/params.go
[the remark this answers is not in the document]
zrbecker: That comment was from yesterday before the code was changed. Was already addressed.

zrbecker:
Everything looks good.

Looks like there was some confusion as some of your responses were to my comments from the other day before the code was changed. Those all look addressed to me, so everything seems good.

cason:
Great work on this, I really appreciated the option for a "final" and architecture independent solution.

Do you mind updating, with the existence of maximum values for those parameters:

- https://docs.cometbft.com/v1.0/explanation/core/proposer-based-timestamps
- https://github.com/cometbft/cometbft/blob/main/spec/core/data_structures.md#synchronyparams
- https://github.com/cometbft/cometbft/blob/main/spec/consensus/proposer-based-timestamp/pbts-sysmodel.md#pbts-msg-delay-adaptive0

melekes:
@cason opened https://github.com/cometbft/cometbft/issues/4893

Issue #4735, opened 2024-12-28
Validation Error When Broadcasting Single-Value Transactions in kvstore Example


itsmehdiabdi:
### Setup

**CometBFT version** : 1.0.0

**Have you tried the latest version**: yes

**ABCI app** : built-in

**Environment**:
- **OS** (e.g. from /etc/os-release): 
```
PRETTY_NAME="Ubuntu 22.04.4 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.4 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```
- **Install tools**: Installed from source
- **Others**: N/A
- **node command runtime flags**: Default runtime flags for the kvstore example.

### Config
No modifications to the default configuration.

### What happened?

Attempting to broadcast a transaction using the following command:

```bash
curl -s 'localhost:26657/broadcast_tx_commit?tx="abcd"'
```

results in an error:

```json
{"jsonrpc":"2.0","id":-1,"error":{"code":-32603,"message":"Internal error","data":"broadcast error on transaction validation: tx 88D4266FD4E6338D13B845FCF289579D209C897823B9217DA3E161936F031589 is invalid: code=2, data=, log='', codespace=''}}
```

However, using the second broadcasting method, such as:

```bash
curl -s 'localhost:26657/broadcast_tx_commit?tx="name=satoshi"'
```

works fine, storing name as the key and satoshi as the value.

### What did you expect to happen?

I expected the transaction `curl -s 'localhost:26657/broadcast_tx_commit?tx="abcd"'` to store abcd as both the key and the value in the kvstore, as described in the documentation.

### How to reproduce it

1. Start the kvstore application using CometBFT.
2. Run the following command:
```bash
curl -s 'localhost:26657/broadcast_tx_commit?tx="abcd"'
```

3. Observe the error response.

### Logs
```json
[...]

PR #4824, opened 2025-01-18
docs: Removed the outdated clause where we store only data, updated KV-store doc

faust403:
…re doc

Closes https://github.com/cometbft/cometbft/issues/4735

### DIFF
--- a/docs/guides/app-dev/getting-started.md
+++ b/docs/guides/app-dev/getting-started.md
@@ -156,6 +156,40 @@ Note the value in the result (`c2F0b3NoaQ==`); this is the base64-encoding of th
 echo -n 'c2F0b3NoaQ==' | base64 -d
 ```
 
+Sometimes we only need to know if value exists and that's the point where else you can use KV-store. In such a case we only need to specify that key equals key, for example:
+
+```sh
+curl -s 'localhost:26657/broadcast_tx_commit?tx="value=value"'
+```
+
+And then when query it
+
+```sh
+curl -s 'localhost:26657/abci_query?data="value"'
+```
+
+You get
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": -1,
+  "result": {
+    "response": {
+      "code": 0,
+      "log": "exists",
+      "info": "",
+      "index": "0",
+      "key": "dmFsdWU=",
+      "value": "dmFsdWU=",
+      "proofOps": null,
+      "height": "68935",
+      "codespace": ""
+    }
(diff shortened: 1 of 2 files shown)


docs/guides/app-dev/getting-started.md
```diff
@@ -118,82 +118,79 @@ tool like [jq](https://stedolan.github.io/jq/) or `json_pp`.
...
+```
+
+Now if we query for `name`, we should get `satoshi`, or `c2F0b3NoaQ==`
+in base64:
+
+```sh
+curl -s 'localhost:26657/abci_query?data="name"'
 ```
 
-Note the single quote (`'`) around the url, which ensures that the
```
melekes, line 124: Why remove this?

docs/guides/app-dev/getting-started.md
```diff
@@ -118,82 +118,79 @@ tool like [jq](https://stedolan.github.io/jq/) or `json_pp`.
 Now let's send some transactions to the kvstore.
 
 ```sh
-curl -s 'localhost:26657/broadcast_tx_commit?tx="abcd"'
+curl -s 'localhost:26657/broadcast_tx_commit?tx="name=satoshi"'
+```
+
```
melekes, line 123: I'd add smth like 

```suggestion

This will store the value `satoshi` under the key `name`.
```

docs/guides/app-dev/getting-started.md
```diff
@@ -118,82 +118,79 @@ tool like [jq](https://stedolan.github.io/jq/) or `json_pp`.
...
 
-We can confirm that our transaction worked and the value got stored by
-querying the app:
+Note the value in the result (`c2F0b3NoaQ==`); this is the base64-encoding of the ASCII of `satoshi`. You can verify this through a simple Bash command:
+
+```bash
+echo -n 'c2F0b3NoaQ==' | base64 -d
+```
+
+Sometimes we only need to know if value exists and that's the point where else you can use KV-store. In such a case we only need to specify that key equals key, for example:
```
melekes, line 159: So, if we want to search by value, we assign this value to a key and search using this key? That seems like an incorrect example, IMHO.

faust403: According to the doc about abci-cli I thought it works so. If it differs, could you help provide more details about it? Before this moment I thought that if it's a Merkle Tree then it should construct a pair (key, value) where 2 values are equal (so it's enough to ensure if values exist)

PR #4886, opened 2025-01-27
docs: fix typo in documentation in debugging.md

mdqst:
#### Description 
I noticed a typo in the documentation where "stage" was used instead of "state" in the following sentence:  
> `inspect` allows operators to query a read-only view of the stage.  

This should correctly refer to the **state** (state store) in CometBFT.

I've updated it to:  
> `inspect` allows operators to query a read-only view of the **state**.  

---

#### PR checklist

- [x] Tests written/updated


zrbecker:
Thanks for the contribution. However, this is a very small contribution, and we generally prefer to accept contributions that address the issue across the documentation as oppose to single typo per PR.

PR #4898, opened 2025-01-31
fix: fixed argument handling in shell script

mdqst:
I’ve corrected an issue where the `$*` variable in the shell script was passing all arguments as a single string.
This caused problems when handling names with spaces or multiple words.
Now, `$@` is used instead to ensure arguments are passed as separate parameters, which resolves the issue.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/scripts/mockery_generate.sh
+++ b/scripts/mockery_generate.sh
@@ -2,4 +2,4 @@
 #
 # Invoke Mockery v2 to update generated mocks for the given type.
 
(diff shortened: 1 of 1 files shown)


zrbecker:
Thanks for the contribute. We need to reevaluate which of these script are important, and remove the ones that are no longer needed or are unused to do errors. So will not be accepting small updates to these scripts at the moment.

PR #4906, opened 2025-02-03
merge commit from fork

melekes:
lower than what was previously reported

### DIFF
--- a/blocksync/pool.go
+++ b/blocksync/pool.go
@@ -361,11 +361,21 @@ func (pool *BlockPool) SetPeerRange(peerID p2p.ID, base int64, height int64) {
 
 	peer := pool.peers[peerID]
 	if peer != nil {
+		if base < peer.base || height < peer.height {
+			pool.Logger.Info("Peer is reporting height/base that is lower than what it previously reported",
+				"peer", peerID,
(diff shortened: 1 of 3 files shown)


zrbecker (approved):

blocksync/pool.go
```diff
@@ -361,11 +361,21 @@ func (pool *BlockPool) SetPeerRange(peerID p2p.ID, base int64, height int64) {
...
 	peer := pool.peers[peerID]
 	if peer != nil {
+		if base < peer.base || height < peer.height {
+			pool.Logger.Info("Peer is reporting height/base that is lower than what it previously reported",
+				"peer", peerID,
+				"height", height, "base", base,
+				"prevHeight", peer.height, "prevBase", peer.base)
+			// RemovePeer will redo all requesters associated with this peer.
+			pool.removePeer(peerID)
+			pool.banPeer(peerID)
```
yihuang, line 371: do we have to ban the peer, it's possible that the peer node is just restored from an old snapshot?

melekes: It's possible, yes, but very rare in practice. Can you think of another solution to the problem?

PR #4868, opened 2025-01-23
feat(storage)!: remove `cometbft-db`

melekes:
Closes #4486.

PR #4915, opened 2025-02-03
chore: remove `db_backend` config option and CLI argument

melekes:
also

- add a changelog entry
- remove `experimental-compact-goleveldb` command
- fix rpc and light client tests https://github.com/cometbft/cometbft/pull/4868#issuecomment-2611646305


node/node.go
```diff
@@ -711,6 +706,11 @@ func (n *Node) OnStop() {
 			n.Logger.Error("Pprof HTTP server Shutdown", "err", err)
 		}
 	}
+	if n.indexerService != nil {
```
zrbecker, line 709: Why was the ordering here changed?

zrbecker (approved):

node/node.go
melekes: The test was failing because of it. There's an implicit assumption that DB is always closed after the indexer, which was not the case earlier.

melekes: I'd add a comment.

state/txindex/indexer_service.go
```diff
@@ -74,7 +77,14 @@ func (is *IndexerService) OnStart() error {
 				batch := NewBatch(numTxs)
 
 				for i := int64(0); i < numTxs; i++ {
-					msg2 := <-txsSub.Out()
+					var msg2 cmtpubsub.Message
```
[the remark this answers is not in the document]
melekes: No, it was a general improvement. Not related to this PR.

PR #4925, opened 2025-02-06
fix: backport security fixes from `v0.37.15`

michaelfig:
Manually backport the fixes released in [v0.37.15](https://github.com/cometbft/cometbft/compare/v0.37.14..v0.37.15).

This PR may be relevant to teams that are upgrading from v0.34.x, but must also apply these fixes as soon as possible to their production chains.

---

#### PR checklist

- [x] Tests written/updated


zrbecker:
Thanks for the contribution. However, since v0.34.x is at end of life we are not backporting security fixes to it.

If your team depends on v0.34.x we recommend maintaining a fork and using a replace in your go.mod to redirect go's module system to use that version.

We are recommending all teams to upgrade to at least v0.38.x to keep up with the latest security patches.

faddat:
While I support the EOL of 0.34.x, I think that we should merge this PR.  

We should be reality-based and keep in mind that about half of the cosmos ecosystem uses v0.34.x, still, today, in prod, and that any of those teams can cause an unexpected issue in their fork, increasing attack surface area by the number of forks and the economic value on those chains.  

So by far the best plan is to continue to provide support for security patches for 0.34.x until it is fewer than 1/3rd of cosmos networks, since that's how consensus works anyway.

PR #4946, opened 2025-02-10
ci: remove govulncheck

### LINKED ISSUE #4945: proposal: remove govulncheck CI job
Undoubtedly, [govulncheck](https://go.dev/blog/govulncheck) is helpful for node operators and app devs who want to ensure their apps are secure. But it doesn't make much sense for CometBFT - a library (even if it's run as a separate binary and ABCI communicates with it via sockets).

`govulncheck` forces CometBFT to bump the min Go version (MSRV), which applies upstream pressure on app devs and node operators. The argument is that node operators and app devs should themselves watch over Go vulnerabilities and bump the Go version accordingly.

melekes:
Closes #4945


melekes:
`govulncheck` must be removed from the `required` jobs list.

zrbecker (approved):

melekes:
> `govulncheck` must be removed from the `required` jobs list by the repo admin.

cc @zrbecker

Zygimantass:
was it possible to exclude the MSRV check and keep all of the other ones?

melekes:
> was it possible to exclude the MSRV check and keep all of the other ones?

nope.

PR #4953, opened 2025-02-13

maximevtush:
#### PR checklist

- [ ] Tests written/updated


zrbecker:
Thanks for the contribution, but I don't think this is all possible fixes.

If we are going to accept something it needs to be a comprehensive solution. Ideally, we need to add a system that does some sort of basic spell checking at least in documentation.

I don't think it's a good use of our time to review spell check fixes a handful at a time.

PR #4978, opened 2025-02-24
fix(statesync): if statesync fails, handshake with the app

### LINKED ISSUE #4977: statesync: if statesync fails or reaches max_discovery_time, it must handshake with the app
Otherwise, the node would use the genesis state, which would be incorrect.

discovered in https://github.com/cometbft/cometbft/actions/runs/13426543714/job/37510433419

Introduced in https://github.com/cometbft/cometbft/pull/3965

melekes:
before starting blocksync.

Closes #4977

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -653,50 +653,60 @@ func (n *Node) OnStart() error {
 			return ErrSwitchStateSync
 		}
 
-		newState, err := startStateSync(n.stateSyncReactor, n.stateSyncProvider,
+		stateCh, errc, err := startStateSync(n.stateSyncReactor, n.stateSyncProvider,
 			n.config.StateSync, n.stateStore, n.blockStore, n.stateSyncState, n.config.Storage.ExperimentalKeyLayout)
-
-		switch err {
-		case nil:
-			n.Logger.Info("statesync complete. switching to blocksync with new state")
-
-			err = bcR.SwitchToBlockSync(newState)
-			if err != nil {
-				return ErrStartBlockSync{Err: err}
-			}
-		case statesync.ErrNoSnapshots:
-			n.Logger.Info("no snapshots found. switching to blocksync with genesis state")
-
-			const handshakeTimeout = 5 * time.Second
-			ctx, cancel := context.WithTimeout(context.Background(), handshakeTimeout)
-			defer cancel()
-			if err := doHandshake(ctx, n.stateStore, n.stateSyncState, n.blockStore,
-				n.stateSyncGenDoc, n.eventBus, n.appInfoResponse, n.proxyApp, n.Logger); err != nil {
-				return ErrHandshake{Err: err}
-			}
-
-			// Reload the state. It will have the Version.Consensus.App set by the
-			// Handshake, and may have other modifications as well (ie. depending on
-			// what happened during block replay).
-			state, err := n.stateStore.Load()
-			if err != nil {
-				return sm.ErrCannotLoadState{Err: err}
-			}
-
-			err = bcR.SwitchToBlockSync(state)
-			if err != nil {
-				return ErrStartBlockSync{Err: err}
-			}
-		default:
+		if err != nil {
 			return ErrStartStateSync{Err: err}
 		}
-	}
 
-	////////////////////////////
-	// FREE GENESIS DOC MEMORY!
-	n.stateSyncGenDoc = nil
-	n.appInfoResponse = nil
-	////////////////////////////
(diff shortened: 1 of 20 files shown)


melekes:
> Looks like there are a number of CodeQL messages here regarding logging the error object. Do these sometimes contain sensitive info?

no

node/node.go
```diff
@@ -636,17 +646,66 @@ func (n *Node) OnStart() error {
...
+		stateCh, errc, err := startStateSync(n.stateSyncReactor, n.stateSyncProvider,
+			n.config.StateSync, n.stateStore, n.blockStore, n.stateSyncState, n.config.Storage.ExperimentalKeyLayout)
 		if err != nil {
 			return ErrStartStateSync{Err: err}
 		}
+
+		// Wait until statesync is over or errors, and start blocksync in a separate goroutine.
+		go func() {
+			defer func() {
+				// FREE GENESIS DOC MEMORY!
```
melekes, line 665: this line, I'm not proud, but don't think there's another way.

node/node.go
```diff
@@ -636,17 +646,66 @@ func (n *Node) OnStart() error {
...
-			n.config.StateSync, n.stateStore, n.blockStore, n.stateSyncGenesis, n.config.Storage.ExperimentalKeyLayout)
+
+		stateCh, errc, err := startStateSync(n.stateSyncReactor, n.stateSyncProvider,
+			n.config.StateSync, n.stateStore, n.blockStore, n.stateSyncState, n.config.Storage.ExperimentalKeyLayout)
 		if err != nil {
 			return ErrStartStateSync{Err: err}
 		}
+
+		// Wait until statesync is over or errors, and start blocksync in a separate goroutine.
+		go func() {
```
jmalicevic, line 663: I was a bit unsure whether the parallelism here and the channels open a way for introducing errors but I can't think of a scenario atm.
The things I was trying to understand:
1. Can we have `state` returned on the `stateCh` before the Syncing function sets the syncing flag to false (which should not happen after we switch to blocksync) - but I think it is not possible. 
2. Could the note somehow setup wrongly, but again I am not seeing it with this code. 

Writing this just in case something clicks on your end when reading this :D

test/e2e/runner/setup.go
```diff
@@ -324,7 +324,6 @@ func MakeConfig(node *e2e.Node) (*config.Config, error) {
 		if len(cfg.StateSync.RPCServers) < 2 {
 			return nil, errors.New("unable to find 2 suitable state sync RPC servers")
 		}
-		cfg.StateSync.MaxDiscoveryTime = 30 * time.Second
```
jmalicevic, line 327: Are we sure that by removing this we are testing all possible corner cases of statesync now?

node/node.go
```diff
@@ -636,17 +646,66 @@ func (n *Node) OnStart() error {
...
+			case newState := <-stateCh:
+				n.Logger.Info("statesync complete. switching to blocksync with new state")
+
+				err = bcR.SwitchToBlockSync(newState)
+				if err != nil {
+					n.Logger.Error("failed to switch to blocksync", "err", err)
+					return
+				}
+			case err := <-errc:
+				n.Logger.Warn("statesync failed. switching to blocksync with genesis state", "err", err)
```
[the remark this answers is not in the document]
melekes: Don't see why we can't try blocksync if statesync fails. We log the error. I.e. logging the error and trying blocksync is better than ONLY logging the error (note Comet doesn't exit upon statesync error).

test/e2e/runner/setup.go
melekes: Do you know why it was introduced in the first place?

node/node.go
melekes: I will change it to wait. I just noticed that we're sleeping in `Node.Start` 😄

jmalicevic: Where is the sleep?

zrbecker (approved):

node/node.go
jmalicevic: But is this not only an exception and an error case while in the regular case tehre is no sleep?

melekes: True. Still avoiding extra goroutines gives app developers more control over the `Start` function.

jmalicevic: Tbh I would avoid changing the behaviour that was there for some years w/o being sure it hsa no implications. For example, someone expecting `node.Start()` to return w/o statesync might be doing other things that could facilitate StateSync execution. 

So I would put statesync in a separate go routine as it was.

PR #4987, opened 2025-02-27
feat(libs): restore os and rand, add deprecation notices

### LINKED ISSUE #4986: Restore libs/os and libs/rand packages
The libs/os and libs/rand packages were previously moved to internal/, making them inaccessible to external consumers. This issue tracks restoring them to libs/ while marking them as deprecated to provide a transition period before removal in a future release.

zrbecker:
Closes #4986 

Summary
Restores libs/os and libs/rand from internal/ to temporarily maintain backward compatibility. Adds deprecation notices to indicate their removal in a future release.

Changes
- Restored libs/os and libs/rand from internal/ (just did a copy)
- Added Deprecated: comments to both packages (only added comments)

#### PR checklist

- [x] Tests written/updated


melekes (approved):
👍

PR #5014, opened 2025-03-03
ci(testapp-docker): release two images, not one

melekes:
Previously, we would only push a single arm64 image. Evidence: https://hub.docker.com/r/cometbft/e2e-node/tags


zrbecker (approved):

melekes:
https://github.com/cometbft/cometbft/actions/runs/13671672657/job/38223317330 ✅

PR #5025, opened 2025-03-10
fix(rpc/test): misuse wrong err variable

suiyuan1314:
`err` has been checked before with `require.NoError(t, err)`, actually we are not aim to check `err` but `msg.Error` :)


melekes (approved):
👍

PR #5038, opened 2025-03-22
refactor(all): replacing []byte(fmt.Sprintf...) by fmt.Appendf(nil, ...)

lfgtwo:
---

#### PR checklist

replacing `[]byte(fmt.Sprintf...)` by `fmt.Appendf(nil, ...)`, added in go1.19;

- [ ] Tests written/updated


melekes:
I don't see how this is better than the previous version.

lfgtwo:
> I don't see how this is better than the previous version.

While browsing the code, I noticed the gopl hints.
image


ref: https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize#fmtappendf

PR #5043, opened 2025-03-24
refactor: use slices.Contains to simplify code

xinhangzhou:
There is a [new function](https://pkg.go.dev/slices@go1.21.0#Contains) added in the go1.21 standard library, which can make the code more concise and easy to read.

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/p2p/peer.go
+++ b/p2p/peer.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"reflect"
 	"runtime/debug"
+	"slices"
 	"time"
 
 	"github.com/cosmos/gogoproto/proto"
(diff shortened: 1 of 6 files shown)


fcecin:
I think it's better to just delete the `intInSlice` functions and where they are called just use `slices.Contains` directly.

The `HasChannel` fix is OK.

xinhangzhou:
> I think it's better to just delete the `intInSlice` functions and where they are called just use `slices.Contains` directly.
> 
> The `HasChannel` fix is OK.

Thanks. Modified.

fcecin (approved):

lfgtwo:
https://github.com/cometbft/cometbft/pull/5039 already checked all

xinhangzhou:
> ** ** approved

Thanks for your  approved.

PR #5102, opened 2025-04-28
refactor: use slices.Contains to simplify code

CoolCu:
---

#### PR checklist

- [ ] Tests written/updated

Use a more modern writing style.

https://pkg.go.dev/slices@go1.21.0#Contains

https://pkg.go.dev/slices@go1.21.0#ContainsFunc


zrbecker:
This seems unnecessary. The any and not functions don't look like they are even being used, so maybe we should delete them. However, this is exceptionally low on our priority to address.

PR #5156, opened 2025-06-03
[adr-118]: fix link to cache

kks-code:
if my change is not correct - let me know! i will fix it.


zrbecker:
We have merged a bunch of PRs that fix 1-2 links in the past few days. I am more than happy to merge corrections, and recognize that there is a task worth doing to correct the links in the documentation, however, I am resistant to merge the links one at a time when the actual task should be to scan the entire documentation for broken links, and correct all of the broken links.

I am open to merging a PR that solves the more general problem. Otherwise it will need to wait until we have bandwidth to prioritize it.

PR #5138, opened 2025-05-27
feat: extend `privval` to support signing raw bytes

rach-id:
Added a new method to the PrivValidator interface that enables validators to securely sign arbitrary raw bytes.

Implements https://github.com/cometbft/cometbft/issues/5126 in v0.38.x

I will open a PR for main after this one gets merged or when it's reviewed and the changes are agreed upon.

#### PR checklist

- [ ] Tests written/updated
- [ ] Changelog entry added in `.changelog` (we use [unclog](https://github.com/informalsystems/unclog) to manage our changelog)
- [ ] Updated relevant documentation (`docs/` or `spec/`) and code comments

PR #5232, opened 2025-06-23
feat: add types for signing raw digests

rach-id:
As part of forward porting https://github.com/cometbft/cometbft/pull/5138 to v1, this PR adds the necessary types in v1 
before opening the PR that adds their implementation.

#### PR checklist

- [ ] Tests written/updated


melekes (approved):
LGTM

zrbecker:
This looks like it's not adding types, it is renaming types and fields, am I misunderstanding?

rach-id:
> This looks like it's not adding types, it is renaming types and fields, am I misunderstanding?

Could be considered both. In fact, `main` had an implementation of raw bytes signing that was incomplete and wouldn't be possible to implement by downstream key signers like horcrox. This PR updates the types to reflect what is in the PR to `v0.38.x` and provide the correct types that can be used.

Happy to update the PR name to `feat: update the types for signing raw digests` if you like :D

zrbecker:
We released v2 so not to introduce backwards incompatible changes. We are having an internal discussion regarding how we want to incorporate external contributions, but since we just put a huge effort into upgrading cosmos-sdk to new proto formats, I am pretty hesitant to accept further breaking changes to protos.

rach-id:
Agreed about the breaking changes. But currently for main, I see the existing protos for signing raw bytes as invalid. So we're not really adding a new feature that breaks, but fixing a bug that requires a breaking change. With the current protos, downstream signers won't be supported because they require a chainID in the request, which is not provided in this interface. Some gymnastics can be done to get it from the payload etc. But I guess the proposed changes fixes this in a clean way.

rach-id:
@tony-iqlusion 100%. If that's something the comet team wants so this PR gets merged, I am down. cc @aljo242

swift1337:
Hey! Closing due to inactivity.

PR #5244, opened 2025-06-27
refactor:Opt LRUTxCache.Remove to reduce lock contention and map access

songgaoye:
---
Calculating key := tx.Key() outside the c.mtx.Lock() in the LRUTxCache.Remove method is feasible and can slightly optimize performance by reducing the time spent inside the critical section. The tx.Key() method (defined in tx.go) computes a SHA-256 hash of the transaction, which is a pure computation that doesn't access shared state or depend on the cache's internal data structures.

Streamlined the Remove method in LRUTxCache by combining the map lookup and deletion into a single operation using the map's return values. This reduces redundant operations and improves performance, especially when the transaction is not in the cache. The behavior remains unchanged, ensuring thread-safety and correct removal of transactions from both the cache map and linked list.
#### PR checklist

- [ ] Tests written/updated


songgaoye:
@melekes Hi sir, could please help me review this PR

mattac21 (approved):

PR #5309, opened 2025-08-25
fix(internal/consensus/reactor.go): reject oversized proposals

arsushi:
---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/.changelog/unreleased/bug-fixes/5275-pruning-ext-commit-fix.md
+++ b/.changelog/unreleased/bug-fixes/5275-pruning-ext-commit-fix.md
@@ -1,0 +1,2 @@
+- `[store]` Prune extended commits properly
+  ([5275](https://github.com/cometbft/cometbft/issues/5275)
--- a/store/store.go
+++ b/store/store.go
@@ -531,6 +531,13 @@ func (bs *BlockStore) PruneBlocks(height int64, state sm.State) (uint64, int64, 
 			}
 			bs.blockCommitCache.Remove(h)
 		}
+		if h < evidencePoint {
+			if err := batch.Delete(bs.dbKeyLayout.CalcExtCommitKey(h)); err != nil {
+				return 0, -1, ErrDBOpt{Err: err}
+			}
+			bs.blockExtendedCommitCache.Remove(h)
+		}
+
 		if err := batch.Delete(bs.dbKeyLayout.CalcSeenCommitKey(h)); err != nil {
 			return 0, -1, ErrDBOpt{Err: err}
 		}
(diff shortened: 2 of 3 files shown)


internal/consensus/reactor.go
```diff
@@ -331,6 +331,16 @@ func (conR *Reactor) Receive(e p2p.Envelope) {
 		}
 		switch msg := msg.(type) {
 		case *ProposalMessage:
+			maxBytes := conR.conS.state.ConsensusParams.Block.MaxBytes
```
aljo242, line 334: Is there a way we can test this?

internal/consensus/reactor.go
```diff
@@ -331,6 +331,16 @@ func (conR *Reactor) Receive(e p2p.Envelope) {
...
+			if maxBytes == -1 {
+				maxBytes = int64(types.MaxBlockSizeBytes)
+			}
+			totalParts := int64(msg.Proposal.BlockID.PartSetHeader.Total)
+			if totalParts > (maxBytes-1)/int64(types.BlockPartSizeBytes)+1 {
```
almkx:
```suggestion
             chunks := (maxBytes-1) / int64(types.BlockPartSizeBytes)+1
			if totalParts > chunks {
```

almk-dev (commented):
Can you explain the -1/+1 logic in a comment?

arsushi:
> Can you explain the -1/+1 logic in a comment?

It is based on https://github.com/cometbft/cometbft/blob/88e9ea6552269006f80431206994d48d33873a4e/internal/consensus/state.go#L2075-L2082

internal/consensus/reactor.go
arsushi: Fixed in https://github.com/cometbft/cometbft/pull/5309/commits/621956b49721d2834a0508a5a65fb40fa8f29d55

internal/consensus/reactor.go
arsushi: Added test case in https://github.com/cometbft/cometbft/pull/5309/commits/6639886dd0cc08c66043c6d7fc6e0e2c3a3a2131

aljo242:
@arsushi can we also get a changelog entry for this?

internal/consensus/byzantine_test.go
```diff
@@ -610,3 +610,58 @@ func (br *ByzantineReactor) Receive(e p2p.Envelope) {
...
+	propBlockID.PartSetHeader.Total = 4294967295
+
+	proposal := types.NewProposal(height, round, -1, propBlockID, block.Time)
+	p := proposal.ToProto()
+	if err := cs.privValidator.SignProposal(cs.state.ChainID, p); err != nil {
+		t.Error(err)
+	}
+	proposal.Signature = p.Signature
+
+	go sendProposalAndParts(height, round, cs, targetPeer, proposal, block, block.Hash(), blockParts)
```
mattac21, line 666: I'm not sure this test case is actually testing anything. This looks to me that it creates the oversized proposal, and then sends the proposal and parts to the peer (where it will be rejected), but it doesnt actually asset that the peer rejected it, it returns before anything is asserted about the peer state.

PR #5324, opened 2025-09-02
fix(consensus/reactor): reject oversized proposals

aljo242:
---
Updates the consensus reactor to validate that a received proposal will not contain more parts than the amount of chunks that it would take to build a block whos size is equal to `ConsensusParams.Block.MaxBytes`.

Original PR is here https://github.com/cometbft/cometbft/pull/5309, but reopened since the contributor stopped replying.

#### PR checklist

- [ ] Tests written/updated


consensus/reactor.go
```diff
@@ -329,6 +329,19 @@ func (conR *Reactor) Receive(e p2p.Envelope) {
...
+			if maxBytes == -1 {
+				maxBytes = int64(types.MaxBlockSizeBytes)
+			}
+			totalParts := int64(msg.Proposal.BlockID.PartSetHeader.Total)
+			chunks := (maxBytes-1)/int64(types.BlockPartSizeBytes) + 1
+			if totalParts > chunks {
+				conR.Logger.Error("Rejecting oversized proposal", "peer", e.Src, "height", msg.Proposal.Height)
+				conR.Switch.StopPeerForError(e.Src, ErrProposalTooManyParts)
+				return
+			}
```
almkx, line 340: Nit: we can avoid the explicit return if we invert the clause, and log if it doesn't match:

```suggestion
			if totalParts <= chunks { 
				ps.SetHasProposal(msg.Proposal)
			    conR.conS.peerMsgQueue <- msgInfo{msg, e.Src.ID()}
			} else {
			    conR.Logger.Error("Rejecting oversized proposal", "peer", e.Src, "height", msg.Proposal.Height)
				conR.Switch.StopPeerForError(e.Src, ErrProposalTooManyParts)
			}
```

almk-dev (approved):

consensus/reactor.go
mattac21: Interesting I personally prefer the early return within the `if` rather than adding in an `else` case just to remove the return. I think early returns are quite nice and are easier to read in my head. It looks the pattern of using guard clauses (returning early in an `if` when we reach a bad/error state) is used quite often, so imo it makes sense to keep it as is.

aljo242: yes i am aligned with keeping this as is to make it clear that this conditional return is an "exceptional" case out of the happy path

types/proposal.go
```diff
@@ -79,6 +79,22 @@ func (p *Proposal) ValidateBasic() error {
 	return nil
 }
 
+// ValidateBlockSize block size ensures that a proposal block is not larger
+// than a maximum number of bytes, based on the total amount of parts reported
+// in the PartSetHeader. If -1 is passed as the maxBlockSizeBytes,
+// types.MaxBlockSizeBytes will be used as the maximum.
+func (p *Proposal) ValidateBlockSize(maxBlockSizeBytes int64) error {
```
aljo242, line 86: Can we add a minimal unit test for this function?

mattac21: added https://github.com/cometbft/cometbft/pull/5324/commits/87b9b3b963f67ab0b88e3ea82ca9b24597541b57

PR #5326, opened 2025-09-02
chore: reset branch to v0.38.x

aljo242:
Our main thinking in this decision is that we want to get all of the features from v1 and v2 that were useful to folks (performance, peering, BLS sigs, etc) without introducing the large set of API breakages and risks that I'm sure your team is familiar with given your upgrade.
We'll be sharing a more detailed roadmap for the stack soon, but driving cometBFT performance will be one of our main priorities.
The reality is that v1 has also been released officially and adopted by some folks, so we will be sure to continue to provide support or smooth upgrade paths

we have https://github.com/cometbft/cometbft/tree/main-backup to cover all of the previous work that has been done in v1 and v2. 

We will be selecting work from this branch to backport, but in order to maintain stability, we will not be introducing a lot of the large-scale refactors that have been done


almk-dev (commented):
Is this meant to be at full parity with the `0.38.x` branch, or partial/selective? If it's meant to be a perfect one to one match, this diff might not be exactly right:
```
❯ git diff v0.38.x | wc -l
     833
```

aljo242:
```
 git diff v0.38.x | wc -l       
      47
```

Fixed: now the only changes are `Makefile` changes we want

PR #5358, opened 2025-09-04
chore: backport bls

aljo242:


common.mk
```diff
@@ -40,9 +46,21 @@ ifeq (rocksdb,$(findstring rocksdb,$(COMETBFT_BUILD_OPTIONS)))
...
+
+# handle secp256k1eth
+ifeq (secp256k1eth,$(findstring secp256k1eth,$(COMETBFT_BUILD_OPTIONS)))
+  BUILD_TAGS += secp256k1eth
+endif
+
+# handle nodebug
+ifeq (nodebug,$(findstring nodebug,$(COMETBFT_BUILD_OPTIONS)))
+  CGO_ENABLED=1
+  BUILD_TAGS += nodebug
```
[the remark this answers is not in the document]
mattac21: this feels like a real thing, we have the same condition twice

mempool/reactor_test.go
```diff
@@ -234,14 +234,109 @@ func TestBroadcastTxForPeerStopsWhenReactorStops(t *testing.T) {
 // Note: in this test we know which gossip connections are active or not because of how the p2p
 // functions are currently implemented, which affects the order in which peers are added to the
 // mempool reactor.
-func TestMempoolReactorMaxActiveOutboundConnections(t *testing.T) {
+// func TestMempoolReactorMaxActiveOutboundConnections(t *testing.T) {
```
mattac21: why did this get commented out? should we just remove?

aljo242: looks like some odd change that got introduced  ill see if i can add back in or why it was removed

PR #5360, opened 2025-09-05
chore: backport perf

mattac21:
---
Backport selected performance improvement PRs made after v0.38 that were not previously back ported. Slight modifications to some to make them not breaking (keeping existing functions instead of removing and renaming a new function, etc).

#### PR checklist

- [ ] Tests written/updated


.changelog/unreleased/breaking-changes/2246-p2p-peerset.md
```diff
@@ -0,0 +1,3 @@
+- `[p2p]` Rename `IPeerSet#List` to `Copy`, add `Random`, `ForEach` methods.
```
aljo242: need to remove all of these and just move the entries to `./CHANGELOG.md`

CHANGELOG.md
```diff
@@ -21,8 +79,14 @@
 
 ### API-BREAKING
 
+- `[p2p]` Rename `IPeerSet#List` to `Copy`, add `Random`, `ForEach` methods.
+   Rename `PeerSet#List` to `Copy`, add `Random`, `ForEach` methods.
+   ([\#2246](https://github.com/cometbft/cometbft/pull/2246))
```
mattac21, line 84: This is API breaking since it removes `List`. Do we care about the break here? I believe we can just keep the `List` method and leave it unused and put a deprecation notice on it (I'm not really sure what the comet process is for deprecation of functions/apis).

CHANGELOG.md
```diff
@@ -21,8 +79,14 @@
...
 - `[crypto]` Remove Sr25519 curve
   ([\#3646](https://github.com/cometbft/cometbft/pull/3646))
+- `[rpc]` The endpoints `broadcast_tx_*` now return an error when the node is
+  performing block sync or state sync.
+  ([\#785](https://github.com/cometbft/cometbft/issues/785))
```
mattac21, line 89: This change was listed as breaking in the previous changelog, it is API breaking in the sense that it changes the response of broadcasting txs while the node is syncing. Do we care about this kind of break? I originally didnt include this PR in the list but without it, including [the change in this PR](https://github.com/cometbft/cometbft/pull/3211) broke the e2e tests (some mutex changes were needed across them)

light/provider/http/http_test.go
```diff
@@ -60,7 +60,7 @@ func TestProvider(t *testing.T) {
 	lb, err := p.LightBlock(context.Background(), 0)
 	require.NoError(t, err)
 	require.NotNil(t, lb)
-	assert.True(t, lb.Height < 1000)
+	assert.True(t, lb.Height < 10000)
```
mattac21, line 63: This test was flakey when using height 1000. When adding some logging to debug, the in memory chain was sometimes at a height > 1000 here.

p2p/switch.go
```diff
@@ -295,21 +298,46 @@ func (sw *Switch) Broadcast(e Envelope) chan bool {
 	return successChan
 }
 
+// BroadcastAsync runs a go routine for each attempted send, which will block
+// trying to send for defaultSendTimeoutSeconds.
+//
+// NOTE: BroadcastAsync uses goroutines, so order of broadcast may not be
+// preserved.
+func (sw *Switch) BroadcastAsync(e Envelope) {
```
mattac21, line 305: There change was previously breaking that modified the signature of `Broadcast`. Instead of breaking, I created `BroadcastAsync` for the new implementation, and changed all invocations of `Broadcast` to `BroadcastAysnc`. `Broadcast` is now unused.

.changelog/unreleased/breaking-changes/2246-p2p-peerset.md
mattac21: done 🙏

CHANGELOG.md
aljo242: The process will be whatever we make it.

I'm aligned with keeping the method, leaving unused and putting that deprecation notice (then removing in v0.40)

CHANGELOG.md
aljo242: Is it a protobuf api?  Adding a field to that should not be breaking right?

p2p/switch.go
aljo242: nonbreaking king

CHANGELOG.md
mattac21: It's not adding a field to the response, from what I can tell it's just changing the response behavior when syncing (broadcast tx will now error instead of broadcasting txs). Idk exactly why it is listed as breaking (and the PR isn't labeled with `breaking` and they dont use a `!` in the PR title, but the changelog entry was put under breaking), but I can't see any other reason why this would be breaking change besides that? Maybe because nodes will now not be processing tx's while syncing, the set of txs that the application sees while syncing will be different? So possibly state breaking? Not super familiar with what qualifies as state breaking, but this doesnt feel like it off the top of my head though.

CHANGELOG.md
mattac21: added https://github.com/cometbft/cometbft/pull/5360/commits/75e06bfb5a879ac60ebca728a9b6f73ca0677fe6

types/vote_set_test.go
```diff
@@ -121,6 +121,56 @@ func TestVoteSet_AddVote_Bad(t *testing.T) {
 
 }
 
+func Benchmark_2_3_Maj(b *testing.B) {
+	height, round := int64(1), int32(0)
+
+	voteProto := &Vote{
+		ValidatorAddress: nil, // NOTE: must fill in
+		ValidatorIndex:   -1,  // NOTE: must fill in
```
aljo242, line 128: what does this mean?

types/signature_cache.go
```diff
@@ -0,0 +1,36 @@
...
+
+type SignatureCache interface {
+	Add(key string, value SignatureCacheValue)
+	Get(key string) (SignatureCacheValue, bool)
+	Len() int
+}
+
+type signatureCache struct {
+	cache map[string]SignatureCacheValue
+}
```
aljo242: Feels premature to have this as an interface with a single implementation.  are there any other alternative ways wed even want to implement a signature cache?

statesync/reactor.go
```diff
@@ -275,7 +275,7 @@ func (r *Reactor) Sync(stateProvider StateProvider, discoveryTime time.Duration)
 		r.Logger.Debug("Requesting snapshots from known peers")
 		// Request snapshots from all currently connected peers
 
-		r.Switch.Broadcast(p2p.Envelope{
+		r.Switch.BroadcastAsync(p2p.Envelope{
```
aljo242, line 278: huge

state/execution.go
```diff
@@ -415,15 +425,36 @@ func (blockExec *BlockExecutor) Commit(
...
-
-	return res.RetainHeight, err
+	if err != nil {
+		// We panic in this case, out of legacy behavior. Before we made the mempool
+		// update complete asynchronously from Commit, we would panic if the mempool
+		// update failed. This is because we panic on any error within commit.
+		// We should consider changing this behavior in the future, as there is no
+		// need to panic if the mempool update failed. The most severe thing we
+		// would need to do is dump the mempool and restart it.
+		panic(fmt.Sprintf("client error during mempool.Update; error %v", err))
```
aljo242, line 455: do we know where this panic is potentially caught? or does this crash?

state/execution.go
```diff
@@ -188,9 +191,11 @@ func (blockExec *BlockExecutor) ProcessProposal(
 // Validation does not mutate state, but does require historical information from the stateDB,
 // ie. to verify evidence from a validator at an old height.
 func (blockExec *BlockExecutor) ValidateBlock(state State, block *types.Block) error {
-	err := validateBlock(state, block)
-	if err != nil {
-		return err
+	if !blockExec.lastValidatedBlock.HashesTo(block.Hash()) {
```
aljo242, line 193: is there potential for NPE?

rpc/core/net.go
```diff
@@ -13,19 +13,23 @@ import (
 // NetInfo returns network info.
 // More: https://docs.cometbft.com/v0.38/spec/rpc/#netinfo
 func (env *Environment) NetInfo(*rpctypes.Context) (*ctypes.ResultNetInfo, error) {
-	peersList := env.P2PPeers.Peers().List()
-	peers := make([]ctypes.Peer, 0, len(peersList))
-	for _, peer := range peersList {
+	peers := make([]ctypes.Peer, 0)
```
aljo242: why not preallocate this?

rpc/core/consensus.go
```diff
@@ -55,24 +58,29 @@ func (env *Environment) Validators(
 // More: https://docs.cometbft.com/v0.38.x/rpc/#/Info/dump_consensus_state
 func (env *Environment) DumpConsensusState(*rpctypes.Context) (*ctypes.ResultDumpConsensusState, error) {
 	// Get Peer consensus states.
-	peers := env.P2PPeers.Peers().List()
-	peerStates := make([]ctypes.PeerStateInfo, len(peers))
-	for i, peer := range peers {
+	peerStates := make([]ctypes.PeerStateInfo, 0)
```
aljo242: why not preallocate?

p2p/conn/evil_secret_connection_test.go
```diff
@@ -255,7 +258,8 @@ func TestMakeSecretConnection(t *testing.T) {
 		{"share bad ethimeral key", newEvilConn(true, true, false, false), "wrong wireType"},
 		{"refuse to share auth signature", newEvilConn(true, false, false, false), "EOF"},
 		{"share bad auth signature", newEvilConn(true, false, true, true), "failed to decrypt SecretConnection"},
-		{"all good", newEvilConn(true, false, true, false), ""},
+		// fails with the introduction of changes PR #3419
+		// {"all good", newEvilConn(true, false, true, false), ""},
```
aljo242: follow up for this?

CHANGELOG.md
Eric-Warehime: Seems like a change to behavior that could potentially show errors in cases where it used to not, but I don't see a problem with including it. It's definitely not consensus breaking and doesn't affect RPC request/response types.

types/vote_set_test.go
mattac21: It's just noting that those fields in the vote proto must eventually be set before signing the vote. this initialization leaves them unset, then the helper `withValidator` populates the same vote proto with a different validator to test different vote scenarios. Basically just a convenience in the test.

types/signature_cache.go
mattac21: Agree, doubt we will ever implement another one of these that isn't just a map, I'll remove it.

state/execution.go
mattac21: i believe it crashes. `asyncUpdateMempool` is called directly as a goroutine (`go asyncUpdateMempool(...)`), and there is no recover in `asyncUpdateMempool`, so comet will crash on this.

state/execution.go
mattac21: Shouldn't be, `lastValidatedBlock.HashesTo(...)` has an internal check that if `lastValidatedBlock == nil`, it will return `false` before it tries to deref the pointer.

rpc/core/net.go
mattac21: done here https://github.com/cometbft/cometbft/pull/5360/commits/5aa0e5542a8a44678f7e0cd83859ab25dc1ffdbf

rpc/core/consensus.go
mattac21: done here https://github.com/cometbft/cometbft/pull/5360/commits/5aa0e5542a8a44678f7e0cd83859ab25dc1ffdbf

state/execution.go
aljo242: does that seem ok or potentially exploitable?

p2p/conn/evil_secret_connection_test.go
mattac21: I dont really have any followup, you can see the [discussion here to comment it out](https://github.com/cometbft/cometbft/pull/3419#discussion_r1664890295). Everyone seemed to agree that the effect to make the test work wasn't worth it and the test was unnecessary.

state/execution.go
mattac21: My gut reaction would be that this is bad, but going back to look at how this `asyncUpdateMempool` is used (only within `BlockExecutor.Commit`) think it seems ok only because they are basically replicating the old behavior prior to this async update pr.

Previously, if the mempool update inside of `Commit` errored, we would return the error to the caller. Looking at all of the references of `Commit`, every caller would eventually panic on this error a few layers up the call stack. Now with this async change, the mempool update is now inside a goroutine. So if we simply return/log the error from the goroutine, it wont be propagated to the caller of `Commit` and they will get a `nil` error if the mempool update fails, making the caller eventually not panic in the case of a failed mempool update, deviating from the previous behavior. So this is why they opted to simply panic here, to replicate the old behavior of every caller panicking themselves on the mempool update failing.

You can see the original discussion here https://github.com/cometbft/cometbft/pull/3008#discussion_r1609947096

p2p/conn/evil_secret_connection_test.go
aljo242: ok let's just remove the test and lgtm

mattac21: remove here https://github.com/cometbft/cometbft/pull/5360/commits/1f3305dc33dc89eb70269e06f87a549b7bdb1387

PR #5400, opened 2025-09-16
chore: misleading error return in boundary check

mdqst:
noticed that the function was returning `(false, err)` even though `err` is always `nil` at this point. this could give the impression that an error was possible here when it actually isn’t.

updated the code to return `(false, nil)` instead, which makes it clear that hitting the boundary is not an error case.

#### PR checklist

- [ ] Tests written/updated


mattac21 (approved):
Looks good to me!

PR #5412, opened 2025-09-19
chore/(revert): revert change to remove sr25519 curve

mattac21:
---
Reverts a previous change to remove the sr25519 curve. It was determined that no one is using this curve in production and it should be removed. However this lead to downstream dependencies breaking (the sdk). This adds the curve back and marks it as depurated. I also removed mentions of it from documentation so that we are no longer recommending it as an option.

#### PR checklist

- [ ] Tests written/updated


crypto/sr25519/privkey.go
```diff
@@ -0,0 +1,177 @@
...
+	if err != nil {
+		panic("sr25519: failed to serialize public key: " + err.Error())
+	}
+
+	return PubKey(b)
+}
+
+// Equals - you probably don't need to use this.
+// Runs in constant time based on length of the keys.
+func (privKey PrivKey) Equals(other crypto.PrivKey) bool {
```
mattac21, line 88: An equals function got added to the `PrivKey` interface, so I had to add this on. Same for the `sr25519.PubKey.Equals()`.

crypto/sr25519/privkey.go
```diff
@@ -0,0 +1,177 @@
...
+	// PrivKeySize is the number of bytes in an Sr25519 private key.
+	PrivKeySize = 32
+
+	KeyType = "sr25519"
+)
+
+// PrivKey implements crypto.PrivKey.
+//
+// Deprecated: This key type will be removed, do not use.
+type PrivKey struct {
```
mattac21, line 32: note the deprecation notice here

crypto/sr25519/pubkey.go
```diff
@@ -0,0 +1,69 @@
...
+	PubKeySize = 32
+
+	// SignatureSize is the size of a Sr25519 signature in bytes.
+	SignatureSize = 64
+)
+
+// PubKey implements crypto.PubKey for the Sr25519 signature scheme.
+//
+// Deprecated: This key type will be removed, do not use.
+type PubKey []byte
```
mattac21, line 26: Note the deprecation notice here

mattac21 (commented):
most of the docs changes I am not adding back in since this key is now deprecated, I thought it made the most sense to keep references of it removed from the docs, but if others think differently and I can add them back in.

CHANGELOG.md
```diff
@@ -97,8 +97,6 @@
 - `[p2p]` Rename `IPeerSet#List` to `Copy`, add `Random`, `ForEach` methods.
    Rename `PeerSet#List` to `Copy`, add `Random`, `ForEach` methods.
    ([\#2246](https://github.com/cometbft/cometbft/pull/2246))
-- `[crypto]` Remove Sr25519 curve
-  ([\#3646](https://github.com/cometbft/cometbft/pull/3646))
```
mattac21, line 101: not sure if I should be removing this from the changelog or add another entry for reverting the change, this felt easier and more clear if the change just never happened, but since some of the docs changes stayed unreverted it may make sense to add in a new entry, or maybe remove this entry and add in a new entry to say that it is now deprecated?

aljo242: i would remove and add new one that e deprecated

internal/keytypes/keytypes.go
```diff
@@ -1,48 +0,0 @@
-package keytypes
-
-import (
```
aljo242, line 3: maybe we should keep thiw file and add sr25519 to it?

test/e2e/pkg/manifest.go
```diff
@@ -49,7 +49,7 @@ type Manifest struct {
 	Nodes map[string]*ManifestNode `toml:"node"`
 
 	// KeyType sets the curve that will be used by validators.
-	// Options are ed25519, secp256k1 and bls12381.
+	// Options are ed25519, secp256k1 and sr25519.
```
aljo242, line 52:
```suggestion
	// Options are ed25519, secp256k1, bls12381 and sr25519.
```

crypto/sr25519/privkey.go
aljo242: do we have tests for these functions on the impls?

PR #5421, opened 2025-09-24
feat: cache validator workload in BuildLastCommitInfo for 28x performance improvement

### LINKED ISSUE #3152: Cache the workload in state.buildLastCommitInfo
## Feature Request

Every block has, as a consensus blocking operation, state.BuildLastCommitInfo (twice)
image

This workload is almost entirely repeated across blocks upon inspection. 
- state.loadValidatorsInfo should be hitting caches, not our DB.
- PubkeyFromProto and PubkeyToProto should hit caches, but could potentially be entirely unneeded
- tm2pb.Validator, as noticed in #3149 is just rehashing an already-provided address. 

### Summary

Lower consensus delay, and I/O work.

### Proposal

Investigate caching and lower load of relevant workloads in `state.BuildLastCommitInfo`

CreeptoGengar:
## Summary
Implements caching for `state.BuildLastCommitInfo` to eliminate performance bottlenecks in consensus operations.

## Changes
- Add validator and ABCI validator caches to `BlockExecutor`
- Create optimized `BuildLastCommitInfoFromStoreWithCache` and `BuildLastCommitInfoWithCache` methods
- Update all call sites in `applyBlock`, `ExtendVote`, `ProcessProposal`
- Add `ClearValidatorCache()` for memory management

## Performance Impact
- **30.7x faster**: 8462 ns/op → 275.7 ns/op
- **9.7x less memory**: 3711 B/op → 384 B/op  
- **3.4x fewer allocations**: 37 → 11 allocs/op

Fixes #3152

### DIFF
--- a/state/execution.go
+++ b/state/execution.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	abci "github.com/cometbft/cometbft/abci/types"
@@ -43,6 +44,14 @@ type BlockExecutor struct {
 	// 1-element cache of validated blocks
 	lastValidatedBlock *types.Block
 
+	// cache for validators to avoid repeated DB lookups
+	validatorCache      map[int64]*types.ValidatorSet
+	validatorCacheMutex sync.RWMutex
+
+	// cache for ABCI validators to avoid repeated conversions
+	abciValidatorCache      map[string]abci.Validator
+	abciValidatorCacheMutex sync.RWMutex
+
 	logger log.Logger
 
 	metrics *Metrics
@@ -68,14 +77,16 @@ func NewBlockExecutor(
 	options ...BlockExecutorOption,
 ) *BlockExecutor {
 	res := &BlockExecutor{
-		store:      stateStore,
-		proxyApp:   proxyApp,
-		eventBus:   types.NopEventBus{},
-		mempool:    mempool,
-		evpool:     evpool,
-		logger:     logger,
-		metrics:    NopMetrics(),
-		blockStore: blockStore,
+		store:              stateStore,
+		proxyApp:           proxyApp,
+		eventBus:           types.NopEventBus{},
+		mempool:            mempool,
+		evpool:             evpool,
+		logger:             logger,
+		metrics:            NopMetrics(),
+		blockStore:         blockStore,
+		validatorCache:     make(map[int64]*types.ValidatorSet),
+		abciValidatorCache: make(map[string]abci.Validator),
 	}
 
 	for _, option := range options {
@@ -87,6 +98,17 @@ func NewBlockExecutor(
 
 func (blockExec *BlockExecutor) Store() Store {
 	return blockExec.store
+}
+
+// ClearValidatorCache clears the validator caches to free memory
+func (blockExec *BlockExecutor) ClearValidatorCache() {
+	blockExec.validatorCacheMutex.Lock()
+	blockExec.validatorCache = make(map[int64]*types.ValidatorSet)
+	blockExec.validatorCacheMutex.Unlock()
+
+	blockExec.abciValidatorCacheMutex.Lock()
+	blockExec.abciValidatorCache = make(map[string]abci.Validator)
+	blockExec.abciValidatorCacheMutex.Unlock()
 }
 
 // SetEventBus - sets the event bus for publishing block related events.
@@ -170,7 +192,7 @@ func (blockExec *BlockExecutor) ProcessProposal(
 		Height:             block.Height,
 		Time:               block.Time,
 		Txs:                block.Txs.ToSliceOfBytes(),
-		ProposedLastCommit: buildLastCommitInfoFromStore(block, blockExec.store, state.InitialHeight),
+		ProposedLastCommit: blockExec.BuildLastCommitInfoFromStoreWithCache(block, state.InitialHeight),
 		Misbehavior:        block.Evidence.Evidence.ToABCI(),
 		ProposerAddress:    block.ProposerAddress,
 		NextValidatorsHash: block.NextValidatorsHash,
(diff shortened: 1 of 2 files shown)


state/execution.go
```diff
@@ -800,6 +912,48 @@ func ExecCommitBlock(
 	return resp.AppHash, nil
 }
 
+// ExecCommitBlockWithCache is an optimized version that uses caching for better performance
+func (blockExec *BlockExecutor) ExecCommitBlockWithCache(
```
mattac21: this is unused. looks like the previous `ExecCommitBlock` is only used when replaying blocks. do we want to use that there?

state/execution.go
```diff
@@ -89,6 +100,17 @@ func (blockExec *BlockExecutor) Store() Store {
 	return blockExec.store
 }
 
+// ClearValidatorCache clears the validator caches to free memory
+func (blockExec *BlockExecutor) ClearValidatorCache() {
```
mattac21: this is unused except in tests, so the part of your test that is testing cache clearing isn't really testing anything

state/execution.go
```diff
@@ -68,14 +77,16 @@ func NewBlockExecutor(
...
+		store:              stateStore,
+		proxyApp:           proxyApp,
+		eventBus:           types.NopEventBus{},
+		mempool:            mempool,
+		evpool:             evpool,
+		logger:             logger,
+		metrics:            NopMetrics(),
+		blockStore:         blockStore,
+		validatorCache:     make(map[int64]*types.ValidatorSet),
+		abciValidatorCache: make(map[string]abci.Validator),
```
mattac21, line 92: in real execution, no values from these caches are ever removed, so these will grow forever. We will need someway to periodically remove elements from them.

state/execution.go
```diff
@@ -513,6 +567,64 @@ func BuildLastCommitInfo(block *types.Block, lastValSet *types.ValidatorSet, ini
...
+			commitSize, valSetLen, block.Height, block.LastCommit.Signatures, lastValSet.Validators,
+		))
+	}
+
+	votes := make([]abci.VoteInfo, block.LastCommit.Size())
+	for i, val := range lastValSet.Validators {
+		commitSig := block.LastCommit.Signatures[i]
+
+		// Create cache key for this validator
+		cacheKey := fmt.Sprintf("%s_%d", val.PubKey.Address(), val.VotingPower)
```
mattac21: from my understanding the point of the `abciValidatorCache` is to not have to do the expensive `val.PubKey.Address()` call. however you are calling `val.PubKey.Address()` when generating the `cacheKey`. doesn't this eliminate most of the point of the cache?

CreeptoGengar:
Had some race conditions in the validator caching that were causing CI failures, so I added proper double-checked locking and moved the cache cleanup outside the mutex locks to prevent deadlocks. 

Also toned down the cache cleanup frequency since it was interfering with mempool rechecking and updated the test accordingly.

CreeptoGengar:
@mattac21 @aljo242 

please check it out and correct me if needed

I appreciate your support and corrections :)

state/execution.go
```diff
@@ -89,6 +104,45 @@ func (blockExec *BlockExecutor) Store() Store {
...
+}
+
+// cleanupOldCacheEntries removes old entries from caches to prevent memory leaks
+func (blockExec *BlockExecutor) cleanupOldCacheEntries() {
+	// Only cleanup if cache is significantly over the limit to avoid frequent cleanup
+	blockExec.validatorCacheMutex.Lock()
+	if len(blockExec.validatorCache) > blockExec.maxCacheSize*2 {
+		// Simple cleanup: clear half of the cache
+		// Since Go maps don't guarantee iteration order, we'll clear the entire cache
+		// and let it rebuild naturally. This is simpler and avoids the FIFO issue.
```
mattac21: cursor pointed this out as well but the first line of the comment here is misleading, it clears the whole cache, not half

mattac21: same below

state/execution.go
```diff
@@ -89,6 +104,45 @@ func (blockExec *BlockExecutor) Store() Store {
...
+// SetMaxCacheSize sets the maximum cache size for testing
+func (blockExec *BlockExecutor) SetMaxCacheSize(size int) {
+	blockExec.maxCacheSize = size
+}
+
+// cleanupOldCacheEntries removes old entries from caches to prevent memory leaks
+func (blockExec *BlockExecutor) cleanupOldCacheEntries() {
+	// Only cleanup if cache is significantly over the limit to avoid frequent cleanup
+	blockExec.validatorCacheMutex.Lock()
+	if len(blockExec.validatorCache) > blockExec.maxCacheSize*2 {
```
mattac21: why `*2` here? that feels misleading against what the `maxCacheSize` variable implies (that we will not grow past it). Since you already initialize the `maxCacheSize` to a default of `1000`, why not just set it to `2000` by default?

state/execution.go
```diff
@@ -474,6 +528,46 @@ func buildLastCommitInfoFromStore(block *types.Block, store Store, initialHeight
...
+		var err error
+		lastValSet, err = blockExec.store.LoadValidators(height)
+		if err != nil {
+			panic(fmt.Errorf("failed to load validator set at height %d: %w", height, err))
+		}
+
+		// Double-checked locking: acquire write lock and check again
+		blockExec.validatorCacheMutex.Lock()
+		// Check again in case another goroutine added it
+		if existingValSet, exists := blockExec.validatorCache[height]; exists {
```
mattac21: im not following why you would need to grab from the cache again here? wouldn't this always be the same as what you just loaded from the store on line 549? I get that another goroutine may have just cached it and its now available in the cache, but why do we need to fetch it from the cache when we already did the work of loading it from the store above?

state/execution.go
```diff
@@ -89,6 +104,45 @@ func (blockExec *BlockExecutor) Store() Store {
...
+
+// SetMaxCacheSize sets the maximum cache size for testing
+func (blockExec *BlockExecutor) SetMaxCacheSize(size int) {
+	blockExec.maxCacheSize = size
+}
+
+// cleanupOldCacheEntries removes old entries from caches to prevent memory leaks
+func (blockExec *BlockExecutor) cleanupOldCacheEntries() {
+	// Only cleanup if cache is significantly over the limit to avoid frequent cleanup
+	blockExec.validatorCacheMutex.Lock()
```
mattac21: this will be called quite frequently and you dont actually need a write lock on this except for the single time that the cache needs to be cleared. it may make more sense to acquire a read only lock when checking the condition, and change to a write lock only when the cache actually needs to be written to

mattac21: you could also use your helper `GetCacheSize` to get the sizes with a read only lock

state/execution.go
mattac21: same question for the `BuildLastCommitInfoWithCache`

PR #5422, opened 2025-09-25
chore: remove unused min func

cargoedit:
Go 1.21 adds built-in functions min and max.   https://tip.golang.org/doc/go1.21
Therefore, we can directly remove our own implementations and use the built-in function with the same names. 

---

#### PR checklist

- [ ] Tests written/updated


cargoedit:
@aljo242 Could you please review this PR at your convenience? Thank you very much.

PR #5445, opened 2025-10-06
refactor: remove MsgToProto function and inline conversion in WALToProto

hawkadrian:
Remove the MsgToProto function that was marked with TODO for removal.

The function was only used by WALToProto for converting consensus messages to protobuf format. 

This change eliminates code duplication by inlining the message conversion logic directly into WALToProto, making the code more maintainable and removing the unnecessary abstraction layer.


aljo242:
CI failing

PR #5466, opened 2025-10-21
refactor: use slices.Contains to simplify code and remove no-used function

wyrapeseed:
There is a [new function](https://pkg.go.dev/slices@go1.21.0#Contains) added in the go1.21 standard library, which can make the code more concise and easy to read.

---

#### PR checklist

- [ ] Tests written/updated


libs/strings/string.go
```diff
@@ -2,17 +2,13 @@ package strings
...
 
 // StringInSlice returns true if a is found in the list.
 func StringInSlice(a string, list []string) bool {
-	for _, b := range list {
-		if b == a {
-			return true
-		}
-	}
-	return false
+	return slices.Contains(list, a)
```
aljo242: we just don't need this function

state/indexer/block/kv/util.go
```diff
@@ -23,13 +24,7 @@ type HeightInfo struct {
...
 
 func intInSlice(a int, list []int) bool {
-	for _, b := range list {
-		if b == a {
-			return true
-		}
-	}
-
-	return false
+	return slices.Contains(list, a)
```
aljo242: dont need this function

state/txindex/kv/utils.go
```diff
@@ -21,12 +22,7 @@ type HeightInfo struct {
 
 // IntInSlice returns true if a is found in the list.
 func intInSlice(a int, list []int) bool {
-	for _, b := range list {
-		if b == a {
-			return true
-		}
-	}
-	return false
+	return slices.Contains(list, a)
```
aljo242: don't need this function

libs/strings/string.go
wyrapeseed: Thanks for your suggestions. Removed.

state/txindex/kv/utils.go
wyrapeseed: Removed.

state/indexer/block/kv/util.go
wyrapeseed: Thanks for your suggestions. Removed.

wyrapeseed:
@aljo242 Thanks for your review!  Modified.  

If there's anything need I to do, please feel free to let me know.

aljo242 (approved):

PR #5494, opened 2025-11-11
feat(lib-p2p): add priority queue for inbound envelopes

swift1337:
This PR adds a priority queue for inbound `Reactor` envelopes before they arrive at the subsequent consumer pool. The implementation relies on a thread-safe leveled linked list instead of a max heap because we have only 10 priorities in Comet. Messages with the same priority are processed as FIFO

image

---

#### PR checklist

- [x] Tests written/updated

---

> [!NOTE]
> Introduce a thread-safe priority queue for inbound reactor envelopes, piped into an autoscaling consumer pool, and wire priorities through switch/reactor; add queue package and tests plus pool stop hook.
> 
> - **lp2p/reactor pipeline**:
>   - Replace per-reactor `envelopeQueue` channel with a `queue.PriorityQueue` and pipe to an autoscaling `autopool` consumer pool.
>   - Add `Receive(reactorName, messageType, envelope, priority)` to enqueue by priority and FIFO within priority; track `sequence` and enqueue time.
>   - Introduce `pipeQueues`, per-reactor scaler `newReactorScaler`, and refactor receive path to `receiveQueued` with updated metrics/logging.
>   - Start/stop flow updated: pool `OnStop` closes pipes/queues; removed direct channel closes.
> - **Switch**:
>   - Use `ChannelDescriptor.Priority` and call `reactors.Receive(..., priority)`.
> - **autopool**:
>   - Add `WithOnStop` option and invoke `onStop` in `Pool.Stop()`.
> - **queue package**:
>   - New thread-safe FIFO `Queue` and leveled `PriorityQueue` (higher value = higher priority) with tests/bench-like metrics logging.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cca041581daa48adc3ff50867bdcf3fb7c1130c0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


lp2p/queue/queue.go
```diff
@@ -0,0 +1,129 @@
+package queue
```
aljo242: should move to an `internal` package

aljo242: if this is useful outside of lp2p  also would make sense to put in a  top level `internal` package

internal/autopool/queue.go
```diff
@@ -0,0 +1,129 @@
...
+			return v, ok
+		}
+	}
+
+	return nil, false
+}
+
+func (q *PriorityQueue) updateHighestNonEmpty(lastLevel int) {
+	// noop
+	if q.levels[lastLevel].Len() > 0 {
```
aljo242, line 121: potential panic?

lp2p/autopool/pool.go
```diff
@@ -105,6 +110,10 @@ func (p *Pool[T]) Stop() {
 		return
 	}
 
+	if p.onStop != nil {
+		p.onStop()
+	}
+
```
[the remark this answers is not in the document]
mattac21: this seems like a real bug to me that we should fix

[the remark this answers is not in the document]
mattac21: seems like the easiest solution to fix would be to just remove the call to stop the pool if a worker sees that the pool channel is closing, since you know that the pool is already stopping if that channel is closed

lp2p/reactor_set.go
```diff
@@ -243,33 +309,43 @@ func (rs *reactorSet) newReactorQueue(
...
 
-	concurrencyCounter := rs.switchRef.metrics.MessageReactorQueueConcurrency.With("reactor", reactorName)
+func pipeQueues(producer *queue.PriorityQueue, consumer chan pendingEnvelope) chan struct{} {
+	stop := make(chan struct{})
+
+	go func() {
+		for {
+			value, ok := producer.Pop()
+			if !ok {
+				time.Sleep(10 * time.Millisecond)
```
mattac21: Curious if you thought about trying to avoid the sleep here? If (`queue.PriorityQueue`) exposed a channel that notified whenever a value was available to be popped (whenever `pq.Add` is called, we push an empty struct to this channel, telling consumers they can call `Pop` and will get a value back) then we could remove the sleep and only `Pop` when the notify chan tells us to. 

This does have a drawback (that probably needs some more thinking on) where if `pipeQueues` isn't ready, then `Add` could block trying to push onto the notify channel which feels quite bad. Originally I thought this wouldn't be an issue if we make this notify channel have a large capacity, but if we see a back up in consumption from the `consumer` chan that this function pushes onto, then that would delay `pipeQueues` from `Pop`ing, which would then delay `Add` calls to the priority queue.

lp2p/reactor_set.go
```diff
@@ -243,33 +309,43 @@ func (rs *reactorSet) newReactorQueue(
...
+				// should never happen
+				panic("unexpected type in priority queue")
+			}
+
+			select {
+			case <-stop:
+				// stop chan called before consumer close
+				return
+			default:
+			}
```
mattac21: What is the reasoning for having the stop `select` after the call to `Pop`? as of right now, if `stop` is closed but no values are pushed via the producer, then this goroutine will never exit. It feels like the `select` should go to the stop of the loop before the call to `Pop`

lp2p/reactor_set.go
```diff
@@ -243,33 +309,43 @@ func (rs *reactorSet) newReactorQueue(
...
+
+			consumer <- pe
```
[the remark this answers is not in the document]
mattac21: agree with this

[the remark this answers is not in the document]
mattac21: if you want to preserve the behavior of always stopping before pushing to the consumer, but avoiding the bug cursor mentioned, you could do this:

The top select prioritizes the check to see if we need to stop (if they were in a single select statement and both stop and push to consumer are ready, they are chosen randomly), and then in the inner select we try and push to the consumer (but if the consumer isn't ready and then we close stop, we will still stop and not potentially block forever).
```go
select {
	case <-stop:
		// stop chan called before consumer close
		return
	default:
        select {
            case <-stop:
                return
            case consumer <- pe:
                // nothing
        }
}

```

lp2p/reactor_set.go
```diff
@@ -20,13 +22,15 @@ type reactorSet struct {
 
 	// [protocol_id => reactorProtocol] mapping
 	protocols map[protocol.ID]reactorProtocol
+
+	sequence atomic.Uint64
```
mattac21: sequence is just used for debug logging, right? took me a little bit to figure that out it may be good to add a comment that this isn't actually used for any logic.

lp2p/reactor_set.go
```diff
@@ -184,56 +196,110 @@ func (rs *reactorSet) SubmitReceive(reactorName, messageType string, envelope p2
...
+	priorityQueue := queue.NewPriorityQueue(priorities)
+
+	// 2. create a queue for message processing (chan)
+	// messages from the first queue will be published here for concurrent processing
+	const concurrentPoolCapacity = 512
+
+	poolQueue := make(chan pendingEnvelope, concurrentPoolCapacity)
+
+	// 3. create a pipe from priority queue to the pool queue
+	pipeStopCh := pipeQueues(priorityQueue, poolQueue)
```
mattac21: curious if you thought about integrating the `autopool` directly with the priority queue? or created a small wrapper around the `autopool` that integrates with the priority queue? If we did that, then it would eliminate the need for this intermediate adapter goroutine that consumes from the priority queue and produces to the single channel that the `autopool` expects as its producer type.

lp2p/reactor_set.go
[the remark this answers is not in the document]
swift1337: Currently, it's indeed possible to leak 1 go-routine, but only during node shutdown. However, I agree with the comment, will make it leak-free

lp2p/autopool/pool.go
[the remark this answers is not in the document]
swift1337: good catch

lp2p/reactor_set.go
swift1337: Good call, initially seq was used to order the message, now it's redundant. Removing it.

lp2p/reactor_set.go
swift1337: > curious if you thought about integrating the autopool directly with the priority queue? 

Yes, but I thought that it would make `autopool` too complicated (dynamic scaling + priority queue + worker pool, ...). I'll try to simplify the API

internal/autopool/queue.go
swift1337: No, this is an internal function. `priority` is validated in `Push() error`

PR #5509, opened 2025-11-24
feat(mempool): implement AppMempool

swift1337:
---

This PR implements **AppMempool** based on new ABCI methods. The API is simple:

```go
// InsertTx inserts a tx into app-side mempool. The call is blocking, but thread-safe.
// Concurrent calls are expected and are caller's responsibility to handle.
func (m *AppMempool) InsertTx(tx types.Tx) error {
 // call abci.InsertTx(...)
}

// TxStream spins up a channel that streams valid transactions from app-side mempool.
// The expectation is that the caller would share it with other peers to gossip transactions.
func (m *AppMempool) TxStream(ctx context.Context, capacity int) <-chan types.Tx {
 // go loop { abci.ReapTxs(...) }
}
```

**UPD**: after a while, I found `StatelessMempool` term too confusing and thought that `AppMempool` sounds way more straightforward

Closes STACK-1838

#### PR checklist

- [x] Tests written/updated

---

> [!NOTE]
> Adds an app-managed mempool that proxies ABCI InsertTx/ReapTxs with retry handling and supporting tests.
> 
> - **Mempool**:
>   - **`mempool/AppMempool`**: New app-side mempool that proxies ABCI.
>     - `InsertTx`: validates size, dedups via LRU `seen` cache, forwards to app, updates metrics.
>     - Retry handling: drops from `seen` and returns error when code >= `abci.CodeTypeRetry`.
>     - `TxStream`: periodic `ReapTxs` loop to emit txs via channel; graceful ctx handling and logging.
>     - `FlushAppConn` implemented; other `Mempool` methods are no-ops as app builds blocks.
> - **ABCI types**:
>   - Add `CodeTypeRetry` (>= 32000) to mark retryable tx results.
> - **Tests**:
>   - `app_mempool_test.go`: covers `InsertTx` paths (ok/seen/empty/retryable) and `TxStream` reaping.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7f054e673110b8a6aef156265c88e18863cc18d1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/mempool/app_mempool.go
+++ b/mempool/app_mempool.go
@@ -98,9 +98,9 @@ func (m *AppMempool) InsertTx(tx types.Tx) error {
 		return ErrEmptyTx
 	}
 
-	if m.config.MaxBatchBytes > 0 && txSize > m.config.MaxBatchBytes {
+	if m.config.MaxTxBytes > 0 && txSize > m.config.MaxTxBytes {
 		return ErrTxTooLarge{
-			Max:    m.config.MaxBatchBytes,
+			Max:    m.config.MaxTxBytes,
 			Actual: txSize,
 		}
 	}


mempool/app_mempool.go
```diff
@@ -0,0 +1,228 @@
...
+		m.metrics.AlreadyReceivedTxs.Add(1)
+		return ErrSeenTx
+	}
+
+	code, err := m.insertTx(tx)
+
+	// todo (@swift1337): should we define more codes and handle them respectively?
+	// todo: remove tx from seen is app returns "is full" code --> retry later
+	switch {
+	case err != nil:
```
Eric-Warehime, line 118: I think we want to handle the case where the app rejects the transaction, but wants it to be delivered again if possible. e.g. if the processing queue is full. In this case we would not want to enter the tx in the seen cache.

Eric-Warehime (approved):
1 comment. LGTM otherwise.

mempool/app_mempool.go
swift1337: Yeah, I was thinking about the same. This requires a well-defined code enum eg code=123 → "retry-later"

mempool/app_mempool.go
```diff
@@ -0,0 +1,228 @@
...
+		for _, tx := range res.Txs {
+			select {
+			case <-ctx.Done():
+				m.logger.Debug("AppMempool.reapTxs: context done while streaming txs")
+				return
+			case channel <- tx:
+				// all good
+			}
+		}
+	}
```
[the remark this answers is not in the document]
mattac21: this feels real to me, instead of ranging over the ticket we should select on the tickers channel as well as context.done

mempool/app_mempool.go
```diff
@@ -0,0 +1,228 @@
...
+func (m *AppMempool) reapTxs(ctx context.Context, channel chan<- types.Tx) {
+	req := &abci.RequestReapTxs{
+		MaxBytes: reapMaxBytes,
+		MaxGas:   reapMaxGas,
+	}
+
+	ticker := time.NewTicker(reapInterval)
+	defer ticker.Stop()
+
+	for range ticker.C {
```
mattac21: a note here, if reap txs takes a long time, we dont reset the ticket before iterating the for loop, this means that if we tick every 2 seconds, and then app.reaptxs takes 1 second, the next tick will come in 1 second instead of 2. imo we should stop the ticket once we receive a tick from it and then reset it just before we wait on the ticker again

mempool/app_mempool.go
```diff
@@ -0,0 +1,228 @@
...
+			m.logger.Error("AppMempool.reapTxs: error reaping txs", "error", err)
+			continue
+		}
+
+		for _, tx := range res.Txs {
+			select {
+			case <-ctx.Done():
+				m.logger.Debug("AppMempool.reapTxs: context done while streaming txs")
+				return
+			case channel <- tx:
```
mattac21: would the capacity of this channel be == current comet clist mempool size? It feels like it would be very bad if this channel fills up and we block here and are then unable to calll reap txs. This seems like (maybe) the same behavior that would happen if the comet mempool fills up right now, but since comet mempool is a linked list, i dont think it would take up nearly as much memory as it would be to initialize a channel with a length of the current comet mempool size (not 100% sure if that's actually how a channel initialization works, but that's how it works in my head lol).

mattac21: I guess the alternative here would be to push to a massive buffer or linked list with a massive capacity (but requires no memory if not full), vs a channel

mempool/app_mempool.go
```diff
@@ -0,0 +1,228 @@
...
+// AppMempoolClient is the interface for the app-side mempool.
+type AppMempoolClient interface {
+	// InsertTx inserts a tx into app-side mempool
+	InsertTx(ctx context.Context, req *abci.RequestInsertTx) (*abci.ResponseInsertTx, error)
+
+	// ReapTxs reaps txs from app-side mempool
+	ReapTxs(ctx context.Context, req *abci.RequestReapTxs) (*abci.ResponseReapTxs, error)
+
+	// Flush app's connection
+	Flush(context.Context) error
```
mattac21, line 39: what does flush do?

mempool/app_mempool.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+// CheckTx returns an error on purpose for explicitness
+func (m *AppMempool) CheckTx(_ types.Tx, _ func(*abci.ResponseCheckTx), _ TxInfo) error {
+	return ErrNotImplemented
+}
+
+// Update does nothing for an app mempool
+func (m *AppMempool) Update(_ int64, _ types.Txs, _ []*abci.ExecTxResult, _ PreCheckFunc, _ PostCheckFunc) error {
+	return nil
```
mattac21, line 219: we would want to remove from the cache in update right? update is called after a block commits and our two scenarios where we remove from the cache are after we gossip a tx (im assuming this would be defined when we do the reactor refactor for this) and after a block commits that includes the tx

mempool/app_mempool.go
swift1337: In the socket client, it ensures that all reqs/res are flushed and received back

mempool/app_mempool.go
swift1337: > would the capacity of this channel be == current comet clist mempool size?

up to us; it will be only used to broadcast tx to peers, no other processing logic

mempool/app_mempool.go
swift1337: > we would want to remove from the cache in update right?

No, because `seen` is a LRU cache, and if tx is seen and commited, that's totally fine, we don't want to try to insert/check tx from a "lagged" peer again

mattac21: I see yea that makes sense

PR #5511, opened 2025-11-26
add authority parameter to consensus params

vladjdk:
partially closes: https://linear.app/cosmoslabs/issue/STACK-1856/replace-authority-keeper-params-with-consensus-parameters-authority

---

> [!NOTE]
> Introduces `AuthorityParams` to consensus params with default, validation (<=256 chars), full proto/plumbing, and tests, plus changelog update.
> 
> - **Types**:
>   - Add `AuthorityParams` and `ConsensusParams.Authority`; provide `DefaultAuthorityParams` (hash of "gov"), `ValidateBasic` max length 256, `Update`, `ToProto`/`FromProto` wiring.
>   - Extend tests to cover authority validation, updates, and proto round-trip; adjust helpers to accept `authority`.
> - **Proto**:
>   - Introduce `AuthorityParams` message and add `authority` field to `ConsensusParams`; regenerate Go code (getters, marshal/unmarshal, size, equal).
> - **Docs**:
>   - Update `CHANGELOG.md` with validation note for `AuthorityParams.Authority`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6774f2a3c798fb8274e60625d08a1e28ab32bec8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/proto/tendermint/types/params.proto
+++ b/proto/tendermint/types/params.proto
@@ -13,8 +13,8 @@ message ConsensusParams {
   BlockParams block = 1;
   EvidenceParams evidence = 2;
   ValidatorParams validator = 3;
-  VersionParams   version   = 4;
-  ABCIParams      abci      = 5;
+  VersionParams version = 4;
+  ABCIParams abci = 5;
   AuthorityParams authority = 6;
 }
 


types/params.go
```diff
@@ -139,6 +147,12 @@ func DefaultABCIParams() ABCIParams {
 	}
 }
 
+func DefaultAuthorityParams() AuthorityParams {
+	return AuthorityParams{
+		Authority: crypto.AddressHash([]byte("gov")).String(),
```
swift1337: I don't think that it's backward compatible as an existing chain would have `""`

types/params.go
```diff
@@ -374,5 +401,8 @@ func ConsensusParamsFromProto(pbParams cmtproto.ConsensusParams) ConsensusParams
 	if pbParams.Abci != nil {
 		c.ABCI.VoteExtensionsEnableHeight = pbParams.Abci.GetVoteExtensionsEnableHeight()
 	}
+	if pbParams.Authority != nil {
+		c.Authority.Authority = pbParams.Authority.Authority
+	}
```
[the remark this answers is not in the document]
swift1337: agree

types/params_test.go
```diff
@@ -230,14 +253,17 @@ func TestConsensusParamsUpdate_VoteExtensionsEnableHeight(t *testing.T) {
...
+		makeParams(1, 4, 3, 1, valEd25519, 1, ""),
+		makeParams(1, 2, 4, 1, valEd25519, 1, ""),
+		makeParams(2, 5, 7, 1, valEd25519, 1, ""),
+		makeParams(1, 7, 6, 1, valEd25519, 1, ""),
+		makeParams(9, 5, 4, 1, valEd25519, 1, ""),
+		makeParams(7, 8, 9, 1, valEd25519, 1, ""),
+		makeParams(4, 6, 5, 1, valEd25519, 1, ""),
+		makeParams(1, 2, 3, 1, valEd25519, 1, ""),
+		makeParams(1, 2, 3, 1, valEd25519, 1, "governance-module"),
+		makeParams(1, 2, 3, 1, valEd25519, 1, "cosmos10d07y265gmmuvt4z0w9aw880jnsr700j6zn9kn"),
```
swift1337, line 262: Can you add a test case that showcases backward (when Authority is nil)?

types/params.go
vladjdk: It's not backward-compatible. Each chain would need to specifically set an authority during an upgrade. The default is "", which means that there is no authority on the modules.

Since Comet has no knowledge of bech32 addresses, I removed this default completely and we defer to the app-side for validation.

proto/tendermint/types/params.proto
```diff
@@ -90,3 +91,7 @@ message ABCIParams {
   // to the application to use when proposing a block during PrepareProposal.
   int64 vote_extensions_enable_height = 1;
 }
+
+message AuthorityParams {
```
aljo242, line 99: needs doc

types/params_test.go
aljo242: bump

vladjdk: I changed the default to "", which means that when an application upgrades the SDK version to use this parameter, the upgrade is done there, where the application sets a value.

This is also why I can't add a test proving backwards-compatibility—this isn't backwards-compatible and chains need to set the value via UpgradeHandler. This will be handled by https://github.com/cosmos/cosmos-sdk/pull/25607 when it's done.

types/params.go
vladjdk: Changed the default to "" to make it more clear that this is deferred to the app-side

proto/tendermint/types/params.proto
vladjdk: added

PR #5512, opened 2025-11-27
feat(mempool): implement app-side mempool reactor

swift1337:
This PR adds full support for **AppMempool** in Comet. The idea is to move all mempool logic to the app side and only accept/relay mempool txs in a non-blocking & concurrent way.

---

- [x] Simplify mempool's logic by returning `<-chan types.Txs` instead of `<-chan types.Tx` to match ABCI 
- [x] Implement new mempool reactor for receiving and broadcasting txs
- [x] Use mempool txs batching (`batches := chunkTxs(txs, maxBatchSizeBytes)`)
- [x] Add `mempool.type=app` to consensus config, update node's setup
- [x] Mimic `CheckTx` for `rpc.BroadcastTxSync`, `rpc.BroadcastTxAsync`, `rpc.BroadcastTxCommit`
- [x] Add E2E tests
    - [x] app-mempool
    - [x] app-mempool + perturbations
    - [x] app-mempool + lib-p2p
- [x] Add metrics

closes STACK-1851

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/mempool/app_mempool.go
+++ b/mempool/app_mempool.go
@@ -203,6 +203,7 @@ func (m *AppMempool) reapTxs(ctx context.Context, channel chan<- types.Txs) {
 			}
 
 			txs := types.ToTxs(res.Txs)
+			m.metrics.ReapedTxs.Add(float64(len(txs)))
 
 			select {
 			case <-ctx.Done():
--- a/mempool/app_reactor.go
+++ b/mempool/app_reactor.go
@@ -145,6 +145,8 @@ func (r *AppReactor) Receive(e p2p.Envelope) {
 		return
 	}
 
+	r.mempool.metrics.BatchSize.With("dir", "inbound").Observe(float64(len(txs)))
+
 	for _, tx := range txs {
 		r.insertTx(peerID, tx)
 	}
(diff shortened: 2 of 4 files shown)


mempool/app_mempool.go
```diff
@@ -209,9 +235,46 @@ func (m *AppMempool) FlushAppConn() error {
 	return nil
 }
 
-// CheckTx returns an error on purpose for explicitness
-func (m *AppMempool) CheckTx(_ types.Tx, _ func(*abci.ResponseCheckTx), _ TxInfo) error {
-	return ErrNotImplemented
+// CheckTx mimics the behavior of the CListMempool's CheckTx method,
+// but actually only calls the app's InsertTx method. This is required for RPC compatibility.
+// @see: BroadcastTxSync, BroadcastTxAsync, and BroadcastTxCommit.
```
Eric-Warehime, line 240: Should we just prevent this from being a footgun by leaving it as an error and/or not registering the broadcast RPCs if the app mempool is configured?

swift1337: I'd leave it as is, because we still might have regular cosmos txs (gov, staking, etc...). `BroadcastTxAsync` flow remains identical to clist-mempool.

Let's reconsider this after e2e Krakatoa tests, if needed.

swift1337:
> but if we use libp2p, does this make a difference?

I think it benefits both comet/lib-p2p, because batching implies less overhead & concurrency, which is a good thing.

I'd revisit this if we want to implement gossip-sub, which requires per-message uniqueness fully; otherwise, different txs in a batch would result in different msgs → making gossip-sub useless.

But even for gossip-sub I would implement a two-layer model, where we batch+flood between validators and gossip between the rest of the network

PR #5553, opened 2026-01-07
fix(blocksync): stop timer leak in requestRoutine

reallesee:
The `defer retryTimer.Stop()` was placed inside the OUTER_LOOP, which means a new defer was added on each iteration but only executed when the function returned. This caused timer goroutines to accumulate and leak until they expired naturally.

Moved explicit `retryTimer.Stop()` calls to the return paths to properly clean up the timer when exiting the function.

---

> [!NOTE]
> **Low Risk**
> Low risk, behavior is unchanged aside from proper timer cleanup; the main risk is subtle timing differences around `Stop()` on early exits.
> 
> **Overview**
> Fixes a retry timer leak in `bpRequester.requestRoutine` by removing the per-iteration `defer retryTimer.Stop()` inside the outer request loop.
> 
> The timer is now explicitly stopped on early-return paths (`pool.Quit()` / `bpr.Quit()`), preventing accumulated timer goroutines during long-running block sync.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 566b3e7e265984c11b143eaff654c3107b70c672. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


blocksync/pool.go
```diff
@@ -849,16 +849,17 @@ OUTER_LOOP:
 		}
 
 		retryTimer := time.NewTimer(requestRetrySeconds * time.Second)
-		defer retryTimer.Stop()
```
mattac21, line 885: so since the timers were still referenced by the `defer`, they couldn't properly be garbage collected and the unused timers would continue until return, correct?

mattac21 (changes requested):
makes sense, have you seen any impact of this leak in production? also lint is failing but we have fixed it on main. can you merge `main` into your branch?

blocksync/pool.go
reallesee: no, just found it during code review

swift1337, line 885: Good finding. However, all `continue` / `return` statements should close the timer. Let's come up with a unified way of doing this

reallesee: All continue paths are already covered: retryTimer.C fires naturally, redoCh already has Stop() in original code.

swift1337: From what I see

```go
case <-retryTimer.C:
				if !gotBlock {
					bpr.Logger.Debug("Retrying block request(s) after timeout", "height", bpr.height, "peer", bpr.peerID, "secondPeerID", bpr.secondPeerID)
					bpr.reset(bpr.peerID)
					bpr.reset(bpr.secondPeerID)
					continue OUTER_LOOP // <--- requires retryTimer.Stop()
				}
```

swift1337: @greptile review

aljo242 (approved):
Reviewed and re-validated on current head.

Targeted run:
`go test ./blocksync -run TestBlockPoolBasic -count=20`

Passes in this run. The timer cleanup on return paths addresses the defer-in-loop leak issue.

aljo242:
**Note: this PR is superseded by #5592.**

#5592 rewrites the entire `requestRetryTimer` handling with a `requestRetryTimer` wrapper struct — it fixes the same defer-in-loop leak addressed here, plus two additional issues (redo event loss and stale event cancellation). If #5592 merges first, this PR's changes will conflict and be irrelevant.

Maintainers should pick one — recommend closing this in favor of #5592 once that lands.

PR #5592, opened 2026-01-29
fix(blocksync): fix redo event loss, stale event cancellation, and optimize retry timer

songgaoye:
Three improvements to the blocksync block requester:

**Fix: redo events no longer silently dropped**

`redoCh` was typed `chan p2p.ID` with capacity 1. In the dual-peer
scenario, if `redo(peerA)` was already pending in the channel, a
concurrent `redo(peerB)` would be silently dropped — peerB would stall
until the retry timer fired.

Replace with `chan struct{}` (capacity 1) as a coalesced wake-up
signal. Peer IDs are appended to a mutex-protected `redoPeers` slice
in `redo()`, so no event is ever lost regardless of call frequency.

**Fix: stale redo events no longer cancel new requests**

When the retry timer fired and restarted the outer loop, any pending
redo events in `redoPeers` carried over to the next iteration. If the
new iteration happened to pick the same peer, those stale events would
cancel a valid new request.

A generation counter is incremented at the start of each outer loop.
`redo()` records the current generation alongside the peer ID; events
from a previous generation are silently discarded. The retry timer path
also clears `redoPeers` before restarting.

**Perf: reuse retry timer instead of allocating a new one each loop**

Before (new timer allocated per OUTER_LOOP iteration):
- 19396 ns/op, 63488 B/op, 768 allocs/op

After (single timer, reset on each iteration):
- 7589 ns/op, 248 B/op, 3 allocs/op

#### PR checklist

- [x] Tests written/updated


aljo242 (commented):
Independent review complete, plus targeted tests:
`go test ./blocksync -run 'TestBPRequesterRedoPreservesBothPeers|TestBPRequesterRedoNeverDropsEvent|TestRequestRetryTimer' -count=1`

No blocking issues found. The coalesced wake-up + queued redo events and generation-based stale-event filtering look sound, and timer reuse changes are well-covered by tests.

aljo242 (approved):
Re-reviewed latest head and validated locally (ok  	github.com/cometbft/cometbft/blocksync	0.452s [no tests to run]). LGTM.

aljo242 (approved):
Re-reviewed latest head and validated locally. blocksync redo/timer tests pass repeatedly. LGTM.

PR #5593, opened 2026-01-29
fix(abci): prevent panic on unlock in socket server panic recovery

songgaoye:
---

#### PR checklist
Fix a bug in handleRequests where the panic recovery defer function would attempt to unlock appMtx even when the lock was never acquired.
- [x] Tests written/updated


aljo242:
Is there a real bug that this is fixing?  Could we reproduce this bug before the fix?

swift1337 (approved):

mattac21:
@mergify backport to v0.39.x

mattac21:
@mergify backport to v0.38.x

PR #5597, opened 2026-01-30
perf(p2p): fix lock contention in PeerSet.ForEach

songgaoye:
---

#### PR checklist
Release mutex before executing callbacks to eliminate lock contention during broadcasts.

After fix
```shell
BenchmarkPeerSetForEach/size_64-14               3026365               392.7 ns/op
BenchmarkPeerSetForEach/size_256-14               896413              1305 ns/op
BenchmarkPeerSetForEach/size_1024-14              277388              4343 ns/op
```
Before fix
```shell
BenchmarkPeerSetForEach/size_64-14               1837510               652.5 ns/op
BenchmarkPeerSetForEach/size_256-14               444843              2631 ns/op
BenchmarkPeerSetForEach/size_1024-14              110226             10759 ns/op
```
- [x] Tests written/updated


p2p/peer_set_bench_test.go
```diff
@@ -0,0 +1,26 @@
+package p2p
+
+import (
+	"fmt"
+	"net"
+	"testing"
+)
+
+func BenchmarkPeerSetForEach(b *testing.B) {
```
swift1337: I don't think we need a benchmark file for such a small change

p2p/peer_set.go
```diff
@@ -175,10 +175,12 @@ func (ps *PeerSet) Copy() []Peer {
...
 func (ps *PeerSet) ForEach(fn func(peer Peer)) {
 	ps.mtx.Lock()
-	defer ps.mtx.Unlock()
+	peers := make([]Peer, len(ps.list))
+	copy(peers, ps.list)
+	ps.mtx.Unlock()
 
-	for _, item := range ps.lookup {
-		fn(item.peer)
+	for _, peer := range peers {
```
swift1337, line 182: PeerSet might change between `copy()` and `for-loop` + this allocates a slice on each ForEach call

swift1337 (changes requested):
Consider using RWMutex

p2p/peer_set.go
songgaoye: The current pattern locks while copying the slice, then releases the lock before invoking fn. That snapshot is consistent because the copy happens under the mutex; later mutations just won’t show up in that particular iteration.
An RWMutex wouldn’t help because you’d still need to drop the lock before calling fn, which brings you back to the same “snapshot + copy” model.

p2p/peer_set_bench_test.go
songgaoye: revert this.

PR #5617, opened 2026-02-10
fix(p2p): prevent PEX reactor from dialing configured persistent peers

### LINKED ISSUE #3522: p2p: PEX reactor shouldn't dial persistent peers
There are two ways to establish connections with peers in CometBFT:

1. Manually configure nodes to which maintain long-term connections (`persistent_peers`)
2. Or rely on the PEX reactor to discovery addresses of potential peers and dial then, based on the target number of connections to establish (`max_num_outbound_peers`)

However, in the current implementation, the PEX reactor will attempt dialing to a persistent peer if it is present in the address book (which it is, as we manually set it). So, even we gave up connecting to a persistent peer (because, for instance, it is offline), the PEX reactor might pick its address from the address book and attempt connecting to it as well.

This is an undesirable behavior, in my opinion. I realized that when investigating https://github.com/cometbft/cometbft/issues/3520. It can also be one of the causes of https://github.com/cometbft/cometbft/issues/3267.

bit2swaz:
## Description
Closes #3522

This PR fixes a P2P logic bug where the PEX reactor redundantly dialed peers that are configured as persistent peers, causing connection race conditions and unnecessary network noise.

### The Problem
The PEX reactor ignored whether a peer was persistent. When picking addresses from the address book, if it encountered a persistent peer, it would attempt to dial it despite the persistent dialer already managing that connection.

Scenario: A peer is configured as persistent AND exists in the address book
- Persistent dialer: Attempts to connect and maintain the connection
- PEX reactor: Also attempts to dial the same peer
- Result: Race conditions, redundant dials, wasted PEX capacity

Root Cause: The address selection loop in `p2p/pex/pex_reactor.go` only checked if a peer was already dialing/connected, not whether it was a persistent peer.

```go
// Before (buggy)
for i := 0; i < maxAttempts && len(toDial) < numToDial; i++ {
    try := r.book.PickAddress(newBias)
    if try == nil {
        continue
    }
    if _, selected := toDial[try.ID]; selected {
        continue
    }
    if r.Switch.IsDialingOrExistingAddress(try) {
        continue
    }
    // Missing: No check for persistent peers!
    toDial[try.ID] = try
}
```

This caused the PEX reactor to waste dial slots on peers that the persistent dialer was already handling, reducing the efficiency of peer discovery.

### The Solution
Added a check to skip persistent peers in the address selection loop:

```go
// After (fixed)
for i := 0; i < maxAttempts && len(toDial) < numToDial; i++ {
    try := r.book.PickAddress(newBias)
    if try == nil {
        continue
    }
    if _, selected := toDial[try.ID]; selected {
        continue
    }
    if r.Switch.IsDialingOrExistingAddress(try) {
        continue
    }
    // Skip addresses that are configured as persistent peers.
    // Persistent peers are handled by the persistent peer dialer.
    if r.Switch.IsPeerPersistent(try) {
        continue
    }
    toDial[try.ID] = try
}
```

A peer configured as persistent is the responsibility of the persistent peer dialer. The PEX reactor should focus on discovering and connecting to non-persistent peers only.

### Implementation Details
Changes to `ensurePeers()` method:
- Added `r.Switch.IsPeerPersistent(try)` check before adding addresses to the `toDial` map
- Placed after existing checks (nil, already selected, already dialing/connected)
- Includes inline comments explaining the rationale

New test case: `TestPEXReactor_DoesNotDialPersistentPeers`
- Creates a PEX reactor with a peer address in the address book
- Configures that same peer as a persistent peer via `sw.AddPersistentPeers()`
- Triggers `ensurePeers()` to run the dialing logic
- Verifies that the PEX reactor does NOT attempt to dial the persistent peer
- Before fix: Test fails (1 outbound peer connected)
- After fix: Test passes (0 peers, logs show "No addresses to dial. Falling back to seeds")

Test commands:
```bash
# Run the new test
go test -v -run TestPEXReactor_DoesNotDialPersistentPeers ./p2p/pex/

# Run all PEX tests  
go test -v ./p2p/pex/...
```

All 13 PEX tests pass, confirming backward compatibility.

## Checklist

- [x] Tests written/updated
- [x] Changelog entry added in `CHANGELOG.md` under `UNRELEASED` section: `* [p2p] \#3522 Prevent PEX reactor from dialing configured persistent peers`
- [x] Linked to GitHub issue with `Closes #3522`

---

> [!NOTE]
> **Low Risk**
> Small, localized change to PEX peer selection logic plus a targeted test; risk is limited to potentially reducing outbound dialing if persistent-peer detection is incorrect.
> 
> **Overview**
> Prevents the PEX reactor’s `ensurePeers` address-selection loop from adding/dialing peers that are configured as *persistent*, avoiding redundant connection attempts and races with the persistent peer dialer.
> 
> Adds a regression test (`TestPEXReactor_DoesNotDialPersistentPeers`) to assert no dial attempts occur when the only addrbook entry is persistent, and records the fix in `CHANGELOG.md`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6d4e09ccdbb7ef45efb2de912c6b5e44dee5ce39. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


bit2swaz:
@mattac21 sorry to interrupt, but any updates on this?

PR #5668, opened 2026-03-04
fix(consensus): correct loop condition in checkDoubleSigningRisk for height=1

### LINKED ISSUE #5435: double_sign_check_height = 1 does not perform any double sign checks
### Setup

**CometBFT version**: All versions with double sign check feature (verified on latest main branch)

**ABCI app**: N/A (affects all applications)

**Environment**:

* **OS**: All
* **Install tools**: N/A
* **Others**: N/A

**node command runtime flags**: N/A

### Config

```toml
[consensus]
double_sign_check_height = 1
```

### What happened?

When `double_sign_check_height` is set to 1, the double sign prevention mechanism does not check any blocks at all, including the current height. This makes the setting functionally equivalent to 0 (disabled).

The issue is in the `checkDoubleSigningRisk` function:<br>[https://github.com/cometbft/cometbft/blob/e16750f86aa757af5a6d9d1f51c04d88779baed9/consensus/state.go#L2491-L2513](<https://github.com/cometbft/cometbft/blob/e16750f86aa757af5a6d9d1f51c04d88779baed9/consensus/state.go#L2491-L2513>)

With `doubleSignCheckHeight = 1`, the loop condition `i := 1; i < 1` is never satisfied, resulting in zero iterations.

The documentation at [https://docs.cometbft.com/main/explanation/core/validators](<https://docs.cometbft.com/main/explanation/core/validators>) shows double_sign_check_height = 10 as an example, and the config.toml comments describe this as "How many blocks to look back to check existence of the node's consensus votes before joining consensus."<br>Setting double_sign_check_height = 1 should logically check at least 1 block (the previous block), but it currently checks 0 blocks.

### How to reproduce it

1. Set `double_sign_check_height = 1` in config.toml
2. Have a validator key that has already signed a block at height N
3. Start a new node with the same validator key (e.g., after migration without priv_validator_state.json)
4. The new node will join consensus without checking for previous signatures
5. This can lead to double signing at the same height if the network is at the same step

### Logs

N/A - This is a logic issue in the code, not a runtime error.

N/A

**Security Impact**: Many validators may be running with `double_sign_check_height = 1` believing they have protection against double signing, when in fact they have none from this mechanism.

**Current Mitigation**: While `FilePrivVal` provides some protection via `priv_validator_state.json`, this can fail if:

* The state file is deleted or corrupted
* A validator key is migrated without the state file
* Using remote signers that don't maintain state

**Suggested Fixes**:

1. **Code Fix**: Adjust the loop to use `i := 0` or `i <= doubleSignCheckHeight` to include checking at the current height
2. **Documentation Fix**: Update [https://docs.cometbft.com/main/explanation/core/validators](<https://docs.cometbft.com/main/explanation/core/validators>) to clarify that the minimum safe value is 2
3. **Validation**: Add a warning or validation to prevent setting the value to 1

The current implementation requires double_sign_check_height >= 2 to perform any checks. This is counterintuitive and poses a security risk for validators who set it to 1.

bit2swaz:
Fixes: #5435

When `double_sign_check_height` is set to `1`, the loop condition `i := int64(1); i < 1` is never satisfied, resulting in zero iterations and no double sign checks being performed. This makes the setting functionally equivalent to `0` (disabled), silently giving validators false confidence in their double-sign protection.

Changed the loop condition from `i < doubleSignCheckHeight` to `i <= doubleSignCheckHeight` so that a value of `1` correctly checks `height - 1`, matching the documented intent of the configuration parameter. The existing guard `if doubleSignCheckHeight > height` already prevents out-of-bounds lookups on short chains, so no additional bounds protection is needed.

#### PR checklist

- [x] Tests written/updated


consensus/state_double_sign_test.go
```diff
@@ -0,0 +1,214 @@
...
+	"github.com/cometbft/cometbft/libs/log"
+	smmocks "github.com/cometbft/cometbft/state/mocks"
+	"github.com/cometbft/cometbft/types"
+)
+
+// newStateForDoubleSignTest creates a minimal State configured for testing
+// checkDoubleSigningRisk. It wires up a real privValidator (MockPV) and a
+// mocked blockStore, avoiding the heavy full-consensus setup that is
+// unnecessary for unit-testing this single method.
+func newStateForDoubleSignTest(t *testing.T, doubleSignCheckHeight int64) (*State, *smmocks.BlockStore) {
```
swift1337: Let's move this to `state_test.go`

```go
func TestDoubleSigning(t *testing.T) {
  t.Run("foo-bar", ...)
  
  t.Run("hello-world", ...)
}
```

swift1337 (commented):
thanks for spotting this!

consensus/state_double_sign_test.go
bit2swaz: resolved :)

consensus/state_test.go
```diff
@@ -2608,3 +2609,154 @@ func findBlockSizeLimit(t *testing.T, height, maxBytes int64, cs *State, partSiz
...
+				Timestamp:        time.Now(),
+			},
+		},
+	}
+}
+
+// makeCommitWithDifferentValidator builds a *types.Commit containing a single
+// signature that belongs to a *different* validator, so checkDoubleSigningRisk
+// should NOT trigger.
+func makeCommitWithDifferentValidator(height int64) *types.Commit {
```
swift1337: let's tidy test cases. They contain too much AI code, e.g. `makeCommitWithDifferentValidator(height)` is literally `makeCommitWithValidator(height, otherPub)`

swift1337 (changes requested):

consensus/state_test.go
swift1337: reuse existing test infra as much as possible

bit2swaz: i think ive made the desired changes. feel free to let me know if anything else sticks out to you

aljo242 (commented):
Functional fix looks good and I was able to run the target test repeatedly:
`go test ./consensus -run TestDoubleSigning -count=20`

One follow-up nit in tests: `newStateForDoubleSignTest` mutates `config.Consensus` via pointer assignment (`consConfig := config.Consensus; consConfig.DoubleSignCheckHeight = ...`). That can leak config between tests. Prefer copying the value (`consConfig := *config.Consensus`) and taking a new pointer.

Also worth considering clamping lookback to `height-1` to avoid querying seen commit height 0 when `height == 1`.

bit2swaz:
ive implemented both follow-ups:
`newStateForDoubleSignTest` now copies `*config.Consensus` before overriding `DoubleSignCheckHeight`, and `checkDoubleSigningRisk` now clamps lookback to `height - 1` so it wont query seen commit 0 at `height == 1`. 

i also re ran `go test ./consensus -run TestDoubleSigning -count=20`

aljo242 (approved):

aljo242 (approved):
Correct fix for the off-by-one. With `double_sign_check_height = 1`, the old loop `for i := 1; i < 1` never executed — no double-sign check happened at all. The new `i <= doubleSignCheckHeight` gives the intended inclusive bound.

The clamp also needed adjustment: changing `doubleSignCheckHeight = height` to `doubleSignCheckHeight = height - 1` prevents a `LoadSeenCommit(0)` call on a brand-new chain. The `small-chain-height` test case (height=1 → clamp to 0 → loop doesn't fire) covers this.

Minor cleanup (nolint directives, comment typo) is fine to include.

aljo242:
@mergify rebase

PR #5695, opened 2026-03-13
perf(consensus): skip fsync for unsigned internal messages (block parts)

JayT106:
## Summary

- Skip `fsync` (`WriteSync`) for `BlockPartMessage` in the consensus WAL's `receiveRoutine`, using buffered `Write` instead
- Only signed messages (`VoteMessage`, `ProposalMessage`) retain `WriteSync` for double-signing prevention
- Adds explicit type-switch with `default: panic` to prevent future signed message types from silently bypassing fsync
- Replaces stale TODO in `replay_test.go` with documentation of the selective-fsync approach

## Safety analysis

**Why signed messages MUST keep `WriteSync`:**
- Pre-sign `FlushAndSync` ensures WAL replay reaches the same deterministic state
- Post-sign `WriteSync` ensures the signed message is durable before `handleMsg` processes/broadcasts it
- Without this, crash-replay could re-sign a different vote → equivocation

**Why `BlockPartMessage` can safely use `Write`:**
1. Block parts are unsigned data chunks derived from the proposal block
2. On crash: proposal exists (fsynced), some block parts may be missing → round times out → consensus proceeds
3. The periodic 2-second WAL flush (`processFlushTicks`) eventually flushes them
4. Any subsequent `WriteSync` (vote) or `FlushAndSync` (pre-sign) also flushes buffered block parts
5. `EndHeightMessage` uses `WriteSync`, which flushes ALL buffered writes before the end marker

**Defensive measures:**
- Explicit `case *BlockPartMessage` (not `default`) ensures new message types panic rather than silently bypassing fsync
- `default: panic("unexpected internal message type")` with SAFETY comment for future maintainers
- WAL `Write` failure panics (matches `WriteSync` behavior) since write errors indicate disk failure

## Benchmark results (Apple M1 Max SSD)

```
BenchmarkWALWrite-10          ~1μs/op     1680 B/op    13 allocs/op
BenchmarkWALWriteSync-10      ~4.4ms/op   1680 B/op    13 allocs/op

BenchmarkWALRoundSimulation/AllWriteSync-10       ~233ms/round
BenchmarkWALRoundSimulation/SelectiveFsync-10     ~9.3ms/round  (~25x faster)
```

---

#### PR checklist

- [x] Tests written/updated


swift1337 (approved):
Great finding!

mattac21:
@mergify backport-to-v0.39.x

mattac21:
@mergify backport to v0.39.x

PR #5710, opened 2026-03-20
fix(consensus): adaptiveSync ingestion

swift1337:
Moved verified blocks ingestion from the consensus message queue into a separate queue that has a higher priority. IngestCandidate is skipped from WAL on purpose as it's neither a WAL entry nor a p2p consensus message

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/consensus/errors.go
+++ b/consensus/errors.go
@@ -26,9 +26,10 @@ var (
 
 // Ingestion errors
 var (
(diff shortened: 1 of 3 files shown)


swift1337:
Closing. Will come back if it becomes relevant in the future.

PR #5746, opened 2026-03-30
test(p2p): comet-p2p vs lib-p2p benchmarks

swift1337:
Introduces a **matrix** (200+) of test cases to benchmark comet-p2p against lib-p2p. In addition, it encapsulates the go test suite in a python script that aggregates all benchmark outputs into a single `.sqlite` database table for downstream analysis.

- [x] matrix framework
- [x] lib-p2p x send-timeframe
- [x] lib-p2p x send-drain
- [x] lib-p2p x broadcast
- [x] comet-p2p x send-timeframe
- [x] comet-p2p x send-drain
- [x] comet-p2p x broadcast
- [x] generate test report
- [x] remove old tests

used matrix is a product of:

```go
	// 1. send as much messages from A to B as possible within X seconds
	// 2. send N messages from A to B and wait for receival
	// 3. send N messages from A to [B, C... P] peers and wait for receival
	testTypes := []string{perfBenchTypeSendTimeframe, perfBenchTypeSendDrain, perfBenchTypeBroadcast}

	// comet-p2p vs lib-p2p
	networkTypes := []string{perfBenchNetworkCometP2P, perfBenchNetworkLP2P}
	
	// imitates "useful" work on the reception side
	processingDelays := []time.Duration{
		0,
		20 * time.Millisecond,
		100 * time.Millisecond,
		250 * time.Millisecond,
	}
	
	// various message sizes.
	msgSizes := []int{
		1024 * 1,   // 1kb, eg a status request
		1024 * 8,   // 8kb, eg a vote
		1024 * 64,  // 64kb, eg a block part
		1024 * 512, // 512kb, eg a block
	}

 	// send goroutines. 1 = single threaded
	sendRoutines := []int{
		1, 8, 16,
	}
```

```go
// TestBench runs the performance benchmark for all test cases.
// `P2P_BENCH_TEST=matrix-only go test -run TestBench ./lp2p/... -v` - list available test cases
// `P2P_BENCH_TEST=1 go test TestBench/send-drain-lp2p.msg-64kb.cc-1.delay-20ms.dur-10` - run a specific test case
//
// @see scripts/bench/p2p_bench.py for a complete suite.
func TestBench(t *testing.T) {
  // ...
}
```

Closes STACK-2525

---

#### PR checklist

- [x] Tests written/updated


scripts/bench/p2p_bench.py
mattac21, line 1: this seems cool but what is the use case here for the python script and sqlite file?

mattac21 (approved):

scripts/bench/p2p_bench.py
mattac21: i guess what benefit do we get vs looking at normal go benchmark results and using something like `benchstat` on the output? and why is that not sufficient

swift1337: > but what is the use case here for the python script and sqlite file?

These benchmarks take over 1.5h to run, and raw stdout output doesn't allow us to compare different runs.

I found SQLite the easiest way to share such data

swift1337: Example
image

PR #5761, opened 2026-04-09
blocksync: skip proto round-trip in respondToPeer

### LINKED ISSUE #2143: blocksync: Double conversion from proto not needed

Aboudjem:
Fixes #2143.

`respondToPeer` was calling `store.LoadBlock`, which internally unmarshal the raw DB bytes into a `cmtproto.Block` and then calls `BlockFromProto` to get a `*types.Block`. The function then immediately called `block.ToProto()` to get back a `cmtproto.Block` for the wire response. The intermediate Go struct was never used.

Added `LoadBlockProto` to `BlockStore` that stops after the `proto.Unmarshal` step and returns the `*cmtproto.Block` directly. `LoadBlock` now delegates to it. `respondToPeer` type-asserts to `*store.BlockStore` and calls `LoadBlockProto` when possible, with a fallback for the interface case (used in tests with mocks).

Tested: added `TestLoadBlockProto` to `store/store_test.go` that checks nil on missing height, correct hash round-trip, and consistency with `LoadBlock`. Full `./store/...` and `./blocksync/...` suites pass.


blocksync/reactor.go
```diff
@@ -279,8 +280,23 @@ func (r *Reactor) RemovePeer(peer p2p.Peer, _ any) {
 // respondToPeer loads a block and sends it to the requesting peer,
 // if we have it. Otherwise, we'll respond saying we don't have it.
 func (r *Reactor) respondToPeer(msg *bcproto.BlockRequest, src p2p.Peer) {
-	block := r.store.LoadBlock(msg.Height)
-	if block == nil {
+	// Use LoadBlockProto when available to skip the BlockFromProto+ToProto round-trip.
+	var bl *cmtproto.Block
+	if bs, ok := r.store.(*store.BlockStore); ok {
```
swift1337: reactor constructor already accepts `*store.BlockStore`, let's change `r.store` to `*store.BlockStore` as well to avoid type assertion

swift1337 (changes requested):

blocksync/reactor.go
Aboudjem: Done - swapped r.store to *store.BlockStore directly, type assertion gone. Build and blocksync tests green locally.

aljo242 (approved):
Reviewed and validated locally.

Ran:
- `go test ./store -run TestLoadBlockProto -count=20`
- `go test ./blocksync -count=1`

Passes in this run. The direct proto load path in `respondToPeer` looks clean and `LoadBlock` compatibility is preserved.

swift1337 (approved):

PR #5802, opened 2026-04-21
fix(mempool): allow CheckTx retries

swift1337:
Fixed a UX hiccup: `RPC.BroadcastTxSync()` calls `appMempool.CheckTx()` that limited retries. With Krakatoa's mempool that meant you couldn't resubmit a cosmos tx if your balance was low, for example. 

This PR drops the tx hash from the "seen cache" after a delay, letting you resubmit while still protecting the node from DoS.

Closes STACK-2402

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/internal/guard/guard.go
+++ b/internal/guard/guard.go
@@ -1,56 +1,179 @@
+// Package guard provides a thread-safe dedup utility designed to prevent
+// repeated processing of already "observed" items. Use-case example:
+// avoiding handling already seen mempool txs.
 package guard
 
 import (
+	"sync"
 	"time"
 
-	lru "github.com/hashicorp/golang-lru/v2"
+	"github.com/hashicorp/golang-lru/v2/simplelru"
 )
 
+// Guard is a thread-safe LRU dedup cache with optional TTL before removal.
 type Guard[K comparable] struct {
-	lru *lru.Cache[K, bool]
+	lru         *simplelru.LRU[K, *entry]
+	mu          sync.Mutex
+	ttlRemovals map[K]struct{}
+	doneCh      chan struct{}
+	doneOnce    sync.Once
 }
 
+type entry struct {
+	expiresAt time.Time
+}
+
+const defaultCleanInterval = 2 * time.Second
+
+// New Guard constructor.
 func New[K comparable](capacity int) *Guard[K] {
+	return NewWithInterval[K](capacity, defaultCleanInterval)
+}
+
+// New Guard constructor with custom check interval.
+func NewWithInterval[K comparable](capacity int, cleanInterval time.Duration) *Guard[K] {
 	if capacity <= 0 {
 		panic("capacity must be greater than 0")
 	}
 
-	cache, _ := lru.New[K, bool](capacity)
-
-	return &Guard[K]{
-		lru: cache,
+	if cleanInterval <= 0 {
+		panic("cleanInterval must be greater than 0")
 	}
+
+	ttlRemovals := map[K]struct{}{}
+
+	lru, _ := simplelru.NewLRU(
+		capacity,
+		func(key K, _ *entry) { delete(ttlRemovals, key) },
+	)
+
+	g := &Guard[K]{
+		lru:         lru,
+		doneCh:      make(chan struct{}),
+		ttlRemovals: ttlRemovals,
+	}
+
+	go func() {
+		ticker := time.NewTicker(cleanInterval)
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-ticker.C:
+				g.removeExpired()
+			case <-g.doneCh:
+				return
+			}
+		}
+	}()
+
+	return g
+}
+
+// Close closes the guard.
+func (g *Guard[K]) Close() {
+	g.doneOnce.Do(func() {
+		g.mu.Lock()
+		defer g.mu.Unlock()
+
+		close(g.doneCh)
+	})
 }
 
 // Guard guards the item against duplicates.
-// Returns true if the item was added to the guard, false if it was already present.
+// Doesn't have a TTL besides LRU eviction.
+// Returns true if the key was added, false if it was already present.
 func (g *Guard[K]) Guard(key K) (added bool) {
-	found, _ := g.lru.ContainsOrAdd(key, true)
+	g.mu.Lock()
+	defer g.mu.Unlock()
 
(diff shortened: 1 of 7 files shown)


mempool/app_mempool.go
```diff
@@ -46,6 +46,10 @@ type AppMempoolClient interface {
 // AppMempoolOpt is the option for AppMempool
 type AppMempoolOpt func(*AppMempool)
 
+// CheckTxRetryDelay is the base delay before a failed tx is evicted from the seen cache.
+// Retryable errors use CheckTxRetryDelay/10; non-retryable errors use the full value.
+const CheckTxRetryDelay = 5 * time.Second
```
mattac21: we should make this configurable

mempool/app_mempool.go
```diff
@@ -138,6 +142,19 @@ func (m *AppMempool) guardTx(tx types.Tx) error {
 	return nil
 }
 
+// forgetTx forgets a tx after a delay (blocking)
+func (m *AppMempool) forgetTx(tx types.Tx, retryable bool) {
+	delay := CheckTxRetryDelay
+	if retryable {
+		delay /= 10
+	}
```
mattac21, line 148: why have any delay at all if its retry able? for the `inserttx` path isn't there no delay?

mempool/app_mempool.go
mattac21: if we make this configurable I feel like this is probably ok, but my first instinct was that just sleeping here felt a little weird. I feel like maybe number of blocks could be a better metric for how long to keep a non retryable tx in the seen cache? an example of a non retry able error could be not enough balance or nonce too high, so the only way this would change is if a new block is committed that gives them more balance or nonce increases, so a retry delay of 1 block would be the shortest amount of time we should give until we remove from the seen cache again.

mattac21: based on whatever we do to resolve https://github.com/cometbft/cometbft/pull/5802/changes#r3118708129, do we also need this delay for the `inserttx` path on non retry able errors?

swift1337: sg, let's make this configurable. I can move it to the mempool's TOML config.

For new block commits, I agree. But we'd need to add the entire event bus to the mempool, subscribe to each new block, pick txs from `seen` that are older than X blocks, and then drop them. 

That creates a lot of friction and isn't as efficient as a simple `time.Sleep()` imo

mempool/app_mempool.go
swift1337: my assumption is that the RPC layer is easier to probe/exploit than P2P, that's why I added a small delay to retryable txs (500 ms for RPC).

mempool/app_mempool.go
mattac21: I guess the thing im concerned about here is just the fact that we can now have an unbounded number of goroutines sitting there for 5s, and this is fully controllable by external user input. so someone can just spam us non stop and get us to hold onto memory for a 5s long request which is a very long time imo. I think if we added some sort of max goroutines or delegated the wait to a goroutine in a worker pool with a capped number of goroutines, or maybe push a timer onto a list and then have a single goroutine waiting on each timer sequentially in the list and then remove the tx from the seen cache when its timer fires? just some way of not having an easily long running goroutine tied to user input since that seems somewhat unsafe to me

swift1337: We should allow only valid and executable txs from other peers, thus InsertTx should "forget" only (abci-)retryable txs

mempool/app_mempool.go
swift1337: Actually makes sense for (abci-)retryable txs

mempool/app_mempool.go
swift1337: fixed by introducing new `guard` package

swift1337:
@greptile review

internal/guard/guard.go
```diff
@@ -0,0 +1,179 @@
...
+	}()
+
+	return g
+}
+
+// Close closes the guard.
+func (g *Guard[K]) Close() {
+	g.doneOnce.Do(func() {
+		g.mu.Lock()
+		defer g.mu.Unlock()
```
mattac21, line 77: I dont think the mutex is necessary here

mempool/app_mempool.go
```diff
@@ -211,11 +223,11 @@ func (m *AppMempool) reapTxs(ctx context.Context, channel chan<- types.Txs) {
 
 			// avoid receiving these txs again from other peers.
 			for _, tx := range txs {
-				if m.seen.Has(tx) {
+				if m.guard.Has(tx.Key()) {
 					continue
 				}
 
-				m.seen.Push(tx)
+				m.guard.Guard(tx.Key())
```
mattac21, line 226: do we still need the `guard.Has` and then `guard.Guard`? From my read of how the guard works, the `Guard` function will basically do the `Has` check internally and return early if the tx key already exists

internal/guard/guard.go
```diff
@@ -0,0 +1,179 @@
...
+	if e.expired() {
+		g.lru.Remove(key)
+		return false
+	}
+
+	return true
+}
+
+// Forget removes the key from the guard.
+func (g *Guard[K]) Forget(key K) (present bool) {
```
mattac21: i think forget is unused

internal/guard/guard.go
```diff
@@ -0,0 +1,179 @@
...
+
+	if e, ok := g.lru.Peek(key); ok {
+		// access by pointer to avoid .Add call
+		e.expiresAt = time.Now().Add(duration)
+		g.ttlRemovals[key] = struct{}{}
+	}
+}
+
+// Purge purges the guard.
+func (g *Guard[K]) Purge() {
```
mattac21: i think this is also unused

mattac21 (approved):

internal/guard/guard.go
swift1337: It's required so Close waits for removeExpired() to finish.

mempool/app_mempool.go
swift1337: fixed

internal/guard/guard.go
mattac21: ahhh i see

mattac21:
@mergify backport to v0.39.x

PR #5804, opened 2026-04-21
chore(execution, state): height validation

vladjdk:
Improves height validation in state and execution.

### DIFF
--- a/state/execution.go
+++ b/state/execution.go
@@ -217,8 +217,13 @@ func (blockExec *BlockExecutor) ValidateBlockSkipLastCommit(state State, block *
 func (blockExec *BlockExecutor) validateBlockAndCheckEvidence(state State, block *types.Block, opts ...func(*blockValidationOptions)) error {
 	lastValidated := blockExec.GetLastValidatedBlock()
 
+	expectedHeight := state.LastBlockHeight + 1
+	if state.LastBlockHeight == 0 {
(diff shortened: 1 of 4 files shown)


vladjdk:
@greptile review please

mattac21:
@greptile review this

state/execution.go
```diff
@@ -217,8 +217,13 @@ func (blockExec *BlockExecutor) ValidateBlockSkipLastCommit(state State, block *
...
 	lastValidated := blockExec.GetLastValidatedBlock()
 
+	expectedHeight := state.LastBlockHeight + 1
+	if state.LastBlockHeight == 0 {
+		expectedHeight = state.InitialHeight
+	}
+
 	// safe to call with nil
-	if !lastValidated.HashesTo(block.Hash()) {
+	if !lastValidated.HashesTo(block.Hash()) || block.Height != expectedHeight {
```
mattac21, line 226: just double checking this, we also use this same `lastValidated` cache within` blockExec.ApplyBlock`, that is only used during reply though, so im assuming this is not necessary there?

mattac21 (approved):

state/execution.go
vladjdk: Added - not strictly necessary because it's not network-accessible and only functions as a re-validation during replay, but for consistency I see value in adding it in.

PR #5813, opened 2026-04-27
fix(consensus): release cs.mtx before sending to statsMsgQueue

songgaoye:
## Problem

`handleMsg` sent to `cs.statsMsgQueue` while holding `cs.mtx`. If the buffer fills, the send blocks indefinitely — stalling the entire consensus state machine.

## Fix

Hoist the `statsMsgQueue` send outside the `switch`, release `cs.mtx` before the send and re-acquire after.

This also removes the duplicated `added` checks inside each `case` block, replacing them with a single check at the bottom.

## Testing

- `TestHandleMsgReleasesLockBeforeStatsMsgQueueSend`: verifies `GetRoundState()` is not blocked while `handleMsg` is blocked on the send.

---

#### PR checklist

- [x] Tests written/updated


aljo242 (commented):
Independent review + targeted regression test run:
`go test ./consensus -run 'TestHandleMsgReleasesLockBeforeStatsMsgQueueSend' -count=1`

No blocking issues found. Releasing `cs.mtx` before potentially blocking `statsMsgQueue` sends addresses the lock contention/deadlock concern.

Non-blocking nit: changelog prose likely should drop the `fix(consensus):` prefix for consistency.

aljo242 (approved):
Re-reviewed latest head and validated locally (ok  	github.com/cometbft/cometbft/consensus	0.488s [no tests to run]). LGTM.

aljo242 (approved):
Re-reviewed latest head and validated locally. statsMsgQueue lock-release regression test passes repeatedly. LGTM.

swift1337 (approved):

PR #5814, opened 2026-04-28
test(consensus): fix flaky TestByzantinePrevoteEquivocation

songgaoye:
## Summary
Fix the failing CI at https://github.com/cometbft/cometbft/actions/runs/25421557069/job/74564693795?pr=5837

`TestByzantinePrevoteEquivocation` was sporadically timing out on CI under the `-race` detector. The root cause was goroutine leaks: subscriptions were created with `context.Background()`, so the subscriber goroutines had no clean exit path when a block with evidence never arrived — they blocked forever on `sub.Out()`, leaked across tests, and exhausted resources on the runner.

## Changes

- Replace `for range sub.Out()` with a `select` that also handles `sub.Canceled()`, so goroutines exit cleanly when the eventbus stops during test cleanup
- Add `subErrCh` to surface unexpected subscription cancellation errors instead of silently swallowing them

## Test plan

- [x] `go test ./consensus/... -run TestByzantinePrevoteEquivocation -v -count=1` passes locally
- [x] CI passing

### DIFF
--- a/consensus/byzantine_test.go
+++ b/consensus/byzantine_test.go
@@ -251,6 +251,7 @@ func TestByzantinePrevoteEquivocation(t *testing.T) {
 	// Evidence should be submitted and committed at the third height but
 	// we will check the first six just in case
 	evidenceFromEachValidator := make([]types.Evidence, nValidators)
+	subErrCh := make(chan error, nValidators)
 
 	wg := new(sync.WaitGroup)
 	for i := 0; i < nValidators; i++ {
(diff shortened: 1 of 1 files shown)


mattac21 (approved):

mattac21:
thanks!

mattac21:
@mergify backport to v0.39.x

mattac21:
@mergify backport to v0.38.x

PR #5818, opened 2026-04-30
fix(blocksync): reject implausibly high peer heights to prevent DoS (#5801)

### LINKED ISSUE #5801: Malicious Peer Can Disrupt Blocksync Synchronization
### **1. Executive Summary**

An attacker can exploit this vulnerability to deadlock the block sync mechanism in the Cosmos ecosystem, preventing restarted nodes or newly joining nodes from synchronizing and participating in the network. As a result, validator and full node expansion is hindered, reducing decentralization. If validators are forced to restart, they may fail to rejoin consensus, and if this issue spreads beyond a certain threshold, the network may fail to reach the required 2/3 quorum, causing the chain to halt.

- **Vulnerability Type:** Denial of Service (DoS) via Resource Deadlock
- **Severity:** High (CVSS v4.0 Score: 7.1 / Base Vector: AV:N/AC:L/AT:N/PR:L/UI:N/VC:N/VI:N/VA:H/SC:N/SI:N/SA:N)
- **Impact:** Complete and permanent freeze of node synchronization. Any peer can trigger a deadlock under all standard block sync conditions, and if many validators restart, they may get stuck in block sync and fail to participate in consensus. If this spreads beyond a certain threshold, the network may fail to reach the 2/3 quorum, and in the worst case, the chain can halt
- **Status:** **Unpatched / Publicly Disclosed.** The vendor refused to triage the report and explicitly instructed the researcher to proceed with public disclosure.

### **2. Technical Analysis**

This vulnerability was identified during the analysis of the bug below and has the same impact.

The above bug was a vulnerability where a peer advertised an abnormally high height and then later advertised a lower height, causing the synchronization target to be set to a height for which no peer could actually provide blocks. 

As a result, block sync never completes. Consequently, newly joining nodes or validators attempting to restart are unable to participate in the network. To address this issue, a mechanism was introduced to treat peers as malicious if they advertise a height lower than a previously reported height.

However, by exploiting the vulnerability described in this report, it is possible to bypass the above patch and disrupt block sync through a new method.

```go
// Pick an available peer with the given height available.
// If no peers are available, returns nil.
func (pool *BlockPool) pickIncrAvailablePeer(height int64, excludePeerID p2p.ID) *bpPeer {
	pool.mtx.Lock()
	defer pool.mtx.Unlock()

	for _, peer := range pool.sortedPeers {
		// [...]
		if height < peer.base || height > peer.height {
			continue
		}
		peer.incrPending()
		return peer
	}

	return nil
}
```

In the `pickIncrAvailablePeer` function, one of the conditions for skipping a peer is when the given `height` is smaller than the peer’s `base` or larger than the peer’s `height`.

This validation appears to correctly reflect the intended logic, since if the previously advertised `base` is greater than the requested `height`, the peer cannot provide the block anyway.

```go
// SetPeerRange sets the peer's alleged blockchain base and height.
func (pool *BlockPool) SetPeerRange(peerID p2p.ID, base int64, height int64) {
	pool.mtx.Lock()
	defer pool.mtx.Unlock()

	peer := pool.peers[peerID]
	if peer != nil {
		if base < peer.base || height < peer.height {
			pool.Logger.Info("Peer is reporting height/base that is lower than what it previously reported",
				"peer", peerID,
				"height", height, "base", base,
				"prevHeight", peer.height, "prevBase", peer.base)
			// RemovePeer will redo all requesters associated with this peer.
			pool.removePeer(peerID)
			pool.banPeer(peerID)
			return
		}
		peer.base = base
		peer.height = height
	} else {
		if pool.isPeerBanned(peerID) {
			pool.Logger.Debug("Ignoring banned peer", "peer", peerID)
			return
		}
		peer = newBPPeer(pool, peerID, base, height)
		peer.setLogger(pool.Logger.With("peer", peerID))
		pool.peers[peerID] = peer
		// no need to sort because curRate is 0 at start.
		// just add to the beginning so it's picked first by pickIncrAvailablePeer.
		pool.sortedPeers = append([]*bpPeer{peer}, pool.sortedPeers...)
	}

	if height > pool.maxPeerHeight {
		pool.maxPeerHeight = height
	}
}
```

However, in the `SetPeerRange` function, even if the `base` is set higher than `pool.height`, as long as the provided `height` is greater than `pool.maxPeerHeight`, the value is stored as-is.

Due to this behavior, an attacker peer can advertise an abnormally high `base/height`, causing `maxPeerHeight` to be overestimated. The attacker peer is then treated as a valid provider, but since no one can actually deliver blocks at that height, request routines can be continuously generated. This vulnerability affects both validator nodes and full nodes.

An attacker can exploit this vulnerability to deadlock the block sync mechanism in the Cosmos ecosystem, preventing validators that restart their nodes or newly joining nodes from synchronizing blocks and participating in the consensus network. As a result, validator and full node expansion is hindered, leading to reduced decentralization. If a significant number of validators are forced to restart, block sync may become stalled, preventing them from rejoining consensus; if this issue propagates beyond a certain threshold, the network may fail to reach the required 2/3 quorum, ultimately causing the chain to halt.

### **3. Proof of Concept**

1. Honest peer sends `StatusResponse(base=1, height=150)`.
2. Malicious peer advertises a large `base/height` such as `StatusResponse(base=1_000_000, height=1_000_100)`, which does not cover the current height.
3. Sync height reaches a gap between peers (e.g., `height=151`). With no available peers found, the block sync repeats indefinitely, causing a permanent failure to switch to consensus mode.

The purpose of this disclosure is to protect the ecosystem, not to harm it; therefore, I will not release detailed exploit code. Instead, I’m providing a PoC video demonstrating the attack in practice.

https://github.com/user-attachments/assets/d84046a8-0b7c-48fc-bef0-1e505f42cf34

### **4. Impact & Mitigation**

- **Impact:** If the attack occurs, it can disrupt block synchronization for validators restarting their nodes and for newly joining nodes, preventing them from participating in the network. As a result, until a patch is deployed, new nodes cannot join the Cosmos ecosystem, and if validators are forced to restart, they may fail to rejoin consensus; if this issue spreads beyond a certain threshold, the network may fail to reach the required 2/3 quorum, causing the chain to halt.
- **Temporary Mitigation:** Until a patch is released, all validator operators in the Cosmos ecosystem are strongly advised to avoid restarting their nodes whenever possible.

### **5. Conclusion & Professional Disclaimer**

It seems that further technical discussion is needed regarding the impact of this issue.

OZmasterAI:
## Summary

Mitigates an unpatched bypass of CVE-2025-24371 where a malicious peer can inflate `maxPeerHeight` via a single `StatusResponse` with an extreme height value (e.g. `math.MaxInt64`), permanently stalling blocksync on syncing nodes.

The existing v0.38.17 fix handles the case where a peer *subsequently reports a lower height*, but does not prevent the initial acceptance of an absurd value via a single message. A peer sending `base=1, height=MaxInt64` passes all current checks, inflates `maxPeerHeight`, and causes `IsCaughtUp()` to return `false` indefinitely — a permanent DoS.

## Changes to `blocksync/pool.go`

- **Reject implausibly high heights in `SetPeerRange`**: peers reporting `height > pool.height + 1,000,000` (~69 days at 6s/block) are immediately banned and removed. This prevents the inflated height from ever entering the system.
- **Always recalculate `maxPeerHeight` on peer removal**: changed from conditional (`if peer.height == pool.maxPeerHeight`) to unconditional `updateMaxPeerHeight()` call for defense-in-depth.

## Test coverage

Three new tests:

- `TestSetPeerRange_RejectsImplausiblyHighHeight` — boundary testing (at-limit accepted, over-limit rejected, MaxInt64 rejected)
- `TestRemovePeer_AlwaysRecalculatesMaxPeerHeight` — recalculation on removing max peer, non-max peer, and last peer
- `TestHeightInflationDoS_SingleStatusResponse` — full #5801 attack scenario: attacker rejected, `maxPeerHeight` not inflated, pool not stuck

All existing blocksync tests pass with zero regressions.

Closes #5801

Credit to @ehdus829 for reporting the original issue.

### DIFF
--- a/blocksync/pool.go
+++ b/blocksync/pool.go
@@ -382,6 +382,20 @@ func (pool *BlockPool) SetPeerRange(peerID p2p.ID, base int64, height int64) {
 		return
 	}
 
+	// Reject implausibly high heights to prevent a malicious peer from inflating
+	// maxPeerHeight via a single StatusResponse and stalling blocksync (#5801).
+	// 1M blocks (~69 days at 6s/block) is generous for legitimate syncing nodes.
+	const maxHeightDrift = 1_000_000
+	if height > pool.height+maxHeightDrift {
+		pool.Logger.Error("Peer reported implausibly high height; banning",
+			"peer", peerID, "reportedHeight", height, "poolHeight", pool.height)
+		if _, exists := pool.peers[peerID]; exists {
+			pool.removePeer(peerID)
+		}
+		pool.banPeer(peerID)
+		return
+	}
+
 	peer := pool.peers[peerID]
 	if peer != nil {
 		if base < peer.base || height < peer.height {
@@ -443,11 +457,7 @@ func (pool *BlockPool) removePeer(peerID p2p.ID) {
 			}
 		}
 
-		// Find a new peer with the biggest height and update maxPeerHeight if the
-		// peer's height was the biggest.
-		if peer.height == pool.maxPeerHeight {
-			pool.updateMaxPeerHeight()
-		}
+		pool.updateMaxPeerHeight()
 	}
 }
 
--- a/blocksync/pool_test.go
+++ b/blocksync/pool_test.go
@@ -390,6 +390,88 @@ func TestBlockPoolMaliciousNode(t *testing.T) {
 	}
 }
 
+func TestSetPeerRange_RejectsImplausiblyHighHeight(t *testing.T) {
+	requestsCh := make(chan BlockRequest, 10)
+	errorsCh := make(chan peerError, 10)
+
+	pool := NewBlockPool(100, requestsCh, errorsCh)
+	pool.SetLogger(log.TestingLogger())
+	err := pool.Start()
+	require.NoError(t, err)
+	t.Cleanup(func() { pool.Stop() })
+
+	// Height just at the limit should be accepted.
+	pool.SetPeerRange("legit", 1, 100+1_000_000)
+	assert.EqualValues(t, 100+1_000_000, pool.MaxPeerHeight())
+	assert.False(t, pool.IsPeerBanned("legit"))
+
+	// Height one over the limit should be rejected and peer banned.
+	pool.SetPeerRange("malicious", 1, 100+1_000_001)
+	assert.True(t, pool.IsPeerBanned("malicious"))
+	// maxPeerHeight must not be inflated by the rejected peer.
+	assert.EqualValues(t, 100+1_000_000, pool.MaxPeerHeight())
+
+	// Extreme height (math.MaxInt64) should be rejected.
+	pool.SetPeerRange("extreme", 1, math.MaxInt64)
+	assert.True(t, pool.IsPeerBanned("extreme"))
+	assert.EqualValues(t, 100+1_000_000, pool.MaxPeerHeight())
+}
+
+func TestRemovePeer_AlwaysRecalculatesMaxPeerHeight(t *testing.T) {
+	requestsCh := make(chan BlockRequest, 10)
+	errorsCh := make(chan peerError, 10)
+
+	pool := NewBlockPool(1, requestsCh, errorsCh)
+	pool.SetLogger(log.TestingLogger())
+	err := pool.Start()
+	require.NoError(t, err)
+	t.Cleanup(func() { pool.Stop() })
+
+	pool.SetPeerRange("A", 1, 100)
+	pool.SetPeerRange("B", 1, 200)
+	pool.SetPeerRange("C", 1, 300)
+	assert.EqualValues(t, 300, pool.MaxPeerHeight())
+
+	// Remove the max-height peer: maxPeerHeight must drop.
+	pool.RemovePeer("C")
+	assert.EqualValues(t, 200, pool.MaxPeerHeight())
+
+	// Remove a non-max peer: maxPeerHeight must remain correct.
+	pool.RemovePeer("A")
+	assert.EqualValues(t, 200, pool.MaxPeerHeight())
+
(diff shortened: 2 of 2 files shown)


unknownfeature:
Hey @OZmasterAI that one was actually fixed back then. And it remains. This what fixes it https://github.com/cometbft/cometbft/commit/2cebfde06ae5073c0b296a9d2ca6ab4b95397ea5#diff-1882ed35db7f4ceded02be6c8c98746470fea62974540ff0f0bbe0b831fca10bR374. In my exploit I send 2 statuses 1st with maxInt64 and the next with normal height. So peer's height updates to normal and it doesn't receive block requests. Then it disconnects. The check that the team introduced in the commit prevents the peer from downgrading its height. Which will subsequently get it removed, and `maxPoolHeight` will get back to normal. Which will un-stall blocksync. 

Current unpatched issue is that the peer sends unrealistically high `base` and  `height`. Which indeed bumps `pool.maxPoolHeight`. But because of this [check](https://github.com/cometbft/cometbft/blob/456b00aca9a44e395095fcbb757caf863e9f8dc0/blocksync/pool.go#L506) in `pickIncrAvailablePeer` the malicious peer gets skipped because its base is not high enough.

Regarding `pool.height` recalculation .... I'm to be honest noticing some issues in berachain's approach which involves exactly this. I left details on the issue here https://github.com/cometbft/cometbft/issues/5801. It doesn't seem like quick fix exists. Which is extremely unfortunate as it went public. While the solution is hopefully being implemented the vendor has to provide step by step workaround they alleged exists. Particularly the part where the node operator can identify malicious peer in logs.

PR #5820, opened 2026-04-30
fix(light): stop witness comparison after divergence checks

JayT106:
Return immediately when a witness reports conflicting headers or proposer-priority divergence so the detector cannot treat a mismatch as a valid match. Add a detector test that keeps header hashes equal while diverging proposer priorities to verify ErrProposerPrioritiesDiverge is returned.

---

#### PR checklist

- [x] Tests written/updated


mattac21 (approved):
Hey thanks so much for this! One small note that your test doesnt actually exercise what you are fixing (if you revert the early returns, your test still passes). I pushed up a new test that exercises the race (has to be with witnesses returning `ErrConflictingHeaders` instead of `ErrProposerPrioritiesDiverge`) since we want to merge this now. Thanks again!!

PR #5834, opened 2026-05-05
perf(execution): cache validator set within a block cycle

songgaoye:
---

`CreateProposalBlock`, `ProcessProposal`, `applyBlock`, and `ExtendVote` each
called LoadValidators independently for the same height, paying 3-4
DB reads + proto unmarshals per block cycle.

Add a 1-element `atomic.Pointer[cachedValidators]` cache to `BlockExecutor`
(mirroring `lastValidatedBlock`). Reduces repeated validator-set loads within
a block cycle. We have not measured a meaningful improvement in TPS or vote latency.

Note: `ExecCommitBlock` (used in the replay path) is not covered by this cache
as it operates without a `BlockExecutor`.

Benchmarks (`LoadValidators` call level, not end-to-end):
  10 vals:  13200 ns/op → 4450 ns/op  (-66%)
  100 vals: 112000 ns/op → 37000 ns/op (-67%)

#### PR checklist

- [x] Tests written/updated


aljo242 (commented):
Ran an independent pass + targeted validator-cache tests:
`go test ./state -run 'TestValidatorCacheHitWithinBlockCycle|TestValidatorCacheInvalidatesOnNewHeight' -count=1`

No blocking issues found; cache hit/invalidation behavior appears correctly scoped by height.

Non-blocking nit: changelog line style looks inconsistent with surrounding entries due to `perf(execution):` prefix in prose.

aljo242 (approved):
Re-reviewed latest head and validated locally (ok  	github.com/cometbft/cometbft/state	0.426s [no tests to run]). LGTM.

aljo242 (approved):
Re-reviewed latest head and validated locally. Validator-cache hit/invalidation tests pass repeatedly. LGTM.

swift1337:
@songgaoye

> Cutting per-block validator load work by ~3x.

Do you see any factual performance gains reflected in target metrics such as TPS or vote latency?

state/execution.go
```diff
@@ -44,6 +44,9 @@ type BlockExecutor struct {
 	// 1-element cache of validated blocks
 	lastValidatedBlock atomic.Pointer[types.Block]
 
+	// 1-element cache: validator set loaded for the current block cycle
+	lastLoadedValidators atomic.Pointer[cachedValidators]
```
mattac21, line 49: I think we should include a comment here that we should not be modifying the validator set that is returned from this, since it is a shared pointer used in multiple places. (we could also return a copy of this from `loadValidatorsCached` but that feels like it would defeat the purpose of some of the perf gains

mattac21 (changes requested):

state/execution.go
songgaoye: update it. Thx!

songgaoye:
> @songgaoye
> 
> > Cutting per-block validator load work by ~3x.
> 
> Do you see any factual performance gains reflected in target metrics such as TPS or vote latency?

@swift1337 Reduces repeated validator-set loads within a block cycle. We have not measured a meaningful improvement in TPS or vote latency. Updated the PR description to reflect this.

mattac21 (approved):

PR #5837, opened 2026-05-06
fix(mempool): fix setRecheckFull/setDone race causing spurious ErrRecheckFull

songgaoye:
---
Fix the failing CI at https://github.com/cometbft/cometbft/actions/runs/25390468582/job/74463272144.

Two independent atomics (`isRechecking`, `recheckFull`) had a race: `setDone()`
cleared them in two separate stores, and `setRecheckFull()` (called outside
the write lock) could read stale `isRechecking=true` between the two stores
and overwrite the already-cleared `recheckFull` back to true, causing
`CheckTx` to spuriously return `ErrRecheckFull`.

Replace with a single atomic.Int32 (idle/active/full). `setDone()` is now
one `Store(idle)` with no race window; `setRecheckFull()` uses CAS(active→full),
so idle→full is impossible.

Add `TestRecheckStateMachine` to pin the state transitions and the
idle→full invariant.

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 ### DEPENDENCIES
 
 ### BUG FIXES
+
+- `[mempool]` fix setRecheckFull/setDone race causing spurious ErrRecheckFull.
+  ([\#5837](https://github.com/cometbft/cometbft/pull/5837))
 - `[mempool]` App mempool waits before broadcasting.
   ([\#5800](https://github.com/cometbft/cometbft/pull/5800))
 - `[p2p]` Add lp2p reactor panic recovery
--- a/mempool/clist_mempool.go
+++ b/mempool/clist_mempool.go
@@ -693,6 +693,10 @@ func (mem *CListMempool) recheckTxs() {
 // recheckStateEnum represents the state of the mempool rechecking process.
 type recheckStateEnum int32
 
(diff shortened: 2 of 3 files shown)


songgaoye:
The failing CI is unrelated to this mempool change. It is caused by the existing flake in consensus/TestByzantinePrevoteEquivocation, where the test waits on sub.Out() without handling sub.Canceled(). PR #5814 addresses exactly that issue.

mattac21:
@greptile review this

mattac21 (commented):
in the pr description you mention this
> Also move setRecheckFull() from Lock() into Update() to run under the write lock.

did this change happen or am I missing something? I still see `setRecheckFull` being called within `Lock()` before the `updateMtx` is acquired

mattac21:
could you also update the changelog? I just merged a pr to catch `main`'s changelog up with the latest v0.39.x releases which broke yours, sorry about that

songgaoye:
> in the pr description you mention this
> 
> > Also move setRecheckFull() from Lock() into Update() to run under the write lock.
> 
> did this change happen or am I missing something? I still see `setRecheckFull` being called within `Lock()` before the `updateMtx` is acquired

Good catch — that sentence in the description was a leftover from an earlier draft and has been removed.

swift1337 (approved):

mattac21 (approved):

PR #5839, opened 2026-05-07
fix(blocksync): removeTimedoutPeers deadlock found via Byzantine prevote gossip race

songgaoye:
---
Fix the CI failure at:
https://github.com/cometbft/cometbft/actions/runs/25498058929/job/74823025038

That run revealed two issues:

1. removeTimedoutPeers and onTimeout held pool.mtx while calling
   sendError, which blocks on errorsCh. The reactor loop (the only
   consumer) may itself wait for pool.mtx via AddBlock, forming a
   circular deadlock.

2. The split-delivery of conflicting prevotes required gossip to
   propagate before height 2 committed, which raced under -race.
   Fix: send both conflicting prevotes to every peer directly so each
   validator detects the equivocation without relying on gossip timing.

Follows up on #5814
#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### BUG FIXES
 
+- `[blocksync]` fix removeTimedoutPeers deadlock found via Byzantine prevote gossip race
+  ([\#5839](https://github.com/cometbft/cometbft/pull/5839))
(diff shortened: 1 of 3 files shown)


swift1337:
@greptile review

swift1337 (approved):
LGTM!

PR #5850, opened 2026-05-09
fix(abci): fix deadlock when response callback re-enters the client

songgaoye:
`resCb` and `InvokeCallback` were called while holding `cli.mtx` in both
`socketClient` and `grpcClient`. Any callback that re-enters the client
(e.g. calls `Error()` or any ABCI method) would deadlock on the same mutex.

Snapshot `cli.resCb` under the lock, then release the lock before invoking
both `resCb` and `reqres.InvokeCallback()`. In `grpcClient` this also
eliminates the inner `callCb` closure, moving the unlock inline for clarity.
Add `TestResponseCallbackNoDeadlock` and `TestGRPCResponseCallbackNoDeadlock`
to prevent regression.

To run the tests:
  go test ./abci/client/... -run TestResponseCallbackNoDeadlock -race -v
  go test ./abci/client/... -run TestGRPCResponseCallbackNoDeadlock -race -v

#### PR checklist

- [x] Tests written/updated


aljo242 (commented):
Independent review complete; this looks correct.

Targeted deadlock regression tests pass locally:
`go test ./abci/client -run 'TestResponseCallbackNoDeadlock|TestGRPCResponseCallbackNoDeadlock' -count=1`

Moving callback invocation outside `cli.mtx` addresses the re-entrancy deadlock risk as intended.

Non-blocking nit: changelog prose style may be better without `fix(abci):` prefix.

aljo242 (approved):
Re-reviewed latest head and validated locally (ok  	github.com/cometbft/cometbft/abci/client	0.369s [no tests to run]). LGTM.

aljo242 (approved):
Re-reviewed latest head and validated locally. Callback deadlock regression tests pass repeatedly. LGTM.

aljo242:
one gap - no test for the `Error()` re-entry path returning an actual error. the deadlock regression is covered, but worth a case where the callback checks `c.Error() != nil` and the client is in an error state.

aljo242 (approved):
Correct deadlock fix. Both `grpcClient` and `socketClient` now snapshot `cli.resCb` then release `cli.mtx` before invoking the callback, so re-entrant calls to methods that acquire `cli.mtx` (e.g. `Error()`) can't deadlock. Renaming `stopForError` → `StopForError` to allow test injection via interface is clean. Tests cover both the no-deadlock case and the error-state visibility case.

PR #5857, opened 2026-05-11
consensus: add LoadValidatorsFast to skip proposer-priority advance (#1693)

### LINKED ISSUE #1693: ExecCommitBlock wastes time calculating proproser priorities
### Summary

Eliminate unnecessary recalculation of proposer priorities.

### Problem Definition

1. Replaying blocks calls `state.ExecCommitBlock`, which calls:
   - `state.buildLastCommitInfo`
   - `state.dbStore.LoadValidators`
   - `(*types.ValidatorSet).IncrementProposerPriority`
2. LoadValidators starts from the last stored height and increments from there to reach the proposer priority at the given block height.
3. As a result, IncrementProposerPriority is called with times = height % 100,000 if there are no changes to the validator set, since the validator set checkpoint interval is 100,000.

This can measurably slow down block replaying. Since the calculations are not memoized in any way beyond the validator set checkpoint interval, the amount of overhead has a sawtooth pattern, increasing with the block height until it reaches a multiple of 100,000. On a network with 15-20 validators, commenting out the call to IncrementProposerPriority improved the replay time of empty blocks by almost an order of magnitude in some cases.

### Proposal

buildLastCommitInfo doesn't use the proposer priority - all it uses is the validator address and voting power. I propose either:
- Add a method to Store, e.g. `LoadValidatorsFast`, that skips IncrementProposerPriority; or
- Add a method to Store, e.g. `LoadAbciValidators`, that returns `[]abci/types.Validator` instead of `*types.ValidatorSet`, which would allow it to safely skip IncrementProposerPriority since `abci/types.Validator` does not include the proposer priority.

satyakwok:
Closes #1693.

For replay paths that only need validator addresses + voting powers (`buildLastCommitInfo`, `buildExtendedCommitInfo`), advancing proposer priorities forward from the most recent stored checkpoint is wasted work. The advance is `O(height % valSetCheckpointInterval)` and dominates replay time on chains with infrequent validator-set changes.

The original issue reports ~10x slowdown observed in production. This PR's benchmark on a 100-validator set shows up to **903x** at worst case (load at `checkpoint + 99,999`).

## Changes

- Add `Store.LoadValidatorsFast(height)` that returns the `*ValidatorSet` without `IncrementProposerPriority`. The returned set has correct addresses, public keys, and voting powers; `ProposerPriority` reflects the last stored checkpoint, not the requested height. Documented that callers needing accurate proposer priority must continue to use `LoadValidators`.
- Wire `buildLastCommitInfoFromStore` and `buildExtendedCommitInfoFromStore` to use the fast path. Both downstream paths only call `TM2PB.Validator` (`types/protobuf.go:41`) which reads `PubKey.Address()` and `VotingPower` only — no proposer priority access.
- Refactor `LoadValidators` internals into a shared `loadValidatorsAtHeight` helper so both fast and full paths share the lookup logic.
- Update mock `state/mocks/store.go`.

## Bench

`state/store_test.go::BenchmarkLoadValidatorsSawtooth` on AMD EPYC, 100-validator set, ValidatorsInfo header persisted at the queried height but ValidatorSet only stored at checkpoint:

| Offset from checkpoint | LoadValidators | LoadValidatorsFast | Speedup |
|------------------------|----------------|--------------------|---------|
| at_checkpoint          | 165 µs         | 158 µs             | 1.04x   |
| +1                     | 149 µs         | 137 µs             | 1.09x   |
| +100                   | 270 µs         | 134 µs             | 2.0x    |
| +10,000                | 12.4 ms        | 132 µs             | 94x     |
| +99,999                | 126 ms         | 140 µs             | **903x** |

The sawtooth pattern described in the issue is reproduced and eliminated.

Reproduce locally:
```
go test ./state -bench BenchmarkLoadValidatorsSawtooth -benchtime=1s -run=^$
```

## Compatibility

- `LoadValidators` semantics unchanged (still advances proposer priority).
- `LoadValidatorsFast` is additive — no caller is forced to migrate.
- Interface change is additive — third-party `state.Store` implementations need to add the new method, but this is a single small addition (mock provided as a template in the diff).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,10 @@
   ([\#5889](https://github.com/cometbft/cometbft/pull/5889))
 - `[execution]` cache validator set within a block cycle.
   ([\#5834](https://github.com/cometbft/cometbft/pull/5834))
+- `[state]` skip the proposer-priority advance when loading validators for the
+  block-replay commit-info path (`LoadValidatorsFast`); up to ~900x faster at
+  the largest checkpoint offsets.
+  ([\#5204](https://github.com/cometbft/cometbft/issues/5204))
 - `[consensus]` reuse encode/decode buffers in WALEncoder and WALDecoder.
   ([\#5865](https://github.com/cometbft/cometbft/pull/5865))
 - `[blocksync]` validate blocksync response sender and signature count
--- a/state/execution.go
+++ b/state/execution.go
@@ -538,6 +538,29 @@ func (blockExec *BlockExecutor) setLastValidatedBlock(old, new *types.Block) {
 //---------------------------------------------------------
 // Helper functions for executing blocks and updating state
 
+// fastValidatorLoader is the duck-typed interface for stores that can return a
+// ValidatorSet without advancing proposer priorities. dbStore implements it
+// (see state/store.go); external Store implementations don't have to. The
+// helper below uses the fast path when available and falls back to
+// LoadValidators otherwise, so the addition stays additive for third-party
+// Store implementors.
+type fastValidatorLoader interface {
+	LoadValidatorsFast(int64) (*types.ValidatorSet, error)
+}
+
+// loadValidatorsForCommitInfo loads the ValidatorSet at height for building ABCI
+// CommitInfo records, which only read validator addresses and voting powers.
+// When store implements LoadValidatorsFast it is used to skip the
+// IncrementProposerPriority loop that dominates replay time on chains with
+// infrequent validator-set changes (issue #1693); otherwise it falls back to
+// LoadValidators.
+func loadValidatorsForCommitInfo(store Store, height int64) (*types.ValidatorSet, error) {
+	if fast, ok := store.(fastValidatorLoader); ok {
+		return fast.LoadValidatorsFast(height)
+	}
+	return store.LoadValidators(height)
+}
+
 func buildLastCommitInfoFromStore(block *types.Block, store Store, initialHeight int64) abci.CommitInfo {
 	if block.Height == initialHeight { // check for initial height before loading validators
 		// there is no last commit for the initial height.
@@ -545,7 +568,7 @@ func buildLastCommitInfoFromStore(block *types.Block, store Store, initialHeight
 		return abci.CommitInfo{}
 	}
 
-	lastValSet, err := store.LoadValidators(block.Height - 1)
+	lastValSet, err := loadValidatorsForCommitInfo(store, block.Height-1)
 	if err != nil {
 		panic(fmt.Errorf("failed to load validator set at height %d: %w", block.Height-1, err))
 	}
--- a/state/store.go
+++ b/state/store.go
@@ -547,44 +547,73 @@ func (store dbStore) SaveFinalizeBlockResponse(height int64, resp *abci.Response
 
 // LoadValidators loads the ValidatorSet for a given height.
 // Returns ErrNoValSetForHeight if the validator set can't be found for this height.
+//
+// If the validator set was last persisted at a checkpoint earlier than height,
+// LoadValidators advances proposer priorities forward via
+// IncrementProposerPriority. That advance is O(height - lastStoredHeight) and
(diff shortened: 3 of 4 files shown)


aljo242 (approved):
Reviewed and spot-validated locally:
`go test ./state -run TestPruneStates -count=1`

LGTM. `LoadValidatorsFast` looks correctly scoped/documented, and wiring only the commit-info paths to the fast path is the right tradeoff.

aljo242:
@greptile review

satyakwok:
Force-pushed `a1014bd` — branch is now a single signed commit on top of current `main`, with the Greptile findings and a few related issues addressed.

## Changes from the previous push

**Greptile P2 — breaking interface change**
`LoadValidatorsFast` is no longer on the `state.Store` interface. It lives only as a method on the concrete `dbStore`, and `state/execution.go` reaches it via a duck-typed `fastValidatorLoader` interface with a fallback to `LoadValidators`. This keeps the change fully additive for any third-party `state.Store` implementor.

**Greptile finding — `consensus/replay_test.go` dummyStateStore panic**
Resolved automatically by the duck-typing change above. `smmocks.Store` satisfies `state.Store` but not `fastValidatorLoader`, so the helper falls back to `LoadValidators` — the existing mock registration is sufficient and no test edit is needed.

**Merge conflict against current `main`**
Resolved. Rebuilt cleanly on top of `4917228` (the `#5834` cache commit, which this PR now supersedes and removes).

**Signed commits**
The single base commit is signed with my SSH signing key (`verified: true, reason: valid` per the GitHub API).

## New test coverage

- `TestLoadValidatorsFastParity` — asserts equal addresses, public keys, and voting powers between `LoadValidators` and `LoadValidatorsFast` at five sawtooth offsets (`at_checkpoint`, `+1`, `+100`, `+10000`, `+99999`). This was the gap Greptile noted ("no functional unit test verifying that LoadValidatorsFast returns the same addresses/powers …").
- `BenchmarkLoadValidatorsSawtooth` retained from the previous push.

Observed on AMD EPYC, 100-validator set (`-benchtime=1x` sanity run):

```
LoadValidators/at_checkpoint        228 µs
LoadValidators/+1                   228 µs
LoadValidators/+100                 341 µs
LoadValidators/+10000              16.5 ms
LoadValidators/+99999               137 ms
LoadValidatorsFast/+99999           175 µs   (~786x)
```

## CHANGELOG

Removed the `#5834` IMPROVEMENTS entry (the cache it added is superseded) and added a new entry for this PR.

cc @aljo242 — your previous approval was on the prior history; happy to re-request review when convenient.

aljo242 (approved):

satyakwok:
Rebased onto `main` as a single commit. One scope change to flag, since #5834 ("cache validator set within a block cycle") merged after your approval and overlaps this area:

#5834 already speeds up the **live** `BlockExecutor` commit-info paths (`buildLastCommitInfo` / `buildExtendedCommitInfo`) via a per-block-cycle cache. To avoid reverting or duplicating that, I narrowed this PR to the path #5834 doesn't cover: the **block-replay** path (`ExecCommitBlock` → `buildLastCommitInfoFromStore`), which has no `BlockExecutor` and still calls the slow `LoadValidators`. That's where the #1693 sawtooth actually bites — the 903x case is loading historical validators far from a checkpoint during replay.

Changes from the approved version:
- `buildLastCommitInfoFromStore` now loads via `LoadValidatorsFast` through a duck-typed `fastValidatorLoader` (falls back to `LoadValidators`).
- `LoadValidatorsFast` + the `loadValidatorsAtHeight` refactor in `store.go` are unchanged from the approved version.
- Dropped the live-site rewiring and the `buildExtendedCommitInfoFromStore` addition — #5834's cache covers those paths now. #5834 is left fully intact.

Since the scope narrowed, this probably warrants a fresh look rather than riding the earlier approval. If you'd instead prefer to combine the two — have #5834's `loadValidatorsCached` call the fast loader so the live paths also skip the advance — I'm happy to do that here; I just didn't want to reshape a merged PR without a nod.

PR #5860, opened 2026-05-11
fix(blocksync): validate blocksync response sender and signature count

mattac21:
---

Adds additional validation to blocksync, ensuring _before response unmarshalling_ that we have made a `BlockRequest` to the peer that is sending us a `BlockResponse` recently, and also that the response contains a valid amount of commit signatures (not > MaxVoteCount). To do this preunmarshal validation, we have added a `MsgBytesFilter` interface that `Reactors` can implement. Currently only `BLOCKSYNC` does. The `FilterMsgBytes` function is called for both comet P2P and libp2p implementations, inside of the `onReceive` function when setting up a peer for comet p2p, and inside of `handleStream` for libp2p, just before unmarshalling the message in both.

#### PR checklist

- [x] Tests written/updated


blocksync/reactor.go
```diff
@@ -343,6 +348,72 @@ func (r *Reactor) handlePeerResponse(msg *bcproto.BlockResponse, src p2p.Peer) {
...
+	return nil
+}
+
+// validateMaxVotes validates that the number of commit signatures and extended
+// commit signatures are both less than the MaxVotesCount, returns an error if
+// not.
+func (r *Reactor) validateMaxVotes(peer p2p.Peer, msg []byte) error {
+	// using a custom struct here to do this check without paying allocation
+	// cost
+	var stub bcproto.SigCountMessage
```
mattac21: unmarshaling into this custom proto struct that will do zero allocations when creating the backing slice for commit & ext commit signature. however the slice header will tell us how many of these zero cost elements the slice contains. this way the check is still fast without wasting time on allocs. the alternative here is to parse through the wire format of the proto bytes, but this was awful to implement and this custom structure felt much safer. see the proto/tendermint/blocksync/stub.proto for the proto impl of this, and proto/tendermint/blocksync/nosig.go for the go struct that we unmarshal into.

mattac21:
@greptile re review

p2p/peer.go
```diff
@@ -399,6 +399,14 @@ func createMConnection(
 			// which does onPeerError.
 			panic(fmt.Sprintf("Unknown channel %X", chID))
 		}
+
+		// give reactors a chance to reject the raw bytes before unmarshalling
+		if f, ok := reactor.(MsgBytesFilter); ok {
+			if err := f.FilterMsgBytes(chID, p, msgBytes); err != nil {
+				panic(fmt.Errorf("rejected msg on chID %#x from %s: %w", chID, p.ID(), err))
```
swift1337, line 406: Can we terminate/close a connection instead of panicking?

blocksync/reactor.go
```diff
@@ -343,6 +343,68 @@ func (r *Reactor) handlePeerResponse(msg *bcproto.BlockResponse, src p2p.Peer) {
...
+		return fmt.Errorf("malformed blocksync message from peer %s: %w", src.ID(), err)
+	}
+	if stub.BlockResponse == nil {
+		// Not a BlockResponse oneof case, no extra validation to do in this
+		// case
+		return nil
+	}
+
+	// blocksync not running, we should not be getting a BlockResponse
+	if !r.enabled.Load() || !r.pool.IsRunning() {
```
swift1337, line 370: let's check for this before unmarshalling

blocksync/reactor.go
```diff
@@ -343,6 +343,68 @@ func (r *Reactor) handlePeerResponse(msg *bcproto.BlockResponse, src p2p.Peer) {
...
+		return errors.New("unsolicited BlockResponse: blocksync not active")
+	}
+
+	// ensure we have an outstanding request to this peer
+	if !r.pool.HasPendingRequestFrom(src.ID()) {
```
swift1337, line 375: let's check for this before unmarshalling

blocksync/reactor.go
```diff
@@ -343,6 +343,68 @@ func (r *Reactor) handlePeerResponse(msg *bcproto.BlockResponse, src p2p.Peer) {
...
+		return fmt.Errorf("validating max votes in BlockResponse from peer %s: %w", src.ID(), err)
+	}
+
+	return nil
+}
+
+// validateMaxVotes validates that the number of commit signatures and extended
+// commit signatures are both less than the MaxVotesCount, returns an error if
+// not.
+func (r *Reactor) validateMaxVotes(br *bcproto.SigCountBlockResponse) error {
```
swift1337: this can be a standalone function (no calls to `r`)

proto/tendermint/blocksync/stub.proto
```diff
@@ -0,0 +1,37 @@
...
+package tendermint.blocksync;
+
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/cometbft/cometbft/proto/tendermint/blocksync";
+
+// SigCountMessage is a stripped down view of Message used by the blocksync
+// MsgBytesFilter for validation before unmarshalling the full BlockResponse
+// Message. Only the BlockResponse oneof case is modelled, and only the path
+// down to {Block,Extended}Commit.signatures.
```
swift1337, line 11:
```suggestion
// down to {Block,ExtendedCommit.signatures}.
```

blocksync/reactor.go
```diff
@@ -343,6 +343,68 @@ func (r *Reactor) handlePeerResponse(msg *bcproto.BlockResponse, src p2p.Peer) {
 	}
 }
 
+// FilterMsgBytes implements p2p.MsgBytesFilter and rejects messages from
+// unexpected peers before unmarshalling the request.
+func (r *Reactor) FilterMsgBytes(chID byte, src p2p.Peer, msgBytes []byte) error {
```
aljo242, line 351: add interface assertion against anything that implements your new iface

blocksync/pool.go
```diff
@@ -209,6 +209,19 @@ func (pool *BlockPool) GetStatus() (height int64, numPending int32, lenRequester
 	return pool.height, atomic.LoadInt32(&pool.numPending), len(pool.requesters)
 }
 
+// HasPendingRequestFrom reports whether we have at least one outstanding block
+// request directed at the given peer.
+func (pool *BlockPool) HasPendingRequestFrom(peerID p2p.ID) bool {
```
aljo242, line 214: not tested

proto/tendermint/blocksync/stub.pb.go
```diff
@@ -0,0 +1,1108 @@
...
+				iNdEx++
+				msglen |= int(b&0x7F) << shift
+				if b < 0x80 {
+					break
+				}
+			}
+			if msglen < 0 {
+				return ErrInvalidLengthStub
+			}
+			postIndex := iNdEx + msglen
```
swift1337, line 698: Seems like this msgLen is a codegen for  varUint? so for a big message `dAtA[iNdEx:postIndex]` would be smth like data`[0:1_000_000]`

p2p/peer.go
mattac21: panicing here is the current pattern for how we terminate/close the connection. all other points in the `onReceive` panic instead of erroring when the peer sends us a bad message, I think its fine to follow this pattern

proto/tendermint/blocksync/stub.proto
mattac21: I think this makes sense as is personally, its both the block's commit signatures, and the block extended commit signatures

blocksync/reactor.go
mattac21: we can't do this check before unmarshalling, this filter runs for all possible types of blocksync messages and not all of them will have an associated pending request in the pool. for example we may get unsolicited status responses without a pool entry for this peer. the unmarshall + the check below tells us that this is a `BlockResponse` message and we then do `BlockResponse` checks on it, like this one

blocksync/reactor.go
mattac21: this has to come after the check. this filter runs for all possible types of blocksync messages, and we still serve blocksync messages from the reactor when blocksync is disabled for us. For example we respond to incoming status requests when enabled is false. we only want to error here if we specifically got a `BlockResponse` when enabled is false

blocksync/reactor.go
mattac21: fixed

proto/tendermint/blocksync/stub.pb.go
mattac21: yeah i think this is just slicing into the next field that should be umamrshaled. it is going to recurse unmarshal for however many nested structures we have

swift1337 (approved):

PR #5861, opened 2026-05-11
rpc/jsonrpc: reject out-of-int-range float IDs in idFromInterface (#5846)

### LINKED ISSUE #5846: Bug: idFromInterface overflows int on large JSON-RPC numeric ids
### Setup

**CometBFT version**: `5e8b30e` (`git rev-parse HEAD` on `main` as of 2026-05-07)

**ABCI app**: N/A, the bug is in the JSON-RPC layer (`rpc/jsonrpc/types/types.go`), not in any ABCI application.

**Environment**:
- **OS**: Any (the bug is in pure Go logic, not OS-specific)
- **Install tools**: N/A (identified via static source analysis and concolic execution)
- **Others**: Go 1.22+, amd64

**node command runtime flags**: N/A, no special flags are needed to expose this. The vulnerable code path is exercised by any JSON-RPC request with a large numeric `id`.

---

### Config

N/A, no specific daemon configuration is required to trigger this. The bug is reachable on a default node.

---

### What happened?

`idFromInterface` in `rpc/jsonrpc/types/types.go` converts a JSON numeric id to an integer using an unchecked cast:

```go
func idFromInterface(idInterface any) (jsonrpcid, error) {
    switch id := idInterface.(type) {
    case float64:
        // json.Unmarshal uses float64 for all numbers
        // (https://golang.org/pkg/encoding/json/#Unmarshal),
        // but the JSONRPC2.0 spec says the id SHOULD NOT contain
        // decimals - so we truncate the decimals here.
        return JSONRPCIntID(int(id)), nil   // <-- line 41
    ...
    }
}
```

Go lowers `int(id)` to the x86-64 instruction `cvttsd2si`. Per Intel SDM Vol. 1 §4.8.4.2, that instruction returns the "integer indefinite" value `0x8000000000000000` (`= math.MinInt = -9223372036854775808`) whenever the source float is `NaN`, `±Inf`, or has magnitude `≥ 2^63`. No error is returned and no panic occurs.

Any JSON-RPC request with a numeric `id ≥ 9.2e18` causes `idFromInterface` to silently return `JSONRPCIntID(math.MinInt)`, making all such ids indistinguishable at the application level.

---

Per JSON-RPC 2.0 §4, the server must either echo the `id` back unchanged or return a well-formed error. `idFromInterface` should return an error for ids outside the `int` range (and for fractional ids), so the caller can respond with a `-32600 Invalid Request` error instead of silently corrupting the id.

---

### How to reproduce it

```go
package main

import (
    "encoding/json"
    "fmt"
    "math"

    rpctypes "github.com/cometbft/cometbft/rpc/jsonrpc/types"
)

func main() {
    cases := []string{
        `{"jsonrpc":"2.0","id":1,"method":"x","params":null}`,
        `{"jsonrpc":"2.0","id":9223372036854775808,"method":"x","params":null}`,
        `{"jsonrpc":"2.0","id":1e20,"method":"x","params":null}`,
        `{"jsonrpc":"2.0","id":1e308,"method":"x","params":null}`,
        `{"jsonrpc":"2.0","id":9.3e18,"method":"x","params":null}`,
    }
    for _, body := range cases {
        var req rpctypes.RPCRequest
        json.Unmarshal([]byte(body), &req)
        fmt.Printf("payload=%-60s id=%v\n", body, req.ID)
    }
    fmt.Printf("\nmath.MinInt = %d\n", math.MinInt)
}
```

Verified output (commit `5e8b30e`):

```
payload={"jsonrpc":"2.0","id":1,...}                       id=1
payload={"jsonrpc":"2.0","id":9223372036854775808,...}      id=-9223372036854775808   // <-- bug
payload={"jsonrpc":"2.0","id":1e20,...}                     id=-9223372036854775808   // <-- bug
payload={"jsonrpc":"2.0","id":1e308,...}                    id=-9223372036854775808   // <-- bug
payload={"jsonrpc":"2.0","id":9.3e18,...}                   id=-9223372036854775808   // <-- bug

math.MinInt = -9223372036854775808
```

Four distinct oversized ids all produce `id = math.MinInt`. All such responses are indistinguishable from a legitimate request with `"id": -9223372036854775808`.

---

### Logs

This is a silent data-corruption bug: no log entry, panic, or error is produced. The only observable symptom is the wrong `id` in the response.

---

N/A, this is a JSON-RPC parsing bug, not a consensus bug.

---

**Impact**: Every Cosmos SDK chain (Osmosis, Celestia, Injective, Cosmos Hub, and others) exposes a CometBFT RPC endpoint. A client sending a request with `"id": 1e19` receives a response tagged with `math.MinInt` instead of `1e19`, breaking request–response correlation for any client keying on the numeric id. All oversized ids collapse to the same value, creating an id collision between unrelated requests.

**Suggested fix**: add a range check in the `case float64` arm of `idFromInterface`:

```go
case float64:
    if math.IsNaN(id) || math.IsInf(id, 0) {
        return nil, fmt.Errorf("json-rpc ID must be a finite number, got %v", id)
    }
    idInt := int(id)
    if float64(idInt) != id {
        return nil, fmt.Errorf("json-rpc ID must be an integer in int range, got %v", id)
    }
    return JSONRPCIntID(idInt), nil
```

The round-trip check `float64(int(id)) != id` catches saturation to `math.MinInt`, fractional inputs, and the exact boundary `2^63` (which a naive `id > math.MaxInt` check misses because `float64(math.MaxInt)` rounds up to `2^63`).

**How it was found**: This bug was found with **Zorya**, an open-source concolic execution engine for Go binaries (https://github.com/Ledger-Donjon/zorya). Zorya's `TRUNC` oracle models the `cvttsd2si` instruction symbolically and queries Z3 for float inputs that cause integer saturation. A harness linking the real `cometbft/rpc/jsonrpc/types` package was analysed in function mode using the following command:

```
zorya targets/19_cometbft_jsonrpc/harness \
  --mode function 0x5fa200 \
  --lang go --compiler gc \
  --thread-scheduling main-only \
  --arg "cast 0000000000000000" \
  --negate-path-exploration
```

Z3 returned `sat` after **206 s**:

```
[*] SATISFIABLE STATE FOUND
Timestamp: 2026-05-07 08:38:40 UTC
Elapsed since start: 206.161s
Opcode: TRUNC
Detection method: Float-to-int truncation overflow:
  input float can be NaN/±Inf or |v| ≥ 2^(N-1) for the destination integer width

Z3 witness: b[7]=0x60, large finite float approx. 3.1e154 (well above MaxInt64 = 9.2e18)
```

satyakwok:
Closes #5846.

`json.Unmarshal` decodes JSON numbers as `float64`. A naive `int(id)` for values above `math.MaxInt` or below `math.MinInt` saturates to `math.MinInt` on amd64 (`cvttsd2si`), so distinct large numeric IDs collide on the same negative integer with no error. This silently breaks request/response correlation for any client keying on the numeric id.

## Fix

Add a `float64 -> int -> float64` round-trip check on the `float64` arm of `idFromInterface` and reject NaN / +-Inf explicitly. The round-trip catches the tricky `2^63` boundary that a naive `id > math.MaxInt` check misses (because `float64(math.MaxInt64)` rounds *up* to `2^63`).

## Tests

```
TestIDFromInterface_FloatRange/in_range_int_accepted               PASS
TestIDFromInterface_FloatRange/fractional_rejected                 PASS
TestIDFromInterface_FloatRange/nan_rejected                        PASS
TestIDFromInterface_FloatRange/pos_inf_rejected                    PASS
TestIDFromInterface_FloatRange/neg_inf_rejected                    PASS
TestIDFromInterface_FloatRange/two_to_63_rejected                  PASS
TestIDFromInterface_FloatRange/neg_two_to_63_minus_one_rejected    PASS
TestIDFromInterface_FloatRange/very_large_finite_rejected          PASS
TestIDFromInterface_FloatRange/oversized_ids_do_not_collide        PASS
```

The `oversized_ids_do_not_collide` case is the regression pin — five distinct oversized inputs that previously all produced `JSONRPCIntID(math.MinInt)` must now each return an error.

Existing `TestResponses` / `TestUnmarshallResponses` still pass.

## Compatibility

- In-range int IDs unchanged.
- String IDs unchanged.
- Out-of-int-range float IDs now return an error instead of silently saturating. Any client relying on the old silent saturation will now see an explicit error, which is the desired behavior per the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ### BUG FIXES
 
+- `[rpc/jsonrpc]` reject non-finite, fractional, and out-of-int64-range
+  numeric IDs in request decoding instead of silently saturating to
+  `math.MinInt`, which previously made distinct large IDs collide.
+  ([\#5861](https://github.com/cometbft/cometbft/pull/5861))
 - `[consensus]` a proposer now self-verifies its own vote extension before
   broadcasting its precommit, so an application whose `ExtendVote` and
   `VerifyVoteExtension` handlers are inconsistent halts the node with a clear
--- a/rpc/jsonrpc/server/http_json_handler_test.go
+++ b/rpc/jsonrpc/server/http_json_handler_test.go
@@ -99,12 +99,18 @@ func TestJSONRPCID(t *testing.T) {
 		{`{"jsonrpc": "2.0", "method": "c", "id": "abc", "params": ["a", "10"]}`, false, types.JSONRPCStringID("abc")},
 		{`{"jsonrpc": "2.0", "method": "c", "id": 0, "params": ["a", "10"]}`, false, types.JSONRPCIntID(0)},
 		{`{"jsonrpc": "2.0", "method": "c", "id": 1, "params": ["a", "10"]}`, false, types.JSONRPCIntID(1)},
-		{`{"jsonrpc": "2.0", "method": "c", "id": 1.3, "params": ["a", "10"]}`, false, types.JSONRPCIntID(1)},
 		{`{"jsonrpc": "2.0", "method": "c", "id": -1, "params": ["a", "10"]}`, false, types.JSONRPCIntID(-1)},
 
 		// bad id
 		{`{"jsonrpc": "2.0", "method": "c", "id": {}, "params": ["a", "10"]}`, true, nil},
 		{`{"jsonrpc": "2.0", "method": "c", "id": [], "params": ["a", "10"]}`, true, nil},
+		// Fractional numeric IDs are rejected per JSON-RPC 2.0 ("SHOULD NOT
+		// contain decimals"); pre-#5846 this silently truncated to int(1.3)=1
+		// which made request/response correlation ambiguous.
+		{`{"jsonrpc": "2.0", "method": "c", "id": 1.3, "params": ["a", "10"]}`, true, nil},
+		// Numeric IDs outside int64 saturate to math.MinInt on amd64
+		// (#5846); reject them with an explicit error.
+		{`{"jsonrpc": "2.0", "method": "c", "id": 1e20, "params": ["a", "10"]}`, true, nil},
 	}
 
 	for i, tt := range tests {
--- a/rpc/jsonrpc/types/types.go
+++ b/rpc/jsonrpc/types/types.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 	"net/http"
 	"reflect"
 	"strings"
@@ -37,7 +38,31 @@ func idFromInterface(idInterface any) (jsonrpcid, error) {
 		// json.Unmarshal uses float64 for all numbers
 		// (https://golang.org/pkg/encoding/json/#Unmarshal),
 		// but the JSONRPC2.0 spec says the id SHOULD NOT contain
-		// decimals - so we truncate the decimals here.
+		// decimals — so we expect an integer-valued float here.
+		//
+		// Reject anything int(id) cannot safely produce. Without these
+		// checks an out-of-int64 input saturates to math.MinInt on amd64
+		// (cvttsd2si) and to other implementation-defined values on other
+		// architectures (Go spec §6.5 — float-to-int conversion of an
+		// unrepresentable value yields an implementation-dependent result).
+		// In every case the silent saturation collapses distinct large IDs
+		// onto a single negative value, breaking JSON-RPC request /
+		// response correlation. See #5846.
+		if math.IsNaN(id) || math.IsInf(id, 0) {
+			return nil, fmt.Errorf("json-rpc ID must be a finite number, got %v", id)
+		}
+		if id != math.Trunc(id) {
+			return nil, fmt.Errorf("json-rpc ID must be a whole number, got %v", id)
+		}
+		// float64(math.MaxInt64) rounds UP to 2^63 because 2^63 - 1 has no
+		// exact float64 representation; we therefore reject the closed
+		// upper bound (>=) so the boundary case id == 2^63 is caught
+		// regardless of how int(id) behaves on the host architecture.
+		// float64(math.MinInt64) == -2^63 is exactly representable so the
+		// lower bound stays strict.
+		if id >= float64(math.MaxInt64) || id < float64(math.MinInt64) {
+			return nil, fmt.Errorf("json-rpc ID out of int64 range, got %v", id)
+		}
 		return JSONRPCIntID(int(id)), nil
 	default:
 		typ := reflect.TypeOf(id)
--- a/rpc/jsonrpc/types/types_test.go
+++ b/rpc/jsonrpc/types/types_test.go
@@ -4,9 +4,11 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"math"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 type SampleResult struct {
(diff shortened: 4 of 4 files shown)


aljo242 (changes requested):
Thanks for the fix attempt here. I found a blocker in local verification.

I ran:
`go test ./rpc/jsonrpc/types -run TestIDFromInterface_FloatRange -count=1`

Current result fails:
- `two_to_63_rejected`
- `oversized_ids_do_not_collide`

So the boundary/collision case is still not fully rejected in practice.

I also noticed this PR currently includes the `LoadValidatorsFast` state-store changes (same scope as #5857) in addition to JSON-RPC ID logic. Please either:
1) rebase/cherry-pick so this PR contains only the JSON-RPC fix, or
2) clearly stack on #5857 and ensure this branch is reviewed/merged in that order.

Once the float-ID edge case is fixed and branch scope is clean, happy to re-review.

aljo242 (changes requested):
Re-tested on latest head; still seeing a blocker.

I ran:
`go test ./rpc/jsonrpc/types -run TestIDFromInterface_FloatRange -count=20`

Still fails on:
- `two_to_63_rejected`
- `oversized_ids_do_not_collide`

So the 2^63/collision edge case is still not handled correctly in practice.

Also, this PR still carries unrelated `LoadValidatorsFast`/state-store changes (same scope as #5857). Please keep this branch scoped to the JSON-RPC ID fix (or clearly stack/retarget it) before merge.

satyakwok:
Force-pushed `c4a3b5d` — single signed commit on top of current `main`, scope is now strictly the JSON-RPC ID fix.

## Re your reviews (#4321051387, #4328986690)

**Test failures on `two_to_63_rejected` and `oversized_ids_do_not_collide`:** root-caused.

The previous attempt relied on a `float64 -> int -> float64` round-trip check. That check only catches the `2^63` boundary if `int(2^63)` saturates to `math.MinInt64` so that `float64(MinInt64) != 2^63`. Per the Go spec (§6.5), float-to-int conversion of an unrepresentable value is **implementation-dependent**. On amd64 with `cvttsd2si` the residue happens to be `MinInt64` and the check works; on other targets (or future Go versions) it can produce other values, and the check passes incorrectly. That's why the same assertions passed on one host and failed on yours.

This version stops depending on conversion behavior:

```go
if math.IsNaN(id) || math.IsInf(id, 0) { reject "finite number" }
if id != math.Trunc(id) { reject "integer in int range" }
if id >= float64(math.MaxInt64) || id < float64(math.MinInt64) { reject "integer in int range" }
return JSONRPCIntID(int(id)), nil
```

The closed `>=` upper bound is the key: `float64(math.MaxInt64) == 2^63` (because `2^63 - 1` is unrepresentable in float64 and rounds up), so `id == 2^63` is caught regardless of what `int(id)` would have produced.

**Branch scope (`LoadValidatorsFast` leaking in):** removed. This branch now contains only `rpc/jsonrpc/types/types.go`, `rpc/jsonrpc/types/types_test.go`, `rpc/jsonrpc/server/http_json_handler_test.go`, and a `CHANGELOG.md` entry. The `LoadValidatorsFast` work is back in scope on #5857 only.

## Verification

```
go test ./rpc/jsonrpc/types -run TestIDFromInterface_FloatRange -count=20
ok  	github.com/cometbft/cometbft/rpc/jsonrpc/types	0.422s

go test ./rpc/... -count=1
ok  	github.com/cometbft/cometbft/rpc/client                    60s
ok  	github.com/cometbft/cometbft/rpc/core
ok  	github.com/cometbft/cometbft/rpc/grpc
ok  	github.com/cometbft/cometbft/rpc/jsonrpc
ok  	github.com/cometbft/cometbft/rpc/jsonrpc/client
ok  	github.com/cometbft/cometbft/rpc/jsonrpc/server
ok  	github.com/cometbft/cometbft/rpc/jsonrpc/types
```

## Heads-up on a behavior change

The pre-existing `TestJSONRPCID` fixture asserted that `id: 1.3` was silently truncated to `JSONRPCIntID(1)`. Both this PR and the previous push reject fractional IDs (per the JSON-RPC 2.0 "SHOULD NOT contain decimals" guidance). I've moved that fixture from the good-id list to the bad-id list and added a row exercising `id: 1e20` end-to-end through the HTTP handler. Flagging in case you want to debate the strictness — happy to relax `fractional_rejected` to a silent truncation if you prefer to keep the surface unchanged and only fix the int64-overflow part.

Commit is signed (`verified: true, reason: valid`).

aljo242:
minor suggestion: the fractional and out-of-range cases share the same error message ("must be an integer in int range"), which makes them indistinguishable to callers. a value like 1.5 is in int range — it just has a decimal. consider splitting:

```go
if id != math.Trunc(id) {
    return nil, fmt.Errorf("json-rpc ID must be a whole number, got %v", id)
}
if id >= float64(math.MaxInt64) || id < float64(math.MinInt64) {
    return nil, fmt.Errorf("json-rpc ID out of int64 range, got %v", id)
}
```

not blocking — the fix is correct either way.

aljo242 (approved):

satyakwok:
Thanks, addressed the suggestion in 307a41f98. Fractional numeric IDs now return `json-rpc ID must be a whole number`, while finite out-of-int64-range IDs return `json-rpc ID out of int64 range`.

Re-ran:
- `go test ./rpc/jsonrpc/types -run TestIDFromInterface_FloatRange -count=20`
- `go test ./rpc/... -count=1`

aljo242 (approved):
Correct fix. The root issue is that `json.Unmarshal` yields `float64` for all JSON numbers, and bare `int(id)` on an out-of-range input is implementation-defined (Go spec §6.5) — on amd64 it saturates to `math.MinInt` via `cvttsd2si`, collapsing distinct large IDs onto the same value and silently corrupting request/response correlation.

The three guard conditions are each necessary:
- `IsNaN/IsInf`: rejects non-finite inputs
- `id != math.Trunc(id)`: rejects fractional IDs per JSON-RPC 2.0 spec
- Range check: uses `>=` for the upper bound because `float64(math.MaxInt64)` rounds up to exactly `2^63` (which doesn't fit in int64), while the lower bound is strict `<` because `-2^63` is exactly representable and is the valid int64 minimum

The `TestIDFromInterface_FloatRange` suite is thorough, including the regression pin for five previously-colliding oversized inputs and the `math.MinInt64` acceptance case.

PR #5867, opened 2026-05-14
blocksync: fix flaky TestBlockPoolBasic deadlock under -race

songgaoye:
## Summary

Fix the CI failure at:
https://github.com/cometbft/cometbft/actions/runs/25839950951/job/75922993075

Sending to `inputChan` directly inside the `select` case blocked the drain loop when the buffer filled up under `-race`, preventing `requestsCh` from being drained and causing all `bpRequester` goroutines to deadlock.

The fix replaces per-request goroutines with a single dedicated dispatcher goroutine that is the sole consumer of `requestsCh` for the full test lifetime. Teardown order is: `pool.Stop()` first (so the pool stops sending to `requestsCh`), then signal the dispatcher, wait for it to exit, then close peer channels.

## Test plan

- [x] Re-run `TestBlockPoolBasic` with `-race` to confirm no timeout


songgaoye:
CI failure at https://github.com/cometbft/cometbft/actions/runs/25897020257/job/76112147530?pr=5867 is unrelated to this pr, it's a flaky test related to evidence. fix in this https://github.com/cometbft/cometbft/pull/5870

aljo242:
@greptile review

aljo242 (changes requested):
Thanks for tackling this flaky test. I have one blocking concern in the new dispatcher lifecycle.

After `stopDispatcher()` closes `done`, the dispatcher goroutine exits and no longer reads `requestsCh`. Any requester goroutine that is already attempting a send to `pool.requestsCh` can remain blocked indefinitely, which risks a goroutine leak in this test path.

Can we adjust shutdown ordering/coordination so in-flight senders cannot get stuck after dispatcher exit?

Also, the changelog item appears to describe a test-only fix in user-facing bug-fix notes; please confirm this follows current project changelog policy.

songgaoye:
> Thanks for tackling this flaky test. I have one blocking concern in the new dispatcher lifecycle.
> 
> After `stopDispatcher()` closes `done`, the dispatcher goroutine exits and no longer reads `requestsCh`. Any requester goroutine that is already attempting a send to `pool.requestsCh` can remain blocked indefinitely, which risks a goroutine leak in this test path.
> 
> Can we adjust shutdown ordering/coordination so in-flight senders cannot get stuck after dispatcher exit?
> 
> Also, the changelog item appears to describe a test-only fix in user-facing bug-fix notes; please confirm this follows current project changelog policy.

fixed. PTAL!

aljo242 (approved):
Re-reviewed latest updates and re-ran the flaky target test locally:
`go test ./blocksync -run TestBlockPoolBasic -count=20`

Passes consistently in this run. The quit-aware channel sends address the blocked-sender/leak concern from prior review.

aljo242 (approved):
Re-reviewed latest head and validated locally (ok  	github.com/cometbft/cometbft/blocksync	42.627s). LGTM.

aljo242 (approved):
Re-reviewed latest head and validated locally. Targeted blocksync flaky test passes repeatedly. LGTM.

songgaoye:
https://github.com/cometbft/cometbft/pull/5888 fix the ci failure

aljo242 (approved):
Correct fix. `sendRequest`/`sendError` were blocking channel sends that could deadlock if the pool stopped while a requester goroutine was mid-send. The select-on-Quit pattern is the standard fix. Test cleanup refactor eliminates the race between the dispatcher goroutine and the test teardown.

PR #5868, opened 2026-05-14
fix(node): use kernel-assigned ephemeral ports and fix OnStart cleanup

songgaoye:
## Summary
Fix ci failure at: https://github.com/cometbft/cometbft/actions/runs/25806158383/job/75809368250

Replace the `testFreeAddr/randPort` TOCTOU pattern (bind, record port,
close, re-bind) with OS-assigned port 0 for P2P, RPC, and pprof listen
addresses in node tests.

For pprof, refactor `startPprofServer` to call `net.Listen` explicitly
before starting the goroutine, returning the listener so the caller
holds it as a local and the actual bound address is readable after
`Start()` returns.

Fix `OnStart` to clean up any services it started if a later step fails,
using local temporaries committed to node fields only on success.

## Test plan

- [x] Run `go test ./node/... -run "TestNodeStartStop|TestNodeDelayedStart|TestPprofServer|TestNodeNewNodeCustomReactors" -count=1` — all four tests should pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,8 @@
   ([\#5837](https://github.com/cometbft/cometbft/pull/5837))
 - `[abci]` fix deadlock when response callback re-enters the client.
   ([\#5850](https://github.com/cometbft/cometbft/pull/5850))
+- `[node]` fix(node): use kernel-assigned ephemeral ports and fix OnStart cleanup
+  ([\#5868](https://github.com/cometbft/cometbft/pull/5868))
 - `[node]` close partial listeners on startRPC failure
   ([\#5869](https://github.com/cometbft/cometbft/pull/5869))
 - `[consensus]` release cs.mtx before sending to statsMsgQueue
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@
   ([\#5837](https://github.com/cometbft/cometbft/pull/5837))
 - `[abci]` fix deadlock when response callback re-enters the client.
   ([\#5850](https://github.com/cometbft/cometbft/pull/5850))
-- `[node]` fix(node): use kernel-assigned ephemeral ports and fix OnStart cleanup
+- `[node]` use kernel-assigned ephemeral ports and fix `OnStart` cleanup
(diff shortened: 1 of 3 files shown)


aljo242:
@greptile review

aljo242 (changes requested):
Nice progress on reducing node test flakiness. I have a few requested fixes before merge:

1. In `TestPprofServer`, `assert.NotNil` allows execution to continue; if `n.pprofLn` is nil, the next `n.pprofLn.Addr()` call can panic and obscure the real failure. Please switch those to `require.NotNil`.

2. `startPprofServer()` now returns `(nil, nil)` on listen failure, and `OnStart` continues successfully. That makes `pprof_laddr` failures effectively non-fatal despite explicit config. Please either propagate an error (preferred) or document/justify the intended non-fatal behavior.

3. Please normalize the changelog line style to match surrounding entries (drop `fix(node):` prefix in prose).

songgaoye:
> * `il, nil)` on listen failure, and `OnStart` continues successfully. That makes `pprof_laddr` failures effectively non-fatal despite explicit config. Please either propagate an error (preferred) or document/justify the intended non-fatal behavior.
> * Please normalize the changelog line style to match surrounding entries (drop `fix(node):` prefix in prose).

fixed. PTAL

aljo242 (approved):
Re-reviewed latest updates and re-ran targeted node tests:
`go test ./node -run 'TestPprofServer|TestNodeStartStop|TestNodeDelayedStart' -count=10`

Passes in this run. The pprof startup path now propagates listen errors and test assertions are hardened.

aljo242 (approved):
Re-reviewed latest head and validated locally (ok  	github.com/cometbft/cometbft/node	11.224s). LGTM.

aljo242 (approved):
Re-reviewed latest head and validated locally. Targeted node tests pass with current changes. LGTM.

aljo242 (approved):
Correct. Pre-binding with `net.Listen(...:0)` and passing the listener directly eliminates the TOCTOU race in port assignment. The deferred cleanup in `OnStart` properly rolls back all partially-started services on any error path. `newNodeTestConfig` helper eliminates hardcoded port reuse across tests.

PR #5869, opened 2026-05-14
fix(node): close partial listeners on startRPC failure

songgaoye:
---
If `startRPC` opened some listeners before a later Listen call failed,
those listeners were leaked. Fix by deferring cleanup behind a success
flag.
#### PR checklist

- [x] Tests written/updated


aljo242:
@greptile review

aljo242 (changes requested):
Thanks for the fix here. I have one merge-blocking concern on the test coverage for the cleanup behavior.

`TestStartRPCCleansUpOnFailure` currently only asserts that `startRPC()` returns an error. That outcome happens regardless of whether partial listeners are actually cleaned up, so the test doesn't prove the cleanup path and won't catch a regression that leaks the first listener.

Can we strengthen the test to verify cleanup explicitly (e.g., capture the first bound ephemeral port and assert it can be rebound after `startRPC()` fails)?

aljo242:
Non-blocking changelog style nit: the entry currently uses commit-prefix prose (`fix(node): ...`).

For consistency with surrounding entries, please use plain descriptive wording after the module tag (e.g. `[node]` close partial listeners on startRPC failure).

aljo242 (approved):
Re-reviewed latest updates and re-ran local stress on the cleanup test:
`go test ./node -run TestStartRPCCleansUpOnFailure -count=20`

Passes, and the test now explicitly validates that the first listener is reclaimable after failure.

PR #5870, opened 2026-05-18
evidence: fix flaky TestReactorsGossipNoCommittedEvidence

songgaoye:
---
Summary

Fix the CI failure at: https://github.com/cometbft/cometbft/actions/runs/25897020257/job/76112147530?pr=5867 .

Setting the peer state after `pools[0].Update()` caused a race where the
broadcast goroutine woke up with a stale peer height, skipped `evList[1]`,
and blocked for 10s, making the test flaky.

Also replace fixed time.Sleep calls with require.`Eventually/require.Never`
for more robust synchronization and better failure diagnostics.

#### PR checklist

- [x] Tests written/updated


aljo242:
@greptile review

aljo242 (changes requested):
Thanks for the fix attempt here — I reproduced a regression locally that looks blocking for merge.

I ran the target test repeatedly on this PR branch:
`go test ./evidence -run TestReactorsGossipNoCommittedEvidence -count=10`

On this PR branch, it fails intermittently (multiple failures) at:
`require.Never(..., \"second reactor should not have received evidence\")`

Running the same command on `main` passes.

Given this PR is intended to reduce flakiness, can we tighten the synchronization so this no longer flakes under repeated local runs?

Also +1 to the follow-ups from review comments:
1. Add explicit failure messages to both `require.Eventually(...)` checks for better diagnostics.
2. Normalize the changelog entry wording to match surrounding style (drop `test(evidence):` prefix in the changelog line).

songgaoye:
> Thanks for the fix attempt here — I reproduced a regression locally that looks blocking for merge.
> 
> I ran the target test repeatedly on this PR branch: `go test ./evidence -run TestReactorsGossipNoCommittedEvidence -count=10`
> 
> On this PR branch, it fails intermittently (multiple failures) at: `require.Never(..., \"second reactor should not have received evidence\")`
> 
> Running the same command on `main` passes.
> 
> Given this PR is intended to reduce flakiness, can we tighten the synchronization so this no longer flakes under repeated local runs?
> 
> Also +1 to the follow-ups from review comments:
> 
> 1. Add explicit failure messages to both `require.Eventually(...)` checks for better diagnostics.
> 2. Normalize the changelog entry wording to match surrounding style (drop `test(evidence):` prefix in the changelog line).

Fixed.PTAL.
ci failure https://github.com/cometbft/cometbft/actions/runs/26162639891/job/76958211130?pr=5870 fix in the pr https://github.com/cometbft/cometbft/pull/5867

aljo242 (approved):
Re-reviewed latest updates and re-ran the target flake test locally:
`go test ./evidence -run TestReactorsGossipNoCommittedEvidence -count=20`

Passes consistently in this run. The test sequencing/connectivity changes address the race window I previously reproduced.

PR #5875, opened 2026-05-19
feat: Add ml-dsa 65 key type

Eric-Warehime:
---
Adds ML-DSA-65 key type as an option for validators looking for post-quantum secure signing algorithms.

* Uses the `circl` library for all of the crypto.
* Minimally integrates into the places it needs to in comet as an option for validator keys. 
* Adds e2e tests for networks fully based on the new key type

#### PR checklist

- [ ] Tests written/updated


Eric-Warehime:
@greptile review

aljo242 (commented):
Did an initial pass and targeted validations:
`go test ./crypto/mldsa65 ./crypto/encoding ./types ./test/e2e/pkg -run 'TestMlDsa65Network|TestSignAndVerify|TestDeterministicFromSeed|TestRoundTripBytes|TestAddressIs20Bytes|TestCachingVerifierHitsCache|TestBlockMaxDataBytes|TestBlockMaxDataBytesNoEvidence' -count=1`

These checks pass locally.

Eric-Warehime:
> Did an initial pass and targeted validations: `go test ./crypto/mldsa65 ./crypto/encoding ./types ./test/e2e/pkg -run 'TestMlDsa65Network|TestSignAndVerify|TestDeterministicFromSeed|TestRoundTripBytes|TestAddressIs20Bytes|TestCachingVerifierHitsCache|TestBlockMaxDataBytes|TestBlockMaxDataBytesNoEvidence' -count=1`
> 
> These checks pass locally.
> 
> Two things I’d like maintainer confirmation on before final LGTM:
> 
> 1. `types.MaxCommitSigBytes` changed from `const` to `var` (source-level API break for downstream constant expressions). I see the API-breaking changelog note, but this is still a notable compatibility shift.
> 2. Changelog has the same ml-dsa entry under both `FEATURES` and `API-BREAKING` — likely intentional, but worth confirming desired release-note style.

Thank you for the review Alex-Bot.   
The const to var change is intentional--I don't imagine this will break very many people and the fix for it is trivial.  
Changelog entry duplication is also intentional.

aljo242:
ran a local benchmark comparing \`Sign()\` vs ed25519:

\`\`\`
BenchmarkMLDSA65Sign     ~700 µs/op   77 KB/op
BenchmarkEd25519Sign      ~19 µs/op    64 B/op
\`\`\`

~37x slower because \`Sign()\` calls \`sk.UnmarshalBinary(privKey)\` on every call (re-parses the 4032-byte key). \`VerifySignature\` already has an LRU cache to avoid this — \`Sign()\` doesn't.

benchmark used:

\`\`\`go
func BenchmarkMLDSA65Sign(b *testing.B) {
	priv, err := mldsa65.GenPrivKey()
	if err != nil {
		b.Fatal(err)
	}
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		if _, err := priv.Sign(msg); err != nil {
			b.Fatal(err)
		}
	}
}

func BenchmarkEd25519Sign(b *testing.B) {
	priv := ed25519.GenPrivKey()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		if _, err := priv.Sign(msg); err != nil {
			b.Fatal(err)
		}
	}
}
\`\`\`

not a blocker for now, but worth revisiting if this lands on a consensus hot path in production.

aljo242 (approved):

crypto/mldsa65/key.go
```diff
@@ -0,0 +1,255 @@
...
+// ===============================================================================================
+
+var _ crypto.PrivKey = PrivKey{}
+
+// PrivKey is the packed ML-DSA-65 private key (FIPS 204 form).
+type PrivKey []byte
+
+// GenPrivKey generates a fresh ML-DSA-65 key using OS randomness.
+func GenPrivKey() (PrivKey, error) {
+	return genPrivKey(rand.Reader)
```
mattac21: nit: there is a comet crypto lib that other key impls use to get a random reader instead of crypto/rand directly

test/e2e/generator/generate.go
```diff
@@ -15,16 +15,21 @@ import (
...
 	"github.com/cometbft/cometbft/crypto/ed25519"
+	"github.com/cometbft/cometbft/crypto/mldsa65"
 	"github.com/cometbft/cometbft/crypto/secp256k1"
 	e2e "github.com/cometbft/cometbft/test/e2e/pkg"
 	"github.com/cometbft/cometbft/version"
 )
 
+// optLarge is the shared "large" choice token used by both the topology and
+// ABCI-delay dimensions.
+const optLarge = "large"
```
mattac21, line 26: the "large" network timeouts here are necessary because the signing and verification with these new keys is too slow for normal timeouts? I wonder if we use the greptile suggestion to try and fetch from the cache verifier whenever we unmarshal if we could get rid of this

types/block.go
```diff
@@ -588,18 +588,18 @@ const (
...
+
+// 4 bytes for field tags + 2 bytes for signature LEN (varint; 2 bytes covers
+// signatures up to 16383 bytes, which fits ML-DSA-65's 3309-byte sigs) +
+// 1 byte for validator address LEN + 1 byte for timestamp LEN + 3 misc.
+const maxCommitSigProtoEncOverhead = 4 + 2 + 1 + 1 + 3
+
+// MaxCommitSigBytes is the max commit sig size: MaxSignatureSize bytes for the
+// signature, 20 bytes for the address, 1 byte for the flag, 14 bytes for the
+// timestamp, plus proto framing overhead.
+var MaxCommitSigBytes = int64(MaxSignatureSize + 20 + 1 + 14 + maxCommitSigProtoEncOverhead)
```
mattac21, line 602: Agree this is a tiny change and not a big deal of a break, but why did this change to a var in the first place? this is still technically a const right? the new key types size isnt actually a var so the result of MaxSignatureSize should be constant even tho its declared as a var

mattac21 (approved):

types/block.go
Eric-Warehime: `MaxSignatureSIze` is computed via function calls [here](https://github.com/cometbft/cometbft/blob/7cb35171b176ad9f8079345e89d7d5544fd2c9c5/types/signable.go#L12) so it can't be const since golang doesn't support compile time functions.

test/e2e/generator/generate.go
Eric-Warehime: No, it's just a lint fix.

PR #5878, opened 2026-05-20
p2p: fix race in TestTransportMultiplexAcceptNonBlocking

songgaoye:
Fix the CI failure at: https://github.com/cometbft/cometbft/actions/runs/25838130793/job/75917496713

`TestTransportMultiplexAcceptNonBlocking` was flaky due to two issues:

- **Race:** the slow peer started its broken handshake after the fast peer finished dialing, but before `Accept()` returned — so its auth error could land on `mt.acceptc` first. Fix: introduce `acceptedc`, closed after `Accept()` returns, so the slow peer only proceeds after the fast peer is already accepted.
- **Goroutine leak:** an early `return` on dial error skipped `close(fastc)`, leaving the slow peer goroutine blocked. Fix: restructure so `errc` is always closed, and drain with `for range errc` to wait for both goroutines to exit.


aljo242 (approved):
Reviewed and validated locally.

Targeted stress run:
`go test ./p2p -run TestTransportMultiplexAcceptNonBlocking -count=30`

Passes in this run. The accepted-channel sequencing and err-channel draining look like the right fix for both the race and goroutine leak.

aljo242 (approved):
Re-reviewed latest head and validated locally (ok  	github.com/cometbft/cometbft/p2p	0.904s). LGTM.

aljo242 (approved):
Re-reviewed latest head and validated locally. Targeted p2p flaky test now passes repeatedly. LGTM.

PR #5879, opened 2026-05-20
fix(lp2p): fallback to conn remote addr when resolving inbound peer

songgaoye:
## Summary

`handleStream` calls `resolvePeer` to look up or provision the incoming peer, but `resolvePeer` relies on the peerstore having the peer's address. libp2p's identify protocol populates the peerstore asynchronously — there is a window where A has opened a CometBFT stream to B but B's peerstore is still empty for A, causing `resolvePeer` to fail and the first message to be dropped.

The fix passes the connection's remote multiaddr directly into `resolvePeer` as a fallback. When the peerstore has no addresses for the peer, `resolvePeer` constructs the `addrInfo` inline from the live connection instead of returning an error. This closes the race without any peerstore side effects.

---

#### PR checklist

- [x] Tests written/updated


aljo242 (approved):
Reviewed and validated locally (ok  	github.com/cometbft/cometbft/lp2p	42.972s). LGTM.

aljo242 (approved):
Reviewed and validated locally. Targeted lp2p tests passed repeatedly. LGTM.

lp2p/switch.go
```diff
@@ -404,6 +404,10 @@ func (s *Switch) handleStream(stream network.Stream) {
 	}
 
 	// 3. Retrieve the peer from the peerSet (or provision if it's not).
+	// Identify may not have populated the peerstore yet; seed it from
+	// the active connection so resolvePeer can provision the peer.
+	s.host.Peerstore().AddAddr(peerID, stream.Conn().RemoteMultiaddr(), peerstore.TempAddrTTL)
```
swift1337: Hey! I need some time to assess whether we should add this or not. No actions required from your end.

swift1337 (changes requested):

lp2p/switch.go
songgaoye: Make sense. My intent here was only to bridge the window where the inbound stream arrives before identify has populated the peerstore, using the address from the live connection and only with `TempAddrTTL`. I did not see this as widening the long-term peer persistence semantics, but happy to revisit if you think should be handled at a different layer.

lp2p/switch_test.go
```diff
@@ -582,6 +584,68 @@ func TestSwitch(t *testing.T) {
 	})
 }
 
+// TestHandleStreamResolvePeerBeforeIdentify reproduces the race where B receives
+// a CometBFT stream from A before B's identify goroutine has populated B's peerstore
+// with A's address. We slow A's identify responder to hold that goroutine open while
+// A's own identify completes and A opens the stream.
+func TestHandleStreamResolvePeerBeforeIdentify(t *testing.T) {
```
swift1337: Move `TestHandleStreamResolvePeerBeforeIdentify` to TestSwitch's `t.Run("ResolvePeerBeforeIdentity", ...)` and follow other code conventions from this codebase (arrange, act, assert).

lp2p/switch.go
swift1337: I think this patch is more suitable:
```patch
diff --git a/lp2p/switch.go b/lp2p/switch.go
index cf96d502f..a068523b2 100644
--- a/lp2p/switch.go
+++ b/lp2p/switch.go
@@ -16,6 +16,7 @@ import (
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/libp2p/go-libp2p/core/peerstore"
 	"github.com/libp2p/go-libp2p/core/protocol"
+	ma "github.com/multiformats/go-multiaddr"
 	"github.com/pkg/errors"
 )
 
@@ -404,11 +405,7 @@ func (s *Switch) handleStream(stream network.Stream) {
 	}
 
 	// 3. Retrieve the peer from the peerSet (or provision if it's not).
-	// Identify may not have populated the peerstore yet; seed it from
-	// the active connection so resolvePeer can provision the peer.
-	s.host.Peerstore().AddAddr(peerID, stream.Conn().RemoteMultiaddr(), peerstore.TempAddrTTL)
-
-	peer, err := s.resolvePeer(peerID)
+	peer, err := s.resolvePeer(peerID, stream.Conn().RemoteMultiaddr())
 	if err != nil {
 		s.Logger.Error("Failed to resolve peer", "protocol", protocolID, "peer_id", peerID.String(), "err", err)
 		return
@@ -469,7 +466,7 @@ func (s *Switch) handleStream(stream network.Stream) {
 	s.reactors.Receive(reactor.name, messageType, envelope, priority)
 }
 
-func (s *Switch) resolvePeer(id peer.ID) (p2p.Peer, error) {
+func (s *Switch) resolvePeer(id peer.ID, connRemoteAddr ma.Multiaddr) (p2p.Peer, error) {
 	key := peerIDToKey(id)
 
 	// peer exists (99% of the time)
@@ -477,9 +474,16 @@ func (s *Switch) resolvePeer(id peer.ID) (p2p.Peer, error) {
 		return peer, nil
 	}
 
+	// addrInfo most likely exists...
 	addrInfo := s.host.Peerstore().PeerInfo(id)
+
+	// ... but peer identity happens async to the stream handling,
+	// thus peerInfo might be resolved later. fallback to connRemoteAddr.
 	if len(addrInfo.Addrs) == 0 {
-		return nil, errors.New("peer has no addresses in peerstore")
+		addrInfo = peer.AddrInfo{
+			ID:    id,
+			Addrs: []ma.Multiaddr{connRemoteAddr},
+		}
 	}
 
 	var isPrivate, isPersistent, isUnconditional bool
```

swift1337 (changes requested):

lp2p/switch.go
songgaoye: Nice catch @swift1337  — passing it as a fallback parameter is much cleaner than side-effecting the peerstore. I like it. Thank you!

aljo242:
@swift1337 to look at again

aljo242 (changes requested):
`switch.go:477-481`: comment uses the wrong variable name and unclear phrasing:

```go
// addrInfo most likely exists...
addrInfo := s.host.Peerstore().PeerInfo(id)

// ... but peer identity happens async to the stream handling,
// thus peerInfo might be resolved later. fallback to connRemoteAddr.
```

`peerInfo` in the second comment should be `addrInfo`. "peer identity happens async" should be "peer identification runs asynchronously". "resolved later" means "not yet populated".

aljo242 (approved):

aljo242 (approved):
LGTM — fallback to conn remote addr is correct and the test cleanly reproduces the race.

swift1337 (approved):

PR #5889, opened 2026-05-22
refactor(blocksync): replace numPending int32 with atomic.Int32 and document `BlockPool` field ownership

songgaoye:
## Summary

- Replace `numPending int32` + `atomic.LoadInt32`/`atomic.AddInt32` calls
  with `atomic.Int32` and its methods for type-safe atomic access
- Group BlockPool fields with ownership comments: Immutable, Protected by
  mtx, Thread-safe (channels)

---

#### PR checklist

- [ ] Tests written/updated


aljo242 (changes requested):
`numPending atomic.Int32` is left without a section comment, floating between "Protected by mtx" and "Thread-safe (channels)". but it's accessed without the mutex in `AddBlock` and `bpRequester.reset` — so it's not mtx-protected. given the stated goal of this PR is to document field ownership, it needs its own label, e.g. `// Accessed atomically; no mtx required.`

aljo242 (approved):

swift1337 (approved):

PR #5864, opened 2026-05-11
consensus: proposer self-verifies its own vote extension before broadcast (#5204)

satyakwok:
Closes #5204.

## Bug

If the application's `ExtendVote` produces data that its own `VerifyVoteExtension` rejects, the chain deadlocks: the proposer happily advances to the next height while every other validator loops verifying the same invalid extension and never reaches 2/3 precommits. Issue reporter 's log trace shows exactly this — proposer at height 12, peers stuck at height 11 with repeated `VerifyVoteExtensionHandler Price=0` errors.

## Fix

Per the maintainer-blessed **Option A** ([@liangping](https://github.com/liangping) in the issue thread): immediately after `ExtendVote` returns on the proposer, call `VerifyVoteExtension` on the resulting extension. On self-rejection, panic with a clear error pointing at the inconsistency between the application's `ExtendVote` and `VerifyVoteExtension` handlers — better operator signal than a silent network-wide stall.

```go
ext, err := cs.blockExec.ExtendVote(context.TODO(), vote, block, cs.state)
if err != nil {
    return nil, err
}
vote.Extension = ext

if len(vote.Extension) > 0 {
    if err := cs.blockExec.VerifyVoteExtension(context.TODO(), vote); err != nil {
        panic(fmt.Errorf(
            "proposer's own vote extension failed VerifyVoteExtension at height %d: %w; "+
                "application ExtendVote and VerifyVoteExtension handlers are inconsistent",
            vote.Height, err,
        ))
    }
}
```

The `len(vote.Extension) > 0` guard preserves existing behavior for the "absent but required" case — that path is caught downstream by `SignAndCheckVote` and produces a recoverable error (the `TestVoteExtensionEnableHeight/extension_absent_but_required` subtest passes unchanged).

## Tests

```
$ go test ./consensus/ -run "TestStateVoteExtension|TestVoteExtensionEnableHeight|TestVerifyVoteExtensionNotCalledOnAbsentPrecommit" -count=1
ok  	github.com/cometbft/cometbft/consensus	0.312s
```

All five `TestVoteExtensionEnableHeight` subtests pass:
- `extension_present_but_not_enabled`
- `extension_absent_but_not_required`
- `extension_present_and_required`
- `extension_absent_but_required`
- `extension_absent_but_required_in_future_height`

[...]

PR #5890, opened 2026-05-23
consensus: proposer self-verifies its own vote extension before broadcast (#5204) [v2]

### LINKED ISSUE #5204: abci: Consensus Deadlock Caused by Proposer Vote Extension Verification Failure
### Setup

`
github.com/cometbft/cometbft@v0.38.12
`

NO

https://github.com/sideprotocol/side

**Environment**:
- **OS** (e.g. from /etc/os-release): Linux/Macos

### What happened?

The blockchain will stop progressing if the proposer's vote extension fails to pass the VerifyVoteExtensionHandler. This occurs because the proposer does not call VerifyVoteExtensionHandler on its own extension, whereas all other validators do. As a result, the proposer node advances to the next height, while the other nodes repeatedly call VerifyVoteExtensionHandler, causing a consensus deadlock.

All nodes should call `VerifyVoteExtensionHandler` to verify data before it included in voteExt.

### How to reproduce it

```go

func (h *PriceOracleVoteExtHandler) ExtendVoteHandler() sdk.ExtendVoteHandler {
	return func(ctx sdk.Context, req *abci.RequestExtendVote) (*abci.ResponseExtendVote, error) {

		if req.Height < ctx.ConsensusParams().Abci.VoteExtensionsEnableHeight {
			return &abci.ResponseExtendVote{}, nil
		}

		proposer := hex.EncodeToString(req.ProposerAddress)
		h.logger.Warn("Prepare vote extension", "proposer", CURRENT_PROPOSER, "my", LOCAL_ADDRESS)

		prices := make(map[string]string)
                // add a mock price data for non-proposer nodes
		if proposer != LOCAL_ADDRESS {
			prices["ATOMUSD"]="123"
		}

		voteExt := types.OracleVoteExtension{
			Height:   req.Height,
			Prices:   prices,
			Proposer: proposer,
		}

		bz, err := voteExt.Marshal()
		if err != nil {
			return nil, fmt.Errorf("failed to marshal vote extension: %w", err)
		}
		
		return &abci.ResponseExtendVote{VoteExtension: bz}, nil
	}
}
```

```go

func (h *PriceOracleVoteExtHandler) VerifyVoteExtensionHandler() sdk.VerifyVoteExtensionHandler {
	return func(ctx sdk.Context, req *abci.RequestVerifyVoteExtension) (*abci.ResponseVerifyVoteExtension, error) {

		if !voteExtensionEnabled(ctx, req.Height) {
			return &abci.ResponseVerifyVoteExtension{Status: abci.ResponseVerifyVoteExtension_ACCEPT}, nil
		}

		if len(req.VoteExtension) == 0 {
			return &abci.ResponseVerifyVoteExtension{Status: abci.ResponseVerifyVoteExtension_REJECT}, nil
		}

		validator := hex.EncodeToString(req.ValidatorAddress)

		var voteExt types.OracleVoteExtension
		err := voteExt.Unmarshal(req.VoteExtension)
		if err != nil {
			return nil, fmt.Errorf("failed to unmarshal vote extension: %w", err)
		}

		if voteExt.Height != req.Height {
			return nil, fmt.Errorf("vote extension height does not match request height; expected: %d, got: %d", req.Height, voteExt.Height)
		}

                 // Since the propoer's price data is empty, will be rejected by other nodes
		if len(voteExt.Prices) == 0  {
			h.logger.Error("VerifyVoteExtensionHandler", "height", voteExt.Height, "validator", validator, "Price", len(voteExt.Prices), "proposer", voteExt.Proposer)
			return &abci.ResponseVerifyVoteExtension{Status: abci.ResponseVerifyVoteExtension_REJECT}, nil
		}

		h.logger.Info("VerifyVoteExtensionHandler", "height", voteExt.Height, "validator", validator, "Price", len(voteExt.Prices), "proposer", voteExt.Proposer)


		return &abci.ResponseVerifyVoteExtension{Status: abci.ResponseVerifyVoteExtension_ACCEPT}, nil
	}
}
```

### Logs

Logs on the proposer: (**height=12**)
```
5:08PM INF Replay: Timeout dur=3000 height=12 module=consensus round=0 step=RoundStepPropose
5:08PM INF Replay: New Step height=12 module=consensus round=0 step=RoundStepPropose
5:08PM INF Replay: Done module=consensus
5:08PM INF Timed out dur=3000 height=12 module=consensus round=0 step=RoundStepPropose

```
Logs on other nodes: (**height=11**)
```
5:05PM ERR VerifyVoteExtensionHandler Price=0 height=0 module=server proposer=93a7c8f465b9a42e63a93be16fdaa76e59330eac validator=93a7c8f465b9a42e63a93be16fdaa76e59330eac
5:05PM INF VerifyVoteExtensionHandler Price=3 height=11 module=server proposer=93a7c8f465b9a42e63a93be16fdaa76e59330eac validator=1b0abcf98fb55f7cffa8ff62a90ac101164e979f
5:05PM INF VerifyVoteExtensionHandler Price=3 height=11 module=server proposer=93a7c8f465b9a42e63a93be16fdaa76e59330eac validator=1b0abcf98fb55f7cffa8ff62a90ac101164e979f
5:05PM ERR VerifyVoteExtensionHandler Price=0 height=0 module=server proposer=93a7c8f465b9a42e63a93be16fdaa76e59330eac validator=93a7c8f465b9a42e63a93be16fdaa76e59330eac
... ...
5:05PM ERR VerifyVoteExtensionHandler Price=0 height=0 module=server proposer=93a7c8f465b9a42e63a93be16fdaa76e59330eac validator=93a7c8f465b9a42e63a93be16fdaa76e59330eac
... ...
```

satyakwok:
Supersedes #5864.

## Why a new PR

#5864 was opened with one unsigned base commit (`d184ae5`) plus five `Merge branch 'main'` commits. Per @mattac21's request, this PR rebases the work onto current `main` as **a single signed commit** with no merge noise.

## What changed since #5864

Greptile flagged two valid issues on the previous PR; both addressed here:

1. **Panic message no longer says "proposer's"** (`consensus/state.go`)
   `signVote` runs on every validator, not just the proposer, so the old wording would mislead operators when a non-proposer hit the panic. The message now leads with `validator %X failed self-verification of its own vote extension at height %d` so the trace identifies the offending node regardless of role.

2. **Self-verify path now has dedicated test coverage** (`consensus/state_test.go`)
   New `TestSelfVerifyVoteExtensionRejectPanics` drives `signVote` directly from the test goroutine (bypassing `receiveRoutine`'s `defer recover()`), with a mock app whose `ExtendVote` returns a non-empty extension that `VerifyVoteExtension` rejects. Asserts the panic fires with the expected message and that both ABCI methods were called.

   The `TestVoteExtensionEnableHeight` `VerifyVoteExtension` mock is **restored** to `Times(numValidators - 1)` — that test's `ExtendVote` mock returns an empty extension, so the self-verify guard is never entered there and the exact count holds. The misleading `// scheduling-dependent count` comment from #5864 is gone.

## Original bug (unchanged from #5864)

If the application's `ExtendVote` produces data its own `VerifyVoteExtension` rejects, the chain deadlocks: the issuing validator advances locally while every peer loops rejecting the invalid extension and 2/3 precommits are never reached. This PR adds an immediate self-verification in `signVote`: after `ExtendVote` returns a non-empty extension, `VerifyVoteExtension` is called on it before broadcasting. On rejection, the node panics with a clear handler-inconsistency message.

Empty extensions are skipped — the absent-but-required case is caught downstream by `SignAndCheckVote` with a recoverable error, preserving existing behavior.

## Validation

```
go test ./consensus -run 'TestSelfVerifyVoteExtensionRejectPanics|TestStateVoteExtension|TestVoteExtensionEnableHeight|TestVerifyVoteExtensionNotCalledOnAbsentPrecommit|TestExtendVoteCalledWhenEnabled' -count=1
ok  	github.com/cometbft/cometbft/consensus	1.042s
```

Closes #5204


aljo242 (changes requested):
: use `context.Background()` instead of `context.TODO()`. `TODO` signals a known placeholder to be replaced; if there's no intent to thread a real context through `signVote`, `Background` is correct.

test coverage gap: `TestSelfVerifyVoteExtensionRejectPanics` covers the reject/panic path. `TestVoteExtensionEnableHeight` uses an empty extension so it skips self-verify entirely. there's no test for the happy path: non-empty extension that `VerifyVoteExtension` accepts. a test confirming `signVote` succeeds and doesn't panic in that case would complete the coverage.

aljo242 (changes requested):
`state.go:2474`: use `context.Background()` instead of `context.TODO()`. `TODO` signals a known placeholder to be replaced; if there's no intent to thread a real context through `signVote`, `Background` is correct.

test coverage gap: `TestSelfVerifyVoteExtensionRejectPanics` covers the reject/panic path. `TestVoteExtensionEnableHeight` uses an empty extension so it skips self-verify entirely. there's no test for the happy path: non-empty extension that `VerifyVoteExtension` accepts. a test confirming `signVote` succeeds and doesn't panic in that case would complete the coverage.

satyakwok:
Thanks, addressed both in 10e33fb3a.

- Replaced the self-verify `context.TODO()` with `context.Background()`.
- Added `TestSelfVerifyVoteExtensionAccepts` to cover the happy path where a non-empty extension is accepted by `VerifyVoteExtension` and `signVote` completes without panic.

Targeted test:
`go test ./consensus -run 'TestSelfVerifyVoteExtensionRejectPanics|TestSelfVerifyVoteExtensionAccepts|TestStateVoteExtension|TestVoteExtensionEnableHeight|TestVerifyVoteExtensionNotCalledOnAbsentPrecommit|TestExtendVoteCalledWhenEnabled' -count=1`

aljo242:
missing changelog entry — nothing in `CHANGELOG.md` for this fix.

one behavior question: `signVote` is called from `receiveRoutine`, which wraps its body in `defer recover()`. the panic will be caught there and logged rather than crashing the process. the pr description frames this as non-recoverable, but in practice the validator just silently skips casting its precommit for the round. is that acceptable, or should this be a fatal log + `os.Exit` to force the operator to fix their app?

satyakwok:
Thanks for the review.

**Changelog:** added an entry under `UNRELEASED` → `### BUG FIXES`.

**On the recover behavior:** the panic isn't actually swallowed-and-skipped. `signVote` runs inside `receiveRoutine`, whose deferred `recover` doesn't continue the loop — it logs `CONSENSUS FAILURE!!!` with the stack and then calls `onExit(cs)`, which stops the WAL and closes `cs.done`, halting the consensus routine. The comment there states the intent directly:

> NOTE: We most probably shouldn't be running any further when there is some unexpected panic. [...] halting the chain upon unexpected consensus bugs sounds like the better option.

So the validator doesn't quietly skip its precommit and move on — consensus stops, loudly, which is the operator-visible "go fix your app" outcome you're describing.

It also keeps this consistent with how `signVote` already handles the same class of bug: it panics on a `non-recoverable error when signing vote` (just below the new code) and `applyCommit`/vote-extension absence-vs-presence mismatch panic the same way. A bespoke `os.Exit` here would diverge from that pattern and skip the clean WAL shutdown the recover path gives us.

So I've kept the panic. If you'd still prefer an explicit fatal exit for this specific case I'm happy to switch, but I'd lean toward staying consistent with the surrounding `signVote` panics and the existing recover → `onExit` halt.

Also rebased onto `main` and signed all commits to clear the failing `check-signed-commits`.

aljo242 (approved):
Both of @aljo242's CHANGES_REQUESTED items are addressed in this v2:
1. `context.Background()` is used at state.go:2474 (not `TODO`).
2. `TestSelfVerifyVoteExtensionAccepts` covers the happy path — non-empty extension accepted by `VerifyVoteExtension` signs without panic.

The guard logic is correct: only fires when `len(vote.Extension) > 0`, panics on reject, preserves the empty-extension recoverable-error path downstream. Panic message leads with `validator %X` (not "proposer's"), which is correct since every validator signs precommits.

⚠️ This is a fork PR — the full CI suite (lint, unit tests, e2e) has not run. A maintainer needs to approve CI before merging.

PR #5900, opened 2026-05-26
Problem: node could get stuck in blocksync reactor

### LINKED ISSUE #5801: Malicious Peer Can Disrupt Blocksync Synchronization

yihuang:
Closes: #5801

Solution:
- introduce an objective blocksync tracker that decide when to escape blocksync loop without rely on peer reported information.

The basic ideas:
* only rely on local wall clock and block times to decide if we need to escape blocksync loop.
* use moving average on block times to estimate the actual block production rate.
* use local wall clock time and synced tip time to decide if we are behind or close to the tip.
* If we are far behind, use bigger timeout to tolerate network glitch.
* If we are close to tip, use smaller timeouts but still make sure smaller than actual block production rate, to defend against the case that attacker feed freshly produced blocks.

```
   isCaughtUpAt(now):
     receiveInterval = now - lastBlockWallTime

     ┌─ Far behind (blockAge > prodInterval × 100):
     │   timeout = max(noBlockTimeout, blockAge / 100)
     │   return receiveInterval ≥ timeout
     │   → 1h behind → 36s timeout; 10min → 6s timeout (floor)
     │
     ├─ Near tip (≥ 3 production samples):
     │   return receiveInterval ≥ prodInterval × 0.6
     │   → catches throttling while tolerating transient glitches
     │
     └─ Floor:
         return receiveInterval ≥ noBlockTimeout
         → catches complete stalls
 ```

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/blocksync/moving_average.go
+++ b/blocksync/moving_average.go
@@ -15,8 +15,8 @@ func NewMovingAverage(windowSize int) *MovingAverage {
 	return &MovingAverage{buf: make([]time.Duration, windowSize)}
 }
 
-// Add records a new value and returns the current average.
-// Returns 0, false if no values have been added yet.
+// Add records a new value and returns the updated average.
+// Always returns ok=true; use Avg to check whether any values exist.
 func (ma *MovingAverage) Add(val time.Duration) (avg time.Duration, ok bool) {
 	if ma.count < len(ma.buf) {
 		ma.count++
--- a/blocksync/moving_average.go
+++ b/blocksync/moving_average.go
@@ -16,8 +16,7 @@ func NewMovingAverage(windowSize int) *MovingAverage {
 }
 
 // Add records a new value and returns the updated average.
-// Always returns ok=true; use Avg to check whether any values exist.
-func (ma *MovingAverage) Add(val time.Duration) (avg time.Duration, ok bool) {
+func (ma *MovingAverage) Add(val time.Duration) time.Duration {
 	if ma.count < len(ma.buf) {
 		ma.count++
 	} else {
@@ -26,7 +25,7 @@ func (ma *MovingAverage) Add(val time.Duration) (avg time.Duration, ok bool) {
 	ma.buf[ma.index] = val
 	ma.sum += val
 	ma.index = (ma.index + 1) % len(ma.buf)
-	return ma.sum / time.Duration(ma.count), true
+	return ma.sum / time.Duration(ma.count)
 }
 
 // Avg returns the current moving average, or 0,false if empty.
--- a/blocksync/moving_average.go
+++ b/blocksync/moving_average.go
@@ -1,0 +1,46 @@
+package blocksync
+
+import "time"
+
+// MovingAverage maintains a sliding window average of time.Duration values.
+type MovingAverage struct {
+	buf   []time.Duration
+	index int
+	count int
+	sum   time.Duration
+}
+
+// NewMovingAverage creates a moving average with the given window size.
+func NewMovingAverage(windowSize int) *MovingAverage {
+	return &MovingAverage{buf: make([]time.Duration, windowSize)}
+}
+
+// Add records a new value and returns the current average.
+// Returns 0, false if no values have been added yet.
+func (ma *MovingAverage) Add(val time.Duration) (avg time.Duration, ok bool) {
+	if ma.count < len(ma.buf) {
+		ma.count++
+	} else {
+		ma.sum -= ma.buf[ma.index]
+	}
+	ma.buf[ma.index] = val
+	ma.sum += val
+	ma.index = (ma.index + 1) % len(ma.buf)
+	return ma.sum / time.Duration(ma.count), true
+}
+
+// Avg returns the current moving average, or 0,false if empty.
+func (ma *MovingAverage) Avg() (time.Duration, bool) {
+	if ma.count == 0 {
+		return 0, false
+	}
+	return ma.sum / time.Duration(ma.count), true
+}
+
+// Len returns the number of values in the window.
+func (ma *MovingAverage) Len() int { return ma.count }
+
+// Reset clears all values.
+func (ma *MovingAverage) Reset() {
+	ma.index, ma.count, ma.sum = 0, 0, 0
+}
--- a/blocksync/pool.go
+++ b/blocksync/pool.go
@@ -56,6 +56,9 @@ const (
 
 var peerTimeout = 15 * time.Second // not const so we can override with tests
 
+// not const so we can override with tests
+var noBlockTimeout = 30 * time.Second
+
 /*
 	Peers self report their heights when we join the block pool.
 	Starting from our latest pool.height, we request blocks
@@ -77,10 +80,12 @@ type BlockPool struct {
 	// Set once in OnStart, then immutable.
 	startTime time.Time
 
-	// Protected by mtx.
-	mtx           cmtsync.Mutex
-	requesters    map[int64]*bpRequester
-	height        int64 // the lowest key in requesters.
+	mtx         cmtsync.Mutex
+	syncTracker SyncTracker // tracks sync rate vs production rate for escape detection
+	// block requests
+	requesters map[int64]*bpRequester
+	height     int64 // the lowest key in requesters.
+	// peers
 	peers         map[p2p.ID]*bpPeer
 	bannedPeers   map[p2p.ID]time.Time
 	sortedPeers   []*bpPeer // sorted by curRate, highest first
@@ -110,6 +115,7 @@ func NewBlockPool(start int64, requestsCh chan<- BlockRequest, errorsCh chan<- p
 		requesters:  make(map[int64]*bpRequester),
 		height:      start,
 		startHeight: start,
+		syncTracker: *NewSyncTracker(noBlockTimeout),
 
 		requestsCh: requestsCh,
 		errorsCh:   errorsCh,
@@ -226,7 +232,6 @@ func (pool *BlockPool) HasPendingRequestFrom(peerID p2p.ID) bool {
 }
 
 // IsCaughtUp returns true if this node is caught up, false - otherwise.
-// TODO: relax conditions, prevent abuse.
 func (pool *BlockPool) IsCaughtUp() bool {
 	pool.mtx.Lock()
 	defer pool.mtx.Unlock()
@@ -237,15 +242,18 @@ func (pool *BlockPool) IsCaughtUp() bool {
 		return false
 	}
 
-	// Some conditions to determine if we're caught up.
-	// Ensures we've either received a block or waited some amount of time,
-	// and that we're synced to the highest known height.
-	// Note we use maxPeerHeight - 1 because to sync block H requires block H+1
-	// to verify the LastCommit.
-	receivedBlockOrTimedOut := pool.height > 0 || time.Since(pool.startTime) > 5*time.Second
-	ourChainIsLongestAmongPeers := pool.maxPeerHeight == 0 || pool.height >= (pool.maxPeerHeight-1)
-	isCaughtUp := receivedBlockOrTimedOut && ourChainIsLongestAmongPeers
-	return isCaughtUp
+	// No-progress check: after peerConnWait, delegate to the sync tracker
+	// which uses block header timestamps and wall-clock time to determine
+	// whether we're truly caught up or just being throttled by attackers.
+	sinceStart := time.Since(pool.startTime)
+	if sinceStart >= peerConnWait {
+		if pool.height > pool.startHeight {
+			return pool.syncTracker.IsCaughtUp()
+		}
+		// Genesis: wait for peerConnWait + noBlockTimeout
+		return sinceStart >= peerConnWait+noBlockTimeout
+	}
+	return false
 }
 
 // PeekTwoBlocks returns blocks at pool.height and pool.height+1. We need to
@@ -270,7 +278,7 @@ func (pool *BlockPool) PeekTwoBlocks() (first, second *types.Block, firstExtComm
 }
 
 // PopRequest removes the requester at pool.height and increments pool.height.
-func (pool *BlockPool) PopRequest() {
+func (pool *BlockPool) PopRequest(blockTime time.Time) {
 	pool.mtx.Lock()
 	defer pool.mtx.Unlock()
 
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

**Comment by aljo242** [comment]
ci has lint failures to fix:
- `sync_tracker.go:16` - gofmt
- `pool_test.go:146,209,360,478` - `QF1008`: `second.Header.Time` -> `second.Time`

bug in `pool.go:466`:
```go
pool.Logger.Debug("Removing peer", peerID)
```
`peerID` is passed as a key, not a value. should be `pool.Logger.Debug("Removing peer", "peer", peerID)`

design question on the 0.6x threshold in `isCaughtUpAt`: the `hasProd && s.prodMA.Len() >= 3` branch fires after only 3 production samples. a node syncing at 1.25x production rate (1 block/4s on a 5s chain) gets `receiveInterval=4s >= 5s*0.6=3s` and escapes after just 3 blocks even if it's 100 blocks behind. the `blockAge > prodInterval*100` guard only protects if the last synced header is old enough, but drops out quickly once a few recent blocks are processed. worth thinking through whether the 0.6 multiplier or the 3-sample floor needs adjustment.

yihuang:
> design question on the 0.6x threshold in isCaughtUpAt: the hasProd && s.prodMA.Len() >= 3 branch fires after only 3 production samples. a node syncing at 1.25x production rate (1 block/4s on a 5s chain) gets receiveInterval=4s >= 5s*0.6=3s and escapes after just 3 blocks even if it's 100 blocks behind. the blockAge > prodInterval*100 guard only protects if the last synced header is old enough, but drops out quickly once a few recent blocks are processed. worth thinking through whether the 0.6 multiplier or the 3-sample floor needs adjustment.

I did slight adjustment to make the timeout smoothly transition from `noBlockTimeout`(30s) at 100 blocks interval to 0 at chain tip.

aljo242:
@greptile review

blocksync/pool_test.go
```diff
@@ -143,7 +143,7 @@ func TestBlockPoolBasic(t *testing.T) {
 			}
 			first, second, _ := pool.PeekTwoBlocks()
 			if first != nil && second != nil {
-				pool.PopRequest()
+				pool.PopRequest(second.Time)
```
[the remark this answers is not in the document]
yihuang: fixed

blocksync/sync_tracker.go
```diff
@@ -0,0 +1,115 @@
...
+	// Regime 2: Near tip. Smooth transition from aggressive rate check
+	// to the noBlockTimeout floor as we get further from the tip.
+	//   At blockAge = 0:              threshold = prodInterval × 0.6
+	//   At blockAge = prodInterval × 100: threshold = noBlockTimeout
+	// This avoids a hard cliff at the regime boundary.
+	if hasProd && s.prodMA.Len() >= 3 {
+		ratio := float64(now.Sub(s.lastBlockHeaderTime)) / float64(prodInterval*100)
+		if ratio > 1 {
+			ratio = 1
+		}
```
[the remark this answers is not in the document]
yihuang: fixed

blocksync/sync_tracker.go
```diff
@@ -0,0 +1,115 @@
...
+	if hasProd {
+		blockAge := now.Sub(s.lastBlockHeaderTime)
+		if blockAge > prodInterval*100 {
+			timeout := blockAge / 100
+			if timeout < s.noBlockTimeout {
+				timeout = s.noBlockTimeout
+			}
+			return receiveInterval >= timeout
+		}
+	}
```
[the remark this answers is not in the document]
yihuang: fixed

blocksync/pool.go
```diff
@@ -234,15 +239,18 @@ func (pool *BlockPool) IsCaughtUp() bool {
...
+	// whether we're truly caught up or just being throttled by attackers.
+	sinceStart := time.Since(pool.startTime)
+	if sinceStart >= peerConnWait {
+		if pool.height > pool.startHeight {
+			return pool.syncTracker.IsCaughtUp()
+		}
+		// Genesis: wait for peerConnWait + noBlockTimeout
+		return sinceStart >= peerConnWait+noBlockTimeout
+	}
+	return false
```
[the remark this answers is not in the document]
yihuang: fixed

aljo242:
ci is red (lint + e2e `networks/ci`) — needs to be clean before merge.

also missing a changelog entry — pr checklist marks it unchecked, confirmed nothing in `CHANGELOG.md`.

blocksync/pool.go
```diff
@@ -646,6 +665,7 @@ func (peer *bpPeer) resetMonitor() {
 }
 
 func (peer *bpPeer) resetTimeout() {
+	peer.logger.Debug("Resetting peer timeout", "peer", peer.id, "numPending", peer.numPending)
```
yihuang:
```suggestion
```

yihuang:
@greptile review

blocksync/moving_average.go
```diff
@@ -0,0 +1,45 @@
+package blocksync
```
swift1337: let's move it to `internal/movavg`

blocksync/pool.go
```diff
@@ -56,6 +56,9 @@ const (
 
 var peerTimeout = 15 * time.Second // not const so we can override with tests
 
+// not const so we can override with tests
+var noBlockTimeout = 30 * time.Second
```
swift1337: It should be an argument inside the constructor

blocksync/pool.go
```diff
@@ -267,7 +285,7 @@ func (pool *BlockPool) PeekTwoBlocks() (first, second *types.Block, firstExtComm
 }
 
 // PopRequest removes the requester at pool.height and increments pool.height.
-func (pool *BlockPool) PopRequest() {
+func (pool *BlockPool) PopRequest(blockTime time.Time) {
```
swift1337: Why do we need to provide this arg if a block always comes from `pool.PeekTwoBlocks()` ?

swift1337 (changes requested):

blocksync/pool.go
yihuang: good point, I've changed it to get block information from `BlockPool` itself.

blocksync/pool.go
yihuang: fixed.

blocksync/moving_average.go
yihuang: done

blocksync/reactor.go
```diff
@@ -131,7 +134,7 @@ func NewReactor(
 	if startHeight == 1 {
 		startHeight = state.InitialHeight
 	}
-	pool := NewBlockPool(startHeight, requestsCh, errorsCh)
+	pool := NewBlockPool(startHeight, requestsCh, errorsCh, defaultNoBlockTimeout)
```
swift1337, line 137: What are the consequences of having a short or a long noBlockTimeout? In my opinion, 30s is a lot for most chains. What do you think about calculating a "subjective" noBlockTimeout based on consensus timeouts (from config), so the value would be relative to block times?

yihuang: > What are the consequences of having a short or a long noBlockTimeout?

Just different level of tolerance for network latency during blocksync.

> In my opinion, 30s is a lot for most chains. What do you think about calculating a "subjective" noBlockTimeout based on consensus timeouts (from config), so the value would be relative to block times?

Good point, I think there are two parts here:
1. Use consensus timeouts as the initial estimated block time.
2. noBlockTimeout based on block time.

The risk is the consensus timeout config in fullnode may not be faithful to the network condition.

swift1337: > noBlockTimeout based on block time

I think using timeouts based on our _local_ consensus configs is better, as we don't have to trust other peers' variables on this.

aljo242:
a few things blocking merge:

1. `blocksync/pool.go:84`: `syncTracker` field is space-indented instead of tab — fails gofmt. one-char fix.
2. `blocksync/reactor.go:35`: `// defaultNoBlockTimeout is the` comment is truncated mid-sentence.
3. swift1337's open question about deriving `noBlockTimeout` from consensus config (rather than hardcoding 30s) is still unresolved in the thread.
4. CHANGELOG entry is missing (checklist item unchecked).

aljo242 (approved):
Well-designed fix for the malicious-peer blocksync stall. Two-regime `SyncTracker` (far-behind proportional timeout + near-tip rate check at 0.6× production interval, with `noBlockTimeout` floor) is more robust than a pure timer. Moving average window of 10 is reasonable. `TestBlockPoolTwoMaliciousPeersStaggered` covers the exact attack scenario (two staggered reconnecting peers at MaxInt64). The `noBlockTimeout` parameter addition to `NewBlockPool` is a clean API change — all callers updated.

PR #5901, opened 2026-05-29
fix(state)!: median time no longer considers nil precommits

vladjdk:
Discard nil and absent votes from the median BFT time calculation to match the behaviour in VerifyCommit tallies.

===

credit to @ehdus829 for discovering, reporting, and recommending a patch for the bug.


aljo242 (approved):

vladjdk:
@greptile review again

PR #5904, opened 2026-06-01
feat(abci/server): accept pre-bound listener in socket and gRPC servers

songgaoye:
## Summary

- Add `NewSocketServerWithListener` and `NewGRPCServerWithListener` constructors to the ABCI socket and gRPC servers, following the same convention as `net/http.Serve` and `grpc.Server.Serve`
- `OnStart` skips `net.Listen` when a listener is already set, so both constructors share the same start path

## Motivation

The existing constructors always call `net.Listen` inside `OnStart`, which has two drawbacks:

- **TOCTOU**: there is a race between selecting a port and binding it — another process can grab the port in that window. A pre-bound listener eliminates the race entirely.
- **Testability**: tests must pick a fixed port upfront, risking conflicts in parallel runs. With `net.Listen("tcp", "127.0.0.1:0")` the OS assigns an ephemeral port, and the caller knows the exact address before the server starts.

Socket-activated processes (e.g. systemd) also rely on passing pre-bound file descriptors, which this enables.

## Test plan

- [x] `TestNewSocketServerWithListener`: binds an ephemeral port, starts the server via the new constructor, dials it, stops cleanly
- [x] `TestNewGRPCServerWithListener`: same pattern for the gRPC server
- [ ] Existing tests continue to pass

---

#### PR checklist

- [x] Tests written/updated


aljo242 (approved):
clean, well-motivated change. nil-listener guard in OnStart is the right pattern — existing callers unaffected, new constructors tested for both socket and grpc.

PR #5905, opened 2026-06-03
feat: Comet kms 

Eric-Warehime:
---
Add kms support to cometbft


Eric-Warehime:
@greptile review

aljo242:
missing test for backend sign/pubkey error path — `DefaultValidationRequestHandler` embeds a `RemoteSignerError` in the response when signing fails, but there's no test verifying that path end-to-end. a test where the backend `Sign` returns an error and asserting the client receives the embedded error (not a connection drop) would close the gap.

reconnect, double-sign, and state-persistence tests all look solid.

aljo242 (approved):

PR #5907, opened 2026-06-03
feat: Add secp256k1eth key type

Eric-Warehime:
---

Adds Eth compatible validator key type.


aljo242 (changes requested):
added a go-ethereum cross-compatibility vector test (commit 50999cd) — uses go-ethereum's `testPrivHex` key from `crypto/signature_test.go`, verifies the derived address matches geth's `testAddrHex`, and asserts the deterministic RFC 6979 signature matches the known-good vector.

one remaining ask before merge: the malleable-signature check in `VerifySignature` uses an indirect DER roundtrip to detect high-S:

```go
modified, parseErr := ecdsa.ParseDERSignature(sig.Serialize())
if !sig.IsEqual(modified) { return false }
```

this works (the high-S test passes) but relies on decred normalizing S in its DER encoder, which is non-obvious to future readers. the direct check is clearer:

```go
var s secp256k1.ModNScalar
s.SetByteSlice(sigStr[32:64])
if s.IsOverHalfOrder() { return false }
```

easy swap, no behavior change.

aljo242 (approved):
added a go-ethereum cross-compatibility vector test (commit 50999cd) — uses go-ethereum's `testPrivHex` key from `crypto/signature_test.go`, verifies the derived address matches geth's `testAddrHex`, and asserts the deterministic RFC 6979 signature matches the known-good vector.

suggestion (non-blocking): the malleable-signature check in `VerifySignature` uses an indirect DER roundtrip to detect high-S. the direct check is clearer:

```go
var s secp256k1.ModNScalar
s.SetByteSlice(sigStr[32:64])
if s.IsOverHalfOrder() { return false }
```

test/e2e/networks/secp256k1eth.toml
```diff
@@ -0,0 +1,27 @@
...
+# making it a useful smoke test for the validator key type integration.
+#
+# To run end-to-end (requires docker):
+#   cd test/e2e
+#   make runner
+#   ./build/runner -f networks/secp256k1eth.toml setup
+#   ./build/runner -f networks/secp256k1eth.toml start
+#   ./build/runner -f networks/secp256k1eth.toml test
+
+key_type = "secp256k1eth"
```
swift1337, line 6: Could you also add the same test, but with libp2p enabled?

test/e2e/networks/secp256k1eth.toml
```diff
@@ -0,0 +1,27 @@
...
+# signature scheme (secp256k1eth). It exercises the consensus, gossip, and
+# privval paths with the Ethereum-style 65-byte [R || S || V] signature format,
+# making it a useful smoke test for the validator key type integration.
+#
+# To run end-to-end (requires docker):
+#   cd test/e2e
+#   make runner
+#   ./build/runner -f networks/secp256k1eth.toml setup
+#   ./build/runner -f networks/secp256k1eth.toml start
+#   ./build/runner -f networks/secp256k1eth.toml test
```
swift1337: this comment is not needed (🤖)

aljo242:
the head branch `eric/ethsecp` no longer exists on the remote, so the pr can't be rebased or merged as-is. please re-push the branch rebased on current main to unblock.

PR #5910, opened 2026-06-04
fix!: require recoverable secp256k1eth signatures and clarify validator hashing

gjermundgaraba:
---

#### Why this is needed

- Makes `secp256k1eth` verification match Ethereum recoverable signature semantics: exact 65-byte `[R || S || V]` signatures, canonical `V` in `{0,1}`, low-S enforcement, and recovered signer checked against the validator public key.
- Clarifies validator hashing for EVM/light-client consumers: `ValidatorSet.Hash()` commits the protobuf-encoded validator public key and voting power; validator addresses and proposer priorities are not included.
- Documents the `secp256k1eth` address/signature formats and adds tests so CometBFT hashing and signature rules stay aligned with the implementation.

#### PR checklist

- [x] Tests written/updated


proto/tendermint/types/validator.proto
```diff
@@ -31,6 +31,6 @@ message Validator {
 }
 
 message SimpleValidator {
-  tendermint.crypto.PublicKey pub_key = 1;
+  bytes address = 1;
```
Eric-Warehime: This is pretty breaking, no? I'm unsure if we can actually change the field like this.

gjermundgaraba: yeah, no, you're right. Let me take a stab at this from a different angle.

Eric-Warehime (approved):

crypto/secp256k1eth/key.go
```diff
@@ -167,42 +167,30 @@ func (PubKey) Type() string {
...
 
-	sig := signatureFromBytes(sigStr[:64])
-	// Reject malleable signatures: decred does not enforce low-S but
-	// libsecp256k1 (and thus go-ethereum's Sign output) does. If Serialize()
-	// changes the signature, S was in the upper half -> reject.
-	modified, parseErr := ecdsa.ParseDERSignature(sig.Serialize())
-	if parseErr != nil {
+	var r secp256k1.ModNScalar
+	var s secp256k1.ModNScalar
+	if r.SetByteSlice(sigStr[:32]) || s.SetByteSlice(sigStr[32:64]) {
```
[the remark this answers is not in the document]
gjermundgaraba: validated that this is safe as written

PR #5911, opened 2026-06-04
feat: support libp2p's noise connection protocol for privval/kms

Eric-Warehime:
---

Updates kms to support libp2p Noise connection protocol.

* Adds functionality to comet kms to support either of secret conn or noise
* Updates the cometbft privval to create a noise listener instead of using secret conn based on the passed in remote addr
* Adds a comet utility to print the node ID from the local node key for convenience


Eric-Warehime:
@greptile review

Eric-Warehime:
@greptile review

cmd/cometbft/commands/show_libp2p_id.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/cometbft/cometbft/lp2p"
+	"github.com/cometbft/cometbft/p2p"
+)
+
+// ShowLibp2pIDCmd dumps the node's libp2p peer ID (derived from the node key).
+// This is the identity presented by the libp2p Noise privval transport and the
+// lp2p host.
+var ShowLibp2pIDCmd = &cobra.Command{
```
swift1337: This is already implemented here: https://github.com/cometbft/cometbft/blob/d27fe6a814434d506480380117a35a9ffe6d2045/cmd/cometbft/commands/show_node_id.go#L32

kms/README.md
```diff
@@ -190,6 +192,90 @@ key_file  = "/secrets/osmosis_priv_validator_key.json"
...
+```
+
+This reads `identity.json` from `<home>` and prints the corresponding libp2p
+peer ID.
+
+**Validator peer ID** (give this to the KMS operator so they can pin it in
+`[[validator]].addr`):
+
+```sh
+cometbft show-libp2p-id --home ~/.cometbft
```
swift1337: same here

kms/cmd/cometkms/main.go
```diff
@@ -42,10 +43,40 @@ func main() {
...
+	}
+	id, err := lp2p.IDFromPrivateKey(key)
+	if err != nil {
+		return "", err
+	}
+	return id.String(), nil
+}
+
+func peerIDCmd() *cobra.Command {
+	var home string
```
swift1337: All of these commands use `home`. can we move common variables to `var runtimeConfig = struct{Home string, ...}` and then use `root.PersistentFlags()`?

privval/noise_listener.go
```diff
@@ -0,0 +1,128 @@
...
+func ParseNoiseAddr(addr string) (peer.ID, string, error) {
+	if !strings.HasPrefix(addr, noiseScheme) {
+		return "", "", fmt.Errorf("privval: not a noise address: %q", addr)
+	}
+	rest := strings.TrimPrefix(addr, noiseScheme)
+	at := strings.LastIndex(rest, "@")
+	if at < 0 {
+		return "", "", fmt.Errorf("privval: noise address missing peer id: %q", addr)
+	}
+	pidStr, hostport := rest[:at], rest[at+1:]
```
swift1337, line 122: Should we also check for explicit port presence? Should we validate that the host is ipv4? Is it possible to specify a domain name here?

Eric-Warehime: I think it's probably best to change this such that we just use noise if a multiaddr is supplied and the regular tcp listener with the secret conn on `tcp://`...

I'll change this.

kms/internal/config/config.go
```diff
@@ -50,3 +53,28 @@ func Load(path string) (*Config, error) {
...
 // ReconnectEnabled reports the effective reconnect setting (default true).
 func (v Validator) ReconnectEnabled() bool { return v.Reconnect == nil || *v.Reconnect }
+
+// Transport identifies the privval connection transport selected by a validator
+// address scheme.
+type Transport int
+
+const (
+	// TransportTCP is tcp:// with cometbft SecretConnection (the default).
+	TransportTCP Transport = iota
```
swift1337, line 63: nit: it's better to have `TransportUnknown Transport = iota` so the default int value becomes invalid (eliminates some potential bugs)

swift1337 (approved):

kms/internal/config/config.go
Eric-Warehime: If the transport is not Noise we just default to the existing tcp. In that case it's technically unknown. If it's actually malformed it will probably fail on dial.

aljo242:
heads up: `kms/internal/app/build.go` will conflict with #5912 on merge. this pr has `Build` returning `(*manager.Manager, error)` while #5912 has it returning `(*manager.Manager, func(), error)` with a cleanup function. both are based on main where the kms module doesn't exist yet, so whichever lands second will need a rebase to reconcile the signatures.

PR #5912, opened 2026-06-04
feat: Support pkcs11 backend for comet kms

Eric-Warehime:
---

Support PKCS#11 backend for comet kms.

* Uses miekg/pkcs11 library
* Testing done via SoftHSM2 (skipped if it is not supported since you need the shared library)

I'd like to also validate manually using aws cloud hsm service--todo


Eric-Warehime:
@greptile review

aljo242:
`findObject` silently uses the first match when multiple keys share the same label or id. worth logging a warning or returning an error when both `KeyLabel` and `KeyID` are set and more than one handle comes back, so operators get a signal on misconfiguration rather than a silent wrong-key scenario.

also, the pr body notes aws cloudhsm validation is still todo. fine to merge without it, but worth tracking as a follow-up issue before recommending this backend for production hsm use.

Eric-Warehime:
@greptile review

aljo242 (approved):

aljo242 (approved):
KMS pkcs11 backend support looks good. e2e tests all pass (networks/ci, networks/libp2p-ci, networks_regressions). The `kms/` nested module is properly isolated with its own `go.mod`, CGO_ENABLED=1 build is gated in CI, and the pkcs11 interface wraps `miekg/pkcs11` cleanly. LGTM.

PR #5921, opened 2026-06-12
fix(rpc): escape request host in endpoints listing page

dxbjavid:
**unescaped request host in the rpc endpoints page**

The endpoints listing served from the rpc root builds its links from the request Host, which comes straight off the wire, and writes it into the href attributes and link text with no escaping. A host carrying markup characters then sits in an HTML context verbatim, so anything able to influence that header (a fronting proxy, say) can drop attribute or tag content into the page. Escaping it with html.EscapeString before the links are built keeps it contained, and leaves ordinary host values untouched.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 ### BUG FIXES
 
+- `[rpc]` escape the request `Host` in the endpoints listing page so it cannot
+  break out of the generated HTML
+  ([\#5921](https://github.com/cometbft/cometbft/pull/5921))
 - `[consensus]` Fix `double_sign_check_height = 1` performing no double-sign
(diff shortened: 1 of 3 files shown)


aljo242 (approved):
Correct XSS fix. `r.Host` is attacker-controlled and was landing raw in the generated HTML. `html.EscapeString` applied before both link-building loops. Test covers the \">` breakout vector and verifies entity-escaping. Minimal change, no behaviour regression.

dxbjavid:
cheers for the look. happy to squash the commits down before merge if that's preferred.

dxbjavid:
Thanks for the merge! Appreciated

PR #5929, opened 2026-06-15
test(flowrate): fix flaky TestWriter by tolerating clock quantization in Idle

songgaoye:
## Description
Fix CI failure at:  https://github.com/cometbft/cometbft/actions/runs/27431355897/job/81082257035?pr=5879
`statusesAreEqual` compared `Status.Idle` with exact equality (`==`) while every
other time-derived field (`Duration`, `TimeRem`, the rates) allowed a deviation.
`Idle` is the difference of two timestamps each rounded to the 20ms `clockRate`,
so scheduling jitter on loaded runners can push the two reads onto adjacent
ticks, producing a spurious 20ms idle and failing the strict `==`.

This compares `Idle` with the same `maxDeviationForDuration` tolerance, and makes
`durationsAreEqual` symmetric (absolute diff) so it bounds drift in both
directions.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 ### BUG FIXES
 
+- `[flowrate]` fix flaky `TestWriter` by comparing `Idle` with a duration
+  tolerance instead of exact equality
+  ([\#5929](https://github.com/cometbft/cometbft/pull/5929))
 - `[rpc]` escape the request `Host` in the endpoints listing page so it cannot
   break out of the generated HTML
   ([\#5921](https://github.com/cometbft/cometbft/pull/5921))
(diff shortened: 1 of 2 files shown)


aljo242 (approved):
Correct fix. The  field is computed from two timestamp reads, so it's subject to the same clock-tick rounding as . Using  with the existing tolerance is exactly right. The  fix to  is also correct — the old one-sided check would silently pass when .

aljo242 (approved):
Correct fix. Idle is computed from two timestamp reads so it has the same clock-tick rounding as Duration. Using durationsAreEqual with the existing tolerance is right. The abs fix to durationsAreEqual is also correct — the old one-sided check would pass silently when d2 < d1.

PR #5931, opened 2026-06-16
fix(blocksync): fix deadlock in AddBlock caused by holding pool.mtx during sendError

songgaoye:
## Summary

Fix CI failure at: https://github.com/cometbft/cometbft/actions/runs/27214146296/job/80350996164

- `AddBlock` held `pool.mtx` via a deferred `Unlock` while calling `sendError` on an unbuffered channel. Any concurrent caller that also needed `pool.mtx` would deadlock until the channel was drained.
- Fix by collecting the error in `sendErr` and dispatching it inside the `defer` after the mutex is released.
- Adds a regression test `TestAddBlockDoesNotDeadlockOnSendError`.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@
   `VerifyVoteExtension` handlers are inconsistent halts the node with a clear
   `CONSENSUS FAILURE` instead of stalling the whole network
   ([\#5204](https://github.com/cometbft/cometbft/issues/5204))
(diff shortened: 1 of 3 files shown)


aljo242 (approved):
Correct. The defer pattern releases pool.mtx before sendError blocks on the channel write, which eliminates the deadlock. Regression test is solid — unbuffered errorsCh ensures the old code would hang reliably.

PR #5934, opened 2026-06-17
fix: Fix race condition in privval shutdown

Eric-Warehime:
---

Updates the privval RetrySignerClient to check on shutdown before retrying.  
Without this change it is possible to get into the following shutdown state:
* Node is requesting a signature from a remote signer
* Shutdown happens on remote signer
* Shutdown happens on node

In this case we have to wait `retries × (timeoutAccept + timeout) ≈ 155s` before the node will shut down.

With these changes we will wait a maximum of a single retry timeout before seeing that `Close` has been called (~3s) and shutdown will happen after that.


dhfang (approved):

dhfang:
Oh I'm not stack team, my approval doesn't mean anything.

swift1337 (approved):

PR #5937, opened 2026-06-22
fix(guard): refresh LRU recency and stop cleanup goroutine on Close

songgaoye:
## Overview

Tighten `internal/guard` behavior around LRU access and shutdown.

This PR keeps the app-mempool retry timeout semantics unchanged: duplicate
`Guard` hits do not clear or extend any pending `ForgetAfter` deadline.

Changes:

- Refresh LRU recency on duplicate `Guard` hits.
  `Guard` previously used `Peek`, so observing an already-guarded key did not
  update its LRU position. Since `Guard` is the public duplicate-observation
  path, it now uses `Get` for present keys. This only affects LRU ordering; it
  does not refresh or cancel `expiresAt`.

- Stop the cleanup goroutine deterministically on `Close`.
  `Close` now closes `doneCh` and waits for the background cleanup loop to exit.
  The previous mutex in `Close` did not provide that guarantee.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 ### BUG FIXES
 
+- `[guard]` refresh LRU recency on duplicate hits and stop the background
+  cleanup goroutine gracefully on `Close`
+  ([\#5937](https://github.com/cometbft/cometbft/pull/5937))
 - `[flowrate]` fix flaky `TestWriter` by comparing `Idle` with a duration
   tolerance instead of exact equality
   ([\#5929](https://github.com/cometbft/cometbft/pull/5929))
--- a/internal/guard/guard.go
+++ b/internal/guard/guard.go
@@ -17,6 +17,7 @@ type Guard[K comparable] struct {
 	ttlRemovals map[K]struct{}
 	doneCh      chan struct{}
 	doneOnce    sync.Once
+	wg          sync.WaitGroup
 }
 
 type entry struct {
@@ -53,30 +54,34 @@ func NewWithInterval[K comparable](capacity int, cleanInterval time.Duration) *G
 		ttlRemovals: ttlRemovals,
 	}
 
-	go func() {
+	g.wg.Go(func() {
 		ticker := time.NewTicker(cleanInterval)
 		defer ticker.Stop()
 
 		for {
 			select {
-			case <-ticker.C:
-				g.removeExpired()
 			case <-g.doneCh:
 				return
+			case <-ticker.C:
+				// doneCh wins ties: don't start a pass once closing
+				select {
+				case <-g.doneCh:
+					return
+				default:
+				}
+				g.removeExpired()
 			}
 		}
-	}()
+	})
 
 	return g
 }
 
-// Close closes the guard.
+// Close stops the background cleanup and waits for it to exit.
 func (g *Guard[K]) Close() {
 	g.doneOnce.Do(func() {
-		g.mu.Lock()
-		defer g.mu.Unlock()
-
 		close(g.doneCh)
+		g.wg.Wait()
 	})
 }
 
--- a/internal/guard/guard.go
+++ b/internal/guard/guard.go
@@ -80,14 +80,16 @@ func (g *Guard[K]) Close() {
 	})
 }
 
-// Guard guards the item against duplicates.
-// Doesn't have a TTL besides LRU eviction.
+// Guard guards the item against duplicates. A newly added key has no TTL
+// (only LRU eviction), but an existing key may carry a pending ForgetAfter
+// deadline; Guard must not clear it on a re-guard, or pending retries break.
 // Returns true if the key was added, false if it was already present.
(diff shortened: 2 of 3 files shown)


swift1337 (changes requested):
What is the original intention of this PR? 

1) "removeExpired drops stale ttlRemovals entries — prevents a slow leak if the map ever drifts from the LRU."

This is done automatically within LRU logic, see the constructor
```go
lru, _ := simplelru.NewLRU(
		capacity,
		func(key K, _ *entry) { delete(ttlRemovals, key) }, // onEvict()
	)
```

2) "Refresh LRU recency on duplicate Guard hits" - Keeping the old TTL is intentional. We don't want to refresh the TTL because that would indefinitely prolong the "guard" if a user tries to resubmit something from broadcastSync, for example. We want to allow duplicate requests after that timeout, not punish users for retrying too early.

songgaoye:
> What is the original intention of this PR?
> 
> 1. "removeExpired drops stale ttlRemovals entries — prevents a slow leak if the map ever drifts from the LRU."
> 
> This is done automatically within LRU logic, see the constructor
> 
> ```go
> lru, _ := simplelru.NewLRU(
> 		capacity,
> 		func(key K, _ *entry) { delete(ttlRemovals, key) }, // onEvict()
> 	)
> ```
> 
> 2. "Refresh LRU recency on duplicate Guard hits" - Keeping the old TTL is intentional. We don't want to refresh the TTL because that would indefinitely prolong the "guard" if a user tries to resubmit something from broadcastSync, for example. We want to allow duplicate requests after that timeout, not punish users for retrying too early.

Thanks, that makes sense.

For `removeExpired`, agreed. The `ttlRemovals` entry should normally be removed by the LRU `onEvict` callback. I’ll drop that part from the PR.

For the duplicate `Guard` hit behavior, I also agree that the `ForgetAfter` TTL must not be refreshed or canceled. That was not the intended behavior. I updated the change so duplicate hits only refresh the LRU recency via `Get`; they leave the existing `expiresAt` untouched. So an early `broadcastSync` retry still gets rejected, but the tx becomes retryable after the original timeout as before.

If you think duplicate hits should not affect LRU ordering either, I can drop that part too and keep the PR scoped to the `Close` goroutine shutdown fix.

swift1337:
> Stop the cleanup goroutine deterministically on Close

@songgaoye, right now `guard.Close()` is only called in test cases and doesn't affect the runtime code path

> I updated the change so duplicate hits only refresh the LRU recency via Get

This changes the recency of the cache key and actually mutates LRU

---

Could you elaborate on the problems your team faced with the app mempool? Do you see any areas of improvement that we need to reflect in gh issues? Happy to help.

songgaoye:
> > Stop the cleanup goroutine deterministically on Close
> 
> @songgaoye, right now `guard.Close()` is only called in test cases and doesn't affect the runtime code path
> 
> > I updated the change so duplicate hits only refresh the LRU recency via Get
> 
> This changes the recency of the cache key and actually mutates LRU
> 
> Could you elaborate on the problems your team faced with the app mempool? Do you see any areas of improvement that we need to reflect in gh issues? Happy to help.

@swift1337  Thanks, this makes sense.

For the LRU part, I updated duplicate hits to use `Get`, so repeated observations refresh the key's recency and intentionally mutate the LRU state.

For `guard.Close()`, you are right that this PR only makes the guard cleanup goroutine stoppable, but by itself it does not affect the production shutdown path unless the app mempool calls it. I opened #5942 to wire that into the actual app mempool lifecycle. I can either keep that as a follow-up PR, or fold it into this PR if you prefer the deterministic cleanup to be fully covered here.

On the app mempool issues: the main pain points we have seen are around observability and behavior under load. It can be hard to tell why txs are rejected/delayed/retried, and the contract between CometBFT's app mempool proxying and the application's own admission/retry logic is subtle. I think useful follow-up issues could be better metrics/logging around app mempool rejection/retry paths, clearer docs for expected app mempool semantics, and perhaps integration tests covering duplicate/retry/eviction behavior.

Happy to help turn those into concrete issues.

swift1337:
>  It can be hard to tell why txs are rejected/delayed/retried

This feedback is helpful-thank you!

We made Comet's app-mempool concise and trimmed its features so that an arbitrary app-side mempool can expose any functionality, metrics, etc. The best way to handle this is to create a concrete GitHub issue where we can move this discussion and agree on what should be improved on the app-mempool side and what needs clarification in the docs.

Would you like to create the issue?

songgaoye:
> > It can be hard to tell why txs are rejected/delayed/retried
> 
> This feedback is helpful-thank you!
> 
> We made Comet's app-mempool concise and trimmed its features so that an arbitrary app-side mempool can expose any functionality, metrics, etc. The best way to handle this is to create a concrete GitHub issue where we can move this discussion and agree on what should be improved on the app-mempool side and what needs clarification in the docs.
> 
> Would you like to create the issue?

sure, I will do it. Thank you!

PR #5942, opened 2026-06-22
fix(mempool): AppMempool ABCI response handling, callback/cleanup, and lifecycle

songgaoye:
## Summary

`AppMempool` proxies txs to the app over ABCI (`InsertTx`, `CheckTx`, `ReapTxs`) and tracks seen txs in a TTL guard cache. This PR fixes robustness bugs in that path.

### Changes

- **Nil ABCI responses** (no more panics, correct semantics):
  - `ReapTxs`: nil response skipped instead of dereferencing `res.Txs`.
  - `InsertTx`: nil response with no error is an invariant violation — now panics.
  - `CheckTx`:
    - On ABCI client error: `forgetTx` and callback were both skipped — tx stuck in cache, RPC caller hung; now fires the callback with a synthetic error response. Tx stays in seen cache (no retry) to avoid DDoS vector.
    - On nil response (no error): `callback(nil)` was called causing a downstream panic; now fires the callback with a synthetic error response. Tx stays in seen cache (no retry) to avoid DDoS vector.
- **CheckTx callback/cleanup**: move `forgetTx` before the callback so a panicking callback can't skip cleanup.
- **InsertTx error path**: transport errors no longer clear the seen cache to avoid DDoS vector.
- **Metrics**: retryable codes no longer inflate `RejectedTxs`.
- **Lifecycle**: add `AppMempool.Close()` to stop the guard's background goroutine and call it from `AppReactor.OnStop()` (was leaking on shutdown).

---

#### PR checklist

- [x] Tests written/updated


swift1337 (changes requested):
What problem does this PR solve?

songgaoye:
> What problem does this PR solve?

Sorry, I'd opened this without a description — I've just updated it with the full context. In short: it fixes several robustness bugs in the app-side mempool's ABCI response handling — panics on `nil InsertTx/CheckTx/ReapTxs `responses, a nil InsertTx response being silently treated as success, RPC callbacks/seen-cache cleanup being skipped on app errors and callback panics, retryable codes inflating `RejectedTxs`, and a guard goroutine leaking on shutdown. Details in the updated PR description above.

mempool/app_mempool.go
```diff
@@ -262,16 +277,27 @@ func (m *AppMempool) CheckTx(tx types.Tx, callback func(res *abci.ResponseCheckT
...
+			if callback != nil {
+				callback(&abci.ResponseCheckTx{Code: 1, Log: "nil CheckTx response"})
+			}
 			return
 		}
 
 		// app mempool doesn't execute the tx, so we ALWAYS return an empty response here.
 		// This will most likely break many clients. Clients should rely on app-specific
 		// broadcasting endpoints (think of eth_sendRawTransaction, etc...).
-		if callback != nil {
```
thomas-nguy, line 272: comment needs to be fixed

songgaoye: fixed!

mempool/app_mempool.go
```diff
@@ -103,11 +103,12 @@ func (m *AppMempool) InsertTx(tx types.Tx) error {
 	switch {
 	case err != nil:
 		m.metrics.FailedTxs.Add(1)
+		m.forgetTx(tx, true)
```
swift1337: This creates a DDoS vector

mempool/app_mempool.go
```diff
@@ -160,6 +161,12 @@ func (m *AppMempool) insertTx(tx types.Tx) (uint32, error) {
 		return 0, err
 	}
 
+	if resp == nil {
```
swift1337, line 163: This should be treated as invariant breaking. What's the use case of having NO error and `nil` response?

mempool/app_mempool.go
```diff
@@ -189,13 +196,15 @@ func (m *AppMempool) reapTxs(ctx context.Context, channel chan<- types.Txs) {
 		MaxGas:   m.config.ReapMaxGas,
 	}
 
+	ticker := time.NewTicker(m.config.ReapInterval)
```
swift1337, line 196: Can you explain this change?

mempool/app_mempool.go
```diff
@@ -261,16 +276,23 @@ func (m *AppMempool) CheckTx(tx types.Tx, callback func(res *abci.ResponseCheckT
...
 
-		// app mempool doesn't execute the tx, so we ALWAYS return an empty response here.
-		// This will most likely break many clients. Clients should rely on app-specific
-		// broadcasting endpoints (think of eth_sendRawTransaction, etc...).
-		if callback != nil {
-			callback(res)
+		if res == nil {
+			// nil response, no error: forget the tx so it can be retried.
+			m.logger.Error("AppMempool.CheckTx: nil response", "tx", txHash(tx))
+			m.forgetTx(tx, true)
```
swift1337: Same comment regarding the DDoS

swift1337 (changes requested):

mempool/app_mempool.go
songgaoye: `time.NewTicker` reuses a single timer object across iterations; `time.After` allocates a new one per loop cycle.

mempool/app_mempool.go
songgaoye: fixed!

mempool/app_mempool.go
songgaoye: fixed!

mempool/app_mempool.go
songgaoye: fixed!

PR #5946, opened 2026-06-30
fix(mempool): Implement `MsgBytesFilter` in Reactor to prevent heap amplification

thomas-nguy:
Implements the optional p2p.MsgBytesFilter interface (defined in p2p/base_reactor.go) for this reactor.

When provided, this filter runs against the raw message bytes before unmarshalling, allowing oversized or malformed messages to be rejected up front. This closes a heap-amplification vector where a small, count-packed wire payload could be unmarshalled into a disproportionately large in-memory structure, exhausting memory.

Motivation:

Previously the message bytes went straight to unmarshal with no pre-decode size/validity check. A crafted gossip message could exploit this to trigger excessive heap allocation. Rejecting bad messages before unmarshal eliminates the attack surface.

---

#### PR checklist

- [ ] Tests written/updated

### DIFF
--- a/internal/protowire/protowire.go
+++ b/internal/protowire/protowire.go
@@ -1,0 +1,127 @@
+// Package protowire provides a minimal, allocation-free reader for walking
+// raw protobuf wire bytes. Every read is bounds-checked and advances a cursor.
+package protowire
+
+import (
+	"errors"
+	"fmt"
+)
+
+// Wire types (the low 3 bits of a field tag).
+const (
+	WireVarint  = 0
+	WireFixed64 = 1
+	WireBytes   = 2
+	WireFixed32 = 5
+)
+
+var (
+	// ErrVarintOverflow is returned when a varint does not terminate within
+	// the 10 bytes that can hold a 64-bit value.
+	ErrVarintOverflow = errors.New("protowire: varint overflow")
+	// ErrTruncatedVarint is returned when the buffer ends mid-varint.
+	ErrTruncatedVarint = errors.New("protowire: truncated varint")
+	// ErrOutOfBounds is returned when a length prefix or fixed-width field
+	// would run past the end of the buffer.
+	ErrOutOfBounds = errors.New("protowire: length out of bounds")
+	// ErrIllegalFieldNumber is returned when a tag decodes to a field number
+	// that is not a legal protobuf field number (>= 1).
+	ErrIllegalFieldNumber = errors.New("protowire: illegal field number")
+	// ErrUnsupportedWireType is returned when a tag carries a wire type the
+	// cursor does not know how to skip.
+	ErrUnsupportedWireType = errors.New("protowire: unsupported wire type")
+)
+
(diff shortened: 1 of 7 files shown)


mempool/filter.go
```diff
@@ -0,0 +1,238 @@
+package mempool
```
swift1337, line 1: Let's generalize most parts of this code into `internal/protowire` package with `WireCursor` and exported errors

mempool/filter.go
```diff
@@ -0,0 +1,238 @@
...
+// gossipBatchByteBudget is the max raw tx bytes a peer may pack into one gossip
+// message. A single tx can be MaxTxBytes even when MaxBatchBytes is smaller, so
+// the budget is the larger of the two. A non-positive result means unbounded.
+func gossipBatchByteBudget(cfg *config.MempoolConfig) int {
+	return max(cfg.MaxTxBytes, cfg.MaxBatchBytes)
+}
+
+// txLimits holds the size ceilings applied to a gossip batch. A non-positive
+// value disables that limit.
+type txLimits struct {
```
swift1337: This struct looks redundant

mempool/app_reactor.go
```diff
@@ -134,6 +136,14 @@ func (r *AppReactor) EnableInOutTxs() {
 	close(r.waitForSwitchingOnCh)
 }
 
+// FilterMsgBytes implements p2p.MsgBytesFilter.
+func (r *AppReactor) FilterMsgBytes(chID byte, _ p2p.Peer, msgBytes []byte) error {
+	if chID != MempoolChannel || len(msgBytes) == 0 {
+		return nil
+	}
+	return filterMempoolMsgBytes(msgBytes, r.config.MaxTxBytes, gossipBatchByteBudget(r.config))
```
swift1337: `max()` can be moved into `filterMempoolMsgBytes()`, making `gossipBatchByteBudget()` redundant
```suggestion
	return filterMempoolMsgBytes(msgBytes, r.config.MaxTxBytes, r.config.MaxBatchBytes)
```

swift1337 (changes requested):
Thank you for contributing! This is indeed a useful improvement

Left some comments

thomas-nguy:
Thanks for your review and comments @swift1337

internal/protowire/protowire.go
```diff
@@ -0,0 +1,127 @@
...
+	WireVarint  = 0
+	WireFixed64 = 1
+	WireBytes   = 2
+	WireFixed32 = 5
+)
+
+var (
+	// ErrVarintOverflow is returned when a varint does not terminate within
+	// the 10 bytes that can hold a 64-bit value.
+	ErrVarintOverflow = errors.New("protowire: varint overflow")
```
swift1337: nit: remove `protowire:` prefix from errors

swift1337 (approved):

internal/protowire/protowire.go
swift1337: Done — removed the `protowire:` prefix from all error strings in 5e8128b65.

PR #5948, opened 2026-07-01
fix(mempool): truncate proto field number to int32 in filter's ReadTag

thomas-nguy:
Fixing a small bug in PR https://github.com/cometbft/cometbft/pull/5946

ReadTag decoded protobuf field numbers as a platform-width int `int(v>>3)`, but the generated gogoproto unmarshaller uses `int32(wire>>3)`. On 64-bit platforms this let a crafted tag whose high bits differ (e.g. 8a 80 80 80 80 01, field 2^32+1) be classified as a different field by the pre-scan than by proto.Unmarshal

As a result, the filter skips the entry while Unmarshal still decoded it as the real repeated`txs` field, slipping empty entries past the empty-entry rejection as part of the filter rule

As an evidence
https://github.com/cometbft/cometbft/blob/main/proto/tendermint/mempool/types.pb.go#L327

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,8 @@
   ([\#5879](https://github.com/cometbft/cometbft/pull/5879))
 - `[consensus]` release cs.mtx before sending to statsMsgQueue
   ([\#5813](https://github.com/cometbft/cometbft/pull/5813))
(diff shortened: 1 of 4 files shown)


dianab-cl:
@greptile review

swift1337 (approved):

PR #5950, opened 2026-07-02
fix(blocksync): tolerate redundant near-tip BlockResponse from a second peer

thomas-nguy:
Commit `4590e7ac`  introduced a pre-unmarshal MsgBytesFilter on the blocksync reactor. Before unmarshalling a BlockResponse, the filter calls BlockPool.HasPendingRequestFrom(src) to confirm we actually have an outstanding request to the sending peer, rejecting (and disconnecting) any peer that sends an unsolicited response.

Problem

The pool requests each height from two peers (peerID + secondPeerID) for redundancy, but PopRequest deletes the requester the instant that height commits. So when the faster peer's block commits first and the second peer's (solicited, expected) response arrives just after, no live requester remains, the filter can't distinguish "already committed" from "never requested" and rejects the response. On the comet p2p transport that rejection panics the peer's receive routine (p2p/peer.go), which _recover catches and tears down the connection (libp2p instead calls StopPeerForError) 

The fix 

`HasPendingRequestFrom` now also accepts any response whose height falls within the current sync window [startHeight, height], in addition to peers with an explicit outstanding request. The pre-unmarshal filter extracts the block height from the wire stub and passes it in.

Evidence
https://github.com/cometbft/cometbft/blob/main/blocksync/pool.go#L383

#### PR checklist

- [x] Tests written/updated


blocksync/pool.go
```diff
@@ -215,11 +215,15 @@ func (pool *BlockPool) GetStatus() (height int64, numPending int32, lenRequester
...
-// HasPendingRequestFrom reports whether we have at least one outstanding block
-// request directed at the given peer.
-func (pool *BlockPool) HasPendingRequestFrom(peerID p2p.ID) bool {
+// HasPendingRequestFrom reports whether a BlockResponse for the given height
+// from the given peer should be accepted. It returns true for an outstanding request
+// to the peer or for any height in the current sync window [startHeight, height].
+func (pool *BlockPool) HasPendingRequestFrom(peerID p2p.ID, height int64) bool {
 	pool.mtx.Lock()
 	defer pool.mtx.Unlock()
+	if height >= pool.startHeight && height <= pool.height {
```
swift1337, line 224: This change introduces the assumption that if `height` is within the boundary, then the node definitely expects a block from `peerID`. How can we ensure that this invariant holds? At first glance, I'm not 100% confident.

mattac21: I agree with @swift1337 here, this change was introduced to stop blocksync from blindly accepting responses from peers that looked somewhat real (like these that fall within the expected height range), but are spam or malicious. I do agree this is a bug that we should be fixing however but we should take a different approach.

Maybe instead of removing both the primary and secondary peers assigned to a request once we receive the primaries response, we keep the entry around for some duration, giving the secondary time to respond, then only after that duration do we remove from the requesters set?

mattac21 (changes requested):

PR #5832, opened 2026-05-05
fix(mempool): allow CheckTx retries (backport #5802)

swift1337:
Backport #5802

Closes STACK-2402

PR #5952, opened 2026-07-03
fix(mempool): fire callback and release seen-guard on app.CheckTx error

JayT106:
`AppMempool.CheckTx` dispatches the ABCI call in a goroutine and returns `nil` immediately. The goroutine is expected to invoke the caller-supplied callback on completion — this is how `BroadcastTxSync` and `BroadcastTxCommit` receive the result.

## Problem

On any transport or app error (`err != nil` from `app.CheckTx`), the goroutine logged and returned without:
1. Invoking the callback — `BroadcastTxSync`/`BroadcastTxCommit` block on `resCh` forever, returning only when the client context expires ("broadcast confirmation not received").
2. Calling `forgetTx` — the tx remains in the seen-guard (added by `guardTx` at entry), so identical resubmissions hit `ErrSeenTx` until LRU eviction or restart.

The `#5802`/`#5832` retry fix only covers the `res.Code != OK` path, not the `err != nil` early return.

## Fix

On `err != nil`, call `forgetTx(tx, true)` (retryable delay) and fire the callback with `CodeTypeRetry` so RPC callers return promptly and the tx can be resubmitted.

---

#### PR checklist

- [x] Tests written/updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,9 +42,9 @@
   ([\#5867](https://github.com/cometbft/cometbft/pull/5867))
 - `[blocksync]` fix removeTimedoutPeers deadlock found via Byzantine prevote gossip race
   ([\#5839](https://github.com/cometbft/cometbft/pull/5839))
-- `[mempool]` fix `AppMempool.CheckTx` not invoking callback or releasing
-  seen-guard on app error, causing `BroadcastTxSync`/`BroadcastTxCommit` to
-  hang until context expiry and blocking tx resubmission.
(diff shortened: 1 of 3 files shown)


swift1337:
CheckTx is implemented in AppMempool only as a fallback. The intention is to use an app-specific RPC (e.g., eth_rpc for Cosmos EVM). Otherwise, use CListMempool.

JayT106:
> Member

I think the issue when standard Cosmos SDK txs submitted via
  CometBFT RPC still go through CheckTx, not the evm TX

PR #5953, opened 2026-07-03
docs(blocksync): document adaptive_sync equivocation risk for validators

JayT106:
With `adaptive_sync=true`, consensus starts immediately (`consensusWaitForSync=false`) and `signAddVote` fires whenever the node is in the validator set. A catching-up validator can sign a vote for block B at height H while the blocksync ingestor concurrently commits a different block B' at H (already decided by the network). The HRS file and KMS are not sufficient backstops — the conflicting messages are at different (H,R,step) tuples so `CheckHRS` never fires.

This is a known tradeoff: validators on high-performance chains (sub-1000ms blocks, high TPS) benefit from adaptive sync to avoid falling behind and being jailed. The risk is deliberate and operator-owned.

## Fix

Documentation only — no behavior change:
- `docs/core/block-sync.md`: add explicit equivocation risk warning to the Adaptive Sync section
- `config/toml.go`: add the same warning to the `adaptive_sync` config comment

---

#### PR checklist

- [ ] Tests written/updated


swift1337 (changes requested):
We tested Comet with high-performing cosmos-evm chains (both comet-p2p and lib-p2p) during load spikes; even validator nodes can sometimes fall behind consensus. 

Adaptive sync helps them "catch up" with the whole network instead of lagging behind blocks and being jailed. I'm talking about block times under 1000 ms and over 1500 TPS.

If your use case shows a risk, you can disable adaptive sync on your validators; I don't feel this PR addresses your concern holistically.

cc @mattac21 @Eric-Warehime

JayT106:
> We tested Comet with high-performing cosmos-evm chains (both comet-p2p and lib-p2p) during load spikes; even validator nodes can sometimes fall behind consensus.
> 
> Adaptive sync helps them "catch up" with the whole network instead of lagging behind blocks and being jailed. I'm talking about block times under 1000 ms and over 1500 TPS.
> 
> If your use case shows a risk, you can disable adaptive sync on your validators; I don't feel this PR addresses your concern holistically.
> 
> cc @mattac21 @Eric-Warehime

Update doc instead of changes.

swift1337 (approved):

PR #5955, opened 2026-07-03
fix(lp2p): cap per-reactor priority queue to prevent memory-exhaustion DoS

JayT106:
### Problem

The lp2p `handleStream` path closes the stream before enqueuing the decoded message. libp2p connection/stream limits therefore provide no backpressure: if a reactor's `Receive` is slow (e.g. mempool under load), the priority queue grows without bound until OOM.

### Fix

- Cap each reactor's priority queue via a new `max_queue_size` config field (default 200,000 — large enough for normal traffic, bounded enough to prevent OOM).
- Once full, evict the oldest lower-priority item instead of rejecting the new one, so a burst of low-priority traffic can't starve out higher-priority messages.
- Configurable globally via `p2p.libp2p.scaler.max_queue_size` and per-reactor via `overrides[].max_queue_size`. `0` = unlimited.
- Preserve an unset override's `max_queue_size` across a config write/reload round-trip, so it stays "inherit the global cap" instead of collapsing to an explicit `0` (unbounded).
- Count evicted messages in the `p2p_messages_reactor_dropped` metric, not just outright rejections, and log drops at `Error` level.

---

#### PR checklist

- [x] Tests written/updated


lp2p/reactor_set.go
```diff
@@ -300,7 +308,7 @@ func (rs *reactorSet) newReactorPriorityQueue(
 		autopool.WithOnShrink[pendingEnvelope](func() {
 			concurrencyCounter.Add(-1)
 		}),
-		autopool.WithPriorityQueue[pendingEnvelope](autopool.NewPriorityQueue(priorities)),
+		autopool.WithPriorityQueue[pendingEnvelope](autopool.NewPriorityQueueWithMax(priorities, concurrentPoolCapacity)),
```
swift1337: This change doesn't address the problem and breaks correctness.

Let's say node A and node B each send the same 1000 valid mempool transactions to node C, one per p2p message (a valid comet case). With this change, most of the transactions will be rejected before they are even deduplicated by the mempool guard/cache logic.

And on top of that, suppose we receive a valid p2p vote from node D. Because the time window between receiving the messages and the vote is short, the autoscaler won't have time to scale the concurrency of consumers, causing the vote to be lost.

Currently, the LRU is indeed unbound, but most of the throughput will be rejected by the lp2p limiters. We could consider a cap on this queue, but definitely not 512. 

e.g. it's acceptable to grow RAM usage by 200 MB, but limiting it to 512 messages (<1 MB RAM) w/ cons that I listed ^^ is impractical

swift1337 (changes requested):

dhfang:
@greptile review

lp2p/reactor_set.go
```diff
@@ -283,6 +290,20 @@ func (rs *reactorSet) newReactorPriorityQueue(
...
 
+	// priority queue max: use reactor-specific override if set, else global config.
+	// 0 means unbounded — only acceptable in tests.
+	scalerCfg := rs.switchRef.host.config.Scaler
+	maxQueueSize := scalerCfg.MaxQueueSize
+	for _, override := range scalerCfg.Overrides {
+		if strings.EqualFold(override.Reactor, reactorName) && override.MaxQueueSize > 0 {
+			maxQueueSize = override.MaxQueueSize
+			break
+		}
```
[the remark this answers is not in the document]
JayT106: fixed in [fix(lp2p): apply scaler override unconditionally, fix log severity](https://github.com/cometbft/cometbft/pull/5955/commits/3379b9c3e8c3e215c4f90ab56a30b51e2919d3e3)

lp2p/reactor_set.go
```diff
@@ -283,6 +290,20 @@ func (rs *reactorSet) newReactorPriorityQueue(
...
+	}
+	if maxQueueSize == 0 {
+		rs.switchRef.Logger.Error("Reactor priority queue is unbounded; node is vulnerable to memory-exhaustion DoS", "reactor", reactorName)
+	}
```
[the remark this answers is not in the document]
JayT106: fixed in [fix(lp2p): apply scaler override unconditionally, fix log severity](https://github.com/cometbft/cometbft/pull/5955/commits/3379b9c3e8c3e215c4f90ab56a30b51e2919d3e3)

dhfang:
@greptile re-review

config/toml.go
```diff
@@ -366,12 +366,17 @@ max_workers = {{ .P2P.LibP2PConfig.Scaler.MaxWorkers }}
 # scale up when observed latency is below this value, scale down when above it.
 threshold_latency = "{{ .P2P.LibP2PConfig.Scaler.ThresholdLatency }}"
 
+# Maximum number of messages buffered per-reactor before drops begin (default: 200000).
+# 0 means unlimited (risks OOM under a sustained flood from a fast peer).
```
swift1337:
```suggestion
# zero means unlimited 
```

lp2p/reactor_set.go
```diff
@@ -283,6 +290,20 @@ func (rs *reactorSet) newReactorPriorityQueue(
 		concurrentPoolCapacity = 512
 	)
 
+	// priority queue max: use reactor-specific override if set, else global config.
+	// 0 means unbounded — only acceptable in tests.
```
swift1337:
```suggestion
```

lp2p/reactor_set.go
```diff
@@ -283,6 +290,20 @@ func (rs *reactorSet) newReactorPriorityQueue(
...
+	maxQueueSize := scalerCfg.MaxQueueSize
+	for _, override := range scalerCfg.Overrides {
+		if strings.EqualFold(override.Reactor, reactorName) {
+			maxQueueSize = override.MaxQueueSize
+			break
+		}
+	}
+	if maxQueueSize == 0 {
+		rs.switchRef.Logger.Info("Reactor priority queue is unbounded; node is vulnerable to memory-exhaustion DoS", "reactor", reactorName)
+	}
```
swift1337: Please review AI code before pushing.
```suggestion

```

lp2p/reactor_set_test.go
```diff
@@ -174,6 +174,43 @@ func TestReactorSet(t *testing.T) {
 		assert.Len(t, reactorB.receivedEnvelopes(), 0)
 	})
 
+	t.Run("receive_drops_when_queue_full", func(t *testing.T) {
```
swift1337: rename to camelCase

swift1337 (changes requested):

config/toml.go
JayT106: Fixed!

lp2p/reactor_set.go
JayT106: Removed!

lp2p/reactor_set.go
JayT106: Removed.

lp2p/reactor_set_test.go
JayT106: Updated.

internal/autopool/queue.go
```diff
@@ -94,9 +106,14 @@ func (q *PriorityQueue) Push(value any, priority int) error {
 	q.mu.Lock()
 	defer q.mu.Unlock()
 
+	if q.maxSize > 0 && q.size >= q.maxSize {
```
songgaoye: a global count across every priority level — and it rejects the incoming message regardless of its priority

JayT106: added evict lower-priority logic.

dhfang:
@greptile re-review

dhfang (approved):

PR #5956, opened 2026-07-03
fix(mempool): include proto framing overhead in AppReactor batch size to prevent peer teardown

JayT106:
### Problem

`AppReactor.chunkTxs` sized batches by raw `len(tx)` bytes, ignoring the 2–4 bytes of protobuf wire overhead each entry adds (1-byte field tag + varint length prefix). A batch of small txs near `MaxTxBytes` serializes to `MaxTxBytes + N*(2–4)` bytes, exceeding the receiver's `RecvMessageCapacity`. The connection layer returns "received message exceeds available capacity" and tears down the peer.

Example overflows (default `MaxTxBytes = 1 MB`):
- 1 000-byte txs: ~+2 564 bytes over capacity
- 1-byte txs: ~+2 097 149 bytes over capacity

Second issue: `GetChannels` sized `RecvMessageCapacity` for a single `MaxTxBytes`-sized tx while `OnStart` used `MaxBatchBytes` when set, so configs with `MaxBatchBytes > MaxTxBytes` would overflow at the capacity check itself.

### Fix

- Add `internal/protowire.RepeatedBytesEntrySize(dataLen int) int` — exact wire bytes proto adds per `repeated bytes` entry (1-byte tag + varint(len)), verified against `proto.Size` at all varint boundary values; panics on negative input.
- `chunkTxs` accounts for this overhead per tx so no encoded batch exceeds `RecvMessageCapacity`.
- Extract `effectiveMaxBatchBytes()` to deduplicate the `MaxBatchBytes`/`MaxTxBytes` fallback logic shared by `OnStart` and `GetChannels`.
- `GetChannels` derives `RecvMessageCapacity` from `effectiveMaxBatchBytes()`, keeping the advertised capacity in sync with the actual send ceiling.

---

#### PR checklist

- [x] Tests written/updated


mempool/app_reactor.go
```diff
@@ -263,7 +273,6 @@ func chunkTxs(txs types.Txs, maxBatchSizeBytes int) []types.Txs {
 				chunks = append(chunks, lastChunk)
 			}
 
-			// reset chunk size
```
swift1337, line 266: let's keep code that is unrelated to this PR as is

mempool/app_reactor.go
```diff
@@ -247,13 +256,14 @@ func chunkTxs(txs types.Txs, maxBatchSizeBytes int) []types.Txs {
 		return nil
 	}
 
-	chunks := []types.Txs{}
+	var chunks []types.Txs
```
swift1337: let's keep code that is unrelated to this PR as is

mempool/app_reactor.go
```diff
@@ -272,14 +281,25 @@ func chunkTxs(txs types.Txs, maxBatchSizeBytes int) []types.Txs {
...
 	if len(lastChunk) > 0 {
 		chunks = append(chunks, lastChunk)
 	}
 
 	return chunks
 }
 
+// protoRepeatedBytesOverhead returns the wire bytes proto adds per entry in a
+// repeated bytes field: 1 byte field tag + varint(dataLen).
+func protoRepeatedBytesOverhead(dataLen int) int {
```
swift1337: Should we move that to recently added `internal/protowire` ?

mempool/app_reactor_test.go
```diff
@@ -207,6 +212,88 @@ func TestChunkTxs(t *testing.T) {
...
+					effectiveSize += len(tx) + protoRepeatedBytesOverhead(len(tx))
+				}
+				require.LessOrEqual(t, effectiveSize, tt.size,
+					"chunk #%d effective size %d exceeds limit %d", i, effectiveSize, tt.size)
+			}
+		})
+	}
+}
+
+func TestEffectiveMaxBatchBytes(t *testing.T) {
```
swift1337: Move to TestAppReactor as `t.Run("effectiveMaxBatchBytes")`

mempool/app_reactor_test.go
```diff
@@ -207,6 +212,88 @@ func TestChunkTxs(t *testing.T) {
...
+			}}
+			require.Equal(t, tt.want, r.effectiveMaxBatchBytes())
+		})
+	}
+}
+
+// TestGetChannelsCapacityCoversMaxBatch is a regression test for the peer-teardown bug:
+// every batch produced by chunkTxs must serialize to a Message that fits within
+// the RecvMessageCapacity advertised by GetChannels.
+func TestGetChannelsCapacityCoversMaxBatch(t *testing.T) {
```
swift1337: Move to TestAppReactors as `t.Run("GetChannels")`

swift1337 (approved):

PR #5957, opened 2026-07-03
fix(lp2p): base custom resource manager on AutoScale; fix System.Conns arithmetic

JayT106:
### Problem

The custom branch of `ResourceManagerFromConfig` built its limits on top of `InfiniteLimits` instead of the AutoScale defaults. That left `System.Memory`, `System.Streams`, the whole `Transient` scope, and even `PeerDefault.Memory`/`FD` at `MaxInt64` — only `PeerDefault.Conns`/`Streams` and the `Service`/`Protocol` maps got explicit values. Custom mode ended up strictly weaker than default mode, which defeats the point of picking it to harden a node. In practice this meant a single peer already accepted under `max_peers` could grow its QUIC receive-window memory without any limit, and even with per-peer memory capped, enough peers together could still exceed available RAM since there was no aggregate ceiling either.

On top of that, `System.Conns` was computed as `PeerDefault.Conns * maxPeerStreams` — connections multiplied by streams, a straight unit error (the code comment even said "per peer × max peers," but the variable used was `maxPeerStreams`). With `max_peers=10` and `max_peer_streams=50` that produced 400 instead of the intended 80. And because the final config was built with `InfiniteLimits` as the fallback, any field left unset anywhere in the chain silently resolved to `MaxInt64` rather than a sane default.

### Fix

Custom mode now starts from `scaled.ToPartialLimitConfig()` (AutoScale) instead of `InfiniteLimits`, so `System.Memory/Streams/FD`, `Transient`, and `PeerDefault.Memory/FD` are all finite by default. The `System.Conns` formula is fixed to `maxConnsPerPeer * maxPeers` (with `ConnsInbound`/`ConnsOutbound` capped the same way) — `maxConnsPerPeer` is a small constant (4) rather than AutoScale's `PeerDefault.Conns` (~8), since QUIC peers hold about one connection each and 4 covers the active connection plus a reconnect in flight (per [@swift1337's suggestion](https://github.com/cometbft/cometbft/pull/5957#issuecomment-4936678132)). The config is built with `scaled` as the fallback instead of `InfiniteLimits`. The old explicit copies of `Service`, `Protocol`, and `PeerDefault.Conns*` are gone too — they come for free from the AutoScale base now.

**Before vs after (MaxPeers=10, MaxPeerStreams=50, 69 GB RAM host):**

| Field | Before (bug) | After (fix) |
|---|---|---|
| \`System.Memory\` | MaxInt64 | ~8.7 GB (totalRAM/8 via AutoScale) |
| \`System.Streams\` | MaxInt64 | 18432 (AutoScale) |
| \`System.Conns\` | 400 (8×50) | 40 (4×10) |
| \`Transient.Streams\` | MaxInt64 | 2304 (AutoScale) |
| \`Transient.Memory\` | MaxInt64 | ~1.1 GB (AutoScale) |
| \`PeerDefault.Memory\` | MaxInt64 | ~1.1 GB (AutoScale) |

---

#### PR checklist

- [x] Tests written/updated


swift1337:
> the inversion defeats the purpose of choosing custom mode to harden a node. A malicious peer could drive the node to OOM since no aggregate memory/stream backstop existed.

Custom resource manager's limits are tied to Comet's network model (channels, streams, and peers). They effectively limit all resources. Can you provide a concrete example of an "Uncapped system scope"?

Regarding the "System.Conns unit error", I agree we need to think about how to limit this further. The assumption for the current values was that `system.max_conns` in the worst case equals `max_peers` * `max_peer_streams`, assuming each stream has its own connection. This is impractical, and in most cases two peers would only have a single connection with QUIC's built-in multiplexer.

JayT106:
Correction on the Transient scope point — I dug into the QUIC code path and that mechanism doesn't actually hold. In QUIC, the handshake finishes before `Accept()` even returns, and `SetPeer` fires right after the rcmgr scope opens, before `ConnGater` runs. So the Transient window is just a few microseconds, not something that spans the handshake like I said.

The real gap is simpler: in the old code, `PeerDefault.Memory` and `PeerDefault.FD` were never copied over from `AutoScale()` — only `Conns` and `Streams` were. I rebuilt the old config to check, and `PeerDefault.Memory` comes out as `MaxInt64`. That means any of the peers already let in under `max_peers` could still grow their QUIC receive-window memory without bound.`PeerDefault` now inherits everything from `AutoScale()` instead of just those two fields.

On the `System.Conns` formula — still think `max_peers × max_peer_streams` reads as a worst-case ceiling, but with QUIC multiplexing the realistic number is closer to `max_peers × 1`. The fix currently uses AutoScale's `PeerDefault.Conns` (8) as a middle ground. Curious which direction you'd prefer:

1. `max_peers × PeerDefault.Conns` — what's there now
2. `max_peers × small_constant` (2-4, covering active + one reconnect)
3. A new `max_system_conns` config field for explicit control

swift1337:
@JayT106 I think this is a good middle ground:

> max_peers × small_constant (2-4, covering active + one reconnect)

JayT106:
> @JayT106 I think this is a good middle ground:
> 
> > max_peers × small_constant (2-4, covering active + one reconnect)

updated!

lp2p/host_test.go
```diff
@@ -361,24 +361,34 @@ func TestResourceManager(t *testing.T) {
...
 		t.Logf("systemLimits: %T: %+v", systemLimits, systemLimits)
+		t.Logf("transientLimits: %T: %+v", transientLimits, transientLimits)
 		t.Logf("peerLimits: %T: %+v", peerLimits, peerLimits)
 		t.Logf("serviceLimits(identify): %T: %+v", serviceLimits, serviceLimits)
 		t.Logf("peerPingLimits: %T: %+v", peerPingLimits, peerPingLimits)
 
-		// no limits on "system" scope...
-		require.Equal(t, math.MaxInt64, systemLimits.GetStreamTotalLimit())
+		// system scope must be finite (backed by AutoScale, not InfiniteLimits)
+		require.Less(t, systemLimits.GetStreamTotalLimit(), 1_000_000)
```
swift1337, line 378:
```
 go test ./lp2p/... -run ^TestResourceManager$/^custom$ -v -count=1
=== RUN   TestResourceManager
=== RUN   TestResourceManager/custom
    host_test.go:371: systemLimits: *rcmgr.BaseLimit: &{Streams:14336 StreamsInbound:7168 StreamsOutbound:14336 Conns:40 ConnsInbound:40 ConnsOutbound:40 FD:92160 Memory:6576668672}
    host_test.go:372: transientLimits: *rcmgr.BaseLimit: &{Streams:1792 StreamsInbound:896 StreamsOutbound:1792 Conns:256 ConnsInbound:128 ConnsOutbound:256 FD:23040 Memory:838860800}
    host_test.go:373: peerLimits: *rcmgr.BaseLimit: &{Streams:500 StreamsInbound:500 StreamsOutbound:500 Conns:8 ConnsInbound:8 ConnsOutbound:8 FD:1440 Memory:872415232}
    host_test.go:374: serviceLimits(identify): *rcmgr.BaseLimit: &{Streams:896 StreamsInbound:448 StreamsOutbound:448 Conns:0 ConnsInbound:0 ConnsOutbound:0 FD:0 Memory:29360128}
    host_test.go:375: peerPingLimits: *rcmgr.BaseLimit: &{Streams:4 StreamsInbound:2 StreamsOutbound:3 Conns:0 ConnsInbound:0 ConnsOutbound:0 FD:0 Memory:8590458880}
--- PASS: TestResourceManager (0.00s)
    --- PASS: TestResourceManager/custom (0.00s)

```

Regardless of `cfg.Limits.MaxPeerStreams = 500 // or 50 or 5000` total N of streams is now capped at `14336` (on my machine), which is wrong for custom mode.

swift1337 (changes requested):

PR #5959, opened 2026-07-03
fix(blocksync): tolerate late BlockResponse from honest peers after switching to consensus

JayT106:
### Problem

`FilterMsgBytes` returned an error for any `BlockResponse` once the block pool had stopped (`!r.pool.IsRunning()`). That error panicked inside `p2p/peer.go`'s `onReceive` and was recovered into `StopPeerForError` — a hard disconnect. Because requests for the last blocks are typically still in-flight when the catch-up ticker fires and `pool.Stop()` is called, both request targets (primary and secondary peer) were disconnected on every normal sync completion, exactly when the node needs them for consensus gossip.

Root cause introduced by #5860 (`FilterMsgBytes`). Not covered by #5950, which handles the second-peer race while the pool is still running.

### Fix

Split the single `||` guard into two branches:

- `!r.enabled.Load()` — node never ran blocksync; any `BlockResponse` is genuinely unsolicited → reject (unchanged behaviour)
- `!r.pool.IsRunning()` — pool stopped for consensus switch; requests we sent before the transition are still in-flight; the peers are honest → allow, but still run `validateMaxVotes` on the response

The `AddBlock` call that follows returns `"already committed block"` and logs it; no peer state is mutated and no disconnect occurs.

`pool.sendError` is already a no-op once the pool is stopped (it guards on `!pool.IsRunning()`), so the pre-fix punishment for genuinely out-of-window responses in this window was already silently dropped the moment `poolEventsRoutine` exited.

The post-stop branch skips the `HasPendingRequestFrom` peer-identity check (any peer, not just the one we solicited, may legitimately answer a stale request) but still enforces `validateMaxVotes` — once the pool stops, any connected peer can reach this path, and the commit-signature cap exists to prevent a heap-amplification DoS regardless of sender. Caught by review on this PR; added a dedicated regression test (`rejects oversized BlockResponse after pool stops for consensus switch`) so an oversized response is still rejected in this window.

---

#### PR checklist

- [x] Tests written/updated


dianab-cl:
@greptile review

dianab-cl (changes requested):
Please check and address Greptile findings

dianab-cl:
@greptile review

dianab-cl:
lgtm

dianab-cl (changes requested):
To fix conflicts

dianab-cl:
@greptile review

swift1337 (approved):

dianab-cl (approved):

PR #5975, opened 2026-07-13
fix(blocksync): don't treat already-committed block from slow peer as an error

songgaoye:
## Summary
With dual requests for blocks near the pool's height, a slow second peer can
deliver a block already committed via PopRequest. AddBlock returned a generic
error for this, causing a spurious "Failed to add block" log and flaking
TestBlockPoolBasic.

Add a sentinel `ErrAlreadyCommittedBlock` so callers can recognize this benign
race via`errors.Is`: the peer is not punished, and the reactor logs it at Debug.

---

#### PR checklist

- [x] Tests written/updated


blocksync/reactor.go
```diff
@@ -343,7 +343,11 @@ func (r *Reactor) handlePeerResponse(msg *bcproto.BlockResponse, src p2p.Peer) {
 	}
 
 	if err := r.pool.AddBlock(src.ID(), bi, extCommit, msg.Block.Size()); err != nil {
-		r.Logger.Error("Failed to add block", "peer", src, "err", err)
+		if errors.Is(err, ErrAlreadyCommittedBlock) {
```
swift1337, line 346: We should drop this peer if it attempts to submit an unexpectedly old block, e.g., ban it when `latestBlock.height - block.height > N`, where N is, for example, 16.

songgaoye: The concern about repeated stale responses makes sense, but I’m not sure a fixed 16-block gap is a safe ban criterion. A legitimately requested response from the slower peer in a dual request may arrive after blocksync has advanced by more than 16 blocks. I’d prefer tracking requests by (peerID, height) and penalizing unsolicited or repeated stale responses. Since this PR only changes the logging of a benign late response, could we handle that separately?

songgaoye:
add CI failure at: https://github.com/cometbft/cometbft/actions/runs/35451016029/job/105918053951

PR #5989, opened 2026-07-22
chore: Bump max tx size default

Eric-Warehime:
---

Bumps max tx bytes to account for the potential of larger txs due to the need to incorporate validator signatures such as IBC txns or evidence.

In terms of picking a sane default, a chain w/ 100 validators all using mldsa65 signatures a `MsgSubmitMisbehaviour` which carries 2 full block headers would take ~1.5MB. 4MB here gives us plenty of overhead.


dianab-cl (approved):

dianab-cl:
To add CHANGELOG entry, as greptile says

PR #5990, opened 2026-07-22
feat: Add unmarhsal json to secp256k1eth

Eric-Warehime:
---

Adds json decoder to secp256k1eth key types which will end up being called during genesis validation. This change will result in genesis validation errors as opposed to panics during startup on a genesis that otherwise would not fail validation.


Eric-Warehime:
@greptile review

crypto/secp256k1eth/key.go
```diff
@@ -138,6 +139,32 @@ func (privKey PrivKey) Sign(msg []byte) ([]byte, error) {
...
+	if len(rawBytes) != PrivKeySize {
+		return fmt.Errorf(
+			"secp256k1eth: invalid private key size, expected %d bytes, got %d",
+			PrivKeySize, len(rawBytes),
+		)
+	}
+	var d secp256k1.ModNScalar
+	overflow := d.SetByteSlice(rawBytes)
+	zero := d.IsZero()
+	d.Zero()
```
mattac21: this feels like dead code?

mattac21 (approved):

crypto/secp256k1eth/key.go
Eric-Warehime: Defensive coding--zeroing out memory where the privkey bytes were before returning. Agree it's kinda useless here.

PR #6024, opened 2026-08-05
rpc/client/http: avoid blocking event dispatch on unbuffered subscriptions

### LINKED ISSUE #6023: rpc/client/http: unbuffered subscription can stall WebSocket event dispatch
### Setup

Current `main` at `5bba72c14`.

**Environment**:

This issue is independent of the operating system.

### What happened?

`WSEvents.Subscribe` allows callers to specify the capacity of the returned
event channel:

```go
Subscribe(ctx, subscriber, query, outCapacity...)
```

Passing 0 creates an unbuffered channel. When eventListener dispatches an event to that channel, it performs a blocking send while holding w.mtx.RLock():

```go
w.mtx.RLock()
if out, ok := w.subscriptions[result.Query]; ok {
    if cap(out) == 0 {
        out <- *result
    } else {
        select {
        case out <- *result:
        default:
            w.Logger.Error(
                "wanted to publish ResultEvent, but out channel is full",
                "result", result,
                "query", result.Query,
            )
        }
    }
}
w.mtx.RUnlock()
```

If no receiver is ready, the send blocks indefinitely.

This stalls the single WebSocket event listener, so events for all other subscriptions on the same client stop being dispatched. Operations that need the subscription map's write lock, including Subscribe, Unsubscribe, and UnsubscribeAll, can also no longer complete their local map updates.

Because the goroutine is blocked inside the send, it also cannot return to the outer select to observe w.Quit().

One subscription whose consumer is not ready should not indefinitely stall event delivery and subscription management for the entire WebSocket client.

The package documentation already describes event delivery as best-effort, and buffered subscriptions currently drop an event when their channel is full.

### How to reproduce it

1. Start an HTTP RPC client.
2.Create a subscription with outCapacity set to 0.
3.Do not start a receiver for the returned channel.
4.Deliver an event matching that subscription.
5.Attempt to unsubscribe or create another subscription.
6.Observe that the event listener remains blocked and the subscription map
cannot be updated.

zjuzhongwen:
Closes https://github.com/cometbft/cometbft/issues/6023

This PR prevents an unbuffered subscription from indefinitely blocking the shared WebSocket event listener.

`Subscribe(..., 0)` creates an unbuffered event channel. Previously, `eventListener` performed a blocking send to that channel while holding the subscription map's read lock. If the subscriber had no active receiver, this stalled event delivery for every subscription on the same client and prevented subscription map updates from acquiring the write lock.

The event dispatch now uses a non-blocking send for both buffered and unbuffered subscription channels. If the channel is not ready, the event is dropped and an error is logged, consistent with the package's best-effort delivery model.

Logging is performed after releasing the subscription lock.

Tests cover an unread unbuffered subscription and verify that:

- the event listener remains responsive;
- unsubscription can complete;
- subsequent events can still be delivered to another subscription.

#### PR checklist

- [x] Tests written/updated
- [ ] Updated relevant documentation and code comments, if required

### DIFF
--- a/rpc/client/http/http.go
+++ b/rpc/client/http/http.go
@@ -652,6 +652,9 @@ func (w *WSEvents) OnStop() {
 //
 // Channel is never closed to prevent clients from seeing an erroneous event.
 //
+// Event delivery is best-effort. An event is dropped if the channel is not
+// ready, including when an unbuffered channel has no waiting receiver.
+//
 // It returns an error if WSEvents is not running.
 func (w *WSEvents) Subscribe(ctx context.Context, _, query string,
 	outCapacity ...int,
@@ -741,6 +744,24 @@ func isErrAlreadySubscribed(err error) bool {
 	return strings.Contains(err.Error(), cmtpubsub.ErrAlreadySubscribed.Error())
 }
 
+func (w *WSEvents) publishEvent(result *ctypes.ResultEvent) {
+	dropped := false
+
+	w.mtx.RLock()
+	if out, ok := w.subscriptions[result.Query]; ok {
+		select {
+		case out <- *result:
+		default:
+			dropped = true
+		}
+	}
+	w.mtx.RUnlock()
+
+	if dropped {
+		w.Logger.Error("wanted to publish ResultEvent, but out channel is not ready", "result", result, "query", result.Query)
+	}
+}
+
 func (w *WSEvents) eventListener() {
 	for {
 		select {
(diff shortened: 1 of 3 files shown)


zjuzhongwen:
Hi @marbar3778,

Could you please take a look at this PR when you have a chance?

This change touches the RPC websocket subscription behavior and fixes a potential deadlock caused by unbuffered subscriptions blocking the event dispatcher.

Your feedback on the event delivery semantics would be highly appreciated.

Thanks!

PR #6039, opened 2026-08-24
perf(state): use fast validator loading for BlockExecutor commit info

songgaoye:
## Summary

`LoadValidatorsFast` avoids reconstructing proposer priorities when callers only need validator addresses and voting powers. It was introduced in #5857 for replay, but the normal `BlockExecutor` commit-info paths still use `LoadValidators`.

This PR uses the fast loader for:

- `ProcessProposal` / `buildLastCommitInfo`
- `CreateProposalBlock` / `buildExtendedCommitInfo`

Stores without `LoadValidatorsFast` continue to fall back to `LoadValidators`.

This is safe because `CommitInfo` and `ExtendedCommitInfo` only use validator addresses and voting powers. They do not use proposer priorities.

Related to #1693 and #5857.

## Benchmark

The benchmark exercises the normal `BlockExecutor.ProcessProposal` path with 100 validators. Each iteration forces the validator-cache miss that occurs at a new block height.

Apple M4, `-benchtime=200ms -count=3`:

| Height after checkpoint | Before | After | Improvement |
| ---: | ---: | ---: | ---: |
| 100 | 54.5–57.1 µs | 26.2–27.8 µs | ~2.1x |
| 10,000 | 2.92–2.96 ms | 25.9–26.1 µs | ~113x |
| 99,999 | 29.1–29.5 ms | 26.1–26.4 µs | ~1,110x |

These results measure the affected `ProcessProposal` path, not overall node throughput.

## Testing

- Added tests confirming both commit-info paths use `LoadValidatorsFast`
- Verified fallback compatibility with stores that only implement `LoadValidators`
- `go test ./state -count=1`
- Targeted tests with `-race`

#### PR checklist

- [x] Tests written/updated
- [x] Updated relevant code comments
- [ ] Documentation update (not required)


dhfang:
@greptile review

dhfang (approved):

state/execution.go
```diff
@@ -576,23 +576,23 @@ func buildLastCommitInfoFromStore(block *types.Block, store Store, initialHeight
 	return BuildLastCommitInfo(block, lastValSet, initialHeight)
 }
 
-func (blockExec *BlockExecutor) loadValidatorsCached(height int64) (*types.ValidatorSet, error) {
-	if entry := blockExec.lastLoadedValidators.Load(); entry != nil && entry.height == height {
+func (blockExec *BlockExecutor) loadValidatorsForCommitInfoCached(height int64) (*types.ValidatorSet, error) {
```
swift1337, line 580:
```suggestion
// note: set.Validators[i].ProposerPriority and set.Proposer are stale to avoid extra computations
func (blockExec *BlockExecutor) loadValidatorsForCommitInfoCached(height int64) (*types.ValidatorSet, error) {
```

swift1337 (approved):

PR #6044, opened 2026-09-01
fix(state): kv indexers no longer silently drop a non-numeric tx.height/block.height equality

gomesalexandre:
## What it says on the box

The KV tx and block indexers silently DROP a `tx.height`/`block.height` equality condition whenever its query argument isn't a valid number, instead of correctly treating it as unsatisfiable. In a compound `AND` query, this widens `/tx_search` and `/block_search` results to rows the query explicitly excludes.

## The bug

`dedupHeight()` in `state/txindex/kv/utils.go` and its identical-shape sibling in `state/indexer/block/kv/util.go`:

```go
hFloat := c.Arg.Number()
if hFloat != nil {
    ...
    heightCondition = append(heightCondition, c)
}
// no else -- when hFloat is nil the condition is added to NEITHER
// heightCondition NOR dedupConditions, and simply vanishes
```

An unsatisfiable filter (e.g. `tx.height = 'something'`) should cause the whole query to match nothing. Instead it drops out of consideration entirely, so a compound query like `account.number = 1 AND tx.height = 'something'` returns whatever the first condition matched on its own.

This only manifests when the broken height condition is ANDed with another condition that independently matches something -- `tx.height = 'abc'` alone (no other condition) already correctly returns 0, which is why casual single-condition probing misses it.

## Repro

```
q="account.number >= 1 AND account.number = 'something'"  -- control, correctly returns 0
q="account.number >= 1 AND tx.height     = 'something'"  -- BUG, returns 1
q="end_event.foo = 100 AND block.height  = 'nope'"       -- BUG, returns a real height
q="end_event.foo = 100 AND block.height  = 3"            -- control, works correctly
```

Full test output (before/after) is in the receipts section below.

## The fix

Route a non-numeric height-equality condition through the normal per-condition `match()` path instead of dropping it. Both indexers already index `tx.height`/`block.height` as real, queryable keys (confirmed by reading the write path -- `types/event_bus.go:187` for tx.height, `heightKey()` for block.height), so a string-valued search key for an equality condition never collides with a real height-keyed entry and the query correctly returns zero matches -- exactly the same mechanism that already makes `account.number = 'something'` correctly return 0 today.

The numeric-ness check now runs unconditionally, before the existing `found`/`heightRangeExists` dedup logic, not just when neither is already true. An earlier version of this fix (adversarially reviewed before opening) only handled the *first* occurrence of a non-numeric height condition -- a *later* duplicate (`tx.height = 1 AND tx.height = 'something'`) still fell through the pre-existing `continue` guard and got silently dropped. Restructuring the check to run first, regardless of ordering, closes that gap too.

## Provenance note

`v0.37.x`'s equivalent code has a different, worse defect on the same input -- an unchecked type assertion that panics. The `if hFloat != nil` guard visible on `main` appears to have been added specifically to stop that panic, with the wrong recovery behavior picked (silently drop the condition) instead of the correct one (treat it as an unsatisfiable match).

## Honest scoping

This is a read-only RPC query-widening bug -- not a crash, not unbounded resource consumption, not a consensus-path issue. No real client emitting a query shaped this way was directly observed; what's proven is the code-level defect and its mechanical trigger condition.

## Receipts

New regression tests, run against the actual pre-fix source (confirmed via `git stash`) and the fixed source:

```
$ go test ./state/txindex/kv/... -run TestTxSearch -v 2>&1 | grep -E "PASS|FAIL" | grep -i "something"
--- unfixed ---
    --- FAIL: TestTxSearch/account.number_=_1_AND_tx.height_=_'something'
    --- FAIL: TestTxSearch/tx.height_=_'something'_AND_account.number_=_1
    --- FAIL: TestTxSearch/tx.height_=_1_AND_tx.height_=_'something'
--- fixed ---
    --- PASS: TestTxSearch/account.number_=_1_AND_tx.height_=_'something'
    --- PASS: TestTxSearch/tx.height_=_'something'_AND_account.number_=_1
    --- PASS: TestTxSearch/tx.height_=_1_AND_tx.height_=_'something'

$ go test ./state/indexer/block/kv/... -run TestBlockIndexer$ -v 2>&1 | grep -E "PASS|FAIL" | grep -i nope
--- unfixed ---
    --- FAIL: TestBlockIndexer/end_event.foo_=_100_AND_block.height_=_'nope'
    --- FAIL: TestBlockIndexer/block.height_=_1_AND_block.height_=_'nope'
--- fixed ---
    --- PASS: TestBlockIndexer/end_event.foo_=_100_AND_block.height_=_'nope'
    --- PASS: TestBlockIndexer/block.height_=_1_AND_block.height_=_'nope'
```

Full existing suites, zero regressions:

```
$ go test ./state/txindex/... ./state/indexer/...
ok  	github.com/cometbft/cometbft/state/txindex	0.853s
ok  	github.com/cometbft/cometbft/state/txindex/kv	0.343s
ok  	github.com/cometbft/cometbft/state/indexer/block/kv	0.352s
ok  	github.com/cometbft/cometbft/state/indexer/sink/psql	2.617s
```

`gofmt -l` and `go vet` clean on all changed files.

## Review

Adversarially reviewed with a second model (run read-only against this diff) before opening -- it correctly caught the order-dependent duplicate-condition gap described above, which is why the fix is structured the way it is rather than the more naive version I started with.

No merge, no self-approval -- this is a draft-then-ready PR per my usual workflow.


state/txindex/kv/kv_test.go
```diff
@@ -111,6 +111,15 @@ func TestTxSearch(t *testing.T) {
 		// search by range (upper bound)
 		{"account.number <= 5", 1},
 		{"account.number <= 1", 1},
+		// search using a non-numeric tx.height equality: the condition is
+		// unsatisfiable (no real height can equal a non-numeric literal) and
+		// must not be silently dropped, which would otherwise leave the
+		// other AND-ed condition to match on its own
+		{"account.number = 1 AND tx.height = 'something'", 0},
```
swift1337: We should simply fail to compile such queries, ensuring `.height` is `uint64`

swift1337 (changes requested):

state/txindex/kv/kv_test.go
gomesalexandre: Pushed a revised fix (f224e55).

Went with an equivalent to "fail to compile" that's actually enforceable given how the query grammar is built: `.height` is parsed through `libs/pubsub/query/syntax`, a shared, tag-agnostic query DSL with no per-tag type system (used for way more than height conditions), so a real parser-level `uint64` constraint would mean expanding that shared package specifically for this one tag - more invasive than "simply" implies.

Instead, `Search()` now validates height conditions up front (both `state/txindex/kv` and `state/indexer/block/kv` - `Search` already returns `(..., error)` in both) and rejects the whole query immediately if `tx.height`/`block.height` is compared against anything non-numeric, e.g. `tx.height = 'something'` now returns an error instead of running at all. `dedupHeight()` is reverted back to its original form since the malformed case never reaches it anymore.

Updated `kv_test.go` in both packages so the three/two previously-affected cases now assert an error instead of an empty result. Confirmed red against the old behavior and green with the fix; full `state/txindex/...` and `state/indexer/...` suites pass (excluding the docker-dependent psql sink tests, unrelated to this change and no docker daemon available here to run them).

state/indexer/block/kv/util.go
```diff
@@ -147,6 +147,22 @@ func parseEventSeqFromEventKey(key []byte) (int64, error) {
...
 
+// validateHeightConditions rejects a query outright if it contains a
+// block.height condition whose argument is not a valid numeric height (e.g.
+// a quoted string literal or a malformed number). block.height is a uint64
+// block height, so a non-numeric comparison can never match anything real --
+// per review feedback (PR #6044), that should fail the query up front with a
+// clear error, rather than being handled deep inside dedupHeight as an
+// always-false condition.
+func validateHeightConditions(conditions []syntax.Condition) error {
+	for _, c := range conditions {
```
swift1337: 1) You are iterating over all parsed conditions. Height validation should be incorporated into query parsing (see https://github.com/cometbft/cometbft/blob/main/libs/pubsub/query/syntax/parser.go#L162).
2) Please note that low-effort AI PRs will be closed, as yours contains substantial slop.

state/txindex/kv/kv_test.go
gomesalexandre: Moved: `tx.height`/`block.height` equality now fails to parse in `libs/pubsub/query/syntax` parseCond (50042a1), not in the kv indexers.

Can't import `types` there for the reserved key constants (`types` -> `libs/pubsub/query` -> `syntax`, would cycle), so `tx.height`/`block.height` are local string consts in the syntax package mirroring `types.TxHeightKey`/`types.BlockHeightKey`.

Used `strconv.ParseUint` instead of `Arg.Number() == nil`, so it also now catches `tx.height = 5.5` and overflow, which the old check let through to silently truncate via `dedupHeight`'s `hFloat.Int64()`.

`validateHeightConditions` and its call sites in `state/txindex/kv` and `state/indexer/block/kv` are gone - `Query.Syntax()` conditions only ever come from `syntax.Parse`, nothing constructs `syntax.Condition` directly, so the post-parse check was dead weight. Those two packages are back to their pre-PR shape.

swift1337 (approved):

swift1337:
@gomesalexandre please sign your commits (see CI)

gomesalexandre:
@swift1337 fixed, thanks for flagging. Root cause: `cab709f9` was authored with `contact@alexandregomes.fr`, which doesn't match my GPG key's UID (`17035424+gomesalexandre@users.noreply.github.com`) - so it was cryptographically signed and verified fine locally, but GitHub's signature check requires the commit author email to match a UID on the signing key, and correctly flagged it as unverified.

Amended that commit's author to the matching email and re-signed (`git commit --amend --author=... -S`), rebased the other two on top, force-pushed. All 3 commits now show `Good signature` locally against the matching UID.

The re-run is currently sitting in `action_required` pending approval on the fork workflow run - happy to wait for that, or feel free to approve it whenever's convenient.

Issue #4054, opened 2024-09-10
CometBFT should notify upon fatal errors and ensure node termination


lucaslopezf:
### Setup

**CometBFT version** (use `cometbft version` or `git rev-parse --verify HEAD` if installed from source):
- v0.34.1

**ABCI app** (name for built-in, URL for self-written if it's publicly available):
- Cosmos-SDK

**Environment**:
- **OS** (e.g. from /etc/os-release): Ubuntu 20.04
- **Install tools**: Docker, Kubernetes
- **Others**: N/A

### What happened?

When there is a failure in the consensus layer, such as during an upgrade that requires a binary update, CometBFT stops the blockchain, but Cosmos-SDK is never informed of this. This results in the node continuing to respond to queries with the blockchain halted.

In the case of a panic, the code enters this recover: [state.go#L812](https://github.com/cometbft/cometbft/blob/a170d312b28a429d1fecc32cd94d91b01c5ad8f7/internal/consensus/state.go#L812). As the function comment states, the intention is to stop the chain. However, there is no communication to Cosmos-SDK about this, and the `isRunning()` method ([service.go#L45](https://github.com/cometbft/cometbft/blob/5e4ab3c747558cdea85cdd3fc320fe6b5c363930/libs/service/service.go#L45)) still returns true.

This results in an inconsistency in the application, as described in the following issue:

- [Cosmos SDK Issue #20745](https://github.com/cosmos/cosmos-sdk/issues/20745)

### What did you expect to happen?

I expected CometBFT to notify Cosmos-SDK upon encountering a fatal error and ensure that the node terminates properly, reflecting the halted state accurately.

### How to reproduce it

1. Start a new node with `simd` in Cosmos.
2. Submit an upgrade proposal.
3. Vote on the proposal.
4. Once the proposal period has passed, observe the issue.

### Logs

```bash
[...]

PR #6050, opened 2026-09-11
fix(node): defer eventBus.Stop to final step of OnStop lifecycle (#4054)

Tyagiquamar:
## Problem
During node teardown or fatal error shutdown, \ventBus\ was closed before background reactors and service listeners finished cleaning up, causing event subscribers to miss shutdown events and potential channel closed panics.

## Root Cause
\
.eventBus.Stop()\ was executed as the very first step of \Node.OnStop()\, before stopping reactors and RPC listeners.

## Fix
Moved \
.eventBus.Stop()\ to the final step of \Node.OnStop()\ after all reactors, services, and listeners have completed teardown.

## Verification
Ran \go test -v ./node -run TestNodeStartStop\ green.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -737,9 +737,6 @@ func (n *Node) OnStop() {
 	n.Logger.Info("Stopping Node")
 
 	// first stop the non-reactor services
-	if err := n.eventBus.Stop(); err != nil {
-		n.Logger.Error("Error closing eventBus", "err", err)
-	}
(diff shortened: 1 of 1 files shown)


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

PR #6067, opened 2026-09-19
fix(libs/autofile): unregister SIGHUP channel before closing it

kriss39:
`AutoFile.Close` closed `af.hupc` while the channel was still registered with `signal.Notify`. The runtime keeps delivering `SIGHUP` to every registered channel, so the next `SIGHUP` after `Close` panics the signal-dispatch goroutine with `send on closed channel` and terminates the process. `AutoFile` exists specifically for logrotate-style `SIGHUP` handling, so repeated `SIGHUP`s are the expected operating mode. The same defect makes `TestSIGHUP` crash the package binary under `go test -count=2`.

This calls `signal.Stop` before closing the channel. `TestSIGHUPAfterClose` opens and closes an `AutoFile`, then sends `SIGHUP` to the process; on `main` it panics, with the fix the signal is delivered only to the test's own channel.

---

#### PR checklist

- [x] Tests written/updated


dianab-cl:
@greptile review

dianab-cl (changes requested):
Please sign the commit(s)

dianab-cl (approved):

PR #6071, opened 2026-09-19
fix(abci): fail in-flight socket client requests on graceful stop

Bruce039:
## Problem

Stopping the ABCI socket client gracefully (`Stop()`, not `StopForError`) leaves `cli.err == nil`, because `StopForError` bails out when the service is already not running. `flushQueue()` then wakes every in-flight `ReqRes` without a response, and the sync wrappers return

```go
return reqRes.Response.GetCheckTx(), cli.Error()   // (nil, nil)
```

Node shutdown goes through `proxy.multiAppConn.stopAllClients()` with no draining, so anything mid-call at that moment gets `(nil, nil)`. With `mempool.type = "app"` that is `AppMempool.CheckTx`, and `BroadcastTxSync` / `BroadcastTxCommit` dereference the nil response.

## Change

- `flushQueue` records `ErrClientStopped` if no error is set yet; `StopForError`'s error still wins
- new sentinel `abcicli.ErrClientStopped`
- test: `CheckTx` against a slow app, `Stop()` while it is pending → expects `ErrClientStopped`, nil response (returns `(nil, nil)` on `main`)

## Relation to #6054

#6054 fixes the double-`Done()` panic in the same loop and adds its own `ErrClientStopped` for requests queued *after* stop. This one covers requests that were already sent. Happy to rebase onto #6054 and reuse its sentinel if it merges first.

## Checklist

- [x] tests
- [x] changelog
- [ ] docs (n/a)


songgaoye:
update https://github.com/cometbft/cometbft/pull/6054/changes/23b427d57af7316bb983adb08e8e2ebabadddebb. Thank you!

PR #6074, opened 2026-09-19
fix(lp2p): derive System.Conns from max_peers in custom limits mode

Bruce039:
## Problem

In `ResourceManagerFromConfig`, custom mode:

```go
// 2.1 limit max system connections to (max conns per peer * max peers)
limits.System.Conns = partialDefaults.PeerDefault.Conns * maxPeerStreams
```

The comment says `max peers`, the code multiplies by `max_peer_streams`. `max_peers` is not used at all for this limit, so the system connection cap scales with a setting that has nothing to do with connections. Concretely, with `max_peers = 10`: `max_peer_streams = 5` → 40 conns, `max_peer_streams = 500` → 4000 conns.

This was part of #5957, which went stale; in that thread @swift1337 suggested `max_peers × small constant (2–4)` as the right shape, since peers normally share one multiplexed QUIC connection. Credit to @JayT106 for the original report.

## Change

- `limits.System.Conns = max_peers * 4` (constant `systemConnsPerPeer`; 4 = one active connection plus headroom for a reconnect while the old one drains)
- nothing else from #5957 is included — only this line
- `TestResourceManager/custom` now asserts the system conn limit equals `max_peers * 4` and does not move when `max_peer_streams` is multiplied by 10 (on `main`: expected 40, actual 400)

`max_peers` is already required to be > 0 in custom mode by `LibP2PLimits.ValidateBasic`, so the limit can't collapse to zero.

## Checklist

- [x] tests
- [x] changelog
- [ ] docs (n/a)


dianab-cl:
@greptile review

dianab-cl (changes requested):
Please sign the commit(s)

Bruce039:
Done — the commit is signed and the branch is rebased on `main` (563fadb6, same content; the changelog entry now sits next to the #6070 line that landed in the meantime). Could you take another look?

dianab-cl (approved):

swift1337 (approved):

PR #6077, opened 2026-09-19
fix(rpc): cancel the per-event write context in Subscribe after each write

0xShadowX:
In `Environment.Subscribe` the goroutine that forwards events to the websocket does, for every event:

```go
writeCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
defer cancel()
if err := ctx.WSConn.WriteRPCResponse(writeCtx, resp); err != nil {
```

`defer` inside the `for` loop doesn't run per iteration; it runs when the goroutine returns, i.e. when the subscription is canceled. Until then every delivered event keeps its context and its 10s timer alive. A client subscribed to `tm.event='Tx'` on a busy chain accumulates one of these per transaction for the lifetime of the connection.

The fix is to call `cancel()` right after `WriteRPCResponse` returns. Nothing else changes.

The added test subscribes with a stub `WSRPCConnection` that records the context passed to each write, publishes two events and asserts the first write's context is already canceled; on `main` it is still live.

This is the same defect PR #5808 pointed at earlier this year; that PR was closed by its author without review, so re-submitting with a test.

- [x] Tests written/updated


swift1337:
low-effort AI PR

